PR #145, opened 2021-06-15
Clarify reference to Debian Rodete in docs

gidoBOSSftw5731:
Debian Rodete (ROlling DEbian TEsting) is an internal Google codename for what is essentially Debian Testing, and therefore there is no public release of it. To lessen any confusion, it seems advisable to point external users to the closest publicly available release while also leaving in Rodete if any internal Googlers are reading the docs and do not understand the origin of Rodete.

For more details please see

### DIFF
--- a/docs/contributing/build-instructions.md
+++ b/docs/contributing/build-instructions.md
@@ -154,7 +154,7 @@ Android.bp file. If you are adding a new target, add a new entry to the
 
 ## Supported platforms
 
-**Linux desktop** (Debian Rodete)
+**Linux desktop** (Debian Testing/Rodete)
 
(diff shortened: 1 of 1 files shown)


gidoBOSSftw5731:
@googlebot I signed it!

primiano:
Ah makes sense .Thanks for the fix. 
Can I only ask you to send this on AOSP gerrit.
As the pull request template suggest (it's still copied there in your text above :) ), patches must go via AOSP, see https://perfetto.dev/docs/contributing/getting-started for instructions.
GitHub is a read-only mirror of AOSP

gidoBOSSftw5731:
Ah, thanks for the mention, I'll add it there momentarily.

PR #180, opened 2021-10-01
Update trace-processor.md

yuminchang0830:
Use SELECT FROM for counter table  & join process_counter_track so should be slice

For more details please see


primiano:
As the template explains we don't take PRs from github. Please use AOSP.

Thanks,
Primiano

PR #402, opened 2022-12-08
ui: make pan and zoom independent of keyboard layout

ehermann:
Hello,
I'm using an azerty keyboard, and pan and zoom is impracticable with the current shortcuts.
It can be fixed buy using event code instead of event key as explained in this page. https://www.bram.us/2022/03/31/wasd-controls-on-the-web/


LalitMaganti:
Sorry we don't accept pull requests through GitHub. Please follow the information at https://perfetto.dev/docs/contributing/getting-started#quickstart for sending a patch to Android AOSP Gerrit.

Issue #455, opened 2023-02-25
Inconsistent number of python and shell results


wanglong001:
> The slice id is not stable, it changes as new tracks are added/removed across different versions of perfetto and hence trace processor. VEry likely you are using a different version of TraceProcessor across python API and shell

@primiano 
Not only the ids are different between python and shell, the count of slice table in python (count= 963750) is greater than shell (count=954036). I found duplicated items in python slice table (different ids with same content).

_Originally posted by @wanglong001 in https://github.com/google/perfetto/issues/454#issuecomment-1443530843_

LalitMaganti:
Do you have a trace I can check what's happening with? Without that, this is completely unactionable.

wanglong001:
## I examined the source code and discovered why the results are inconsistent
By default, mmap is used in the shell to read files, and 128ul * 1024 * 1024 bytes are read each time. The default in Python is to read 1 * 1024 * 1024 bytes.
I discovered that the size of each read is related to this problem. This problem won't be activated in my data if each read is less than 3M.


**_I discovered that the length of the TraceEvent field causes this issue._**
I attempted to fix the bug and discovered that the fix was successful under our data. I'll push a pull request; please review it for me.

aMayzner:
We don't support contributing via Github. If you want to contribute please
use https://perfetto.dev/docs/contributing/getting-started.

PR #457, opened 2023-02-27
[bugfix] Correct the issue where the TraceEvent field is too long, resulting in data duplication.

wanglong001:
[bugfix] Correct the issue where the TraceEvent field is too long, resulting in data duplication.


LalitMaganti:
As described at https://github.com/google/perfetto/issues/455, please send a patch through our AOSP contribution process as we do not accept patches on GitHub.

wanglong001:
> please send a patch through our AOSP contribution process as we do not accept patches on GitHub.

Okay， I will send a patch through AOSP contribution process

PR #868, opened 2024-08-24
Update BUILD.gn

omsandippatil:

For more details please see


primiano:
1. as written in the bug template, we don't take PRs from github
2. This change makes no sense, you are removing comments for no good reason

PR #1040, opened 2025-03-14
Github test CL1

primiano:
GitHub test with commits CL1

PS1

### DIFF
--- a/src/tools/busy_threads/busy_threads.cc
+++ b/src/tools/busy_threads/busy_threads.cc
@@ -63,30 +63,30 @@ void PrintUsage(const char* bin_name) {
 #endif
 }
 
-__attribute__((noreturn)) void BusyWait(int64_t tstart,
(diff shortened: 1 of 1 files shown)


src/tools/busy_threads/busy_threads.cc
```diff
@@ -63,30 +63,30 @@ void PrintUsage(const char* bin_name) {
 #endif
 }
 
-__attribute__((noreturn)) void BusyWait(int64_t tstart,
+__attribute__((noreturn)) void BusyWait(int64_t tstart_ns,  // PS1
```
LalitMaganti: Should be PS2

LalitMaganti (commented):
Looks OK

src/tools/busy_threads/busy_threads.cc
LalitMaganti: Test

PR #1060, opened 2025-03-27
ci: move to GitHub Actions

primiano:
Move the CI over to GH Actions


.github/workflows/analyze.yml
```diff
@@ -0,0 +1,84 @@
...
+  # 1. continuous
+  schedule:
+    # Run every 6 hours
+    - cron: "0 */6 * * *"
+
+  # 2. postsubmit
+  push:
+    branches:
+      - main
+      - test
```
LalitMaganti: Do we want to actually run on test?

.github/workflows/analyze.yml
```diff
@@ -0,0 +1,84 @@
...
+
+  # 3. presubmit
+  pull_request:
+    types: [opened, synchronize]
+    branches:
+      - main
+      - test
```
LalitMaganti: As above.

infra/ci/sandbox/Dockerfile
```diff
@@ -55,16 +51,41 @@ RUN set -ex; \
...
+    rm -rf /root/.cache/ /usr/share/man/* /usr/share/doc/*;
+
+WORKDIR /opt/ci
 
+COPY sandbox_entrypoint.sh ./
+COPY artifacts_uploader.py ./
+COPY tmp/config.py ./
+
+# Download GitHub Actions runner.
+# We should upkeep GHAR_VER with the latest version in
```
LalitMaganti, line 75: are we planning on doing this in a followup or just do it as and when necessary?

infra/ci/worker/gce-startup-script.sh
```diff
@@ -15,105 +15,37 @@
...
-  -v $ARTIFACTS_DIR:$ARTIFACTS_DIR \
-  --env SHARED_WORKER_CACHE="$SHARED_WORKER_CACHE" \
-  --env SANDBOX_NETWORK_ARGS="$SANDBOX_NETWORK_ARGS" \
-  --env ARTIFACTS_DIR="$ARTIFACTS_DIR" \
-  --env SANDBOX_TMP="$SANDBOX_TMP" \
-  --env WORKER_HOST="$(hostname)" \
-  --name worker-$i \
-  --hostname worker-$i \
+  --name worker \
+  --privileged \
```
LalitMaganti, line 45: Do we need this? Seems like this contradicts the comment above about not wanting this.

infra/ci/worker/sandbox_runner.py
```diff
@@ -0,0 +1,131 @@
...
+  # for a downgraded service account that we pass to the sandbox. The sandbox
+  # service account is allowed only storage object creation for untrusted CI
+  # artifacts.
+  sandbox_svc_token = subprocess.check_output([
+      'gcloud',
+      'auth',
+      'application-default',
+      'print-access-token',
+      '--impersonate-service-account=%s' % SANDBOX_SVC_ACCOUNT,
+      # TODO(primiano): uncomment below after b/406184705 is fixed.
```
LalitMaganti, line 57: This is fixed.

LalitMaganti (approved):
LGTM % nits

.github/workflows/analyze.yml
primiano: Ah fair point. I was using it only for testing :) removed

infra/ci/sandbox/Dockerfile
primiano: Yeah this needs to be some upkeep we do once/twice per year.
Can't think of a better option tbh

infra/ci/worker/gce-startup-script.sh
primiano: you mean the --privilege?
yeah the worker (i would rename it, but it's a pain) is really the root image. it's a docker container purely to make deployment and testing easy, but it's by all itnent and purposes as privileged as GCE.
it's just a way to make it easier to update the image (managing a raw linux image on GCE is a majestic pain)

infra/ci/worker/sandbox_runner.py
primiano: Unfortunately seems to have not propagated yet:

root@gh-us-west1-7svl:/opt/ci# gcloud auth application-default print-access-token --impersonate-service-account=gce-ci-sandbox@perfetto-ci.iam.gserviceaccount.com --lifetime=21600
ERROR: (gcloud.auth.application-default.print-access-token) There was a problem refreshing your current auth tokens: ('Unable to acquire impersonated credentials', '{\n  "error": {\n    "code": 400,\n    "message": "Access tokens for service account gce-ci-sandbox@perfetto-ci.iam.gserviceaccount.com can not have lifetimes of more than 1 hour. Use the Org Policy Constraint constraints/iam.allowServiceAccountCredentialLifetimeExtension to allow up to 12 hours.",\n    "status": "INVALID_ARGUMENT"\n  }\n}\n')

PR #1061, opened 2025-03-27
tprobes: debuggable android builds: optionally serialise bad pages

rsavitski:
To assist internal debugging. The best way would've been to retain the page until end-of-trace, as otherwise we'll miss cases that rotate out of ring buffer modes. However that proved to be too much plumbing for a debugging feature (cpu_reader -> OnFtraceWritten -> FtraceDataSource, or static storage + accessors).

I can be convinced to change things, but this feels like the minimal addition.

Bug: 401855064


src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -158,6 +158,14 @@ void WriteAndSetParseError(CpuReader::Bundler* bundler,
   proto->set_status(status);
 }
 
+void SerialiseOffendingPage([[maybe_unused]] CpuReader::Bundler* bundler,
+                            [[maybe_unused]] const uint8_t* page,
+                            [[maybe_unused]] size_t size) {
+#if PERFETTO_BUILDFLAG(PERFETTO_ANDROID_USERDEBUG_BUILD)
```
ddiproietto: I would remove the USERDEBUG only limitation.

On user build, shell already has access to the raw ftrace pages anyway. For non shell (i.e. field, dev telemetry tracing api), we can disable this through the config. 

If you disagree, at least, let's leave this enabled on non android.

ddiproietto (approved):

ddiproietto (approved):

src/traced/probes/ftrace/cpu_reader.cc
rsavitski: I've also pondered this decision for a bit. Generally this has the potential to break cross-DS isolation (you can see all events in the ring buffer), which could be surprising to users.

It's not a problem on android since we run with equivalent tracefs permissions, there's future changes to consider. Basically, I wanted this debugging feature to be as out of the way as possible so that we don't won't need to keep thinking of its implications if/when we change the ftrace permission model.

However I've changed this to all-android because: (1) we understand the permissions, (2) it'll be helpful if ABI issues come up on release devices. The feature is still turned off by default via the config.

PR #1083, opened 2025-03-28
tp, stdlib: Add trigger_clone_flush_latency table

ddiproietto:
This does two things:

* It moves all the logic of the clone_duration metric into a stdlib table.
* It introduces another standard library table that takes into account the time it takes to process the trigger. 

Bug: https://b.corp.google.com/issues/406981933

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -14623,6 +14623,7 @@ genrule {
         "src/trace_processor/perfetto_sql/stdlib/stack_trace/jit.sql",
         "src/trace_processor/perfetto_sql/stdlib/stacks/cpu_profiling.sql",
         "src/trace_processor/perfetto_sql/stdlib/time/conversion.sql",
+        "src/trace_processor/perfetto_sql/stdlib/trace/stats.sql",
         "src/trace_processor/perfetto_sql/stdlib/v8/jit.sql",
         "src/trace_processor/perfetto_sql/stdlib/viz/flamegraph.sql",
         "src/trace_processor/perfetto_sql/stdlib/viz/slices.sql",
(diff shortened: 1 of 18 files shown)


Android.bp
```diff
@@ -14623,6 +14623,7 @@ genrule {
         "src/trace_processor/perfetto_sql/stdlib/stack_trace/jit.sql",
         "src/trace_processor/perfetto_sql/stdlib/stacks/cpu_profiling.sql",
         "src/trace_processor/perfetto_sql/stdlib/time/conversion.sql",
+        "src/trace_processor/perfetto_sql/stdlib/trace/stats.sql",
```
LalitMaganti: Can we please call the package traced to make it a little more obvious that this only for traces collected using Perfetto tools.

src/trace_processor/perfetto_sql/stdlib/trace/stats.sql
```diff
@@ -0,0 +1,87 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+-- Reports the duration of the flush operation for cloned traces (for each
+-- buffer).
+CREATE PERFETTO TABLE clone_flush_latency (
```
LalitMaganti: Can we please add a prefix this with traced? This is part of the presubmit checks in tools/run_presubmit we are missing.

Android.bp
ddiproietto: Good idea, done

src/trace_processor/perfetto_sql/stdlib/trace/stats.sql
ddiproietto: Good idea, done

LalitMaganti (approved):

PR #1085, opened 2025-03-29
Add stdlib table for sysui CUJs.

devianb:
As part of the stdlib migration, add a new table to track the jank/latency CUJs for SystemUI/Launcher. More details about the migration can be found here:
go/perfetto-jank-stdlib-migration

Bug: 400855183
Test: tools/diff_test_trace_processor.py --name-filter ".*sysui_cujs.*" out/linux_clang_debug/trace_processor_shell
Change-Id: I665fda17c79606f4a0bf1d03218287d753afcebf


src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,203 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE android.frames.timeline;
+
+-- Table tracking all latency and jank CUJs related to SystemUI and Launcher.
+CREATE PERFETTO TABLE sysui_cujs (
```
LalitMaganti: This table should be prefixed with android

LalitMaganti: Please make sure you run tools/run_presubmit (we will make this a mandatory check with time).

test/trace_processor/diff_tests/metrics/android/atrace_utils.py
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from os import sys, path
+import synth_common
+
+def add_main_thread_atrace(trace, ts, ts_end, buf, pid):
```
LalitMaganti: Why are we doing this instead of relying on synth_common? Feels like this is just replicating a lot of the functionality of that.

test/trace_processor/diff_tests/include_index.py
```diff
@@ -258,6 +259,7 @@ def fetch_all_diff_tests(index_path: str) -> List['testing.TestCase']:
       AndroidGpu,
       AndroidStdlib,
       CpuClusters,
+      Cujs,
```
nicomazz: nit: AndroidCujs? does this include both latency and jank?

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,203 @@
...
+    SUBSTR(slice.name, 3, LENGTH(slice.name) - 3) AS cuj_name,
+    ts,
+    dur,
+    ts + dur AS ts_end
+  FROM slice
+  JOIN process_track
+    ON slice.track_id = process_track.id
+  JOIN process USING (upid)
+  WHERE
+    slice.name GLOB 'J<*>' OR slice.name GLOB 'L<*>'
```
nicomazz: for latency cujs we don't want to snap to the frame boundaries: is that what's happening here?

devianb: Yes, that is how I have written the logic for both Jank and latency CUJs. Should the latency CUJs just have the slice boundary instead?

test/trace_processor/diff_tests/include_index.py
devianb: Yes, it includes both jank and latency cujs.

test/trace_processor/diff_tests/metrics/android/atrace_utils.py
devianb: This file was not meant to be committed in this CL. I will update the synth_common file with the necessary functions and upload it as part of the migration CL.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
nicomazz: Latency I was shouldn't snap to frame boundaries

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

test/trace_processor/diff_tests/include_index.py
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done. Made separate tables for jank and latency CUJs, with only the jank cujs latching to the frame boundaries.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,344 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE android.frames.timeline;
+
+-- Exhaustive list of jank and latency CUJs.
+CREATE PERFETTO TABLE _cujs AS
```
LalitMaganti: Even if these are private _cujs is too generic a name and will certainly clash. Please add some other prefix to this and all below.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
LalitMaganti: Please resolve comments if they're done.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

LalitMaganti (approved):
LGTM.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,348 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE android.frames.timeline;
+
+-- Exhaustive list of jank and latency CUJs.
+CREATE PERFETTO TABLE _sysui_cujs AS
```
nicomazz: I would add a warning in the doc here, or even reword the table name: my fear is that people will start using this table instead of the proper one, despite the ts and dur are potentially different than the ones we use for montoring.

What about renaming this to _sysui_cujs_slices instead? Then, at least for latency CUJs you can generate the table directly from this one.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,348 @@
...
+        )
+        -- frame end within cuj
+        OR (
+          (
+            frame.ts + frame.dur
+          ) >= cuj.ts AND (
+            frame.ts + frame.dur
+          ) <= cuj.ts_end
+        )
+      )
```
nicomazz, line 209: nit: the indentation looks a bit off: if also parenthesis fit in a single line, I think we can keep them there

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,348 @@
...
+      *
+    FROM frames_in_cuj
+    WHERE
+      frame_idx = 1 OR frame_idx = frame_cnt
+  )
+SELECT
+  cuj.cuj_id,
+  cuj.upid,
+  cuj.process_name,
+  -- Extracts "CUJ_NAME" from "J<CUJ_NAME>"
```
nicomazz, line 244: can we also keep the "slice_cuj_name" with J< and L<? I think we need it in several plugins, and if we keep it here we can avoid joining more tables there and have cleaner code

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,348 @@
...
+  dur DURATION,
+  -- State of the CUJ whether it was completed/cancelled.
+  state STRING,
+  -- layer id associated with the actual frame.
+  layer_id LONG,
+  -- vysnc id of the first frame that falls within the CUJ boundary.
+  begin_vsync LONG,
+  -- vysnc id of the last frame that falls within the CUJ boundary.
+  end_vsync LONG,
+  -- thread id of the UI thread.
```
nicomazz: latency cujs don't have an UI thread

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,348 @@
...
+  -- Duration of the CUJ calculated based on the ts and ts_end values.
+  dur DURATION,
+  -- State of the CUJ whether it was completed/cancelled.
+  state STRING,
+  -- layer id associated with the actual frame.
+  layer_id LONG,
+  -- vysnc id of the first frame that falls within the CUJ boundary.
+  begin_vsync LONG,
+  -- vysnc id of the last frame that falls within the CUJ boundary.
+  end_vsync LONG,
```
nicomazz: We actually don't care about frame infromation for latency cujs

test/trace_processor/diff_tests/include_index.py
```diff
@@ -116,6 +116,7 @@
 from diff_tests.stdlib.android.heap_profile_tests import HeapProfile
 from diff_tests.stdlib.android.memory import AndroidMemory
 from diff_tests.stdlib.android.startups_tests import Startups
+from diff_tests.stdlib.android.sysui_cujs_test import AndroidCujs
```
nicomazz: Should we be consistent and call them SystemuiCujs instead? (this would also keep it coherent with the file name)

nicomazz (changes requested):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Removed.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Removed.

test/trace_processor/diff_tests/include_index.py
devianb: Changed.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Actually this weird formatting is done by the format_stdlib script. Without running it, the perfetto CI check will fail. @LalitMaganti can we bypass this somehow?

nicomazz: let's not bypass it if this was from a formatting script: even if it looks weird at least we have coherent formatting.

test/trace_processor/diff_tests/include_index.py
```diff
@@ -116,6 +116,7 @@
 from diff_tests.stdlib.android.heap_profile_tests import HeapProfile
 from diff_tests.stdlib.android.memory import AndroidMemory
 from diff_tests.stdlib.android.startups_tests import Startups
+from diff_tests.stdlib.android.sysui_cujs_test import SystemuiCujs
```
nicomazz: nit: SystemUI

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Actually the ui_thread field would still be needed, as it is being used in the android_blocking_call_cuj_metric

test/trace_processor/diff_tests/include_index.py
devianb: Done

src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_metric.sql
```diff
@@ -13,74 +13,22 @@
...
-        process_name,
-        process_metadata,
-        cuj_name,
-        tb.ts,
-        tb.dur,
-        tb.ts_end
-    FROM android_jank_cuj_main_thread_cuj_boundary tb
-        JOIN android_jank_cuj cuj USING (cuj_id)
-        JOIN android_jank_cuj_main_thread t USING (cuj_id)
+DROP TABLE IF EXISTS android_jank_latency_cujs;
```
nicomazz: Can we add this to the stdlib as well? In several cases we want both latency and jank cujs. We could have a "type" field maybe saying if it's one or the other.

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_different_ui_thread.out
```diff
@@ -10,7 +10,7 @@ android_blocking_calls_cuj_metric {
       is_kernel_task: false
     }
     ts: 0
-    dur: 18000000
+    dur: 16000000
```
nicomazz: Why are these tests changing?

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_metric.out
```diff
@@ -1,6 +1,6 @@
 android_blocking_calls_cuj_metric {
   cuj {
-    id: 1
+    id: 2
     name: "TEST_SYSUI_LATENCY_EVENT"
```
nicomazz, line 4: Why from 2? What happened to the one with Id 1?

test/trace_processor/diff_tests/metrics/graphics/android_jank_cuj.py
```diff
@@ -198,16 +198,24 @@ def add_actual_surface_frame_events(ts,
...
 trace.add_package_list(
     ts=0, name="com.android.systemui", uid=10001, version_code=1)
 trace.add_package_list(ts=0, name="android", uid=1000, version_code=1)
+
 trace.add_process(pid=PID, ppid=1, cmdline="com.android.systemui", uid=10001)
 trace.add_process(
     pid=SF_PID, ppid=1, cmdline="/system/bin/surfaceflinger", uid=1000)
+
+# trace.add_thread(
+#       tid=PID, tgid=PID, cmdline="MainThread", name="MainThread")
```
nicomazz: Expected?

nicomazz (changes requested):
Is it possible to change the metrics and keep tests as before to have confidence in the changes? It would be great if test refactors was happening separately, if possible

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_different_ui_thread.out
devianb: Earlier the CUJ boundary defined by the CUJ slice. Now the boundary is defined by the frame boundary, hence the values have changed. I added frame information in all the test traces, and according to these frame boundaries, the metric values have changed.

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_metric.out
devianb: The _sysui_cuj_slices table in stdlib has the list of all CUJ slices with the 'J<' or 'L<' prefix. In this case the cuj_id 1 is given to the 'J<IGNORED>' CUJ in the trace file.

test/trace_processor/diff_tests/metrics/graphics/android_jank_cuj.py
devianb: Accidentally added while testing. Removed.

src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_metric.sql
devianb: Done.

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_different_ui_thread.out
nicomazz: > Earlier the CUJ boundary defined by the CUJ slice

At least for jank CUJs, I think we were using frame boudaries aleady, the same we were using for alerting. Are you sure we were using the CUJ slice before?

We should be using the CUJ slice only for latency CUJs, but not for jank ones

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_metric.out
nicomazz: >  In this case the cuj_id 1 is given to the 'J' CUJ in the trace file.

I don't completely understand: what's the "J" cuj in the trace file?

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_different_ui_thread.out
devianb: > At least for jank CUJs, I think we were using frame boudaries aleady, the same we were using for alerting. Are you sure we were using the CUJ slice before?

Yes we were using the frame boundaries for jank CUJs already. The difference now is that the start ts of the CUJ is the start of the actual frame(from the android_frames_layers table), whereas earlier it used to be the start of the expected frame.

> We should be using the CUJ slice only for latency CUJs, but not for jank ones.
Yes, that is how the current logic is, frame boundary for jank, and CUJ slice for latency.

In this particular test, the CUJ under consideration is actually a jank CUJ, and going by the logic mentioned above, the duration 16000000 is correct.

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_metric.out
devianb: Sorry incomplete msg, I mean J\<IGNORED\>.
https://cs.android.com/android/platform/superproject/main/+/main:external/perfetto/test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_metric.py;drc=a7ee98c8d3cdd8bd824f15b8aa1c1cc04839b9ac;l=274

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_different_ui_thread.out
nicomazz: > The difference now is that the start ts of the CUJ is the start of the actual frame(from the android_frames_layers table), whereas earlier it used to be the start of the expected frame.

Why are we using the actual frame right now? If the ui thread is stuck with runnables the frame will be delayed, and will have less time to reach the deadline. I wonder if we should use the expected timeline for the cuj beginning timestamp, and the last frame in the actual timeline for the CUJ end ts?

Anyway, would you mind keeping the existing behaviour for this cl, and changing how it works in a follow up? Otherwise we risk to just revert this commit if we have alerts regressions

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_metric.out
nicomazz: I see, the indexing is done on the plain slices, while in the output we filter out certain cujs. I would prefer to keep the previous behaviour, but I don't have strong opinions here, this is fine as well (at the end, we're not using that id at all anyway)

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_cuj_different_ui_thread.out
devianb: > I wonder if we should use the expected timeline for the cuj beginning timestamp, and the last frame in the actual timeline for the CUJ end ts.

We can use the expected timeline for CUJ start ts. But for the end ts, the existing logic considers the end ts of the Choreographer doFrame, and not the actual frame timeline. That is actually what is causing the change in the tests.

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_on_jank_cuj_metric.out
devianb, line 1: @nicomazz This change happened because in the existing logic, there is no layer_id check while deciding which frames lie within CUJ boundary. In the [android_jank_cuj.py](https://cs.android.com/android/platform/superproject/main/+/main:external/perfetto/test/trace_processor/diff_tests/metrics/graphics/android_jank_cuj.py;drc=00485344d140cab813ebd0d0a240c93482d0ef8b;l=729) file, because there is no actual frame(no layer id), this last frame is not considered in the new logic, hence the CUJ duration has changed.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
nicomazz: it seems this is using the cuj_event table ui_thread, but there won't be any ui_thread for latency cujs as we're not emitting the "ui_thread" instant event. What's in this right now? are we sure the android_blocking_call_cuj_metric works as intended and considers the correct thread? (for latency cujs, it should be the main thread

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,418 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE android.frames.timeline;
+
+-- Exhaustive list of jank and latency CUJs.
```
nicomazz: Let's add that these are just the slices emitted, but the proper CUJs and their boudnaries are computed also taking into account instant events and frame boundaries.

Otherwise, looking this comment, I would start using this table from the outside

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,418 @@
...
+FROM slice
+JOIN process_track
+  ON slice.track_id = process_track.id
+JOIN process
+  USING (upid)
+WHERE
+  slice.name GLOB 'J<*>'
+  OR slice.name GLOB 'L<*>'
+  AND (
+    process.name GLOB 'com.google.android*' OR process.name GLOB 'com.android.*'
```
nicomazz: can we extract this in a macro or similar, so we can easily control the processes we consider without having to dig deeper in the tables? I would expect us to reuse this in other places as well in the future

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,418 @@
...
+    FROM _sysui_cuj_state_markers AS csm
+    WHERE
+      csm.cuj_id = cuj.cuj_id AND csm.marker_name GLOB "*#UIThread"
+    LIMIT 1
+  ) AS ui_thread
+FROM _sysui_cuj_state_markers
+JOIN _sysui_cujs_slices AS cuj
+  USING (cuj_id)
+GROUP BY
+  cuj_id;
```
nicomazz: I suspect you can avoid writing subqueries here and just write something like this
```
MAX(
    CASE
      WHEN csm.marker_name GLOB '*layerId#*'
      THEN CAST(STR_SPLIT(csm.marker_name, 'layerId#', 1) AS INTEGER)
      ELSE NULL
    END
  ) AS layer_id,
  -- Extract begin VSync ID from 'beginVsync#<ID>' marker.
  MAX(
    CASE
      WHEN csm.marker_name GLOB '*beginVsync#*'
      THEN CAST(STR_SPLIT(csm.marker_name, 'beginVsync#', 1) AS INTEGER)
      ELSE NULL
    END
  ) AS begin_vsync,
  -- Extract end VSync ID from 'endVsync#<ID>' marker.
  MAX(
    CASE
      WHEN csm.marker_name GLOB '*endVsync#*'
      THEN CAST(STR_SPLIT(csm.marker_name, 'endVsync#', 1) AS INTEGER)
      ELSE NULL
    END
  ) AS end_vsync,
  -- Extract UI thread UTID from 'UIThread' marker.
  MAX(
    CASE
      WHEN csm.marker_type = 'UIThread' THEN csm.utid
      ELSE NULL
    END
  ) AS ui_thread
FROM _sysui_cuj_state_markers AS csm
-- Only consider markers relevant for extracting these values.
WHERE csm.marker_type IN ('layerId', 'beginVsync', 'endVsync', 'UIThread')
```

It should be slighly faster

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,418 @@
...
+  AND (
+    state != 'canceled'
+    -- Older builds don't have the state markers so we allow NULL but filter out
+    -- CUJs that are <4ms long - assuming CUJ was canceled in that case.
+    OR (
+      state IS NULL AND cuj.dur > 4e6
+    )
+  )
+GROUP BY
+  cuj_id
```
nicomazz: is this group by needed? it seems we're not doing any aggregation in the select anyway

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,418 @@
...
+SELECT
+  *,
+  "jank" AS cuj_type
+FROM android_sysui_jank_cujs
+UNION
+SELECT
+  *,
+  NULL AS layer_id,
+  NULL AS begin_vsync,
+  NULL AS begin_vsync,
```
nicomazz: shouldn't this be end_vsync?

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,418 @@
...
+  -- vysnc id of the last frame that falls within the CUJ boundary.
+  end_vsync LONG,
+  -- Type of CUJ, i.e. jank or latency.
+  cuj_type STRING
+) AS
+SELECT
+  *,
+  "jank" AS cuj_type
+FROM android_sysui_jank_cujs
+UNION
```
nicomazz: union all should be more efficient, as it doesn't check for dupes (and here it's guaranteed we won't have them)

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,418 @@
...
+  -- Id of the CUJ slice in perfetto. Keeping the slice id column as part of this table
+  -- as provision to lookup the actual CUJ slice ts and dur. The ts and dur in this table
+  -- might differ from the slice duration, as they are associated with start and end frame
+  -- corresponding to the CUJ.
+  slice_id JOINID(slice.id),
+  -- Start timestamp of the CUJ. Start of the CUJ as defined by the start of the CUJ slice,
+  -- or start of the actual frame, whichever is greater.
+  ts TIMESTAMP,
+  -- End timestamp of the CUJ. Calculated as the end timestamp of the last frame overlapping with the CUJ.
+  ts_end TIMESTAMP,
```
nicomazz: could you clearly state that for Jank CUJ timestamps are adjusted based on frames, while Latency CUJ timestamps use the original slice timing

test/trace_processor/diff_tests/metrics/android/android_blocking_calls_on_jank_cuj_metric.out
nicomazz: sg!

nicomazz (changes requested):

aMayzner (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Fixed.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Fixed.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Fixed. removed  fields from latency cujs table.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done.

LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,420 @@
...
+  cuj_type STRING
+) AS
+SELECT
+  *,
+  "jank" AS cuj_type
+FROM android_sysui_jank_cujs
+UNION ALL
+SELECT
+  *,
+  upid AS ui_thread,
```
nicomazz, line 419: let's add a small comment here that we're using upid as ui_thread as it's the tid of the main thread (otherwise it might seem a mistake from a quick look)

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,420 @@
...
+  -- expected frame.
+  ts TIMESTAMP,
+  -- End timestamp of the CUJ. Calculated as the end timestamp of the last actual frame
+  -- overlapping with the CUJ.
+  ts_end TIMESTAMP,
+  -- Duration of the CUJ calculated based on the ts and ts_end values.
+  dur DURATION,
+  -- State of the CUJ whether it was completed/cancelled.
+  state STRING,
+  -- thread id of the UI thread.
```
nicomazz: let's add that in case of latency cujs, this will always be tha main thread of the process?

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -0,0 +1,420 @@
...
+  slice_id JOINID(slice.id),
+  -- Start timestamp of the CUJ. Start of the CUJ as defined by the start of the first overlapping
+  -- expected frame.
+  ts TIMESTAMP,
+  -- End timestamp of the CUJ. Calculated as the end timestamp of the last actual frame
+  -- overlapping with the CUJ.
+  ts_end TIMESTAMP,
+  -- Duration of the CUJ calculated based on the ts and ts_end values.
+  dur DURATION,
+  -- State of the CUJ whether it was completed/cancelled.
```
nicomazz: Can we write explicitly which are the possible values in this comment? Are they only "completed" and "cancelled"?

e.g. ` -- State of the CUJ. One of: "completed", "cancelled", "..."

test/trace_processor/diff_tests/metrics/graphics/android_jank_cuj.py
```diff
@@ -198,16 +198,24 @@ def add_actual_surface_frame_events(ts,
...
 trace.add_packet()
 trace.add_package_list(
     ts=0, name="com.android.systemui", uid=10001, version_code=1)
 trace.add_package_list(ts=0, name="android", uid=1000, version_code=1)
+
 trace.add_process(pid=PID, ppid=1, cmdline="com.android.systemui", uid=10001)
 trace.add_process(
     pid=SF_PID, ppid=1, cmdline="/system/bin/surfaceflinger", uid=1000)
+
+# trace.add_thread(
```
nicomazz: should we remove this comment?

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done

test/trace_processor/diff_tests/metrics/graphics/android_jank_cuj.py
devianb: Done

LalitMaganti (approved):

PR #1098, opened 2025-03-31
Android Java SDK: use 'java.lang.ref.Cleaner' to cleanup native memory.

KirillTim:
The Cleaner API requires 'minSdkVersion' be 33, so we update it.

Tested:
```
```

Also tested in android tree:
```
cd external/perfetto
git fetch https://github.com/google/perfetto/ dev/ktimofeev/java-android-sdk-free-native-memory && git checkout FETCH_HEAD
atest perfetto_trace_instrumentation_test
```


src/android_sdk/java/main/dev/perfetto/sdk/PerfettoNativeMemoryCleaner.java
```diff
@@ -0,0 +1,98 @@
...
+
+    synchronized void registerAlloc(String target) {
+      allocCount.put(target, allocCount.getOrDefault(target, 0) + 1);
+    }
+
+    synchronized void registerFree(String target) {
+      freeCount.put(target, freeCount.getOrDefault(target, 0) + 1);
+    }
+
+    public synchronized int getAllocCountForTarget(String target) {
```
zezeozue, line 81: Can you add a comment that this is for test?

And any others

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoNativeMemoryCleaner.java
```diff
@@ -0,0 +1,98 @@
...
+      }
+      applyNativeFunction(mFreeFunctionPtr, mPtr);
+    }
+  }
+
+  public static class AllocationStats {
+    private final Map<String, Integer> allocCount = new HashMap<>();
+    private final Map<String, Integer> freeCount = new HashMap<>();
+
+    synchronized void registerAlloc(String target) {
```
zezeozue, line 73: Can we avoid holding a lock? The PerfettoTrackEventExtra is thread local so  maybe the PerfettoCleaner can be a member of PTEE and we have one cleaner per thread and can avoid synchronizing?

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoNativeMemoryCleaner.java
KirillTim: I've add a comment for the whole `AllocationStats` class that is used only in tests.

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoNativeMemoryCleaner.java
KirillTim: We only use this method in tests and on the slow path (memory allocation), so it is fine to have it synchronized.

ddiproietto (approved):
Looks good to me. Let's wait for @zezeozue to have a look as well.

zezeozue (approved):

PR #1105, opened 2025-03-31
tools: fix presubmit script formatting

LalitMaganti:


LalitMaganti:

- #1105 ◀

tools/run_presubmit
```diff
@@ -69,20 +69,40 @@ def run_command(args: List[str],
...
+    tracking_result = run_command(args, check=True)
+    tracking_branch = tracking_result.stdout.strip()
+
+    # If we have a tracking branch, check if it's local
+    if tracking_branch:
+      # Check if this is a local branch by trying to get its ref in refs/heads/
+      try:
+        _ = run_command(
+            ["git", "show-ref", "--verify", f"refs/heads/{tracking_branch}"],
+            check=True)
```
primiano, line 88: add `stderr=subprorcess.DEVNULL` so we don't see confusing errors?

Same below

tools/run_presubmit
```diff
@@ -69,20 +69,40 @@ def run_command(args: List[str],
 
 
 def get_merge_base() -> Optional[str]:
-  """Gets the merge base with the upstream branch (@{u}). Returns None if not
-  found."""
+  result = run_command(["git", "symbolic-ref", "--short", "HEAD"], check=True)
```
primiano: nit: use single quotes (or generally format this file with yapf)

tools/run_presubmit
```diff
@@ -69,20 +69,40 @@ def run_command(args: List[str],
...
+  branch = result.stdout.strip()
+
+  # First check if there's a tracking branch
   try:
-    run_command(['git', 'rev-parse', '--abbrev-ref', '@{u}'],
-                check=True,
-                stderr=subprocess.DEVNULL)
-    result = run_command(['git', 'merge-base', 'HEAD', '@{u}'])
-    return result.stdout.strip()
+    # More direct way to get the upstream branch
```
primiano, line 77: The nestead try/catch IMHO hurts readability.

WHy dion't we make run_command return none in case of failure? that's makes the code easier to read imho

tools/run_presubmit
LalitMaganti: Done

tools/run_presubmit
LalitMaganti: Not necessary, capture_output means output is implicitly captured.

tools/run_presubmit
LalitMaganti: Would be nice but also it would mean changing this file signifcantly. I'd like to do this in a followup.

primiano (approved):

PR #1106, opened 2025-03-31
tools: copy presubmit hook

LalitMaganti:

PR #1108, opened 2025-03-31
tools: workflow for running presubmit

LalitMaganti:


LalitMaganti:

- #1105
    - #1106
        - #1108 ◀

.github/workflows/presubmit-script.yml
```diff
@@ -0,0 +1,38 @@
+# Copyright (C) 2025 The Android Open Source Project
```
primiano, line 1: I think this should go into the existing repo-cecks rather than yet another run, otherwise it eats too much time for checkout and cache for no good reason

LalitMaganti:
Don't want this anymore, we're going with a different approach.

PR #1133, opened 2025-04-01
Send guest VMs' SystemInfo packet to host traced

jahdiel-alvarez:
During traced_relay connection to the host traced instance, we send the SystemInfo packet of the guest VM. Once received, traced caches the serialized system info and emits it to the trace file at the end of a tracing session.

Having the SystemInfo packet of all secondary VMs allows us to do syscall id mapping to syscall name and also provides useful information to understand all the machines involved during a tracing session.

Test: tools/run_android_test --system-adb out/android perfetto_unittest
Bug: http://b/402546586


LalitMaganti (commented):
Looks good to me but I'd like Chinglin to also take a look.

src/traced_relay/relay_service.cc
```diff
@@ -81,6 +84,88 @@ std::string GenerateSetPeerIdentityRequest(int32_t pid,
   return ipc::BufferedFrameDeserializer::Serialize(ipc_frame);
 }
 
+void SetSystemInfo(protos::gen::InitRelayRequest* request) {
```
cervantesyu, line 86: What this method does is mostly the same as in  TracingServiceImpl::EmitSystemInfo(). Can we move this to some common layer code, e.g. under //base so that TracingServiceImpl and here can share the implementation. //base sounds a plausible choice as "perfetto/ext/base/version.h" and "perfetto/ext/base/android_utils.h" are included there.

src/tracing/ipc/producer/relay_ipc_client.cc
```diff
@@ -54,6 +54,27 @@ void RelayIPCClient::OnDisconnect() {
     listener_->OnServiceDisconnected();
 }
 
+void RelayIPCClient::InitRelay(const InitRelayRequest& init_relay_request) {
+  PERFETTO_DCHECK_THREAD(thread_checker_);
+  if (!connected_) {
```
cervantesyu: IIUC, this branch is unreachable because InitRelay() is synchronously called in RelayService::OnServiceConnected(), which is the |listener_| of this class, and it's set to true in RelayIPCClient::OnConnect(). We should just assert |connected_| here.

src/traced_relay/relay_service.cc
jahdiel-alvarez: The main caveat is that we use two different proto serializer classes. But I guess the underlying proto is the same so we should be able to deduplicate the code.

@LalitMaganti what is your recomendation for where this new shared codebase should live? Should we use something like "perfetto/ext/base/system_info.h"?

src/tracing/ipc/producer/relay_ipc_client.cc
jahdiel-alvarez: Doing runtime assertions doesn't seem to be a convention in perfetto. I'll simply add `PERFETTO_UNLIKELY` in the if statement to hint to the compiler the unlikeliness of this being true.

LalitMaganti: Actually it's very much the convention. We do PERFETTO_CHECK everywhere in the codebase.

include/perfetto/ext/base/BUILD.gn
```diff
@@ -57,6 +57,7 @@ source_set("base") {
     "unix_task_runner.h",
     "utils.h",
     "uuid.h",
+    "version.h",
```
LalitMaganti: Please don't tdo this, this is intentionally separate. Depend on base:version directly.

protos/perfetto/ipc/BUILD.gn
```diff
@@ -33,6 +33,7 @@ perfetto_proto_library("@TYPE@") {
     # .ipc.h depends on the .gen.h coming from the various common/xxx.proto.
     "../common:cpp",
     "../config:cpp",
+    "../trace:minimal_cpp",
```
LalitMaganti: Can you move the system_info file to common so you don't need this edge?

LalitMaganti (changes requested):

LalitMaganti (approved):

PR #1144, opened 2025-04-02
proto: Add generator_visibility to "//protos/perfetto/config/chrome:scenario_@TYPE@"

atetubou:
This is to make the target visible from
https://source.chromium.org/chromium/chromium/src/+/main:content/test/BUILD.gn;l=3625;drc=b0e7f7c12e45613c7ce60ab3ffb540ca158aa3a1
after we change proto_library template not to have unnecessary dependency.

This is similar PR with https://github.com/google/perfetto/pull/1117.

Bug: http://b/issues/402356473


protos/perfetto/config/chrome/BUILD.gn
```diff
@@ -21,4 +21,7 @@ perfetto_proto_library("scenario_@TYPE@") {
   sources = [ "scenario_config.proto" ]
   generate_descriptor = "scenario_config.descriptor"
   descriptor_root_source = "scenario_config.proto"
+  generator_visibility = [
```
LalitMaganti: This needs to be guarded with build_with_chromium.

LalitMaganti: Also @primiano. Honestly this is quite ugly that things are leaking into our codebase like this. Adding Primiano to make the final call on this.

atetubou: If we want to avoid having path of downstream project, we can set `"*"` instead to make generator action public.

LalitMaganti: That would be ideal yes.

atetubou: OK, updated to use `"*"`.

atetubou:
Can I get approval to merge this now?

LalitMaganti (approved):

stevegolton (approved):

LalitMaganti (approved):

PR #1147, opened 2025-04-02
Make it possible to remove JSON from libperfetto

edechamps-Google:
This change makes it possible for GN Perfetto embedders to not bundle the JSON export code when depending on libperfetto. This is useful for some embedders such as Cronet, where that dependency is problematic from a build system perspective because of the complexity of the code generators involved.


BUILD.gn
```diff
@@ -334,7 +334,7 @@ if (build_with_chromium) {
       ]
       public_deps += [ "include/perfetto/ext/tracing/ipc:ipc" ]
     }
-    if (!is_nacl) {
```
LalitMaganti, line 337: There's a todo at the top of this file:
```

# TODO(primiano): there seem to be two "libperfetto" and one
# "libperfetto_client_experimental" targets defined within this BUILD.gn file.
# Rationalize them with eseckler@. For now seems this one is only used from
# chromium and the other one only from the Android tree.
```

Instead of piling more hacks on top of this, can we figure out the proper solution here?  We cannot support more and more of these config options causing an explosion in the state space of what we have to support

edechamps-Google: I don't think these two problems are related. They seem orthogonal to me.

The one thing I tried to do first is remove the JSON dependency from `libperfetto` completely (unconditionally), and move them to their direct parent targets on the Chromium side. This way there is no conditional build flag. Unfortunately that proved harder than I thought, as I ran into various build failures that seemed painful to troubleshoot. This is made even more tedious by the fact that attempting to run Chromium CQ against a different revision of a third-party dependency (multiple third-party dependencies, in fact, as V8 depends on this too) is very tedious/inconvenient. I spent an entire day trying to make this work and then gave up when faced with weird Windows build failures that I cannot even reproduce.

This flag would make this kind of cleanup significantly easier, as it removes the need to coordinate the GN build rules with the Perfetto repo. So one approach could be to bring this change in, and then once rolled into Chromium we can look into making every build type set this flag to false (not just Cronet), at which point it can be removed, leaving a cleaner build dependency tree for everyone. Basically a 3-step migration process.

LalitMaganti: > I don't think these two problems are related. They seem orthogonal to me.

They're related in the fact that there's a bunch of complexity being pushed into Perfetto which really is chromium complexity.

> So one approach could be to bring this change in, and then once rolled into Chromium we can look into making every build type set this flag to false (not just Cronet), at which point it can be removed, leaving a cleaner build dependency tree for everyone. Basically a 3-step migration process.

If we could do this, that would be fantastic.

gn/perfetto.gni
```diff
@@ -102,6 +102,12 @@ if (!defined(perfetto_build_with_embedder)) {
   perfetto_build_with_embedder = build_with_chromium || is_perfetto_embedder
 }
 
+# Controls whether the `libperfetto` target bundles the JSON export code in its
+# dependencies. Can be overriden by embedders to reduce dependency bloat.
+if (!defined(perfetto_libperfetto_includes_json)) {
+  perfetto_libperfetto_includes_json = !is_nacl
```
LalitMaganti: Can we add build_with_chromium here?

LalitMaganti (approved):

gn/perfetto.gni
edechamps-Google: Done.

(Let me know if you want me to rebase and rewrite history with a single clean commit. I have no idea what your preferred git PR workflow is.)

PR #1152, opened 2025-04-02
[ui] Extend trusted origin to support *.proxy.googlers.com hostnames

bhamiltoncx:
This builds on f2511be5e0c05976ccb5233edfa8ee2141923731 to extend support for trusted origins to `*.proxy.googlers.com` hostnames, which are used in addition to `*.c.googlers.com` hostnames.

I extended the unit tests to cover this and confirmed the tests pass with `./ui/run-unittests`.

Bug: http://b/402495918

### DIFF
--- a/ui/src/frontend/post_message_handler.ts
+++ b/ui/src/frontend/post_message_handler.ts
@@ -53,6 +53,7 @@ export function isTrustedOrigin(origin: string): boolean {
   const hostname = new URL(origin).hostname;
   if (hostname.endsWith('.corp.google.com')) return true;
   if (hostname.endsWith('.c.googlers.com')) return true;
+  if (hostname.endsWith('.proxy.googlers.com')) return true;
   if (
     hostname === 'localhost' ||
     hostname === '127.0.0.1' ||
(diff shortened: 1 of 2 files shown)


LalitMaganti:
@primiano can you please take a call on this? Should we be doing this at all?

bhamiltoncx:
> @primiano can you please take a call on this? Should we be doing this at all?

For more context, currently we can't test any Google code which opens a tab to ui.perfetto.dev and posts the Perfetto protobuf over to the other tab. The origin has changed to `*.proxy.googlers.com`, so it doesn't pass this check.

Without this, ui.perfetto.dev loads but it ignores the message from the other tab.

I'm not 100% sure what the trusted origin logic was originally about, but it's been around for a long time.

LalitMaganti:
> I'm not 100% sure what the trusted origin logic was originally about, but it's been around for a long time.

I know but it was a convenience hack which we probably shouldn't have done in the first place honestly.

bhamiltoncx:
> I know but it was a convenience hack which we probably shouldn't have done in the first place honestly.

Makes sense. Hopefully there's an alternative for testing code which opens a Perfetto tab. If not, a lot of code will go untested.

primiano (approved):

LalitMaganti (approved):

PR #1146, opened 2025-04-02
df: add support for setid sorting

LalitMaganti:
Welcome to Perfetto!
Make sure your PR has a bug/issue attached or has at least
a clear description of the problem you are trying to fix.

For more details please see
https://perfetto.dev/docs/contributing/getting-started

PR #1140, opened 2025-04-01
df: add support for null overlays and isnull/isnotnull

LalitMaganti:
This PR adds support for both null overlays on columns and the
isnull and isnotnull operators on all columns.

PR #1088, opened 2025-03-31
tp: add support for string columns

LalitMaganti:

PR #1086, opened 2025-03-30
df: add handling of all numeric column types

LalitMaganti:
This CL adds proper filtering support of all numeric columns. Tests only are implemented for uint32 because all the numeric types basically use the same logic.

PR #1154, opened 2025-04-03
df: add runtime builder class for building dataframes for unknown schemas

LalitMaganti:
This CL adds a builder class for building dataframes where the schema of the
dataframe is not known upfront and should be inferred from the inserted
rows. We try and always be as memory efficient as possible and infer as many
properties (i.e. sorted/id etc) as we can.

### DIFF
--- a/src/trace_processor/dataframe/BUILD.gn
+++ b/src/trace_processor/dataframe/BUILD.gn
@@ -19,6 +19,7 @@ source_set("dataframe") {
     "cursor.h",
     "dataframe.cc",
     "dataframe.h",
+    "runtime_dataframe_builder.h",
   ]
   deps = [
     "../../../gn:default_deps",
@@ -48,6 +49,7 @@ perfetto_unittest_source_set("unittests") {
   testonly = true
   sources = [
     "dataframe_unittest.cc",
+    "runtime_dataframe_builder_unittest.cc",
     "type_set_unittest.cc",
   ]
   deps = [
--- a/src/trace_processor/dataframe/impl/bit_vector.h
+++ b/src/trace_processor/dataframe/impl/bit_vector.h
@@ -51,13 +51,26 @@ struct BitVector {
   // Default constructor creates an empty bit vector.
   BitVector() = default;
 
+  // Movable but not copyable.
+  BitVector(BitVector&&) = default;
+  BitVector& operator=(BitVector&&) = default;
+  BitVector(const BitVector&) = delete;
+  BitVector& operator=(const BitVector&) = delete;
+
   // Allocates a new BitVector with `size` unset bits.
   //
   // size: Size for how many bits to add to the BitVector.
-  // Returns a BitVector with the given size and with all bits set to false.
-  static BitVector CreateWithSize(uint64_t size) {
+  // Returns a BitVector with the given size and with all bits set to ``value`.
+  static BitVector CreateWithSize(uint64_t size, bool value = false) {
+    if (size == 0) {
+      return {};
+    }
     auto words = FlexVector<uint64_t>::CreateWithSize((size + 63u) / 64u);
-    memset(words.data(), 0, words.size() * sizeof(uint64_t));
+    memset(words.data(), value ? 0xFF : 0, words.size() * sizeof(uint64_t));
+    // Clear any bits past `size`.
+    if (size % 64u != 0u) {
+      words.back() &= (1ull << (size % 64u)) - 1ull;
+    }
     PERFETTO_DCHECK(size <= words.size() * 64u);
     return BitVector(std::move(words), size);
   }
--- a/src/trace_processor/dataframe/impl/bit_vector.h
+++ b/src/trace_processor/dataframe/impl/bit_vector.h
@@ -60,7 +60,7 @@ struct BitVector {
   // Allocates a new BitVector with `size` unset bits.
   //
   // size: Size for how many bits to add to the BitVector.
-  // Returns a BitVector with the given size and with all bits set to ``value`.
+  // Returns a BitVector with the given size and with all bits set to `value`.
   static BitVector CreateWithSize(uint64_t size, bool value = false) {
     if (size == 0) {
       return {};
--- a/src/trace_processor/dataframe/impl/query_plan.cc
+++ b/src/trace_processor/dataframe/impl/query_plan.cc
@@ -42,9 +42,10 @@ namespace {
 
 // Calculates filter preference score for ordering filters.
 // Lower scores are applied first for better efficiency.
-uint32_t FilterPreference(const FilterSpec& fs, const ColumnSpec& col) {
+uint32_t FilterPreference(const FilterSpec& fs, const impl::Column& col) {
   enum AbsolutePreference : uint8_t {
     kIdEq,                     // Most efficient: id equality check
+    kSetIdSortedEq,            // Set id sorted equality check
     kIdInequality,             // Id inequality check
     kNumericSortedEq,          // Numeric sorted equality check
     kNumericSortedInequality,  // Numeric inequality check
@@ -53,10 +54,14 @@ uint32_t FilterPreference(const FilterSpec& fs, const ColumnSpec& col) {
     kLeastPreferred,           // Least preferred
   };
   const auto& op = fs.op;
-  const auto& ct = col.column_type;
-  const auto& n = col.nullability;
+  const auto& ct = col.storage.type();
+  const auto& n = col.overlay.nullability();
   if (n.Is<NonNull>() && ct.Is<Id>() && op.Is<Eq>()) {
     return kIdEq;
+  }
+  if (n.Is<NonNull>() && ct.Is<Uint32>() && col.sort_state.Is<SetIdSorted>() &&
+      op.Is<Eq>()) {
+    return kSetIdSortedEq;
   }
   if (n.Is<NonNull>() && ct.Is<Id>() && op.IsAnyOf<InequalityOp>()) {
     return kIdInequality;
@@ -124,14 +129,14 @@ base::Status QueryPlanBuilder::Filter(std::vector<FilterSpec>& specs) {
                    [this](const FilterSpec& a, const FilterSpec& b) {
                      const auto& a_col = columns_[a.column_index];
                      const auto& b_col = columns_[b.column_index];
-                     return FilterPreference(a, a_col.spec) <
-                            FilterPreference(b, b_col.spec);
+                     return FilterPreference(a, a_col) <
+                            FilterPreference(b, b_col);
                    });
 
   // Apply each filter in the optimized order
   for (FilterSpec& c : specs) {
     const Column& col = columns_[c.column_index];
-    const auto& ct = col.spec.column_type;
+    StorageType ct = col.storage.type();
 
     // Get the non-null operation (all our ops are non-null at this point)
     auto non_null_op = c.op.TryDowncast<NonNullOp>();
@@ -190,7 +195,7 @@ void QueryPlanBuilder::Output(uint64_t cols_used) {
       continue;
     }
     const auto& col = columns_[i];
-    switch (col.spec.nullability.index()) {
+    switch (col.overlay.nullability().index()) {
       case Nullability::GetTypeIndex<SparseNull>():
       case Nullability::GetTypeIndex<DenseNull>():
         null_cols.emplace_back(ColAndOffset{i, plan_.params.output_per_row});
@@ -227,7 +232,7 @@ void QueryPlanBuilder::Output(uint64_t cols_used) {
     }
     for (auto [col, offset] : null_cols) {
       const auto& c = columns_[col];
-      switch (c.spec.nullability.index()) {
+      switch (c.overlay.nullability().index()) {
         case Nullability::GetTypeIndex<SparseNull>(): {
           using B = bytecode::StrideTranslateAndCopySparseNullIndices;
           auto reg = PrefixPopcountRegisterFor(col);
@@ -307,7 +312,9 @@ void QueryPlanBuilder::NullConstraint(const NullOp& op, FilterSpec& c) {
   // the caller (i.e. SQLite) knows that we are able to handle the constraint.
   c.value_index = plan_.params.filter_value_count++;
 
-  switch (columns_[c.column_index].spec.nullability.index()) {
+  const auto& col = columns_[c.column_index];
+  uint32_t nullability_type_index = col.overlay.nullability().index();
+  switch (nullability_type_index) {
     case Nullability::GetTypeIndex<SparseNull>():
     case Nullability::GetTypeIndex<DenseNull>(): {
       auto indices = EnsureIndicesAreInSlab();
@@ -333,12 +340,12 @@ void QueryPlanBuilder::NullConstraint(const NullOp& op, FilterSpec& c) {
 
 bool QueryPlanBuilder::TrySortedConstraint(
     const FilterSpec& fs,
-    const ColumnType& ct,
+    const StorageType& ct,
     const NonNullOp& op,
     const bytecode::reg::RwHandle<CastFilterValueResult>& result) {
   const auto& col = columns_[fs.column_index];
-  const auto& n = col.spec.nullability;
-  if (!n.Is<NonNull>() || col.spec.sort_state.Is<Unsorted>()) {
+  const auto& nullability = col.overlay.nullability();
+  if (!nullability.Is<NonNull>() || col.sort_state.Is<Unsorted>()) {
     return false;
   }
   auto range_op = op.TryDowncast<RangeOp>();
@@ -354,7 +361,7 @@ bool QueryPlanBuilder::TrySortedConstraint(
       base::unchecked_get<bytecode::reg::RwHandle<Range>>(indices_reg_);
 
   // Handle set id equality with a specialized opcode.
-  if (ct.Is<Uint32>() && col.spec.sort_state.Is<SetIdSorted>() && op.Is<Eq>()) {
+  if (ct.Is<Uint32>() && col.sort_state.Is<SetIdSorted>() && op.Is<Eq>()) {
     using B = bytecode::Uint32SetIdSortedEq;
     auto& bc = AddOpcode<B>();
     bc.arg<B::val_register>() = result;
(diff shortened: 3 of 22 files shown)


LalitMaganti:

- #1079
    - #1086
        - #1087
            - #1088
                - #1140
                    - #1146
                        - #1154 ◀

src/trace_processor/dataframe/impl/bit_vector.h
```diff
@@ -51,13 +51,30 @@ struct BitVector {
...
+  BitVector& operator=(BitVector&&) = default;
+  BitVector(const BitVector&) = delete;
+  BitVector& operator=(const BitVector&) = delete;
+
   // Allocates a new BitVector with `size` unset bits.
   //
   // size: Size for how many bits to add to the BitVector.
-  // Returns a BitVector with the given size and with all bits set to false.
+  // Returns a BitVector with the given size and with all bits set to ``value`.
+  template <bool value = false>
```
aMayzner: Why are we templating everything now? Is it really such a performance gain?

src/trace_processor/dataframe/impl/bit_vector.h
```diff
@@ -51,13 +51,30 @@ struct BitVector {
...
+    if (size == 0) {
+      return {};
+    }
     auto words = FlexVector<uint64_t>::CreateWithSize((size + 63u) / 64u);
-    memset(words.data(), 0, words.size() * sizeof(uint64_t));
+    if constexpr (value) {
+      memset(words.data(), 0xFF, words.size() * sizeof(uint64_t));
+      if (size % 64u != 0u) {
+        words.back() &= (1ull << (size % 64u)) - 1;
+      }
```
aMayzner: Please add a comment explaining what you are doing here

src/trace_processor/dataframe/impl/bit_vector.h
```diff
@@ -51,13 +51,30 @@ struct BitVector {
...
+  BitVector(const BitVector&) = delete;
+  BitVector& operator=(const BitVector&) = delete;
+
   // Allocates a new BitVector with `size` unset bits.
   //
   // size: Size for how many bits to add to the BitVector.
-  // Returns a BitVector with the given size and with all bits set to false.
+  // Returns a BitVector with the given size and with all bits set to ``value`.
+  template <bool value = false>
   static BitVector CreateWithSize(uint64_t size) {
```
aMayzner: Please test this code. In a BitVector unittest.

src/trace_processor/dataframe/impl/query_plan.cc
```diff
@@ -50,8 +50,8 @@ uint32_t FilterPreference(const FilterSpec& fs, const ColumnSpec& col) {
     kLeastPreferred,           // Least preferred
   };
   const auto& op = fs.op;
-  const auto& ct = col.column_type;
-  const auto& n = col.nullability;
+  const auto& ct = col.storage.column_type();
```
aMayzner: It seems like it shouldn't be `col.storage.column_type()` but rather `col.storage.type()` or `col.storage.data_type()`

src/trace_processor/dataframe/impl/query_plan.cc
```diff
@@ -39,7 +39,7 @@ namespace {
 
 // Calculates filter preference score for ordering filters.
 // Lower scores are applied first for better efficiency.
-uint32_t FilterPreference(const FilterSpec& fs, const ColumnSpec& col) {
+uint32_t FilterPreference(const FilterSpec& fs, const impl::Column& col) {
   enum AbsolutePreference : uint8_t {
     kIdEq,                     // Most efficient: id equality check
     kIdInequality,             // Id inequality check
```
aMayzner, line 49: Is there no specialised preference for SetIdCols?

src/trace_processor/dataframe/impl/query_plan.cc
```diff
@@ -121,14 +121,14 @@ void QueryPlanBuilder::Filter(std::vector<FilterSpec>& specs) {
...
-                            FilterPreference(b, b_col.spec);
+                     return FilterPreference(a, a_col) <
+                            FilterPreference(b, b_col);
                    });
 
   // Apply each filter in the optimized order
   for (FilterSpec& c : specs) {
     const Column& col = columns_[c.column_index];
-    const auto& ct = col.spec.column_type;
+    ColumnType ct = col.storage.column_type();
```
aMayzner: ColumnType shouldn't be a property of storage

src/trace_processor/dataframe/impl/query_plan.cc
```diff
@@ -296,7 +296,7 @@ void QueryPlanBuilder::NullConstraint(const NullOp& op, FilterSpec& c) {
   // the caller (i.e. SQLite) knows that we are able to handle the constraint.
   c.value_index = plan_.params.filter_value_count++;
 
-  switch (columns_[c.column_index].spec.nullability.index()) {
+  switch (columns_[c.column_index].overlay.nullability().index()) {
```
aMayzner: I don't like this, what does `nullability().index()` mean? Please increase the readability of this.

src/trace_processor/dataframe/impl/query_plan.cc
```diff
@@ -326,8 +326,8 @@ bool QueryPlanBuilder::TrySortedConstraint(
     const NonNullOp& op,
     const bytecode::reg::RwHandle<CastFilterValueResult>& result) {
   const auto& col = columns_[fs.column_index];
-  const auto& n = col.spec.nullability;
-  if (!n.Is<NonNull>() || col.spec.sort_state.Is<Unsorted>()) {
+  const auto& n = col.overlay.nullability();
```
aMayzner: Not n.

src/trace_processor/dataframe/impl/types.h
```diff
@@ -199,18 +210,22 @@ class Overlay {
...
   Variant data_;
 };
 
 // Combines column specification with storage implementation.
 // Represents a complete column in the dataframe.
 struct Column {
-  ColumnSpec spec;  // Column specifications (name, type, etc.)
-  Storage storage;  // Physical storage for column data
-  Overlay overlay;  // Optional overlay data for special properties
+  std::string name;      // Column name
```
aMayzner: Those comments are not adding much. Please remove if that's the case

src/trace_processor/dataframe/BUILD.gn
```diff
@@ -16,9 +16,10 @@ import("../../../gn/test.gni")
 
 source_set("dataframe") {
   sources = [
+    "cursor.h",
     "dataframe.cc",
     "dataframe.h",
-    "cursor.h",
+    "dataframe_builder.h",
```
aMayzner: maybe just builder? It's already both in dataframe folder and namespace

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+// This class allows constructing a `Dataframe` incrementally. It infers
+// column types (`int64_t`, `double`, `StringPool::Id`) based on the first
+// non-null value encountered in each column. Null values are tracked
+// efficiently using a `BitVector` (created only if nulls exist), and the
+// underlying data storage only stores non-null values (SparseNull
+// representation).
+//
+// Upon calling `Build()`, the builder analyzes the collected data to:
+// - Determine the final optimal storage type for integer columns (downcasting
+//   `int64_t` to `uint32_t` or `int32_t` if possible, or using `Id` type).
```
aMayzner, line 58: Not `setid` type?

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+  //
+  // Implementation wise, the collected data for each column is analyzed to:
+  // - Determine the final optimal storage type (e.g., downcasting int64_t to
+  //   uint32_t/int32_t if possible, using Id type if applicable).
+  // - Determine the final nullability overlay (NonNull or SparseNull).
+  // - Determine the final sort state (IdSorted, SetIdSorted, Sorted, Unsorted)
+  //   by analyzing the collected non-null values.
+  // - Construct and return the final `Dataframe` instance.
+  base::StatusOr<Dataframe> Build() && {
+    if (!current_status_.ok()) {
```
aMayzner: RETURN_IF_ERROR?

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+  //    (i.e. Id, uint32_t, int32_t) initially as int64_t, with potential
+  //    downcasting occurring during Build().
+  // 2) Tracks null values sparsely: only non-null values are appended to the
+  //    internal data storage vectors. A BitVector is created and maintained
+  //    only if null values are encountered for a column.
+  // 3) Performs strict type checking against the inferred type for subsequent
+  //    rows. If a type mismatch occurs, sets an error status (retrievable via
+  //    status()) and returns false.
+  template <typename ValueFetcherImpl>
+  bool AddRow(ValueFetcherImpl* fetcher) {
```
aMayzner, line 142: Shouldn't you check somewhere that row being added has the same length as the column_names_? Also, maybe it would be more readable if, rather that doing column_names_.size() constantly, you would cache col_count_ at the beginning

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+      default:
+        current_status_ = base::ErrStatus(
+            "Type mismatch in column '%s' at row %u. Existing type != "
+            "fetched type.",
+            column_names_[i].c_str(), row_count_);
+        return nullptr;
+    }
+  }
+
+  static impl::Storage CreateIntegralStorage(impl::FlexVector<int64_t> data,
```
aMayzner: Integral?

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+    if (IsInBounds<uint32_t>(min, max)) {
+      return impl::Storage{impl::Storage::Uint32{Downcast<uint32_t>(data)}};
+    }
+    if (IsInBounds<int32_t>(min, max)) {
+      return impl::Storage{impl::Storage::Int32{Downcast<int32_t>(data)}};
+    }
+    return impl::Storage{impl::Storage::Int64{std::move(data)}};
+  }
+
+  static impl::Overlay GetOverlay(std::optional<impl::BitVector> null_overlay) {
```
aMayzner: CreateOverlayFromBitVector?

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+    if (null_overlay) {
+      return impl::Overlay{impl::Overlay::SparseNull{*std::move(null_overlay)}};
+    }
+    return impl::Overlay{impl::Overlay::NoOverlay{}};
+  }
+
+  template <typename T>
+  static bool IsInBounds(int64_t min, int64_t max) {
+    return min >= std::numeric_limits<T>::min() &&
+           max <= std::numeric_limits<T>::max();
```
aMayzner, line 343: Do we want to have max inclusive? It's quite different than usual approach. If yes, please add a comment that we are looking inclusive bounds

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+
+  static impl::Overlay GetOverlay(std::optional<impl::BitVector> null_overlay) {
+    if (null_overlay) {
+      return impl::Overlay{impl::Overlay::SparseNull{*std::move(null_overlay)}};
+    }
+    return impl::Overlay{impl::Overlay::NoOverlay{}};
+  }
+
+  template <typename T>
+  static bool IsInBounds(int64_t min, int64_t max) {
```
aMayzner: This is very unreadable to me. It should be sth like: Bounds in Type Numeric Limits. I would have never guessed that 
```
uint32_t s = 10;
uint32_t e = 20;
IsInBounds(10, 20); 
```
would mean that we check if 10 is bigger than numeric limits min() and 20 lower than numeric limits max().

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+    return min >= std::numeric_limits<T>::min() &&
+           max <= std::numeric_limits<T>::max();
+  }
+
+  template <typename T>
+  static impl::FlexVector<T> Downcast(const impl::FlexVector<int64_t>& data) {
```
aMayzner: Hmm... What if we can't downcast? We will never figure out the error. Please at least add a comment mentioning that. And call this function `DowncastFromInt64`

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+    auto res = impl::FlexVector<T>::CreateWithSize(data.size());
+    for (uint32_t i = 0; i < data.size(); ++i) {
+      res[i] = static_cast<T>(data[i]);
+    }
+    return res;
+  }
+
+  static constexpr SortState GetDoubleSortState(
```
aMayzner: GetSortStateForDoubles or sth like that

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+      bool is_nullable,
+      const impl::FlexVector<double>& data) {
+    if (is_nullable) {
```
aMayzner, line 359: This kinda feels like this check should live outside of this function, and this function should only check if `data` is sorted.

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+    }
+    for (uint32_t i = 1; i < data.size(); ++i) {
+      if (data[i - 1] > data[i]) {
+        return SortState{Unsorted{}};
+      }
+    }
+    return SortState{Sorted{}};
+  }
+
+  static SortState GetStringSortState(
```
aMayzner, line 370: As above.

src/trace_processor/dataframe/runtime_dataframe_builder.h
```diff
@@ -0,0 +1,412 @@
...
+        if (prev > curr) {
+          return SortState{Unsorted{}};
+        }
+        prev = curr;
+      }
+    }
+    return SortState{Sorted{}};
+  }
+
+  static SortState GetIntegeralSortState(bool is_nullable,
```
aMayzner: This function is significantly different to the above functions, as we don't check if data is sorted or not, and we just get this as arguments to this function. Please change the name to cover this.

src/trace_processor/dataframe/type_set.h
```diff
@@ -132,6 +133,18 @@ class TypeSet {
     return idx;
   }
 
+  // Equality operator for TypeSet.
+  constexpr bool operator==(const TypeSet& other) const {
```
aMayzner: That's not what I would think an equality would be. Why only the active type, not any other type? Plus, it's a stupid question, if the TypeSets would be created with different orders of arguments, wouldn't it mean that even if the active type is the same, they might have different indexes?

src/trace_processor/perfetto_sql/engine/dataframe_module.cc
```diff
@@ -66,17 +66,8 @@ int DataframeModule::Connect(sqlite3* db,
...
-  // TODO(lalitm): take this from the engine rather than hardcoding it here.
-  auto df = dataframe::Dataframe(
-      {CS{"id", dataframe::Id(), dataframe::IdSorted(), dataframe::NonNull()}},
-      &pool);
-  for (int i = 0; i < 10000; ++i) {
-    df.InsertRow<ConstantValueFetcher>();
-  }
-  std::unique_ptr<Vtab> res =
-      std::make_unique<Vtab>(Vtab{{}, &pool, std::move(df)});
+  // TOOO(lalitm): fix this.
```
aMayzner: Fix what.

src/trace_processor/dataframe/BUILD.gn
LalitMaganti: For codesearching, builder.h is too generic, will match unnecesarily broad range of things.

src/trace_processor/dataframe/runtime_dataframe_builder.h
LalitMaganti: That's covered in the next line.

src/trace_processor/dataframe/runtime_dataframe_builder.h
LalitMaganti: > Shouldn't you check somewhere that row being added has the same length as the column_names_? 

It's up to the implementation of ValueFetcherImpl to do bounds checking

> Also, maybe it would be more readable if, rather that doing column_names_.size() constantly, you would cache col_count_ at the beginning

Can you explain how so? It's used in a single place in the loop. What exactly would be more readable?

src/trace_processor/dataframe/runtime_dataframe_builder.h
LalitMaganti: This makes the Build significantly more complicated.

src/trace_processor/dataframe/runtime_dataframe_builder.h
LalitMaganti: This makes the Build significantly more complicated.

src/trace_processor/dataframe/type_set.h
LalitMaganti: You're missing the fact that TypeSet is guaranteed to be identical in the types it can represent by the compiler. You can omit the template parameter for class functions and it will be the same parameter as the instance of the templated class.

I've added it explicitly to make that more clear.

src/trace_processor/dataframe/impl/bit_vector.h
```diff
@@ -51,13 +51,26 @@ struct BitVector {
...
+  BitVector& operator=(BitVector&&) = default;
+  BitVector(const BitVector&) = delete;
+  BitVector& operator=(const BitVector&) = delete;
+
   // Allocates a new BitVector with `size` unset bits.
   //
   // size: Size for how many bits to add to the BitVector.
-  // Returns a BitVector with the given size and with all bits set to false.
-  static BitVector CreateWithSize(uint64_t size) {
+  // Returns a BitVector with the given size and with all bits set to ``value`.
```
aMayzner: Remove one `

aMayzner (approved):

PR #1155, opened 2025-04-03
Add slow start insights track with slices of interest

haipingyang71:
Change-Id: Ieb1ccae6e8a5763f6a08839e168a0166e37ec480

Bug: 376308737

Test: ui/run-dev-server with ptraces

UI snapshot:  https://screenshot.googleplex.com/6Fqk5FHFtDJKnwb

### DIFF
--- a/ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
@@ -1,0 +1,747 @@
+// Copyright (C) 2024 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Trace} from '../../public/trace';
+import {STR, LONG, NUM} from '../../trace_processor/query_result';
+import {createQuerySliceTrack} from '../../components/tracks/query_slice_track';
+import {TrackNode} from '../../public/workspace';
+
+// The container that keeps track of packages that have slow start reason.
+interface Startup {
+  // The startup id.
+  id: number;
+  // The package name.
+  package: string;
+  // Time start
+  ts: bigint;
+  // Time end
+  ts_end: bigint;
+}
+
+// The slices for slow start reasons.
+interface Slice {
+  // Time start
+  ts: bigint;
+  // Duration
+  dur: bigint;
+  // Sice name
+  name : string;
+}
+
+// The log tag
+const tag = 'SlowStartInsights';
+
+/**
+ * Returns a track node that contains slow start status
+ * for the packages that started up in a trace.
+ * @param trace The loaded trace.
+ * @returns a track node with the slow start status.
+ * `undefined` if there are no app startups detected.
+ */
+export async function slowInsightsTrack(
+  trace: Trace,
+): Promise<TrackNode | undefined> {
+  const startups: Array<Startup> = [];
+
+  // Find app startups
+  let result = await trace.engine.query(
+    `
+        INCLUDE PERFETTO MODULE android.startup.startups;
+        SELECT startup_id AS id, package, ts, ts_end FROM android_startups;`,
+    tag,
+  );
+
+  const it = result.iter({id: NUM, package: STR, ts: LONG, ts_end: LONG});
+  for (; it.valid(); it.next()) {
+    startups.push({
+      id: it.id,
+      package: it.package,
+      ts: it.ts,
+      ts_end: it.ts_end,
+    });
+  }
+
+  if (startups.length === 0) {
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
```diff
@@ -0,0 +1,747 @@
...
+async function getOpenDexFilesFromOatTrack(
+  trace: Trace,
+  startups: Startup[],
+): Promise<TrackNode | undefined> {
+  // The log tag
+  let tag = 'OpenDexFilesFromOat';
+  let slice_glob = `'OpenDexFilesFromOat*'`;
+  let result = await trace.engine.query(
+    `
+    SELECT RUN_METRIC('android/startup/slow_start_thresholds.sql');
```
LalitMaganti, line 502: Using RUN_METRIC in plugins is not allowed. There should be a presubmit check warning you about this if you did tools/install-build-deps. Please port all relevant things you need to the standard library first before trying to use them here.

LalitMaganti: This applies here and for all other usages of RUN_METRIC in this class.

LalitMaganti (changes requested):
Requesting changes as above.

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
```diff
@@ -0,0 +1,747 @@
+// Copyright (C) 2024 The Android Open Source Project
```
stevegolton, line 1: Nit: 2025

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
stevegolton, line 1: We use camel_case for file names. This should be `slow_insights.ts`

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
```diff
@@ -0,0 +1,747 @@
...
+const tag = 'SlowStartInsights';
+
+/**
+ * Returns a track node that contains slow start status
+ * for the packages that started up in a trace.
+ * @param trace The loaded trace.
+ * @returns a track node with the slow start status.
+ * `undefined` if there are no app startups detected.
+ */
+export async function slowInsightsTrack(
```
stevegolton, line 52: If you already have the data already available in a table/view, you should just create the track directly from that by crafting the correct SQL query instead of pulling it all out and injecting it back in again.

I.e.
```
const track = await createQuerySliceTrack({
    trace: trace,
    uri: uri,
    data: {
      sqlSource: `
        SELECT
          startup_id AS id,
          ts,
          ts_end - ts as dur
        FROM android_startups
      `,
      columns: ['ts', 'dur', 'name'],
    },
  });
```

You only need to inject the data if you don't have it available in trace processor already.

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
```diff
@@ -0,0 +1,747 @@
...
+  trace: Trace,
+): Promise<TrackNode | undefined> {
+  const startups: Array<Startup> = [];
+
+  // Find app startups
+  let result = await trace.engine.query(
```
stevegolton, line 58: I think most if not all of these `let`s can be `const`s. Eslint should complain if lets are not reassigned.

Our presubmit checks are still up in the air since the move to github, but you should be able to run eslint with `ui/format-sources`.

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
```diff
@@ -0,0 +1,747 @@
...
+  return getSlowReasonTrack(trace, slices, uri, title);
+}
+
+async function getBroadcastReceivedTrack(
+  trace: Trace,
+  startups: Startup[],
+): Promise<TrackNode | undefined> {
+  // The log tag
+  let tag = 'Broadcast received';
+  let slice_glob = `'broadcastReceiveReg*'`;
```
stevegolton, line 443: Variables should be in `snakeCase`. There's quite a few instances of this here.

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
```diff
@@ -0,0 +1,747 @@
...
+  threshold_pct = Number(it.threshold_pct);
+
+  let slices = await getMainThreadSlices(trace, tag, slice_glob, 0, threshold_pct, startups);
+
+  const uri = `/android_startups/monitor_contention`;
+  const title = `Monitor Contention`;
+  return getSlowReasonTrack(trace, slices, uri, title);
+}
+
+async function getBindApplicationTrack(
```
stevegolton, line 593: In general there's a lot of duplicated code in this file that makes it very difficult to understand exactly what you're trying to do. Could you remove some of this duplication please?

stevegolton (changes requested):

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
haipingyang71: Done.

ui/src/plugins/dev.perfetto.AndroidStartup/slowInsights.ts
haipingyang71: Done

PR #1160, opened 2025-04-03
Perfetto Android Java SDK: extract 'PerfettoTrackEventBuilder'.

KirillTim:
We decouple the event builder from the native object wrappers, that simplifies the code and allows us to enable the debug asserts in builder at runtime.

Tested:
```
```

### DIFF
--- a/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
+++ b/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
@@ -38,7 +38,7 @@ public final class PerfettoTrace {
   private static final int PERFETTO_TE_TYPE_INSTANT = 3;
   private static final int PERFETTO_TE_TYPE_COUNTER = 4;
 
-  private static boolean isDebug = false;
+  private static boolean sIsDebug = false;
 
   static class NativeAllocationRegistry {
     public static NativeAllocationRegistry createMalloced(
@@ -198,7 +198,7 @@ public final class PerfettoTrace {
    * @param eventName The event name to appear in the trace.
    */
   public static PerfettoTrackEventBuilder instant(Category category, String eventName) {
-    return PerfettoTrackEventBuilder.newEvent(PERFETTO_TE_TYPE_INSTANT, category, isDebug)
+    return PerfettoTrackEventBuilder.newEvent(PERFETTO_TE_TYPE_INSTANT, category, sIsDebug)
         .setEventName(eventName);
   }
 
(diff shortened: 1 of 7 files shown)


src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrackEventBuilder.java
```diff
@@ -0,0 +1,679 @@
...
+  }
+
+  private PerfettoTrackEventBuilder(boolean isCategoryEnabled, PerfettoTrackEventBuilder parent) {
+    mIsCategoryEnabled = isCategoryEnabled;
+    if (!mIsCategoryEnabled) {
+      // No fields of this builder will be used, no need to initialize them.
+      return;
+    }
+    if (parent == null) {
+      // Crete a root builder, saved into thread-local variable.
```
zezeozue: Is this a stale comment?

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -38,6 +38,8 @@ public final class PerfettoTrace {
   private static final int PERFETTO_TE_TYPE_INSTANT = 3;
   private static final int PERFETTO_TE_TYPE_COUNTER = 4;
 
+  private static boolean isDebug = false;
```
zezeozue: nit: sIsDebug (android framework convention)

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrackEventBuilder.java
KirillTim: No, it isn't.

For the builder class we have a two step initialization process:
1. Call the class constructor
2. Call `initNewEvent` or `initChildBuilderForProto` on the existing object.

The purpose of these comments in the constructor is to highlight why do we have three different code branches.

If the comment is confusing and make more harm than good then I can delete it, should I?

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
KirillTim: Will fix, thanks!
Side question: should we follow android framework code convention after migration from framework repo? 🤔

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrackEventBuilder.java
zezeozue: Maybe rephrase to: (we are) Creating a root builder which will be saved in thread local storage.

I misread it as an additional action that will happen in the following line. (also typo on 'create').

The comment is useful. Please keep

zezeozue (approved):

ddiproietto (approved):

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
zezeozue: I guess we don't have to. But we should at least be consistent

PR #1163, opened 2025-04-03
CI: skip linux/android/bazel tests on docs-only changes

primiano:


LalitMaganti (approved):

.github/workflows/analyze.yml
```diff
@@ -45,7 +45,15 @@ jobs:
         with:
           fetch-depth: 2
 
-      - name: Check if the change is UI/docs only (skip linux/android tests)
+      - name: Check if the change is trivial (docs/infra only)
+        run: |
+          if !(git diff --name-only HEAD^1 HEAD | egrep -qv '^(docs|infra|tools)/'); then
```
LalitMaganti: Are you sure this is a good diff. what if there are n commits?

LalitMaganti (approved):

.github/workflows/analyze.yml
primiano: Ah you are right. I copied from the existing one, but it's also wrong. I think the new version should be fixed.

PR #1166, opened 2025-04-04
Detect, ignore, and log weird results from SuspendControlService.

simonmacm:
Bug: 404239369

We're seeing rare, strange results in traces that are consistent with SuspendControlService rolling back values or returning 0. If we see this occurring, ignore what it told us and increment a counter.

### DIFF
--- a/src/kernel_utils/kernel_wakelock_errors.h
+++ b/src/kernel_utils/kernel_wakelock_errors.h
@@ -1,0 +1,25 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_KERNEL_UTILS_KERNEL_WAKELOCK_ERRORS_H_
+#define SRC_KERNEL_UTILS_KERNEL_WAKELOCK_ERRORS_H_
+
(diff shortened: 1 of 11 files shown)


src/kernel_utils/kernel_wakelock_errors.h
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_KERNEL_UTILS_KERNEL_WAKELOCK_ERRORS_H_
+#define SRC_KERNEL_UTILS_KERNEL_WAKELOCK_ERRORS_H_
+
+#include <cinttypes>
+
+constexpr uint64_t kKernelWakelockErrorZeroValue = 1;
```
rsavitski, line 22: `inline constexpr uint64_t ...`

An alternative would be to use protobuf enums. But I think this is ok as well.

src/traced/probes/android_kernel_wakelocks/android_kernel_wakelocks_data_source.cc
```diff
@@ -167,10 +169,17 @@ void AndroidKernelWakelocksDataSource::WriteKernelWakelocks() {
...
   for (auto it = totals.GetIterator(); it; ++it) {
     KernelWakelockInfo* info = wakelocks_.Find(it.key());
-
     uint64_t total = it.value();
-    if (it.value() != info->last_value) {
-      uint64_t last_value = info->last_value;
+    uint64_t last_value = info->last_value;
+
+    if (total == 0) {
+      error_flags &= kKernelWakelockErrorZeroValue;
```
rsavitski: Should be `=` or `|=`.

rsavitski (changes requested):
Sorry, missed this, still learning how to navigate github.

src/traced/probes/android_kernel_wakelocks/android_kernel_wakelocks_data_source.cc
simonmacm: Oops, thanks for the catch. Done.

src/kernel_utils/kernel_wakelock_errors.h
simonmacm: OK, I'll leave as is.

PR #1169, opened 2025-04-04
tp: fix handling of begin/rollback for sqlite modules

LalitMaganti:


src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h
```diff
@@ -300,6 +300,12 @@ class PerfettoSqlEngine {
                                  const std::string& key,
                                  const PerfettoSqlParser&);
 
+  // Called when a transaction is committed.
```
aMayzner: This comment is not verbose enough - please explain here (in a short way) what is the transaction and a commit. I know those are sqlite concepts, but i can only hope that some work on the engine can still be done if someone is not up to speed on the transactions/commits/rollbacks

src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h
```diff
@@ -300,6 +300,12 @@ class PerfettoSqlEngine {
...
+  int OnCommit();
+
+  // Called when a transaction is rolled back.
```
aMayzner: as above

src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.cc
```diff
@@ -1087,6 +1097,21 @@ base::Status PerfettoSqlEngine::ExecuteCreateMacro(
   return base::OkStatus();
 }
 
+int PerfettoSqlEngine::OnCommit() {
```
aMayzner, line 1100: why not return void at this point? (this code is already significantly different than the rest of the engine). Here and below

src/trace_processor/sqlite/module_lifecycle_manager.h
```diff
@@ -73,7 +75,7 @@ class ModuleStateManager {
     friend class ModuleStateManager<Module>;
 
     ModuleStateManager* manager;
-    bool disconnected = false;
+    bool create_committed = false;
```
aMayzner: We never before used the commit/transactions language here. Can you please add a small comment to this here?

src/trace_processor/sqlite/module_lifecycle_manager.h
```diff
@@ -134,8 +169,16 @@ class ModuleStateManager {
...
 
  private:
-  base::FlatHashMap<std::string, std::unique_ptr<PerVtabState>> state_by_name_;
+  using StateMap =
+      base::FlatHashMap<std::string, std::unique_ptr<PerVtabState>>;
+  StateMap state_by_name_;
+  StateMap destroyed_state_by_name_;
+};
+
+struct Foo {
```
aMayzner: ?

src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.cc
LalitMaganti: SQLite requires I return a value for this function. Either I can do it here or split the logic between here and the lambda. I figured it was best to do it all in one place.

aMayzner (approved):

PR #1171, opened 2025-04-04
tracing: remove logspam from stalling messages

LalitMaganti:
These are totally harmless logs but they confuse poeople a lot when they appear. See http://yaqs/1389768403857702912


rsavitski (approved):
Thank you, been meaning to send the same patch to Daniele :--)

PR #1172, opened 2025-04-04
Add unified code formatters (part 1)

primiano:
- Add unified archiecture to invoke code formatters
- All the code to detect changed files, --all etc is shared
- All the formatter behave in the same way
  1. If no option is passed, they format only files changed from upstream (or origin/main if no upstream is set)
  2. If a list of files is passed, they format the list of files passed
  3. If --all is passed, they format all the versioned files
- Introduce a unified tools/format-sources which invokes them all

What's missing (and upcoming in next CLs):
- Invoke formatters from tools/run_presubmit
- Merge the SQL formatter
- YAPF / Python formatter (which today depends on depot_tools)


python/tools/code_format_clang.py
```diff
@@ -0,0 +1,43 @@
...
+
+from code_format_utils import CodeFormatterBase, run_code_formatters
+
+
+class ClangFormat(CodeFormatterBase):
+
+  def __init__(self):
+    super().__init__(
+        name='clang-format',
+        exts=['.cc', '.cpp', '.c', '.h', '.inc', '*.java', '*.proto'])
```
LalitMaganti, line 27: Simon was complaining that clang-format was fighting with gen_merged_protos. Were we also using clang-format before and if so why were we not getting errors?

LalitMaganti (approved):

python/tools/code_format_clang.py
primiano: Ok i'll check this in the next cl when i update presubmits.

PR #1176, opened 2025-04-06
tp: make sure all virtual table modules receive Commit/Rollback hooks

LalitMaganti:
This CL ensures that the commit/rollback hooks is plumbed through to every virtual table module in trace processor, not just the ones which are considered part of the engine.

To do this, we needed to rework a little bit how the ModuleStateManager works to be an object you inherit from instead of one you put in your context. We also need to extract out a base class from ModuleStateManager so the engine can keep a hold of the type erased pointers (with the added benefit of less binary size).

In the process, we remove several hacky use of sqlite_engine() now that RegisterVirtualTableModule is directly plumbed into the PerfettoSqlEngine.


src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h
```diff
@@ -92,6 +94,50 @@ class PerfettoSqlEngine {
   base::StatusOr<SqliteEngine::PreparedStatement> PrepareSqliteStatement(
       SqlSource sql);
 
+  // Registers a virtual table module with the given name.
```
aMayzner, line 97: Why is this comment formatted this way? The space between variable name and contents is very wide

aMayzner (approved):

src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h
LalitMaganti: Done

PR #1177, opened 2025-04-06
tp: remove dependency of span join and window on ModuleStateManager

LalitMaganti:
This CL removes the complexity of managing state from these operators: 1) SpanJoin: span join's arguments uniquely identify everything about
   how it should work so there's no reason to have some extra state.
   The only thing we save is on a Connect call, not having to run all
   the code for creation again but frankly this happens very rarely
   and it's not worth all the subtle issues which crop up because of
   having to manage state.
2) Window: window only requires state because it supports an update
   based interface. In the old days, window was a very important
   table as it was used by the UI to control what should be rendered.
   However, with the mipmap operators, this is not necessary anymore
   and window only exists really for testing purposes with zero
   uses in production. Therefore, we make it readonly with all
   arguments passed in the constructor.


stevegolton (approved):
LGTM

PR #1179, opened 2025-04-07
Add slow start insights track

haipingyang71:
Change-Id: I797f65ca6092edccc98a441d5879b648c19438bd

Bug: 376308737

Test: ui/run-dev-server

UI snapshot: https://screenshot.googleplex.com/B5TSyBxDNNp7WDM

### DIFF
--- a/ui/src/plugins/dev.perfetto.AndroidStartup/slow_insights.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidStartup/slow_insights.ts
@@ -1,0 +1,831 @@
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.AndroidStartup/slow_insights.ts
```diff
@@ -0,0 +1,831 @@
...
+      selected = true;
+      sliceId = slices.length;
+    }
+    addResultToSlices(result, slices);
+  }
+
+  const uri = `/android_startups/unlock_running`;
+  const title = `Unlock Running`;
+  return await 
+    getSlowReasonTrackAndSelect(trace, slices, uri, title, selected, sliceId);
```
stevegolton, line 228: Again, I wouldn't pull the slices out of TP just to push them back in again for the purpose of creating the track.

Please create the track directly from the query (here and all functions below).

haipingyang71: Since slow start reason tracks will be added only if the query results are not empty and createQuerySliceTrack() does not make the result of the sql query available, creating the track from query directly would require to run the query and check the result first before calling createQuerySliceTrack(). That means some queries have to run twice, is that how we want to proceed here?

stevegolton: That's fine, the main thing is to avoid pulling all the track out into javascript.

PR #1180, opened 2025-04-07
Remove obsolete CheckBadCppPatterns from tools/run_presubmit

ddiproietto:
This was introduced in 0fd92a9496e0("service: disallow ReadBuffers for CLONE_SNAPSHOT traces"), but it never worked.

I checked the current complaints and they're all false positives.

I'm also removing the check from PRESUBMIT.py, even though this should not be used anymore.

### DIFF
--- a/tools/run_presubmit
+++ b/tools/run_presubmit
@@ -682,34 +682,6 @@ def CheckBannedCpp(changed_files: List[str]) -> List[str]:
   return errors
 
 
-def CheckBadCppPatterns(changed_files: List[str]) -> List[str]:
-  # File Regex (for filtering), Code Regex (for checking), Message tuple
-  # File pattern needs to be fnmatch compatible
-  bad_patterns: List[Tuple[str, str, str]] = [
-      (
-          '*/tracing_service_impl.cc',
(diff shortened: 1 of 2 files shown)


tools/run_presubmit
```diff
@@ -682,34 +682,6 @@ def CheckBannedCpp(changed_files: List[str]) -> List[str]:
   return errors
 
 
-def CheckBadCppPatterns(changed_files: List[str]) -> List[str]:
```
LalitMaganti, line 685: I think we still want to keep this check we should just turn it into a rachet (count how many instances there are and then make sure it doesn't increase).

It's still the case we should *not* be using trigger_config in the service except for implmenting GetTriggerMode itself.

LalitMaganti (changes requested):

tools/run_presubmit
ddiproietto: No, this is too complicated.

Also, we can use `trigger_config()`, we're using it correctly in many places. The presubmit check was wrong to begin with. This, in my opinion is proof of the fact that stuff like this doesn't belong in presubmits.

LalitMaganti (approved):
OK going to leave it to your judgement :)

PR #1185, opened 2025-04-07
tprobes: generic events as self-describing protos (1/3)

rsavitski:
Preliminary fixups, split out for reviewers' convenience.


protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -154,6 +154,12 @@ message FtraceConfig {
   // For more context: go/rss-stat-throttled.
   optional bool throttle_rss_stat = 15;
 
+  // If true, use self-describing proto messages when writing events not known
+  // at compile time (aka generic events). Each event bundle will have a set of
+  // serialised proto descriptors for events within that bundle.
+  // Added in: perfetto v50.
+  optional bool denser_generic_event_encoding = 32;
```
LalitMaganti: Is the plan to make this true by default when it's more stable?

src/traced/probes/ftrace/cpu_reader.h
```diff
@@ -166,6 +177,9 @@ class CpuReader {
     const bool compact_sched_enabled_;
     CompactSchedBuffer* const compact_sched_buf_;
     uint64_t initial_previous_bundle_end_ts_;
+    // Keyed by proto field id within |FtraceEvent|.
+    base::FlatSet<uint32_t> generic_descriptors_to_write_;
+    const std::map<uint32_t, std::vector<uint8_t>>* generic_pb_descriptors_;
```
LalitMaganti: Should this be unordered_map or flathashmap?

src/traced/probes/ftrace/cpu_reader_fuzzer.cc
```diff
@@ -50,17 +50,20 @@ void FuzzCpuReaderProcessPagesForDataSource(const uint8_t* data, size_t size) {
   memcpy(g_page, data, std::min(size_t(base::GetSysPageSize()), size));
 
   FtraceMetadata metadata{};
```
LalitMaganti: There seems a lot of weird diffs here because of code formatting. Can we do a format pass first?

LalitMaganti (changes requested):
Right now it's very hard to review because of the number of different changes in this CL. Can we split into three pieces:
1) formatting pass
2) refactoring only
3) actual changes

Overall though seems very reasonable direction.

LalitMaganti (changes requested):
Sending to make it clearer

PR #1186, opened 2025-04-07
Fix proto formatter, roll clang-format add ignores

primiano:
- Fix a small bug I introduced in format-cpp-sources which would ignore proto and java files (by having an accidental *)
- Roll clang-format to the version used by chromium. Among the various things, this supports .clang-format-ignore files
- Add a few known files to the ignore list. Other might follow in a follow up cl.


tools/run_presubmit
```diff
@@ -288,7 +275,7 @@ def CheckPatchFormatted(changed_files: List[str],
   try:
     # Use --dry-run or -n, and -Werror to exit with error if changes are needed
     cmd = ['clang-format', '--dry-run', '-Werror'] + abs_files_to_format
-    run_command(cmd)
+    print(cmd)
```
LalitMaganti: This seems unintentional?

LalitMaganti (approved):

tools/run_presubmit
primiano: yes fixed

LalitMaganti (approved):

PR #1194, opened 2025-04-08
ui: Add flag to patch around hotkey collisions in dvorak

stevegolton:
This PR adds a 'Dvorak keyboard layout' flag which disables:

- '.' hotkey to go to the next slice
- ',' hotkey to go to the previous slice

This approach is a quick fix. The proper fix is to allow users to configure hotkeys, but we're a long way off this.


LalitMaganti:
Can I ask if you can quickly add a settings page and make this the first option there? This is a classic case of "if you do the first step, every other step will be a lot easier". 

Feel free not to though I'll approve.

stevegolton:
@LalitMaganti Normally I would agree with you (e.g. the status bar), but seeing as these options will be burned into localstorage, it makes things somewhat more difficult to change once we've started, so I'd rather not rush things.

PR #1202, opened 2025-04-08
Redaction - Use ftrace event `pid` with rename events

vaage:
Before this, redaction relied on the `pid` value in the rename task and asserted its presence. This broke when Perfetto stopped emitted the field. The field was removed because the rename task's `pid` would match the ftrace event's `pid`.

To safely remove the assertion, redaction will now use:

```
return rename_task.has_pid() ? rename_task.pid() : ftrace_event.pid()
```

### DIFF
--- a/src/trace_redaction/redact_process_events.cc
+++ b/src/trace_redaction/redact_process_events.cc
@@ -107,6 +107,13 @@ base::Status RedactProcessEvents::OnFtraceEvent(
                            protos::pbzero::FtraceEvent::kTimestampFieldNumber);
   }
 
+  auto pid = decoder.FindField(protos::pbzero::FtraceEvent::kPidFieldNumber);
+
+  if (!pid.valid()) {
+    return base::ErrStatus("RedactProcessEvents: missing pid %d",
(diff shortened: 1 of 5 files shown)


src/trace_redaction/redact_process_events.cc
```diff
@@ -265,64 +278,62 @@ base::Status RedactProcessEvents::OnProcessRename(
...
 
-  auto pid = decoder.pid();
   auto new_comm = decoder.newcomm();
   auto old_comm = decoder.oldcomm();
   auto oom_score_adj = decoder.oom_score_adj();
 
+  // The rename task's pid *should* always match the ftrace event's pid. To
+  // support backwards compatibility but assume the ftrace event's pid can be
+  // used, the rename task's pid will be used if it is present, otherwise it'll
+  // use the ftrace event's pid.
```
LalitMaganti, line 308: Please add a link here to the bug you linked to me over chat.

LalitMaganti (approved):

LalitMaganti (approved):

stevegolton (approved):

PR #1215, opened 2025-04-09
prevent tabnabbing attack with rel="noopener noreferrer"

jaworek:
Fix potential tabnabbing attack by adding `rel="noopener noreferrer"` to `a` tags with `target="_blank"`

### DIFF
--- a/ui/src/assets/bigtrace.html
+++ b/ui/src/assets/bigtrace.html
@@ -35,11 +35,11 @@ Please file a bug (details below) and try these remediation steps:
 
 * <a href="javascript:clearAllCaches();">Clear all the site storage and caches</a> and reload the page.
 
-* Clear the site data and caches from devtools, following <a target="_blank" href="https://developers.google.com/web/tools/chrome-devtools/storage/cache#deletecache">these instructions</a>.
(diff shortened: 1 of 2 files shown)


primiano:
Can you expand on how this attack would work in this scenario? it's not obvious to me.

Also this CL seems very arbitrary. you picked 2 links and added noreferrer... how about all the other links in the codebase? Why these two are special?

i'm very surprised that, among all the possible links, you picked bigtrace.html. that file is really unused.

Is this some AI generated PS?

primiano:
I'm doing a pass on my CLs.
This CL honestly mainly adds clutter with little value:

1) All the URLs that you are touching are trusted, we don't open arbitrary pages with target=_blank, which would be the problematic bit here.

2) As you linked, this has been long fixed in modern browsers. Browsers are generally very good at updating, and if somebody runs an outdated browser, ui.perfetto.dev is their very last problem

3) I'm quite convinced Chrome had cross origin isolation across window.opener for a loooong time now, far earlier than 2023.

4) This change is inconsistent beacuse if you codesearch for `target: 'blank'` you'll see far more instances in the codebase than these two (most of them come from mithril) so it's unclear and confusing why we should only patch those two. and i would not patch the others because it's mostly noise at this point.

PR #1216, opened 2025-04-09
tracing_service_impl: Record stop_delay_ms for TracePacket.trigger

ddiproietto:
`stop_delay_ms` is mirrored from the config. We shouldn't read the config from trace_processor.

### DIFF
--- a/src/tracing/service/tracing_service_impl_unittest.cc
+++ b/src/tracing/service/tracing_service_impl_unittest.cc
@@ -100,6 +100,7 @@ using ::testing::InSequence;
 using ::testing::Invoke;
 using ::testing::InvokeWithoutArgs;
 using ::testing::IsEmpty;
+using ::testing::IsSupersetOf;
 using ::testing::Mock;
 using ::testing::Ne;
 using ::testing::NiceMock;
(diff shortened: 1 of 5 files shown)


src/tracing/service/tracing_service_impl_unittest.cc
```diff
@@ -938,6 +945,19 @@ TEST_F(TracingServiceImplTest, EmitTriggersWithStopTracingTrigger) {
       HasTriggerMode(protos::gen::TraceConfig::TriggerConfig::STOP_TRACING));
   EXPECT_THAT(GetReceivedTriggers(packets),
               UnorderedElementsAre("trigger_name", "trigger_name_3"));
```
rsavitski, line 947: nit: remove this expectation since it's covered by the new block?

rsavitski (approved):

rsavitski:
Also note the failing tests, at least PerfettoCmdlineTest.MultipleTriggersCloneSnapshot.

src/tracing/service/tracing_service_impl_unittest.cc
ddiproietto: This still has some value: we check that there are only two triggers.

ddiproietto:
Thanks, fixed the tests

PR #1217, opened 2025-04-09
tracing_service_impl: Record stop_delay_ms for TracePacket.clone_snapshot_trigger

ddiproietto:
`stop_delay_ms` is mirrored from the config. We shouldn't read the config from trace_processor.


rsavitski (approved):
Nit: the PR description seems like it was torn during copy-pasting. Please reword both the title and the body of the patch description before merging.

LalitMaganti (commented):
Something seems wrong with the PR description. Can we please fix?

Also this does not seem to be connected to anything? e.g. where are we actually computing this value to put here?

ddiproietto:
I update the PR description, thanks.

> Also this does not seem to be connected to anything? e.g. where are we actually computing this value to put here?

The value comes from `TriggerInfo::trigger_delay_ms`, which was already populated on main.

In this latest push, I updated some field names to remove `mono_`.

LalitMaganti (approved):

PR #1225, opened 2025-04-09
tprobes: generic events as self-describing protos (2/3)

rsavitski:
The encoding of ftrace events that do not have a protobuf description in
the perfetto repo is very space inefficient. This is because every
GenericFtraceEvent proto has a string description of the field names and
added layers of proto nesting. This was good enough for local
exploration, but is often used by accident when configs use category
globs. Also, long-term it'd be nice to move away from requiring protos
to be checked into our repo outside of truly well-known cases.

This patch adds a feature flag to enable such generic events to be
serialised using self-describing protos, with the aim of being a bit
more space efficient. The FtraceEvent proto will have a submessage under
a field id not covered by the compile-time oneof fields (similar to
proto extensions). The FtraceEventBundle will have an
id->ProtoDescriptor map, once per generic event type.

Notes:
* every unknown event type will have its own proto field id, and a
  bundle-level proto descriptor.
* the mapping stays consistent within all bundles in  a tracing session.
* there is no global mapping of event->id across traces.
* we still use the FtraceEvent container proto, with the upsides of
  being able to do streaming serialisation of bundles, and the downsides
  of protobuf encoding overheads due to deep nesting (and extra missing
  bits like delta timestamps).
* we do not use incremental state, every bundle describes the generic
  events. The space cost is amortised by bundles being relatively
  infrequent.
* the new encoding is behind a flag, and the default case continues to
  be GenericFtraceEvent protos.

//////////////
For reviewers:

This might help visualise the proto structure:
  message FtraceEvent {
    optional uint64 timestamp = 1;
    optional uint32 pid = 2;

    oneof event {
      PrintFtraceEvent print = 3;
      SchedSwitchFtraceEvent sched_switch = 4;
      ...
    }

    // synthesised at runtime, similar to an extension range:
    optional GenericOne unknown_event_one = 65536;
    optional GenericTwo unknown_event_two = 65537;
  }

I started by reserving a high range of field ids within the FtraceEvent
proto for the synthesised-on-the-fly submesssages.  Would've been nice
to use something with one fewer byte for pb tags (e.g. 1k-2k), but that
is a bit too close to the compile-time proto range, which already goes
up to ~600. The field id range does matter since we use "id>65k" as a
way of quickly identifying that an event is generic within the
proto_translation_table and cpu_reader.

proto_translation_table: I changed the |events_| mapping (from ftrace id
to a struct describing the event) to generate proper per-field proto
descriptions of events, and moved the GenericFtraceEvent specifics to
cpu_reader's serialisation path.

cpu_reader::Bundler changes: when parsing, we need to remember the set
of referenced generic events to emit their descriptors.

### DIFF
--- a/src/traced/probes/ftrace/cpu_reader.cc
+++ b/src/traced/probes/ftrace/cpu_reader.cc
@@ -27,6 +27,7 @@
 #include "perfetto/ext/base/metatrace.h"
 #include "perfetto/ext/base/utils.h"
 #include "perfetto/ext/tracing/core/trace_writer.h"
+#include "perfetto/protozero/proto_utils.h"
 #include "src/kallsyms/kernel_symbol_map.h"
 #include "src/kallsyms/lazy_kernel_symbolizer.h"
 #include "src/traced/probes/ftrace/ftrace_config_muxer.h"
@@ -35,6 +36,7 @@
 #include "src/traced/probes/ftrace/ftrace_print_filter.h"
 #include "src/traced/probes/ftrace/proto_translation_table.h"
 
+#include "protos/perfetto/common/descriptor.pbzero.h"
 #include "protos/perfetto/trace/ftrace/ftrace_event.pbzero.h"
 #include "protos/perfetto/trace/ftrace/ftrace_event_bundle.pbzero.h"
 #include "protos/perfetto/trace/ftrace/ftrace_stats.pbzero.h"  // FtraceParseStatus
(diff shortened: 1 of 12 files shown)


rsavitski:
Carrying over a comment from the other PR: yes, the intent is to make this the default (and hopefully the only implementation) eventually.

src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -348,6 +351,23 @@ void CpuReader::Bundler::StartNewPacket(
   }
 }
 
+void CpuReader::Bundler::WriteGenericEventDescriptors() {
+  if (!bundle_)
+    return;
+
+  for (uint32_t proto_id : generic_descriptors_to_write_) {
+    PERFETTO_DCHECK(generic_pb_descriptors_->Find(proto_id));
```
LalitMaganti, line 359: Should this not just be `PERFETTO_DCHECK(pb_descriptor)`

LalitMaganti (approved):

src/traced/probes/ftrace/cpu_reader.cc
rsavitski: wanted to keep the DCHECK out of the way of the main logic, the if-condition looked cluttered with it in the middle.

PR #1232, opened 2025-04-10
Allow users of DatasetSliceTrack to entirely override what's shown on the tooltip

stevegolton:
Allow users of DatasetSliceTrack to entirely override what's shown on the tooltip & add `category` if not null

### DIFF
--- a/ui/src/plugins/dev.perfetto.AndroidLog/logs_track.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidLog/logs_track.ts
@@ -77,7 +77,7 @@ export function createAndroidLogTrack(trace: Trace, uri: string) {
     colorizer: (row) => DEPTH_TO_COLOR[row.depth],
     // It would be nice to show the message on the tooltip too, but loading a
     // message for each event may balloon memory, so we just show the tag.
-    tooltip: (row) => (row.tag === null ? [] : [row.tag]),
+    tooltip: (slice) => slice.row.tag,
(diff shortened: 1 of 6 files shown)


ui/src/plugins/dev.perfetto.AndroidLog/logs_track.ts
```diff
@@ -77,7 +77,7 @@ export function createAndroidLogTrack(trace: Trace, uri: string) {
     colorizer: (row) => DEPTH_TO_COLOR[row.depth],
     // It would be nice to show the message on the tooltip too, but loading a
     // message for each event may balloon memory, so we just show the tag.
```
LalitMaganti, line 79: Does this apply anymore? Frankly seems a bit excessive concern.

LalitMaganti (approved):

ui/src/plugins/dev.perfetto.AndroidLog/logs_track.ts
stevegolton: This PR doesn't affect this comment in particular. The concern here is that we're loading the message for each event that's loaded into the track into memory, as tooltips need to be synchronously available. Maybe it's not a huge deal though...

Having said that, rendering DOM tooltips does make it easier to format the potentially long message strings with proper wrapping or cropping, which was an issue with the canvas rendering approach.

I'll try it out and upload in a followup if it works out.

PR #1233, opened 2025-04-10
Remove non UTF-8 characters in systrace parser

elrakadm:
Remove non UTF-8 characters in systrace parser

A trace can contain some "bytes" in "print" ftrace events which are not valid UTF8. sqlite expects strings to be UTF-8.

Tested via 
```
tools/ninja -C out/linux_clang_debug && out/linux_clang_debug/perfetto_unittests &&  tools/diff_test_trace_processor.py  out/linux_clang_debug/trace_processor_shell
```

### DIFF
--- a/src/trace_processor/importers/systrace/systrace_parser.cc
+++ b/src/trace_processor/importers/systrace/systrace_parser.cc
@@ -51,7 +51,8 @@ void SystraceParser::ParsePrintEvent(int64_t ts,
                                      uint32_t pid,
                                      base::StringView event) {
   systrace_utils::SystraceTracePoint point{};
-  switch (ParseSystraceTracePoint(event, &point)) {
+  std::string tmp = base::RemoveInvalidUTF8(event.ToStdString());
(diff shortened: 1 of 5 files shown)


src/trace_processor/importers/systrace/systrace_parser.cc
```diff
@@ -51,7 +51,8 @@ void SystraceParser::ParsePrintEvent(int64_t ts,
                                      uint32_t pid,
                                      base::StringView event) {
   systrace_utils::SystraceTracePoint point{};
-  switch (ParseSystraceTracePoint(event, &point)) {
+  std::string tmp = base::RemoveInvalidUTF8(event.ToStdString());
```
LalitMaganti, line 54: Hmmmmm I'm not sure about doing it here honestly. I'd suggest instead doing it below.

Also creating a std::string on every parse point is non-ideal. I'd suggest instead having an std::string stored in SystraceParser as a field and then pass the reference/pointer to that in RemoveInvalidUTF8

PR #1241, opened 2025-04-10
Dev/wusamuel/wattson UI toggle unit

wusamuel6:
ui: Add toggle for Wattson uW/mW
    
Create a custom Wattson AggregationPanel that has a button toggle for
switching between displaying mW and uW units.
    
Bug: 409794084


ui/src/plugins/org.kernel.Wattson/aggregation_panel.ts
```diff
@@ -0,0 +1,139 @@
...
+  implements m.ClassComponent<AggregationPanelAttrs>
+{
+  private scaleNumericData: boolean = false;
+
+  view({attrs}: m.CVnode<AggregationPanelAttrs>) {
+    return m(
+      '.details-panel',
+      m(
+        '.details-panel-heading.aggregation',
+        m('.button.pf-button--filled',
```
stevegolton: Please use the Button component - helps keep the look and feel of the UI consistent.

Might I suggest the SegmentedButton component might work better here.

ui/src/components/aggregation_adapter.ts
```diff
@@ -67,3 +76,19 @@ export function createAggregationToTabAdaptor(
...
 }
+
+export function createAggregationToTabAdaptor(
+  trace: Trace,
+  aggregator: AreaSelectionAggregator
+): AreaSelectionTab {
+  return createBaseAggregationToTabAdaptor(trace, aggregator, AggregationPanel);
+}
+
+export function createWattsonAggregationToTabAdaptor(
```
stevegolton: Can you move `createWattsonAggregationToTabAdaptor` into the wattson plugin? It feels very wattson specific + I don't want to have multiple tab adapter types floating around.

stevegolton (changes requested):

ui/src/plugins/org.kernel.Wattson/aggregation_panel.ts
wusamuel6: SegmentedButton is much better, thanks for suggestion! Done.

ui/src/components/aggregation_adapter.ts
wusamuel6: Done.

PR #1244, opened 2025-04-10
[ui] Enable power aggregation tab plugin by default

cphlipot1:
The power aggregations tab provides users a better UX when
aggregating over power rail. Enable this plugin by default so that
users get this better UX without having to enable the plugin
manually.


stevegolton (approved):
LGTM

primiano (approved):
This is so cool. I just tried it out... spot on!
Thanks so much for the contrib!

LalitMaganti (approved):

PR #1248, opened 2025-04-11
tp: add support for limit/offset

LalitMaganti:


src/trace_processor/dataframe/impl/bytecode_instructions.h
```diff
@@ -359,6 +359,18 @@ struct Distinct : Bytecode {
                                      indices_register);
 };
 
+// Limits the number of indices in a span and applies an offset.
```
aMayzner: Maybe change the order? First you do the offset, then limit

aMayzner (approved):

Issue #1181, opened 2025-04-07
Support binding on a custom IPs with trace_processor_shell --httpd


sydnash:
as the trace_processor_shell --help only shows that it can set the listen port through --http-port, but i cannot found a way to set the listen ips.

sydnash:
I've write some code to add the capabilities to listen on the specific ip by --http-ip, could some can give some advise to it? @ddiproietto

LalitMaganti:
This has landed, thanks for the contribution!

PR #1259, opened 2025-04-14
tp: support listening on specified ip

sydnash:
make trace_processor_shell support listen on specific ip other than localhost.
issuse: https://github.com/google/perfetto/issues/1181

users can use --http-ip 0.0.0.0 to specific an ip or a hostname.  
trace_processor_shell will listen on 0.0.0.0 by default.
this make it can be accessed by localhost:9001 even trace_processor_shell is started at a docker container by port forward.

like this: docker run -it --rm   -p 9001:9001 image:tag trace_processor_shell -D --http-port 9001

### DIFF
--- a/include/perfetto/ext/base/unix_socket.h
+++ b/include/perfetto/ext/base/unix_socket.h
@@ -74,7 +74,7 @@ enum class SockPeerCredMode {
 #endif
 };
 
-// Returns the socket family from the full addres that perfetto uses.
+// Returns the socket family from the full address that perfetto uses.
 // Addr can be:
 // - /path/to/socket : for linked AF_UNIX sockets.
 // - @abstract_name  : for abstract AF_UNIX sockets.
@@ -83,13 +83,22 @@ enum class SockPeerCredMode {
 // - vsock://-1:3000 : for VM sockets.
 SockFamily GetSockFamily(const char* addr);
 
+// NetAddrInfo is a wrapper for a full address and it's family
+// and type.
 struct NetAddrInfo {
   NetAddrInfo(const std::string& ip_port, SockFamily family, SockType type)
       : ip_port_(ip_port), family_(family), type_(type) {}
-  std::string ip_port_;
-  SockFamily family_;
-  SockType type_;
+  std::string ip_port_;  // full address with ip and port
+  SockFamily family_;    // socket family
+  SockType type_;        // socket type
 };
+
+// Returns a list of NetAddrInfo from ip and port which
+// ip can be an ipv4 or an ipv6 or a domain.
+// 127.0.0.1    : ipv4 address
+// localhost    : domain
+// ::1          : ipv6 address
+// port is a normal tcp port as string.
 std::vector<NetAddrInfo> GetNetAddrInfo(const std::string& ip,
                                         const std::string& port);
 
--- a/include/perfetto/ext/base/unix_socket.h
+++ b/include/perfetto/ext/base/unix_socket.h
@@ -83,6 +83,16 @@ enum class SockPeerCredMode {
 // - vsock://-1:3000 : for VM sockets.
 SockFamily GetSockFamily(const char* addr);
 
+struct NetAddrInfo {
+  NetAddrInfo(const std::string& ip_port, SockFamily family, SockType type)
+      : ip_port_(ip_port), family_(family), type_(type) {}
+  std::string ip_port_;
+  SockFamily family_;
+  SockType type_;
+};
+std::vector<NetAddrInfo> GetNetAddrInfo(const std::string& ip,
+                                        const std::string& port);
+
 // Returns whether inter-process shared memory is supported for the socket.
 inline bool SockShmemSupported(SockFamily sock_family) {
 #if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
--- a/src/base/http/http_server.cc
+++ b/src/base/http/http_server.cc
@@ -73,6 +73,38 @@ void HttpServer::Start(int port) {
   }
 }
 
+void HttpServer::Start(const std::string& listen_ip, int port) {
+  std::string ip = listen_ip;
+  if (listen_ip.empty()) {
+    ip = "0.0.0.0";
+  }
+  std::string port_str = std::to_string(port);
+  std::vector<NetAddrInfo> addr_infos = GetNetAddrInfo(ip, port_str);
+  for (NetAddrInfo& info : addr_infos) {
+    if (info.family_ == SockFamily::kInet) {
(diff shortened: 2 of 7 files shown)


src/base/http/http_server.cc
```diff
@@ -73,6 +73,38 @@ void HttpServer::Start(int port) {
...
 
+void HttpServer::Start(const std::string& listen_ip, int port) {
+  std::string ip = listen_ip;
+  if (listen_ip.empty()) {
+    ip = "0.0.0.0";
+  }
+  std::string port_str = std::to_string(port);
+  std::vector<NetAddrInfo> addr_infos = GetNetAddrInfo(ip, port_str);
+  for (NetAddrInfo& info : addr_infos) {
+    if (info.family_ == SockFamily::kInet) {
```
LalitMaganti, line 88: This is just duplicating a lot of code already in main. Can you please refactor the existing code instead of duplicating?

src/base/unix_socket.cc
```diff
@@ -299,6 +320,31 @@ SockFamily GetSockFamily(const char* addr) {
...
 }
 
+std::vector<NetAddrInfo> GetNetAddrInfo(const std::string& ip,
+                                        const std::string& port) {
+  struct addrinfo hints, *serv_info, *p;
+  memset(&hints, 0, sizeof hints);
+  hints.ai_family = AF_UNSPEC;
+  hints.ai_socktype = SOCK_STREAM;
+  hints.ai_flags = AI_PASSIVE;
+  PERFETTO_CHECK(getaddrinfo(ip.c_str(), port.c_str(), &hints, &serv_info) ==
```
LalitMaganti, line 336: This is not going to work on Windows.

LalitMaganti: I also don't understand why we need this. Why are we not just directly binding to the IP address and letting the OS deal with all of this?

src/trace_processor/trace_processor_shell.cc
```diff
@@ -750,6 +751,7 @@ General purpose:
 Behavioural:
  -D, --httpd                          Enables the HTTP RPC server.
  --http-port PORT                     Specify what port to run HTTP RPC server.
+ --http-ip ip                         Specify what ip to run HTTP RPC server.
```
LalitMaganti: ip address

LalitMaganti (changes requested):

src/base/http/http_server.cc
sydnash: > This is just duplicating a lot of code already in main. Can you please refactor the existing code instead of duplicating?

do you mean this duplicated with HttpServer::Start(int port) function?

src/base/unix_socket.cc
sydnash: there are tow reasons:
1. use  the getaddrinfo to check the ip address is ipv4 or ipv6.
2. through this, user can specific a domoain such as localhost or hostname, it will auto get the ipv4 and ipv6 address.

sydnash: > This is not going to work on Windows.

what's the error report?  the MakeSockAddr in unix_socket.cc use the getaddrinfo too.

sydnash: > > This is not going to work on Windows.
> 
> what's the error report? the MakeSockAddr in unix_socket.cc use the getaddrinfo too.

I think it's because i call getaddrinfo before WSAStartup, can you help to check if this ok now?

src/base/http/http_server.cc
```diff
@@ -56,23 +56,48 @@ void HttpServer::Start(int port) {
...
   bool ipv6_listening = sock6_ && sock6_->is_listening();
   if (!ipv6_listening) {
-    PERFETTO_PLOG("Failed to listen on IPv6 socket: \"%s\"", ipv6_addr.c_str());
+    PERFETTO_PLOG("Failed to listen on IPv6 socket: \"%s\"", ip_addr.c_str());
     sock6_.reset();
   }
 }
 
+void HttpServer::Start(const std::string& listen_ip, int port) {
+  std::string ip = listen_ip;
```
LalitMaganti: I think if listen_ip is empty, we should default to localhost *not* 0.0.0.0.

Also this can be simplified to:
```
std::string ip = listen_ip.empty() ? "localhost" : listen_ip;
```

src/trace_processor/rpc/httpd.h
```diff
@@ -29,7 +29,9 @@ class TraceProcessor;
 // The unique_ptr argument is optional. If non-null, the HTTP server will adopt
 // an existing instance with a pre-loaded trace. If null, it will create a new
 // instance when pushing data into the /parse endpoint.
-void RunHttpRPCServer(std::unique_ptr<TraceProcessor>, const std::string&);
+void RunHttpRPCServer(std::unique_ptr<TraceProcessor>,
```
LalitMaganti: These parameters should be named and need some comments. Two unnamed parameters are very mysterious.

src/trace_processor/trace_processor_shell.cc
```diff
@@ -750,6 +751,7 @@ General purpose:
 Behavioural:
  -D, --httpd                          Enables the HTTP RPC server.
  --http-port PORT                     Specify what port to run HTTP RPC server.
+ --http-ip ip                         Specify what ip address to run HTTP RPC server.
```
LalitMaganti: http-ip-address

include/perfetto/ext/base/unix_socket.h
```diff
@@ -83,6 +83,16 @@ enum class SockPeerCredMode {
...
 SockFamily GetSockFamily(const char* addr);
 
+struct NetAddrInfo {
+  NetAddrInfo(const std::string& ip_port, SockFamily family, SockType type)
+      : ip_port_(ip_port), family_(family), type_(type) {}
+  std::string ip_port_;
+  SockFamily family_;
+  SockType type_;
+};
+std::vector<NetAddrInfo> GetNetAddrInfo(const std::string& ip,
```
LalitMaganti, line 102: Similar to other functions, this needs some comments.

src/base/http/http_server.cc
LalitMaganti: Yes. Much better now.

src/base/unix_socket.cc
LalitMaganti: Ack looks OK now.

LalitMaganti (changes requested):

src/base/http/http_server.cc
sydnash: good, this can keep the previous behavier.

LalitMaganti (approved):

PR #1266, opened 2025-04-14
perfetto: fix system_info proto name

LalitMaganti:
Now system_info proto is in common, there are build system clashes. Fix
it by just renaming this proto.


rsavitski (approved):
Consider adding the following to the merged patch description:
Fixes: 53afaf1

Issue #275, opened 2022-04-13
Chromium v8 JavaScript (JS) stacks in Perfetto?


ivberg:
It seems that the tracing probes and infrastructure are highly shared between F12 perf dev tools & Perfetto.

It seems F12 Dev Tools [enables some categories](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/devtools/front_end/timeline/TimelineController.js;drc=977fa8c33a878dd5e9f33ef5aa036aff86932e24;l=46) that Perfetto UI Chrome tracing supports. Things like these categories

- v8.runtime_stats_sampling
- v8.cpu_profiler

Again these can be collected by Perfetto. Additionally, a F12 profile with JS stacks can be saved and loaded by Perfetto.

However, it doesn't seem like Perfetto has support for these. There is no UI to show them and the SQL query interface does now show /expose data under [cpu_profile_stack_sample](https://perfetto.dev/docs/analysis/sql-tables#cpu_profile_stack_sample) no matter if the trace was taken in Perfetto or F12 Profile JSON

primiano:
> However, it doesn't seem like Perfetto has support for these. There is no UI to show them and the SQL query interface does now show /expose data under [cpu_profile_stack_sample](https://perfetto.dev/docs/analysis/sql-tables#cpu_profile_stack_sample) no matter if the trace was taken in Perfetto or F12 Profile JSON

That's correct, perfetto UI and trace processor never had any support for JS callstack samples. The data is really just plumbed through (only some paths even, IIRC, only the JSON exported) and then the chrome devtools backend ingests it.

Historical reasons, if you ask. The v8 profilers were written by devtools and tracing was not fully there IIRC. Then tracing came along but nobody ever merged the two paths. There is no big plan to do so, because v8 profilers are typically aimed at web developers (who use devtools) and trace events are aimed at browser engineers.

Don't get me wrong, it would be nice if they were consistently supported in all tools. But this is by no means a priority for anybody, and as such will probably not bubble up in the next N years.

ivberg:
Thx @primiano . It's always refreshing & nice to hear the context / history here and the reality with respect to priorities and feature support. I appreciate it! 

Since the title of this issue is "Chromium v8 JavaScript (JS) stacks in Perfetto?" I think that's 1 interesting future scenario. However, mixed with this is what you also responded to.
>  The data is really just plumbed through (only some paths even, IIRC, only the JSON exported) and then the chrome devtools backend ingests it.

[...]

PR #1267, opened 2025-04-14
tp: fix parsing of Chrome Devtools/NodeJS CPU profiling data

LalitMaganti:
Turns out they use a separate format to V8 with a more standardised
system [1]. Since supporting it is quite simple, just do it.

See discussion at https://github.com/google/perfetto/issues/275 for what prompted this.

[1] https://chromedevtools.github.io/devtools-protocol/tot/Profiler/#type-ProfileNode

### DIFF
--- a/src/trace_processor/importers/common/legacy_v8_cpu_profile_tracker.cc
+++ b/src/trace_processor/importers/common/legacy_v8_cpu_profile_tracker.cc
@@ -19,7 +19,9 @@
 #include <cstdint>
 #include <optional>
 #include <utility>
+#include <vector>
 
+#include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
 #include "perfetto/ext/base/flat_hash_map.h"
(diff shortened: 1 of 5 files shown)


stevegolton (approved):

aMayzner (approved):

hardfist:
@LalitMaganti sorry to bother, but it seems this only fixes the json_trace_tokenizer and not fixes the track_event_tokenizer, when I try to migrate from json format trace to proto format trace, it seems the track_event tokenizer not support NodeJS style CPU profiling yet.
the following is pftrace contains NodeJS CPU profiling but seems not working
[trace.pftrace.zip](https://github.com/user-attachments/files/20626308/trace.pftrace.zip)

LalitMaganti:
Unfortunately I don't have any bandwidth to work on this as I'm in the midst of some major projects. Patches welcome if you want to contribute, with the existing code it should be pretty straightforward to see how to do it.

PR #1270, opened 2025-04-14
Android Java SDK: allow to extend 'PerfettoTrace$Category'.

KirillTim:
In framework, we need to do additional checks in 'Category#isEnabled()'.

Tested:
```
```


zezeozue (approved):

src/android_sdk/java/test/dev/perfetto/sdk/test/PerfettoTraceTest.java
```diff
@@ -628,8 +628,45 @@ public void testRegister() throws Exception {
...
+        super(name);
+      }
+
+      @Override
+      public boolean isEnabled() {
+        return false;
+      }
+    }
+
+    Category disabledFooCategory = new DisabledCategory(FOO);
```
ddiproietto: Nit: different name?

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -50,7 +50,7 @@ public final class PerfettoTrace {
    * Perfetto category a trace event belongs to. Registering a category is not sufficient to capture
    * events within the category, it must also be enabled in the trace config.
    */
-  public static final class Category implements PerfettoTrackEventExtra.PerfettoPointer {
```
ddiproietto, line 53: Discussed offline:

it would be nice to have some benchmarks for this change. If this turns out to make things slower, we could think about changing the behavior by not registering the category (which is a slowpath). Not registering the category will make it always disabled.

ddiproietto (approved):

PR #1273, opened 2025-04-15
tp: support multiple --add-sql-package in cmdline

cervantesyu:
This revision fixes the bug that trace_processor_shell allows only one `--add-sql-module` path. Also rename arguments to --add-sql-package and --override-sql-package since the args are paths for loading SQL modules and act as SQL packages.

Bug: 410753691
Test: diff_test
Change-Id: Ia45032ded2b9a06a5e5a0e110a874f067073f45a


src/trace_processor/trace_processor_shell.cc
```diff
@@ -774,11 +774,13 @@ General purpose:
...
                                       executed after the selected metrics and
                                       the metrics output is suppressed.
- --add-sql-module PACKAGE_PATH         Files from the directory will be treated
+ --add-sql-module PACKAGE_PATH
+ --add-sql-package PACKAGE_PATH       Files from the directory will be treated
                                       as a new SQL package and can be used for
                                       INCLUDE PERFETTO MODULE statements. The
                                       name of the directory is the module name.
- --override-sql-module PACKAGE_PATH   Will override trace processor package with
+ --override-sql-module PACKAGE_PATH
```
LalitMaganti: Can you please move these to the Advanced section and say they are the same as --add-sql-package/--override-sql-package but with more acccurate names and are only kept around for backwards compatbility?

src/trace_processor/trace_processor_shell.cc
```diff
@@ -1718,10 +1723,13 @@ base::Status MaybeUpdateSqlModules(const CommandLineOptions& options) {
     }
   }
 
-  if (!options.sql_module_path.empty()) {
-    auto status = IncludeSqlModule(options.sql_module_path, false);
-    if (!status.ok())
-      return base::ErrStatus("Couldn't add SQL module: %s", status.c_message());
+  if (!options.sql_module_paths.empty()) {
+    for (const auto& add_sql_module_path : options.sql_module_paths) {
+      auto status = IncludeSqlModule(add_sql_module_path, false);
```
LalitMaganti: While you're here please refactor `IncludeSqlModule` (specifically Module -> Package).

LalitMaganti (changes requested):

src/trace_processor/trace_processor_shell.cc
cervantesyu: Done.

src/trace_processor/trace_processor_shell.cc
cervantesyu: Done.

src/trace_processor/trace_processor_shell.cc
```diff
@@ -705,7 +705,7 @@ struct CommandLineOptions {
 
   std::string query_file_path;
   std::string query_string;
-  std::string sql_module_path;
+  std::vector<std::string> sql_module_paths;
```
LalitMaganti: This and below need to be renamed as well.

cervantesyu: Now that we are here, I made more renames of module -> package in the file to reflect that
package is the root dir to host subdirs or SQL source files.
module is an individual SQL source file.

PTAL. Thanks.

PR #1275, opened 2025-04-15
protovm: initial implementation

rukkal:


src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
+/*
```
primiano: Can you rename this file (and class) to SlabAllocator

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+      slabs_.push_back(std::move(slab));
+      next_free_slot_ = &slabs_.back()[0];
+    }
+
+    auto* slot = next_free_slot_;
+    next_free_slot_ = slot->next;
+    return &slot->element;
+  }
+
+  void Deallocate(void* p) {
```
primiano: s/Deallocate/Free/? it seems a quite standard term in allocator-land

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+    slot->next = next_free_slot_;
+    next_free_slot_ = slot;
+  }
+
+  static constexpr size_t kSlabCapacity = 64;
+
+ private:
+  union Slot {
+    Slot* next;
+    std::aligned_storage_t<ElementSize, ElementAlign> element;
```
primiano: I learned recently stD::aligned_storage is deprecated in C++23.
See aosp/3439585 for alterative (use alignas)

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+      if (!slab) {
+        return nullptr;
+      }
+      slabs_.push_back(std::move(slab));
+      next_free_slot_ = &slabs_.back()[0];
+    }
+
+    auto* slot = next_free_slot_;
+    next_free_slot_ = slot->next;
+    return &slot->element;
```
primiano: memset(0)  it before returning? can avoid subtle bugs, and I don't think we are so perf sensitive that the zeroing will make any practical difference to perf

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  void Deallocate(void* p) {
+    auto* slot = static_cast<Slot*>(p);
+    slot->next = next_free_slot_;
+    next_free_slot_ = slot;
+  }
+
+  static constexpr size_t kSlabCapacity = 64;
```
primiano: make this a template arugment as well (and just assign it by default?)

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+//   hold multiple elements.
+//
+// - Free list management: a free list keeps track of available elements within
+//   the allocated slabs. When a request for allocation comes in, the allocator
+//   simply takes an element from the free list. Deallocation returns the
+//   element back to the free list.
+template <size_t ElementSize, size_t ElementAlign>
+class FreeListAllocator {
+ public:
+  explicit FreeListAllocator(size_t& remaining_memory_bytes)
```
primiano: a bit clearer if the ctor argument is called "memory_limit_bytes" (but still keep the member variable called remaining_memory_byes) .
Also add a DCHECK that max_bytes % ElementSize == 0

primiano: why the & here? seems very weird to pass a reference

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+// - Slab allocation: instead of requesting memory for each individual element,
+//   it allocates large chunks of memory (slabs) upfront, where each slab can
+//   hold multiple elements.
+//
+// - Free list management: a free list keeps track of available elements within
+//   the allocated slabs. When a request for allocation comes in, the allocator
+//   simply takes an element from the free list. Deallocation returns the
+//   element back to the free list.
+template <size_t ElementSize, size_t ElementAlign>
+class FreeListAllocator {
```
primiano: This class is great, I only have one quesiton: why does it do only "raw" memory management and delegates the placement new/dtor to the outside?
In other words, why this is not instead a FreeListAllocator<T>, where T is the actual class (so you use sizeof(T), alignemtnof(T) here, and do placement new and placement dtor on Allocate/Free)

The only good reason to do this is if you want to use the same SlabAllocator for 2 different classes hat happen to use have the same size (but are you actually going to do that?)
And even if that was the case, I'd still move the placement new/dtor here, and make the Allocate/Free methods templated (i.e. Allocate<T>())

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+  static constexpr size_t kSlabCapacity = 64;
+
+ private:
+  union Slot {
+    Slot* next;
+    std::aligned_storage_t<ElementSize, ElementAlign> element;
+  };
+
+  std::unique_ptr<Slot[]> CreateSlab() {
+    if (sizeof(Slot[kSlabCapacity]) > remaining_memory_bytes_) {
```
primiano: Add a small comment:
// The new slab might go beyond our left budget.

Or maybe this is a bit more readable if you make this:

static constexpr size_t kSlabSize = sizeof(Slot[kSlabCapacity])
if (kSlabSize > remaining_memory_byes_) ...

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+      return nullptr;
+    }
+
+    auto slab = std::unique_ptr<Slot[]>(new Slot[kSlabCapacity]);
+    remaining_memory_bytes_ -= sizeof(Slot[kSlabCapacity]);
+
+    for (size_t i = 0; i + 1 < kSlabCapacity; ++i) {
+      auto& slot = slab[i];
+      auto& next_slot = slab[i + 1];
```
primiano: technically speaking this is a UB on the last entry, because you are taking the address of someting beyond bounds. clang became stricter over time.

Just stop the iteration in the for loop at `i < kSlabCapacity - 2` ?

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    auto& last_slot = slab[kSlabCapacity - 1];
+    last_slot.next = nullptr;
+
+    return slab;
+  }
+
+  size_t& remaining_memory_bytes_;
+  Slot* next_free_slot_;
```
primiano: = nullptr

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+      slot.next = &next_slot;
+    }
+
+    auto& last_slot = slab[kSlabCapacity - 1];
+    last_slot.next = nullptr;
+
+    return slab;
+  }
+
+  size_t& remaining_memory_bytes_;
```
primiano: why this reference? 
1. If you need it, this is very obscure. Make it a pointer. The problem of using references instead of pointers like this is that you don't realize they are references in the places where they get used, and that can lead to bugs when toucing the code afterwards. (a reference member is a pointer that pretends to look like an innocent  variable) 
2. in general keeping references as member variable is very discouraged (I thought some compilers even complain about it)

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+
+    auto* slot = next_free_slot_;
+    next_free_slot_ = slot->next;
+    return &slot->element;
+  }
+
+  void Deallocate(void* p) {
+    auto* slot = static_cast<Slot*>(p);
+    slot->next = next_free_slot_;
+    next_free_slot_ = slot;
```
primiano: Well something must free up the slab, if this was the last active element of the slab, right?
Otherwise this allocator will just leak entire slabs if you keep allocating and deallocating

src/protovm/free_list_allocator.h
```diff
@@ -0,0 +1,104 @@
...
+
+    auto& last_slot = slab[kSlabCapacity - 1];
+    last_slot.next = nullptr;
+
+    return slab;
+  }
+
+  size_t& remaining_memory_bytes_;
+  Slot* next_free_slot_;
+  std::vector<std::unique_ptr<Slot[]>> slabs_;
```
primiano: I'd probably arrange this as

struct Slab {
   Slot slots[kCapacity]
   // any other per-slab count/metadata you might need
}

and then
std::list<Slab> slabs

2 reasons:
1. std::list guarantees you pointer stability, so there is no need of an extra unique_ptr
2. std::list emits way less code than vector (check on godbolt, vector spews soooo much assembly). And you don't really need random indexing here

src/protovm/allocator.h
```diff
@@ -0,0 +1,66 @@
...
+namespace perfetto {
+namespace protovm {
+
+// A centralized allocator to manage all the memory requests from RwProto. It
+// enforces a strict memory usage limit, thus allowing fine-grained control of
+// the overall proto VM's memory footprint.
+class Allocator {
+ public:
+  explicit Allocator(size_t memory_limit_bytes);
+  StatusOr<Node::Bytes> CreateBytes(protozero::ConstBytes data);
```
primiano: Can you add some commentS? It's unclear by looking at this what's the differnece between CreateBytes and AllocateBytes.
Very subtle

src/protovm/allocator.h
```diff
@@ -0,0 +1,66 @@
...
+  void Delete(Node* node);
+  void DeleteReferencedData(Node* node);
```
primiano, line 47: Ditto here, unclear what's Delete vs DeleteRefData

src/protovm/allocator.h
```diff
@@ -0,0 +1,66 @@
...
+  UniquePtr<void> AllocateBytes(size_t size);
+  void DeallocateBytes(UniquePtr<void> p, size_t size);
+
+  template <class T, class... Args>
+  StatusOr<UniquePtr<Node>> CreateNode(Args&&... args) {
+    auto* p = free_list_.Allocate();
+    if (!p) {
+      PERFETTO_ABORT(
```
primiano: I don't know what PEFETTO_ABORT is (I guess another commit introduces it). but imho this should be called PERFETTO_VM_ABORT as I guess it's specific to the VM thing

Maybe can it be an inline function so can just be Abort in the protovm:: namespace?

src/protovm/allocator.h
```diff
@@ -0,0 +1,66 @@
...
+          "Failed to allocate node (%zu bytes). Memory limit: %zu [bytes]. "
+          "Remaining: %zu "
+          "[bytes].)",
+          sizeof(Node), memory_limit_bytes_, remaining_memory_bytes_);
+    }
+    new (p) Node{T{std::forward<Args>(args)...}};
```
primiano, line 77: Can you add some comment here, this seems a bit subtle.
You initialize the object as T (fine) but you then cast it to Node?

Is it the case that you expect all Ts to be subclasses of Node? 
If that's the case I think it becomes more readable and explicit if you add

static_assert(std::is_base_of<Node, T>::value);

src/protovm/allocator.h
```diff
@@ -0,0 +1,66 @@
...
+    return UniquePtr<Node>(static_cast<Node*>(p));
+  }
+
+ private:
+  size_t memory_limit_bytes_;
```
primiano: = 0 to both

src/protovm/allocator.cc
```diff
@@ -0,0 +1,90 @@
...
+
+namespace perfetto {
+namespace protovm {
+
+Allocator::Allocator(size_t memory_limit_bytes)
+    : memory_limit_bytes_{memory_limit_bytes},
+      remaining_memory_bytes_{memory_limit_bytes},
+      free_list_{remaining_memory_bytes_} {}
+
+StatusOr<Node::Bytes> Allocator::CreateBytes(protozero::ConstBytes data) {
```
primiano: Aaah okay I see, this is a dup() equivalent.
It allocates a node of that size and fills it.
Maybe call it AllocateAndCopyBytes ?

src/protovm/allocator.cc
```diff
@@ -0,0 +1,90 @@
...
+  if (data.size == 0) {
+    return Node::Bytes{nullptr, 0};
+  }
+
+  auto copy = AllocateBytes(data.size);
+  if (!copy) {
```
primiano, line 39: I think AllocateBytes itself is going to return an Abort statusOr right? so her eyou can just "return copy" I thikn (or use std::move or whatever retains the abort message, without creating a new one)

src/protovm/allocator.cc
```diff
@@ -0,0 +1,90 @@
...
+    }
+  } else if (auto* map_node = node->GetIf<Node::MapNode>()) {
+    Delete(map_node->value.release());
+  } else if (auto* bytes = node->GetIf<Node::Bytes>()) {
+    DeallocateBytes(std::move(bytes->data), bytes->size);
+  }
+}
+
+UniquePtr<void> Allocator::AllocateBytes(size_t size) {
+  if (size > remaining_memory_bytes_) {
```
primiano: it's a bit odd that both FreeListAllocator and this do the limit accounting. imho should be just one of them.
EDIT: aaaaah now I understand why you used a refernce in FLA.
Okay this is too subtle. Either:

1) Make it a pointer in FLA
2) Add an explicit feature in FLA, something like add/remove_extra_bytes() and explain that is used to acocunt for bytes allocated directly outside of the FLA (so all the math stays in FLA)

src/protovm/allocator.cc
```diff
@@ -0,0 +1,90 @@
...
+    PERFETTO_ELOG("Reached memory limit");
```
primiano: this log is too generic, and realistically we won't ever see it.
just remove it?
Why the other allocation failure is a StatusOr and this is an ELOG, where does that StatusOr message go?

primiano (commented):
Comments inline + an overall comment.
Maybe it's more future proof if we call that UniquePtr Handle (or AllocatorHandle) so in future can be compacted (and pointers moved) without having to impact the caller.

I tried to sketch a possible implementation just to give an idea.
Again, we don't need the actual compacting allocator here, I'd just introduce a (very dumb) Hadle class to be future-proof.

You are already in the right state as  in your design you need to std::move back the handle into Deallocate() to free the object, which is great.


```

// An allocator that reserves the right to move pointers around
class CompactingAllocator {

  std::vector<T> objects;

  Handle<T> Allocate() {
    refs.emplace_back(...);
    ObjRef ref = refs.back();
    return Handle<T>(ref)
  }

  std::list<ObjRef<T>> refs;
}

struct ObjRef {
  T* ptr;
}

// A handle is just a pointer to an Object holder.
// The handle itslef is guarnateed to be a stable pointer
// (the std::list guarantees that) but the pointed memory
// can move.
// Make it move-only, delete copy ctors
struct Handle<T> {
  T* operator->() {
    return static_cast<T>(obj->ptr);
  }

  private:
  Obj<T>* obj;
}
```

src/protovm/free_list_allocator.h
rukkal: Done

src/protovm/free_list_allocator.h
rukkal: Done

src/protovm/free_list_allocator.h
rukkal: Done

src/protovm/free_list_allocator.h
rukkal: Done

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+// IndexedRepeatedField: Represents a protobuf repeated field where elements
+//                       are accessed by index.
+//
+//
+// MappedRepeatedField: Represents a protobuf repeated field where elements
+//                      are accessed by key.
+
+class RwProto {
+ public:
+  class Cursor {
```
primiano: Can this be moved to a "sibling" class (RwProtoCursor + using Cursor = RwProtoCursor)?
Or do you access private fields of Cursor from RwProto (which would justify the nested class)

Just I find nested classes harder to read (and also impossible to fwd declare if needed).
IIRC the general style guideline is to avoid nested classes if possible, and use "using" aliases to put them within the namespace of the class

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+                                               UniquePtr<Node> map_value);
+    StatusOr<uint64_t> ReadScalarField(Node& node, uint32_t field_id);
+
+    Node* node_;
+    std::pair<IntrusiveMap*, Node::MapNode*> holding_map_and_node_ = {nullptr,
+                                                                      nullptr};
+    Allocator* allocator_;
+  };
+
+  explicit RwProto(Allocator& allocator);
```
primiano: Take a pointer, not a reference.
Same reasoning of the above CL, references don't make it obvious you are mutating the object at the call site, they look like a "pass by value"

IMHO references should only show up as "const X&" (or in locally scoped things like auto& foo = ..." but not in function signatures

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+  ~RwProto();
+  Cursor Root();
```
primiano: nit: root() or GetRoot()

src/protovm/free_list_allocator.h
rukkal: Sg. Done

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+  std::string SerializeAsString() const;
+  void SerializeField(uint32_t field_id,
```
primiano, line 147: 1) What does this do? (add a comment, it's not obviouS)
2) If  Node is an output argument, take it by pointer, not reference

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+//                      are accessed by key.
+
+class RwProto {
+ public:
+  class Cursor {
+   public:
+    class RepeatedFieldIterator {
+     public:
+      RepeatedFieldIterator();
+      explicit RepeatedFieldIterator(Allocator& allocator,
```
primiano: Allocator* (see other comment about refs vs pointers)

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+                                     IntrusiveMap::Iterator it);
+      RepeatedFieldIterator& operator++();
+      Cursor operator*();
```
primiano: should this return a Cursor& or you actually want a copy? if you wnat a copy make an explicit method.
operator* is the one case when you actually want a reference (unless, again,  you want to make a new Cursor obj every time)

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+     private:
+      Allocator* allocator_;
+      IntrusiveMap::Iterator it_;
+    };
+
+    Cursor();
+    explicit Cursor(Node* node, Allocator* allocator);
+    StatusOr<bool> HasField(uint32_t field_id);
+    StatusOr<void> EnterField(uint32_t field_id);
+    StatusOr<void> EnterRepeatedFieldByIndex(uint32_t field_id, uint32_t index);
```
primiano: Maybe just call it EnterRepeatedFieldAt (or EnterRepeatedFieldAtOffset)

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+    StatusOr<RepeatedFieldIterator> IterateRepeatedField(uint32_t field_id);
+    StatusOr<void> EnterRepeatedFieldByKey(uint32_t field_id,
```
primiano: this deserves a comment (I know what it does, but it's the most complicate to guess for other people)

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+                                           uint32_t map_key_field_id,
+                                           uint64_t key);
+    StatusOr<Scalar> GetScalar() const;
+    StatusOr<void> SetBytes(protozero::ConstBytes data);
+    StatusOr<void> SetScalar(Scalar scalar);
+    StatusOr<void> Merge(protozero::ConstBytes data);
```
primiano: Add a comment to explain what this does. I don't recall on top of my head (or without looking at the impl).
Does it "append" bytes? or does it a recursive merge? Or does it only do a 1 level merge (i.e. creates new fields if not exist, overwrites existing fields. but does not recurse in nested submessages)

I think the latter, but please add a  comment on it as this is subtle

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+    StatusOr<IntrusiveMap::Iterator> FindOrCreateMessageField(
+        Node& node,
+        uint32_t field_id);
+    StatusOr<IntrusiveMap::Iterator> FindOrCreateIndexedRepeatedField(
+        Node& node,
+        uint32_t index);
+    StatusOr<IntrusiveMap::Iterator> FindOrCreateMappedRepeatedField(
+        Node& node,
+        uint64_t key);
+    StatusOr<IntrusiveMap::Iterator> MapInsert(IntrusiveMap& map,
```
primiano: nit: IntrusiveMap*

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+                                               uint64_t key,
+                                               UniquePtr<Node> map_value);
+    StatusOr<uint64_t> ReadScalarField(Node& node, uint32_t field_id);
+
+    Node* node_;
```
primiano: = nullptr

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+    std::pair<IntrusiveMap*, Node::MapNode*> holding_map_and_node_ = {nullptr,
+                                                                      nullptr};
```
primiano: Just = {} should do the same and avoids a line wrapping

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,216 @@
...
+    Allocator* allocator_;
```
primiano: = nullptr

src/protovm/allocator.cc
rukkal: Definitely! We both got confused by that name earlier this morning

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+    : allocator_{&allocator}, it_{it} {}
+
+RwProto::Cursor::RepeatedFieldIterator&
+RwProto::Cursor::RepeatedFieldIterator::RepeatedFieldIterator::operator++() {
+  ++it_;
+  return *this;
+}
+
+RwProto::Cursor RwProto::Cursor::RepeatedFieldIterator::operator*() {
+  return Cursor{it_->value.get(), allocator_};
```
primiano: Okay it seems you deliberately want to create a new cursor every time. In this case plz have an explicit GetCursor() method. Returning a new object from just * is too subtle.
The general expectation is that * just dereferences some pointer you have, doesn't create new things.

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+
+  auto status_conversion = ConvertToMappedRepeatedFieldIfNeeded(
+      *message_field->value, map_key_field_id);
+  PERFETTO_RETURN_IF_STATUS_NOT_OK(status_conversion);
+
+  auto status_or_repeated_field =
+      FindOrCreateMappedRepeatedField(*message_field->value, key);
+  PERFETTO_RETURN_IF_STATUS_NOT_OK(status_or_repeated_field);
+
+  holding_map_and_node_ = {
```
primiano: I'm missing something here: who does the "iterate through all entries until you find a matching value?" I was expecting to see this code somehow

src/protovm/allocator.cc
rukkal: Maybe the best would be a dedicated class (MemoryLimiter?) shared between the two allocators. It just felt a bit overkill, but maybe it's worth it to avoid confusion

src/protovm/allocator.h
rukkal: Why? Those members are initialized in the constructor

src/protovm/free_list_allocator.h
rukkal: `next_slot` never goes out of bounds. We basically do `next_slot = slab[x]` where `x < slab size`, no?

src/protovm/free_list_allocator.h
rukkal: Removed the memory-tracking counter in FLA and pushed all the memory accounting into Allocator. As discussed offline.

rukkal: The parameter got removed

src/protovm/free_list_allocator.h
rukkal: As discussed offline, this class is currently ProtoVM-agnostic and we'll try to keep it as it is. FLA (renamed to SlabAllocator) could even be moved under src/base at some point.

src/protovm/free_list_allocator.h
rukkal: These lines got removed (memory accounting pushed into Allocator)

src/protovm/free_list_allocator.h
rukkal: Got rid of this member variable (memory accounting pushed into Allocator)

src/protovm/allocator.h
rukkal: Renamed to AllocateAndCopyBytes as suggested somewhere else

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+
+StatusOr<void> RwProto::Cursor::EnterField(uint32_t field_id) {
+  PERFETTO_DCHECK(node_);
+
+  auto status_or_it = FindOrCreateMessageField(*node_, field_id);
+  PERFETTO_RETURN_IF_STATUS_NOT_OK(status_or_it);
+  auto it = *status_or_it;
+
+  if (it->value->GetIf<Node::IndexedRepeatedField>()) {
+    PERFETTO_ABORT(
```
primiano: I might have added this comment elsewhere, but I'd rename this to PERFETTO_VM_ABORT as, IIRC, this macro is tied to your special protovm::StatusOr

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+    PERFETTO_ABORT("Attempted \"get scalar\" operation but node has type %s",
+                   node_->GetTypeName());
+  }
+  return *scalar;
+}
+
+StatusOr<void> RwProto::Cursor::SetBytes(protozero::ConstBytes data) {
+  PERFETTO_DCHECK(node_);
+
+  if (node_->GetIf<Scalar>()) {
```
primiano: discussed offline, I'd prefer if the checks were "positive" (i.e. list explicitly the allowed cases).
This code here leaves me with the doubt of "The fact that you allow message<>bytes conversions is deliberate or is a coding accident?)" (I know it's deliberate, but the code doesn't make it clear)

I'd write this as
if (!(GetIf<Empty> || GetIf<Bytes> || GetIf<Message>)) {
 ABORT
}

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+  allocator_->DeleteReferencedData(node_);
+  node_->value = Node::Bytes{std::move(*status_or_bytes)};
+
+  return StatusOr<void>::Ok();
+}
+
+StatusOr<void> RwProto::Cursor::SetScalar(Scalar scalar) {
+  PERFETTO_DCHECK(node_);
+
+  if (node_->GetIf<Node::Bytes>() || node_->GetIf<Node::Message>()) {
```
primiano: same here and everywhere else in this file

src/protovm/allocator.h
rukkal: Added comments

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+StatusOr<void> RwProto::Cursor::Delete() {
+  PERFETTO_DCHECK(node_);
+
+  bool is_root_node = !holding_map_and_node_.first;
+  if (is_root_node) {
+    node_->value = Node::Empty{};
+    return StatusOr<void>::Ok();
+  }
+
+  auto [holding_map, map_node] = holding_map_and_node_;
```
primiano: auto& ? Or do you intend to actually copy ?

src/protovm/allocator.cc
rukkal: Pushed the memory accounting into Allocator, as discussed offline

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+  Node::Message message;
+
+  protozero::ProtoDecoder decoder(protozero::ConstBytes{
+      static_cast<const uint8_t*>(bytes->data.get()), bytes->size});
+
+  for (auto field = decoder.ReadField(); field.valid();
+       field = decoder.ReadField()) {
+    auto status_or_map_value = CreateNodeFromField(field);
+    if (!status_or_map_value.IsOk()) {
+      Node node_tmp{message};
```
primiano: Can you add a comment here? This bit is not fully clear to me what it does

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+
+    auto it = message.field_id_to_node.Find(field.id());
+
+    // First occurrence of this field id. Just insert a new field into the
+    // map.
+    if (!it) {
+      auto status_or_it = MapInsert(message.field_id_to_node, field.id(),
+                                    std::move(*status_or_map_value));
+      if (!status_or_it.IsOk()) {
+        Node node_tmp{message};
```
primiano: Okay yo useem to do this a lot in here.

I'd much rather use base::OnScopeExit (see other usages in the codebase) rather than trying to handle manually all the cleanup cases

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+    protozero::Field field) {
+  if (field.type() == protozero::proto_utils::ProtoWireType::kLengthDelimited) {
+    auto status_or_bytes = allocator_->AllocateAndCopyBytes(field.as_bytes());
+    PERFETTO_RETURN_IF_STATUS_NOT_OK(status_or_bytes);
+    auto status_or_node =
+        allocator_->CreateNode<Node::Bytes>(std::move(*status_or_bytes));
+    if (!status_or_node.IsOk()) {
+      Node node_tmp{std::move(*status_or_bytes)};
+      allocator_->DeleteReferencedData(&node_tmp);
+      PERFETTO_RETURN(status_or_node);
```
primiano: PERFETTO_VM_RETURN (or call these PROTOVM_RETURN / PROTOVM_ABORT)

PERFETTO_XXX is too likely to conflict in future /  or confuse people

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+  if (node.GetIf<Node::MappedRepeatedField>()) {
+    return StatusOr<void>::Ok();
+  }
+
+  if (node.GetIf<Node::Empty>()) {
+    node.value = Node::MappedRepeatedField{};
+    return StatusOr<void>::Ok();
+  }
+
+  if (node.GetIf<Node::Bytes>() || node.GetIf<Node::Message>()) {
```
primiano: can you add some comments here? I think I can guess what this is doing, but some parts are not obvious, especiall the ReadScalarField below (if the node is bytes, what does ReadScalarField do?)

src/protovm/allocator.cc
rukkal: The StatusOr message goes into the stacktrace (owned by StatusOr). This PERFETTO_ELOG is likely a left over. Got rid of it

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+  PERFETTO_DCHECK(map_node);
+  holding_map->Remove(*map_node);
+  allocator_->Delete(&GetOuterNode(*map_node));
+
+  node_ = nullptr;  // Delete operation invalidates cursor
+
+  return StatusOr<void>::Ok();
+}
+
+StatusOr<void> RwProto::Cursor::ConvertToMessageIfNeeded(Node& node) {
```
primiano: Please pass argument either by const& or pointer, not reference, because it's too hard to figure out who makes changes on who while reading the code

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+      allocator_->CreateNode<Scalar>(field.type(), field.as_uint64());
+  if (!status_or_node.IsOk()) {
+    PERFETTO_RETURN(status_or_node);
+  }
+
+  return std::move(*status_or_node);
+}
+
+StatusOr<void> RwProto::Cursor::ConvertToMappedRepeatedFieldIfNeeded(
+    Node& node,
```
primiano: ditto: const Node& or Node*

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+    map_value->value = std::move(node.value);
+    node.value = Node::MappedRepeatedField{};
+
+    auto status_or_it =
+        MapInsert(node.GetIf<Node::MappedRepeatedField>()->key_to_node,
+                  *status_or_key, std::move(map_value));
+    PERFETTO_RETURN(status_or_it);
+  }
+
+  if (auto* indexed = node.GetIf<Node::IndexedRepeatedField>()) {
```
primiano: Okay this whole function needs comment plz. I can only understand half of it and the rest is wild guesses

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+
+StatusOr<void> RwProto::Cursor::ConvertToIndexedRepeatedFieldIfNeeded(
+    Node& node) {
+  if (node.GetIf<Node::IndexedRepeatedField>()) {
+    return StatusOr<void>::Ok();
+  }
+
+  if (node.GetIf<Node::MappedRepeatedField>()) {
+    PERFETTO_ABORT(
+        "Attempted \"convert to intexed repeated field\" operation but "
```
primiano: typo: "intexed"

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+  node.value = Node::IndexedRepeatedField{};
+
+  auto status_or_it =
+      MapInsert(node.GetIf<Node::IndexedRepeatedField>()->index_to_node, 0,
+                std::move(map_value));
+  PERFETTO_RETURN(status_or_it);
+}
+
+StatusOr<IntrusiveMap::Iterator> RwProto::Cursor::FindOrCreateMessageField(
+    Node& node,
```
primiano: ditto (cons& or *)

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+
+  auto status_or_map_value = allocator_->CreateNode<Node::Empty>();
+  PERFETTO_RETURN_IF_STATUS_NOT_OK(status_or_map_value);
+
+  return MapInsert(node.GetIf<Node::IndexedRepeatedField>()->index_to_node,
+                   index, std::move(*status_or_map_value));
+}
+
+StatusOr<IntrusiveMap::Iterator>
+RwProto::Cursor::FindOrCreateMappedRepeatedField(Node& node, uint64_t key) {
```
primiano: ditto: const& or *

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+
+  auto status_or_map_value = allocator_->CreateNode<Node::Empty>();
+  PERFETTO_RETURN_IF_STATUS_NOT_OK(status_or_map_value);
+
+  return MapInsert(node.GetIf<Node::MappedRepeatedField>()->key_to_node, key,
+                   std::move(*status_or_map_value));
+}
+
+StatusOr<IntrusiveMap::Iterator> RwProto::Cursor::MapInsert(
+    IntrusiveMap& map,
```
primiano: ditto

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+      allocator_->CreateNode<Node::MapNode>(key, std::move(map_value));
+  if (!status_or_map_node.IsOk()) {
+    allocator_->Delete(map_value.release());
+    PERFETTO_RETURN(status_or_map_node, "Failed to allocate node");
+  }
+
+  auto [it, inserted] =
+      map.Insert(*status_or_map_node->release()->GetIf<Node::MapNode>());
+  if (!inserted) {
+    allocator_->Delete(map_value.release());
```
primiano: Doesn't the UniquePtr delete itself from the allocator?
i think this needs some dicussion. If you have to delete manualy, then the wrapper object should not be called uniqueptr, because UniquePtr comes with an expectation that out-of-scope == deletion

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+    allocator_->Delete(status_or_map_node->release());
+    PERFETTO_ABORT(
+        "Failed to insert intrusive map entry (key = %d). Duplicated key?",
+        static_cast<int>(key));
+  }
+
+  return it;
+}
+
+StatusOr<uint64_t> RwProto::Cursor::ReadScalarField(Node& node,
```
primiano: ditto: const& or *

src/protovm/rw_proto.cc
```diff
@@ -0,0 +1,705 @@
...
+  for (auto it = message->field_id_to_node.begin(); it; ++it) {
+    auto field_id = static_cast<uint32_t>(it->key);
+    SerializeField(field_id, *it->value, *proto.get());
+  }
+
+  return proto.SerializeAsString();
+}
+
+void RwProto::SerializeField(uint32_t field_id,
+                             Node& node,
```
primiano: ditto (here I think node could be const&) but proto should be a ptr

src/protovm/allocator.cc
rukkal: Closing this as AllocateBytes got inlined in the meantime

src/protovm/allocator.h
rukkal: It's not a cast. It is constructing Node passing the internal value of type T (one of Bytes, Empty, Message, ...)

src/protovm/rw_proto.h
rukkal: Done

rukkal: Done. I factored out the cursor into rw_proto_cursor.h

src/protovm/free_list_allocator.h
rukkal: Done in follow up PR #1348

src/protovm/free_list_allocator.h
rukkal: Resolving this comment as the implementation radically changed in follow up PR #1348

src/protovm/allocator.h
rukkal: Added comments

src/protovm/allocator.h
rukkal: PERFETTO_ABORT (and a couple of other macros) are macros because the capture the __FILE__ and __LINE__ information. I renamed them to PERFETTO_VM_*

src/protovm/allocator.h
rukkal: Added comments

src/protovm/rw_proto.h
rukkal: Done

src/protovm/rw_proto.h
rukkal: SerializeField() is an internal helper recursively called during SerializeAsString(). Just made it private. My bad.

src/protovm/rw_proto.h
rukkal: On the other hand non-const references help with correctness as it can't be null, but you also have a good point. Anyways... your codebase your rules ;) I changed it to pointer

src/protovm/rw_proto.h
rukkal: Done

src/protovm/rw_proto.h
rukkal: uhm the thing is that there isn't an existing cursor for which we can return a reference. We only have internal raw nodes (Node*), which we don't want to expose to the rest of the VM hence we wrap them with Cursor.

src/protovm/rw_proto.h
rukkal: Agree. Added the comment.

src/protovm/rw_proto.h
rukkal: Renamed to EnterRepeatedFieldAt

src/protovm/rw_proto.h
rukkal: done

src/protovm/rw_proto.h
rukkal: done

src/protovm/rw_proto.h
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.h
rukkal: Ok, I have seen the other comment where you got to the implementation details. Renamed function to GetCursor() as you suggested there.

src/protovm/rw_proto.h
rukkal: done. Also done with all the other non-const references passed around in RwProtoCursor

src/protovm/rw_proto.cc
rukkal: it is a copy because holding_map and map_node are raw pointers. Alas, we can't do `auto* [a, b] = ...`

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: Renamed to OwnedPtr

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/free_list_allocator.h
rukkal: Not a leak, but the HWM never goes down.

Anyways.. there is the follow up #1348 with all the allocator improvements we discussed offline

src/protovm/allocator.h
rukkal: Made the change btw

src/protovm/rw_proto.h
rukkal: done

src/protovm/rw_proto.cc
rukkal: The "find matching value" is performed by the FindOrCreateMappedRepeatedField() above. There is no "iteration" involved though, as the values are kept in our IntrusiveTree ordered by key.

The only "iteration" is done in ConvertToMappedRepeatedFieldIfNeeded() (a bit more above) and only once, when a message (e.g. raw bytes) is scanned and each field is added into the IntrusiveTree.

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: I added a couple of helpers to Allocator, so I could get rid of those ugly `Node node_tmp` and hopefully reduce some confusion.

About base::OnScopeExit... how did you think to use it exactly?

One way would be extending it with something like a Release() member:
```
auto message = Node::Message{};

auto deleter = base::OnScopeExit([&](){
   allocator_->DeleteReferencedData(message);
});

...

// Success!
deleter.Release();
return node;
```

another option is making the `allocator_->DeleteReferencedData(message);` a no op in case of success:
```
auto message = Node::Message{};

auto deleter = base::OnScopeExit([&](){
   allocator_->DeleteReferencedData(message);
});

...

// Success!
node.value = std::move(message); // message becomes empty and calling DeleteReferencedData() has no effect
return node;
```

I guess option 2 is cleaner. To make it work we'll have to extend Node::Message and possibly the embedded base::IntrusiveTree with move semanthics

src/protovm/rw_proto.cc
rukkal: done

src/protovm/rw_proto.cc
rukkal: Added a bunch of explanation. Hope they help.

src/protovm/rw_proto.cc
rukkal: let's continue the discussion below

rukkal:
Pushed a bunch of minor changes 

> **Comments inline + an overall comment for Patchset 1 (the allocator part)**
> 
> Maybe it's more future proof if we call that UniquePtr Handle (or AllocatorHandle) so in future can be compacted (and pointers moved) without having to impact the caller.
> 
> I tried to sketch a possible implementation just to give an idea. Again, we don't need the actual compacting allocator here, I'd just introduce a (very dumb) Hadle class to be future-proof.
> 
> You are already in the right state as in your design you need to std::move back the handle into Deallocate() to free the object, which is great.
> 
> ```
> 
> // An allocator that reserves the right to move pointers around
> class CompactingAllocator {
> 
>   std::vector<T> objects;
> 
>   Handle<T> Allocate() {
>     refs.emplace_back(...);
>     ObjRef ref = refs.back();
>     return Handle<T>(ref)
>   }
> 
>   std::list<ObjRef<T>> refs;
> }
> 
> struct ObjRef {
>   T* ptr;
> }
> 
> // A handle is just a pointer to an Object holder.
> // The handle itslef is guarnateed to be a stable pointer
> // (the std::list guarantees that) but the pointed memory
> // can move.
> // Make it move-only, delete copy ctors
> struct Handle<T> {
>   T* operator->() {
>     return static_cast<T>(obj->ptr);
>   }
> 
>   private:
>   Obj<T>* obj;
> }
> ```

As discussed offline, I'd rather postpone the addition of Handle till we find out whether we actually need it, i.e. till we find out whether Scudo can properly handle our blob allocations or not. The RwProto refactoring to introduce Handle should be fairly simple anyways.

src/protovm/rw_proto.cc
rukkal: fyi the code moved to rw_proto_cursor.cc

protos/perfetto/protovm/vm_program.proto
```diff
@@ -0,0 +1,127 @@
...
+
+// Copies SRC into DST. If a message, replaces the DST node, discarding any
+// pre-existing field.
+message VmOpSet {}
+
+// Delete the field or message pointed by DST.
+message VmOpDel {}
+
+enum VmCursorEnum {
+  SRC = 0;
```
primiano: ah i just noticed this sorry. Some google3 linters are very pedantic on wanting the 0 value to be UNSPECIFIED=0, see https://protobuf.dev/best-practices/dos-donts/#unspecified-enum

also can you either nest this enum under VmOpSelect and/or rename these to VM_CURSOR_UNSPECIFIED=0, VM_CURSOR_SRC=1, ...

as-is right now the codegen it's going to create ::perfetto::protos::SRC = 0, which is likely to accumulate naming conflicts over time

src/protovm/allocator.h
```diff
@@ -0,0 +1,93 @@
...
+// the overall proto VM's memory footprint.
+class Allocator {
+ public:
+  explicit Allocator(size_t memory_limit_bytes);
+  StatusOr<Node::Bytes> AllocateAndCopyBytes(protozero::ConstBytes data);
+
+  // Deeply delete a node and all the referenced data. E.g. if the node holds a
+  // Node::Message, recursively delete the message fields and finally delete the
+  // node.
+  void Delete(Node* node);
```
primiano, line 42: out of curiosity, shouldn't this take an OwnedPtr that you move-into delete?

src/protovm/error_handling.h
```diff
@@ -0,0 +1,248 @@
...
+
+  std::unique_ptr<Stacktrace>& GetStacktraceUniquePtr() {
+    PERFETTO_DCHECK(IsAbort());
+    return *std::launder(
+        reinterpret_cast<std::unique_ptr<Stacktrace>*>(storage_));
+  }
+
+  Status status_;
+
+  static constexpr size_t kStorageAlign =
```
primiano, line 238: why all this stuff? why storagE_ is not just a std::unique_ptr<Stacktrace> ? rather than all these placement news etc etc

EDIT: we spoke offline, imho this is a bit overkill. all these placement-new and std::launder to save the space for a std::unique_ptr?

anyhow let's keep this for another cl

src/protovm/executor.h
```diff
@@ -0,0 +1,72 @@
...
+namespace perfetto {
+namespace protovm {
+
+using CursorEnum = perfetto::protos::pbzero::VmCursorEnum;
+
+struct Cursors {
+  RoCursor src;
+  RwProto::Cursor dst;
+  CursorEnum selected;
+  bool create_if_not_exist;
```
primiano: default-initialize these members  to avoid bugs where we forget to initialize them=?

src/protovm/executor.h
```diff
@@ -0,0 +1,72 @@
...
+};
+
+// Executor's methods are virtual to be overridden in tests
+class Executor {
+ public:
+  virtual ~Executor();
+  virtual StatusOr<void> EnterField(Cursors& cursors, uint32_t field_id) const;
+  virtual StatusOr<void> EnterRepeatedFieldAt(Cursors& cursors,
```
primiano: same comment of elsewhere: either `const Cursors&` or `Cursors*`

src/protovm/owned_ptr.h
```diff
@@ -0,0 +1,39 @@
...
+namespace perfetto {
+namespace protovm {
+
+// Custom deleter for std::unique_ptr to ensure the resource is released and its
+// memory deallocated upon destruction.
+struct Deleter {
+  void operator()(void* p) const { PERFETTO_DCHECK(p == nullptr); }
+};
+
+template <class T>
```
primiano, line 46: Add some comment to explain why OwnedPtr exists (the fact that you have a cusotm allocator) and when one is supposed to use this. Otherwise feels like you are just aliasing unique_ptr

src/protovm/rw_proto.h
```diff
@@ -0,0 +1,158 @@
...
+  ~RwProto();
+  Cursor GetRoot();
+  std::string SerializeAsString() const;
+
+ private:
+  void SerializeField(uint32_t field_id,
+                      const Node& node,
+                      protozero::Message* proto) const;
+
+  Allocator* allocator_;
```
primiano: either:
1)make it Allocator* const allocator_;
2) make it = nullptr

to enforce that nobody will forget to initialize this

src/protovm/scalar.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace perfetto {
+namespace protovm {
+
+struct Scalar {
+  static Scalar Fixed32(uint32_t value);
+  static Scalar Fixed64(uint64_t value);
+  static Scalar VarInt(uint64_t value);
+
+  bool operator==(const Scalar& other) const;
```
primiano: honestly tese are so small that is better inlining these functions in the header and getting rid of the .cc. I bet the code that does the check is smaller than emitting a function call, register spill etc etc

same for the static functions above

src/protovm/rw_proto_cursor.h
```diff
@@ -0,0 +1,126 @@
...
+   public:
+    RepeatedFieldIterator();
+    explicit RepeatedFieldIterator(Allocator* allocator,
+                                   IntrusiveMap::Iterator it);
+    RepeatedFieldIterator& operator++();
+    RwProtoCursor GetCursor();
+    explicit operator bool() const;
+
+   private:
+    Allocator* allocator_;
```
primiano: Allocator* const or = nullptr (to prevent accidenally leaving it uninitialized)

src/protovm/rw_proto_cursor.h
```diff
@@ -0,0 +1,126 @@
...
+
+namespace perfetto {
+namespace protovm {
+
+class RwProtoCursor {
+ public:
+  class RepeatedFieldIterator {
+   public:
+    RepeatedFieldIterator();
+    explicit RepeatedFieldIterator(Allocator* allocator,
```
primiano: I don't think explicit is needed here as you have 2 args

src/protovm/scalar.h
rukkal: Agree. I think somehow I got the impression there was a stronger "no header-only function policy" in place

rukkal: Done

src/protovm/rw_proto_cursor.h
rukkal: done

src/protovm/rw_proto_cursor.h
rukkal: Nice catch. Must be some left over

protos/perfetto/protovm/vm_program.proto
rukkal: done

src/protovm/allocator.h
rukkal: Just tried to make the change, but turns out that most of the deletions are performed on raw pointers and it is more convenient to keep `Delete(Node* node)`.

The thing is that OwnedPtr is extensively used within RwProto to clarify ownership while nodes are being passed around between functions, i.e. while the hierarchy is being modified. However, once the hierarchy is stable, most of the nodes are stored as raw pointers within IntrusiveTree.

etiennep-chromium:
This broke chrome autoroller: 
https://ci.chromium.org/ui/p/chromium/builders/try/linux-perfetto-rel/20035/overview
Presumably it's missing #include <new> for std::launder.
Could you please send a fix or revert?

PR #1287, opened 2025-04-15
Improve dataset search efficiency

stevegolton:
This change improves efficiency of the dataset search algorithm by grouping together batches of tracks that have the same dataset `src` running one query per group, rather than a big union all of all tracks.

In addition, rather than embedding the track URI in the query, we resolve the owning track for each event by doing a reverse lookup of the track's partition (filter).

This leads to much more efficient query(s), despite having to run more than one query (one per track group).

### DIFF
--- a/ui/src/core/search_manager.ts
+++ b/ui/src/core/search_manager.ts
@@ -1,31 +1,17 @@
-// Copyright (C) 2024 The Android Open Source Project
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//      http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


ui/src/core/search_manager.ts
```diff
@@ -91,11 +77,16 @@ export class SearchManagerImpl {
...
       this._searchInProgress = true;
       this._limiter.schedule(async () => {
+        const startTime = performance.now();
         if (DATASET_SEARCH.get()) {
           await this.executeDatasetSearch();
         } else {
           await this.executeSearch();
         }
+        const timeTaken = performance.now() - startTime;
+        console.log(
```
LalitMaganti: nit: remove?

ui/src/core/search_manager.ts
```diff
@@ -1,31 +1,17 @@
-// Copyright (C) 2024 The Android Open Source Project
```
LalitMaganti: You probable didn't want to remove this?

LalitMaganti (approved):

ui/src/core/search_manager.ts
stevegolton: Good catch. Reverted.

ui/src/core/search_manager.ts
stevegolton: Done

PR #1293, opened 2025-04-16
Add venv also to Cloud Build docker image

primiano:


primiano:
NOTE: I already pushed this container, just forgot to send a CL to update the dockerfile

LalitMaganti (approved):

PR #1298, opened 2025-04-16
Add option to ignore allowlist for battery plugin

ryan-zuklie:


ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
```diff
@@ -833,10 +836,23 @@ const BT_ACTIVITY = `
...
 `;
 
+let ignoreAllowlist: Setting<boolean> | undefined;
+
 export default class implements PerfettoPlugin {
   static readonly id = 'dev.perfetto.AndroidLongBatteryTracing';
   private readonly groups = new Map<string, TrackNode>();
 
+  static onActivate(app: App) {
+    ignoreAllowlist = app.settings.register({
```
LalitMaganti: Instead of the double negative here, should we instead do the positive (enable session allow list) and then set that to default true?

ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
```diff
@@ -833,10 +836,23 @@ const BT_ACTIVITY = `
   from step2
 `;
 
+let ignoreAllowlist: Setting<boolean> | undefined;
```
LalitMaganti: Steve: is this the right way to do this?

stevegolton: Unfortunately yes. There is no nice way to pass objects from `onActivate` to `onTraceLoad`. You could move the global into the plugin class as a static variable, but you'd need to name the plugin class in order to access it in `onTraceLoad`. Either way the variable is still effectively global, and you have to deal with it possibly being undefined.

An alternative would be to just initialize the setting in `onTraceLoad()`, but then the setting won't be available until a trace loads, so it won't show up in the settings page before a trace is loaded. This may or may not be a problem, depending on how discoverable you want the setting to be.

ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
```diff
@@ -833,10 +836,23 @@ const BT_ACTIVITY = `
...
   private readonly groups = new Map<string, TrackNode>();
 
+  static onActivate(app: App) {
+    ignoreAllowlist = app.settings.register({
+      id: 'dev.perfetto.AndroidLongBatteryTracing#ignoreAllowlist',
+      name: 'Ignore AndroidLongBatteryTracing Session Allowlist',
+      description:
+        'When true, the AndroidLongBatteryTracing plugin will run regardless of session type. This may result in partial or misleading data.',
+      schema: z.boolean(),
+      defaultValue: false,
```
stevegolton: You might want to add `requiresReload` to this setting. See https://github.com/google/perfetto/blob/main/ui/src/plugins/com.example.Settings/index.ts#L99C7-L99C21.

All it does is prompt the user to restart when they change the setting, but I'm assuming it'll be helpful here.

ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
rzuklie: Sure! Done.

ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
rzuklie: I don't particularly mind if the setting isn't available until the trace loads, however, I couldn't figure out how to access the app.settings object in onTraceLoad (which afaict only gets a Trace object). Is this worth trying to chage?

ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
rzuklie: Nice! Done.

ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
LalitMaganti: The trace interface extends app so you should just have .settings available on the trace object.

rzuklie: Ah sorry, I missed that. I updated the code to put the flag entirely in onTraceLoad.

LalitMaganti (approved):

ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
stevegolton: Oh I _think_ I know why the CI is failing. There's a bug in the settings manager where it doesn't remove trace scoped settings before loading a new trace, and registering a new setting with the same ID causes a crash.

I'm OOO until Thursday. I'll take a look then.

If you wanted to land this sooner maybe move back to the previous model (register in onActivate). I think there are other advantages to having the setting available before the trace loads. E.g. quick linking to it via URL args which I'm planning on adding soon.

stevegolton: Actually, the issue should be addressed in: https://github.com/google/perfetto/pull/1330

rzuklie: Thanks for the fix! The presubmit passes now after syncing

Issue #1299, opened 2025-04-16
:trace_processor fails to build on Bazel 8


tomrenn:
Trying to update a Bazel project that uses @perfetto// as an external repository, and getting the following error

```
Traceback (most recent call last):
  File "/usr/local/google/home/tomrenn/.cache/bazel/_bazel_tomrenn/341b79c0f899b42ec9b96031ae4c892a/sandbox/linux-sandbox/18860/execroot/_main/bazel-out/k8-opt-exec-ST-c2fa0d4f630c/bin/external/perfetto/src_trace_processor_tables_tables_python_tool.runfiles/_main/../perfetto/tools/gen_tp_table_headers.py", line 107, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/google/home/tomrenn/.cache/bazel/_bazel_tomrenn/341b79c0f899b42ec9b96031ae4c892a/sandbox/linux-sandbox/18860/execroot/_main/bazel-out/k8-opt-exec-ST-c2fa0d4f630c/bin/external/perfetto/src_trace_processor_tables_tables_python_tool.runfiles/_main/../perfetto/tools/gen_tp_table_headers.py", line 78, in main
    for table in parse_tables_from_modules(modules):
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/tomrenn/.cache/bazel/_bazel_tomrenn/341b79c0f899b42ec9b96031ae4c892a/sandbox/linux-sandbox/18860/execroot/_main/bazel-out/k8-opt-exec-ST-c2fa0d4f630c/bin/external/perfetto/src_trace_processor_tables_tables_python_tool.runfiles/perfetto/python/generators/trace_processor_table/util.py", line 254, in parse_tables_from_modules
    imported = importlib.import_module(module)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/google/home/tomrenn/.cache/bazel/_bazel_tomrenn/341b79c0f899b42ec9b96031ae4c892a/execroot/_main/prebuilts/python/linux-x86/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
[...]

PR #1300, opened 2025-04-16
Fix python tables generation on Bazel 8

tomrenn:
This is aimed at fixing https://github.com/google/perfetto/issues/1299

Previously, build targets like `@perfetto//:src_trace_processor_tables_tables_python_gen` worked because somewhere on the python path was `external/perfetto/src/trace_processor/tables/`, but when starting to use Bazel 8, this `external/perfetto/.*` prefix is no longer available. The python path has `src/trace_processor/tables` available on it.

Instead of trying to fiddle with relative paths and try to identify the repo_name inside starlark, this change modifies the --inputs parameter to pass the modules names directly.


tomrenn:
It looks like builds are failing due to 

AttributeError: 'str' object has no attribute 'removesuffix'

removesuffix was added in python 3.9 - I wonder if it's possible to update the python version on CI? Otherwise I can move that logic to a helper method

LalitMaganti:
We need to stick with Python 3.8 for now until it hits EOL.

tomrenn:
Ack, I updated the commit so that it works on python 3.8. Thanks for the quick review!

LalitMaganti (approved):

Issue #1301, opened 2025-04-17
the trace_process_shell and dev-server ui cannot work on firefox


sydnash:
while i use firefox with the trace_process_shell, there are two problem with it:
1.  firefox send the websocket handshake package with Connection header as:   Connection: keep-alive, Upgrade which case the trace_process_shell cannot recongnize the websocket handshake message.
```
    } else if (hdr_name.CaseInsensitiveEq("connection")) {
        conn->keepalive_ = hdr_value.CaseInsensitiveEq("keep-alive");
        http_req.is_websocket_handshake =
            hdr_value.CaseInsensitiveEq("upgrade");
      }
```

2. in firefox, the websocket message process will out of order because the follow async logic.
```
assertExists(e.data as Blob)
      .arrayBuffer()
      .then((buf) => {
        super.onRpcResponseBytes(new Uint8Array(buf));
      });
```

primiano:
Ah well spotted. Yes these are real bugs as you claim.
I see that you are sending a PR for it, super! thanks so much

PR #1302, opened 2025-04-17
Fix/trace process shell/erorr on firefox

sydnash:
fix firefox with trace_process_shell

issue: https://github.com/google/perfetto/issues/1301

### DIFF
--- a/src/base/http/http_server.cc
+++ b/src/base/http/http_server.cc
@@ -221,9 +221,20 @@ size_t HttpServer::ParseOneHttpRequest(HttpServerConnection* conn) {
         if (IsOriginAllowed(hdr_value))
           conn->origin_allowed_ = hdr_value.ToStdString();
       } else if (hdr_name.CaseInsensitiveEq("connection")) {
-        conn->keepalive_ = hdr_value.CaseInsensitiveEq("keep-alive");
-        http_req.is_websocket_handshake =
-            hdr_value.CaseInsensitiveEq("upgrade");
+        size_t semicolon = hdr_value.find(',');
+        if (semicolon == StringView::npos) {
+          conn->keepalive_ = hdr_value.CaseInsensitiveEq("keep-alive");
+          http_req.is_websocket_handshake =
+              hdr_value.CaseInsensitiveEq("upgrade");
+        } else {
+          auto first = hdr_value.substr(0, semicolon);
+          auto second = hdr_value.substr(semicolon + 2);
+          http_req.is_websocket_handshake =
+              first.CaseInsensitiveEq("upgrade") ||
+              second.CaseInsensitiveEq("upgrade");
+          conn->keepalive_ = first.CaseInsensitiveEq("keep-alive") ||
+                             second.CaseInsensitiveEq("keep-alive");
+        }
       }
     }
   }
--- a/src/base/http/http_server.cc
+++ b/src/base/http/http_server.cc
@@ -221,20 +221,12 @@ size_t HttpServer::ParseOneHttpRequest(HttpServerConnection* conn) {
         if (IsOriginAllowed(hdr_value))
           conn->origin_allowed_ = hdr_value.ToStdString();
       } else if (hdr_name.CaseInsensitiveEq("connection")) {
-        size_t semicolon = hdr_value.find(',');
-        if (semicolon == StringView::npos) {
-          conn->keepalive_ = hdr_value.CaseInsensitiveEq("keep-alive");
-          http_req.is_websocket_handshake =
-              hdr_value.CaseInsensitiveEq("upgrade");
-        } else {
-          auto first = hdr_value.substr(0, semicolon);
-          auto second = hdr_value.substr(semicolon + 2);
-          http_req.is_websocket_handshake =
-              first.CaseInsensitiveEq("upgrade") ||
-              second.CaseInsensitiveEq("upgrade");
-          conn->keepalive_ = first.CaseInsensitiveEq("keep-alive") ||
-                             second.CaseInsensitiveEq("keep-alive");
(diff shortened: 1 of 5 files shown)


src/base/http/http_server.cc
```diff
@@ -221,9 +221,20 @@ size_t HttpServer::ParseOneHttpRequest(HttpServerConnection* conn) {
         if (IsOriginAllowed(hdr_value))
           conn->origin_allowed_ = hdr_value.ToStdString();
       } else if (hdr_name.CaseInsensitiveEq("connection")) {
-        conn->keepalive_ = hdr_value.CaseInsensitiveEq("keep-alive");
-        http_req.is_websocket_handshake =
-            hdr_value.CaseInsensitiveEq("upgrade");
+        size_t semicolon = hdr_value.find(',');
```
LalitMaganti: Can we use base::SplitString to find all the specified options and then handle each one of them instead of doing this ad-hoc one-off splitting?

ui/src/trace_processor/http_rpc_engine.ts
```diff
@@ -86,11 +88,21 @@ export class HttpRpcEngine extends EngineBase {
   }
 
   private onWebsocketMessage(e: MessageEvent) {
-    assertExists(e.data as Blob)
```
LalitMaganti, line 89: Not convinced about adding yet another queue here. Why not just do a await on the arrayBuffer and then call onRpcResponseBytes directly?

src/base/http/http_server.cc
primiano: Or honestly... just look for the presence of the substring, without doing full string matching. THe chance of colliding with something else like "dontkeep-alive" is minimal really. it isn't worth the trouble of splitting even.

ui/src/trace_processor/http_rpc_engine.ts
primiano: I think the theory here (which is realistic) is that in firefox the following sequence can happen:

onWebSocketMessage(1)
arrayBuffer[asyncBegin](1)
onWebSocketMessage(2)
arrayBuffer[asyncBegin](2)
arrayBuffer[asyncEnd](2)  -> calls onRpcResponseBytes(2)
arrayBuffer[asyncEnd](1) -> calls onRpcResponseBytes(1)

ui/src/trace_processor/http_rpc_engine.ts
```diff
@@ -32,6 +32,8 @@ export class HttpRpcEngine extends EngineBase {
   private websocket?: WebSocket;
   private connected = false;
   private disposed = false;
+  private queue: Blob[] = [];
+  private isProcessing = false;
```
primiano: can you call this isProcessingQueue (or isQueueBusy or anything with Queue) so it's clear this is tied to the `queue`?

ui/src/trace_processor/http_rpc_engine.ts
```diff
@@ -86,11 +88,21 @@ export class HttpRpcEngine extends EngineBase {
...
+    this.queue.push(blob);
+    this.processQueue();
+  }
+
+  private async processQueue() {
+    if (this.isProcessing) return;
+    this.isProcessing = true;
+    while (this.queue.length > 0) {
+      const blob = this.queue.shift();
+      if (!blob) continue;
```
primiano: Is there any case when the blob would be falsy?
if the queue length > 0, blob must be defined right?
So this should be a `const blob = assertExists(this.queue.shift())` rather than being defensive

src/base/http/http_server.cc
primiano: Anyhow, either the splitString or the strpos... or base::Contains from string_util.h... you choose

primiano (commented):
Thanks a lot for spotting the issue and sending a patch to this. 
Few comments but overall looks sensible

primiano:
(See comments still open)

sydnash:
> (See comments still open)

I add some comments in the review, maybe we can have a discuss on it.

LalitMaganti:
> I add some comments in the review, maybe we can have a discuss on it.

Sorry but most of our comments do not seem to have been addressed. Please don't mark the comments as resolved unless you have actually resolved them.

primiano (approved):

ui/src/trace_processor/http_rpc_engine.ts
```diff
@@ -86,11 +88,20 @@ export class HttpRpcEngine extends EngineBase {
...
+    this.processQueue();
+  }
+
+  private async processQueue() {
+    if (this.isProcessingQueue) return;
+    this.isProcessingQueue = true;
+    while (this.queue.length > 0) {
+      const blob = assertExists(this.queue.shift());
+      const buf = await blob.arrayBuffer();
+      super.onRpcResponseBytes(new Uint8Array(buf));
```
stevegolton: If anything throws inside this handler could we end up leaving ourselves in an invalid state?

ui/src/trace_processor/http_rpc_engine.ts
```diff
@@ -86,11 +88,30 @@ export class HttpRpcEngine extends EngineBase {
...
         super.onRpcResponseBytes(new Uint8Array(buf));
-      });
+      }
+    } finally {
+      this.isProcessingQueue = false;
+      // if the queue is not empty, we need start a new task
+      // to continue process the remain websocket message.
+      if (this.queue.length > 0) {
+        this.processQueue();
+      }
```
primiano, line 106: isn't this what the while does? can you explain in which case this is needed?
of is it for the exception case?

stevegolton: I think the try..catch could be moved inside the while loop and we could omit this bit

stevegolton (approved):

PR #1306, opened 2025-04-17
Remove default argument from ProducerEndpoint::CreateTraceWriter

ddiproietto:
Default arguments are banned in virtual functions by the Google C++ style.

I didn't find any use of the default argument outside of the perfetto repo.

Welcome to Perfetto!

For more details please see


rsavitski (commented):
LGTM with two points:
1) please remove the "welcome to perfetto" part of the PR description next time (and make sure it doens't end up as part of the merged commit)

2) I think a couple of chromium tests might break, so we'll need to send them a patch before this patch can be autorolled? e.g. https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/tracing/test/tracing_integration_test.cc;l=261;drc=2ea511fc39b2923a38a4d01ab063f4a6530047d6

rsavitski (approved):

PR #1312, opened 2025-04-17
tracing: Add configurable buffer exhausted policy

ddiproietto:
This commit introduces a new field in the config
(DataSourceConfig.buffer_exhausted_policy) that can be used to change the behavior of a data source instance when the shared memory buffer is exhausted.

N.B.: Only data sources explicitly registered with BufferExhaustedPolicy::kConfigurable (or
PERFETTO_BUFFER_EXHAUSTED_POLICY_CONFIGURABLE) respect this config field.

Bug: 384007571
Tested: unsubmitted (it wastes 20 seconds) integration test


LalitMaganti (approved):

include/perfetto/public/abi/data_source_abi.h
```diff
@@ -227,6 +227,9 @@ enum PerfettoDsBufferExhaustedPolicy {
   // it will stall, retry and eventually drop data if a free chunk is not
   // acquired after a few seconds.
   PERFETTO_DS_BUFFER_EXHAUSTED_POLICY_STALL_AND_DROP = 2,
+  // Equivalent to PERFETTO_DS_BUFFER_EXHAUSTED_POLICY_DROP by default, but can
+  // be changed using the tracing configuration.
+  PERFETTO_DS_BUFFER_EXHAUSTED_POLICY_CONFIGURABLE = 3,
```
primiano: TBH the fact that this is DROP by default but it's another entry is a bit arbitray.
IMHO would be a bit less surprising if conceptually the "configurable" is orthogonal (i.e. it's a boolean on the side, not an alternative to the policy).
I would still use PerfettoDsSetBufferExhaustedPolicy by doing

PerfettoDsSetBufferExhaustedPolicy(DROP | CONFIGURABLE)
PerfettoDsSetBufferExhaustedPolicy(STALL_AND_DROP | CONFIGURABLE)

and by making CONFIGURABLE  a high bit, like 0x80000000

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -161,6 +161,18 @@ struct CompareBackendByType {
...
 
+BufferExhaustedPolicy ParseBufferExhaustedPolicy(const DataSourceConfig& cfg) {
+  switch (cfg.buffer_exhausted_policy()) {
+    case DataSourceConfig::BUFFER_EXHAUSTED_DROP:
+      return BufferExhaustedPolicy::kDrop;
+    case DataSourceConfig::BUFFER_EXHAUSTED_STALL:
+      return BufferExhaustedPolicy::kStall;
+    case DataSourceConfig::BUFFER_EXHAUSTED_STALL_THEN_DROP:
+      return BufferExhaustedPolicy::kStallThenDrop;
+  }
```
primiano, line 180: Here the fact that you don't handle CONFIGURABLE seems a  bug, rather than being deliberate.
IMHO a dedicate boolean is clearer

include/perfetto/public/abi/data_source_abi.h
ddiproietto: Good point.

I'm using a separate boolean instead.

src/tracing/internal/tracing_muxer_impl.cc
ddiproietto: Fair. I'm using a separate boolean now

protos/perfetto/config/data_source_config.proto
```diff
@@ -108,7 +108,25 @@ message DataSourceConfig {
   // DO NOT SET in consumer as this will be overridden by the service.
   optional uint64 tracing_session_id = 4;
 
-  // Keeep the lower IDs (up to 99) for fields that are *not* specific to
+  enum BufferExhaustedPolicy {
+    // The data source will drop packets when there's no space in the shared
+    // memory buffer.
+    BUFFER_EXHAUSTED_DROP = 0;
```
primiano: This needs to be BUFFER_EXHAUSTED_UNSPECIFIED = 0 as per https://protobuf.dev/best-practices/dos-donts/#unspecified-enum

primiano (approved):
LGTM % UNSPECIFIED=0

protos/perfetto/config/data_source_config.proto
ddiproietto: Good idea, done

PR #1325, opened 2025-04-18
Add track descriptions

stevegolton:
This PR adds track descriptions to track definitions.

Plugins can attach descriptions (short strings of text) that are displayed in the track details popup.

In the future we could make this style of help text more visible and more consistent throughout the UI, but for now this change provides the plumbing to allow plugins to specify descriptions.

image

### DIFF
--- a/ui/src/frontend/viewer_page/track_view.ts
+++ b/ui/src/frontend/viewer_page/track_view.ts
@@ -637,6 +637,8 @@ function renderTrackDetailsMenu(node: TrackNode, descriptor?: Track) {
       }),
       m(TreeNode, {left: 'Path', right: fullPath}),
       m(TreeNode, {left: 'Title', right: node.title}),
+      descriptor &&
+        m(TreeNode, {left: 'Description', right: descriptor?.description}),
       m(TreeNode, {
         left: 'Workspace',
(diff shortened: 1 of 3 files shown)


LalitMaganti:
While descriptions are definitely a start and we should merge this, I think the concept of a "help text" is different from a description. 

A description to me is what information the track contains. This can come from the trace itself. The help text I think is more "this is how this track should be used to solve x y z problems, maybe link to more rich docs".

stevegolton:
> While descriptions are definitely a start and we should merge this, I think the concept of a "help text" is different from a description. 
> 
> A description to me is what information the track contains. This can come from the trace itself. The help text I think is more "this is how this track should be used to solve x y z problems, maybe link to more rich docs". 

I see what you mean, though in practice I think it might be a stretch to ask plugin developers to add both a description _and_ help text.

I assume that the plugins that plumb descriptions from the trace itself would mostly be the automatic bulk track adding plugins such as TrackEvent and friends, in which case where would the help text come from?

And plugins that add plugin-only tracks would add descriptions manually, in which case they would have to add both descriptions and help text which I don't think they'd bother to do.

Let's land this and see if we can make do with one field, then we can always add another if necessary.

PR #1328, opened 2025-04-21
Update proto_library template to propagate indirect input deps correctly.

atetubou:
This is necessary for https://crrev.com/c/6442961


atetubou:
Can I get additional reviewer for this PR?

PR #1334, opened 2025-04-22
ftrace: Add new "linux.frozen_ftrace" data source

mhiramat:
Add new "linux.frozen_ftrace" data source for reading the remaining previous boot data from the ftrace persistent ring buffer. Note that this data source does not actively record any trace data for the current boot. Thus you must not mix this with other data sources in the same session.

This new data source requires 2 parameters;

 - instance_name: Specifies an instance keeping the previous boot data.

 - event_format_path: Specifies a directory which stores the event format files saved in the previous boot time.

Change-Id: Ic1f61f6141d019c569b6f67425f331632c1985b0

Welcome to Perfetto!

For more details please see


src/traced/probes/ftrace/frozen_ftrace_procfs.h
```diff
@@ -0,0 +1,71 @@
...
+#include <memory>
+#include <set>
+#include <string>
+#include <vector>
+
+#include "src/traced/probes/ftrace/ftrace_procfs.h"
+
+namespace perfetto {
+
+class FrozenFtraceProcfs : public FtraceProcfs {
```
rsavitski: Per b/409925392#comment11, we shouldn't be subclassing any ftrace classes.

You should use FtraceProcfs directly and trust that ProtoTranslationTable will not call the writing APIs. Consider that the implementation currently only calls one method, and that has no reason to change: https://osscs.corp.google.com/search?q=f:proto_translation%20trace_procfs_&sq=&ss=chromium%2Fchromium%2Fsrc:third_party%2Fperfetto%2Fsrc%2Ftraced%2Fprobes%2Fftrace%2F.

As far as I understand the |event_format_path| is also no longer necessary, so there's even less reason for this class to exist.

(You could've made a new "readonly procfs" interface that FtraceProcfs itself implements, and changed ProtoTranslationTable to expect any implementer of that interface, but it's not worth the refactoring effort here).

src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -228,12 +226,66 @@ size_t CpuReader::ReadCycle(
...
+    ProcessBatch(data_source->trace_writer(), data_source->mutable_metadata(),
+                 data_source->parsing_config(),
+                 data_source->mutable_parse_errors(),
+                 data_source->mutable_bundle_end_timestamp(cpu_), parsing_bufs,
+                 pages_read);
+  }
+  return pages_read;
+}
+
+size_t CpuReader::ReadFrozen(
```
rsavitski: Please use the structure from [1]. We would very strongly prefer to not touch /any/ of the existing CpuReader functions for this patch, even if something looks like something could be refactored to be shared. 

We would prefer to maintain a straight-line function that is mostly a copy of ReadCycle + ReadAndProcessBatch, but which can be simplified because the tracefs instance is supposed to be stopped (which in turn means that a lot of ReadAndProcessBatch is not relevant to your use-case - this code won't be continuously chasing the kernel writer).

[1] https://github.com/google/perfetto/commit/5c67ea4159bc2ff2d84521faa14ee6f06f33fbcf#diff-e1430d29cfe69b77e177a7bcf04110bade9258d096ec87dbf49f2791f095614e

protos/perfetto/config/ftrace/frozen_ftrace_config.proto
```diff
@@ -0,0 +1,28 @@
...
+
+package perfetto.protos;
+
+// Next id: 3
+message FrozenFtraceConfig {
+  // The instance name which stores the previous boot ftrace data.
+  required string instance_name = 1;
+  // The event format directory (absolute path) which will be used for
+  // searching event ids and formats.
+  optional string event_format_path = 2;
```
rsavitski: I believe we can drop "event_format_path" from the config now that we've established that it's fine to use the instance's events/.../format files (at least for static kernel tracepoints)?

src/traced/probes/ftrace/ftrace_procfs.cc
```diff
@@ -463,6 +463,17 @@ bool FtraceProcfs::SetCpuBufferSizeInPages(size_t pages) {
   return WriteNumberToFile(path, pages * (base::GetSysPageSize() / 1024ul));
 }
 
+size_t FtraceProcfs::GetCpuBufferSizeInPages() {
```
rsavitski, line 468: Passing observation, this output is possible for a fresh instance:

/sys/kernel/tracing > cat buffer_size_kb
7 (expanded: 1408)

Though I suppose returning 0 for that case is good enough, as we're not likely to hit this case in practice.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
```diff
@@ -0,0 +1,208 @@
...
+    return;
+
+  // Enable all events in the translation table and all syscalls because
+  // the previous boot trace data may record any events.
+  EventFilter event_filter;
+  for (const auto& event : translation_table_->events()) {
+    event_filter.AddEnabledEvent(event.ftrace_event_id);
+  }
+
+  SyscallTable syscalls = SyscallTable::FromCurrentArch();
```
rsavitski: I don't believe you need the syscall_filter at all. It's for setting up tracepoint filters so that a only subset of syscalls is traced via 'raw_syscalls/sys_{enter,exit}' (which otherwise can be quite high volume since they cover all syscalls on the system).

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
```diff
@@ -0,0 +1,208 @@
...
+  if (!translation_table_) {
+    PERFETTO_ELOG("Failed to create translation table.");
+    return;
+  }
+
+  // Assumes the same core count as currently. If not, the previous boot
+  // data is cleared because of the failure of buffer metadata validation.
+  size_t num_cpus = tracefs_->NumberOfCpus();
+
+  // TODO: assumes the tracefs instance is using CLOCK_BOOTTIME (encoded as
```
rsavitski: To clarify, the "TODO:" comments from my prototype are not to be submitted, I was just leaving notes for you. If you do want to leave a todo, use the "TODO(mhiramat):" prefix.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
```diff
@@ -0,0 +1,208 @@
...
+  // ipc/tasks to run inbetween.
+  task_runner_->PostTask([weak_this = weak_factory_.GetWeakPtr()] {
+    if (weak_this)
+      weak_this->ReadTask();
+  });
+}
+
+void FrozenFtraceDataSource::ReadTask() {
+  // For the previous boot trace data, we don't have any metadata.
+  FtraceMetadata metadata{};
```
rsavitski: Then there's not much point to be passing it into ReadFrozen, right?

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
```diff
@@ -0,0 +1,208 @@
...
+
+  bool all_cpus_done = true;
+  for (size_t i = 0; i < cpu_readers_.size(); i++) {
+    size_t max_pages = cpu_page_quota_[i];
```
rsavitski: Please re-read the following comment (I believe I rephrased the same thing elsewhere on the bug/prototype/chat): https://osscs.corp.google.com/chromium/chromium/src/+/main:third_party/perfetto/src/traced/probes/ftrace/ftrace_controller.cc;l=298;drc=072b5d54d45f770c280c3ed933d12c37da34b6db

We want to break up the work into multiple ReadTask() calls, whereas this seems to be trying to read all pages in one go.

src/traced/probes/ftrace/frozen_ftrace_data_source.h
```diff
@@ -0,0 +1,98 @@
...
+  std::unique_ptr<FrozenFtraceProcfs> tracefs_;
+  std::unique_ptr<ProtoTranslationTable> translation_table_;
+  std::unique_ptr<FtraceDataSourceConfig> parsing_config_;
+  CpuReader::ParsingBuffers parsing_mem_;
+  std::vector<CpuReader> cpu_readers_;
+
+  std::vector<size_t> cpu_page_quota_;
+
+  base::FlatSet<protos::pbzero::FtraceParseStatus> parse_errors_;
+  std::vector<uint64_t> bundle_end_ts_by_cpu_;
```
rsavitski: You don't need this at all, it's for dealing with streaming recording and ring buffers (note that your current patch set doesn't make use of the values).

perfetto.rc
```diff
@@ -193,3 +193,27 @@ service perfetto_persistent_sysui_tracing_for_bugreport /system/bin/perfetto -c
     restart_period 60
     user nobody
     group nobody
+
+##############################################################################
```
rsavitski: I need to check with a teammate for a second opinion, but expect pushback on this entire block. Adding a data source to perfetto that can convert a stopped tracefs instance into a trace makes sense, but creating tracefs instances and kicking off the recording should probably live in device-specific rc files.

perfetto.rc
```diff
@@ -193,3 +193,27 @@ service perfetto_persistent_sysui_tracing_for_bugreport /system/bin/perfetto -c
...
+# perfetto_prev_boot_dump_for_bugreport - Dump the previous boot tracing data
+# from the ftrace persistent ring buffer.
+##############################################################################
+
+on perfetto_maybe_trace_dump_on_boot && property:persist.debug.perfetto.prev_boot_dump=1 && property:persist.traced.enable=1
+    # If there is no "boot_mapped" instance, make it.
+    mkdir /sys/kernel/tracing/instances/boot_mapped 0770 root shell
```
rsavitski: This looks to be defeating the entire purpose of tracefs ACLs on android, shell is not supposed to have access to a raw unredacted tracefs instance (and selinux will stop you, rightfully so).

perfetto.rc
```diff
@@ -193,3 +193,27 @@ service perfetto_persistent_sysui_tracing_for_bugreport /system/bin/perfetto -c
     restart_period 60
     user nobody
     group nobody
+
+##############################################################################
+# perfetto_prev_boot_dump_for_bugreport - Dump the previous boot tracing data
+# from the ftrace persistent ring buffer.
+##############################################################################
+
+on perfetto_maybe_trace_dump_on_boot && property:persist.debug.perfetto.prev_boot_dump=1 && property:persist.traced.enable=1
```
rsavitski: For the "persist.debug.perfetto.prev_boot_dump" sysprop, it's unclear:
* who sets this?
* who's allowed to set it? (debug.* is generally unrestricted, while you probably want a system-owned sysprop)
* are you intending to convert the boot_mapped instance on every reboot, regardless of there being a crash?

Let's follow up outside of the patch review.

rsavitski (changes requested):
Made a pass, and will follow up on a chat about things not strictly related to the patch.

rsavitski:
Oh and also, you may find "gn check out/your_output_dir" useful to fix at least the GN presubmit errors on github actions.

src/traced/probes/ftrace/frozen_ftrace_procfs.h
mhiramat: OK. BTW, just out of curiosity, when we need to support backed up eventfs for modules/dynamic events, we'll revisit to drive the FtraceProcfs, or keep it and just add a method to specify |event_format_path|?

protos/perfetto/config/ftrace/frozen_ftrace_config.proto
mhiramat: Yes, I just kept the "event_format_path" for future use. But to simplify the code for this commit, it should be dropped. Thanks!

src/traced/probes/ftrace/ftrace_procfs.cc
mhiramat: Ah, I forgot that case, maybe better to drop after `(`?. Shouldn't we return 1 or 2 (round up) in such case?

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: I would not like to drop such syscall events if it is recorded in the persistent ring buffer. This depends on user usage. If perfetto can set both `instance_name` and syscall events in the same ftrace data source, user can record some specific syscalls. If perfetto only supports syscall recording on the root instance, I agree with dropping this filter (as same as atrace filters).

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: OK, let me fix that. Thanks!

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: Good catch! I didn't notice that I can pass nullptr to it. Thanks!

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: OK, I'll introduce a kMaxReadPages=32 to limit the max_pages for this layer.

src/traced/probes/ftrace/frozen_ftrace_data_source.h
mhiramat: Got it. Let me drop it. Thanks!

perfetto.rc
mhiramat: OK, it makes sense. I might be better to move it to device specific rc file with kernel cmdline option. Thanks,

perfetto.rc
mhiramat: OK, anyway if I can move tihs into the device specific rc, I think we don't need it because boot_mapped should always be there.

perfetto.rc
mhiramat: Thanks for the comment! I agree it should be designed  more carefully.

src/traced/probes/ftrace/cpu_reader.cc
mhiramat: OK, then I just clone the code.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: It seems that the metadata should not be a nullptr. See CpuReader::Bundler. The comment said // Never nullptr.

mhiramat: It seems metadata stores parsed common field like pids. So it is better to use a common object. Let me make it a member of FrozenFtraceDataSource.

src/traced/probes/ftrace/ftrace_procfs.cc
mhiramat: I decided to just notice that in the comment. Anyway, that is not a problem in this case.

src/traced/probes/ftrace/frozen_ftrace_procfs.h
rsavitski: IMO, it's probably at that point that we can split out a interface for "minimal Tracefs abstraction methods for ProtoTranslationTable's needs", and make both the FtraceProcfs and the new FrozenFtraceProcfs implement that interface.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
rsavitski: What I meant is: syscall_filter is used to _configure_ recording. It does not affect parsing the tracefs buffers and writing protobufs right now.

If `raw_syscalls/...` events exist in the frozen instance, they will be written into the trace. Regardless of this filter.

It's a case of misleading code structure on our side, even though it's part of "FtraceDataSourceConfig" passed to CpuReader, the latter doesn't actually use the value. We'll look into refactoring this later.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
rsavitski: Fine to keep as is, but my point is that there isn't a point to keeping onto the FtraceMetadata. Consider the comment from [1] - the metadata is supposed to be a temporary container used to hand over miscellaneous information from cpu_reader back to other data sources in traced_probes (concretely - the procfs scraper that gathers extra info about referenced thread ids).

You could construct the `FtraceMetadata` as a stack variable in CpuReader::ReadFrozen, so that it gets thrown away even if it's a mandatory input to `ProcessPagesForDataSource`.

But to reiterate, if you want this metadata object to live as long as the FrozenFtraceDataSource, I'm fine with it.

[1] https://osscs.corp.google.com/chromium/chromium/src/+/main:third_party/perfetto/src/traced/probes/ftrace/ftrace_metadata.h;l=35?q=ftrace_metadata&ss=chromium%2Fchromium%2Fsrc:third_party%2Fperfetto%2F

src/traced/probes/ftrace/ftrace_procfs.cc
rsavitski: Yeah, it'd technically be more correct to return 1 or 2 pages for that case, but I don't think this corner case will come up in practice. (Plus I believe it means that the trace buffer is empty anyway?).

protos/perfetto/config/ftrace/frozen_ftrace_config.proto
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Next id: 3
```
rsavitski: nit: either "2" or remove the comment altogether, it's only helpful when a proto file grows big.

protos/perfetto/config/ftrace/frozen_ftrace_config.proto
rsavitski: Thanks, we prefer not to commit "for future" definitions or commented out code, as it hurts maintainability.

prev_boot_dump.pbtxt
```diff
@@ -0,0 +1,15 @@
+buffers {
+      size_kb: 65536
```
rsavitski: nit: never tabs, always 2-char space indentation.

Android.bp.extras
```diff
@@ -255,6 +255,13 @@ prebuilt_etc {
     src: "persistent_cfg.pbtxt",
 }
 
+prebuilt_etc {
```
rsavitski: I would /prefer/ if this was defined with the rest of the boot_mapped tracing rc files, and not on core perfetto team to maintain, especially if it'll be referenced by those rc files. 

But if there's no other place to put it, you can leave this here.

src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -1136,4 +1145,61 @@ void CpuReader::ParseSchedWakingCompact(const uint8_t* start,
...
+    }
+    pages_read += 1;
+  }
+
+  if (pages_read == 0)
+    return pages_read;
+
+  // Inputs that we will throw away since we only need a subset of what
+  // FtraceDataSource does.
+  // TODO: consider making use of |parse_errors|.
```
rsavitski: nit: the todo is no longer relevant

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
```diff
@@ -0,0 +1,197 @@
...
+  const auto ftrace_clock =
+      protos::pbzero::FtraceClock::FTRACE_CLOCK_UNSPECIFIED;
+
+  // To avoid reading pages more than expected, record remaining pages.
+  size_t initial_page_quota = tracefs_->GetCpuBufferSizeInPages() + 1;
+
+  PERFETTO_CHECK(cpu_readers_.empty());
+  cpu_readers_.reserve(num_cpus);
+  for (size_t cpu = 0; cpu < num_cpus; cpu++) {
+    // TODO: the two nullptrs are ok, the implementation treats those pointers
```
rsavitski: nit: stale TODO, probably remove it altogether.

docs/data-sources/previous-boot-trace.md
```diff
@@ -0,0 +1,71 @@
...
+data_sources {
+ config {
+   name: "linux.frozen_ftrace"
+   frozen_ftrace_config {
+     instance_name: "boot_mapped"
+    }
+  }
+}
+
+duration_ms: 5000
```
rsavitski, line 70: Tell us if a reliance on a hardcoded duration becomes a problem. Right now the implementation is technically a bit racy since the data source doesn't have a way to delay itself from being stopped (if it hasn't finished reading out all of the buffers).

The "data source isn't done yet" would normally be handled by StopDataSource() delaying the ACK to the service, but ProbesProducer [1] currently doesn't let ProbesDataSource implementers from achieving that. (Existing data sources instead delay the Flush(), but it doesn't apply to this data source since there is no concept of "commit data up to current timestamp" for it).

[1] https://osscs.corp.google.com/chromium/chromium/src/+/main:third_party/perfetto/src/traced/probes/probes_producer.cc;l=500;drc=6a643f3561eceafe50d0c5dfed09d665523c4b74

src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -1136,4 +1145,61 @@ void CpuReader::ParseSchedWakingCompact(const uint8_t* start,
   compact_buf->sched_waking().common_flags().Append(common_flags);
 }
 
+size_t CpuReader::ReadFrozen(
```
rsavitski, line 1148: Thanks for the updates, I think this looks reasonable.

For programmatic tests, you may want to add one to cpu_reader_unittest.cc for ReadFrozen. The last couple of test cases in that file should be a reasonable template. For example: https://osscs.corp.google.com/chromium/chromium/src/+/main:third_party/perfetto/src/traced/probes/ftrace/cpu_reader_unittest.cc;l=3504;drc=4c920851b488f8f33348a3b7b6a055da0b999b5d

(with the gn+ninja dev environment, you can run the tests using: 
```
ninja -C out/OUTPUTDIR perfetto_unittests && ninja out/OUTPUTDIR/perfetto_unittests --gtest_filter="CpuReaderTest*"
```

For testing FrozenFtraceDataSource, you /could/ take a look at [ftrace_controller_unittest.cc](https://osscs.corp.google.com/chromium/chromium/src/+/main:third_party/perfetto/src/traced/probes/ftrace/ftrace_controller_unittest.cc), but the mocking-based tests often end up not testing much in reality. So if you can't come up with a consise test for the DataSource, then we can skip tests for it for now.

rsavitski (changes requested):
Looking much better, thank you. See the comment on CpuReader::ReadFrozen for suggestions on programmatic tests.

Android.bp.extras
mhiramat: Aaah, good catch! Yes, I should move this with init script part. Thanks!

docs/data-sources/previous-boot-trace.md
mhiramat: Yeah, when I tried with "duration_ms: 1000", it failed to dump all data.

Perhups, I should try "write_into_file" option.

prev_boot_dump.pbtxt
mhiramat: OK. It seems that only this part used a tab.

protos/perfetto/config/ftrace/frozen_ftrace_config.proto
mhiramat: OK, thanks!

src/traced/probes/ftrace/cpu_reader.cc
mhiramat: OK

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: OK.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: Ah, I thought that it also filtered the input. OK, then remove it. Thanks!

src/traced/probes/ftrace/cpu_reader.cc
mhiramat: OK, let me add a unittests.

mhiramat: Hmm, I found that the ReadFrozen is not good for actual test because it directly read() from tracefs. We may need an intermadiate object or test file. I guess this is why `ReadAndProcessBatch` also doesn't have a unit test.

docs/data-sources/previous-boot-trace.md
mhiramat: `write_into_file` does not help because of permission design. Accoding to my observation, with this option, traced will write into file instead of perfetto, and that is not permitted.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: OK. If there is no negative effect, let's keep this as it is.

src/traced/probes/ftrace/ftrace_procfs.cc
mhiramat: OK, I found `FtraceConfigMuxer::ComputeCpuBufferSizeInPages()` returns >=1. So it should return more than 1.

mhiramat:
I added a patch to add a unit test of `CpuReader::ReadFrozen()`. A unique point of this test is that made a temporary per-cpu file to read.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
```diff
@@ -0,0 +1,187 @@
...
+      weak_factory_(this) {
+  // This should check the required parameters.
+  ds_config_.ParseFromString(ds_config.frozen_ftrace_config_raw());
+}
+
+void FrozenFtraceDataSource::Start() {
+  parsing_mem_.AllocateIfNeeded();
+
+  tracefs_ = FtraceProcfs::CreateGuessingMountPoint(
+      "instances/" + ds_config_.instance_name() + "/");
```
rsavitski: We should validate the instance_name similar to [1]. There is never a reason for the path to escape the tracefs (i.e. /sys/kernel/tracing), and we don't want to deal with the possibility of confused deputy attacks via specially-constructed paths.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/traced/probes/ftrace/ftrace_controller.cc;l=869;drc=072b5d54d45f770c280c3ed933d12c37da34b6db

docs/data-sources/previous-boot-trace.md
rsavitski: > traced will write into file instead of perfetto, and that is not permitted.

What destination paths are you testing with? On android, /data/misc/perfetto-traces is the canonical place for trace outputs, and is writeable by `traced`.

But write_info_file is unrelated to the issue described by my original post (which still has an accurate summary of why there's a need for "long-enough" duration_ms at the moment).

src/traced/probes/ftrace/cpu_reader_unittest.cc
```diff
@@ -3859,5 +3859,93 @@ TEST(CpuReaderTest, GenericEventSelfDescribingFormat) {
   // (skip verifying last switch)
 }
 
+// Tests that |ReadFrozen()| reads the specified file and parse it correctly.
+TEST(CpuReaderTest, FrozenPageRead) {
```
rsavitski, line 3864: couple of things:
1) We're testing CpuReader, which needs a file descriptor to read from (it'll assume that it is a trace pipe from tracefs). In the test, we can just pass a temporary file, without needing to create a fake tracefs directory structure. See below for a suggestion which just uses base::TempFile, and doesn't require adding anything to tmp_dir_tree.{h,cc}. See bottom of this message for a suggestion.

2) The assertion for `FTRACE_STATUS_PARTIAL_PAGE_READ` seems misleading. While you hit this codepath with a test file, the actual trace pipe doesn't behave this way. We expect an EAGAIN if there are no more events, not a read of zero bytes. Therefore we have two options: either make the second read for exactly 6 pages, or continue asking for 32 pages, but do not assert a specific parsing error. I leave it up to you.

Here's the version with a base::TempFile:

```
// Tests that |ReadFrozen()| reads the specified file and parse it correctly.
TEST(CpuReaderTest, FrozenPageRead) {
  auto page_ok = PageFromXxd(g_switch_page);
  auto page_loss = PageFromXxd(g_switch_page_lost_events);

  size_t page_sz = base::GetSysPageSize();
  base::TempFile test_pages = base::TempFile::CreateUnlinked();
  base::WriteAll(test_pages.fd(), page_ok.get(), page_sz);
  base::WriteAll(test_pages.fd(), page_ok.get(), page_sz);
  base::WriteAll(test_pages.fd(), page_ok.get(), page_sz);
  base::WriteAll(test_pages.fd(), page_loss.get(), page_sz);
  base::WriteAll(test_pages.fd(), page_loss.get(), page_sz);
  base::WriteAll(test_pages.fd(), page_ok.get(), page_sz);
  base::WriteAll(test_pages.fd(), page_ok.get(), page_sz);
  base::WriteAll(test_pages.fd(), page_ok.get(), page_sz);
  ASSERT_NE(lseek(test_pages.fd(), 0, SEEK_SET), -1);

  ProtoTranslationTable* table = GetTable("synthetic");
  CpuReader cpu_reader(0, test_pages.ReleaseFD(), table,
                       /*symbolizer=*/nullptr,
                       protos::pbzero::FTRACE_CLOCK_UNSPECIFIED,
                       /*ftrace_clock_snapshot=*/nullptr);

  // build cfg requesting sched/sched_switch
  FtraceMetadata metadata{};
  FtraceDataSourceConfig ftrace_cfg = EmptyConfig();
  ftrace_cfg.event_filter.AddEnabledEvent(
      table->EventToFtraceId(GroupAndName("sched", "sched_switch")));

  // invoke ReadFrozen
  TraceWriterForTesting trace_writer;
  base::FlatSet<protos::pbzero::FtraceParseStatus> parse_errors;
  CpuReader::ParsingBuffers parsing_mem;
  parsing_mem.AllocateIfNeeded();

  // read 2 pages because max_pages limits it.
  size_t pages_read = cpu_reader.ReadFrozen(
      &parsing_mem, 2, &ftrace_cfg, &metadata, &parse_errors, &trace_writer);
  EXPECT_EQ(2u, pages_read);
  EXPECT_EQ(0u, parse_errors.size());

  // first 2 pages has 2 events without any data lost.
  auto first_packets = trace_writer.GetAllTracePackets();
  ASSERT_EQ(1u, first_packets.size());

  EXPECT_FALSE(first_packets[0].ftrace_events().lost_events());
  EXPECT_EQ(2u, first_packets[0].ftrace_events().event().size());

  // read remaining pages.
  pages_read = cpu_reader.ReadFrozen(&parsing_mem, 6, &ftrace_cfg, &metadata,
                                     &parse_errors, &trace_writer);
  EXPECT_EQ(6u, pages_read);
  EXPECT_EQ(0u, parse_errors.size());

  // Each packet should contain the parsed contents of a contiguous run of pages
  // without data loss.
  // So we should get three packets (each page has 1 event):
  //   [2 events (previous ReadFrozen)] [1 events] [1 event] [4 events].
  auto packets = trace_writer.GetAllTracePackets();
  ASSERT_EQ(4u, packets.size());
  EXPECT_FALSE(packets[0].ftrace_events().lost_events());
  EXPECT_EQ(2u, packets[0].ftrace_events().event().size());

  EXPECT_FALSE(packets[1].ftrace_events().lost_events());
  EXPECT_EQ(1u, packets[1].ftrace_events().event().size());

  EXPECT_TRUE(packets[2].ftrace_events().lost_events());
  EXPECT_EQ(1u, packets[2].ftrace_events().event().size());

  EXPECT_TRUE(packets[3].ftrace_events().lost_events());
  EXPECT_EQ(4u, packets[3].ftrace_events().event().size());
}
```

src/base/test/tmp_dir_tree.h
```diff
@@ -47,6 +48,12 @@ class TmpDirTree {
   // succeeds.
   void AddFile(const std::string& relative_path, const std::string& content);
 
+  // Creates a file at `relative_path` which contains binary data `content`.
+  // All the parent directories should have been created. PERFETTO_CHECK()s that
+  // the operation succeeds.
+  void AddBinaryFile(const std::string& relative_path,
```
rsavitski: see the comment on cpu_reader_unittest.cc, I believe we don't need this and can just use base::TempFile.

rsavitski (changes requested):

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: OK, let me validate the instance_name.

src/traced/probes/ftrace/cpu_reader_unittest.cc
mhiramat: > the actual trace pipe doesn't behave this way.

Ah, good catch!

OK, I'll rewrite it.

src/base/test/tmp_dir_tree.h
mhiramat: OK, I'll revert it.

mhiramat:
Thanks for review! I updated according to the comments.

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: Thanks, I added the check.

rsavitski (approved):
Looks good to me, thank you for the patience with the slow review.

I'll try to get to the internal .rc patch later today.

mhiramat:
Thanks for review and merging !

PR #1346, opened 2025-04-23
[ui] Fix passing classes to input elements

altimin:
Ensure that the class name is passed properly to a TextInput div
rather than being passed as a string child element.


stevegolton (approved):
Nice catch!

betasheet (approved):

PR #1348, opened 2025-04-24
protovm: slab allocator towards production quality

rukkal:
- Make slabs 4KB-aligned (direct memory page allocation).
- Keep a map of 4KB-aligned blocks to corresponding slab. Used to map an element to be deallocated to the corresponding slab in O(1).
- Keep a list of non-full slabs. Used to access a slab ready for allocations in O(1).
- Deallocate slabs once they become empty.

### DIFF
--- a/src/protovm/slab_allocator.h
+++ b/src/protovm/slab_allocator.h
@@ -17,76 +17,266 @@
 #ifndef SRC_PROTOVM_SLAB_ALLOCATOR_H_
 #define SRC_PROTOVM_SLAB_ALLOCATOR_H_
 
-#include <algorithm>
 #include <cstdlib>
-#include <vector>
+#include <new>
 
-namespace perfetto {
-namespace protovm {
+#include <sys/mman.h>
+
+#include "perfetto/base/logging.h"
+#include "perfetto/ext/base/flat_hash_map.h"
+#include "perfetto/ext/base/paged_memory.h"
+#include "src/base/intrusive_list.h"
 
 // An efficient allocator for elements with fixed size and alignment
 // requirements.
 //
+// Design doc: go/perfetto-protovm-implementation
+//
 // Key features:
 //
-// - Slab allocation: instead of requesting memory for each individual element,
-//   it allocates large chunks of memory (slabs) upfront, where each slab can
-//   hold multiple elements.
+// - Slab allocation: Instead of requesting memory for each individual element,
+//   this allocator pre-allocates large memory chunks (slabs). Each slab is
+//   designed to hold multiple elements.
 //
-// - Free list management: a free list keeps track of available elements within
-//   the allocated slabs. When a request for allocation comes in, the allocator
-//   simply takes an element from the free list. Deallocation returns the
-//   element back to the free list.
-template <size_t ElementSize, size_t ElementAlign, size_t SlabCapacity = 64>
-class SlabAllocator {
+// - Element free list: A free list tracks available elements within each
+//   individual slab, allowing for O(1) access time during allocation.
+//
+// - Slab intrusive lists: Slabs are managed within one of two intrusive lists.
+//   The "non-full slabs" list and the "full slabs" list. This organization
+//   allows "non-full" slabs (those with available space for new allocations) to
+//   be accessed in O(1) time.
+//
+// - Block-to-Slab hash map: A hash map links 4KB-aligned memory blocks to their
+//   corresponding slab. This enables O(1) mapping of an element back to its
+//   slab during deallocation.
+//
+// Allocation process:
+// 1. If there is no free slab
+//    1.1 Allocate a new slab, add it to the “non-full slabs” list, populate the
+//        block-to-slab map
+// 2. Pick any slab from the “non-full slabs” list
+// 3. Allocate the element
+// 4. If needed, move the slab to the “full slabs” list
+//
+// Deallocation process:
+// 1. Find slab from the block-to-slab map
+// 2. Free element
+// 3. If needed, move to the “non-full slabs” list
+// 4. If needed, deallocate the slab.
+
+namespace perfetto {
+namespace protovm {
+
+namespace internal {
+static constexpr size_t k4KiloBytes = static_cast<size_t>(4096);
+}
+
+template <size_t ElementSize, size_t ElementAlign, size_t Blocks4KB>
+class Slab {
  public:
-  explicit SlabAllocator() : next_free_slot_{nullptr} {}
+  struct IntrusiveListTraits {
+    static constexpr size_t node_offset() {
+      return offsetof(Slab, intrusive_list_node_);
+    }
+  };
+
+  Slab() : paged_memory_(base::PagedMemory::Allocate(ElementSize * kCapacity)) {
+    PERFETTO_CHECK(paged_memory_.IsValid());
+
+    // Expect allocated memory to be always 4KB-aligned
+    PERFETTO_CHECK(reinterpret_cast<uintptr_t>(paged_memory_.Get()) %
+                       internal::k4KiloBytes ==
+                   0);
+
+    ConstructSlots();
+    InitializeSlotsFreeList();
+  }
+
+  ~Slab() { DestroySlots(); }
 
   void* Allocate() {
-    if (!next_free_slot_) {
-      auto slab = CreateSlab();
-      if (!slab) {
-        return nullptr;
-      }
-      slabs_.push_back(std::move(slab));
-      next_free_slot_ = &slabs_.back()[0];
-    }
-
+    PERFETTO_DCHECK(next_free_slot_);
     auto* slot = next_free_slot_;
-    next_free_slot_ = slot->next;
-    memset(&slot->element, 0, ElementSize);
-    return &slot->element;
+    next_free_slot_ = next_free_slot_->next;
+    ++size_;
+    return slot;
   }
 
   void Free(void* p) {
     auto* slot = static_cast<Slot*>(p);
+    PERFETTO_DCHECK(slot >= slots() && slot < slots() + kCapacity);
     slot->next = next_free_slot_;
     next_free_slot_ = slot;
+    --size_;
+  }
+
+  bool IsFull() const { return size_ == kCapacity; }
+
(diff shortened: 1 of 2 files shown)


src/protovm/slab_allocator.h
```diff
@@ -19,49 +19,132 @@
...
+//
+// - Block-to-Slab hash map: A hash map links 4KB-aligned memory blocks to their
+// corresponding slab. This enables O(1) mapping of an element back to its slab
+// during deallocation.
+
+namespace perfetto {
+namespace protovm {
+
+namespace internal {
+static constexpr size_t k4KiloBytes = static_cast<size_t>(4) * 1024;
```
primiano: I mean, given this value is never going to ever be anyting != 4096, I'd just use 4096 in the code. This just makes the code harder to read IMHO bringing no benefit.

src/protovm/slab_allocator.h
```diff
@@ -19,49 +19,132 @@
...
-      auto slab = CreateSlab();
-      if (!slab) {
-        return nullptr;
-      }
-      slabs_.push_back(std::move(slab));
-      next_free_slot_ = &slabs_.back()[0];
+  static Slab* New() {
+    size_t rounded_up_size = RoundUpToSysPageSize(sizeof(Slab));
+    PERFETTO_CHECK(rounded_up_size >= sizeof(Slab));
+    void* ptr = mmap(nullptr, rounded_up_size, PROT_READ | PROT_WRITE,
```
primiano: can you use base::PagedMemory? It's already cross-platform, otherwise we have to port this to windows to use it in traced

src/protovm/slab_allocator.h
```diff
@@ -19,49 +19,132 @@
...
+namespace internal {
+static constexpr size_t k4KiloBytes = static_cast<size_t>(4) * 1024;
+}
+
+inline size_t RoundUpToSysPageSize(size_t req_size) {
+  const size_t page_size = base::GetSysPageSize();
+  return (req_size + page_size - 1) & ~(page_size - 1);
+}
+
 template <size_t ElementSize, size_t ElementAlign, size_t SlabCapacity = 64>
```
primiano: Imho the deffault capacity should be way higher. Ideally you want to have >4KB per page. mmap/unmap is slow-ish and causes VM fragmentation, so ideally I'd make it so we end up doing blocks of 64KB or more. less is going to be just a lot of overhead imho

Maybe this SlabCapacity sould be something like round(64*1024/ElementAlign) or similar

primiano (commented):
Some initial comments. I have to go for the day, i'll resume on monday

src/protovm/slab_allocator.h
rukkal: This is a discussion we started offline, but we didn't continue. So the VM has to run cross-platform as well. Fine. I just have two concerns with using base::PagedMemory as it is:
1. it allocates two extra pages around the usable memory that act as canaries/guards. Do we want that overhead?
2. it adds an extra layer of indirection that we don't need. Now SlabAllocator does a single mmap and there we construct our Slab. With base::PagedMemory we'll likely have to do an extra "new". new base::PagedMemory(), which internally does the mmap.

Wdyt? What if we factor out a couple of helper functions (e.g. AllocatedPagedMemory() + DeallocatePagedMemory()) to be reused by both base::PagedMemory and SlabAllocator?

src/protovm/slab_allocator.h
rukkal: done

src/protovm/slab_allocator.h
rukkal: mmh forget about 1). I just realized that it is basically free.

src/protovm/slab_allocator.h
rukkal: Sorry, this `= 64` default was just a leftover and I removed it.

The relevant parameter is the `Blocks4KBPerSlab` in the class below:
```
template <size_t ElementSize, size_t ElementAlign, size_t Blocks4KBPerSlab = 1>
class SlabAllocator {
```

which defaults to 1 and I agree it is too small for most of the use cases. I changed it to 16 for now.

primiano (changes requested):
So I finally had a look at this. 
IMHO it's not worth complicating our life with trying to make Slab* self-allocating, and using intrusive list for the slabs themselves.
Slabs are big and hence infrequent. I don't think an extra malloc-managed heap object per slab.

I'd much rather make Slab a heap object, so then you can use a std::list<Slab> in SlabAllocator and internally you can use PagedMemory and can use move semantics.

The current code is too full of subtleties which IMHO are not really required here.

rukkal:
> So I finally had a look at this. IMHO it's not worth complicating our life with trying to make Slab* self-allocating, and using intrusive list for the slabs themselves. Slabs are big and hence infrequent. I don't think an extra malloc-managed heap object per slab.
> 
> I'd much rather make Slab a heap object, so then you can use a std::list in SlabAllocator and internally you can use PagedMemory and can use move semantics.
> 
> The current code is too full of subtleties which IMHO are not really required here.

As discussed offline, let's:
- change Slab to use base::PagedMemory (fine to have one extra heap allocation per slab)
- keep the base::IntrusiveList<Slab> as it allows moving slabs between lists without allocations. Using std::list would involve one extra heap allocation per element in the worst-case scenario)

rukkal:
> As discussed offline, let's:
> 
> - change Slab to use base::PagedMemory (fine to have one extra heap allocation per slab)
> - keep the base::IntrusiveList as it allows moving slabs between lists without allocations. Using std::list would involve one extra heap allocation per element in the worst-case scenario)

Made this changes

src/protovm/slab_allocator.h
```diff
@@ -17,76 +17,236 @@
...
+  Slot* slots() {
+    return std::launder(reinterpret_cast<Slot*>(paged_memory_.Get()));
+  }
 
-    for (size_t i = 0; i + 1 < SlabCapacity; ++i) {
-      auto& slot = slab[i];
-      auto& next_slot = slab[i + 1];
-      slot.next = &next_slot;
+  Slot* next_free_slot_{nullptr};
+  std::size_t size_{0};
```
primiano: nit: no need for std:: for size_t (we don't do anywheer else in the codebase)

src/protovm/slab_allocator.h
```diff
@@ -17,76 +17,236 @@
...
+
   union Slot {
     Slot* next;
     alignas(ElementAlign) unsigned char element[ElementSize];
   };
 
-  std::unique_ptr<Slot[]> CreateSlab() {
-    auto slab = std::unique_ptr<Slot[]>(new Slot[SlabCapacity]);
+  Slot* slots() {
+    return std::launder(reinterpret_cast<Slot*>(paged_memory_.Get()));
```
primiano, line 171: mostly curious: what' the std::launder for? I try reading cppreference but is full of c++ legalese

src/protovm/slab_allocator.h
```diff
@@ -17,76 +17,236 @@
...
+    DeleteSlabs(slabs_full_);
+  }
+
+  void* Allocate() {
+    // Create new slab if needed
+    if (slabs_non_full_.empty()) {
+      auto* slab = new SlabType();
+      if (!slab) {
+        return nullptr;
+      }
```
primiano: you can't really fail the new. The thing that might actually fail is the mmap inside (but PagedMeemory aborts unless you pass kMayFail)

src/protovm/slab_allocator.h
```diff
@@ -17,76 +17,236 @@
...
+    auto ptr_4KB_aligned = reinterpret_cast<uintptr_t>(ptr) &
+                           ~(static_cast<uintptr_t>(internal::k4KiloBytes) - 1);
+    SlabType** slab = block_4KB_aligned_to_slab_.Find(ptr_4KB_aligned);
+    PERFETTO_CHECK(slab);
+    PERFETTO_CHECK(*slab);
+    return **slab;
+  }
+
+  void DeleteSlabs(
+      base::IntrusiveList<SlabType, typename SlabType::IntrusiveListTraits>&
```
primiano: make this a typedef (using SlabList = base::IntrusiveList....)

primiano (approved):
It looks good, only few comments.

As a general top level comment, can you leave more code comments around?
For instnace, the hashtable. I know what that is because I suggested it... but other people will find it a bit mysterious so a comment there woudl be nice

src/protovm/slab_allocator.h
rukkal: afiu all that legalese means "hey compiler, forget what you thought you knew about this address (e.g. other objects living there), this is **the object** living at this address now and don't perform any optimizations based on previous knowledge"

So here I am telling the compiler "forget this was a raw piece of memory, now it is a Slot object". Tbh in this case I am not sure we could ever run into UB without the std::launder, but it is a safe choice. I think std::launder makes total sense in code like StatusOr (basically a variant) where on the same piece of memory we could create different objects.

rukkal: btw speaking of new objects I realize that with the latest refactoring we lost the actual construction/destruction of Slot objects :upside_down_face:  adding them back now

src/protovm/slab_allocator.h
rukkal: But the new also allocates+creates a Slab object on the heap... and then the paged memory is mmap'ed

So the new could fail. However, I initially assumed it would return nullptr, given that we build with -fno-exceptions (right?). But now I am thinking more about it and it should abort instead, because we are not passing a std::nothrow_t tag. I'll remove the if(!slab)

src/protovm/slab_allocator.h
rukkal: done

src/protovm/slab_allocator.h
rukkal: done

rukkal:
> It looks good, only few comments.
> 
> As a general top level comment, can you leave more code comments around? For instnace, the hashtable. I know what that is because I suggested it... but other people will find it a bit mysterious so a comment there woudl be nice

There is a "top-level" comment (line 30) that provides an overview of the allocator (including why the hashtable). Now I just extended it with two sections "allocation process" + "deallocation process".

PR #1353, opened 2025-04-25
base: add IntrusiveList

rukkal:


src/base/intrusive_list.h
```diff
@@ -0,0 +1,151 @@
...
+
+namespace internal::intrusive_list {
+
+struct Node {
+  Node* prev;
+  Node* next;
+};
+
+// Base class to factor out type-independent code (avoid binary bloat)
+class Base {
```
primiano: ListOps? calling a class Base always makes me feel asking for toubles :) (thinking about random C headers out there, like windows.h, which might be doing `#define base ...`)

src/base/intrusive_list.h
```diff
@@ -0,0 +1,151 @@
...
+// Unlike std::list<>, the entries being inserted into the list need to
+// explicitly declare an IntrusiveListNode structure (one for each list they are
+// part of). The user must specify a Traits struct for each list the entry is
+// part of. The traits struct defines how to get to the IntrusiveListNode from
+// the outer object.
+//
+// Usage example:
+// class Person {
+//  public:
+//   struct Traits {
```
primiano: I love the design, just can it be ListTraits rather than Traits?

src/base/intrusive_list.h
```diff
@@ -0,0 +1,151 @@
...
+//     static constexpr size_t NodeOffset() { return offsetof(Person, node); }
```
primiano: node_offset() to make it clear is a constexpr/cheap thing

src/base/intrusive_list.h
```diff
@@ -0,0 +1,151 @@
...
+// }
+//
+// IntrusiveList<Person, Person::Traits> list;
+// Person person;
+// list.PushBack(person);
+// ...
+
+namespace perfetto::base {
+
+namespace internal::intrusive_list {
```
primiano: very small bikeshedding, I would have probably just done

namespace internal {
 struct ListNode...
 struct ListOps...
}

simply because there is no matchin non-internal base::intrusive_list namespace (and it's good that way)

src/base/intrusive_list.cc
```diff
@@ -0,0 +1,63 @@
...
+
+void Base::PopFront() {
+  PERFETTO_DCHECK(front_);
+  front_ = front_->next;
+
+  if (front_) {
+    front_->prev = nullptr;
+  }
+
+  --size_;
```
primiano, line 44: add a DCHECK(size_> 0) before the --?

same below

src/base/intrusive_list.h
```diff
@@ -0,0 +1,151 @@
...
+  using value_type = T;
+  using const_pointer = const T*;
+
+  void PushFront(T& entry) {
+    internal::intrusive_list::Base::PushFront(nodeof(&entry));
+  }
+
+  void PopFront() { internal::intrusive_list::Base::PopFront(); }
+
+  T& Front() {
```
primiano: nit: lowercase  front(), as it's trivial

primiano (approved):
Looks great... and very well timed. I might use it very soon next week ;)

LGTM with few minor comments

src/base/intrusive_list.h
rukkal: done

src/base/intrusive_list.h
rukkal: done

src/base/intrusive_list.cc
rukkal: done

src/base/intrusive_list.h
rukkal: done.  For empty() and size() as well

src/base/intrusive_list.h
rukkal: done

src/base/intrusive_list.h
rukkal: > thinking about random C headers out there, like windows.h, which might be doing #define base ...

omg :sweat_smile: 

changed as suggested below:

namespace internal {
struct ListNode...
struct ListOps...
}

PR #1357, opened 2025-04-25
Move timestamp and duration format settings to new settings subsystem

stevegolton:
This change migrates the timestamp and duration format settings, which are currently handled and persisted separately, to the new settings subsystem.

These settings are slightly different to most, as they need to be made available to the entire UI. While this could be done by fetching them via their key from the settings subsystem, this would not be type safe. Instead, the settings are exposed as part of the `Trace` interface (under the `Timeline` section) and the settings are simply registered as part of the trace object initialization.

No attempt has been made to migrate the settings from the previous storage mechanism so users will need to reconfigure these settings when migrating.


LalitMaganti:
> No attempt has been made to migrate the settings from the previous storage mechanism so users will need to reconfigure these settings when migrating.

This feels somewhat sad but fair if it would be a bunch of fragile code.

LalitMaganti (approved):

PR #1360, opened 2025-04-25
Support reading files using symlinks

aMayzner:
Support files with symlinks. The change was suggested in http://yaqs/2182006114088910848


LalitMaganti:
> the ideal solution would be to use something like fts_read, at least on linux, so that:

I agree this is an ideal solution but given that that code is already quite tricky to reason about with the presence of ifdefs for NACL and Windows, I think complicating the code even more is not a great idea. Looking at the man page (https://man7.org/linux/man-pages/man3/fts.3.html) this looks a very complex API and getting all the details right here will take quite some time.

Also this is actively blocking folks (http://yaqs/2182006114088910848#a1n1) so I'm inclined to merge this PR. This code is currently only used for trace processor so if we think this is not high quality enough to have in base, happy to move it to trace_processor instead.

Alternatively, if we think the priority is high enough, let me know and I'll rewrite this with fts. But I want to merge this to not block people.

LalitMaganti (approved):

LalitMaganti:
Fix courtesy of afontan@ internally. Just copied and posted on behalf of them.

PR #1365, opened 2025-04-25
Remove `python_version` from python/BUILD file

aMayzner:
The copybara worker is blocked on wrong formatting of the python/BUILD file: we can't mention `python_version` anymore and the LSC is removing it. 

Editing gen_bazel to stop adding it to autogenerated python BUILD file.

tools/install_build_deps install bazel with version 7.4. I built it locally and it seemed to build fine


tools/gen_bazel
```diff
@@ -309,12 +309,10 @@ class PythonBuildGenerator:
     label.comment = target.name
     label.srcs += (self.python_label_to_path(x) for x in target.sources)
     label.data += (self.python_data_to_path(x) for x in target.data)
-    label.main = target.python_main
-    label.python_version = 'PY3'
```
LalitMaganti, line 313: Please remove handling of python_version everywhere else in this file, it should no longer be necessary.

LalitMaganti (changes requested):

tools/gen_bazel
aMayzner: Done

LalitMaganti (approved):

rsavitski (approved):

PR #1367, opened 2025-04-27
Update index.ts

xiangshuangxi:
Complete perfetto internal systemui name

Because the underlying AndroidRuntime::setArgv0 will truncate thread names longer than 15 to keep 15, so this is probably intentional. Because the underlying setArgv0 is consistent with the kernel thread name length, this modification is to keep the first letter capitalized with other variables in TRACKS_TO_PIN, and to make the code clearer after completion, and easier for users to understand.

Signed-off-by:
Shuangxi Xiang <xiangshuangxi@xiaomi.corp-partner.google.com>

### DIFF
--- a/test/trace_processor/diff_tests/metrics/graphics/composer_execution.py
+++ b/test/trace_processor/diff_tests/metrics/graphics/composer_execution.py
@@ -30,7 +30,7 @@ trace.add_thread(15000, 10335, "worker thread")
 trace.add_ftrace_packet(1)
 
 # unskipped validation
-trace.add_atrace_begin(ts=100, tid=10335, pid=10335, buf="onMessageRefresh")
+trace.add_atrace_begin(ts=100, tid=10335, pid=10335, buf="CompositionEngine::present")
 trace.add_atrace_begin(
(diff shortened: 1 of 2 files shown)


xiangshuangxi (commented):
Please help to review this modification, thanks.

primiano:
@nicomazz do you have any opinion on this?

@xiangshuangxi what I don't understand is: if you add the 'A' there, it is not going to match anymore the track title.
So it will be trying to pin a track that does not exist.

I cannot understand your explanation honestly

xiangshuangxi:
@primiano 
Hi, 
Because the kernel intercepts and retains the thread name from the back to the front according to the kernel thread name length requirement (a maximum of 15 are retained), such as the thread name of com.android.systemui on systrace is ndroid.systemui. Therefore, adding an 'A' here is just a small alignment optimization on the code, and will not affect the thread name (ndroid.systemui) of com.android.systemui on systrace.

xiangshuangxi:
Please help review whether this modification is reasonable and provide information.
208a6d6bebafb80b8d5f30f8de1dc338ead770a6

primiano:
These changes seem completely random.

PR #1373, opened 2025-04-28
tracing: Move buffer_exhausted_policy to instance state

ddiproietto:
A future commit will allow configuring the buffer_exhausted_policy from the config. This is a prerequisite for that.

### DIFF
--- a/include/perfetto/tracing/internal/tracing_muxer.h
+++ b/include/perfetto/tracing/internal/tracing_muxer.h
@@ -38,6 +38,8 @@ class TracingSession;
 namespace internal {
 
 struct DataSourceParams {
+  BufferExhaustedPolicy default_buffer_exhausted_policy =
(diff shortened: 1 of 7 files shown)


include/perfetto/tracing/internal/tracing_muxer.h
```diff
@@ -38,6 +38,8 @@ class TracingSession;
 namespace internal {
 
 struct DataSourceParams {
+  BufferExhaustedPolicy default_buffer_exhausted_policy =
```
primiano, line 44: Can you add a comment saying "this is the policy that is defined in the C++ code. This will become DataSourceState.buffer_exhausted_policy, unless the data source says that the policy is configurable (e.g. see XXX api you are introducing) and the TraceConfig.YYY changes it"

ddiproietto: Done, thanks

PR #1376, opened 2025-04-28
Add machine id column to AndroidLog panel

jahdiel-alvarez:
In multi-VM traces, it is valuable to differentiate from which machine each Android log is coming from. For this we added a machine id column to the table in the AndroidLog panel. This columns is conditional to there being more than one machine, if not then the log panel behaves the same way as before.

Test: ui/run-unittests
Bug: 406061688

### DIFF
--- a/ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
@@ -51,6 +51,7 @@ interface Pagination {
 
 interface LogEntries {
   offset: number;
+  machineIds: number[];
   timestamps: time[];
   priorities: number[];
   tags: string[];
@@ -86,6 +87,8 @@ export class LogPanel implements m.ClassComponent<LogPanelAttrs> {
       this.reloadData(attrs);
     }
 
+    const hasMachineIds =
+      this.entries && this.entries.machineIds.filter((id) => id).length > 0;
     const hasProcessNames =
       this.entries &&
       this.entries.processName.filter((name) => name).length > 0;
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -171,6 +179,7 @@ export class LogPanel implements m.ClassComponent<LogPanelAttrs> {
           this.entries.isHighlighted[i] && 'pf-highlighted',
         ),
         cells: [
+          ...(hasMachineIds ? [machineIds[i] || ''] : []),
```
stevegolton: Is it intended to show nothing if machine id is 0? I think if we have machine ids then it would make sense to show 0 here?

jahdiel-alvarez: So by convention the machine id of the primary machine is NULL. I set it to zero to be consistent with the type of `machineIds`. In most cases in the UI if the data is coming from the primary machine, we simply don't reference the id. So to be consistent with that style, I am not showing "fake" id of the primary machine.

stevegolton: Ah I see. In that case, maybe change the type of `machineIds` to be `(number | null)[]` and just pass the null value through to the logs panel. This will avoid running into the issue where  if 0 was ever actually pushed through here for whatever reason then it would actually show up as 0 on the display. This is mainly to avoid falling into the JS falsy trap where 0 and null are both falsy.

In terms of the display of the null `machineId` value, just leaving it blank looks like the information is missing, but IIUC we know that 0 (or null if you apply the above) means the primary machine - so maybe try to convey that information somehow in the cell? I still feel like '0' would logically make sense, but perhaps I'm being naive here. Even just printing 'NULL' would be better than empty? WDYT?

stevegolton (approved):

PR #1382, opened 2025-04-29
Update SysUI persistent tracing flag name in comments

PabiGamito:
The flag `persist.debug.perfetto.persistent` no longer exists and we now have `persist.debug.perfetto.persistent_sysui_tracing_for_bugreport` instead to be more descriptive. This PR updates the comments that mention this flag which is no longer used.


ddiproietto (approved):
Thanks a lot!

I think the flag never existed. It was just renamed before code review: https://android-review.git.corp.google.com/c/platform/external/perfetto/+/2967622/2..11/perfetto.rc#b162

PR #1385, opened 2025-04-29
[chrome] Add scenario description string

etiennep-chromium:
This PR adds a description string field to ScenarioConfig, which will be displayed in Chrome internal UI.

### DIFF
--- a/protos/perfetto/config/chrome/scenario_config.proto
+++ b/protos/perfetto/config/chrome/scenario_config.proto
@@ -97,6 +97,8 @@ message ScenarioConfig {
   // Scenario name, unique within the whole field tracing config.
   optional string scenario_name = 1;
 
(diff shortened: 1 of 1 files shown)


protos/perfetto/config/chrome/scenario_config.proto
```diff
@@ -97,6 +97,8 @@ message ScenarioConfig {
   // Scenario name, unique within the whole field tracing config.
   optional string scenario_name = 1;
 
+  optional string scenario_description = 9;
```
khokhlov962, line 101: nit: add a field description; maybe also mention how it is going to be used.

khokhlov962:
LGTM from my side, but you'll need a review from someone from Perfetto team.

betasheet (approved):

PR #1389, opened 2025-04-30
Add machine id filter to AndroidLog

jahdiel-alvarez:
In multi-machine traces doing log analysis might be complicated given there could be thousands of logs from each machine in the same log table. To improve the log navigation, we added a multi-select filter which allows the user to select any subset of machines to only view their respective logs.

Test: ui/run-unittests
Bug: 406061688

### DIFF
--- a/ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
@@ -22,6 +22,12 @@ import {Monitor} from '../../base/monitor';
 import {AsyncLimiter} from '../../base/async_limiter';
 import {escapeGlob, escapeQuery} from '../../trace_processor/query_utils';
 import {Select} from '../../widgets/select';
+import {
+  MultiSelectDiff,
+  MultiSelectOption,
+  PopupMultiSelect,
+} from '../../widgets/multiselect';
+import {PopupPosition} from '../../widgets/popup';
 import {Button} from '../../widgets/button';
 import {TextInput} from '../../widgets/text_input';
 import {VirtualTable, VirtualTableRow} from '../../widgets/virtual_table';
@@ -37,9 +43,15 @@ export interface LogFilteringCriteria {
   tags: string[];
   textEntry: string;
   hideNonMatching: boolean;
+  machineExcludeList: number[];
+}
+
+export interface LogPanelCache {
+  uMachineIds: number[];
 }
 
 export interface LogPanelAttrs {
+  cache: LogPanelCache;
   filterStore: Store<LogFilteringCriteria>;
   trace: Trace;
 }
@@ -61,6 +73,20 @@ interface LogEntries {
   totalEvents: number; // Count of the total number of events within this window
 }
 
+async function getMachineIds(engine: Engine): Promise<number[]> {
+  // A machine might not provide Android logs, even if configured to do so.
+  // Hence, the |cpu| table might have ids not present in the logs. Given this
+  // is highly unlikely and going through all logs is expensive, we will get
+  // the ids from |cpu|, even if filter shows ids not present in logs.
+  const result = await engine.query(`SELECT DISTINCT(machine_id) FROM cpu`);
+  const machineIds: number[] = [];
+  const it = result.iter({machine_id: NUM_NULL});
+  for (let row = 0; it.valid(); it.next(), row++) {
+    machineIds.push(it.machine_id == null ? 0 : it.machine_id);
+  }
+  return machineIds;
+}
+
 export class LogPanel implements m.ClassComponent<LogPanelAttrs> {
   private entries?: LogEntries;
 
@@ -80,6 +106,10 @@ export class LogPanel implements m.ClassComponent<LogPanelAttrs> {
     ]);
 
     this.filterMonitor = new Monitor([() => attrs.filterStore.state]);
+
+    getMachineIds(attrs.trace.engine).then((uMachineIds) => {
+      attrs.cache.uMachineIds = uMachineIds;
+    });
   }
 
   view({attrs}: m.CVnode<LogPanelAttrs>) {
@@ -99,7 +129,11 @@ export class LogPanel implements m.ClassComponent<LogPanelAttrs> {
       {
         title: 'Android Logs',
         description: `Total messages: ${totalEvents}`,
-        buttons: m(LogsFilters, {trace: attrs.trace, store: attrs.filterStore}),
+        buttons: m(LogsFilters, {
+          trace: attrs.trace,
+          cache: attrs.cache,
+          store: attrs.filterStore,
+        }),
       },
       m(VirtualTable, {
         className: 'pf-android-logs-table',
@@ -277,6 +311,7 @@ class FilterByTextWidget implements m.ClassComponent<FilterByTextWidgetAttrs> {
 
 interface LogsFiltersAttrs {
   readonly trace: Trace;
+  readonly cache: LogPanelCache;
   readonly store: Store<LogFilteringCriteria>;
 }
 
@@ -325,7 +360,44 @@ export class LogsFilters implements m.ClassComponent<LogsFiltersAttrs> {
         },
         disabled: attrs.store.state.textEntry === '',
       }),
+      this.renderFilterPanel(attrs),
     ];
+  }
+
+  private renderFilterPanel(attrs: LogsFiltersAttrs) {
+    const machineExcludeList = attrs.store.state.machineExcludeList;
+    const options: MultiSelectOption[] = attrs.cache.uMachineIds.map(
+      (uMachineId) => {
+        return {
+          id: String(uMachineId),
+          name: `Machine ${uMachineId}`,
+          checked: !machineExcludeList.some(
+            (excluded: number) => excluded === uMachineId,
+          ),
+        };
+      },
+    );
+
+    return m(PopupMultiSelect, {
+      label: 'Filter',
+      icon: 'filter_list_alt',
+      popupPosition: PopupPosition.Top,
+      options,
+      onChange: (diffs: MultiSelectDiff[]) => {
+        const newList = new Set<number>(machineExcludeList);
+        diffs.forEach(({checked, id}) => {
+          const machineId = Number(id);
+          if (checked) {
+            newList.delete(machineId);
+          } else {
+            newList.add(machineId);
+          }
+        });
+        attrs.store.edit((draft) => {
+          draft.machineExcludeList = Array.from(newList);
+        });
+      },
+    });
   }
 }
 
@@ -414,6 +486,9 @@ async function updateLogView(engine: Engine, filter: LogFilteringCriteria) {
       where prio >= ${filter.minimumLevel}`;
   if (filter.tags.length) {
     selectedRows += ` and tag in (${serializeTags(filter.tags)})`;
+  }
+  if (filter.machineExcludeList.length) {
+    selectedRows += ` and ifnull(process.machine_id, 0) not in (${filter.machineExcludeList.join()})`;
   }
 
   // We extract only the rows which will be visible.
--- a/ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
@@ -47,7 +47,7 @@ export interface LogFilteringCriteria {
 }
 
 export interface LogPanelCache {
-  uMachineIds: number[];
+  uniqueMachineIds: number[] | null;
 }
 
 export interface LogPanelAttrs {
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -61,6 +73,20 @@ interface LogEntries {
...
+async function getMachineIds(engine: Engine): Promise<number[]> {
+  // A machine might not provide Android logs, even if configured to do so.
+  // Hence, the |cpu| table might have ids not present in the logs. Given this
+  // is highly unlikely and going through all logs is expensive, we will get
+  // the ids from |cpu|, even if filter shows ids not present in logs.
+  const result = await engine.query(`SELECT DISTINCT(machine_id) FROM cpu`);
+  const machineIds: number[] = [];
+  const it = result.iter({machine_id: NUM_NULL});
+  for (let row = 0; it.valid(); it.next(), row++) {
+    machineIds.push(it.machine_id == null ? 0 : it.machine_id);
```
stevegolton: I'd suggest `it.machine_id ?? 0`.

Generally `==` should be avoided in our codebase.

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -61,6 +73,20 @@ interface LogEntries {
...
 
+async function getMachineIds(engine: Engine): Promise<number[]> {
+  // A machine might not provide Android logs, even if configured to do so.
+  // Hence, the |cpu| table might have ids not present in the logs. Given this
+  // is highly unlikely and going through all logs is expensive, we will get
+  // the ids from |cpu|, even if filter shows ids not present in logs.
+  const result = await engine.query(`SELECT DISTINCT(machine_id) FROM cpu`);
+  const machineIds: number[] = [];
+  const it = result.iter({machine_id: NUM_NULL});
+  for (let row = 0; it.valid(); it.next(), row++) {
```
stevegolton: What's `row` for? Looks like it's not used.

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -61,6 +73,20 @@ interface LogEntries {
   totalEvents: number; // Count of the total number of events within this window
 }
 
+async function getMachineIds(engine: Engine): Promise<number[]> {
+  // A machine might not provide Android logs, even if configured to do so.
+  // Hence, the |cpu| table might have ids not present in the logs. Given this
+  // is highly unlikely and going through all logs is expensive, we will get
+  // the ids from |cpu|, even if filter shows ids not present in logs.
+  const result = await engine.query(`SELECT DISTINCT(machine_id) FROM cpu`);
```
stevegolton: Maybe sort by machine_id?

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -37,9 +43,15 @@ export interface LogFilteringCriteria {
   tags: string[];
   textEntry: string;
   hideNonMatching: boolean;
+  machineExcludeList: number[];
+}
+
+export interface LogPanelCache {
+  uMachineIds: number[];
```
stevegolton: Please avoid hungartion notation!

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -80,15 +106,18 @@ export class LogPanel implements m.ClassComponent<LogPanelAttrs> {
     ]);
 
     this.filterMonitor = new Monitor([() => attrs.filterStore.state]);
+
+    getMachineIds(attrs.trace.engine).then((uMachineIds) => {
```
stevegolton: Maybe only run this if the cache is not already populated? Otherwise there's not much point in having the cache

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -325,8 +361,45 @@ export class LogsFilters implements m.ClassComponent<LogsFiltersAttrs> {
         },
         disabled: attrs.store.state.textEntry === '',
       }),
+      ...(hasMachineIds ? [this.renderFilterPanel(attrs)] : []),
```
stevegolton: `hasMachineIds && this.renderFilterPanel(attrs)` works just fine here - mithril omits falsy children for this exact reason.

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -415,6 +488,9 @@ async function updateLogView(engine: Engine, filter: LogFilteringCriteria) {
   if (filter.tags.length) {
     selectedRows += ` and tag in (${serializeTags(filter.tags)})`;
   }
+  if (filter.machineExcludeList.length) {
+    selectedRows += ` and ifnull(process.machine_id, 0) not in (${filter.machineExcludeList.join()})`;
```
stevegolton: I think you want to do `join(', ');`

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -325,8 +361,45 @@ export class LogsFilters implements m.ClassComponent<LogsFiltersAttrs> {
...
+          name: `Machine ${uMachineId}`,
+          checked: !machineExcludeList.some(
+            (excluded: number) => excluded === uMachineId,
+          ),
+        };
+      },
+    );
+
+    return m(PopupMultiSelect, {
+      label: 'Filter',
```
stevegolton: Maybe a more descriptive label like `filter by machine` (or similar)

stevegolton (changes requested):
Overall looks good % comments

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
jahdiel-alvarez: So the `u` stands for `unique`. I added it because we have the `machineIds` field in LogEntries, which contains the machine ids per log. So to differentiate the two concepts I added the `u`, wasn't trying to use Hungarian notation.

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
jahdiel-alvarez: So this is happening in the constructor so only happens ones. So effectively it caches against each rerender. I guess we could add the check in case another LogPanel is added in the future in index.ts, although that seem unlikely. Adding the check seems a bit unnecessary. WDYT?

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
jahdiel-alvarez: In join() if the separator is omitted it uses "," which would still produce a valid SQL.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#separator

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
jahdiel-alvarez: Seems a bit long but I guess it works

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
stevegolton: Ah, looks a lot like hungarian notation for 'unsigned'. Perhaps go for the more explicit `uniqueMachineIds` in that case?

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
stevegolton: In general I would expect mithril components like this to be somewhat ephemeral, getting created and destroyed multiple times throughout the lifetime of the UI.

In this instance, I believe this component is tied to the lifetime of the tab. If the user closes and reopens the tab the LogPanel component will be constructed multiple times. It would make sense to only load the machine ids once on trace load as it's never going to change.

I'm a bit confused as it seems you've gone to the effort of injecting the cache for this exact reason, but you're not really making use of it. Is there another reason for injecting this cache in from the outside?

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
stevegolton: Wow, well I learned something new today!

I would prefer a more explicit `join(',')` or `join(', ')`, but I'm not going to let that block the PR.

stevegolton (changes requested):

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
jahdiel-alvarez: Sure

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
jahdiel-alvarez: Sure, just want to note that we are already using `.join()` a couple of lines below: https://github.com/google/perfetto/blob/main/ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts#L426

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
jahdiel-alvarez: Yes, I've considered this. In `dev.perfetto.AndroidLog/index.ts` we have:
```ts
ctx.tabs.registerTab({
      isEphemeral: false,
      uri: androidLogsTabUri,
      content: {
        render: () => m(LogPanel, {filterStore, cache, trace: ctx}),
        getTitle: () => 'Android Logs',
      },
    });
```
Given `isEphemeral` is set to `false`, my understanding is that LogPanel object would be persisted even if hidden (maybe I am wrong?), therefore I assumed that the constructor of LogPanel would only be executed once created. But regardless to be on the safe side, we should probably guard the call and be explicit in not updating the cache unless necessary.

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -47,7 +47,7 @@ export interface LogFilteringCriteria {
 }
 
 export interface LogPanelCache {
-  uMachineIds: number[];
+  uniqueMachineIds: number[] | null;
```
stevegolton: Nit: By convention we usually prefer to use `undefined` instead of `null` for cases like this. You can use the ? operator for this:

```ts
  uniqueMachineIds?: number[];
```

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
```diff
@@ -109,17 +109,19 @@ export class LogPanel implements m.ClassComponent<LogPanelAttrs> {
 
     this.filterMonitor = new Monitor([() => attrs.filterStore.state]);
 
-    getMachineIds(attrs.trace.engine).then((uMachineIds) => {
-      attrs.cache.uMachineIds = uMachineIds;
-    });
+    if (attrs.cache.uniqueMachineIds === null) {
```
stevegolton: If this constructor is called multiple times in quick succession, we could trigger multiple loads.

I realise this isn't the biggest deal as getMachineIds is a very cheap operation seeing as the `cpu` table is small, but I don't think it hurts to be thorough.

If you see https://github.com/google/perfetto/blob/main/ui/src/plugins/dev.perfetto.Ftrace/ftrace_explorer.ts#L113, which does something similar for the ftrace events, we set the cache state to loading when loading to avoid this issue.

TBH, and after thinking a bit about the ftrace one, I'm not quite sure this complexity is entirely worth the effort. It might be better just to load the machine ides in the plugin's onTraceLoad() function and inject them in pre-filled via the attrs.

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
stevegolton: I mean all of the other controls 'filter' the tab in one way or another, so just calling this 'filter' is too generic.

ui/src/plugins/dev.perfetto.AndroidLog/logs_panel.ts
stevegolton: `isEphemeral: false` just means the tab sticks around in the triple dot context menu even after the tab is closed.

We don't really have direct control over the lifetime of components. Mithril destroys the component the moment we stop rendering it, and re-creates it the moment we start rendering it again. Tab content isn't rendered when the tab is closed so the component will be destroyed when the tab is closed, and recreated when the tab is opened again.

stevegolton (changes requested):

stevegolton (approved):

PR #1397, opened 2025-04-30
Add pluginable overlay subsystem & use for waker lines for cpu tracks

stevegolton:
This change adds the following:
- A new overlay rendering subsystem which allows plugisn to render an overlay on top of the timeline (on top of tracks other tracks).
- The CpuSlices plugin uses this new overlay rendering mechanism to to render the waker the waker information over the top of the timeline when CPU slices are selected.
  - This moves a lot of the custom waker information from the core into the plugin, tidying up the core a lot.

### DIFF
--- a/ui/src/public/track.ts
+++ b/ui/src/public/track.ts
@@ -56,6 +56,15 @@ export interface TrackManager {
    * power to the user compared to e.g. purely filtering by name.
    */
   registerTrackFilterCriteria(filter: TrackFilterCriteria): void;
+
+  /**
+   * Register a timeline overlay renderer.
+   *
+   * Overlays are rendered on top of all tracks in the timeline view and can be
+   * used to draw annotations that span multiple tracks, such as flow arrows or
+   * vertical lines marking specific events.
+   */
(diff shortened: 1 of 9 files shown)


aMayzner (approved):

ui/src/public/track.ts
```diff
@@ -56,6 +56,15 @@ export interface TrackManager {
...
   registerTrackFilterCriteria(filter: TrackFilterCriteria): void;
+
+  /**
+   * Register a timeline overlay renderer.
+   *
+   * Overlays are rendered on top of all tracks in the timeline view and can be
+   * used to draw annotations that span multiple tracks, such as flow arrows or
+   * vertical lines marking specific events.
+   */
+  registerTimelineOverlay(overlay: TimelineOverlay): void;
```
primiano: discussed offline: maybe just registerOverlay. Imho TimelineOverline makes me thing too much of "you can only overlay on the topi bit" ratehr than the whole canvas.

primiano (approved):

ui/src/public/track.ts
stevegolton: Agreed & done

PR #1412, opened 2025-05-02
Reserve 2501 to 3000 for Android XR metrics

zhul8:


LalitMaganti (approved):

aMayzner (approved):
Can you please explain why are you adding this? We are migrating from metrics v1, so you should try to work with metrics v2

LalitMaganti:
Spoke to these folks offline, they have a lot of metrics which are maintained by multiple teams in closed source and it would be a significant challenge to migrate. This is a reasonable change considering no new code is being added to our codebase, it's maintained externally.

aMayzner (approved):

PR #1414, opened 2025-05-05
Remove depth check for filtering frame slices.

devianb:
Current depth check limits filtering Choreographer#doFrame slices to be at depth 0 and 1. There are cases where the frame slice can be at level 2(eg. b/413987043). The depth check was kept earlier to eliminate resynced slices. This check is not required as the current table logic eliminates these slices with the integer cast.

Bug: 413987043
Test: tools/diff_test_trace_processor.py --name-filter ".*frames.*" out/linux_clang_debug/trace_processor_shell

Welcome to Perfetto!

For more details please see

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
@@ -45,16 +45,16 @@ WITH
       upid,
       ts
     FROM thread_slice
-    -- Mostly the frame slice is at depth 0. Though it could be pushed to depth 1 while users
-    -- enable low layer trace e.g. atrace_app.
     WHERE
-      name GLOB $glob_str AND depth IN (0, 1)
+      name GLOB $glob_str
   )
 SELECT
   *
 FROM all_found
--- Casting string to int returns 0 if the string can't be cast.
--- frame_id is -1 indicating an invalid vsync id
+-- Casting string to int returns 0 if the string can't be cast. This eliminates the Choreographer resynced slices
+-- with the format "Choreographer#doFrame - resynced to 1234 in 20.0ms". The stdlib table is to only list the top
+-- level Choreographer#doFrame slices, hence the 'resynced' slices are ignored.
+-- frame_id is -1 indicating an invalid vsync id.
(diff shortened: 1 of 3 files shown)


src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
```diff
@@ -45,16 +45,16 @@ WITH
       upid,
       ts
     FROM thread_slice
-    -- Mostly the frame slice is at depth 0. Though it could be pushed to depth 1 while users
-    -- enable low layer trace e.g. atrace_app.
+    -- Mostly the frame slice is at depth 0. Though it could be pushed to lower depths like 1 or 2
```
nicomazz: do we still need this comment? it seems it refers to the "depth in (0, 1)" part in the previous code, but now that is not there it's not clear what it refers to

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
```diff
@@ -45,16 +45,16 @@ WITH
...
     WHERE
-      name GLOB $glob_str AND depth IN (0, 1)
+      name GLOB $glob_str
   )
 SELECT
   *
 FROM all_found
--- Casting string to int returns 0 if the string can't be cast.
--- frame_id is -1 indicating an invalid vsync id
+-- Casting string to int returns 0 if the string can't be cast. This eliminates the Choreographer resynced slices.
```
nicomazz: can we briefly add why we want to aliminate the recynced slices and how they look like? (as without looking at a trace it might be difficult to understand)

LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
devianb: Removed.

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
devianb: Added more details.

PR #1419, opened 2025-05-06
Replace the deprecated onMessageRefresh with present

xiangshuangxi:
Remove the abandoned onMessageRefresh in old Android versions

In Android T and higher, the atrace event of SurfaceFlinger::onMessageRefresh() is indeed deprecated, so the logic of these test codes needs to be adjusted. If the test is only for Android T+, it is recommended to directly delete all onMessageRefresh related Trace events, because the function has been removed from the code base and its Trace events are no longer generated. In addition, the function of onMessageRefresh has been refactored and optimized in Android T, and its core logic has been integrated into the new process of CompositionEngine. It is recommended to replace onMessageRefresh here with the Trace event of the new function (such as CompositionEngine::present()).

Welcome to Perfetto!

For more details please see

### DIFF
--- a/test/trace_processor/diff_tests/metrics/graphics/composer_execution.py
+++ b/test/trace_processor/diff_tests/metrics/graphics/composer_execution.py
@@ -30,7 +30,7 @@ trace.add_thread(15000, 10335, "worker thread")
 trace.add_ftrace_packet(1)
 
 # unskipped validation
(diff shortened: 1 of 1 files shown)


xiangshuangxi:
I would like invite you to review this change, thank you!

aMayzner (approved):

PR #1422, opened 2025-05-06
Add support for dimension spec in metrics v2

elrakadm:
Add support for dimension spec in metrics v2

* Explicitly specify the expected type of the dimension in the metrics v2 output
* Accept null for all dimension types
* Continue supporting dimensions without explicit types being specified 
*  Read columns based on spec dimension/metric column names not based on order to 
    * Fix bug on wrong output produced in the case different dimension order between the query and the spec
     * Fix bug on wrong metric value emitted, incase of the metric column is not the last
* Add test for simple_slices in metrics v2

Bug: 415966029
Test: `tools/ninja -C out/linux_clang_debug  && tools/diff_test_trace_processor.py  out/linux_clang_debug/trace_processor_shell --name-filter=".*SummaryMetricsV2.*"`

### DIFF
--- a/protos/perfetto/trace_summary/v2_metric.proto
+++ b/protos/perfetto/trace_summary/v2_metric.proto
@@ -38,10 +38,7 @@ import "protos/perfetto/perfetto_sql/structured_query.proto";
 //
 // ```
 // id: "memory_per_process"
-// dimensions_specs: {
-//   name: "process_name"
-//   type: STRING
-// }
+// dimensions: "process_name"
 // value: "avg_rss_and_swap"
 // query: {
 //   table: {
@@ -69,14 +66,8 @@ import "protos/perfetto/perfetto_sql/structured_query.proto";
 //
 // ```
 // id: "memory_per_process_and_cuj"
-// dimensions_specs: {
-//  name: "process_name"
-//  type: STRING
-//}
-// dimensions_specs: {
-//  name: "cuj_name"
-//  type: STRING
-//}//
+// dimensions: "process_name"
+// dimensions: "cuj_name"
 // value: "avg_rss_and_swap"
 // query: {
 //   interval_intersect: {
--- a/protos/perfetto/trace_summary/v2_metric.proto
+++ b/protos/perfetto/trace_summary/v2_metric.proto
@@ -158,11 +158,14 @@ message TraceMetricV2Spec {
 
   message DimensionSpec {
     optional string name = 1;
+    // The type of the dimension. Must be specified.
     optional DimensionType type = 2;
   }
   // The columns from `query` which will act as the "dimensions" for the metric.
   // For a given set of dimensions, there must be exactly *one* value emitted.
   // Optional.
+  // If the `dimensions_specs` field is defined, then the type of each
+  // dimension must be specified.
   repeated DimensionSpec dimensions_specs = 5;
   // Either dimensions or dimensions_specs should be defined, but not both.
   repeated string dimensions = 2;
--- a/protos/perfetto/trace_summary/v2_metric.proto
+++ b/protos/perfetto/trace_summary/v2_metric.proto
@@ -38,7 +38,10 @@ import "protos/perfetto/perfetto_sql/structured_query.proto";
 //
 // ```
 // id: "memory_per_process"
-// dimensions: "process_name"
+// dimensions_specs: {
+//   name: "process_name"
+//   type: STRING
+// }
 // value: "avg_rss_and_swap"
 // query: {
 //   table: {
@@ -66,7 +69,14 @@ import "protos/perfetto/perfetto_sql/structured_query.proto";
 //
 // ```
 // id: "memory_per_process_and_cuj"
-// dimensions: "process_name"
+// dimensions_specs: {
+//  name: "process_name"
+//  type: STRING
+//}
+// dimensions_specs: {
+//  name: "cuj_name"
+//  type: STRING
+//}//
 // value: "avg_rss_and_swap"
 // query: {
 //   interval_intersect: {
@@ -148,10 +158,23 @@ message TraceMetricV2Spec {
   // Required.
   optional string id = 1;
 
+  enum DimensionType {
+    DIMENSION_TYPE_UNSPECIFIED = 0;
+    STRING = 1;
+    INT64 = 2;
+    DOUBLE = 3;
+  }
+
+  message DimensionSpec {
+    optional string name = 1;
+    optional DimensionType type = 2;
+  }
   // The columns from `query` which will act as the "dimensions" for the metric.
   // For a given set of dimensions, there must be exactly *one* value emitted.
   // Optional.
-  repeated string dimensions = 2;
+  repeated DimensionSpec dimensions_specs = 5;
+  // Deprecated, use dimensions_specs instead.
+  repeated string dimensions = 2 [deprecated = true];
 
   // The column from `query` which will act as the "value" for the metric. This
   // must be a column containing only integers/doubles/nulls. Strings are *not*
@@ -230,6 +253,9 @@ message TraceMetricV2 {
     // The dimensions that `value` should be associated with. The order of
     // dimensions matches precisely the order of dimension names given by the
     // `spec`.
+    // The type of the dimension is infered from the sql column type.
+    // In case dimensionSpec is specified, the dimension type must match the
+    // type specified in the spec.
     message Dimension {
       message Null {}
       oneof value_oneof {
--- a/protos/perfetto/trace_summary/v2_metric.proto
+++ b/protos/perfetto/trace_summary/v2_metric.proto
@@ -173,8 +173,8 @@ message TraceMetricV2Spec {
   // For a given set of dimensions, there must be exactly *one* value emitted.
   // Optional.
   repeated DimensionSpec dimensions_specs = 5;
-  // Deprecated, use dimensions_specs instead.
-  repeated string dimensions = 2 [deprecated = true];
+  // Either dimensions or dimensions_specs should be defined, but not both.
+  repeated string dimensions = 2;
 
   // The column from `query` which will act as the "value" for the metric. This
   // must be a column containing only integers/doubles/nulls. Strings are *not*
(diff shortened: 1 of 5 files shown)


protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -148,10 +158,23 @@ message TraceMetricV2Spec {
...
+  message DimensionSpec {
+    optional string name = 1;
+    optional DimensionType type = 2;
+  }
   // The columns from `query` which will act as the "dimensions" for the metric.
   // For a given set of dimensions, there must be exactly *one* value emitted.
   // Optional.
-  repeated string dimensions = 2;
+  repeated DimensionSpec dimensions_specs = 5;
+  // Deprecated, use dimensions_specs instead.
```
LalitMaganti: I wouldn't deprecate this, I would actually ask that either (but not both) be provided. For getting started, it's nice if you can just list the names without having to have all the nested messages.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +112,162 @@ base::Status CreateQueriesAndComputeMetrics(
   }
   protozero::HeapBuffered<protos::pbzero::TraceSummary> summary;
   for (auto m = queries_per_metric.GetIterator(); m; ++m) {
+    auto metric_name = m.key().c_str();
```
LalitMaganti: Const std::string&

You shouldn't also do .c_str here as that creates a copy.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +112,162 @@ base::Status CreateQueriesAndComputeMetrics(
...
-                             m.key().c_str());
+                             metric_name);
     }
     auto* metric = summary->add_metric();
     metric->AppendBytes(protos::pbzero::TraceMetricV2::kSpecFieldNumber,
                         m.value().spec.data, m.value().spec.size);
 
     auto it = processor->ExecuteQuery(m.value().query);
+    protos::pbzero::TraceMetricV2Spec::Decoder spec_decoder(
+        m.value().spec.data, m.value().spec.size);
```
LalitMaganti: Just m.value().spec

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +112,162 @@ base::Status CreateQueriesAndComputeMetrics(
...
+
     uint32_t col_count = it.ColumnCount();
+    base::FlatHashMap<std::string, uint32_t> column_name_to_index;
```
LalitMaganti: Just do a linear scan here instead of creating a map. It's o(n^2) but in practice given the number of columns it'll be faster.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +112,162 @@ base::Status CreateQueriesAndComputeMetrics(
...
+    }
     while (it.Next()) {
       PERFETTO_CHECK(col_count > 0);
-      const auto& value = it.Get(col_count - 1);
-
+      const auto& metric_value_column = it.Get(*metric_value_index);
       // Skip null rows.
-      if (value.is_null()) {
+      if (metric_value_column.is_null()) {
+        PERFETTO_DLOG(
```
LalitMaganti: Don't dlog in trace processor.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +112,162 @@ base::Status CreateQueriesAndComputeMetrics(
...
-                m.key().c_str());
-        }
-      }
-
-      switch (value.type) {
+      // Read metric value.
+      switch (metric_value_column.type) {
         case SqlValue::kLong:
-          row->set_value(static_cast<double>(value.AsLong()));
+          row->set_value(static_cast<double>(metric_value_column.AsLong()));
```
LalitMaganti, line 214: Doing as* will crash tp if the type does not match. There should be a check to make sure the type matches the declared type.

LalitMaganti (commented):
Quick first pass.

src/trace_processor/trace_summary/summary.cc
elrakadm: Done

src/trace_processor/trace_summary/summary.cc
elrakadm: Done

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Sounds good, I add a check to verify that not both are defined.

src/trace_processor/trace_summary/summary.cc
elrakadm: I think having the map will be faster even for low number of columns given there are a lot of rows produce.
Because the map is only created the first row not on each one. But the linear scan will need to be done on each row to figure out the dimension index.

src/trace_processor/trace_summary/summary.cc
elrakadm: Removed

src/trace_processor/trace_summary/summary.cc
elrakadm: The type check is done in reading dimension specs. 
Metric values are expected to be long or double and the check is done to verify it or I am missing something else ?

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -38,7 +38,10 @@ import "protos/perfetto/perfetto_sql/structured_query.proto";
 //
 // ```
 // id: "memory_per_process"
-// dimensions: "process_name"
+// dimensions_specs: {
```
LalitMaganti: please keep it as dimensions.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +112,168 @@ base::Status CreateQueriesAndComputeMetrics(
...
+      }
+
+      if (dimension_specs.empty()) {
+        // Dimensions are defined without an explicit type
+        // Infer the type from the sql value.
+        for (auto dim_name_it = spec_decoder.dimensions(); dim_name_it;
+             ++dim_name_it) {
+          protos::pbzero::TraceMetricV2::MetricRow::Dimension* dimension =
+              row->add_dimension();
+          auto* dim_index =
```
LalitMaganti: We should do this once per metric instead of per row. We should figure out the mapping upfront.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +112,168 @@ base::Status CreateQueriesAndComputeMetrics(
...
+          protos::pbzero::TraceMetricV2::MetricRow::Dimension* dimension =
+              row->add_dimension();
+
+          const auto& dim_spec = dimension_specs[i];
+          std::string dim_name = dim_spec.name().ToStdString();
+          protos::pbzero::TraceMetricV2Spec::DimensionType dimension_type =
+              static_cast<protos::pbzero::TraceMetricV2Spec::DimensionType>(
+                  dim_spec.type());
+
+          auto* dim_index = column_name_to_index.Find(dim_name);
```
LalitMaganti: We should be figuring this out upfront and only store the indexes for use here.

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Done

src/trace_processor/trace_summary/summary.cc
elrakadm: Unified reading dimensions and dimensions spec, the indices now are precomputed

src/trace_processor/trace_summary/summary.cc
elrakadm: Unified reading dimensions and dimensions spec, the indices now are precomputed

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +111,182 @@ base::Status CreateQueriesAndComputeMetrics(
...
-        switch (dim.type) {
-          case SqlValue::kLong:
-            row->add_dimension()->set_int64_value(dim.AsLong());
-            break;
-          case SqlValue::kDouble:
-            row->add_dimension()->set_double_value(dim.AsDouble());
-            break;
-          case SqlValue::kString:
-            row->add_dimension()->set_string_value(dim.AsString());
+    if (spec_decoder.dimensions_specs() && spec_decoder.dimensions()) {
```
LalitMaganti: You should use has_dimension_specs and has_dimensions instead.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -112,68 +111,182 @@ base::Status CreateQueriesAndComputeMetrics(
...
+    if (spec_decoder.dimensions_specs()) {
+      for (auto dim_spec_it = spec_decoder.dimensions_specs(); dim_spec_it;
+           ++dim_spec_it) {
+        protos::pbzero::TraceMetricV2Spec::DimensionSpec::Decoder dim_spec(
+            *dim_spec_it);
+        std::string dim_name = dim_spec.name().ToStdString();
+        protos::pbzero::TraceMetricV2Spec::DimensionType dimension_type =
+            static_cast<protos::pbzero::TraceMetricV2Spec::DimensionType>(
+                dim_spec.type());
+        if (dimension_type ==
```
LalitMaganti, line 159: This feels legit no? Why not treat this the same as if you just specified the name with dimensions?

LalitMaganti (approved):
Looks great thanks!

src/trace_processor/trace_summary/summary.cc
elrakadm: I want TP to be strict, if the summary spec will define a dimension spec and won't list the names (using old dimensions) then you must specify a type.

It will make the returned dimension type known before the metric runs so downstream systems can expect a specific type.

src/trace_processor/trace_summary/summary.cc
elrakadm: Done.

src/trace_processor/trace_summary/summary.cc
LalitMaganti: Can we document this in the proto please?

elrakadm: Done

PR #1423, opened 2025-05-06
bazel: remove allowlist visibility for direct zero targets in favour of amalgams

LalitMaganti:
This makes zero consistent with the other proto types.


ddiproietto (commented):
Makes sense, thanks!

ddiproietto (approved):

PR #1429, opened 2025-05-07
Add DataGrid widget

stevegolton:
This commit adds a new `DataGrid` widget, a versatile component for displaying tabular data.

It supports:
- Support for various data sources (e.g., SQL, in-memory).
- Pagination for handling large datasets.
- Sorting and filtering, with optional controlled mode for external state management.
- Cell-level menus for value-based filtering.
- Column header menus for sorting and null-value filtering.
- A toolbar displaying active filters, pagination controls, and a reset button.
- Customizable cell content rendering.

This widget replaces the current QueryTable implementation, so it can be seen anywhere where the QueryTable was used (e.g. query & explore pages, omnibox query results).

The intent is that this widget will eventually be used anywhere where we need to display large tabular data sets.

Note: To avoid confusion, filtering is disabled for now. This could cause confusion when used with e.g. the explore page and the query page because it's not clear that filters applied to the data grid are ephemeral and local to that data grid instance, rather than being applied higher up to the explore page's node or the query page's query.


aMayzner:
Is there an option to disable the filtering and sorting and have pure data visualisation? I guess it might come in handy when we don't want to confuse people (for example when we won't support filtering from the table in the Explore Page yet)

stevegolton:
> Is there an option to disable the filtering and sorting and have pure data visualisation? I guess it might come in handy when we don't want to confuse people (for example when we won't support filtering from the table in the Explore Page yet)

Not yet but I can certainly add it.

stevegolton:
> We still actually want support for people to use the table to display the filters, it is just that the filters will be displayed much higher up in the page instead of directly in the table.

I think both.

We'd want options to:
1. Hide filters in the toolbar (just like you say). The filters and menus still remain.
2. Disable all filter UI elements entirely. Filters still work, but the user cannot manipulate them.

Initially we will just disable filtering everywhere to avoid confusion (2). This will avoid any initial confusion. Then, after we plumb the filters in on the explore page, we can re-enable filtering support but just hide the filters in the toolbar (1).

ui/src/components/widgets/data_grid/data_grid.ts
```diff
@@ -0,0 +1,680 @@
...
+   * Display applied filters in the toolbar. Set to false to hide them, for
+   * example, if filters are displayed elsewhere in the UI. This does not
+   * disable filtering functionality.
+   *
+   * Defaults to true.
+   */
+  readonly showFiltersInToolbar?: boolean;
+}
+
+const DEFAULT_ROWS_PER_PAGE = 50;
```
aMayzner: Shouldn't global variables be at the top of the page? Also isn't 50 too few? I would rather have 100 and then see the issue with it

ui/src/components/widgets/data_grid/data_grid.ts
```diff
@@ -0,0 +1,680 @@
...
+    onSortByChange: ((sortBy: SortBy) => void) | undefined,
+    onFiltersChange:
+      | ((filters: ReadonlyArray<FilterDefinition>) => void)
+      | undefined,
+    maxRowsPerPage: number,
+    showFilters: boolean,
+  ) {
+    return m('.pf-data-grid__toolbar', [
+      m(Button, {
+        icon: 'restart_alt',
```
aMayzner: Shouldn't we use icons from `Icons.{sth}`? Here and below. This way we can make sure that we use consistent icons for similar stuff

ui/src/components/widgets/data_grid/data_grid.ts
```diff
@@ -0,0 +1,680 @@
...
+            }
+          },
+        }),
+        m(
+          'span.pf-data-grid__toolbar-page',
+          `${(this.currentPage * maxRowsPerPage + 1).toLocaleString()}-${
+            Math.min(
+              (this.currentPage + 1) * maxRowsPerPage,
+              totalRows,
+            ).toLocaleString() || '0'
```
aMayzner: Can we extract this whole maths/string outside for readability?

ui/src/components/widgets/data_grid/in_memory_data_source.ts
```diff
@@ -0,0 +1,240 @@
...
+import {
+  DataGridDataSource,
+  DataSourceResult,
+  FilterDefinition,
+  RowDef,
+  SortBy,
+  SortByColumn,
+} from './common';
+
+export class InMemoryDataSource implements DataGridDataSource {
```
aMayzner, line 25: Can we please test the functionality of this class?

aMayzner (approved):

LalitMaganti (approved):

PR #1437, opened 2025-05-07
traceconv: implement function to go from text -> binary

LalitMaganti:
Implement it in a very trivial way but better than not having it at
all


src/traceconv/main.cc
```diff
@@ -76,6 +81,23 @@ uint64_t StringToUint64OrDie(const char* str) {
   return number;
 }
 
+int TextToTrace(std::istream* input, std::ostream* output) {
+  std::string trace_text((std::istreambuf_iterator<char>(*input)),
+                         (std::istreambuf_iterator<char>()));
```
rsavitski:
```suggestion
  std::string trace_text(std::istreambuf_iterator<char>{*input},  
                         std::istreambuf_iterator<char>{});  
```

src/traceconv/main.cc
```diff
@@ -76,6 +81,23 @@ uint64_t StringToUint64OrDie(const char* str) {
   return number;
 }
 
+int TextToTrace(std::istream* input, std::ostream* output) {
```
rsavitski, line 84: Who's the audience for this, given that protoc and protoscope exist?

Fwiw I tried on a "default" trace and it completely clobbered scheduling because it re-encoded compact_sched in a different way (I didn't investigate further). It only worked with me explicitly turning compact_sched off.

LalitMaganti: Pinged you on a thread.

rsavitski (approved):

PR #1446, opened 2025-05-08
[API] Single data source for all track events

etiennep-chromium:
This CL defines a single TrackEventDataSource that covers
all registries and exposes TrackEvent functions.
Based on https://android-review.git.corp.google.com/c/platform/external/perfetto/+/2171302

Migrating from https://android-review.git.corp.google.com/c/platform/external/perfetto/+/3431685/7

Bug: [347706944](https://issuetracker.google.com/issues/347706944)


src/tracing/internal/track_event_internal.cc
```diff
@@ -209,50 +249,74 @@ protos::pbzero::BuiltinClock TrackEventInternal::clock_ = kDefaultTraceClock;
 // static
 bool TrackEventInternal::disallow_merging_with_system_tracks_ = false;
 
+// static
+void TrackEventInternal::EnableRegistry(
+    const TrackEventCategoryRegistry* registry,
+    const protos::gen::TrackEventConfig& config,
+    uint32_t internal_instance_index) {
+  for (size_t i = 0; i < registry->category_count(); i++) {
```
etiennep-chromium, line 250: Moving comment form gerrit:
> I'm picturing a race condition between registry registration and trace starting.
There should be no memory corruption, but it's possible that a registry category is never going to be enabled, while it should have been.
Can we hold the mutex here?

etiennep-chromium: TL;DR: I don't think a lock achieves anything useful here.

`TrackEventInternal::GetRegistries()` called from `OnSetup()` already takes a mutex to ensure registries is consistent.
I'm guessing you're suggesting to extend the scope of that lock to all of `TrackEventInternal::EnableTracing()`.
When a lock might be necessary here, this means TrackEventInternal::EnableTracing() is racing with `TrackEventInternal::AddRegistry()`, say

- Thread A calls TrackEventDataSource::OnSetup() -> GetRegistries()
- Thread B calls TrackEventInternal::AddRegistry
- Thread A calls TrackEventInternal::EnableTracing -> it now has an old list of registries

Let's say that we extend the scope of lock to EnableTracing(), then Thread A might be forced to finish before Thread B, in which case we end up with the same issue "a registry category is never going to be enabled, while it should have been". However, the premise is that Thread A and B are racing; this would be a problem at the callsite which can't be solved by making TrackEventInternal more thread-safe. By extending the lock scope, we simply push the barrier to a little later, but didn't eliminate a potential racy situation.

etiennep-chromium: > Mmmhh, I think we can fully solve it.

I missed the fact that we call TrackEventInternal::EnableRegistry in AddRegistry; I think this works because Trace() considers an instance active right before OnSetup() is called.

src/tracing/internal/track_event_internal.cc
```diff
@@ -139,6 +143,18 @@ bool NameMatchesPatternList(const std::vector<std::string>& patterns,
...
+  static base::NoDestructor<std::vector<const TrackEventCategoryRegistry*>>
+      registries;
+  return registries.ref();
+}
+
+std::mutex& GetTrackEventCategoryMutex() {
+  static std::mutex mu;
+  return mu;
+}
+
```
ddiproietto, line 155: Nit, optional:

Instead of having two static functions, can we have a single object, like

```
class RegistryState {
 public:
  RegistryState& GetInstance() {
    static base::NoDestructor<RegistryState> state;
    return state.ref();
  }
  
  // ...
  
  // ...

 private:
  std::mutex mu_;
  std::vector<>....
}
```

So that the mutex and the vector it protects are closer together? (I'm taking inspiration from perfetto::shlib::TrackEvent::GlobalState and  TrackEventSessionObserverRegistry). Maybe it makes less sense here, because we often have to use the mutex and the vector outside separately and we don't encapsulate all the behavior yet.

include/perfetto/tracing/internal/track_event_data_source.h
```diff
@@ -220,53 +220,37 @@ struct TrackEventDataSourceTraits : public perfetto::DefaultDataSourceTraits {
...
-  }
+  void OnStart(const DataSourceBase::StartArgs& args) override;
 
   void OnStop(const DataSourceBase::StopArgs& args) override {
     auto outer_stop_closure = args.HandleStopAsynchronously();
     StopArgsImpl inner_stop_args{};
     uint32_t internal_instance_index = args.internal_instance_index;
     inner_stop_args.internal_instance_index = internal_instance_index;
-    inner_stop_args.async_stop_closure = [internal_instance_index,
+    auto registries = TrackEventInternal::GetRegistries();
```
ddiproietto: Thanks for fixing the race condition at OnSetup, with EnableTracing.

I think we should fix the same race condition here and in other places as well.

Basically everywhere we use `TrackEventInternal::GetRegistries()` (lookin at the current use cases I think the function should be removed).

That means we will call user callbacks of Observer under the mutex, but that should be ok, you shouldn't add new registries in those. It's probably a good idea to use an std::recursive_mutex, for this reason.

src/tracing/internal/track_event_internal.cc
etiennep-chromium: Not quite what you suggested, but I moved mu_ and registries as non-static members of TrackEventInternal.

include/perfetto/tracing/internal/track_event_data_source.h
etiennep-chromium: > I think we should fix the same race condition here and in other places as well.

Done for TrackEventInternal::DisableTracing.
The 3 other calls to GetRegistries() are use for ForEachObserverForRegistries.
On one hand, I find this a bit risky because ForEachObserverForRegistries itself takes another lock, TrackEventSessionObserverRegistry::mutex_
On the other hand, it seems we make no effort (before this CL) at calling observers in TrackEventSessionObserverRegistry when they are registered late (e.g. start a tracing session, add an observer). So I don't think having a lock around ForEachObserverForRegistries would help.

PR #1463, opened 2025-05-09
protovm: insert/delete array operations

rukkal:


protos/perfetto/protovm/vm_program.proto
```diff
@@ -117,9 +120,16 @@ message VmOpMerge {}
...
-// Delete the field or message pointed by DST.
+// Deletes the field or message pointed by DST.
+// If DST points to an array element (repeated field), subsequent elements are
+// shifted left by one position.
 message VmOpDel {}
 
+// Inserts SRC into DST. If DST points to an array element, the pointed element
+// (and the subsequent ones) are shited right by one position. In other cases,
+// the operation is semantically equivalent to VmOpSet.
+message VmOpInsert {}
```
rukkal, line 131: @primiano atm VmOpInsert is only used for array insertions, but we might want to keep a generic (array agnostic) name for possible future use cases. E.g.:
- insertion of characters into a string (or bytes)
- a bit more fancy: we add the concept of multimap/multiset (in addition to the current map), where the "insert" operation could add multiple elements with the same value

primiano (approved):
Sorry i thought i reviewed this but somehow I left it pending.
The idea itself LGTM, I am just trying to figure out how you are going to use in practice.
How will you know what's the right index?
Is it because in some cases indexes have a very specific semantic?

rukkal:
> Is it because in some cases indexes have a very specific semantic?

Exactly! In WindowManager there are arrays of "subwindows" where the index corresponds to the z-order

I am putting this on hold for now as we are exploring some ideas that won't need the array operations

primiano (approved):

PR #1468, opened 2025-05-09
perfetto: improve scripts

LalitMaganti:
* new-branch no longer sets upstream, uses origin/main by default
* presubmit always runs against origin/main


stevegolton (approved):
Nice!

PR #1481, opened 2025-05-12
Android Java SDK: don't call native methods in the 'Category' constructor.

KirillTim:
If the Category is stored as a static final field, its constructor is called on the class loading phase. That requires the native library also to be loaded in the class's static initialization block, which is not what we usually want.

In this commit, we also fix the race condition between `#register()`, `#unregister()` and `#isRegisteredAndEnabled()` methods.

To test the new initialization workflow, the `System.loadLibrary` call in `PerfettoTraceTest` class is moved from the `static {}` block to the `setUp` method. This method is called at runtime after all static final fields are created.

Tested:
```
```


src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -124,7 +141,7 @@ public boolean isRegistered() {
 
     /** Returns the native pointer for the category. */
     @Override
-    public long getPtr() {
+    public synchronized long getPtr() {
```
zezeozue: Synchronizing on this I think will lead to contention on every trace point. Can we avoid it?

KirillTim: Oh, wow, indeed.
Will fix it, thanks!

KirillTim: Done. 
We now return volatile `mPtr` from here. I've also fixed a bug here: `mExtraPtr` was used instead of `mPtr`.

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -95,37 +93,57 @@ public Category(String name, List<String> tags) {
...
     }
 
     /** Unregister the category. */
-    public Category unregister() {
-      native_unregister(mPtr);
-      mIsRegistered = false;
+    public synchronized Category unregister() {
+      if (mIsRegistered) {
+        // mIsRegistered == true implies mPtr != 0
+        mIsRegistered = false;
```
zezeozue, line 119: If registered is equivalent to mPtr != 0. Can we remove the registered variable so we don't maintain same internal state twice .  Can add a helper method

zezeozue (approved):

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -95,37 +93,57 @@ public Category(String name, List<String> tags) {
...
-    /** Whether the category is enabled or not. */
-    public boolean isEnabled() {
-      return native_is_enabled(mPtr);
+    /** Whether the category is registered and enabled or not. */
+    public boolean isRegisteredAndEnabled() {
+      // This method is explicitly not synchronized, because we want it to be as fast as possible.
+      // mIsRegistered is volatile and is set only from `#register()` or `#unregister()` methods,
+      // both of them are synchronized.
+      if (mIsRegistered) {
+        // mIsRegistered == true implies mPtr != 0
```
LalitMaganti: I'm not sure this comment makes sense. It's very much possible for another thread to have called unregister between the previous check. I think we should make sure that we can guard against something which is fundamentally racy.

KirillTim: > It's very much possible for another thread to have called unregister between the previous check.

It is fine. The `mIsRegistered` check is just a fast-path optimization to not call the native method if we know in java that the category is not registered.

but if the category is registered (or is unregistered in the process) we go to native and check in native_is_enabled the register and if it is enabled (if we have a consumer) state

LalitMaganti: Sure please add that as a comment then to make it clear we've thought about this case. As it stands, the current comment makes it feel otherwise.

KirillTim: We discussed with @zezeozue and decided to remove the `mIsRegister` check from here and move it to the `PerfettoTrackEventBuilder`, see the https://github.com/google/perfetto/pull/1481/commits/bbe0cc1d5939d169f7efe941c8bd306eab21a65e.

LalitMaganti (approved):

PR #1491, opened 2025-05-12
Fix memory corruption for TAR files >4GB on browser

yohaigevim:
Add a size validation to ensure file sizes do not exceed the maximum value of `size_t`,
preventing overflows and incorrect calculations in 32-bit environments.

Welcome to Perfetto!

For more details please see

### DIFF
--- a/src/trace_processor/importers/archive/tar_trace_reader.cc
+++ b/src/trace_processor/importers/archive/tar_trace_reader.cc
@@ -259,6 +259,18 @@ base::StatusOr<TarTraceReader::ParseResult> TarTraceReader::ParseMetadata() {
                            size.status().message().c_str());
   }
 
+  // Ensure the size fits within `size_t` to prevent issues on 32-bit platforms
+  // (e.g., in-browser environments) where `size_t` is smaller than `uint64_t`.
+  // `metadata_->size` is a `uint64_t`, but passing it to methods like
+  // `buffer_.SliceOff` (which take `size_t`) may cause overflow if it exceeds
+  // the max representable `size_t` value.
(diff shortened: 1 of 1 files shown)


src/trace_processor/importers/archive/tar_trace_reader.cc
```diff
@@ -259,6 +259,18 @@ base::StatusOr<TarTraceReader::ParseResult> TarTraceReader::ParseMetadata() {
                            size.status().message().c_str());
   }
 
+  // Ensure the size fits within `size_t` to prevent issues on 32-bit platforms
+  // (e.g., in-browser environments) where `size_t` is smaller than `uint64_t`.
+  // `metadata_->size` is a `uint64_t`, but passing it to methods like
+  // `buffer_.SliceOff` (which take `size_t`) may cause overflow if it exceeds
+  // the max representable `size_t` value.
+  const uint64_t max_size = std::numeric_limits<size_t>::max();
+  if ((size.value() & (~(max_size))) != 0) {
```
LalitMaganti: This seems a pretty obscure check. Why not just check that it's less than this size (which is what I think you semantically want there).

yohaigevim: In order to avoid this error when building the trace_processor_shell:
error: result of comparison 'unsigned long' > 18446744073709551615 is always false [-Werror,-Wtautological-type-limit-compare]

LalitMaganti: You should be able to fix that by doing `std::less<>()(size.value(), max_size)`

PR #1493, opened 2025-05-13
Add handling for irq_vectors/local_timer_[entry|exit] events

johnstultz-work:
On x86, the local apic is usually what generates timer irqs. Unfortunately, it's "special", so it doesn't register via the standard generic irq code.

Instead, it uses irq_vectors/ which is an x86 specific trace event system.

While one can capture irq_vectors/* and it shows up in the ftrace tracks, it would be nice if the irq_vectors/* entries would be rendered similarly to the IRQ track in the Perfetto UI.

So this pull request implements this feature. Adding the events to the events_list and parsing the events, adding them to the IRQ track.

Bug: 415090521


src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -2747,6 +2755,24 @@ void FtraceParser::ParseIrqHandlerExit(uint32_t cpu,
       });
 }
 
+void FtraceParser::ParseLocalTimerEntry(uint32_t cpu, int64_t timestamp) {
+  TrackId track = context_->track_tracker->InternTrack(kIrqBlueprint,
+                                                       tracks::Dimensions(cpu));
+
+  base::StringView irq_name = "LocalTimer";
```
LalitMaganti: If we're hardcoding this, can you just intern this in the constructor?

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -2747,6 +2755,24 @@ void FtraceParser::ParseIrqHandlerExit(uint32_t cpu,
       });
 }
 
+void FtraceParser::ParseLocalTimerEntry(uint32_t cpu, int64_t timestamp) {
+  TrackId track = context_->track_tracker->InternTrack(kIrqBlueprint,
```
LalitMaganti, line 2760: Just to confirm, is it the case that if this is executing, no other IRQ can execute in parallel?

johnstultz-work: I'll have to look more into this. I'm not sure if irqs are disabled for the entire time the local timer irq is running. I suspect it is the case for simplicity, but I know with IRQ priorities on some hardware you can have another IRQ stack land in IRQ context, but that's usually limited for things like NMI/FIQ style use.

Is there a trick for handling things on the perfetto side if we do have nested irqs?

src/trace_processor/importers/ftrace/ftrace_parser.cc
johnstultz-work: Apologies, I'm very much a novice when it comes to C++.   Do you mean just declaring the variable as:
```
   base::StringView irq_name("LocalTimer")
```
or something similar?  Or do you mean something else?

src/trace_processor/importers/ftrace/ftrace_parser.cc
LalitMaganti: > Is there a trick for handling things on the perfetto side if we do have nested irqs?

Nested IRQs are OK, they'll work out of the box. That is, if all the slice perfectly nests within each other.

The problem comes from when there is overlapping *without* nesting. In that case, you'll need to either use a different track for these events or use a different approach (TrackCompressor) for adding slices.

johnstultz-work: These should only nest. There's no alternating back and forth between different IRQs on the same CPU.

src/trace_processor/importers/ftrace/ftrace_parser.cc
LalitMaganti: See https://github.com/google/perfetto/blob/8f3a02ea772c7f37c94915e358f3320d54764179/src/trace_processor/importers/ftrace/ftrace_parser.cc#L467

Basically you are just having a constant string here, if you intern it in the constructor, you pay the cost of string interning once per trace instead of once per event. For things like IRQs where you expect thousands/millions of events this really adds up.

src/trace_processor/importers/ftrace/ftrace_parser.cc
LalitMaganti: Then LGTM on this approach.

src/trace_processor/importers/ftrace/ftrace_parser.cc
johnstultz-work: Ack. Updated the series with this fix. Let me know if you see anything else that needs tweaking.

LalitMaganti (approved):

aMayzner (approved):

PR #1499, opened 2025-05-13
perfetto/ftrace: Add support for ftrace's max_graph_depth option

johnstultz-work:
In the perfetto config, add a new function_graph_max_depth option, which sets the max depth for the function_graph tracer to follow in its traces.

This is helpful to limit the amount of data that function_graph traces collect, while still providing insight as to what might be taking the time.

Bug: 417357433


protos/perfetto/config/perfetto_config.proto
```diff
@@ -1255,6 +1255,9 @@ message FtraceConfig {
 
   // Previously drain_buffer_percent, perfetto v43-v47.
   reserved 26;
+  // If |enable_function_graph| is true, only trace this many calls down
+  // the stack. Sets the max_graph_depth value in sys/kernel/tracing/
```
LalitMaganti: Please add some comments on which Perfetto version (51) and Android version (25Q3) this feature is available from. Also please add a note that this is only respected for the *first* tracing session that enables funcgraph tracing.

johnstultz-work: Done. Thanks so much for the review and feedback!

rsavitski: Isn't it the /last/ activated funcgraph tracing session that will win here? I'll send a separate patch to fix up the comment.

LalitMaganti (approved):

PR #1511, opened 2025-05-14
frozen_ftrace: Erase the trace data after read

mhiramat:
After read out all data and flush the writer from "linux.frozen_ftrace" data source, make sure to clear trace data in the kernel.

Background of this change is the upstream kernel patch[1], which makes the kernel to recover the trace data that is already read in the previous boot. This actually expects the perfetto AOT use case, because it reads out the kernel data periodically. Thus we can not read most of the data in the persistent ring buffer next boot time.
However, after perfetto read out the recovered previous boot data via "linux.frozen_ftrace" data source, the data will be 
recovered too (unless it is reused.) This means perfetto will read the same data again next boot time. To prevent it, clear the trace data after read via frozen_ftrace data source.

This does not change anything from user's viewpoint, because on the current kernel, the read data is gone. And on the kernel with the patch[1], it is cleared explicitly.

[1] https://lore.kernel.org/all/174709742769.1964496.18203163435305117893.stgit@mhiramat.tok.corp.google.com/

Change-Id: Id4a7852c493fc4ac635396314eceb9d52ac57732


src/traced/probes/ftrace/frozen_ftrace_data_source.cc
```diff
@@ -190,6 +190,8 @@ void FrozenFtraceDataSource::ReadTask() {
 void FrozenFtraceDataSource::Flush(FlushRequestID,
                                    std::function<void()> callback) {
   writer_->Flush(std::move(callback));
+  // Ensure the read data is erased and not recovered in the next boot.
+  tracefs_->ClearTrace();
```
rsavitski: Flush() isn't logically the right place to do this. It's an IPC method [1] that asks the writer to commit (=flush) all data up to the current point in time. Depending on the config, that could be many times per trace and not just before stopping (e.g. [2]).

In this case, could we do the clearing in the `FrozenFtraceDataSource` destructor? I belive that is reasonable since it's part of the cleanup, and there is no need for an error return codepath.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/ipc/consumer_port.proto;l=61;drc=0773e9cfee445991a1d90f681438b4caa3f6fd80

[2] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/config/trace_config.proto;l=259;drc=8beea9a98dbca3394b2d808b661ebbf4528c0335

rsavitski (changes requested):

src/traced/probes/ftrace/frozen_ftrace_data_source.cc
mhiramat: OK, let's move it in the destructor.

PR #1513, opened 2025-05-14
Make Day Explorer blames available in the Perfetto UI.

simonmacm:
Since they can take a while to calculate we hide them behind a command.

To make it more obvious when the command has finished, we open the first group by default.

While we're here, clean up a little bit: import correct objects for iterator types, and check for google3 with a fallthrough only once instead of repeatedly.

Bug: http://b/417417058

### DIFF
--- a/ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
@@ -1738,28 +1738,6 @@ export default class implements PerfettoPlugin {
     return features;
   }
 
-  async addTracks(ctx: Trace, features: Set<string>): Promise<void> {
-    const containedTraces = (ctx.openerPluginArgs?.containedTraces ??
-      []) as ContainedTrace[];
-
-    await ctx.engine.query(PACKAGE_LOOKUP);
-    await this.addNetworkSummary(ctx, features);
-    await this.addBluetooth(ctx, features);
-    await this.addModemRil(ctx, features);
-    await this.addKernelWakelocks(ctx);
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
```diff
@@ -1071,14 +1079,10 @@ export default class implements PerfettoPlugin {
...
-        `INCLUDE PERFETTO MODULE
-            google3.wireless.android.telemetry.trace_extractor.modules.atom_counters_slices`,
-      );
-    } catch (e) {
-      return;
-    }
+    await e.query(
+      `INCLUDE PERFETTO MODULE
+          google3.wireless.android.telemetry.trace_extractor.modules.atom_counters_slices`,
+    );
```
stevegolton, line 1085: Why did you remove the try..catch here and above? Does this cause problems for non-googlers or am I missing another check elsewhere?

simonmacm: Sorry, this was inadequately explained in the description.

We now have a single try/catch that is then used to determine if google3 is there (`features.contains('google3')`), see line 1730.

We then don't enter this method at all if we don't have google3, see line ~1800.

I want to remove these try/catches because they also fail if the module crashes. In future I'll move to having a no-op module for google3 detection.

stevegolton: Ah ok that makes sense. Thanks for clarifying.

stevegolton (approved):

PR #1521, opened 2025-05-14
tp: add query planning and interpreter code for actually using indexes

LalitMaganti:
This CL adds add the necessary logic for making use of dataframe
indexes. We are only using them for point queries at the moment, not
range queries but it works on all column storage types (except id where
it does not make sense) and null types.


src/trace_processor/dataframe/impl/query_plan.h
```diff
@@ -248,13 +251,17 @@ class QueryPlanBuilder {
...
   // Attempts to apply optimized filtering on sorted data.
   // Returns true if the optimization was applied.
-  bool TrySortedConstraint(
-      const FilterSpec& fs,
-      const StorageType& ct,
-      const NonNullOp& op,
-      const bytecode::reg::RwHandle<CastFilterValueResult>& result);
+  bool TrySortedConstraint(FilterSpec& fs,
+                           const StorageType& ct,
+                           const NonNullOp& opF);
```
aMayzner, line 264: Typo?

src/trace_processor/dataframe/impl/query_plan.cc
```diff
@@ -128,12 +130,75 @@ inline uint8_t GetDataSize(StorageType type) {
   }
 }
 
+SimpleNullability NullabilityToSimpleNullability(Nullability nullability) {
```
aMayzner, line 133: I really don't like `SimpleNullability`. Aren't the types wrong? I would expect, from the name, `SimpleNullability` to behave the same way as `Nullability`.

src/trace_processor/dataframe/types.h
```diff
@@ -12,18 +12,22 @@ namespace perfetto::trace_processor::dataframe {
...
-  friend class Dataframe;
+  const std::vector<uint32_t>& columns() const { return columns_; }
 
-  Index(std::vector<uint32_t> _columns,
-        std::shared_ptr<std::vector<uint32_t>> _permutation_vector)
-      : columns(std::move(_columns)),
-        permutation_vector(std::move(_permutation_vector)) {}
+  const std::shared_ptr<std::vector<uint32_t>>& permutation_vector() const {
+    return permutation_vector_;
+  }
```
aMayzner, line 26: Please add a comment

PR #1545, opened 2025-05-16
summary: add support for metric v2 spec templates

LalitMaganti:
Makes it easier to define N metrics based on the same query.


protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -263,8 +296,12 @@ message TraceMetricV2 {
   }
   repeated MetricRow row = 1;
```
elrakadm, line 335: Will MetricRow.value be repeated field ?

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -183,6 +183,39 @@ message TraceMetricV2Spec {
...
 
+// A template for generating multiple v2 trace-based metrics.
+//
+// Often when people start writing more complex sets of metrics, they find that
+// they are repeating themselves a lot. This proto is intended to help with
+// that: it allows you to specify a set of metrics which share a common prefix
+// and a common set of dimensions. The template will then generate a set of
+// metrics which share the same prefix and dimensions but differ in the value
+// column.
+message TraceMetricV2TemplateSpec {
```
elrakadm, line 225: Could the `TraceMetricV2Spec` be expanded instead of creating a new proto ?
 Creating a new repeated field `value_columns` where either `value` or `value_columns` are set. Similar to dimensions vs dimensions_specs change.

As it is not clear what TraceSummary output expanded version of TraceMetricV2Spec .

src/trace_processor/trace_summary/summary.cc
```diff
@@ -53,6 +53,47 @@ struct Metric {
...
+        for (auto dim = tmpl.dimensions(); dim; ++dim) {
+          protozero::ConstChars dim_str = *dim;
+          expanded->add_dimensions(dim_str.data, dim_str.size);
+        }
+        for (auto dim = tmpl.dimensions_specs(); dim; ++dim) {
+          protozero::ConstBytes dim_spec = *dim;
+          expanded->add_dimensions_specs()->AppendRawProtoBytes(dim_spec.data,
+                                                                dim_spec.size);
+        }
+        if (tmpl.has_query()) {
```
elrakadm, line 83: the template must has a query similar to a metricSpec, should it throw an error if it doesn't ?

elrakadm (approved):

src/trace_processor/trace_summary/summary.cc
LalitMaganti: That error will show up later anyway when the code per metric runs. We can add that check in later if necessary.

PR #1546, opened 2025-05-16
[dns] track merging proto for discussion

LalitMaganti:


protos/perfetto/trace/track_event/track_descriptor.proto
```diff
@@ -148,7 +153,69 @@ message TrackDescriptor {
...
+  // requests, scheduler activity etc).
+  //
+  // Note: for tracks where `thread` or `process` are set, this option is
+  // *ignored*. See `parent_uuid` for details.
+  //
+  // Note: this flag only affects the UI for now but we expect this also to be
+  // respected by the trace processor in the future to allow for consistency
+  // between you see and what you query.
+  enum ChildMergeBehavior {
+    // When unspecified or not set, defaults to `SIBLING_MERGE_BEHAVIOR_LEGACY`.
```
etiennep-chromium, line 173: Nit: SIBLING_MERGE_BEHAVIOR_LEGACY -> CHILD_MERGE_BEHAVIOR_LEGACY

protos/perfetto/trace/track_event/track_descriptor.proto
```diff
@@ -148,7 +153,69 @@ message TrackDescriptor {
...
+  }
+  optional ChildMergeBehavior child_merge_behavior = 14;
+
+  // An opaque value which allows specifying whether two sibling tracks should
+  // be merged together. Tracks with the same `sibling_merge_key` will be merged
+  // together. If `sibling_merge_key` is not set, the track with *not* be
+  // merged with any other track. See `sibling_merge_behavior` for more details.
+  //
+  // Note: this option is only relevant for tracks where the parent has
+  // `sibling_merge_behavior` set to
```
etiennep-chromium, line 209: In Chrome, we often have a set of tracks whose parent is the process.
If we want to control merging behavior of this set, without affecting the whole process, I think this behavior makes this impossible.

How about making CHILD_MERGE_BEHAVIOR_LEGACY also honor sibling_merge_key *if specified*

So if `sibling_merge_key` is not specified:
- CHILD_MERGE_BEHAVIOR_LEGACY merges any child tracks with the same `name` *except* when those tracks have a child *themselves*
- CHILD_MERGE_BEHAVIOR_BY_SIBLING_MERGE_KEY doesn't merges child tracks

protos/perfetto/trace/track_event/track_descriptor.proto
```diff
@@ -110,12 +110,17 @@ message TrackDescriptor {
   // |extra_counter_values|).
   optional CounterDescriptor counter = 8;
 
+  // Deprecated: prefer use of `sbling_merge_behavior` instead: this option
```
etiennep-chromium: Nit: child_merge_behavior

etiennep-chromium (commented):
IIUC this adds the protos but doesn't implement the behavior?

protos/perfetto/trace/track_event/track_descriptor.proto
LalitMaganti: > If we want to control merging behavior of this set, without affecting the whole process, I think this behavior makes this impossible.

You can still do this though, just add a new "intermediate" track between the parent process track and your set of child tracks. And you set `CHILD_MERGE_BEHAVIOR_BY_SIBLING_MERGE_KEY` on the intermediate track.

etiennep-chromium: It's possible to work around, but this imposes extra work on the caller, and changes the resulting track hierarchy (which I see as an inconvenient); on the flip side is there an inconvenient to following sibling_merge_key + CHILD_MERGE_BEHAVIOR_LEGACY?

LalitMaganti: For very complex historical reasons, it is the case that any child tracks of tracks with process/thread descriptors do *not* have parent relationships with their parents in trace processor and instead are treated as siblings of their parent. 

Could we have an exception where we say that "parent" semantics are inherited *except* parent_id: sure but it would be somewhat confusing. 

I could be convinced but I need to think about this more.

etiennep-chromium: Oh..
From an API perspective (tracing/track.h), we set the parent of a process-scoped track as the process track itself, so I would expect the merging behavior to derive from that.
Otherwise do process-scoped track in different processes share the same parents? And would the default merging behavior be imposed?

LalitMaganti: > Otherwise do process-scoped track in different processes share the same parents? And would the default merging behavior be imposed?

What happens in TP is that any child of the top level process track (i.e. a child of a track with "process_descriptor" field set) is automatically "promoted" to be a sibling of top level process track *but* inheriting the same process association.  So the way it works is as follows:
top level process scoped track in SDK -> process_track table with parent_id null and upid matching process 
child of top level process scoped track in SDK -> process_track table with parent_id null and upid matching parent track in SDK process
child of non-top level process scoped tracks -> process_track table with parent_id equal to parent and upid matching top level parent track in SDK process.

Does that make sense? If we need to go back and forth on this, let's maybe discuss a bit on chat.

PR #1554, opened 2025-05-16
Fix duration in android_kernel_wakelocks table.

simonmacm:
We accidentally had an extra `WHERE` clause that meant we were only counting awake time towards `dur`. Fix that, and expose `awake_dur` in case anyone actually wants that.


src/trace_processor/perfetto_sql/stdlib/android/kernel_wakelocks.sql
```diff
@@ -73,8 +75,6 @@ WITH
       sum(dur) AS dur,
       sum(iif(power_state = 'awake', dur, 0)) AS awake_dur
     FROM _android_kernel_wakelocks_joined
-    WHERE
-      power_state = 'awake'
     GROUP BY
       1,
```
aMayzner: Please use column names here, this is the preferred styleguide in stdlib.

aMayzner (approved):

Issue #1321, opened 2025-04-17
[Feature][UI] Improve UX for Entity State Residency visualization


cphlipot1:
A few users have expressed some confusion how the [entity state residency](url) tracks are visualized in the UI, and i would like to suggest a few ways to improve this.

<img width="1111" alt="Image" src="https://github.com/user-attachments/assets/d45a8bf3-9bd5-41db-b28d-bfa611e68cc1" />

## Track visualization

Currently the raw data in the track is the cumulative time in milliseconds in that state.

These tracks, like the power rail tracks, are configured to visualize in the UI with the y-axis "rate" mode.

However this causes the the values on the Y-axis to be displayed as (milliseconds in state)/(wallclock time in sec). The units of ms/sec isn't the most intuitive, and is not very obvious to users unless they already have an understanding of the low level representation of this state residency data and how Perfetto UI renders tracks in "rate" mode. 

I propose scaling the Y-axis values by a factor of .001 such that they are in the range 0.0 - 1.0 instead of 0 - 1000. Alternatively we could change this to % and display values 0-100%

## Aggregation Tab

For similar reasons as https://github.com/google/perfetto/issues/910 the UX on the aggregations tab could be improved.

instead of the default aggregations tab columns i recommend displaying:
- **Entity/state name** - (previously "Name")
- **Time in state** - time in sec that was spent in this state (previously "delta value")
- **% time in state** - the percent of wall clock time that spent in this state. (previously "rate /s")

cphlipot1:
> I propose scaling the Y-axis values by a factor of .001 such that they are in the range 0.0 - 1.0 instead of 0 - 1000. Alternatively we could change this to % and display values 0-100%

If we wanted to do something like this, I'm assuming we would be unable to change the values stored in the counter tables themselves without breaking backwards compatibility for clients that use trace processor SQL.

If that is the case, does the UI have any way to scale counter traccks independent of the values in SQL?

cphlipot1:
Also wanted to clarify i would be willing to make contributions to improve some of the above if not too involved.

LalitMaganti:
All makes sense I think this is largely just a matter of creating a new opinionated plugin for these tracks. 

> If that is the case, does the UI have any way to scale counter traccks independent of the values in SQL?

Any time where you want to make changes like this the process is as follows:
1) remove the group from the trace processor UI plugin which acts as a "last resort" solution to directly 1:1 render trace processor data in the UI
[...]

PR #1582, opened 2025-05-20
Add specialized counter tracks for entity state residencies

cphlipot1:
Add specialized tracks for entity state residency that accomplish a few things:
- state tracks are now grouped by Entity this cuts down on the label length for each track and avoids text from getting cut off and provides logical organization of the presented data.
-  yaxis is also now shared for states in each Entity instead of all states across all entities sharing a y-axis
- unit in the track is shifted to seconds to prevent the odd ms/s unit for rate tracks which was non-obvious and confusing
- plugin is enabled by default due to the removal of generic hadling in counter_tracks.ts to prevent the tracks from disappearing entirely.

Bug: https://github.com/google/perfetto/issues/1321

## Before

image

## After

image

Note that with the plugin enabled by default the Entity residency state aggregation will now be on by default as well:
image


ui/src/plugins/dev.perfetto.EntityStateResidency/index.ts
```diff
@@ -31,5 +37,74 @@ export default class implements PerfettoPlugin {
...
+            entity_name, state_name, track_id
+          ORDER BY
+            entity_name, state_name
+        `);
+
+    let entityResidencyGroup: TrackNode | undefined;
+    let currentGroup: TrackNode | undefined;
+    const it = result.iter({entity: STR, state: STR, trackId: NUM});
+    for (; it.valid(); it.next()) {
+      if (entityResidencyGroup == null) {
```
stevegolton: Nit: We usually just do if (!entityResidencyGroup) in situations like this.

ui/src/plugins/dev.perfetto.EntityStateResidency/index.ts
```diff
@@ -31,5 +37,74 @@ export default class implements PerfettoPlugin {
...
+        const powerGroup = ctx.plugins
+          .getPlugin(StandardGroupsPlugin)
+          .getOrCreateStandardGroup(ctx.workspace, 'POWER');
+        entityResidencyGroup = new TrackNode({title: 'Entity Residency'});
+        powerGroup.addChildInOrder(entityResidencyGroup);
+      }
+
+      // Create a track group for the current entity if it does not already
+      // exist.
+      if (currentGroup?.title != it.entity) {
```
stevegolton: Nit: `!==`

stevegolton (approved):
Looks good to me!

LalitMaganti (approved):

cphlipot1:
addressed nits

LalitMaganti (approved):

stevegolton (approved):

PR #1590, opened 2025-05-20
deobfuscation_module: Postpone deobfuscation to NotifyEndOfFile

ddiproietto:
Deobfuscation packets are now stored in DeobfuscationTracker and used at NotifyEndOfFile.

Bug: https://b.corp.google.com/issues/153632336


src/trace_processor/importers/proto/deobfuscation_tracker.cc
```diff
@@ -0,0 +1,14 @@
+#include "src/trace_processor/importers/proto/deobfuscation_tracker.h"
```
LalitMaganti: Here and below: missing copyrihgt headers.

LalitMaganti: This all feels complete overkill. Just inline this into deobfuscation module? lot of boilerpolate for very little gain.

ddiproietto: I had no idea I could keep this in the module. Done, thanks

src/trace_processor/importers/proto/deobfuscation_module.h
```diff
@@ -49,8 +54,9 @@ class DeobfuscationModule : public ProtoImporterModule {
       HeapGraphTracker*);
   void ParseDeobfuscationMappingForProfiles(
       const protos::pbzero::DeobfuscationMapping::Decoder&);
-  void ParseDeobfuscationMapping(protozero::ConstBytes);
+  void ParseDeobfuscationMapping(protozero::ConstBytes, HeapGraphTracker*);
 
+  std::deque<TraceBlob> packets_;
```
LalitMaganti: vector?

PR #1610, opened 2025-05-20
Return only accessible hardcoded atrace events.

KirillTim:
In this CL we filter out all the hardcoded atrace events that are not accessible.
We do it similar to the filtering out the vendor atrace events in `DiscoverAccessibleVendorTracepointsWithFile`.

Bug: b/417757659
Tested: `./out/Default_bin/perfetto_unittests '--gtest_filter=PredefinedTracepointsTest*'`
Tested:
Manually recorded a trace on linux with `tests/config/atrace.cfg`, using the following command:
`sudo out/Default_bin/tracebox -o /tmp/atrace_linux.ptrace --txt -c test/configs/atrace.cfg`
Atrace events are not accessible on linux, so I've checked that `ftrace_setup_errors`
metadata field in the collected trace is empty.


src/traced/probes/ftrace/ftrace_config_muxer.cc
```diff
@@ -662,13 +351,15 @@ FtraceConfigMuxer::FtraceConfigMuxer(
     ProtoTranslationTable* table,
     SyscallTable syscalls,
     std::map<std::string, std::vector<GroupAndName>> vendor_events,
+    std::map<std::string, std::set<GroupAndName>> predefined_events,
```
ddiproietto: Can we use base::FlatSet<GroupAndName> instead of std::set<GroupAndName> (here and everywhere), given that we expect a small number of events per category?

ddiproietto:
Thanks! LGTM

Can I also ask you to test that if you run a trace with a config like:

```
buffers {
  size_kb: 100024
  fill_policy: RING_BUFFER
}

data_sources {
  config {
    name: "linux.ftrace"
    target_buffer: 0
    ftrace_config {
      atrace_categories: "sched"
    }
  }
}

duration_ms: 10000
```

`/sys/kernel/tracing/events/sched/sched_switch/enable` is set to 1 if and only if the trace is running?

ddiproietto (approved):

PR #1612, opened 2025-05-20
tp/stdlib: Add Mali GPU power state counters

wusamuel6:
Parse mali_gpu_power_state trace event and create Mali GPU power state table for use in Perfetto's stdlib

Bug: 394608157
Test: tools/diff_test_trace_processor.py out/linux/trace_processor_shell --name-filter 'AndroidGpu:mali_gpu_power_state'


src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4177,4 +4181,17 @@ void FtraceParser::ParseHrtimerExpireExit(uint32_t cpu,
                                                        tracks::Dimensions(cpu));
   context_->slice_tracker->End(timestamp, track, hrtimer_id_);
 }
+
+constexpr auto kMaliGpuPowerStateBlueprint = tracks::CounterBlueprint(
```
LalitMaganti: Please place inside the `ParseMaliGpuPowerState` function.

LalitMaganti: THis should also be static when moved.

LalitMaganti (approved):

PR #1618, opened 2025-05-21
tp: fix UAF in statsd module

LalitMaganti:
It's not legal to hold onto pointers to descriptors as they can easily
be invalidated as more things are added to the descriptor pool. Instead,
hold onto the index which is stable.


src/trace_processor/importers/proto/statsd_module.cc
```diff
@@ -197,11 +193,12 @@ void StatsdModule::ParseAtom(int64_t ts, protozero::ConstBytes nested_bytes) {
       [&](ArgsTracker::BoundInserter* inserter) {
         ArgsParser delegate(ts, *inserter, *context_->storage);
 
-        const auto& fields = descriptor_->fields();
-        const auto& field_it = fields.find(nested_field_id);
+        const auto& descriptor =
+            context_->descriptor_pool_->descriptors()[*descriptor_idx_];
```
ddiproietto: are we guaranteed here that `descriptor_idx_` is not `nulllopt`?

LalitMaganti: Made it more clear.

ddiproietto (approved):

PR #1626, opened 2025-05-22
Decentralize minimap

stevegolton:
This patch adds a new minimap provider extension point allowing plugins to take control of what is displayed in the minimap area. Only one provider can take control at a time, so we use a priority number to define who 'wins'.

As a result, this patch moves the two minimap sources (sched and slices) out to their relevant plugins in order to make use of this new extension point.

This patch also updates the underlying mechanism behind the sched minimap calculation to use interval_intersect, for more accurate minimap bucketing.


ui/src/plugins/dev.perfetto.Tracks/index.ts
```diff
@@ -12,26 +12,27 @@
...
 
-/**
- * This plugin adds the generic aggregations for slice tracks and counter
- * tracks.
- */
 export default class implements PerfettoPlugin {
-  static readonly id = 'dev.perfetto.GenericAggregations';
+  static readonly id = 'dev.perfetto.Tracks';
+  static description =
+    'Generic bulk operations for tracks originating from the tracks table.';
```
LalitMaganti: Why is this not in the TraceProcessorTrack plugin in that case? That's basically the responsibility of that plugin as well?

LalitMaganti (approved):

ui/src/plugins/dev.perfetto.Tracks/index.ts
stevegolton: Good idea, done!

PR #1628, opened 2025-05-22
protos: tp: ui: add "correlation id" for coloring track events

LalitMaganti:
Have a way for users to define that "I want all of these things to have
the same color" without explicitly saying that in the trace. Instead
they define an implicit "correlation id" and we use that to make
decisions on which slices should be drawn the same color.


stevegolton (approved):
Looks good to me!

PR #1636, opened 2025-05-23
Add layer support in DatasetSliceTrack & BaseSliceTrack

stevegolton:
This allows tracks to separate slices onto different layers where each layer is mipmapped independently.

Very short but important slices which would normally be swamped by any longer but less important slices if they were on the same layer. By moving them onto separate layers, we can mipmap the separately making sure the more important slices show up on the track.

This patch also modifies the 'thread slice' track to take advantage of the layer functionality to ensure non-idle slices are displayed above idle ones.

### DIFF
--- a/ui/src/components/tracks/base_slice_track.ts
+++ b/ui/src/components/tracks/base_slice_track.ts
@@ -320,7 +320,7 @@ export abstract class BaseSliceTrack<
       (
         await this.engine.query(`
           SELECT
-            depth + 1 AS rowCount
+            ifnull(depth, 0) + 1 AS rowCount
           FROM (${this.getSqlSource()})
           ORDER BY depth DESC
           LIMIT 1
--- a/ui/src/components/tracks/base_slice_track.ts
+++ b/ui/src/components/tracks/base_slice_track.ts
@@ -51,12 +51,6 @@ const CHEVRON_WIDTH_PX = 10;
 const DEFAULT_SLICE_COLOR = UNEXPECTED_PINK;
 const INCOMPLETE_SLICE_WIDTH_PX = 20;
 
-// TODO(stevegolton): Large depth values really slow down mipmap performance so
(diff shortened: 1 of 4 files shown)


ui/src/components/tracks/dataset_slice_track.ts
```diff
@@ -92,6 +92,12 @@ export interface DatasetSliceTrackAttrs<T extends DatasetSchema> {
    *   width corresponds to the duration of the slice.
    * - `depth` (NUM): Depth of each event, used for vertical arrangement. Higher
    *   depth values are rendered lower down on the track.
+   * - `layer` (NUM): This layer value influences the mipmap function. Slices in
+   *   different layers will be mipmapped independency of each other, and the
+   *   buckets of higher layers will be rendered on top of lower layers.
+   *   Warning: While using layers, depth values must be kept under 16 (see
```
LalitMaganti: We should enforce this somehow (maybe by just getting the max depth from the dataset) otherwise this will silently fail and be impossible to debug.

stevegolton: Good plan. Done.

ui/src/components/tracks/base_slice_track.ts
```diff
@@ -314,6 +314,17 @@ export abstract class BaseSliceTrack<
     const result = await this.onInit();
     result && this.trash.use(result);
 
+    // Calc the number of rows based on the depth parameter.
+    const rowCount = assertExists(
+      (
+        await this.engine.query(`
+          SELECT
```
LalitMaganti, line 322: WOuld it be better to do the order by depth desc limit 1 trick?

stevegolton: Well spotted!

LalitMaganti (approved):

PR #1641, opened 2025-05-24
tp: speed up JSON tokenization/parsing by ~9x

LalitMaganti:
This CL moves away from using jsoncpp as much as possible in trace processor for performance reasons. Turns out jsoncpp really sucks very badly and with a few hundred lines of code, we can destroy it in benchmarks.

Before:
```
[898.916] processor_shell.cc:1878 Trace loaded: 453.80 MB in 22.06s (20.6 MB/s)
```

After:
```
[928.441] processor_shell.cc:1878 Trace loaded: 453.80 MB in 2.56s (177.1 MB/s)
```

There should be little-no functional change as a result of this CL. I've added a bunch of tests which pass both before and after to try and ensure there is no behavioural change. It's possible that there are some bugs of course but given the huge perf wins here, I think it's worth landing.


LalitMaganti:
+Steve as an FYI.

src/trace_processor/importers/common/parser_types.h
```diff
@@ -57,16 +60,74 @@ struct alignas(8) InlineSchedWaking {
 static_assert(sizeof(InlineSchedWaking) == 16);
 
 struct alignas(8) JsonEvent {
-  struct Begin {};
-  struct End {};
-  struct Scoped {
-    int64_t dur;
+  using StrOrUint64 = std::variant<std::monostate, StringPool::Id, uint64_t>;
```
aMayzner: Why do you need this?

aMayzner (approved):

PR #1645, opened 2025-05-27
PinSysUITracks: Cmd to pin main-thread slice track with good durations

kshitijmg:
- When debugging coroutine execution activity, we might want to use special tools that manipulate smali to get a flamegraph similar to method tracing
- This however isn't perfect and often yields "broken" slices whherein duration is -1 (non-ending slices). This CL adds a command to filter those out, giving us a more readable trace

Bug: 419472502
Test: Run command and ensure all slices on pinned track have valid duration in broken trace
Change-Id: I1182be16430a9d9fee08f01706332c86aa648550


kshitijmg:
Not needed anymore, closing

PR #1630, opened 2025-05-22
protos: tp: ui: add explicit enum for track_event merging

LalitMaganti:
This CL adds an explicit proto for deciding how track_event tracks should be merged together in the UI (and eventually the trace processor). Until now we've relied on a very implicit merging scheme but with this change, we can now properly define this and allow users to choose how they want merging to happen.

PR #1647, opened 2025-05-27
shared_lib: Add PERFETTO_TE_NESTED_TRACKS() param

ddiproietto:
This allows, emitting a PERFETTO_TE() event into a hierachy of multiple nested tracks.

For example

```
PERFETTO_TE(cat, PERFETTO_TE_INSTANT("event"),
            PERFETTO_TE_NESTED_TRACKS(
	        PERFETTO_TE_NESTED_TRACK_PROCESS(),
	        PERFETTO_TE_NESTED_TRACK_NAMED("parent", 2),
	        PERFETTO_TE_NESTED_TRACK_NAMED("child", 2)
	    ));
```

emits "event" in the hierarchy of tracks:

```
Current process
|
└──"parent"
        |
        └──"child"
```

Bug: https://b.corp.google.com/issues/415332664

### DIFF
--- a/examples/shared_lib/example_shlib_track_event.c
+++ b/examples/shared_lib/example_shlib_track_event.c
@@ -101,6 +101,12 @@ int main(void) {
                 PERFETTO_TE_DOUBLE_COUNTER(3.14),
                 PERFETTO_TE_REGISTERED_TRACK(&mycounter),
                 PERFETTO_TE_DYNAMIC_CATEGORY_STRING("physics"));
+    PERFETTO_TE(physics, PERFETTO_TE_COUNTER(),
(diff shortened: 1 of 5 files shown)


primiano (approved):
LGTM. Maybe quickly check with @LalitMaganti, he recently added some fields to determine how sibling tracks should be merged when there is nesting. does this play well with that change? (we don't have to support now, i would not change the scope of this cl. just wondering if adding those enum values he is adding in #1630 play well here)

ddiproietto:
> we don't have to support now, i would not change the scope of this cl. just wondering if adding those enum values he is adding in #1630 play well here

I think this should play well with #1630: we just need to add the new enum in the thread descriptor.

Thanks for the review!

PR #1648, opened 2025-05-27
base: Add RTMutex

primiano:
Add Priority-Inversion aware mutexes

- On Android/Linux: use a std::atomic + FUTEX_PI
- On mac/ios/qnx: use pthread_mutex with  PTHREAD_PRIO_INHERIT
- Other platforms: use std::mutex as a fallback (it still works on windows).
Also introduce a MaybeRtMutex which supports a runtime flag, which will be
wired up in future CLs.

Benchmarks:
On Android/arm64:
```
BM_RtMutex_NoContention<std::mutex>/real_time          131 ns          130 ns      5064449
BM_RtMutex_NoContention<RtFutex>/real_time             131 ns          130 ns      5388690
BM_RtMutex_NoContention<RtPosixMutex>/real_time        136 ns          136 ns      5147596
BM_RtMutex_Contention<std::mutex>/real_time     1000783447 ns       585409 ns            1 Iterations=16.8274M
BM_RtMutex_Contention<RtFutex>/real_time        1000857219 ns       530885 ns            1 Iterations=628.195k
BM_RtMutex_Contention<RtPosixMutex>/real_time   1001905193 ns       966307 ns            1 Iterations=89.063k
```

On Linux/x64:
```
BM_RtMutex_NoContention<std::mutex>/real_time          148 ns          148 ns      4742237
BM_RtMutex_NoContention<RtFutex>/real_time             160 ns          160 ns      4388775
BM_RtMutex_NoContention<RtPosixMutex>/real_time        165 ns          165 ns      4216096
BM_RtMutex_Contention<std::mutex>/real_time   1000416916 ns       186869 ns            1 Iterations=14.2615M
BM_RtMutex_Contention<RtFutex>/real_time      1000231574 ns       132401 ns            1 Iterations=180.89k
BM_RtMutex_Contention<RtPosixMutex>/real_time 1000264068 ns       151657 ns            1 Iterations=181.669k
```

Bug: b/394504259


ddiproietto:
Why can't we use pthread_mutex with PTHREAD_PRIO_INHERIT? We can set PTHREAD_PRIO_INHERIT only if the flag is enabled. On non-linux, sure, let's keep using std::mutex.

Benefits of using pthread_mutex:

* We don't have to deal with the futex syscall details.
* We don't have to deal with memory_order around the cmpxch.
* We don't have to keep an extra thread_local tid, pthread already has to do that for us.
* Less disruptive change: on android with libc++ (and on linux with libstdc++), std::mutex is a small wrapper around pthread_mutex already.
* In this direct futex implementation it's possible to call SetUseFutex() after a mutex has already been created, which can cause problems. In pthread, each mutex remembers if it needs to be a PTHREAD_PRIO_INHERIT.

Benefits of dealing with futex directly:
* lock() can be inline. (unlock() cannot fully be inlined, since you have to check the tid).
* when the flag is false, we're using std::mutex (but we're adding the flag check: I would argue it's the same change as going from libc++ std::mutex to pthread_mutex).

Any other reason I'm missing to prefer futex directly?

ddiproietto (approved):

PR #1652, opened 2025-05-27
Use interval intersect to slice slices to the exact range of the area selection when performing aggregations

stevegolton:
Use interval intersect to intersect slices by the exact area selection range when doing aggregations.

Any panels that use the area selection aggregator adapter (`createAggregationToTabAdaptor()`) will now be passed the pre-intersected dataset if the selected datasets implement a ts, dur and id columns.


LalitMaganti:
Amazing thanks so much for doing this!

I'd love if we also do counters so we can switch everything over atomically.

stevegolton:
> Amazing thanks so much for doing this!
> 
> I'd love if we also do counters so we can switch everything over atomically.

Discussed offline: Counter aggregations are a different beast and it's not really clear how interval_intersect can apply to them. Leaving for now.

PR #1653, opened 2025-05-28
Add bigtrace dependencies to be accesible via pip

edgararriagag:
This is adds necessary dependencies to start using Bigtrace when pip installing perfetto


python/setup.py
```diff
@@ -21,6 +26,8 @@
     keywords=['trace processor', 'tracing', 'perfetto'],
     install_requires=[
         'protobuf',
+        'grpcio',
```
LalitMaganti: These should be optional dependencies, not required.

python/setup.py
```diff
@@ -8,8 +8,13 @@
         'perfetto.common',
         'perfetto.trace_processor',
         'perfetto.trace_uri_resolver',
+        'perfetto.bigtrace',
+        'perfetto.bigtrace.protos',
     ],
-    package_data={'perfetto.trace_processor': ['*.descriptor']},
+    package_data={
+        'perfetto.trace_processor': ['*.descriptor'],
+        'perfetto.bigtrace.protos': ['**/*']
```
LalitMaganti: This should be *.descriptor as well afaik.

edgararriagag: This is the error that happens when that is removed.

```
.../python/venv-ea-perfetto/lib/python3.12/site-packages/perfetto/bigtrace/api.py", line 21, in <module>
    from perfetto.bigtrace.protos.perfetto.bigtrace.orchestrator_pb2 import BigtraceQueryArgs
ModuleNotFoundError: No module named 'perfetto.bigtrace.protos'
```

*.descriptor did not help, as none of the files are of that type, the files types are *.py and *.pyi, I modified to specifically list the file types that are being required instead of the general wildcard, but I can put the wildcard back again if you prefer so.

python/setup.py
edgararriagag: Done

PR #1663, opened 2025-05-28
Update Winscope stats message type

PabiGamito:
All these errors are caused by or effectively lead to data loss in the trace tables and should be reported as such instead of just info.

### DIFF
--- a/src/trace_processor/storage/stats.h
+++ b/src/trace_processor/storage/stats.h
@@ -406,58 +406,58 @@ namespace perfetto::trace_processor::stats {
   F(v8_code_load_missing_code_range,      kSingle,  kError,    kAnalysis,      \
       "V8 load had no code range or an empty one. Event ignored."),            \
   F(winscope_inputmethod_clients_parse_errors,                                 \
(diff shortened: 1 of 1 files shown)


src/trace_processor/storage/stats.h
```diff
@@ -406,58 +406,58 @@ namespace perfetto::trace_processor::stats {
   F(v8_code_load_missing_code_range,      kSingle,  kError,    kAnalysis,      \
       "V8 load had no code range or an empty one. Event ignored."),            \
   F(winscope_inputmethod_clients_parse_errors,                                 \
-                                          kSingle,  kInfo,     kAnalysis,      \
+                                          kSingle,  kDataLoss, kAnalysis,      \
```
LalitMaganti: kDataLoss means the data is not in the trace because it could not be captured. Reading the description, it feels like all of these are errors not losses.

PabiGamito: Ah ok, then yeah, should be kError. Will update.

LalitMaganti (approved):

PR #1670, opened 2025-05-28
Dev/wusamuel/UI wattson gpu aggregation

wusamuel6:
ui: Add estimate aggregator for GPU

Similar to CPU, add an estimate aggregation for the GPU when user draws
a window over the GPU estimation track. Clean up some Wattson UI code.

Bug: 419369871

### DIFF
--- a/ui/src/plugins/org.kernel.Wattson/index.ts
+++ b/ui/src/plugins/org.kernel.Wattson/index.ts
@@ -37,38 +37,18 @@ export default class implements PerfettoPlugin {
   static readonly id = `org.kernel.Wattson`;
 
   async onTraceLoad(ctx: Trace): Promise<void> {
+    const markersSupported = await hasWattsonMarkersSupport(ctx.engine);
     const cpuSupported = await hasWattsonCpuSupport(ctx.engine);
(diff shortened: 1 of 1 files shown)


ui/src/plugins/org.kernel.Wattson/index.ts
```diff
@@ -37,38 +37,18 @@ export default class implements PerfettoPlugin {
   static readonly id = `org.kernel.Wattson`;
 
   async onTraceLoad(ctx: Trace): Promise<void> {
+    const markersSupported = await hasWattsonMarkersSupport(ctx.engine);
     const cpuSupported = await hasWattsonCpuSupport(ctx.engine);
     const gpuSupported = await hasWattsonGpuSupport(ctx.engine);
+
     // Short circuit if Wattson is not supported for this Perfetto trace
-    if (!(cpuSupported || gpuSupported)) return;
+    if (!(markersSupported || cpuSupported || gpuSupported)) return;
```
aMayzner, line 45: nit: `NOT(X || Y || Z)` is strictly less readable to me than `NOT(X) AND NOT(Y) AND NOT(Z)`, but maybe it's just me

PR #1692, opened 2025-05-30
tp: ui: py_api: Remove the metric_ids requirement

aMayzner:
Bug: 418942995

Trace Summary: If computation.metric_ids are not specified execute all metrics

### DIFF
--- a/src/trace_processor/trace_summary/summary.cc
+++ b/src/trace_processor/trace_summary/summary.cc
@@ -358,7 +358,9 @@ base::Status Summarize(TraceProcessor* processor,
               "Metric with empty id field: this is not allowed");
         }
         if (base::CaseInsensitiveEqual(id, "all")) {
-          return base::ErrStatus("Metric with id 'all' is not allowed: ");
+          return base::ErrStatus(
+              "Metric with `id` field value `all` is not allowed. Please "
+              "change the value of the `id` field of the metric spec.");
         }
         if (queries_per_metric.Find(id)) {
           return base::ErrStatus(


include/perfetto/trace_processor/basic_types.h
```diff
@@ -312,6 +312,7 @@ struct SqlPackage {
 struct TraceSummaryComputationSpec {
   // The set of metric ids which should be computed and returned in the
   // `TraceSummary` proto.
+  // If the vector is empty, all metrics will be computed.
   std::vector<std::string> v2_metric_ids;
```
LalitMaganti: This should be changed to an std::optional first. The semantic should be:
1) If nulopt, all metrics should be computed
2) If present empty, no metrics should be computed
3) If present and non-empty, specified metrics should be computed.

protos/perfetto/trace_processor/trace_processor.proto
```diff
@@ -394,6 +394,7 @@ message AnalyzeStructuredQueryResult {
 message TraceSummaryArgs {
   message ComputationSpec {
     repeated string metric_ids = 1;
+    optional bool dont_execute_metrics = 3;
```
LalitMaganti: I really don't like this name. Can we call it `run_all_metrics` or similar.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -346,8 +347,42 @@ base::Status Summarize(TraceProcessor* processor,
...
+    for (const auto& spec : spec_decoders) {
+      for (auto it = spec.metric_spec(); it; ++it) {
+        protos::pbzero::TraceMetricV2Spec::Decoder m(*it);
+        std::string id = m.id().ToStdString();
+        if (id.empty()) {
+          return base::ErrStatus(
+              "Metric with empty id field: this is not allowed");
+        }
+        if (base::CaseInsensitiveEqual(id, "all")) {
+          return base::ErrStatus("Metric with id 'all' is not allowed: ");
```
LalitMaganti: Can you explain what went wrong here and how the user can fix it.

Ideally this would also be factored out into a funcciton so this can grow tith time.

protos/perfetto/trace_processor/trace_processor.proto
aMayzner: Do you think we should guard against providing metric_ids and `run_all_metrics`?

LalitMaganti: Yes we should return an error.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -385,6 +422,7 @@ base::Status Summarize(TraceProcessor* processor,
       }
     }
   }
+  // If `v2_metric_ids` is an empty vector, we will not compute any metrics.
```
LalitMaganti, line 425: Feels out of place?

LalitMaganti (approved):

src/trace_processor/trace_summary/summary.cc
aMayzner: It is the last `if` condition, there is just nothing to do in this situation

PR #1697, opened 2025-05-30
Handle CREATED, DEAD, and DESTROYED events in GenericTaskStateEvents

jahdiel-alvarez:
The thread table is populated with a start_ts when a CREATED event is received for a non-existing thread. If any other state is received for a non-existing thread, then the start_ts is left NULL (since we don't know when the thread was actually created).

Now, threads are only ended when a DEAD event is received. The DESTROYED event is only considered if its received before a DEAD event. If no utid exists associated to a DESTROYED event, then the event is simply ignored.

Also, we introduce a new error stat: generic_task_state_invalid_order, to cover error cases in which events are received in an inproper sequence. For example, the error stat is incremented if a CREATED event is received for an already existing thread or if a DEAD event is received after a DESTROYED event for the same thread.

Bug: 420976270

### DIFF
--- a/src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
+++ b/src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
@@ -56,7 +56,10 @@ void GenericKernelParser::RemovePendingStateInfoForTid(UniqueTid utid) {
 
 GenericKernelParser::GenericKernelParser(TraceProcessorContext* context)
     : context_(context),
+      created_string_id_(context_->storage->InternString("Created")),
       running_string_id_(context_->storage->InternString("Running")),
+      dead_string_id_(context_->storage->InternString("Z")),
+      destroyed_string_id_(context_->storage->InternString("X")),
       task_states_({
           context_->storage->InternString("Unknown"),
           context_->storage->InternString("Created"),
@@ -78,16 +81,40 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
   const uint32_t cpu = static_cast<uint32_t>(task_event.cpu());
   const uint32_t tid = static_cast<uint32_t>(task_event.tid());
   const int32_t prio = task_event.prio();
+  const size_t state = static_cast<size_t>(task_event.state());
 
-  // TODO(jahdiel): Handle the TASK_STATE_CREATED in order to set
-  // the thread's creation timestamp.
-  UniqueTid utid = context_->process_tracker->UpdateThreadName(
-      tid, comm_id, ThreadNamePriority::kGenericKernelTask);
-
-  StringId state_string_id =
-      task_states_[static_cast<size_t>(task_event.state())];
-  if (state_string_id == kNullStringId) {
+  if (state > task_states_.size()) {
     context_->storage->IncrementStats(stats::task_state_invalid);
+  }
+
+  StringId state_string_id = task_states_[state];
+
+  // Handle thread creation
+  auto utid_opt = context_->process_tracker->GetThreadOrNull(tid);
+  if (!utid_opt) {
+    if (state_string_id == destroyed_string_id_) {
+      // Ignore a DESTROYED event for a non-existing thread.
+      return;
+    }
+    utid_opt = context_->process_tracker->StartNewThread(
+        state_string_id == created_string_id_ ? std::optional{ts}
+                                              : std::nullopt,
+        tid);
+    context_->process_tracker->UpdateThreadNameByUtid(
+        *utid_opt, comm_id, ThreadNamePriority::kGenericKernelTask);
+  } else if (state_string_id == created_string_id_) {
+    context_->storage->IncrementStats(stats::generic_task_state_invalid_order);
+  }
+
+  UniqueTid utid = *utid_opt;
+
+  if (state_string_id == dead_string_id_) {
+    if (ThreadStateTracker::GetOrCreate(context_)->GetPrevEndState(utid) ==
+        destroyed_string_id_) {
+      context_->storage->IncrementStats(
+          stats::generic_task_state_invalid_order);
+    }
+    context_->process_tracker->EndThread(ts, tid);
   }
 
   // Given |PushSchedSwitch| updates the pending slice, run this
--- a/src/trace_processor/importers/common/thread_state_tracker.cc
+++ b/src/trace_processor/importers/common/thread_state_tracker.cc
@@ -240,6 +240,15 @@ void ThreadStateTracker::UpdatePendingState(
   }
 }
 
+StringId ThreadStateTracker::GetPrevEndState(UniqueTid utid) {
+  if (!HasPreviousRowNumbersForUtid(utid))
+    return kNullStringId;
+
+  auto row_ref = RowNumToRef(prev_row_numbers_for_thread_[utid]->last_row);
+
+  return row_ref.state();
+}
+
 bool ThreadStateTracker::IsRunning(StringId state) {
   return state == running_string_id_;
 }
--- a/src/trace_processor/importers/common/thread_state_tracker.h
+++ b/src/trace_processor/importers/common/thread_state_tracker.h
@@ -81,6 +81,7 @@ class ThreadStateTracker : public Destructible {
   //
   // - PushThreadState
   // - UpdatePendingState
+  // - GetPrevEndState
   //
   // To update the thread state track accordingly. Updating pending state is
   // necessary in this scenario because single thread state change events don't
@@ -97,6 +98,8 @@ class ThreadStateTracker : public Destructible {
                           std::optional<uint16_t> cpu = std::nullopt,
                           std::optional<UniqueTid> waker_utid = std::nullopt,
                           std::optional<uint16_t> common_flags = std::nullopt);
+
+  StringId GetPrevEndState(UniqueTid utid);
 
  private:
   void AddOpenState(int64_t ts,
--- a/src/trace_processor/importers/generic_kernel/generic_kernel_parser.h
+++ b/src/trace_processor/importers/generic_kernel/generic_kernel_parser.h
@@ -77,7 +77,10 @@ class GenericKernelParser {
   std::vector<std::optional<SchedEventState::PendingSchedInfo>>
       pending_state_per_utid_;
 
+  StringId created_string_id_;
   StringId running_string_id_;
+  StringId dead_string_id_;
+  StringId destroyed_string_id_;
   const std::vector<StringId> task_states_;
 };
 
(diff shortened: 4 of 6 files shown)


LalitMaganti (approved):

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -78,18 +81,42 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
...
-      task_states_[static_cast<size_t>(task_event.state())];
-  if (state_string_id == kNullStringId) {
+  if (state > task_states_.size()) {
     context_->storage->IncrementStats(stats::task_state_invalid);
   }
 
+  StringId state_string_id = task_states_[state];
+
+  // Handle thread creation
+  auto utid_opt = context_->process_tracker->GetThreadOrNull(tid);
```
LalitMaganti: I think you actually want something different here:
1) if you are `created_string_id_`, you should call StartNewThread
2) If you are `destroyed_string_id_` you should call GetThreadOrNull
3) Otherwise you should call GetOrCreateThread.

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -78,18 +81,42 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
...
 
-  // TODO(jahdiel): Handle the TASK_STATE_CREATED in order to set
-  // the thread's creation timestamp.
-  UniqueTid utid = context_->process_tracker->UpdateThreadName(
-      tid, comm_id, ThreadNamePriority::kGenericKernelTask);
-
-  StringId state_string_id =
-      task_states_[static_cast<size_t>(task_event.state())];
-  if (state_string_id == kNullStringId) {
+  if (state > task_states_.size()) {
```
LalitMaganti: >= I think.

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -78,18 +81,42 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
...
     context_->storage->IncrementStats(stats::task_state_invalid);
```
LalitMaganti: You should also return here in this case.

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -78,18 +81,42 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
...
+  } else if (state_string_id == created_string_id_) {
+    context_->storage->IncrementStats(stats::generic_task_state_invalid_order);
+  }
+
+  UniqueTid utid = *utid_opt;
+
+  if (state_string_id == dead_string_id_) {
+    if (ThreadStateTracker::GetOrCreate(context_)->GetPrevEndState(utid) ==
+        destroyed_string_id_) {
+      context_->storage->IncrementStats(
```
LalitMaganti: SHouldn't you return?

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -78,18 +81,42 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
...
+      return;
+    }
+    utid_opt = context_->process_tracker->StartNewThread(
+        state_string_id == created_string_id_ ? std::optional{ts}
+                                              : std::nullopt,
+        tid);
+    context_->process_tracker->UpdateThreadNameByUtid(
+        *utid_opt, comm_id, ThreadNamePriority::kGenericKernelTask);
+  } else if (state_string_id == created_string_id_) {
+    context_->storage->IncrementStats(stats::generic_task_state_invalid_order);
```
LalitMaganti: Return?

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
jahdiel-alvarez: So I think this is already covered.
1. If `created_string_id_` then we already use StartNewThread (line 99).
2. We only care about `destroyed_string_id_` if the thread exists (covered by line 93).
3. GetOrCreateThread simply calls StartNewThread without a ts, which we already (lines 100-101).

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -83,40 +85,67 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
   const int32_t prio = task_event.prio();
   const size_t state = static_cast<size_t>(task_event.state());
 
-  if (state > task_states_.size()) {
+  if (state == 0 || state >= task_states_.size()) {
```
LalitMaganti: You can make this part of the switch below.

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -83,40 +85,67 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
...
     }
-    utid_opt = context_->process_tracker->StartNewThread(
-        state_string_id == created_string_id_ ? std::optional{ts}
-                                              : std::nullopt,
-        tid);
-    context_->process_tracker->UpdateThreadNameByUtid(
-        *utid_opt, comm_id, ThreadNamePriority::kGenericKernelTask);
-  } else if (state_string_id == created_string_id_) {
-    context_->storage->IncrementStats(stats::generic_task_state_invalid_order);
   }
```
LalitMaganti, line 93: I'd refactor this switch into a function. I think if you make a function which is std::optional<UniqueTid> GetThreadForState(State), you'll find this becomes a lot more readable :)

jahdiel-alvarez: In terms of readability, I think the issue is the switch itself. It adds "noise" to the code (no need to list out all the possible states). Also, it forces us to duplicate code (see the `if (!utid_opt)` and `UpdateThreadName`). I don't see how moving the switch into a function addresses this. 

On the other hand, refactoring this out into a function adds additional complexity (less readable) to the API of the new method. For example, when handling the `TASK_STATE_DESTROYED` now we can simply return, which is very straightforward. In the new method, we would need to return an empty optional, which doesn't obviously mean stop executing on the caller side. These extra semantics I believe make the code less clean. I would argue, we roll back to the initial approach. As we continue to develop this method and the complexity increases, then refactoring out might be the best option.

LalitMaganti: Here's how you can write a clean function:
```
std::optional<UniqueTid> GetUtidForState(uint32_t tid, TaskStateEnum state, StringId comm_id) {
  switch (state) {
    case TaskStateEnum::TASK_STATE_CREATED: {
      if (context_->process_tracker->GetThreadOrNull(tid)) {
        context_->storage->IncrementStats(
            stats::generic_task_state_invalid_order);
        return std::nullopt;
      }
      UniqueTid utid = context_->process_tracker->StartNewThread(ts, tid);
      context_->process_tracker->UpdateThreadNameByUtid(
            utid, comm_id, ThreadNamePriority::kGenericKernelTask);
      return utid;
    }
    case TaskStateEnum::TASK_STATE_DESTROYED:
      return context_->process_tracker->GetThreadOrNull(tid);
    case TaskStateEnum::TASK_STATE_DEAD: {
      UniqueTid utid = context_->process_tracker->GetOrCreateThread(tid);
      if (ThreadStateTracker::GetOrCreate(context_)->GetPrevEndState(utid) == destroyed_string_id_) {
        context_->storage->IncrementStats(stats::generic_task_state_invalid_order);
        return std::nullopt;
      }
      context_->process_tracker->UpdateThreadNameByUtid(
            utid, comm_id, ThreadNamePriority::kGenericKernelTask);
      context_->process_tracker->EndThread(ts, tid);
      return utid;
    }
    case TaskStateEnum::TASK_STATE_RUNNABLE:
    case TaskStateEnum::TASK_STATE_RUNNING:
    case TaskStateEnum::TASK_STATE_INTERRUPTIBLE_SLEEP:
    case TaskStateEnum::TASK_STATE_UNINTERRUPTIBLE_SLEEP:
    case TaskStateEnum::TASK_STATE_STOPPED: {
      UniqueTid utid = context_->process_tracker->GetOrCreateThread(tid);
      context_->process_tracker->UpdateThreadNameByUtid(
            utid, comm_id, ThreadNamePriority::kGenericKernelTask);
      return utid;
    }
    case TaskStateEnum::TASK_STATE_UNKNOWN:
      return std::nullopt;
  }
}
```

LalitMaganti: There's a very specific reason I want this switch: it lets me see very explicitly what is happening for each case which I can match up 1:1 with Linux. It really helps with seeing how it's consistent/inconsistent.

For example, I noticed here that it appears that you are creating threads on `TASK_STATE_DEAD` even if the thread doesn't exist: we don't do this on Linux (tbh we don't track DEAD at all, just destroyed). i would have not seen this unless I wrote t out like this.

I don't see the extra verbosity here as noise but as very valuable signal. Being explicit here is a good thing.

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -128,6 +133,57 @@ void GenericKernelParser::ParseGenericTaskStateEvent(
...
+      if (!utid_opt) {
+        utid_opt = context_->process_tracker->UpdateThreadName(
+            tid, comm_id, ThreadNamePriority::kGenericKernelTask);
+      } else if (ThreadStateTracker::GetOrCreate(context_)->GetPrevEndState(
+                     *utid_opt) == destroyed_string_id_) {
+        context_->storage->IncrementStats(
+            stats::generic_task_state_invalid_order);
+        utid_opt = std::nullopt;
+      }
+      context_->process_tracker->EndThread(ts, tid);
```
LalitMaganti, line 166: You don't want to run this on the destroyed state imo.

jahdiel-alvarez: I think if we see DEAD, then we should end the thread (even if prev state was DESTROYED) because of TID reuse. The kernel told us the thread was DEAD so the TID can be reused and we don't want to associate a new thread with the previously problematic one.

LalitMaganti: Ack.

Issue #408, opened 2022-12-18
run python api report error


wjj1928:
I just copied the example from official website. I am running it on ubuntu16. 

python code:
#!/usr/bin/env python3

from perfetto.trace_processor import TraceProcessor

fname="trace.bin"
tp = TraceProcessor(trace=fname)

error message:

Traceback (most recent call last):
  File "/usr/lib/python3.8/urllib/request.py", line 1354, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/lib/python3.8/http/client.py", line 1256, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1302, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1251, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1011, in _send_output
    self.send(msg)
  File "/usr/lib/python3.8/http/client.py", line 951, in send
    self.connect()
  File "/usr/lib/python3.8/http/client.py", line 1425, in connect
    self.sock = self._context.wrap_socket(self.sock,
  File "/usr/lib/python3.8/ssl.py", line 500, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/lib/python3.8/ssl.py", line 1040, in _create
    self.do_handshake()
  File "/usr/lib/python3.8/ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1131)

During handling of the above exception, another exception occurred:

I have installed perfetto with command: pip3 install perfetto
What's wrong with my environment?

LalitMaganti:
So two things:
1) Ubuntu 16 is now ancient and we've never tested on anything nearly that old.
2) It seems like your environment is connecting to a https url instead of HTTP which we use by default. No clue why it's doing that but we use the Python API extensively so it's definitely an environment issue not a library issue.

wjj1928:
I try ubuntu20, results the same! and Just checked the source, I saw manifest file[https://github.com/google/perfetto/blob/master/python/perfetto/prebuilts/manifests/trace_processor_shell.py]()  use https instead of http:

```
    'url':
[...]

PR #1708, opened 2025-06-02
Secure trace_processor download with SSL certificate verification

stmazoni:
This pull request addresses a potential security vulnerability in the `trace_processor` download process, and prevent SSLCertVerificationError in the Mac OS environment.

Related bug: https://github.com/google/perfetto/issues/408

### DIFF
--- a/python/perfetto/trace_processor/platform.py
+++ b/python/perfetto/trace_processor/platform.py
@@ -21,6 +21,8 @@ import socket
 import stat
 import subprocess
 import tempfile
+import ssl
+import certifi
 from typing import Tuple
 from urllib import request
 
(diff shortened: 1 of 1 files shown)


python/perfetto/trace_processor/platform.py
```diff
@@ -21,6 +21,8 @@
 import stat
 import subprocess
 import tempfile
+import ssl
+import certifi
```
LalitMaganti, line 25: This is not part of the standard library of Python so we cannot assume this package will be present.

LalitMaganti (commented):
Unfortunately I don't think we can take this patch in the current state as it relies on a library which is not part of the standard library.

stmazoni:
I understand, thank you for checking this. I will close this request.

PR #1711, opened 2025-06-02
tp: py_api: Provide TraceSummarySpec as bytes

aMayzner:
Extend tp python API to also support bytes (proto) alongside textproto

### DIFF
--- a/python/perfetto/trace_processor/api.py
+++ b/python/perfetto/trace_processor/api.py
@@ -175,7 +175,8 @@ class TraceProcessor:
     an error.
 
     Args:
-      specs: A list of textproto/bytes specs to be used for the summary
+      specs: A list of textproto (as str) or binary proto (as bytes) specs to be
+        used for the summary
       metric_ids: Optional list of metric IDs as defined in TraceMetrics. 
         If `None` all metrics will be executed.
       metadata_query_id: Optional query ID for metadata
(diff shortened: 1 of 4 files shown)


python/perfetto/trace_processor/api.py
```diff
@@ -167,16 +167,17 @@ def metric(self, metrics: List[str]):
     return metrics
 
   def trace_summary(self,
-                    specs: List[str],
+                    specs: Union[List[str], List[bytes]],
```
LalitMaganti: AFAIK the trace processor C++ api supports *both* textproto and proto. Is there any reason we cannot do the same by doing `List[Union[str, bytes]]`

aMayzner: No technical reason - this is the current implementation, I thought it would be cleaner this way, choosing between one or the other

LalitMaganti: I'd prefer to keep it consistent. We should document it as well.

python/perfetto/trace_processor/api.py
```diff
@@ -167,16 +167,17 @@ def metric(self, metrics: List[str]):
...
                     metric_ids: Optional[List[str]] = None,
                     metadata_query_id: Optional[str] = None):
     """Returns the trace summary data corresponding to the passed in metric
     IDs and specs. Raises TraceProcessorException if the response returns with
     an error.
 
     Args:
-      metric_ids: A list of metric IDs as defined in TraceMetrics
-      specs: A list of textproto specs to be used for the summary
+      specs: A list of textproto/bytes specs to be used for the summary
```
LalitMaganti: I'd explicitly comment here that str will be considered as textproto and bytes as a binary proto.

PR #1720, opened 2025-06-03
tp: fix parsing of non-object/array JSON values in proto traces

LalitMaganti:
Turns out the AddJsonValueToArgs function can be called with strings and
literals, not just objects and arrays like I thought. Handle this case
by trying to parse the value directly as a fallback if the
array/iterator approach fails.


victorvianna:
CC @victorvianna

aMayzner (approved):

PR #1729, opened 2025-06-04
tp: introduce specialized storage for "id-like" columns

LalitMaganti:
If a column fits all the criteria for the bitvector + popcount data
structure, then create one to sit alongside and use it for equality
constraints.


.gitignore
```diff
@@ -16,6 +16,7 @@
 .idea
 .project
 .recipe_deps/
```
aMayzner, line 18: Did you want to add it?

LalitMaganti: Yes.

PR #1730, opened 2025-06-04
svg group_key fix

zezeozue:
exporter: Fix group_key name when exporting svgs

Now, the group_key column name used will be returned along with the svgs.

Test: tools/diff_test_trace_processor.py out/android/trace_processor_shell --name-filter '.*svg.*'


LalitMaganti:
Can we close this?

PR #1731, opened 2025-06-04
Bump priority of thread state aggregation tab to show up before slices

stevegolton:
When selecting thread state tracks, the slice aggregation tab showed up before the thread state aggregation tab, which is usually not what we want.

Before:

After:
image


stevegolton:
Slice aggregation no longer shows up when (only) thread state tracks are selected due to a dataset change, so this is a non issue.

PR #1732, opened 2025-06-04
do-not-land: Repro hardened libcxx issue

victorvianna:
1) Reapply Lalit's fixes
2) Build with hardened libcxx


victorvianna:
The last commit from #1728 finally fixes the issue, closing

PR #1739, opened 2025-06-05
tp: StructuredQuery.sql supports multi select SQL field

aMayzner:
Before we had to have preamble to state what queries were needed before executing the statement from the `sql`  field. Now we tokenize the `sql` field into statements and push all except the last one into the `preamble` (and validate that they don't produce any results) and push only the last one into the final query.

### DIFF
--- a/protos/perfetto/perfetto_sql/structured_query.proto
+++ b/protos/perfetto/perfetto_sql/structured_query.proto
@@ -90,17 +90,17 @@ message PerfettoSqlStructuredQuery {
   message Sql {
     // The SQL string. Required.
     //
-    // Note: this string is currently required to be a single well-formed select
-    // statement. This means you cannot use CREATE PERFETTO TABLE/VIEW but *can*
-    // use WITH clauses, UNION and INTERSECT. This restriction may be dropped
-    // in the future.
+    // Note: We expect the last statement to be the only one to produce a
+    // result.
     optional string sql = 1;
 
     // The name of columns which will be returned by the SQL. Required.
     repeated string column_names = 2;
 
-    // SQL string that has to be run before running the SQL. Supports multi
-    // statement queries. Optional.
+    // DEPRECATED, as `sql` field supports multi-statement queries.
+    //
+    // SQL string that has to be run before running the SQL.
+    // Supports multi statement queries. Optional.
     optional string preamble = 3;
   }
 
--- a/protos/perfetto/perfetto_sql/structured_query.proto
+++ b/protos/perfetto/perfetto_sql/structured_query.proto
@@ -90,8 +90,13 @@ message PerfettoSqlStructuredQuery {
   message Sql {
     // The SQL string. Required.
     //
-    // Note: We expect the last statement to be the only one to produce a
-    // result.
+    // `sql` can contain multiple, semi-colon separated statements but must
+    // adhere to the following principles:
+    // 1) Only the final statement can return results (i.e. be a `SELECT`
+    // statement): every other statement must be a statement returning no output
+    // (e.g. INCLUDE PERFETTO MODULE, CREATE PERFETTO TABLE etc.).
+    // 2) The final statement *must* be a valid `SELECT` statement returning
+    // results with at least one column.
     optional string sql = 1;
 
     // The name of columns which will be returned by the SQL. Required.
--- a/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
+++ b/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
@@ -34,6 +34,7 @@
 #include "perfetto/protozero/field.h"
 #include "perfetto/protozero/proto_decoder.h"
 #include "protos/perfetto/perfetto_sql/structured_query.pbzero.h"
+#include "src/trace_processor/perfetto_sql/tokenizer/sqlite_tokenizer.h"
 #include "src/trace_processor/util/status_macros.h"
 
 namespace perfetto::trace_processor::perfetto_sql::generator {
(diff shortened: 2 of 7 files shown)


protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -90,17 +90,17 @@ message PerfettoSqlStructuredQuery {
   message Sql {
     // The SQL string. Required.
     //
-    // Note: this string is currently required to be a single well-formed select
-    // statement. This means you cannot use CREATE PERFETTO TABLE/VIEW but *can*
-    // use WITH clauses, UNION and INTERSECT. This restriction may be dropped
-    // in the future.
+    // Note: We expect the last statement to be the only one to produce a
+    // result.
```
LalitMaganti: This comment should be expanded. Something like:
```
`sql` can contain multiple, semi-colon separated statements but must adhere to the following principles:
1) Only the final statement can return results (i.e. be a `SELECT` statement): every other statement must be a statement returning no output (e.g. INCLUDE PERFETTO MODULE, CREATE PERFETTO TABLE etc.).
2) The final statement *must* be a valid `SELECT` statement returning results with at least one column.
```

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -48,6 +49,75 @@ enum QueryType : uint8_t {
...
+std::pair<std::string, std::string> GetPreambleAndSql(const std::string& sql) {
+  std::string preamble;
+  std::string current_statement_str;
+
+  if (sql.empty()) {
+    return {preamble, current_statement_str};
+  }
+
+  auto tokenizer =
+      SqliteTokenizer(SqlSource::FromTraceProcessorImplementation(sql));
```
LalitMaganti: SqliteTokenizer tokenizer(SqlSource::FromTraceProcessorImplementation(sql));

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -48,6 +49,75 @@ enum QueryType : uint8_t {
   kNested,
 };
 
+std::pair<std::string, std::string> GetPreambleAndSql(const std::string& sql) {
```
LalitMaganti: This code needs non-trivial unit tests.

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -48,6 +49,75 @@ enum QueryType : uint8_t {
...
+      current_tok = tokenizer.Next();
+    }
+
+    // Extract the current statement.
+    // If current_tok is TK_SEMI, we want to exclude it from the statement.
+    // If current_tok is terminal (end of input), we want to include the last
+    // token.
+    SqlSource current_statement_sql_source = tokenizer.Substr(
+        statement_begin_tok, prev_tok, SqliteTokenizer::EndToken::kInclusive);
+    std::string next_statement_candidate = current_statement_sql_source.sql();
```
LalitMaganti: You need to account for empty statements (i.e. multiple semi-colons back to back.

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -48,6 +49,75 @@ enum QueryType : uint8_t {
...
+  if (statement_begin_tok.IsTerminal()) {
+    return {preamble, current_statement_str};
+  }
+
+  while (true) {
+    SqliteTokenizer::Token current_tok = statement_begin_tok;
+    SqliteTokenizer::Token prev_tok = statement_begin_tok;
+    while (current_tok.token_type != TK_SEMI && !current_tok.IsTerminal()) {
+      prev_tok = current_tok;
+      current_tok = tokenizer.Next();
```
LalitMaganti: Shouldn't this be `NextNonWhitespace()`?

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
LalitMaganti: Both of these should return SqlSource *not* std::string. In general in trace processor you should avoid as much as possible leaving the world of "SqlSource".

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -48,6 +49,75 @@ enum QueryType : uint8_t {
...
+
+    // Extract the current statement.
+    // If current_tok is TK_SEMI, we want to exclude it from the statement.
+    // If current_tok is terminal (end of input), we want to include the last
+    // token.
+    SqlSource current_statement_sql_source = tokenizer.Substr(
+        statement_begin_tok, prev_tok, SqliteTokenizer::EndToken::kInclusive);
+    std::string next_statement_candidate = current_statement_sql_source.sql();
+
+    if (!current_statement_str.empty()) {
```
LalitMaganti: This logic is really hard to follow. In my head, this can be much simpler:
1) Keep looping through terminal tokens, keeping track of the last and second last tokens
2) When you hit the EOF, you should Substr into two: one for everything except the last statement, one for the last statement.

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -48,6 +50,64 @@ enum QueryType : uint8_t {
   kNested,
 };
 
+std::pair<SqlSource, SqlSource> GetPreambleAndSql(const std::string& sql) {
+  PERFETTO_ELOG("Parse SQL: %s", sql.c_str());
```
LalitMaganti: Remove?

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -48,6 +50,64 @@ enum QueryType : uint8_t {
...
+      SqlSource(SqlSource::FromTraceProcessorImplementation(""))};
+
+  if (sql.empty()) {
+    return result;
+  }
+
+  SqliteTokenizer tokenizer(SqlSource::FromTraceProcessorImplementation(sql));
+
+  SqliteTokenizer::Token statement_begin_tok = tokenizer.NextNonWhitespace();
+  while (statement_begin_tok.token_type == TK_SEMI) {
```
LalitMaganti: This still feels significantly more complex than it needs to be. Why is this not what I described in the previous one?

LalitMaganti (approved):

PR #1741, opened 2025-06-05
Filter out suspend slices that overlap with following slices

haixiaosugoogle:
b/422676164 - Traces with data losses outputs suspend slices that overlaps.

Filter out slices that overlaps with the following slices since we don't know what the data is like during that period.

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/suspend.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/suspend.sql
@@ -49,7 +49,8 @@ WITH
     WHERE
       track.name = 'Suspend/Resume Latency'
       AND (
-        slice.name = 'syscore_resume(0)' OR slice.name = 'timekeeping_freeze(0)')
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/stdlib/android/suspend.sql
```diff
@@ -38,30 +38,45 @@ WITH
...
+  ),
+  suspend_slice AS (
+    -- Filter out all the slices that overlapped with the following slices. 
+    -- This happens with data loss where we lose start and end slices for suspends.
+    SELECT
+      ts,
+      dur
+    FROM suspend_slice_pre_filter
+    WHERE dur > 0
+  ),
```
LalitMaganti, line 81: Is this a case of something we want to drop at the trace processor level?

haixiaosugoogle: Discussed offline, for this fix we want to fix it in sql.

PR #1745, opened 2025-06-05
tp: table tests for SF parser changes

PriyankaSPX:
Test rect computation at the table level, since it returns table ids and we want to check what data it actually added to the various Winscope tables.

Bug: 411363817

### DIFF
--- a/test/trace_processor/diff_tests/parser/android/surfaceflinger_layers.textproto
+++ b/test/trace_processor/diff_tests/parser/android/surfaceflinger_layers.textproto
@@ -1,30 +1,12 @@
 packet {
   clock_snapshot {
     primary_trace_clock: BUILTIN_CLOCK_BOOTTIME
-    clocks {
-      clock_id: 6
-      timestamp: 0
-    }
-    clocks {
-      clock_id: 2
-      timestamp: 1682359234732002451
-    }
-    clocks {
-      clock_id: 4
-      timestamp: 0
-    }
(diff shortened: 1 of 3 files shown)


LalitMaganti:
Thanks for sending. As it stands I see three problems with this:
1) the amount of code you're adding here means that now winscope is a non trivial percent of the binary size of Perfetto where for almost all users of Perfetto don't care. They are paying the cost for code they will never use. I feel we should start thinking about making winscope an optional module which is disabled by default
2) this patch is far too large to do a good review on. It needs to be split into 3-4 patches, 3500 lines in one go is not reasonable to review
3) I'm in the midst of a major refactor of how trace processor c++ tables work and I'm certain there will be a huge number of conflicts with this patch. I'd like to wait on landing this until that change lands.

PriyankaSPX:
1. Making it optional sounds sensible to me - happy to discuss that further
2. I've misunderstood the commit breakdown as acceptable for different stages of review, will split this into multiple patches
3. Happy to wait - do you have an estimated time frame on when your refactor will land? Ideally was hoping to get these changes landed by EOM

LalitMaganti:
Re ETA: it's a P0 I'm sending CLs on every day. I'm hoping within next few days

PriyankaSPX:
Okay great - making this draft again, will send in multiple patches once the refactor has landed

LalitMaganti (approved):

PR #1746, opened 2025-06-05
Trace Redaction - Expose "process tree adding duplicate entries" issue

vaage:
When redacting a process trees, there are two action:

1. Remove all processes and threads that are not connected to the target package.

2. Add entries for synthetic threads and processes.

Adding synth thread and process entries must happen after removing disconnected threads and processes to avoid removing the new entries.

In this change, we split the two operations into their own primitives to better expose a problem.

Filtering is simple, it optionally copies entries from one trace to the other. This prevents the need for any special logic to respect the trace's structure.

When adding entries, there needs to be additional logic. There is no logic when adding synth threads/processes, so it adds them to every process tree; it should only add them to one.

This change:

 1. Splits the two primitives, making it easier to see how the logic differs.

 2. Updates the tests to expose the issue and documents where the test is fudging values in order to pass.

A later change will be made to correct these issues.

Bug: 400093607


src/trace_redaction/add_synth_threads_to_process_trees.cc
```diff
@@ -0,0 +1,119 @@
...
+
+void AddThreadsToProcessTree(const Context& context,
+                             protos::pbzero::ProcessTree* process_tree) {
+  PERFETTO_DCHECK(context.synthetic_process);
+
+  auto& tids = context.synthetic_process->tids();
+  PERFETTO_DCHECK(tids.size() >= 2);
+
+  auto it = tids.begin();
+  ++it;
```
LalitMaganti: This ++ seems a bit suspicious. Is this intentional? If so, can you please leave a comment about this?

src/trace_redaction/add_synth_threads_to_process_trees.cc
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Copy fields from one process tree to another process tree.
+void CopyProcessTreeEntries(protozero::Field src,
+                            protos::pbzero::ProcessTree* dest) {
+  PERFETTO_DCHECK(src.valid());
+  PERFETTO_DCHECK(src.id() ==
+                  protos::pbzero::TracePacket::kProcessTreeFieldNumber);
+
+  protozero::ProtoDecoder decoder(src.as_bytes());
```
LalitMaganti: This seems quite wasteful to decode only to immediately reencode. Can we not just do AppendRawProtoBytes or similar?

src/trace_redaction/add_synth_threads_to_process_trees.cc
vaage: It was intentional; the first thread was the main thread. I've moved it into its own field. This should make the code easier to read.

src/trace_redaction/add_synth_threads_to_process_trees.cc
vaage: We are building a new process tree. Copying the original processes and threads and then add new ones. I've tried using `process_util::AppendField` but it overrides the previous process tree. Am I missing something?

LalitMaganti: I think this decoder and all the below loop can be replaced with:
```
dest->AppendRawProtoBytes(src.as_bytes());
```

vaage: Thanks. Done.

LalitMaganti (approved):

LalitMaganti (approved):

LalitMaganti (approved):

stevegolton (approved):

LalitMaganti:
@vaage: seems like there is some issue in the test: https://github.com/google/perfetto/actions/runs/15684498164/job/44286221342

PR #1747, opened 2025-06-05
dev/primiano/intrlist

primiano:
- **base: Add PushBack/PopBFront to Intrusive list**
- **Use masked pointers for next/prev**


src/base/intrusive_list.h
```diff
@@ -48,26 +48,56 @@ namespace perfetto::base {
...
   void PushFront(ListNode* node);
+  void PushBack(ListNode* node);
   void PopFront();
+  void PopBack();
   void Erase(ListNode* node);
+  bool empty() const { return head_.next_ == head_masked(); }
 
-  ListNode* front_{nullptr};
-  size_t size_{0};
+  uintptr_t head_masked() const {
```
LalitMaganti: Comment on this or naming something else would be useful. Is the most mysterious feeling part of this CL, not sure it's the best name you could use.

primiano:
Hey mr @rukkal , I started using your IntrusiveList and would like to propose some changes.
TBH there is another change that I would like to do, that is: allowing to erase() a node, without knowing the list it belongs to. In other words being able to do this:

```
 list_.PushFront(p3_);
 list_.PushFront(p2_);
 list_.PushFront(p1_);

p2_.node.erase();
```

That is almost trivial to do, but I need to remove size() (because that's the only thing that right now requires kwowing the list).

TBH size() it's questionable for a linked list.
The only problem is that today your slab allocator seems to use size() to do a +1 check. Is there any chance that could be done elsewhere without relying on the list maintaining its own size?

If you really need it, then I need to make it some sort of trait, where one can chose if you want a list-with-size-but-without-node-erase or list-without-size-but-with-node-erase, although that makes the code uglier.

src/base/intrusive_list.h
```diff
@@ -48,26 +48,56 @@ namespace perfetto::base {
...
+    return reinterpret_cast<uintptr_t>(&head_) | 1;
+  }
+  ListNode head_{head_masked(), head_masked()};
```
LalitMaganti: I think after reading the patch a bunch I understand this but I think calling this "head" is quite misleading. THis is more a "close the loop" node (i.e. prev points to tail, next points to head). I think calling it something different would help with readaiblity.

primiano: I used the name list_head because that's really the same name that the kernel uses for precisely the same reason since the beginning of times:
https://elixir.bootlin.com/linux/2.1.45/source/include/linux/list.h

src/base/intrusive_list.h
primiano: maybe head_with_sentinel? or head_sentinel()?

src/base/intrusive_list.h
LalitMaganti: Fair enough, I still think it's quite misleading but who am I to question the kernel :)

src/base/intrusive_list.h
LalitMaganti: Yeah `head_with_sentinel` or `head_and_sentinel` seems good.

src/base/intrusive_list.h
primiano: renamed to head_and_tail

src/base/intrusive_list.cc
```diff
@@ -21,46 +21,48 @@
...
-  }
-
-  if (next) {
-    next->prev = prev;
-  }
-
-  PERFETTO_DCHECK(size_ > 0);
+  prev->next = next;
+  next->prev = prev;
+  node->prev = node->next = nullptr;
```
rukkal: here you might have to update head_

src/base/intrusive_list.h
```diff
@@ -62,14 +84,32 @@ class ListOps {
...
+  ListNode head_and_tail_{sentinel(), sentinel()};
+  size_t size_ = 0;
 };
 
 }  // namespace internal
 
-using IntrusiveListNode = internal::ListNode;
+template <typename T, typename Traits = typename T::Traits>
+struct IntrusiveListNode : private internal::ListNode {
+  T* prev() {
```
rukkal: This feels a bit redundant considering that we already have the Iterator that can move forwards/backwards. Would make sense to extend iterator with something like IsSentinel() and use the iterator instead?

src/base/intrusive_list.cc
rukkal: nvm. I see this code is completely rewritten in the next commit

rukkal:
> Hey mr @rukkal , I started using your IntrusiveList and would like to propose some changes. TBH there is another change that I would like to do, that is: allowing to erase() a node, without knowing the list it belongs to. In other words being able to do this:
> 
> ```
>  list_.PushFront(p3_);
>  list_.PushFront(p2_);
>  list_.PushFront(p1_);
> 
> p2_.node.erase();
> ```
> 
> That is almost trivial to do, but I need to remove size() (because that's the only thing that right now requires kwowing the list).
> 
> TBH size() it's questionable for a linked list. The only problem is that today your slab allocator seems to use size() to do a +1 check. Is there any chance that could be done elsewhere without relying on the list maintaining its own size?
> 
> If you really need it, then I need to make it some sort of trait, where one can chose if you want a list-with-size-but-without-node-erase or list-without-size-but-with-node-erase, although that makes the code uglier.

Hey! I think it's nice to have an IntrusiveList::size(), but we can definitely remove it to unblock the TB refactoring. I pushed #1751 with the change

src/base/intrusive_list.h
primiano: so the problem I have is that in TraceBufferV2 the following will happen:

For context there will be:
- 1 TraceBuffer (well more than one, one for buffer configured, let's say one here)
- The buffer has N chunks (stored linearly in the mmaped area)
- The buffer has also a map of M per-sequence state (one per data source writer / thread)
- Each sequence state has a list of chunks 
- Each chunk is part of one (and at most one) of those M lists

Now the problem is that I start my walking from the chunk. at that point I don't know which list it belongs to, and it would be nice to be able to do "go to the previous chunk in the sequence, without having to do further lookups"

The problem is that in order to obtain a cursor, I need to have a reference/pointer to the list that owns the node.

primiano:
Let me keep this PR on hold for a little bit. I start wondering if I need more tweaks (or if I should have something custom for TBV2).
I'll come back on this  after i finish the TBV2 cl

primiano:
@rukkal @LalitMaganti can you PTAL now?
I have not finished the TBV2 one yet, but I made enough progress and I think these are the operations I need.

src/base/intrusive_list.h
rukkal: > I need to have a reference/pointer to the list that owns the node

Don't you need to know only the type of the list? And then you can `IntrusiveList<Chunk>::Iterator(node_address)`

primiano:
Ended up not needing this

PR #1758, opened 2025-06-06
UI: Improve Wattson runtime and amalgamate counter slices

wusamuel6:
ui: Improve runtime and QOL

Slightly improve UI runtime for Wattson and make Wattson counters easier to use and view.

Bug: 422853790


LalitMaganti (commented):
Please fix the PR title.

ui/src/plugins/org.kernel.Wattson/process_aggregator.ts
```diff
@@ -43,15 +42,13 @@ export class WattsonProcessSelectionAggregator
...
-      -- CPUs and GROUP BY process
+      -- Group idle attribution by process
       CREATE OR REPLACE PERFETTO TABLE
       wattson_plugin_per_process_idle_attribution AS
       SELECT
         SUM(idle_cost_mws) as idle_cost_mws,
         upid
-      FROM _filter_idle_attribution(${area.start}, ${duration})
-      WHERE cpu in ${cpusCsv}
+      FROM wattson_plugin_idle_attribution
```
stevegolton, line 51: This is a little fragile right? It depends on the thread_aggregator running first and creating the table with the right time span, right?

stevegolton: Discussed offline, this table sharing is actually nothing new so fixing it is out of scope of this PR.

PR #1759, opened 2025-06-06
Show per-machine system info in Info and Stats page

jahdiel-alvarez:
In unified traces it is very valuable to know the system information of each machine. This allows you to reason about the source of the tracing data.

Bug: 406278549

### DIFF
--- a/ui/src/plugins/dev.perfetto.TraceInfoPage/trace_info_page.ts
+++ b/ui/src/plugins/dev.perfetto.TraceInfoPage/trace_info_page.ts
@@ -126,7 +126,7 @@ class StatsSection implements m.ClassComponent<StatsSectionAttrs> {
     return m(
       `section${attrs.cssClass}`,
       m('h2', attrs.title),
-      m('h3', attrs.subTitle),
+      m('h4', attrs.subTitle),
       m(
         'table',
         m('thead', m('tr', m('td', 'Name'), m('td', 'Value'), m('td', 'Type'))),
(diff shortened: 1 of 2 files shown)


jahdiel-alvarez:
UI screenshot:
machine_stats_perfetto

stevegolton (approved):
LGTM,  but why the changes to the h3 sizes?

ui/src/plugins/dev.perfetto.TraceInfoPage/trace_info_page.ts
```diff
@@ -215,6 +215,86 @@ class TraceMetadata implements m.ClassComponent<EngineAttrs> {
...
+
+  view() {
+    const data = this.data;
+    if (data === undefined || data.length <= 1) {
+      return undefined;
+    }
+
+    const machineTables = data.map((row) => {
+      const tableRows = [];
+      for (const key of Object.keys(machineRowSpec)) {
```
stevegolton, line 271: I think you can do this with Object.entries and avoid the type assertion below

jahdiel-alvarez:
@stevegolton I basically moved h3 to be h4 and created a new h3. I did this because the prev h3 (now h4), looked like normal text (see the text description below Machines header in the screenshot). So I needed an intermediate header in between h2 and h3 for the Machine name headers (see Machine 0 in screenshot).

ui/src/plugins/dev.perfetto.TraceInfoPage/trace_info_page.ts
jahdiel-alvarez: Doesn't quite work because we are only iterating through the keys of `machineRowSpec`, the values are taken from the row object not the spec.

stevegolton: Ah yes you're totally right, sorry.

PR #1763, opened 2025-06-09
ui: Explore Page: Always visualise table

aMayzner:
Show table (default in the bottom right side) also when there are no results - do it to keep the layout consistent and fix the bug with SQL source

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/node_data_viewer.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/node_data_viewer.ts
@@ -55,19 +55,22 @@ export class NodeDataViewer implements m.ClassComponent<NodeDataViewerAttrs> {
         attrs.onQueryExecuted();
       });
     };
-    if (attrs.query === undefined) {
-      return m(TextParagraph, {text: `No data to display}`});
-    }
-    if (attrs.query instanceof Error) {
-      return m(TextParagraph, {text: `Error: ${attrs.query.message}`});
-    }
-    if (this.queryResult === undefined) {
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/node_data_viewer.ts
```diff
@@ -106,15 +109,18 @@ export class NodeDataViewer implements m.ClassComponent<NodeDataViewerAttrs> {
...
-        m(
-          'article',
-          m(QueryTable, {
-            trace: attrs.trace,
-            query: queryToRun(attrs.query),
-            resp: this.queryResult,
-            fillParent: false,
-          }),
-        ),
+        queryErrors()
```
stevegolton: Maybe cache the result of this function to avoid the duplicate conditional?

ui/src/plugins/dev.perfetto.ExplorePage/query_builder/node_data_viewer.ts
```diff
@@ -55,19 +55,22 @@ export class NodeDataViewer implements m.ClassComponent<NodeDataViewerAttrs> {
...
+        return `Error: ${attrs.query.message}`;
+      }
+      if (this.queryResult === undefined) {
+        runQuery();
+        return `No data to display`;
+      }
+      if (this.queryResult.error !== undefined) {
+        return `Error: ${this.queryResult.error}`;
+      }
+      return;
```
stevegolton: Maybe explicitly return undefined here - leaving it naked seems a bit confusing

PR #1769, opened 2025-06-10
Add SF and HWUI callback missed as part of android_frames table.

devianb:
Remove intermediate table containing information about the SF and HWUI missed callbacks for frames, and add this information directly in the stdlib android_frames table.

Bug: 400855183
Test: tools/diff_test_trace_processor.py --name-filter ".*android_frames.*" out/linux_clang_debug/trace_processor_shell

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
@@ -137,6 +137,36 @@ WHERE
 GROUP BY
   vsync;
 
+CREATE PERFETTO TABLE _android_jank_cuj_sf_process AS
+SELECT
+  *
+FROM process
+WHERE
+  process.name = '/system/bin/surfaceflinger'
+LIMIT 1;
+
+-- Match the frame timeline on the app side with the frame timeline on the SF side.
+-- In cases where there are multiple layers drawn, there would be separate frame timeline
+-- slice for each of the layers. GROUP BY is used to deduplicate these rows.
+CREATE PERFETTO TABLE android_app_to_sf_vsync_match AS
+SELECT
+  app_timeline.upid AS app_upid,
+  CAST(app_timeline.name AS INTEGER) AS app_vsync,
+  sf_process.upid AS sf_upid,
+  CAST(sf_timeline.name AS INTEGER) AS sf_vsync
+FROM actual_frame_timeline_slice AS app_timeline
+JOIN directly_connected_flow(app_timeline.id) AS flow
+  ON flow.slice_out = app_timeline.id
+JOIN actual_frame_timeline_slice AS sf_timeline
+  ON flow.slice_in = sf_timeline.id
+JOIN _android_jank_cuj_sf_process AS sf_process
+  ON sf_timeline.upid = sf_process.upid
+GROUP BY
+  app_upid,
+  app_vsync,
+  sf_upid,
+  sf_vsync;
+
 -- TODO(b/384322064) Match actual timeline slice with correct draw frame using layer name.
 -- All slices related to one frame. Aggregates `Choreographer#doFrame`,
 -- `actual_frame_timeline_slice` and `expected_frame_timeline_slice` slices.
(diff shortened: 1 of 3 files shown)


LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
```diff
@@ -125,6 +125,18 @@ SELECT
   id
 FROM expected_frame_timeline_slice;
 
+-- Contains frames with missed SF or HWUI callbacks.
+CREATE PERFETTO TABLE _vsync_missed_callback AS
+SELECT
+  CAST(str_split(name, 'Callback#', 1) AS INTEGER) AS frame_id,
```
nicomazz: why renaming this from vsync to frame_id?

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
```diff
@@ -293,7 +305,11 @@ CREATE PERFETTO TABLE android_frames (
   -- process id.
   upid JOINID(process.id),
   -- process name.
-  process_name STRING
+  process_name STRING,
+  -- Indicates whether the frame missed SF callback.
+  sf_callback_missed LONG,
+  -- Indicates whether the frame missed HWUI callback.
+  hwui_callback_missed LONG
```
nicomazz: I think it's not very clear why we need to expose these from android_frames: from the cl it looks like we're never using them? 
While it doesn't hurt much having these here, I wonder if we can keep the final "android_frames" table a bit cleaner without them

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
devianb: Changed to vsync.

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
devianb: > I wonder if we can keep the final "android_frames" table a bit cleaner without them

The original idea was to change to logic in the frames.sql table and use these columns from the android_frames table. But I see your point, and agree that we can just use the _vsync_missed_callback table directly from the stdlib, and not modify the android_frames table. Fixed it.

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
```diff
@@ -125,6 +125,48 @@ SELECT
...
+  CAST(str_split(name, 'Callback#', 1) AS INTEGER) AS vsync,
+  max(name GLOB '*SF*') AS sf_callback_missed,
+  max(name GLOB '*HWUI*') AS hwui_callback_missed
+FROM slice
+WHERE
+  name GLOB '*FT#Missed*Callback*'
+GROUP BY
+  vsync;
+
+CREATE PERFETTO TABLE _android_jank_cuj_sf_process AS
```
nicomazz, line 140: It looks like this is being defined here, but also the previous revision was using "android_jank_cuj_sf_process". Should we delete the previous version, and update refs to refer to this one?

devianb: The original table android_jank_cuj_sf_process has refs in files which we don't intend to use in any jank metrics. I believe updating refs in those files should not be a part of this CL, but rather a separate cleanup CL.

nicomazz: it looks like it's created in sql/android/jank/relevant_threads.sql, and used in the jank tag lib. Probably you can just make the definition there reference this table, with a plain 

```
-- in perfetto/src/trace_processor/metrics/sql/android/jank/relevant_threads.sql
DROP TABLE IF EXISTS android_jank_cuj_sf_process;
CREATE PERFETTO TABLE android_jank_cuj_sf_process AS
SELECT * FROM _android_jank_cuj_sf_process;
```

so at least we have a single definition of it.

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
devianb: Done

LalitMaganti (approved):

PR #1771, opened 2025-06-10
tp: annotate all python tables with read/write access requirements

LalitMaganti:
This is required for upcoming dataframe migration. This change is a noop on the existing table based system as these annotations are just ignored.


src/trace_processor/tables/py_tables_unittest.cc
```diff
@@ -1,282 +0,0 @@
-/*
```
aMayzner, line 1: Why do we remove it?

src/trace_processor/tables/py_tables_unittest.py
```diff
@@ -1,64 +0,0 @@
-# Copyright (C) 2022 The Android Open Source Project
```
aMayzner, line 1: Why do we remove it?

src/trace_processor/tables/py_tables_unittest.cc
LalitMaganti: Because the amount of code in macros now is negligible, all the test should happen at the dataframe level.

src/trace_processor/tables/py_tables_unittest.py
LalitMaganti: Because the amount of code in macros now is negligible, all the test should happen at the dataframe level.

PR #1774, opened 2025-06-10
tp: migrate initial set of static tables to dataframe

LalitMaganti:
Migrate any table which does not need any changes (i.e. is using the table in a non-trivial way).

### DIFF
--- a/python/generators/trace_processor_table/public.py
+++ b/python/generators/trace_processor_table/public.py
@@ -193,6 +193,7 @@ class Table:
   add_implicit_column: bool = True
   tabledoc: Optional[TableDoc] = None
   wrapping_sql_view: Optional[WrappingSqlView] = None
+  # TODO(lalitm): remove once migration sticks.
   use_legacy_table_backend: bool = False
 
 


python/generators/trace_processor_table/public.py
```diff
@@ -194,6 +194,7 @@ class Table:
   add_implicit_column: bool = True
   tabledoc: Optional[TableDoc] = None
   wrapping_sql_view: Optional[WrappingSqlView] = None
+  use_legacy_table_backend: bool = False
```
aMayzner, line 197: Please add a todo to remove it, after the migration

python/generators/trace_processor_table/serialize_new.py
```diff
@@ -0,0 +1,603 @@
...
+from python.generators.trace_processor_table.public import ColumnFlag
+from python.generators.trace_processor_table.public import CppAccess
+from python.generators.trace_processor_table.public import CppAccessDuration
+from python.generators.trace_processor_table.public import SqlAccess
+from python.generators.trace_processor_table.util import ParsedTable
+from python.generators.trace_processor_table.util import ParsedColumn
+from python.generators.trace_processor_table.util import parse_type
+
+
+class ColumnSerializer:
```
aMayzner, line 28: Shouldn't there be some tests for it?

aMayzner (approved):

python/generators/trace_processor_table/serialize_new.py
LalitMaganti: No need for tests, it's a very minimal wrapper over data frame and data frame has some extensive tests.

PR #1784, opened 2025-06-10
add thermal data source to poll temperatures

kev7c7:
Add the "android.thermal" data source. Below is a sample config:

data_sources: {
  config {
    name: "android.thermal"
    android_thermal_config {
      poll_ms: 1000
      sensors: "BIG"
      sensors: "MID"
      sensors: "usb_pwr_therm"
      sensors: "usb_pwr_therm2"
      sensors: "maxfg"
      sensors: "disp_therm"
      sensors: "therm_6"
      sensors: "therm_7"
      sensors: "battery"
      sensors: "battery_cycle"
      sensors: "soc"
      sensors: "smpl_gm"
      sensors: "LITTLE"
      sensors: "vdroop1"
      sensors: "batoilo"
      sensors: "vdroop2"
      sensors: "batt_vs"
      sensors: "mdis_morepower"
      sensors: "thb_hda"
      sensors: "G3D"
      sensors: "ISP"
      sensors: "TPU"
      sensors: "AUR"
      sensors: "neutral_therm"
      sensors: "quiet_therm"
      sensors: "charger_skin_therm"
    }
  }
}

Test: tested on pixel device. Also has been used extensively in Android XR project.
Bug: 423091981


LalitMaganti:
Ryan: can you please review this?

src/traced/probes/thermal/android_thermal_data_source.cc
```diff
@@ -0,0 +1,193 @@
...
+      poll_interval_ms_ - static_cast<uint32_t>(now_ms % poll_interval_ms_));
+
+  WriteTemperatureCountersData();
+}
+
+void AndroidThermalDataSource::WriteTemperatureCountersData() {
+  auto packet = writer_->NewTracePacket();
+  auto timestamp = static_cast<uint64_t>(base::GetBootTimeNs().count());
+  packet->set_timestamp(timestamp);
+  auto* bundle = packet->set_ftrace_events();
```
rsavitski, line 161: This is wrong. You shouldn't be faking ftrace events, as those protobuf messages are specifically for converting data obtained from /sys/kernel/tracing.

There should instead be a dedicated protobuf message/field for this under protos/perfetto/trace/, and src/trace_processor should have explicit parsing of that protobuf field.

(However, see the comment on config.proto because there's already thermal_zone polling in perfetto, rendering this entire file possibly redundant)

protos/perfetto/config/thermal/android_thermal_config.proto
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+message AndroidThermalConfig {
```
rsavitski, line 21: Perfetto already has a sysfs poller with `thermal_zone` support: [link](https://source.chromium.org/search?q=f:%5Ethird_party%2Fperfetto%20%22%2Fsys%2Fclass%2Fthermal%2F%22).

Instead of adding a new data source, could you modify the existing one to support sensor name filtering, and/or whatever else you need?

kev7c7: Thanks! It looks like the poller in sys_stats does exactly what we need; will advise XR folks to use that instead. With that feature, this CL is no longer needed.

PR #1810, opened 2025-06-12
tp: skip sorting when single sort constraint is on ordered column

LalitMaganti:
There's no point sorting if there's only a single order by on a column
which is already sorted

### DIFF
--- a/src/trace_processor/dataframe/dataframe_unittest.cc
+++ b/src/trace_processor/dataframe/dataframe_unittest.cc
@@ -819,6 +819,96 @@ TEST_F(DataframeBytecodeTest, PlanQuery_MinOptimizationApplied) {
                   expected_bytecode, /*cols_used=*/1);
 }
 
+TEST_F(DataframeBytecodeTest, SortOptimizationApplied_SingleAscNonNullSorted) {
+  std::vector<impl::Column> cols = MakeColumnVector(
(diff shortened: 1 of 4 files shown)


src/trace_processor/dataframe/dataframe_unittest.cc
```diff
@@ -819,6 +819,96 @@ TEST_F(DataframeBytecodeTest, PlanQuery_MinOptimizationApplied) {
...
 }
 
+TEST_F(DataframeBytecodeTest, SortOptimizationApplied_SingleAscNonNullSorted) {
+  std::vector<impl::Column> cols = MakeColumnVector(
+      impl::Column{impl::Storage::Uint32{}, impl::NullStorage::NonNull{},
+                   Sorted{}, HasDuplicates{}});
+  std::vector<FilterSpec> filters;
+  std::vector<SortSpec> sorts = {{0, SortDirection::kAscending}};
+  RunBytecodeTest(cols, filters, {}, sorts, {}, R"(
+    InitRange: [size=0, dest_register=Register(0)]
```
aMayzner, line 829: Would it be possible to say in the bytecode that the optimisation has been applied? It will greatly help with debugging in the future

LalitMaganti: Putting it in the bytecode is very much the wrong approach as everything you put here has a cost.

For debugging, we have the metatrace and it'll clearly be indicated there.

PR #1820, opened 2025-06-12
tp: _interval_interesect supports instant events

aMayzner:
_interval_intersect macro will now return instants if there are instant results. Removed the dur > 0 check and added tests. If the interval (ts, dur) is (0, 10) it will return instant when it's inside it, (5, 0), but the ends are excluded, so there is no overlap with (10, 2).

### DIFF
--- a/src/trace_processor/containers/interval_tree.h
+++ b/src/trace_processor/containers/interval_tree.h
@@ -39,6 +39,23 @@ struct Interval {
   Id id;
 };
 
+inline bool IsOverlapping(bool query_is_instant,
+                          Ts s,
+                          Ts e,
+                          const Interval& i) {
+  bool interval_is_instant = (i.start == i.end);
+  if (query_is_instant && interval_is_instant) {
+    return s == i.start;
+  }
+  if (query_is_instant) {
+    return i.start <= s && s < i.end;
+  }
+  if (interval_is_instant) {
+    return s <= i.start && i.start < e;
+  }
+  return e > i.start && s < i.end;
+}
+
 // An implementation of a centered interval tree data structure, designed to
 // efficiently find all overlap queries on a set of intervals. Centered interval
 // tree has a build complexity of O(N*logN) and a query time of O(logN + k),
@@ -60,6 +77,7 @@ class IntervalTree {
   // overlaps)).
   void FindOverlaps(uint64_t s, uint64_t e, std::vector<Id>& res) const {
     std::vector<const Node*> stack{nodes_.data() + root_};
+    bool query_is_instant = (s == e);
     while (!stack.empty()) {
       const Node* n = stack.back();
       stack.pop_back();
@@ -72,7 +90,7 @@ class IntervalTree {
           break;
         }
 
-        if (e > i.start && s < i.end) {
+        if (IsOverlapping(query_is_instant, s, e, i)) {
           res.push_back(i.id);
         }
       }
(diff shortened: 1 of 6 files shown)


src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
```diff
@@ -330,13 +330,21 @@ struct IntervalTreeIntervalsAgg
           ctx, "Interval intersect requires intervals to be sorted by ts.");
       return;
     }
-    agg_ctx.last_interval_start = interval.start;
     int64_t dur = sqlite::value::Int64(argv[2]);
-    if (dur < 1) {
-      sqlite::result::Error(
-          ctx, "Interval intersect only works on intervals with dur > 0");
+
+    // Ignore instants. We shouldn't error out on them, as they might be valid
```
LalitMaganti: IMO instants should be passed through as-is as that's what most people would want/expect.

test/trace_processor/diff_tests/stdlib/intervals/intersect_tests.py
```diff
@@ -328,30 +328,74 @@ def test_single_interval(self):
...
+        FROM _interval_intersect_single!(1, 6, A)
         ORDER BY ts;
         """,
         out=Csv("""
         "id","ts","dur"
         0,1,1
-        1,3,2
+        1,3,0
         2,6,1
         """))
```
LalitMaganti, line 401: Please add some better tests for multi-interval intersection with instants. Also test edge cases (what happens on instant at end of interval/start of interval etc.)

test/trace_processor/diff_tests/stdlib/intervals/intersect_tests.py
```diff
@@ -328,33 +328,274 @@ def test_single_interval(self):
...
+    return DiffTestBlueprint(
+        trace=TextProto(""),
+        #      0 1 2 3 4 5 6 7
+        # A:   _ - - - - - - _
+        # B:   . _ . . _ . _ .
+        # res: _ _ . . _ . _ _
+        query="""
+        INCLUDE PERFETTO MODULE intervals.intersect;
+
+        CREATE PERFETTO TABLE B_instants AS
```
LalitMaganti, line 577: I believe instants on the edges of intervals should be passed through.

test/trace_processor/diff_tests/stdlib/intervals/intersect_tests.py
```diff
@@ -328,33 +328,274 @@ def test_single_interval(self):
...
+        "id","ts","dur"
+        """))
+
+  def test_instants(self):
+    return DiffTestBlueprint(
+        trace=TextProto(""),
+        #      0 1 2 3 4 5 6 7
+        # A:   _ - - - - - - _
+        # B:   - - _ _ _ _ - -
+        # res: _ - _ - - _ - _
```
LalitMaganti: This ascii diagram does not make sense.

src/trace_processor/containers/interval_tree.h
```diff
@@ -39,6 +39,23 @@ struct Interval {
   Id id;
 };
 
+inline bool IsOverlapping(bool query_is_instant,
+                          Ts s,
+                          Ts e,
+                          const Interval& i) {
+  bool interval_is_instant = (i.start == i.end);
+  if (query_is_instant && interval_is_instant) {
```
LalitMaganti, line 47: I believe instants at both the start and end of the interval should be passed through.

PR #1824, opened 2025-06-13
probes: fix subtle bugs with clock snapshots for non-boot ftrace clocks

LalitMaganti:
* correctly snapshot ftrace clocks per instance. The previous approach
  was straight up broken: if your ftrace instance uses a different
  clock to the primary instance, you'd get incorrect snapshots
* fix snapshotting of local clock. local requires reading the time
  per cpu, not just the cpu zero clock

### DIFF
--- a/src/traced/probes/ftrace/ftrace_controller.cc
+++ b/src/traced/probes/ftrace/ftrace_controller.cc
@@ -259,8 +259,15 @@ void FtraceController::StartIfNeeded(FtraceInstanceState* instance,
   PERFETTO_CHECK(instance->cpu_clock_state.empty());
   instance->cpu_clock_state.reserve(num_cpus);
   for (size_t cpu = 0; cpu < num_cpus; cpu++) {
-    instance->cpu_clock_state.emplace_back(
-        CpuClockState{instance->ftrace_procfs->OpenCpuStats(cpu), {}});
+    // If using the boot clock, don't bother opening the per-CPU stats file as
+    // we won't use it. Otherwise, open it for use in MaybeSnapshotFtraceClock.
+    if (instance->ftrace_clock ==
+        protos::pbzero::FtraceClock::FTRACE_CLOCK_UNSPECIFIED) {
+      instance->cpu_clock_state.emplace_back();
+    } else {
+      instance->cpu_clock_state.emplace_back(
+          CpuClockState{instance->ftrace_procfs->OpenCpuStats(cpu), {}});
+    }
   }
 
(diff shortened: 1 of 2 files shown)


src/traced/probes/ftrace/ftrace_controller.cc
```diff
@@ -691,21 +693,21 @@ void FtraceController::DumpFtraceStats(FtraceDataSource* data_source,
...
-void FtraceController::MaybeSnapshotFtraceClock() {
-  if (!cpu_zero_stats_fd_)
+void FtraceController::MaybeSnapshotFtraceClock(FtraceInstanceState* instance) {
+  if (instance->ftrace_clock ==
+      protos::pbzero::FtraceClock::FTRACE_CLOCK_UNSPECIFIED)
     return;
 
-  auto ftrace_clock = primary_.ftrace_config_muxer->ftrace_clock();
-  PERFETTO_DCHECK(ftrace_clock != protos::pbzero::FTRACE_CLOCK_UNSPECIFIED);
+  for (auto& clock : instance->cpu_clock_state) {
```
rsavitski, line 701: Isn't this logically misleading - the tracefs stats file calls into kernel code to get a fresh "current" timestamp, but for a cpu-local clock it means that the result is only relevant to the cpu THIS code is running under (i.e. the cpu where traced_probes is currently scheduled).

Given [tracing_stats_read](https://github.com/torvalds/linux/blob/44a5ab7a7958fbf190ae384b8ef252f23b840c1b/kernel/trace/trace.c#L8749) -> [rb_time_stamp](https://github.com/torvalds/linux/blob/44a5ab7a7958fbf190ae384b8ef252f23b840c1b/kernel/trace/ring_buffer.c#L1090).

That said, I agree with adding support for "local" clocks on a best effort manner. Maybe we can stick to a single snapshot across all cpus as "good enough in practice". I checked a modern arm64 phone and the boot<->local skew is ~2 ms per hour.

rsavitski (changes requested):

src/traced/probes/ftrace/ftrace_controller.cc
rsavitski: (To clarify: I am happy to take over the patch series, but let's agree on an approach)

LalitMaganti: If the drift is really that low, sticking with the existing approach of just using CPU 0's clock just makes sense. Should we just abandon this CL and just make the trace processor change to stop blocking local clock?

Issue #1828, opened 2025-06-13
Don't use protos in Chromium standard library


petrcermak:
The [`chrome_causes_per_scroll` view](https://github.com/google/perfetto/blob/fbed4ffaccc0e754779134993ae8212091582aa9/src/trace_processor/perfetto_sql/stdlib/chrome/scroll_jank/scroll_jank_v3.sql#L477-L508) in `src/trace_processor/perfetto_sql/stdlib/chrome/scroll_jank
/scroll_jank_v3.sql` uses protocol buffers. However, **the standard library is not supposed to use protos**. Only metrics code should use protos.

Since the [`chrome_scroll_jank_v3_intermediate` view](https://github.com/google/perfetto/blob/fbed4ffaccc0e754779134993ae8212091582aa9/src/trace_processor/metrics/sql/chrome/chrome_scroll_jank_v3.sql#L20-L56) in `src/trace_processor/metrics/sql/chrome/chrome_scroll_jank_v3.sql` is the only user of `chrome_causes_per_scroll`, we should just move `chrome_causes_per_scroll` to `src/trace_processor/metrics/sql/chrome/chrome_scroll_jank_v3.sql` (where protos are allowed).

PR #1829, opened 2025-06-13
Move chrome_(scroll_jank_v3_)causes_per_scroll to trace_processor/metrics

petrcermak:
This change copies the definition of `chrome_causes_per_scroll` from `src/trace_processor/perfetto_sql/stdlib/chrome/scroll_jank/scroll_jank_v3.sql` into a new view `chrome_scroll_jank_v3_causes_per_scroll` in `src/trace_processor/metrics/sql/chrome/chrome_scroll_jank_v3.sql`.

Bug: #1828


petrcermak:
> You didn't remove it from stdlib

The SoT for `src/trace_processor/perfetto_sql/stdlib/chrome/scroll_jank/scroll_jank_v3.sql` is in Chromium so I need to make the change there instead.

aMayzner:
Ok, that's fine, you can submit it.

PR #1835, opened 2025-06-13
tp: migrate all winscope tables to dataframe

LalitMaganti:
As part of this change, to debug issues, add DCHECK to base::unchecked_get. To avoid circular deps, create a new variant.h header to contain unchecked_get.


include/perfetto/ext/base/variant.h
```diff
@@ -0,0 +1,70 @@
+/*
```
aMayzner, line 1: Update the CL description to explain what is really happening - this Cl seems to be about a lot more than migrating winscope tables

aMayzner (approved):

PR #1846, opened 2025-06-15
tp: migrate profiler and metadata tables to dataframe

LalitMaganti:


src/trace_processor/storage/trace_storage.h
```diff
@@ -340,16 +340,11 @@ class TraceStorage {
     // of tables. If we wanted to extend this to every table, we'd need to deal
     // with tracking all the tables in the storage: this is not worth doing
     // given most memory is used by these tables.
-    thread_table_.ShrinkToFit();
```
aMayzner, line 343: Is memory usage of those tables not important anymore?

LalitMaganti: Every dataframe is automatically shrunk in memory when finalized.

PR #1847, opened 2025-06-15
tp: migrate all tables except those depended upon by table functions

LalitMaganti:


src/trace_processor/perfetto_sql/stdlib/android/network_packets.sql
```diff
@@ -73,4 +73,6 @@ SELECT
   remote_port,
   packet_icmp_type,
   packet_icmp_code
-FROM __intrinsic_android_network_packets;
+FROM __intrinsic_android_network_packets
```
aMayzner, line 76: Why wouldn't we also expose id in this table?

aMayzner (approved):

src/trace_processor/perfetto_sql/stdlib/android/network_packets.sql
LalitMaganti: I don't know but it wasn't before and I don't want to change public API in the middle of all of this :)

PR #1853, opened 2025-06-16
tp: etm: symbolize vtable

OmarElfouly:
Lands https://android-review.googlesource.com/c/platform/external/perfetto/+/3389389.

Below is the original description:

Adds a vtable to symbolize addresses. It uses the existing symbolizer class under the hoods, which basically means we will spawn a llvm-symbolize process to perform the actual symbolization.

Change-Id: I7d0769ef776256c150d96e7c7b9991cc99680aba


src/trace_processor/trace_processor_impl.cc
```diff
@@ -1350,6 +1351,10 @@ std::unique_ptr<PerfettoSqlEngine> TraceProcessorImpl::InitPerfettoSqlEngine(
       "__intrinsic_etm_decode_trace", storage);
   engine->RegisterVirtualTableModule<etm::EtmIterateRangeVtable>(
       "__intrinsic_etm_iterate_instruction_range", storage);
+  if (config.dev_flags.count("llmv-symbolizer")) {
```
LalitMaganti: nit: typo

LalitMaganti: What is the long term plan for this? Are we planning on building scripts around this? If so, we should look into how we can get this *away* from depending on dev flags.

src/trace_processor/perfetto_sql/intrinsics/operators/symbolize_vtable.cc
```diff
@@ -0,0 +1,369 @@
...
+  std::vector<profiling::SymbolizedFrame> frames_;
+};
+
+SymbolizeVtable::Vtab::Vtab(TraceStorage* storage) : storage_(storage) {
+  llvm_ = std::make_unique<profiling::LLVMSymbolizerProcess>("llvm-symbolizer");
+}
+
+void SymbolizeVtable::Cursor::SymbolizeMapping(MappingId mapping_id,
+                                               uint64_t address) {
+  auto storage = vtab_->storage();
```
LalitMaganti, line 103: Instead of hardcoding that this code only works on these tables, who not instead take as input the file_name and rel_pc as input to this vtable and handle the rest in SQL?

LalitMaganti: Also if you do this, I think turning this into a table function would honestly make a lot more sense than a straight SQLite vtable. You can significantly simplify all of this code if you do this.

As an FYI, I've undertaken a huge rewrite of a lot of this part of trace processor and you can see there are a lot of pending changes of mine waiting for review at ci.perfetto.dev.

src/trace_processor/trace_processor_impl.cc
OmarElfouly: Done.

rasikanavarange: Yes, we will want to build scripts around this so probably should find a way to not use a dev flag. Why does the current implementation need to be behind a dev flag?

LalitMaganti: Well depending on a sub-process is not really a great state to be because it will only work in shell. It won't work in WASM etc. 

If we use LLVM is a library that would mean we don't need to put behind a dev flag.

PR #1855, opened 2025-06-16
Add machine id to global Meminfo and CPU tracks

jahdiel-alvarez:
When visualizing unified traces that contain global Meminfo and CPU tracks, you end up with two tracks with the same name. For example, you end up with two MemFree tracks and there is no way of distinguishing which track belongs to which machine. Added the machine id to the track name to disambiguate to which machine the track belongs to.

Bug: 421946530


jahdiel-alvarez:
Screenshot of the Meminfo tracks with the machine id:

image

stevegolton (approved):

PR #1856, opened 2025-06-16
stdlib: Add Wattson IRQ power attribution

wusamuel6:
Add IRQ power attribution in the same fashion as tasks power
attribution. With an unique identifying ID, each IRQ slice can leverage
the same framework that does idle transition power attribution and tasks
power attribution, since ultimately each task is also just a slice.

All the IRQ slices need to be flattened, such that there is only ever
one "slice" running on a CPU at a given point in time. The stack from
bottom to top is thread/process, soft IRQ, hard-IRQ-0, ..., hard-IRQ-n,
with each subsequent slice being fully contained within the previous
slice.

For the hard IRQs, use _intervals_flatten!() to flatten since those
slices inherently have the parent_id info and ancestor/child info
associated with them. To include soft IRQs and normal threads, use
SPAN_OUTER_JOIN() and an additional post-processing step to remove the
overlapping slices.

Bug: 422478917


src/trace_processor/perfetto_sql/stdlib/wattson/tasks/threads_w_processes.sql
```diff
@@ -38,3 +40,120 @@ LEFT JOIN android_process_metadata AS package
...
+CREATE PERFETTO TABLE _all_irq_slices AS
+SELECT
+  id,
+  ts,
+  dur,
+  parent_id,
+  extract_arg(arg_set_id, 'vec') IS NULL AS is_hard_irq
+FROM slice
+WHERE
+  category = 'irq';
```
LalitMaganti, line 54: This is not a good way to filter for irqs, it's very possible for other people to emit events with a category as generic as "irq".

Suggest instead joining with track and looking for track.type `cpu_softirq` and `cpu_irq`

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/threads_w_processes.sql
```diff
@@ -38,3 +40,120 @@ LEFT JOIN android_process_metadata AS package
...
   dur > 0;
+
+-- Gets all IRQs and labels each slice with hard or soft IRQ
+CREATE PERFETTO TABLE _all_irq_slices AS
+SELECT
+  id,
+  ts,
+  dur,
+  parent_id,
+  extract_arg(arg_set_id, 'vec') IS NULL AS is_hard_irq
```
LalitMaganti, line 51: This is a very inefficient way of doing this. If you do a union all of the two track types, it will be signifcantlyt more efficient.

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/threads_w_processes.sql
```diff
@@ -38,3 +40,120 @@ LEFT JOIN android_process_metadata AS package
   USING (upid)
 WHERE
   dur > 0;
+
+-- Gets all IRQs and labels each slice with hard or soft IRQ
+CREATE PERFETTO TABLE _all_irq_slices AS
```
LalitMaganti, line 45: Can we expose this as a new, public table in the linux package? Feels like it belongs there.

PR #1857, opened 2025-06-17
tp: change metric proto to bundle metrics together

LalitMaganti:
This is a specialized output format really designed for a single important internal customer but likely will be useful beyond that context (though I anticipate quite rarely).

Plumb this through all the relevant places.


protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -249,6 +266,21 @@ message TraceMetricV2TemplateSpec {
...
+  // If true, the specs generated by this template wuill not have
+  // `output_group_id` set.
+  //
+  // By default, template specs, when expanded, will also share a common
+  // `output_group_id` (equal to the `id_prefix` of the template). This means
+  // clients who want to to enable output grouping will receive all
+  // metrics generated by this template in a single group.
+  //
+  // By turning this off, metrics will instead be emitted individually.
+  optional bool disable_output_grouping = 7;
```
elrakadm: To clarify, disabling the output grouping off will return the output as TraceProcessor does today (splitting the template into multiple metric specs and query each individually) ?

protos/perfetto/trace_summary/file.proto
```diff
@@ -55,12 +55,20 @@ message TraceSummarySpec {
...
   // The result of computing a v2 trace metric.
   //
   // See documentation for TraceMetric for usage of this field.
   repeated TraceMetricV2 metric = 1;
 
+  // An alternative encoding of the v2 trace metric where metrics are grouped
+  // by their `group_id` field.
+  //
+  // See documentation for TraceMetricV2Group for usage of this field.
+  repeated TraceMetricV2Group metric_groups = 3;
```
elrakadm: I will suggest to update `TraceMetricV2` to have repeated values and repeated specs instead of creating a new proto output ? it can have one value or more depending on whether there is a group or not.

It will make reading trace summary output consistent.

src/trace_processor/trace_summary/summary_integrationtest.cc
```diff
@@ -206,5 +232,404 @@ TEST_F(TraceSummaryTest, DuplicateMetricIdFromTemplate) {
...
+          column_names: "value_a"
+          column_names: "value_b"
+        }
+      }
+    }
+  )");
+  ASSERT_TRUE(status_or_output.ok());
+  EXPECT_THAT(*status_or_output, EqualsIgnoringWhitespace(R"(
+    metric_bundles {
+      specs {
```
elrakadm, line 317: maybe out of scope of the cl : should this return the metric spec template instead to reduce the verbosity of the output in case of bundling ?
but I can see why it is required to expand the template incase the input summary spec had metrics with the same bundle id.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -135,235 +139,323 @@ base::Status WriteMetadata(TraceProcessor* processor,
...
+      bundle_id = m.id;
+    }
+    metrics_by_bundle[bundle_id].push_back(&m);
+  }
+  for (auto it = metrics_by_bundle.GetIterator(); it; ++it) {
+    RETURN_IF_ERROR(VerifyBundleHasConsistentSpecs(it.key(), it.value()));
+
+    const std::string& bundle_id = it.key();
+    auto* bundle = summary->add_metric_bundles();
+    for (const Metric* metric : it.value()) {
```
elrakadm: nit: it.value() is called multiple times, could you please create a variable for the vector maybe named `bundleMetrics` ?

src/trace_processor/trace_summary/summary.cc
```diff
@@ -135,235 +139,323 @@ base::Status WriteMetadata(TraceProcessor* processor,
...
 
-    if (spec_decoder.has_dimensions_specs() && spec_decoder.has_dimensions()) {
-      return base::ErrStatus(
-          "Both dimensions and dimension_specs defined for metric '%s'. Only "
-          "one is allowed",
-          metric_name.c_str());
+base::Status WriteDimension(
+    const DimensionWithIndex& dim_with_index,
+    const std::string& metric_or_bundle_name,
+    Iterator& it,
```
elrakadm: nit: rename to query_it or query_result_it

src/trace_processor/trace_summary/summary.cc
```diff
@@ -135,235 +139,323 @@ base::Status WriteMetadata(TraceProcessor* processor,
...
+      if (is_unique_dimensions && !seen_dimensions.Insert(hash, true).second) {
+        return base::ErrStatus(
+            "Duplicate dimensions found for metric bundle '%s': this is not "
+            "allowed",
+            bundle_id.c_str());
+      }
+      for (size_t i = 0; i < it.value().size(); ++i) {
+        const auto& metric_value_column = query_it.Get(value_indices[i]);
+        auto* row_value = row->add_values();
+        if (metric_value_column.is_null()) {
```
elrakadm, line 425: This changes the behavior a bit for unbundled metrics, where it will return null values instead of dropping them.

Should TP add rows only if all values in the row != null value ?

LalitMaganti: Added a check for all null rows.

src/trace_processor/trace_summary/summary_integrationtest.cc
LalitMaganti: I would really make templates transparent honestly as much as possible. I don't think verbosity is a big deal, this will only be repeated once per template, not once per row..

LalitMaganti:
Fixed, PTAL.

elrakadm (approved):

PR #1867, opened 2025-06-17
Add NamedTrack::Global static method

altimin:
This allows creating global named tracks similar to Track::Global
and CounterTrack::Global, with an optional id parameter.


altimin:
Not sure how to assign specific folks for review here, so let me just tag @LalitMaganti or @ddiproietto directly.

include/perfetto/tracing/track.h
```diff
@@ -231,6 +231,11 @@ class PERFETTO_EXPORT_COMPONENT NamedTrack : public Track {
     return NamedTrack(std::forward<TrackEventName>(name), id, parent);
   }
 
+  template <class TrackEventName>
+  static NamedTrack Global(TrackEventName name, uint64_t id = 0) {
```
ddiproietto: `TrackEventName&&`, this needs to be a forwarding reference.

(NamedTrack::ThreadScoped was broken as well, I had to send https://github.com/google/perfetto/pull/1879)

Can you also cover this in api_integrationtest.cc, please, just to make sure it compiles? (It would be broken now).

ddiproietto (commented):
Thanks!

include/perfetto/tracing/track.h
altimin: Sorry for the delay, done — PTAL!

src/tracing/test/api_integrationtest.cc
```diff
@@ -2221,22 +2221,32 @@ TEST_P(PerfettoApiTest, TrackEventCustomNamedTrack) {
   });
   thread.join();
 
+  const auto global_track = perfetto::NamedTrack::Global("GlobalTrack");
+  const auto global_track_with_id = perfetto::NamedTrack("MyCustomTrack", 1);
```
ddiproietto: This is not a global track, it's a named track, child of the process track, right? (The name seems and the comments below seem to suggest this is a global track)

LalitMaganti (approved):

PR #1869, opened 2025-06-17
Add usb group to predefined tracepoints

roychl666:
Create a usb group for DWC3 ftrace events.
DWC3 USB controller is a widely adopted SNPS IP, with linux upstream support and extensitve use by companies like QCOM, Samsung and Google. This would be helpful in debugging usb issues and analyzing usb performance.
The counterpart in atrace is in ag/34015648.

Bug: 396043704
Test: out/android/perfetto_unittests
Test: capture and validate the trace

### DIFF
--- a/src/traced/probes/ftrace/predefined_tracepoints.cc
+++ b/src/traced/probes/ftrace/predefined_tracepoints.cc
@@ -379,6 +379,24 @@ base::FlatSet<GroupAndName> GenerateCameraTracePoints(
   return events;
 }
 
+base::FlatSet<GroupAndName> GenerateUsbTracePoints() {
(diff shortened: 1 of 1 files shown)


primiano (changes requested):
So I looked more into this and I feel these should really not go here (neither in perfetto) and go in to the atrace HAL.

those points are very specific to one particular usb controller. why would we put them in atrace or perfetto? We can't list all possible drivers there (The ones that are there are mostly legacy).

These days the proper way to do it is via vendor tracepoints (/vendor/etc/atrace/atrace_categories.txt)

Look for atrace_categories.txt in the internal google repo. that's the place to change, not here

PR #1874, opened 2025-06-18
Show args on debug track details panel

stevegolton:
- Add a new option to the 'add debug track' menu to select a column to use as the arg set id.
- Move the existing 'args' in the details panels (which are really just a printout of the raw columns from the source query), to the left panel and call them 'Raw Columns', and print the real args is present into the right hand panel.

### DIFF
--- a/ui/src/components/tracks/debug_slice_track_details_panel.ts
+++ b/ui/src/components/tracks/debug_slice_track_details_panel.ts
@@ -14,9 +14,9 @@
 
 import m from 'mithril';
 import {duration, Time, time} from '../../base/time';
-import {hasArgs} from '../details/args';
+import {hasArgs, renderArguments} from '../details/args';
 import {getSlice, SliceDetails} from '../sql_utils/slice';
-import {asSliceSqlId, Utid} from '../sql_utils/core_types';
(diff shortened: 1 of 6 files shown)


ui/src/components/tracks/debug_slice_track_details_panel.ts
```diff
@@ -262,4 +298,71 @@ export class DebugSliceTrackDetailsPanel implements TrackEventDetailsPanel {
...
+        value,
+        displayValue: it.displayValue ?? 'NULL',
+      });
+    }
+
+    return args;
+  }
+}
+
+function parseArgValue(it: {
```
LalitMaganti: I feel like displayvalue already basically does this. Why do we need this in the UI?

LalitMaganti (approved):

ui/src/components/tracks/debug_slice_track_details_panel.ts
stevegolton: This is just to get the arguments in a standard format for pushing through the generic rendering function `renderArguments` though this and `loadArgs` are copy pasted actually so I'll just import the one from`sql_utils/args.ts` instead.

PR #1882, opened 2025-06-18
Add a data source which polls for CPU time per UID / cluster

simonmacm:
Bug: http://b/422976190


src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+
+  auto packet = writer_->NewTracePacket();
+  packet->set_timestamp(static_cast<uint64_t>(base::GetBootTimeNs().count()));
+
+  bool first_time;
+  if (previous_times_.size() == 0) {
+    first_time = true;
+    packet->set_sequence_flags(
+        protos::pbzero::TracePacket::SEQ_INCREMENTAL_STATE_CLEARED |
+        protos::pbzero::TracePacket::SEQ_NEEDS_INCREMENTAL_STATE);
```
ddiproietto: Nit: I think we can remove `| protos::pbzero::TracePacket::SEQ_NEEDS_INCREMENTAL_STATE`

(I realize now that the wakelock data source has it and I approved it, but I checked the semantic and it is redundant).

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+    uint32_t uid,
+    std::vector<uint64_t>& cluster_deltas_ms,
+    protozero::PackedVarInt& uid_list,
+    protozero::PackedVarInt& total_time_ms_list) {
+  bool write = false;
+  for (uint32_t cluster = 0; cluster < cluster_deltas_ms.size(); cluster++) {
+    if (cluster_deltas_ms[cluster] != 0) {
+      write = true;
+    }
+  }
```
ddiproietto:
```
  for (uint64_t value : cluster_deltas_ms) {
    if (value != 0) {
      write = true;
    }
  }
```

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+
+  if (write) {
+    uid_list.Append(uid);
+    for (uint32_t cluster = 0; cluster < cluster_deltas_ms.size(); cluster++) {
+      total_time_ms_list.Append(cluster_deltas_ms[cluster]);
+    }
```
ddiproietto:
```
  for (uint64_t value : cluster_deltas_ms) {
    total_time_ms_list.Append(value);
  }
```

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+  MaybeAppendUid(current_uid, cluster_deltas_ms, uid, total_time_ms);
+
+  if (first_time) {
+    proto->set_cluster_count(uint32_t(cluster_deltas_ms.size()));
+  }
+  proto->set_uid(uid);
+  proto->set_total_time_ms(total_time_ms);
+}
+
+void AndroidCpuPerUidDataSource::MaybeAppendUid(
```
ddiproietto: Nit: This doesn't need to be a method, it can be a function in the anonymous namespace in this file, right?

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+    if (time.uid != current_uid) {
+      if (current_uid != kInvalidUid) {
+        MaybeAppendUid(current_uid, cluster_deltas_ms, uid, total_time_ms);
+      }
+      current_uid = time.uid;
+      for (uint32_t i = 0; i < cluster_deltas_ms.size(); i++) {
+        cluster_deltas_ms[i] = 0L;
+      }
+    }
+
```
ddiproietto, line 200: Nit: I would protect against `time.cluster` being out of bounds.

In case the library returns wrong data, we don't want to crash.

```
if (time.cluster >= cluster_deltas_ms.size()) {
  // Data is corrupted
  continue;
}
```

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+  uint32_t current_uid = kInvalidUid;
+
+  // GetCpuTimes() returns values grouped by UID.
+  for (auto& time : cpu_times) {
+    if (first_uid == kInvalidUid) {
+      first_uid = time.uid;
+    }
+
+    if (time.uid == first_uid) {
+      cluster_deltas_ms.push_back(0L);
```
ddiproietto, line 187: I would add a comment that explains that we're using the first uid to understand the number of clusters

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    if (time.uid != current_uid) {
+      if (current_uid != kInvalidUid) {
+        MaybeAppendUid(current_uid, cluster_deltas_ms, uid, total_time_ms);
+      }
+      current_uid = time.uid;
+      for (uint32_t i = 0; i < cluster_deltas_ms.size(); i++) {
+        cluster_deltas_ms[i] = 0L;
+      }
```
ddiproietto, line 198: Nit:

```
for (uint64_t& val : cluster_delta_ms) {
  val = 0;
}
```

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
```diff
@@ -0,0 +1,227 @@
...
+  } else {
+    first_time = false;
+    packet->set_sequence_flags(
+        protos::pbzero::TracePacket::SEQ_NEEDS_INCREMENTAL_STATE);
+  }
+
+  auto* proto = packet->set_cpu_per_uid_data();
+
+  protozero::PackedVarInt uid;
+  protozero::PackedVarInt total_time_ms;
```
ddiproietto: Maybe let's rename these to uid_list and total_time_ms_list? I was confused below when seeing just `uid`

ddiproietto (approved):
Thanks, just nits.

I was thinking what happens with concurrent data sources. I think our code should be fine. Will the library handle correctly multiple calls to android::bpf::getUidsUpdatedCpuFreqTimes with different last_update_ms? I.e. is the function fully reentrant, does it avoid global variables and are the reads non destructive?

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

src/traced/probes/android_cpu_per_uid/android_cpu_per_uid_data_source.cc
simonmacm: Done.

PR #1884, opened 2025-06-19
Decentralize dataset-search into plugins

stevegolton:
This patch introduces a new concept of 'search providers', which provide a way for plugins to extend the search interface subsystem.

This patch also includes modifications to plugins in order to bring the functionality of the dataset search mechanism in line with the functionality of the original search.

That is to say we search:
  - cpu slices by process or thread name
  - slices by name, id or arg string/key
  - android logs by message

### DIFF
--- a/ui/src/plugins/dev.perfetto.TraceProcessorTrack/index.ts
+++ b/ui/src/plugins/dev.perfetto.TraceProcessorTrack/index.ts
@@ -45,6 +45,7 @@ import {
   QueryFlamegraph,
 } from '../../components/query_flamegraph';
 import {AreaSelection, areaSelectionsEqual} from '../../public/selection';
+import {escapeSearchQuery} from '../../trace_processor/query_utils';
 
 export default class implements PerfettoPlugin {
   static readonly id = 'dev.perfetto.TraceProcessorTrack';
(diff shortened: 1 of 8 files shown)


ui/src/plugins/dev.perfetto.TraceProcessorTrack/index.ts
```diff
@@ -444,6 +446,68 @@ export default class implements PerfettoPlugin {
...
+          .filter((t) =>
+            t.renderer.getDataset?.()?.implements({arg_set_id: NUM_NULL}),
+          );
+      },
+      async getSearchFilter(searchTerm) {
+        const searchLiteral = escapeSearchQuery(searchTerm);
+        return {
+          join: `args USING(arg_set_id)`,
+          where: `
+            args.string_value GLOB ${searchLiteral}
```
LalitMaganti, line 503: Should we not also search int and double values as well?

stevegolton: Maybe in the future, for now I'm just emulating exactly what the existing search does.
https://github.com/google/perfetto/blob/main/ui/src/core/search_manager.ts#L240

LalitMaganti: Ack.

PR #1896, opened 2025-06-20
Add a priority boost support to Perfetto.

KirillTim:
This CL adds a feature to conditionally change the priority of the `traced` process
when the trace is running.

This feature is only supported on Android and Linux,
and requires the `traced` to have 'CAP_SYS_NICE' capability (be started as root).

This is the first part of a bigger change.

Bug: b/423561591
Tested: ./perfetto_unittests


include/perfetto/ext/base/sched.h
```diff
@@ -0,0 +1,225 @@
...
+        // For SCHED_IDLE kernel priority is always 120
+        return kKernelDefaultPrio;
+      case SchedPolicy::kOther:
+      case SchedPolicy::kBatch:
+        return static_cast<unsigned int>(kKernelDefaultPrio + nice_);
+      case SchedPolicy::kFifo:
+      case SchedPolicy::kRr:
+        return kKernelMaxRtPrio - 1 - rt_priority_;
+    }
+    PERFETTO_FATAL("Can't be here (Unknown sched policy enum value %d)",
```
primiano, line 119: is this needed? the compiler should realized you are exhausting all the switch cases. just return 0 (Add a dcheck if you want) but don't bloat the binary plz

include/perfetto/ext/base/sched.h
```diff
@@ -0,0 +1,225 @@
...
+#include "perfetto/ext/base/no_destructor.h"
+#include "perfetto/ext/base/status_or.h"
+
+namespace perfetto::base {
+
+struct SchedConfig {
+  enum class SchedPolicy { kIdle, kBatch, kOther, kFifo, kRr };
+  static SchedConfig CreateIdle() { return {SchedPolicy::kIdle, 0, 0}; }
+  static SchedConfig CreateDefaultUserspacePolicy() { return CreateOther(0); }
+  static SchedConfig CreateOther(int nice) {
```
primiano, line 36: why do you have thse convenience methods? these are pure boilerplate.
Can you give me one reason why one should prefer these other than just using the constructor?

These static methods make sense if you have somehting that, for instance, returns a unique_ptr, where you want to return false if somethign fails. 
I don't see a single reason why these static methods should exist

include/perfetto/ext/base/sched.h
```diff
@@ -0,0 +1,225 @@
...
+   * If both priorities have the same `KernelPriority()` value, we compare the
+   * policies.
+   *
+   * Strictly speaking, it is not correct to compare realtime priorities
+   * 'SCHED_FIFO' and 'SCHED_RR' based on their policy, but we do it for
+   * simplicity.
+   */
+  bool operator<(const SchedConfig& other) const {
+    const int inverted_priority =
+        -1 * static_cast<int>(InternalUnifiedPriority());
```
primiano, line 84: Nothing drives me more mad than seeing functions being used only in one place.
What's the reason this InternalUnifiedPriority is a standalone function rather than being code inlined here? It hurts readability. Now if I want to understand waht this function does I have to scroll down and this interrupts my flow. Why?
Code gets read way more times than it gets written. please optimize your code for readability.

include/perfetto/ext/base/sched.h
```diff
@@ -0,0 +1,225 @@
...
+    if (priority >= kMinPriority && priority <= kMaxPriority) {
+      return OkStatus();
+    }
+    return ErrStatus("Invalid priority: %d. Valid range is [%d, %d]", priority,
+                     kMinPriority, kMaxPriority);
+  }
+
+ private:
+  unsigned int InternalUnifiedPriority() const {
+    if (policy_ == SchedPolicy::kIdle) {
```
primiano, line 143: Why this is not part of kernelPriority? why idle is special? WHy are we worrying about idle at all? who is ever going to use it?
Please introduce these things only after there is a valid use case.
You are bloating the codebase with code that is never going to be actively used. that's the worse thing for a project: you see code that you think does something, turns out is never used, you try to use it and might find bugs.
Introduce the code that you need.
Perfetto's base is NOT meant to be, and will never be, a general purpose library like ABSL. plese introduce strictly only the code you need.
All this stuff hurts mainteinability of the codebase. When poeple touch files around they need to ask "uh who is using kIdle?" and then when they don't find the answer they will start looking into google3 and other repos thinking: maybe somebody outside uses it. and they will waste time in this code search

include/perfetto/ext/base/sched.h
```diff
@@ -0,0 +1,225 @@
...
+              const unsigned int priority,
+              const int nice)
+      : policy_(policy), rt_priority_(priority), nice_(nice) {
+    switch (policy) {
+      case SchedPolicy::kIdle:
+        PERFETTO_DCHECK(priority == 0 && nice == 0);
+        break;
+      case SchedPolicy::kOther:
+      case SchedPolicy::kBatch:
+        PERFETTO_DCHECK(ValidateNiceValue(nice).ok() && priority == 0);
```
primiano, line 59: why. please don't create logic that shadows the kernel logic. the kernel will return EINVAL if you get this wrong.

include/perfetto/ext/base/sched.h
```diff
@@ -0,0 +1,225 @@
...
+        break;
+      case SchedPolicy::kFifo:
+      case SchedPolicy::kRr:
+        PERFETTO_DCHECK(ValidatePriority(priority).ok() && nice == 0);
```
primiano, line 63: ditto

include/perfetto/ext/base/sched.h
```diff
@@ -0,0 +1,225 @@
...
+
+  static Status ValidateNiceValue(const int nice) {
+    if (nice >= kMinNice && nice <= kMaxNice) {
+      return OkStatus();
+    }
+    return ErrStatus("Invalid nice value: %d. Valid range is [%d, %d]", nice,
+                     kMinNice, kMaxNice);
+  }
+
+  static Status ValidatePriority(const unsigned int priority) {
```
primiano, line 133: remove

src/tracing/service/tracing_service_impl.cc
```diff
@@ -340,21 +342,24 @@ std::unique_ptr<TracingService> TracingService::CreateInstance(
...
+  return std::make_unique<TracingServiceImpl>(
+      std::move(shm_factory), task_runner, std::move(deps),
+      base::SchedManager::GetInstance(), init_opts);
 }
 
 TracingServiceImpl::TracingServiceImpl(
     std::unique_ptr<SharedMemory::Factory> shm_factory,
     base::TaskRunner* task_runner,
     tracing_service::Dependencies deps,
+    base::SchedManagerInterface& sched_manager,
```
primiano, line 354: this should be passed in tracing_service::Dependencies. There is precendent there to do precisely what you want to do,.
You should pass a non-null sched_maanger only in tests, and by deafault construct the right one here

src/tracing/service/tracing_service_impl.cc
```diff
@@ -934,6 +939,38 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
     }
   }
 
+  std::optional<base::SchedConfig> priority_boost;
```
primiano, line 942: too much code. this should just be

if (cfg.has_priority_boost()) {
  tracing_sesison->priority_boost = ...
  auto res = UpdateCurrentSchedPolicyIfNeeded()
 if (!res) {
    tracing_sessions_.erase(tsid);
   return res;
} 
}

src/tracing/service/tracing_service_impl.cc
```diff
@@ -1383,6 +1425,57 @@ void TracingServiceImpl::StopOnDurationMsExpiry(TracingSessionID tsid) {
   // after |trace_duration_ms| unconditionally.
   FlushAndDisableTracing(tsid);
 }
+base::StatusOr<base::SchedConfig>
+TracingServiceImpl::CreateSchedPolicyFromConfig(
+    const protos::gen::PriorityBoostConfig& config) {
+  if (config.has_nice_value() && config.has_realtime_priority()) {
```
primiano, line 1431: never realy on the has_xxx methods for PODs. You don't want to get into the "is this 0" vs "is this unset". it's a protobuf reccomendation. tools in the middle might accidentally replace unset values with 0. There should never be any sematnic difference between 0 and "Unset" for integers or booleans in protos.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -1383,6 +1425,57 @@ void TracingServiceImpl::StopOnDurationMsExpiry(TracingSessionID tsid) {
...
+    // Android uses 'SCHED_OTHER' as a default userspace priority, and so do we.
+    return base::SchedConfig::CreateOther(config.nice_value());
+  }
+  if (config.has_realtime_priority()) {
+    RETURN_IF_ERROR(
+        base::SchedConfig::ValidatePriority(config.realtime_priority()));
+    // Android uses 'SCHED_FIFO' as a rt-priority, and so do we.
+    return base::SchedConfig::CreateFifo(config.realtime_priority());
+  }
+  return base::ErrStatus(
```
primiano, line 1447: why. just leave it to the default.

src/tracing/service/tracing_service_impl.h
```diff
@@ -946,6 +954,11 @@ class TracingServiceImpl : public TracingService {
   uint64_t chunks_discarded_ = 0;
   uint64_t patches_discarded_ = 0;
 
+  base::SchedManagerInterface& sched_manager_;
+  // This field having value implies that the current platform supports sched
+  // policy updates.
+  std::optional<base::SchedConfig> default_sched_policy_;
```
primiano, line 960: why do we do this? if somebody changes the default priority of traced, and then uses the boost, they are asking for troubles. please let' skeep things simple. Just get back to the default.
This responsiblity should be really be encapsulated in its own class (rather than having 2 classes here) as you need precisely the same logic in traced_probes

primiano (changes requested):
This CL has both too much code, and not enough code.
Some overall comments:

- I'd encapsulate the "muxing differnt priorities" into its own class. Have a class where you can do AddPriority() and that returns a RAII object (which you then put in TracingSession) than when destroyed removes the priority from the list.
- That's the only thing you need to test, and it is nicer to test
- You don't need any test in tracing_Service_impl
- then you use this class both in traced and traced_probes
- then you have an e2e test that reads back the ftrace data to check if the prio of both processes is actually boosted.

include/perfetto/ext/base/sched.h
KirillTim: > is this needed?

This is needed for the GCC.
It can't deduce from the code flow that the function returns on each control flow path, see: https://godbolt.org/z/sorvzzvda
and also other places in our code base: https://source.corp.google.com/search?q=f:external%2Fperfetto%20%22For%20gcc%22&sq=git:googleplex-android%2Fplatform%2Fsuperproject%2Fmain@main

> just return 0 

in our code base we not usually doing this

> but don't bloat the binary plz

you are talking about the size of the error message being add to the `.text` session? Should I just use the shorted `"For GCC"` instead?

src/tracing/service/tracing_service_impl.cc
KirillTim: > There should never be any sematnic difference between 0 and "Unset" for integers or booleans in protos.

What is the right way then to check if there is a value `nice_value` in the `config`? Zero is also a valid `nice_value` value, that can be set in the config proto.

src/tracing/service/tracing_service_impl.cc
KirillTim: If there is an empty `PriorityBoostConfig {}` message in the config, then there is an error in the config.
Why should we silently ignore the error and doing nothing instead of reporting the error?

src/tracing/service/tracing_service_impl.cc
KirillTim: After reading go/totw/199 and your next comment about the default value, I am still not sure of the correct workflow to obtain the config.
Should it be like this:
```
if (cfg.has_priority_boost()) {
  const auto& boost = cfg.priority_boost();
    if (boost.realtime_priority() != 0) {
      // use realtime priority
    } else {
      int nice = boost.nice_value();
      // doesn't matter if the nice_value was set to zero or just missing,
      // we create a userspace priority using nice
      // (nice=0 is a default for the userspace priority anyway)
    }
} else {
  // priority_boost wasn't requested
}
```

include/perfetto/ext/base/sched.h
KirillTim: The kernel indeed will return EINVAL if we pass incorrect value, or combination of `policy` and `nice`/`priority`.
However, before we call the kernel, we need to find the max priority that we want to set. And to correctly find the maximum priority we want to be sure the the  data stored inside `SchedConfig` is correct.
So we validate invariant in constructor with `PERFETTO_DCHECK` and use the separate `ValidateNiceValue` and `ValidatePriority` method to also check the `priority_boost` config we got from the user.

include/perfetto/ext/base/sched.h
KirillTim: > What's the reason this InternalUnifiedPriority is a standalone function rather than being code inlined here?

You suggest to replace it with a lambda, being placed inside the body of the `operator<` function? 
something like 
```
bool operator<(const SchedConfig& other) const {
    const auto inverted_prio = [](const SchedConfig& config) {
      if (config.policy_ == SchedPolicy::kIdle) {
        return -1 * (kKernelMaxPrio + 1);
      }
      return -1 * static_cast<int>(config.KernelPriority());
    };
    return std::forward_as_tuple(inverted_prio(*this), policy_) <
           std::forward_as_tuple(inverted_prio(other), other.policy_);
}
```
?
It seems to me not very idiomatic C++, TBH

include/perfetto/ext/base/sched.h
KirillTim: > Why this is not part of kernelPriority? why idle is special?

The idle is special because of the way the kernel represents it:
```
# normal prio, nice = 0
$cat /proc/self/sched | grep -e prio -e policy 
policy                                       :                    0
prio                                         :                  120

# normal prio, change nice to 19
$ nice -n 19 cat /proc/self/sched | grep -e prio -e policy
policy                                       :                    0
prio                                         :                  139
# idle prio
sudo chrt --idle 0 cat /proc/self/sched | grep -e prio -e policy
policy                                       :                    5
prio                                         :                  120
```
For the IDLE priority the `prio` value (used by kernel to sort both realtime and userspace priorities) is 120, equals to the `prio` for the default policy (SHED_OTHER).
So, to simplify the compare logic we set the value that is lower that the lowest `prio`, used by kernel.


> WHy are we worrying about idle at all? who is ever going to use it?

There is a fact that complicates the priority boost code significantly: we don't fully control the environment in which we are running.

From https://man7.org/linux/man-pages/man7/sched.7.html:
>  SCHED_IDLE: Scheduling very low priority jobs

We only partially control the way our `traced`/`traced_probs` daemons are started on android. What if some vendor will change defaults (are they allowed to do so?) and will start our background daemons with the lowest possible priority for some reason? Or what if the system for some obscure reason decide to change the priority of `traced` after some time it was inactive?

We control the policy we allow user to configure and set (we use FIFO for realtime and OTHER for userspace), but not the policy we are running in.

We can say "IDLE policy is never used on Android, if the current policy is IDLE just report an error", but why not just support it instead, since it costs nothing?

My point is: we implement a tiny subset of the kernel scheduling logic to apply to our process, so we need to implement it fully and correctly.

> Introduce the code that you need.

For the IDLE priority we add like 15-30 lines of the code, but maintain the correctness, I think it is a nice deal.


> All this stuff hurts mainteinability of the codebase. When poeple touch files around they need to ask "uh who is using kIdle?" and then when they don't find the answer they will start looking into google3 and other repos thinking: maybe somebody outside uses it. and they will waste time in this code search

I agree with this, I think if I add the comment that explains why we have the code there it fix the problem.

src/tracing/service/tracing_service_impl.cc
KirillTim: I agree, will fix it.

priyanshscpp (changes requested):
All GPTfied

PR #1897, opened 2025-06-20
Winscope rect and transform trackers

PriyankaSPX:
Winscope rect and transform trackers.
    
For parsing SF/VC/WM data into rect tables.
    
Bug: 411363817

### DIFF
--- a/src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
+++ b/src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
@@ -27,7 +27,7 @@
 #include "src/trace_processor/tables/winscope_tables_py.h"
 #include "src/trace_processor/types/trace_processor_context.h"
 
-namespace perfetto::trace_processor {
+namespace perfetto::trace_processor::winscope {
 
 enum ProtoLogLevel : int32_t {
   DEBUG = 1,
@@ -81,6 +81,6 @@ class ProtoLogMessageDecoder {
   base::FlatHashMap<uint64_t, TrackedMessage> tracked_messages_;
 };
 
-}  // namespace perfetto::trace_processor
+}  // namespace perfetto::trace_processor::winscope
 
 #endif  // SRC_TRACE_PROCESSOR_IMPORTERS_PROTO_WINSCOPE_PROTOLOG_MESSAGE_DECODER_H_
--- a/src/trace_processor/importers/proto/winscope/winscope_geometry.cc
+++ b/src/trace_processor/importers/proto/winscope/winscope_geometry.cc
@@ -31,18 +31,24 @@ static bool IsFloatClose(double a, double b) {
 }
 }  // namespace
 
+Rect::Rect() = default;
+
 Rect::Rect(const protos::pbzero::RectProto::Decoder& rect) {
   x = rect.has_left() ? rect.left() : 0;
   y = rect.has_top() ? rect.top() : 0;
-  w = rect.has_right() ? rect.right() : 0;
-  h = rect.has_bottom() ? rect.bottom() : 0;
+  auto right = rect.has_right() ? rect.right() : 0;
+  auto bottom = rect.has_bottom() ? rect.bottom() : 0;
+  w = right - x;
+  h = bottom - y;
 }
 
 Rect::Rect(const protos::pbzero::FloatRectProto::Decoder& rect) {
   x = rect.has_left() ? static_cast<double>(rect.left()) : 0;
   y = rect.has_top() ? static_cast<double>(rect.top()) : 0;
-  w = rect.has_right() ? static_cast<double>(rect.right()) : 0;
-  h = rect.has_bottom() ? static_cast<double>(rect.bottom()) : 0;
+  auto right = rect.has_right() ? static_cast<double>(rect.right()) : 0;
+  auto bottom = rect.has_bottom() ? static_cast<double>(rect.bottom()) : 0;
+  w = right - x;
+  h = bottom - y;
 }
 
 Rect::Rect(double left, double top, double right, double bottom) {
@@ -65,7 +71,7 @@ Rect Rect::CropRect(const Rect& other) {
   const auto min_right = std::min(x + w, other.x + other.w);
   const auto max_top = std::max(y, other.y);
   const auto min_bottom = std::min(y + h, other.y + other.h);
-  return Rect{max_left, max_top, min_right - max_left, min_bottom - max_top};
+  return Rect(max_left, max_top, min_right, min_bottom);
 }
 
 bool Rect::ContainsRect(const Rect& other) {
@@ -93,8 +99,7 @@ bool Rect::IntersectsRect(const Rect& other) {
     if (y + h > other.y + other.h) {
       new_h = other.h;
     }
-
-    return !Rect{new_x, new_y, new_w, new_h}.IsEmpty();
+    return !Rect(new_x, new_y, new_w + new_x, new_h + new_y).IsEmpty();
   }
   return false;
 }
@@ -127,12 +132,8 @@ Rect TransformMatrix::TransformRect(const Rect& r) {
   const auto rb_prime = TransformMatrix::TransformPoint({r.x + r.w, r.y + r.h});
   const auto x = std::min(lt_prime.x, rb_prime.x);
   const auto y = std::min(lt_prime.y, rb_prime.y);
-  return Rect{
-      x,
-      y,
-      std::max(lt_prime.x, rb_prime.x) - x,
-      std::max(lt_prime.y, rb_prime.y) - y,
-  };
+  return Rect(x, y, std::max(lt_prime.x, rb_prime.x),
+              std::max(lt_prime.y, rb_prime.y));
 }
 
 Region TransformMatrix::TransformRegion(Region region) {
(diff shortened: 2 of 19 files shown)


LalitMaganti:
Before landing this CL, I'd like to discuss what I mentioned a while back re making Winscope an optional module of trace processor which only you folks enable and everyone else does not pay the cost for.

The main thing this involves is ifdeffing all the parts of the code which correspond to Winscope (really should just be the modules and the tables), put them behind a GN flag and then disable by default (and you can enable it when you build the WASM module for Winscope).

You can see how this works for ETM. If you code search for `enable_perfetto_etm_importer` you'll see how the build files are organized and if you search for `PERFETTO_ENABLE_ETM_IMPORTER` you can see how that corresponds to the C++ code.

I would ask that we do this before landing the 4000+ LOC you are planning on land in the coming CLs.

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -27,7 +27,7 @@
 #include "src/trace_processor/tables/winscope_tables_py.h"
 #include "src/trace_processor/types/trace_processor_context.h"
 
-namespace perfetto::trace_processor {
+namespace perfetto::trace_processor::winscope {
```
PriyankaSPX, line 30: i've put WinscopeContext, ProtologMessageDecoder and ShellTransitionsTracker inside the winscope namespace (along with the two new trackers). Left the rest of the parsers and WinscopeModule for now.

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.cc
```diff
@@ -0,0 +1,48 @@
...
+#include "src/trace_processor/importers/proto/winscope/winscope_transform_tracker.h"
+
+#include <algorithm>
+#include "src/trace_processor/storage/trace_storage.h"
+
+namespace perfetto::trace_processor::winscope {
+
+tables::WinscopeTransformTable::Id* WinscopeTransformTracker::GetOrInsertRow(
+    geometry::TransformMatrix& matrix) {
+  auto existing_matrix = std::find_if(
```
LalitMaganti: Do you really want to do a linear scan every time you call this function? Why is this not using a base::FlatHashMap?

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.cc
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+#include "src/trace_processor/importers/proto/winscope/winscope_transform_tracker.h"
+
+#include <algorithm>
+#include "src/trace_processor/storage/trace_storage.h"
+
+namespace perfetto::trace_processor::winscope {
+
+tables::WinscopeTransformTable::Id* WinscopeTransformTracker::GetOrInsertRow(
```
LalitMaganti: Why return a pointer? Why is this not just returning the id by value?

src/trace_processor/importers/proto/winscope/winscope_rect_tracker.cc
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#include "src/trace_processor/importers/proto/winscope/winscope_rect_tracker.h"
+
+#include <algorithm>
+#include "src/trace_processor/storage/trace_storage.h"
+
+namespace perfetto::trace_processor::winscope {
+
+tables::WinscopeRectTable::Id* WinscopeRectTracker::GetOrInsertRow(
```
LalitMaganti: All the same quesitons as below.

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
LalitMaganti: Ack.

LalitMaganti (changes requested):
Please fix the PR title.

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.cc
PriyankaSPX: changed to return the Id itself (assuming you don't mean Id.value)

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.cc
PriyankaSPX: originally wasn't that many rects - but yeah recognise that's much better as we scale this, changed to use base::FlatHashMap, thank you!

src/trace_processor/importers/proto/winscope/winscope_geometry.cc
```diff
@@ -31,18 +31,24 @@ static bool IsFloatClose(double a, double b) {
 }
 }  // namespace
 
+Rect::Rect() {};
```
PriyankaSPX: fixed some bugs in this file. can separate if you'd rather review separately.

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.cc
```diff
@@ -0,0 +1,44 @@
...
+
+#include "src/trace_processor/importers/proto/winscope/winscope_transform_tracker.h"
+
+#include <algorithm>
+#include "src/trace_processor/storage/trace_storage.h"
+
+namespace perfetto::trace_processor::winscope {
+
+tables::WinscopeTransformTable::Id WinscopeTransformTracker::GetOrInsertRow(
+    geometry::TransformMatrix& matrix) {
```
LalitMaganti: Why is this not const reference?

src/trace_processor/importers/proto/winscope/winscope_rect_tracker.cc
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#include "src/trace_processor/importers/proto/winscope/winscope_rect_tracker.h"
+
+#include <algorithm>
+#include "src/trace_processor/storage/trace_storage.h"
+
+namespace perfetto::trace_processor::winscope {
+
+tables::WinscopeRectTable::Id WinscopeRectTracker::GetOrInsertRow(
```
LalitMaganti: Do we really need two different trackers? Why not one geometrt tracker?

PriyankaSPX: wanted to separate them so it's clear that one is only affecting the rects table in storage, the other only the transform table - the transform tracker is only going to be used by the SF layers parser at this stage, whereas the rects tracker will be used by SF/ViewCapture/WindowManager

I didn't want to import a bunch of transform stuff that isn't necessary for the other traces

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.cc
PriyankaSPX: changed

src/trace_processor/importers/proto/winscope/winscope_geometry.cc
```diff
@@ -31,18 +31,24 @@ static bool IsFloatClose(double a, double b) {
 }
 }  // namespace
 
+Rect::Rect() {}
```
LalitMaganti: nit: = default

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.h
```diff
@@ -0,0 +1,60 @@
...
+ private:
+  base::FlatHashMap<geometry::TransformMatrix,
+                    tables::WinscopeTransformTable::Id>
+      rows_;
+};
+
+}  // namespace perfetto::trace_processor::winscope
+
+template <>
+struct std::hash<
```
LalitMaganti: instead of implmenting std::hash, can you have a hasher class instead and pass it as a hahser to FlatHashMap?

src/trace_processor/importers/proto/winscope/winscope_rect_tracker.h
```diff
@@ -0,0 +1,55 @@
...
+  const tables::WinscopeRectTable::Id& GetOrInsertRow(geometry::Rect& rect);
+
+ private:
+  base::FlatHashMap<geometry::Rect, tables::WinscopeRectTable::Id> rows_;
+};
+
+}  // namespace perfetto::trace_processor::winscope
+
+template <>
+struct std::hash<::perfetto::trace_processor::winscope::geometry::Rect> {
```
LalitMaganti: instead of implmenting std::hash, can you have a hasher class instead and pass it as a hahser to FlatHashMap?

src/trace_processor/importers/proto/winscope/winscope_transform_tracker.h
PriyankaSPX: done

src/trace_processor/importers/proto/winscope/winscope_geometry.cc
PriyankaSPX: done

src/trace_processor/importers/proto/winscope/winscope_rect_tracker.h
PriyankaSPX: done - any particular reason why? i see both done in the codebase

PR #1898, opened 2025-06-20
Add layer utils.

PriyankaSPX:
Utils for parsing SF data into extended surfaceflinger_layer table.

Bug: 411363817

### DIFF
--- a/src/trace_processor/importers/proto/winscope/winscope_module.h
+++ b/src/trace_processor/importers/proto/winscope/winscope_module.h
@@ -65,7 +65,7 @@ class WinscopeModule : public ProtoImporterModule {
   winscope::WinscopeContext context_;
   util::ProtoToArgsParser args_parser_;
 
-  SurfaceFlingerLayersParser surfaceflinger_layers_parser_;
+  winscope::SurfaceFlingerLayersParser surfaceflinger_layers_parser_;
   SurfaceFlingerTransactionsParser surfaceflinger_transactions_parser_;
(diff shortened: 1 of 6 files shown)


src/trace_processor/importers/proto/winscope/surfaceflinger_layer_utils.h
```diff
@@ -0,0 +1,253 @@
...
+#include "protos/perfetto/trace/android/surfaceflinger_layers.pbzero.h"
+#include "src/trace_processor/importers/proto/winscope/winscope_rect.h"
+#include "src/trace_processor/importers/proto/winscope/winscope_transform.h"
+
+namespace perfetto {
+namespace trace_processor {
+
+using TransformDecoder = protos::pbzero::TransformProto::Decoder;
+
+class SfTransform {
```
LalitMaganti: This seems like a very "Java" way of writing this code. In C++, if all functions in a class are static, we make it a namespace intstead.

PriyankaSPX: have put all the utils in the winscope::layer, winscope::display or winscope::transform namespaces

src/trace_processor/importers/proto/winscope/winscope_module.h
PriyankaSPX, line 1: I want to move all the parsers and other classes which are used only by WinscopeModule into the winscope namespace. Started by moving the SF layers parser here.

PR #1900, opened 2025-06-20
Make Winscope an optional TP module.

PriyankaSPX:
Bug: 411363817


gn/perfetto.gni
```diff
@@ -360,6 +360,10 @@ declare_args() {
   # builds.
   enable_perfetto_etm_importer =
       is_linux && enable_perfetto_trace_processor && perfetto_build_standalone
+
+  # Enables Winscope features in trace_processor.
+  enable_perfetto_winscope =
```
PriyankaSPX, line 367: Not sure exactly which builds you want to enable by default. We'll need it for the WASM for Winscope, plus for the Flicker prebuilts

src/trace_processor/importers/proto/additional_modules.cc
```diff
@@ -39,7 +40,9 @@
 #include "src/trace_processor/importers/proto/trace.descriptor.h"
 #include "src/trace_processor/importers/proto/translation_table_module.h"
 #include "src/trace_processor/importers/proto/v8_module.h"
+#if PERFETTO_BUILDFLAG(PERFETTO_ENABLE_WINSCOPE)
 #include "src/trace_processor/importers/proto/winscope/winscope_module.h"
```
LalitMaganti, line 45: Please separate these similar to other ifdefs.

src/trace_processor/importers/proto/additional_modules.cc
```diff
@@ -60,7 +63,9 @@ void RegisterAdditionalModules(TraceProcessorContext* context) {
   context->modules.emplace_back(new AndroidCameraEventModule(context));
   context->modules.emplace_back(new MetadataModule(context));
   context->modules.emplace_back(new V8Module(context));
+#if PERFETTO_BUILDFLAG(PERFETTO_ENABLE_WINSCOPE)
```
LalitMaganti: Please put at bottom of this list, separate from the rest.

src/trace_processor/storage/trace_storage.h
```diff
@@ -57,9 +57,11 @@
 #include "src/trace_processor/tables/trace_proto_tables_py.h"
 #include "src/trace_processor/tables/track_tables_py.h"
 #include "src/trace_processor/tables/v8_tables_py.h"
-#include "src/trace_processor/tables/winscope_tables_py.h"
 #include "src/trace_processor/types/destructible.h"
 #include "src/trace_processor/types/variadic.h"
+#if PERFETTO_BUILDFLAG(PERFETTO_ENABLE_WINSCOPE)
```
LalitMaganti, line 63: As above.

src/trace_processor/types/trace_processor_context.h
```diff
@@ -170,6 +168,10 @@ class TraceProcessorContext {
   std::unique_ptr<Destructible> etm_tracker;                            // EtmTracker
   std::unique_ptr<Destructible> elf_tracker;                            // ElfTracker
   std::unique_ptr<Destructible> file_tracker;                           // FileTracker
+#if PERFETTO_BUILDFLAG(PERFETTO_ENABLE_WINSCOPE)
```
LalitMaganti, line 172: Can we move these to the WinscopeModule instead of living here? there's no reason for this to be here if it's only used by one module.

gn/perfetto.gni
LalitMaganti: You should set it to false at https://cs.android.com/android/platform/superproject/main/+/main:external/perfetto/infra/luci/recipes/perfetto.py;l=75 (while you're there, please also set `enable_perfetto_etm_importer` to false and `is_clang` to true).

LalitMaganti: You should also add !perfetto_build_generator.

tools/gen_bazel
```diff
@@ -56,6 +56,7 @@ gn_args_base = {
     'enable_perfetto_traced_perf': 'false',
     'perfetto_force_dcheck': '"off"',
     'enable_perfetto_llvm_demangle': 'true',
+    'enable_perfetto_winscope': 'false',
```
LalitMaganti: You should not need this if you do above.

tools/gen_bazel
```diff
@@ -56,6 +56,7 @@ gn_args_base = {
...
     # TODO(lalitm): we are disabling these temporarily because Studio cannot
     # build them and the we don't have a way to enable/disable these properly
     # at the Bazel level.
```
LalitMaganti, line 61: Since you're here, please remove the etm ones as well and do the same as above (add !perfetto_build_generator in the gn args above).

src/trace_processor/types/trace_processor_context.h
PriyankaSPX: sure - mind if i do this in a followup? i'm adding a couple more trackers in a future CL, so could move them all at the same time

LalitMaganti: Yeah sure that works.

LalitMaganti (approved):

PR #1908, opened 2025-06-23
Make Winscope parsing module optional.

PriyankaSPX:
Tables and SQL should remain in all builds for now.

Bug: 411363817


tools/gen_bazel
```diff
@@ -61,7 +61,6 @@ gn_args_base = {
     # at the Bazel level.
     # Reenable these once we have a way to enable/disable them in Bazel.
     'enable_perfetto_trace_processor_mac_instruments': 'false',
-    'enable_perfetto_etm_importer': 'false',
```
LalitMaganti, line 64: Why is this being removed?

PriyankaSPX: you requested it in https://github.com/google/perfetto/pull/1900/files#diff-ce3a531998ff893995e345cf259fa96df15356db29f1c558601be4141e0caf9dL64

happy to put it back if that was also causing failures?

LalitMaganti: Ah my bad forgot it's moved up there.

Can you also do the same for the mac instruments as well?

PriyankaSPX: sure - removed the comment as well

gn/perfetto.gni
```diff
@@ -359,7 +359,14 @@ declare_args() {
   # this point and it requires extra libraries so we only enable it in selected
```
LalitMaganti, line 359: You need to add !perfetto_build_generator to instruments here too.

PriyankaSPX: done - ptal

LalitMaganti (approved):

PR #1911, opened 2025-06-23
Introduce winscope context.

PriyankaSPX:
For pointers to winscope-specific trackers that were previously in TraceProcessorContext.

In a separate WinscopeContext class to prevent circular dependencies with WinscopeModule and the constituent parsers.

Bug: 411363817

### DIFF
--- a/src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
+++ b/src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
@@ -25,7 +25,6 @@
 #include "perfetto/ext/base/flat_hash_map.h"
 #include "src/trace_processor/storage/trace_storage.h"
 #include "src/trace_processor/tables/winscope_tables_py.h"
-#include "src/trace_processor/types/destructible.h"
 #include "src/trace_processor/types/trace_processor_context.h"
 
 namespace perfetto::trace_processor {
@@ -57,19 +56,9 @@ struct TrackedMessage {
   std::optional<std::string> location;
 };
 
-class ProtoLogMessageDecoder : public Destructible {
+class ProtoLogMessageDecoder {
  public:
-  explicit ProtoLogMessageDecoder(TraceProcessorContext* context);
-  virtual ~ProtoLogMessageDecoder() override;
-
-  static ProtoLogMessageDecoder* GetOrCreate(TraceProcessorContext* context) {
-    if (!context->protolog_message_decoder) {
-      context->protolog_message_decoder.reset(
-          new ProtoLogMessageDecoder(context));
-    }
-    return static_cast<ProtoLogMessageDecoder*>(
-        context->protolog_message_decoder.get());
-  }
+  explicit ProtoLogMessageDecoder(TraceProcessorContext*);
 
   std::optional<DecodedMessage> Decode(
       uint64_t message_id,
--- a/src/trace_processor/importers/proto/winscope/protolog_parser.cc
+++ b/src/trace_processor/importers/proto/winscope/protolog_parser.cc
@@ -42,16 +42,23 @@
 
 namespace perfetto::trace_processor {
 
-ProtoLogParser::ProtoLogParser(TraceProcessorContext* context)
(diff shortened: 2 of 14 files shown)


src/trace_processor/importers/proto/winscope/winscope_context.h
```diff
@@ -0,0 +1,40 @@
...
+#define SRC_TRACE_PROCESSOR_IMPORTERS_PROTO_WINSCOPE_WINSCOPE_CONTEXT_H_
+
+#include <memory>
+#include "src/trace_processor/types/destructible.h"
+
+namespace perfetto {
+namespace trace_processor {
+
+struct WinscopeContext {
+  // ShellTransitionsTracker
```
LalitMaganti: If you are inside your own module, none of these need to be destructible anymore. Just initialize them eagerly in the constructor of WinscopeModule and make all of these the actual types. You can even remove the unique_ptr now.

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -62,13 +63,15 @@ class ProtoLogMessageDecoder : public Destructible {
   explicit ProtoLogMessageDecoder(TraceProcessorContext* context);
   virtual ~ProtoLogMessageDecoder() override;
 
-  static ProtoLogMessageDecoder* GetOrCreate(TraceProcessorContext* context) {
-    if (!context->protolog_message_decoder) {
-      context->protolog_message_decoder.reset(
-          new ProtoLogMessageDecoder(context));
+  static ProtoLogMessageDecoder* GetOrCreate(
```
LalitMaganti: All of these GetOrCreate functions can now be removed now you've decoupled yourself from the context.

src/trace_processor/importers/proto/winscope/protolog_parser.cc
```diff
@@ -42,17 +42,25 @@
 
 namespace perfetto::trace_processor {
 
-ProtoLogParser::ProtoLogParser(TraceProcessorContext* context)
```
LalitMaganti, line 45: Just make the TraceProcessorContext pointer a field on the WinscopeContext? THat way you need to only pass one thing around.

Even better if you can have only the fields you need in the TraceProcessorContext as that will allow you to clearly declare your dependencies.

PriyankaSPX: > Just make the TraceProcessorContext pointer a field on the WinscopeContext? THat way you need to only pass one thing around.

Done

> Even better if you can have only the fields you need in the TraceProcessorContext as that will allow you to clearly declare your dependencies.

There are a few cases where we need to pass the whole context into things like ArgsTracker so probably best to just leave the context pointer exposed for now.

src/trace_processor/importers/proto/winscope/protolog_parser.cc
```diff
@@ -75,6 +82,8 @@ void ProtoLogParser::ParseProtoLogMessage(
     boolean_params.emplace_back(*it);
   }
 
+  auto* trace_processor_context = context_->trace_processor_context_;
```
LalitMaganti: Here and below: seems like you just want storage, I'd consider pulling that out instead.

src/trace_processor/importers/proto/winscope/shell_transitions_parser.cc
```diff
@@ -30,70 +31,76 @@
...
-ShellTransitionsParser::ShellTransitionsParser(TraceProcessorContext* context)
-    : context_(context), args_parser_{*context->descriptor_pool_} {}
+ShellTransitionsParser::ShellTransitionsParser(WinscopeContext* context)
+    : context_(context),
+      args_parser_{*context->trace_processor_context_->descriptor_pool_} {}
 
 void ShellTransitionsParser::ParseTransition(protozero::ConstBytes blob) {
   protos::pbzero::ShellTransition::Decoder transition(blob);
 
+  auto* trace_processor_context = context_->trace_processor_context_;
```
LalitMaganti: AS above

PR #1913, opened 2025-06-24
PinAndroidPerfMetrics: Add support to pin notifications_blocking metrics

kshitijmg:
- Metrics prefixed `perfetto_android_notifications_blocking_call-blocking_calls-name-` are now handled by `dev.perfetto.PinAndroidPerfMetrics` plugin.
- This is already tracked and visualized by other tools. This CL allows perfetto to pin such metrics in the timeline.
- Other tools build the table in `android_sysui_notifications_blocking_calls_metric.sql` and use the value from the table 'android_sysui_notifications_blocking_calls'
- This CL however does not make use of it, to prevent use of `RUN_METRIC` in light of stdlib migration.

Bug: 352266918
Test: `./ui/run-unittests`
Test: Manual - load traces from b/379219510, pin metric "perfetto_android_notifications_blocking_call-blocking_calls-name-ExpNotRow#onNotifUpdated (leaf)-total_dur_ns-mean"
Change-Id: I9827d5726c9ff5133ad9295ce0343125591f9399

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
@@ -93,3 +93,19 @@ SELECT
 FROM android_binder_txns AS tx
 WHERE
   NOT aidl_name IS NULL AND is_sync = 1;
+
+CREATE PERFETTO FUNCTION _is_relevant_notifications_blocking_call(
+    name STRING,
+    dur LONG
+)
+RETURNS BOOL AS
+SELECT
+  $name = 'NotificationStackScrollLayout#onMeasure'
+  AND $dur > 0
+  AND (
+    $name GLOB 'NotificationStackScrollLayout#onMeasure'
+    OR $name GLOB 'NotificationToplineView#onMeasure'
+    OR $name GLOB 'ExpNotRow#*'
+    OR $name GLOB 'NotificationShadeWindowView#onMeasure'
+    OR $name GLOB 'ImageFloatingTextView#onMeasure'
+  );
--- a/ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinNotificationsBlockingCall.ts
+++ b/ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinNotificationsBlockingCall.ts
@@ -1,0 +1,96 @@
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {
+  NotificationsBlockingCallMetricData,
+  MetricHandler,
+} from './metricUtils';
+import {Trace} from '../../../public/trace';
+import {addDebugSliceTrack} from '../../../components/tracks/debug_tracks';
+
+class BlockingCallMetricHandler implements MetricHandler {
+  /**
+   * Matches metric key for notifications blocking call metrics & return parsed data if successful.
+   *
+   * @param {string} metricKey The metric key to match.
+   * @returns {NotificationsBlockingCallMetricData | undefined} Parsed data or undefined if no match.
+   */
+  public match(
+    metricKey: string,
+  ): NotificationsBlockingCallMetricData | undefined {
+    const matcher =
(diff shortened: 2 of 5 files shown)


ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinNotificationsBlockingCall.ts
```diff
@@ -0,0 +1,95 @@
...
+    JOIN thread USING (utid)
+WHERE
+    thread.is_main_thread AND
+    s.name = 'NotificationStackScrollLayout#onMeasure' AND
+    s.dur > 0 AND (
+       s.name GLOB 'NotificationStackScrollLayout#onMeasure'
+    OR s.name GLOB 'NotificationToplineView#onMeasure'
+    OR s.name GLOB 'ExpNotRow#*'
+    OR s.name GLOB 'NotificationShadeWindowView#onMeasure'
+    OR s.name GLOB 'ImageFloatingTextView#onMeasure'
```
nicomazz: can we extract this to a function that is used both by this plugin and the notification metric? Otherwise changes in the metric will result in this file being outdated

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinNotificationsBlockingCall_unittest.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright (C) 2024 The Android Open Source Project
```
nicomazz: nit: 2025

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinNotificationsBlockingCall.ts
kshitijmg: While it is possible, from my understanding, this would need a change in the sql itself, which cannot be sourced anymore via the `RUN_METRIC()` function - this isn't part of perfetto stdlib and frankly doesn't make sense for it to be a part of stdlib either.
I'd like this to be commonized too, but with the deprecation of `RUN_METRIC()` in favor of stdlib, I don't know if its possible.

cc @devianb

nicomazz: you could move this logic in perfetto/src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql and use it from both places, similarly to `_android_critical_blocking_calls`

kshitijmg: Done!

src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
```diff
@@ -93,3 +93,19 @@ SELECT
...
+SELECT
+  $name = 'NotificationStackScrollLayout#onMeasure'
+  AND $dur > 0
+  AND (
+    $name GLOB 'NotificationStackScrollLayout#onMeasure'
+    OR $name GLOB 'NotificationToplineView#onMeasure'
+    OR $name GLOB 'ExpNotRow#*'
+    OR $name GLOB 'NotificationShadeWindowView#onMeasure'
+    OR $name GLOB 'ImageFloatingTextView#onMeasure'
+  );
```
nicomazz, line 111: Looks good, but i wonder if you can add a small test for this as well. I think there is one for the function above as well that you can use as template

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
kshitijmg: I don't see any tests for `_is_relevant_blocking_call()` other than the tests in `pinBlockingCall_unittest.ts`, which I think validate the metricHandler itself.
There is an equivalent `pinNotificationsBlockingCall_unittest.ts` added by this CL :)

nicomazz: oh ok technically the test for `_is_relevant_blocking_call` is in android_blocking_calls_cuj_metric.py

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinNotificationsBlockingCall.ts
```diff
@@ -0,0 +1,96 @@
+// Copyright (C) 2025 The Android Open Source Project
```
stevegolton, line 1: Typescript filenames are typically snake_case

stevegolton (approved):

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinNotificationsBlockingCall.ts
kshitijmg: Currently, everything belonging to `dev.perfetto.PinAndroidPerfMetrics` is `camelCase`.
We can refactor this in another bug/PR, all at once :)

LalitMaganti (approved):

PR #1914, opened 2025-06-24
tp: Optimize ArgsTracker with FlatHashMap

LalitMaganti:
Replaced std::map with base::FlatHashMap in ArgsTracker for array
indexes. This improves performance by using a more efficient hash map
implementation.

### DIFF
--- a/src/trace_processor/importers/common/args_tracker.h
+++ b/src/trace_processor/importers/common/args_tracker.h
@@ -23,6 +23,8 @@
 #include <tuple>
 #include <type_traits>
 
+#include "perfetto/ext/base/flat_hash_map.h"
+#include "perfetto/ext/base/hash.h"
 #include "perfetto/ext/base/small_vector.h"
 #include "src/trace_processor/dataframe/dataframe.h"
 #include "src/trace_processor/db/column.h"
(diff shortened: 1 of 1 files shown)


src/trace_processor/importers/common/args_tracker.h
```diff
@@ -295,7 +297,18 @@ class ArgsTracker {
...
+      base::Hasher hasher;
+      hasher.Update(reinterpret_cast<uint64_t>(std::get<0>(t)));
+      hasher.Update(std::get<1>(t));
+      hasher.Update(std::get<2>(t));
+      hasher.Update(std::get<3>(t).raw_id());
+      return hasher.digest();
+    }
+  };
+  base::FlatHashMap<ArrayKeyTuple, size_t /*next_index*/, Hasher>
+      array_indexes_;
```
primiano, line 311: is this append only, right? remember that FHM has some known issue if you delete a lot due to tombstone pollution

primiano (approved):

src/trace_processor/importers/common/args_tracker.h
LalitMaganti: Yes it's append only.

PR #1922, opened 2025-06-24
Add new metric for Bitmap Count and Bitmap Memory in SysUI and Launcher.

archisha-baranwal:
This metric shows the aggregate values of Bitmap Count and Bitmap Memory for SysUI and Launcher processes.

Bug: 414346795
Test: tools/diff_test_trace_processor.py --name-filter ".*android_bitmap_metric.*" out/linux_clang_debug/trace_processor_shell
Flag: none


LalitMaganti:
As documented at the bottom of metrics.proto, we're not accepting any new metrics here. Metrics v2 should be used for any and all new metrics like this.

devarshimb@ knows exactly what the plan is here, please reach out to him and discuss what should be done.

PR #1925, opened 2025-06-24
tp: Parse heap graph runtime_internal_objects

ddiproietto:
This introduces a new field for Java heap graph objects: runtime internal objects.

The new field will be populated by ART in a separate commit.

runtime internal objects are parsed as references.

https://buganizer.corp.google.com/issues/424237741


ddiproietto:
I need to add diff_tests

src/trace_processor/importers/proto/heap_graph_tracker.cc
```diff
@@ -425,8 +426,26 @@ void HeapGraphTracker::AddObject(uint32_t seq_id,
     }
     any_references = true;
   }
-  if (any_references) {
+  for (size_t i = 0; i < obj.runtime_internal_objects.size(); ++i) {
+    uint64_t owned_object_id = obj.runtime_internal_objects[i];
+    // This is true for unset reference fields.
+    std::optional<ObjectTable::RowReference> owned_row_ref;
+    if (owned_object_id != 0)
```
ilkos: No braces? Probably better to just std::optional<ObjectTable::RowReference> owned_row_ref = owner_object_id != 0 ? GetOrInsert : nullopt anyway

ilkos: Actually this should never be null for the internal objects

ilkos (approved):
lgtm, let's simplify the nullness case tho

src/trace_processor/importers/proto/heap_graph_tracker.cc
ddiproietto: > Actually this should never be null for the internal objects

+1

PR #1926, opened 2025-06-25
docs: fix docs on firefox

LalitMaganti:
1) sscrollIntoViewIfNeeded is not available on Firefox so switch it
   to the standard scrollIntoView.
2) Tab clicking was causing a scroll of the docs upwards which was
   unexpected. Turns out because it was scrollIntoView of a totally
   separate container but because of how it works it causes scroll
   of the main conatiner as well. Switch it to a more direct
   approach

Bug: 427486537


primiano:
thanks a lot!

PR #1938, opened 2025-06-27
tp: change growth factor of flexvector to 1.5x

LalitMaganti:
2x was causing very high memory use when parsing large heap dumps. Use
1.5x instead which is what also folly's vector uses to be a bit more
reasonable.

Also remove a bunch of confusing code where alignment and capacity
multiples were mixed up. We were not actually using custom alignment
capability in prod so remove it and simplify.

### DIFF
--- a/src/trace_processor/dataframe/impl/flex_vector.h
+++ b/src/trace_processor/dataframe/impl/flex_vector.h
@@ -46,7 +46,6 @@ namespace perfetto::trace_processor::dataframe::impl {
 // - Simple API similar to std::vector but for trivially copyable types only
 //
 // Performance characteristics:
-// - Ensures power-of-two capacity for efficient modulo operations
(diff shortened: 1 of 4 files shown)


src/trace_processor/dataframe/impl/flex_vector.h
```diff
@@ -55,17 +54,22 @@ namespace perfetto::trace_processor::dataframe::impl {
 //   for (int i = 0; i < 20; ++i) {
 //     vec.push_back(i);  // Will automatically resize when needed
 //   }
-template <typename T, uint64_t kAlignment = std::max<uint64_t>(alignof(T), 64)>
+template <typename T>
 class FlexVector {
  public:
+  // The capacity should alwauys be a multiple of this value to ensure
```
elrakadm: typo: always

elrakadm (approved):

PR #1954, opened 2025-06-29
ui: Switch to using DataGrid to display aggregations.

stevegolton:
This patch switches out the old aggregation tables for DataGrids.

This has a number of benefits:
- Promotes code reuse.
- Adds pagination (avoids overloading the DOM).
- Adds the ability to add filters.
- Presents a consistent UX for tabular data throughout the UI.

Example:
image


ui/src/plugins/dev.perfetto.Sched/cpu_slice_selection_aggregator.ts
```diff
@@ -81,44 +81,37 @@ export class CpuSliceSelectionAggregator implements Aggregator {
...
         title: 'TID',
         kind: 'NUMBER',
-        columnConstructor: Float64Array,
         columnId: 'tid',
       },
       {
-        title: 'Wall duration (ms)',
-        kind: 'TIMESTAMP_NS',
-        columnConstructor: Float64Array,
+        title: 'Wall duration',
```
LalitMaganti: Why are we removing the units here and not elsewhere? Feels inconsistent.

ui/src/plugins/dev.perfetto.Sched/cpu_slice_selection_aggregator.ts
```diff
@@ -81,44 +81,37 @@ export class CpuSliceSelectionAggregator implements Aggregator {
...
+        kind: 'DURATION_NS',
         columnId: 'total_dur',
         sum: true,
       },
       {
-        title: 'Avg Wall duration (ms)',
-        kind: 'TIMESTAMP_NS',
-        columnConstructor: Float64Array,
+        title: 'Avg Wall duration',
```
LalitMaganti: As above.

LalitMaganti (approved):

ui/src/plugins/dev.perfetto.Sched/cpu_slice_selection_aggregator.ts
stevegolton: The kind hint 'DURATION_NS' tells the aggregator to expect durations in nanoseconds, and the aggregator will take charge of converting to s/ms/µs/etc according to user preferences.

I've left the other units in because:
- They are using units we have converters or centralized user preferences for - e.g. watts, joules
- They not represented as nanoseconds in the source data - usually floating point milliseconds or similar - and are in some 3rd party plugin I don't particularly want to touch.

stevegolton: Ended up rolling back in order to limit the scope of this PR.

PR #1956, opened 2025-06-30
Consider unique frame_id when calculating per-frame metrics.

devianb:
While extracting the frames in a CUJ, there are duplicate entries for a CUJ. This is because currently there is no way to associate an actual frame with a corresponding DrawFrame. At the metric level, this is an immediate fix to make sure that the correct number of frames are used for metric calculation.

Bug: 425562160
Test: tools/diff_test_trace_processor.py --name-filter ".*android_blocking_calls_cuj_per_frame.*" out/linux_clang_debug/trace_processor_shell

### DIFF
--- a/src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
+++ b/src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
@@ -87,7 +87,9 @@ WHERE
       OR
       -- frame end within cuj
       (frame.ts_end >= cuj.ts AND frame.ts_end <= cuj.ts_end)
-   );
+   )
+   -- This grouping is done to de-duplicate identical rows.
+   GROUP BY frame_id;
 
 -- Combine the above two tables to get blocking calls within frame within CUJ.
 DROP TABLE IF EXISTS blocking_calls_frame_cuj;


src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
```diff
@@ -87,7 +87,7 @@ WHERE
       OR
       -- frame end within cuj
       (frame.ts_end >= cuj.ts AND frame.ts_end <= cuj.ts_end)
-   );
+   ) GROUP BY frame_id;
```
LalitMaganti: If you're grouping by frame_id, please be explicit in the aggregate functions you are taking with min/max/other aggregate functions. Implicitly relying on sqlite to pick which row is chosen is not a good idea.

devianb: But in this case the duplicate rows are identical(all column values are the same), so I don't see how I can use the aggregation functions for a specfic column.

LalitMaganti: Please leave a comment then.

devianb: Done.

LalitMaganti (approved):

PR #1970, opened 2025-06-30
prelude: add flows index in prelude

LalitMaganti:


src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/indexes.sql
```diff
@@ -0,0 +1,19 @@
+--
+-- Copyright 2024 The Android Open Source Project
```
aMayzner: 2025

PR #1979, opened 2025-07-01
Update extractor to return LayerDecoder.

PriyankaSPX:
LayerDecoder provides greater clarity than ConstBytes. Also use optional Rect instead of pointer in GetCroppedScreenBounds.

Bug: 411363817

### DIFF
--- a/src/trace_processor/importers/proto/winscope/surfaceflinger_layers_extractor.cc
+++ b/src/trace_processor/importers/proto/winscope/surfaceflinger_layers_extractor.cc
@@ -28,64 +28,64 @@ enum ProcessingStage { VisitChildren, Add };
 
 // When z-order is the same, we sort such that the layer with the layer id
 // is drawn on top.
-void SortByZThenLayerId(std::vector<ConstBytes>& layers) {
-  std::sort(layers.begin(), layers.end(),
-            [](const ConstBytes& a, const ConstBytes& b) {
-              LayerDecoder layer_a(a);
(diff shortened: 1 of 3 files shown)


src/trace_processor/importers/proto/winscope/surfaceflinger_layers_extractor.cc
```diff
@@ -28,117 +28,132 @@ enum ProcessingStage { VisitChildren, Add };
...
+    const std::unordered_map<int, LayerDecoder>& layers_by_id) {
+  std::sort(layer_ids.begin(), layer_ids.end(), [&layers_by_id](int a, int b) {
+    const auto& layer_a = layers_by_id.at(a);
+    const auto& layer_b = layers_by_id.at(b);
+    auto z_val_a = layer_a.z();
+    auto z_val_b = layer_b.z();
+    if (z_val_a != z_val_b) {
+      return z_val_a > z_val_b;
+    }
+    return layer_a.id() >= layer_b.id();
```
LalitMaganti: Btw this is not a valid comparator functions. It expects false for equality.

LalitMaganti: Also consider using std::tie (see examples elsewhere in the codebase

PriyankaSPX: thanks - tie doesnt work here but changed to use make_tuple

PR #1983, opened 2025-07-01
Add support for CPU-per-UID data source to Trace Processor

simonmacm:
Bug: http://b/422976190


src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
```diff
@@ -0,0 +1,152 @@
...
+
+    UpdateCounter(ts, uid, cluster_id, it.value());
+  }
+}
+
+void AndroidCpuPerUidModule::UpdateCounter(int64_t ts,
+                                           uint32_t uid,
+                                           uint32_t cluster,
+                                           uint64_t value) {
+  if (!loaded_package_names_) {
```
LalitMaganti: IMO this is the wrong place to do this. For one, your assumption that a uid can have a unique package is just wrong. Really it should be up to the user at the SQL layer to figure out what the right thing to do here is.

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
```diff
@@ -0,0 +1,152 @@
...
+    loaded_package_names_ = true;
+    const auto& package_list = context_->storage->package_list_table();
+    for (auto row = package_list.IterateRows(); row; ++row) {
+      package_names_.Insert(row.uid(), row.package_name());
+    }
+  }
+
+  static constexpr auto kBlueprint = tracks::CounterBlueprint(
+      "android_cpu_per_uid", tracks::StaticUnitBlueprint("ms"),
+      tracks::DimensionBlueprints(
```
LalitMaganti: Please just have the uid, leave the package name to the SQL layer.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/counter_tracks.ts
```diff
@@ -36,6 +36,13 @@ export const COUNTER_TRACK_SCHEMAS: ReadonlyArray<CounterTrackTypeSchema> = [
     topLevelGroup: 'THERMALS',
     group: 'ACPM Temperature',
   },
+  {
+    type: 'android_cpu_per_uid',
```
LalitMaganti: I'd suggest making your own plugin and having a custom query instead of doing this given your need to join with the package list table.

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
simonmacm: Hmm, I was trying to copy what nettrace does.

LalitMaganti: nettrace was written before plugins were a thing and I let it slide because there was no reasonable way to have "custom SQL track in the UI" at that time. That does not apply anymore.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/counter_tracks.ts
simonmacm: Done.

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
simonmacm: Done.

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
simonmacm: Done.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/counter_tracks.ts
simonmacm: (In the sense of removing this stuff, the plugin will come later.)

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
```diff
@@ -0,0 +1,125 @@
...
+void AndroidCpuPerUidModule::UpdateCounter(int64_t ts,
+                                           uint32_t uid,
+                                           uint32_t cluster,
+                                           uint64_t value) {
+  static constexpr auto kBlueprint = tracks::CounterBlueprint(
+      "android_cpu_per_uid", tracks::StaticUnitBlueprint("ms"),
+      tracks::DimensionBlueprints(tracks::UintDimensionBlueprint("uid"),
+                                  tracks::UintDimensionBlueprint("cluster")),
+      tracks::FnNameBlueprint([](uint32_t uid, uint32_t cluster) {
+        return base::StackString<1024>("%d CL%d", uid, cluster);
```
LalitMaganti: Ideally in the track name, having some indication of the source of the track would be a good idea.

simonmacm: Done, I think.

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
```diff
@@ -0,0 +1,125 @@
...
+    if (uid_with_value_this_packet.count(uid)) {
+      continue;
+    }
+    uint32_t cluster_id = it.key() & 0xffffffff;
+
+    UpdateCounter(ts, uid, cluster_id, it.value());
+  }
+}
+
+void AndroidCpuPerUidModule::UpdateCounter(int64_t ts,
```
LalitMaganti, line 110: This needs some diff tests.

simonmacm: Done

PR #1984, opened 2025-07-01
Introduction of Arm CPU telemetry solution

pan-:
Arm introduced a CPU [telemetry solution](https://developer.arm.com/documentation/109542/latest/) for its cores starting with Arm Neoverse N1. The telemetry solution helps developers understand where CPU cycles are spent, identify performance bottlenecks, and improve execution efficiency.

The CPU telemetry solution is based on [specification files](https://gitlab.arm.com/telemetry-solution/telemetry-solution/-/tree/main/data/pmu/cpu/neoverse?ref_type=heads) tailored for each core. These specification files describe core-specific PMU events and derived metrics, identifying key performance bottlenecks. They also define a metric tree for exploring microarchitectural performance.

Arm has released a CLI tool named [topdown\_tool](https://gitlab.arm.com/telemetry-solution/telemetry-solution/-/tree/main/tools/topdown_tool?ref_type=heads) to capture these metrics and has integrated support into its own products, such as the [Streamline Performance Analyzer](https://developer.arm.com/Tools%20and%20Software/Streamline%20Performance%20Analyzer).

This PR enables support for Arm's CPU telemetry solution in Perfetto, proposing the following changes:

* A new filter for traced perf to limit capture to a specific CPUID.
* Addition of a `CpuInfoManager` in the `App` to manage CPU description files.
* A new support page to visualize and load CPU specification files.
* A new recording section guiding users in configuring PMU captures. Users select metrics to capture, and the system selects the correct PMU events based on the chosen CPU. Multiple configurations can be created simultaneously—for example, to capture events on systems with heterogeneous cores.
* A new telemetry plugin that analyses captured events, computing and displaying metrics based on CPU specification files.

Additionally, this PR introduces a flag allowing Perfetto to connect to an alternative well-known socket instead of the default system socket. This is convenient when working with an updated tracebox/traced_perf on production systems where updating Perfetto itself may not be feasible.

### DIFF
--- a/protos/perfetto/config/perfetto_config.proto
+++ b/protos/perfetto/config/perfetto_config.proto
@@ -2371,7 +2371,7 @@ message FollowerEvent {
 //     }
 //   }
 //
-// Next id: 21
+// Next id: 23
 message PerfEventConfig {
   // What event to sample on, and how often.
   // Defined in common/perf_events.proto.
@@ -2393,6 +2393,12 @@ message PerfEventConfig {
   //
   // Available since: perfetto v50.
   repeated uint32 target_cpu = 20;
+
+  // Allow failure of perf_event_open
+  optional bool ignore_open_failure = 21;
+
+  // Filter based on CPUID
+  repeated string cpuid = 22;
 
   //
   // Kernel <-> userspace ring buffer options:
--- a/protos/perfetto/config/profiling/perf_event_config.proto
+++ b/protos/perfetto/config/profiling/perf_event_config.proto
@@ -36,7 +36,7 @@ package perfetto.protos;
 //     }
 //   }
 //
-// Next id: 21
+// Next id: 23
 message PerfEventConfig {
   // What event to sample on, and how often.
   // Defined in common/perf_events.proto.
@@ -58,6 +58,12 @@ message PerfEventConfig {
   //
   // Available since: perfetto v50.
   repeated uint32 target_cpu = 20;
+
+  // Allow failure of perf_event_open
+  optional bool ignore_open_failure = 21;
+
+  // Filter based on CPUID
+  repeated string cpuid = 22;
 
   //
   // Kernel <-> userspace ring buffer options:
--- a/protos/perfetto/trace/generic_kernel/generic_task.proto
+++ b/protos/perfetto/trace/generic_kernel/generic_task.proto
@@ -71,7 +71,6 @@ message GenericKernelTaskRenameEvent {
   // New command name for the thread.
   optional string comm = 2;
 }
-
 // Metadata about the processes and threads in the trace.
 // The main goal of this proto is to provide a generic kernel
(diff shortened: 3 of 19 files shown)


LalitMaganti (changes requested):
Thanks for the contribution! Can I ask that you please split this into multiple PRs as different folk need to review different parts of this code:
1) the CPU reader parts need to be reviewed by on device folk
2) the recording stuff needs to be handled by UI folk
3) there seems to be some ad-hoc bug fixes which need to be reviewed individually.

primiano:
Thanks a lot for this.
It comes as a bit of a surprise (I didn't know this was happening) but it's certainly a welcome surprise :)
You definitely put a lot of though into this.

I'd echo Lalit's request: could this be split in different PRs? It's going to make it faster to iterate, as there are a bunch of independent discussion to have (e.g. on the socket, i have some questions, and some comment son the UI side), but I would not block a whole PR to have all the discussions, otherwise we'll never converge.

Huge thanks for contributing to this! looks awesome overall

pan-:
I propose to have four PR: 
- The CPUID filtering for traced perf
- The socket flag
- The CPU manager + page + telemetry plugin as they can work standalone without the recording page
- The split panel change used to make tab more usable as widgets.
- The Recording page for PMUs.

Does that sounds right or would you like to subdivide the third item further ?

LalitMaganti:
I think that looks like a very reasonble starting point. Lets go with that and we can discuss on individual PRs if we need any more splitting.

rsavitski:
@pan- Hello, checking in on whether you're waiting on us for a review, or we're waiting for a new version of the patches? Just clarifying, not intending to hurry you.

From me personally, once the patch is ready for review, please split the src/profiling/... changes into a separate PR (there was https://github.com/google/perfetto/pull/2034, maybe reuse that).

pan-:
@rsavitski Thanks for coming back to us, we need to split this PR into multiple pull request. 
I will let you know once it is complete.

abhmen01:
Created PRs according to @pan- 's proposal.

The CPUID filtering for traced perf:  #2034 
The socket flag: #2036 
The CPU manager + page + telemetry plugin:  #5284 
The drawer panel change used to make tab more usable as widgets: #5282 
The Recording page for PMUs: #5286

primiano:
@abhmen01  There seems to be some issues with your PR split. Some of them have duplicted contents (the same diffs on different PRs) and is becoming hard to figure out what goes where.
Also I cannot fully tell which PRs are ready to be reviewed and which ones you are still iterating on.

Can you be more explicit on the ones that are ready for review?

abhmen01:
Hi @primiano 
#2034, #2036, #5282, and #5284 are mutually exclusive and may be reviewed independently.
#5286 has dependencies on certain commits from the above PRs; therefore, those commits have been grouped with it.

PR #1985, opened 2025-07-01
Add nested tracks and flamegraphs for long battery plugin.

rzuklie:
This uses the new generic day explorer tables to form an arbitrarily nested set of tracks, rather than the old fixed set of tracks. This also adds area selection for these tracks which will show a flame graph following the same hierarchy.


simonmacm (approved):
Generally looks great, thank you.

This does make me think that the existing plugin is gathering a load of responsibilities and becoming quite heavyweight, and that we should think about how to break it up soon...

LalitMaganti:
+1 please split it up. There's a bunch of functionality here which has little to do with day explorer. I think there are at least 2-3 plugins hiding in this plugin.

PR #2000, opened 2025-07-02
Make trace processor metatrace thread local, not global

simonmacm:
This makes it possible to use metatrace in a multi-threaded context, as long as we are careful to only use each trace processor instance from a single thread.

Bug: [427909201](https://b.corp.google.com/427909201)

### DIFF
--- a/src/trace_processor/tp_metatrace.h
+++ b/src/trace_processor/tp_metatrace.h
@@ -19,7 +19,6 @@
 
 #include <array>
 #include <functional>
-#include <memory>
 #include <vector>
 
 #include "perfetto/base/time.h"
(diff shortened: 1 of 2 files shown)


src/trace_processor/tp_metatrace.h
```diff
@@ -132,7 +132,7 @@ class RingBuffer {
   void ReadAll(std::function<void(Record*)>);
 
   static RingBuffer* GetInstance() {
-    static RingBuffer* rb = new RingBuffer();
+    thread_local RingBuffer* rb = new RingBuffer();
```
LalitMaganti: This will leak memory. You should use std::unique_ptr<RingBuffer>

LalitMaganti (approved):

PR #2007, opened 2025-07-02
Simplify GetThreadOrNull(tid) in process_tracker

jahdiel-alvarez:
ProcessTracker has two GetThreadOrNull methods. One only takes a tid as an argument and the other takes a tid and an optional pid. Refactored the method so that it is clear that calling GetThreadOrNull(tid, std::nullopt) is the same as GetThreadOrNull(tid).

Bug: 425694654


jahdiel-alvarez:
@LalitMaganti I thought about consolidating into a single `GetThreadOrNull(tid, optional<pid>)` method. Would make the API public though. WDYT?

LalitMaganti:
> I thought about consolidating into a single GetThreadOrNull(tid, optional<pid>) method. Would make the API public though. WDYT?

I don't think that's a good idea. The semantic of that would be something I don't think anyone should use.

PR #2012, opened 2025-07-02
Limited the existence of a TID to one at a time

jahdiel-alvarez:
ProcessTracker permitted the creation of multiple threads with same TID but different UTIDs, mainly to address data loss scenarios. Refactored the code to have better semantics to address these cases. The ClearThread method removes all preexisting UTIDs from the active thread map, acting as a barrier point from which a new thread can be created for a preexisting TID.

Also, updated the API so that GetOrCreateThread becomes the primary thread creation method, forcing a single TID and thread association at any point in time.

Bug: 425694654

### DIFF
--- a/src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
+++ b/src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
@@ -146,7 +146,8 @@ std::optional<UniqueTid> GenericKernelParser::GetUtidForState(int64_t ts,
             stats::generic_task_state_invalid_order);
         return std::nullopt;
       }
-      UniqueTid utid = context_->process_tracker->StartNewThread(ts, tid);
+      UniqueTid utid = context_->process_tracker->GetOrCreateThread(tid);
+      context_->process_tracker->SetThreadStartTs(utid, ts);
       context_->process_tracker->UpdateThreadName(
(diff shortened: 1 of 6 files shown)


src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
```diff
@@ -146,7 +146,8 @@ std::optional<UniqueTid> GenericKernelParser::GetUtidForState(int64_t ts,
             stats::generic_task_state_invalid_order);
         return std::nullopt;
       }
-      UniqueTid utid = context_->process_tracker->StartNewThread(ts, tid);
+      UniqueTid utid = context_->process_tracker->GetOrCreateThread(tid);
```
LalitMaganti, line 151: Shouldn't you ClearThread here?

LalitMaganti (approved):

src/trace_processor/importers/generic_kernel/generic_kernel_parser.cc
jahdiel-alvarez: @LalitMaganti Not really because we are already explicitly checking above that the thread doesn't exist (line 144).

LalitMaganti: Can you add a comment then?

LalitMaganti (approved):

PR #2017, opened 2025-07-03
tp: add msd and lsd radix sort implementations and use them in dataframe

LalitMaganti:
This CL adds two implementation of radix sort:
* LSD based for fixed size keys
* MSD based for variable size keys

We start using it in the appropriate places in the interpreter.

Before:
```
BM_BytecodeInterpreter_SortUint32            298341154 ns    298193315 ns            2
BM_BytecodeInterpreter_SortString            686262567 ns    686039510 ns            1
```

After:
```
BM_BytecodeInterpreter_SortUint32             25955994 ns     25947812 ns           27
BM_BytecodeInterpreter_SortString            173340011 ns    173280245 ns            4
```


src/base/radix_sort_unittest.cc
```diff
@@ -0,0 +1,231 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
```
aMayzner: 2025

PR #2019, opened 2025-07-03
ui: Add thread state by cpu aggregation behind flag

stevegolton:
This patch adds a new aggreagtion tab which breaks down thread states by CPU as well as thread and state.

It is behind a flag as it doubles the computation required when an area selection is made.

Example:
image

Related: https://buganizer.corp.google.com/issues/378644348

### DIFF
--- a/ui/src/plugins/dev.perfetto.Sched/index.ts
+++ b/ui/src/plugins/dev.perfetto.Sched/index.ts
@@ -52,6 +52,9 @@ import {ThreadStateSelectionAggregator} from './thread_state_selection_aggregato
 import {createThreadStateTrack} from './thread_state_track';
 import {WakerOverlay} from './waker_overlay';
 import {Cpu} from '../../components/cpu';
+import {ThreadStateByCpuAggregator} from './thread_state_by_cpu_aggregator';
+import {App} from '../../public/app';
+import {Flag} from '../../public/feature_flag';
 
 function uriForThreadStateTrack(upid: number | null, utid: number): string {
   return `${getThreadUriPrefix(upid, utid)}_state`;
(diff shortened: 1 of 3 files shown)


nicomazz:
nit: is it possible to have the cpu id centered in the column? I wonder if you could also add the cpu sizing there as it might help (but optional). 
Another FR would be the avg frequency of each cpu while in "running" state, but probably for another change

stevegolton:
> nit: is it possible to have the cpu id centered in the column?

Yes! But why? Numerical values are typically right aligned.

Note: The gap to the right is space for the cell context menu button - I plan on improving this as it does look a bit odd, I'm just not quite sure how yet.

> I wonder if you could also add the cpu sizing there as it might help (but optional).

Print cpu size: yes. Would it make sense to group by cpu size then cpu?

> Another FR would be the avg frequency of each cpu while in "running" state, but probably for another change
Definitely for another PR. Could be slow to calc if number of selected threads is large.

nicomazz:
> Yes! But why? Numerical values are typically right aligned.

oh I thought TID and PID were center alligned but actually there was just some right padding. It's fine to keep it right aligned if it's the standard for numbers.

> Would it make sense to group by cpu size then cpu?

I think grouping by cpu size is better: my only concern is whether the cpu size is wrong (because AFAIK it's just an heuistic)

> Could be slow to calc if number of selected threads is large.

ok let's leave it out from this but we have another FR for it already opened actually. Let's talk about it offline

LalitMaganti:
CPU size is no longer a heuristic, we fixed that a while back.

LalitMaganti (approved):

LalitMaganti (approved):

Issue #951, opened 2024-11-27
traceconv get empty file


tingfengwanxiao:
When I use as ./traceconv.exe systrace 1.pftrace 2.ftrace,I get a empty 2.ftrace,
it seems because my 1.pftrace is too large as 8G,so How can I change so large perfetto trace to ftrace,
or how can I directly capture a ftrace by som command as large as 8G?

LalitMaganti:
Are you sure it's actually because the file is large? Have you tried with a smaller file and things work correctly?

I would be a bit surprised if the file being large was an issue albeit it's not impossible I guess. 

Just to be clear, if you have an 8GB Perfetto file, that will turn into a 20GB or more systrace file. What are you planning on doing with that in the end?

tingfengwanxiao:
I want to capture a long-term trace and then divide it into smaller ones at certain time points, but perfetto trace seems unable to be directly divided, so I want to convert it into ftrace and use text processing to divide it. Do you have any good suggestions

LalitMaganti:
Why not just collect many small traces in the first place? Would be a lot easier than capturing a big trace and then trying to split it.

LalitMaganti:
Fine I don't think it's worth discussing more in that direction.

Back to the actual topic at hand: would still be useful to know if this is only an issue on big traces or also happening on smaller traces. Especially interesting is if the cut-off is 4GB.

LalitMaganti:
Do you have enough memory on your system to have the full trace in memory i.e. at least 20GB of free memory? Because systrace conversion works by first parsing the full trace into trace processor tables before then doing the conversion.

LalitMaganti:
Specifically, what happens if you just try and just parse the trace with trace_processor_shell? Does it work?

tingfengwanxiao:
> Do you have enough memory on your system to have the full trace in memory i.e. at least 20GB of free memory? Because systrace conversion works by first parsing the full trace into trace processor tables before then doing the conversion.

I only have 16G,maybe its reason

tingfengwanxiao:
Then how can I do for this large trace? split it  by time for some small trace

LalitMaganti:
There's not really a good solution for this today unfortunately. Everyone who needs this sort of thing today just collects smaller traces in the first place.

tingfengwanxiao:
ok ,thank you !

LalitMaganti:
I've changed this bug to track a better implementation of trace splitting: we might implement it at some point but for us it's not a super high priority.

[...]

PR #2020, opened 2025-07-03
tp: Reduce the number of trace_bounds calculations.

LiangliangSui:
Currently, there is redundancy in the calculation of `trace_bounds`, which is computed multiple times. 
Take `TraceProcessorImpl::Flush` and `TraceProcessorImpl::NotifyEndOfFile()` as examples: 

both functions construct `trace_bounds` by calling `BuildBoundsTable`. The step of `GetTraceTimestampBoundsNs`
is relatively time-consuming as it internally iterates over numerous tables. When dealing with
excessively large trace files, this time consumption becomes particularly noticeable.

```c++
base::Status LoadTrace(const std::string& trace_file_path, double* size_mb) {
  base::Status read_status = ReadTraceUnfinalized(
      g_tp, trace_file_path.c_str(), [&size_mb](size_t parsed_size) {
        *size_mb = static_cast<double>(parsed_size) / 1E6;
        fprintf(stderr, "\rLoading trace: %.2f MB\r", *size_mb);
      });
  g_tp->Flush();

  ......

  return g_tp->NotifyEndOfFile();
}
```

Within the `Flush` method, `BuildBoundsTable` is invoked. Additionally, `NotifyEndOfFile` calls
`BuildBoundsTable` twice (since `NotifyEndOfFile` internally triggers `Flush`).

Perhaps we could retain only the `BuildBoundsTable` call within `NotifyEndOfFile`, calculating
`trace_bounds` just once at the end. This approach would eliminate two invocations of
`GetTraceTimestampBoundsNs`, thereby reducing processing time.

Taking the loading of 25.33MB and 268.26MB trace files via trace_processor_shell as examples,
the execution times before and after optimization are as follows:

**25.33MB**

before
Trace loaded: 25.33 MB in 6.21s (4.1 MB/s)
Trace loaded: 25.33 MB in 6.22s (4.1 MB/s)
Trace loaded: 25.33 MB in 6.20s (4.1 MB/s)
Trace loaded: 25.33 MB in 6.23s (4.1 MB/s)
Trace loaded: 25.33 MB in 6.24s (4.1 MB/s)

after
Trace loaded: 25.33 MB in 5.94s (4.3 MB/s)
Trace loaded: 25.33 MB in 5.95s (4.3 MB/s)
Trace loaded: 25.33 MB in 5.92s (4.3 MB/s)
Trace loaded: 25.33 MB in 5.90s (4.3 MB/s)
Trace loaded: 25.33 MB in 5.89s (4.3 MB/s)

**268.26MB**

before
Trace loaded: 268.26 MB in 76.01s (3.5 MB/s)
Trace loaded: 268.26 MB in 76.56s (3.5 MB/s)
Trace loaded: 268.26 MB in 77.01s (3.5 MB/s)
Trace loaded: 268.26 MB in 76.74s (3.5 MB/s)
Trace loaded: 268.26 MB in 76.91s (3.5 MB/s)

after
Trace loaded: 268.26 MB in 70.89s (3.8 MB/s)
Trace loaded: 268.26 MB in 71.24s (3.8 MB/s)
Trace loaded: 268.26 MB in 71.39s (3.8 MB/s)
Trace loaded: 268.26 MB in 71.30s (3.8 MB/s)
Trace loaded: 268.26 MB in 71.22s (3.8 MB/s)

### DIFF
--- a/src/trace_processor/trace_processor_impl.cc
+++ b/src/trace_processor/trace_processor_impl.cc
@@ -589,8 +589,6 @@ base::Status TraceProcessorImpl::Parse(TraceBlobView blob) {
 
 void TraceProcessorImpl::Flush() {
   TraceProcessorStorageImpl::Flush();
-  BuildBoundsTable(engine_->sqlite_engine()->db(),
-                   GetTraceTimestampBoundsNs(*context_.storage));
 }
 
 base::Status TraceProcessorImpl::NotifyEndOfFile() {
@@ -1405,8 +1403,6 @@ std::unique_ptr<PerfettoSqlEngine> TraceProcessorImpl::InitPerfettoSqlEngine(
     }
   }
 
-  // Fill trace bounds table.
-  BuildBoundsTable(db, GetTraceTimestampBoundsNs(*storage));
   return engine;
 }
 


LalitMaganti:
Actually I'm afraid these calls are necessary; we need bounds to be correctly set after each Flush call. If we want to improve this we need a more targeted change where we detect whether or not the source tables have changed and decide based on that whether or not we should do this.

LiangliangSui:
> Actually I'm afraid these calls are necessary; 

At present, after each Flush, there is a corresponding NotifyEndOfFile, so the correctness can be guaranteed.
In addition, I have a question here: is it stipulated that NotifyEndOfFile must be called at the end?



> we need bounds to be correctly set after each Flush call

trace_bounds should only be used by the UI currently, and it has not been used after each Flush.

LalitMaganti:
> At present, after each Flush, there is a corresponding NotifyEndOfFile, so the correctness can be guaranteed.

Sure but the point is that people don't *have* to do this. Trace processor is a library and we have to care about what people might do in theory as well as our own "reference implementation".

Very practically speaking, when we run symbolisation, we  actually call flush and then run a query *without* calling NotifyEndOfFile. This is a perfectly valid thing to do and in this case trace_bounds would not be updated with your change.

> In addition, I have a question here: is it stipulated that NotifyEndOfFile must be called at the end?

Yes absolutely but people can (and do) write queries before it is called.

>  trace_bounds should only be used by the UI currently, and it has not been used after each Flush.

This is just not true, trace_bounds is a public API and can be used by anyone, anywhere in the world. Even aside from theoretical uses from other people, there are people inside Google who rely on trace_bounds being accurate. This CL would break them.

LalitMaganti:
https://github.com/google/perfetto/pull/2022 is the correct version of this patch. It respects the API expectations of trace_bounds being built on Flush API calls while only making one call on the "main path" of loading traces using trace_processor_shell.

LiangliangSui:
> If we want to improve this we need a more targeted change where we detect whether or not the source tables have changed and decide based on that whether or not we should do this.

https://github.com/google/perfetto/pull/2022 only changed the three calls to BuildBoundsTable into two. Maybe really necessary to detect the changes of tables to decide whether to call BuildBoundsTable.

LalitMaganti:
> https://github.com/google/perfetto/pull/2022 only changed the three calls to BuildBoundsTable into two. Maybe really necessary to detect the changes of tables to decide whether to call BuildBoundsTable.

Can you explain where the two are coming from? There should only be a single call.

LiangliangSui:
Flush and below link
https://github.com/google/perfetto/blob/05523df14a57fdada791d362968e4730a86d8633/src/trace_processor/trace_processor_impl.cc#L623-L629

LalitMaganti:
If you read the change caarefully, Flush is no longer called by trace_processor_shell when not symbolizing/deobfuscating. I'm very sure if you test your traces with that patch, you will see the same perf improvement as your patch.

LiangliangSui:
sorry, I didn't notice the changes in src/trace_processor/trace_processor_shell.cc. The current modifications can achieve the expected results.

LiangliangSui:
I have a question unrelated to this PR.

I want to build a tool to split a large trace into multiple small traces. Could you share your thoughts or suggestions?

LalitMaganti:
See https://github.com/google/perfetto/issues/951. It's a very hard problem to do in a general way, you're going to run in all sorts of subtle issue with "ambient state" (things like which processes are running on the system) if you split traces niavely.

Concretely: at the start of traces, we snapshot the full state of all the processes on the system. If you do a naive splitting tool where you just split the trace based on packet boundaries, you'll drop this packet and then be confused why there in your split trace all the processes are missing.

Now there are probably 20-25 different similar instances of "we snapshot this at the start of the trace and we want to preserve it" and many more of "we want to preserve some data (e.g. the value of a counter) from before the start of the split trace".

So TLDR it's a very hard problem, if you decide to solve it I wish you luck. It'll be something we get to at some point but to do it properly, it's a multi-month project and it's never been a priority so not something we will get to soon.

PR #2030, opened 2025-07-03
base: speed up FlatHashMap::Clear()

LalitMaganti:
We were doing a full run through and rehash for no reason. In the end,
clear will want everything zeroed out and the tags empty. So just do
that directly.

This speeds up some pathalogical cases in trace processor where we were
spending a lot of time in the rehashing codepaths.

### DIFF
--- a/src/base/flat_hash_map_benchmark.cc
+++ b/src/base/flat_hash_map_benchmark.cc
@@ -303,16 +303,16 @@ void BM_HashMap_LookupRandInts(benchmark::State& state) {
 }
 
 template <typename MapType>
-void BM_HashMap_RandomIntsInsertAndClear(benchmark::State& state) {
+void BM_HashMap_RandomIntsClear(benchmark::State& state) {
   std::minstd_rand0 rng(0);
   std::vector<size_t> keys(static_cast<size_t>(num_samples()));
   std::shuffle(keys.begin(), keys.end(), rng);
 
   MapType mapz;
-  for (auto _ : state) {
-    for (const size_t key : keys)
-      mapz.insert({key, key});
+  for (const size_t key : keys)
+    mapz.insert({key, key});
 
+  for (auto _ : state) {
     mapz.clear();
     benchmark::ClobberMemory();
   }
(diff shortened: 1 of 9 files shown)


include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -264,13 +265,19 @@ class FlatHashMap {
     if (PERFETTO_UNLIKELY(capacity_ == 0))
       return;
 
-    for (size_t i = 0; i < capacity_; ++i) {
-      const uint8_t tag = tags_[i];
-      if (tag != kFreeSlot && tag != kTombstone)
-        EraseInternal(i);
+    if (!std::is_trivially_destructible<Key>::value ||
+        !std::is_trivially_destructible<Value>::value) {
```
primiano: I don't the is_trivially_destructible vheck buys anything. if trivially destructible, the whole for loop becomes a no-op. the compiler will figure it out.

https://godbolt.org/z/vKjMva1fj
vs
https://godbolt.org/z/168nEGvPK

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -264,13 +265,19 @@ class FlatHashMap {
...
+        const uint8_t tag = tags_[i];
+        if (tag != kFreeSlot && (AppendOnly || tag != kTombstone)) {
+          keys_[i].~Key();
+          values_[i].~Value();
+        }
+      }
     }
-    // Clear all tombstones. We really need to do this for AppendOnly.
-    MaybeGrowAndRehash(/*grow=*/false);
+    memset(&tags_[0], kFreeSlot, capacity_);
```
primiano: could we call Reset(capacity_) here? does it make a big difference?
I feel it keeps the code more mainteinable.

maybe you can add, in the Reset() a branch like:

size_t old_capacity = capacity_;
... (current reset code)
if (n != old_capacity) {
    keys_ = AlignedAllocTyped<Key[]>(n);      // Deliberately not 0-initialized.
    values_ = AlignedAllocTyped<Value[]>(n);  // Deliberately not 0-initialized.
}

so it doesn't re-allocate.

primiano (approved):

PR #2031, opened 2025-07-03
tp: add logic to guard against too many tombstones in flathashmap

LalitMaganti:
Prevents pathalogical situations where Find operations are slow because
of too many tombstones.


primiano (approved):
I am not sure I understand what this does. isn't that cast implicit given the *0.85?

primiano (approved):
Actually LGTM now that github loaded the diff.
thanks!

PR #2034, opened 2025-07-03
Add CPU filter into traced_perf

pan-:
PMUs can generate large amounts of data which causes an overhead in term of capture and processing. Perfetto capture the same PMUs for all of the CPUs except if CPUs are selected individually.

This change introduce a parameter to restrict the capture on specific CPUIDs. The capture can be configured to restrict data captured to CPUs that actually matters.


pan-:
The elephant in the room in my opinion is the code duplication with https://github.com/google/perfetto/blob/main/src/traced/probes/system_info/system_info_data_source.cc which already parse `/proc/cpuinfo`. 

Where should it be factored ?

LalitMaganti:
Please put it in base. That's where all shared Perfetto code goes.

primiano:
yeah src/base/cpu_info.{h,cc} seems the right place if the code can be refactored between the two places.

also +@rsavitski FYI

pan-:
@primiano @LalitMaganti We are resuming our work to get Arm topdown telemetry in Perfetto. 
@abhmen01 Will work on these patches.

src/base/cpu_info.cc
```diff
@@ -0,0 +1,181 @@
...
+  std::optional<uint32_t> cpu_index;
+  std::optional<uint32_t> implementer;
+  std::optional<uint32_t> architecture;
+  std::optional<uint32_t> variant;
+  std::optional<uint32_t> part;
+  std::optional<uint32_t> revision;
+  uint64_t features = 0;
+
+  auto flush_cpu = [&] {
+    AppendCpuInfo(&cpus, processor, cpu_index, implementer, architecture,
```
primiano: Why do you define a function (AppendCpuInfo) which is called only in one place (here)?
Why the code is not inlined in here? I really dislike when I have to jump through hops just to read some code that is supposed to be linear

src/base/cpu_info.cc
```diff
@@ -0,0 +1,181 @@
...
+                   std::optional<uint32_t> implementer,
+                   std::optional<uint32_t> architecture,
+                   std::optional<uint32_t> variant,
+                   std::optional<uint32_t> part,
+                   std::optional<uint32_t> revision,
+                   uint64_t features) {
+  if (!cpu_index.has_value())
+    return;
+
+  CpuInfo cpu;
```
primiano: `CpuInfo cpu{}` (with trailing {}) so it gets default.initialized if you forget some fields

src/base/cpu_info.cc
```diff
@@ -0,0 +1,181 @@
...
+    implementer = std::nullopt;
+    architecture = std::nullopt;
+    variant = std::nullopt;
+    part = std::nullopt;
+    revision = std::nullopt;
+    features = 0;
+  };
+
+  while (line_start != proc_cpu_info.end()) {
+    auto line_end = std::find(line_start, proc_cpu_info.end(), '\n');
```
primiano: since you are touching this code should we use just base::StringSplitter here?  think some of this boilerplate could be avoided

src/traced/probes/system_info/system_info_data_source.cc
```diff
@@ -82,105 +49,41 @@ void SystemInfoDataSource::Start() {
...
+        parsed_cpu.variant && parsed_cpu.revision) {
+      auto* identifier = cpu->set_arm_identifier();
+      identifier->set_implementer(parsed_cpu.implementer.value());
+      identifier->set_architecture(parsed_cpu.architecture.value());
+      identifier->set_variant(parsed_cpu.variant.value());
+      identifier->set_part(parsed_cpu.part.value());
+      identifier->set_revision(parsed_cpu.revision.value());
+    } else if (parsed_cpu.implementer || parsed_cpu.architecture ||
+               parsed_cpu.part || parsed_cpu.variant || parsed_cpu.revision) {
+      PERFETTO_ILOG("Arm specific fields not found for cpu", PRIu32,
```
primiano: This seems wrong, ti should be "for cpu %" PRIu32. with the %, without the comma.
Also please jsut PERFETTO_DLOG, this seems too noisy for a production log.

src/profiling/perf/perf_producer.cc
```diff
@@ -479,13 +499,35 @@ void PerfProducer::StartDataSource(DataSourceInstanceID ds_id,
   }
 
   std::vector<EventReader> per_cpu_readers;
+  auto cpuids = GetCpuID();
```
primiano: make the type explicit, makes the code more readable.

src/profiling/perf/perf_producer.cc
```diff
@@ -479,13 +499,35 @@ void PerfProducer::StartDataSource(DataSourceInstanceID ds_id,
...
+
   for (uint32_t cpu : target_cpus) {
+    using std::begin;
+    using std::end;
+    // Filter using the CPUID
+    if (!event_config_pb.cpuid().empty()) {
+      const auto& config_ids = event_config_pb.cpuid();
+      const auto& cpuid = cpuids[cpu];
```
primiano: this feels very prone to an out of bound access if the cardinality of the two independent data structures is not matching.

src/profiling/perf/perf_producer.cc
```diff
@@ -264,6 +265,25 @@ uint32_t TimeToNextReadTickMs(DataSourceInstanceID ds_id, uint32_t period_ms) {
...
+          base::Uint64ToHexStringNoPrefix(cpu.part.value());
+      if (cpu.variant) {
+        cpuid += base::Uint64ToHexStringNoPrefix(cpu.variant.value());
+        if (cpu.revision) {
+          cpuid += base::Uint64ToHexStringNoPrefix(cpu.revision.value());
+        }
+      }
+      result.push_back(cpuid);
+    }
+  }
```
primiano: maybe this should be a method (std::string GetCpuID) in base::CpuInfo? or does it only make sense in traced_perf?

protos/perfetto/trace/generic_kernel/generic_task.proto
```diff
@@ -71,7 +71,6 @@ message GenericKernelTaskRenameEvent {
   // New command name for the thread.
   optional string comm = 2;
 }
```
primiano, line 73: why this change? seems a stray change.

primiano (changes requested):

src/base/cpu_info.cc
abhmen01: Function removed and code inlined.

src/base/cpu_info.cc
abhmen01: Changed the code to use base::StringSplitter

src/traced/probes/system_info/system_info_data_source.cc
abhmen01: Fixed.

src/profiling/perf/perf_producer.cc
abhmen01: Fixed by changing GetCpuID to return a map keyed by cpu index.

protos/perfetto/trace/generic_kernel/generic_task.proto
abhmen01: Accidental change. Removed.

src/base/cpu_info.cc
```diff
@@ -0,0 +1,163 @@
...
+std::string ReadFile(const std::string& path) {
+  std::string contents;
+  if (!base::ReadFile(path, &contents))
+    return "";
+  return contents;
+}
+
+}  // namespace
+
+std::vector<CpuInfo> ParseCpuInfo(const std::string& proc_cpu_info) {
```
primiano: remove const& here, better to pass by value as StringSplitter manipulates the string.

src/base/cpu_info.cc
```diff
@@ -0,0 +1,163 @@
...
+    implementer = std::nullopt;
+    architecture = std::nullopt;
+    variant = std::nullopt;
+    part = std::nullopt;
+    revision = std::nullopt;
+    features = 0;
+  };
+
+  for (base::StringSplitter lines(
+           proc_cpu_info, '\n',
```
primiano: I think this might cause some weird bug because you are forcing proc_cpu_info to be a temporary copy (maybe it works, but feels very borderline).
Better to pass the string by value here, so you can manipulate here
(StringSplitter modifies the passed string)

src/profiling/perf/perf_producer.cc
primiano: This comment still holds!

abhmen01: Yes. It makes sense to move it to the src/base. The change is done.

src/base/cpu_info.cc
abhmen01: Changed.

src/base/cpu_info.cc
abhmen01: Fixed.

src/base/cpu_info.h
```diff
@@ -0,0 +1,53 @@
...
+};
+
+// Parses the contents of the input string into per-CPU entries.
+std::vector<CpuInfo> ParseCpuInfo(std::string proc_cpu_info);
+
+// Reads /proc/cpuinfo and parses it into per-CPU entries.
+std::vector<CpuInfo> ReadCpuInfo();
+
+// Reads /proc/cpuinfo and returns CPU IDs keyed by cpu index.
+std::map<uint32_t, std::string> GetCpuID();
```
primiano: This is an odd API how has been implemented:
1. Why this is a map and not an array with the same cardinality of the vector redurned by ReadCpuInfo
2. 
3. More importantly, why this is not a method on the CpuInfo struct itself as i originally suggested? It seems you can work it out purely from the data from the struct. I mean this should be `std::string CpuInfo::GetArmCPUID` (and just return empty string for other CPUs)

even better, the cpuid is so cheap to compute and small, that I'd just compute it always when you read the cpuinfo and stash it into a `char arm_cpuid[32]{}` (how many chars is it in practice?)

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -2394,6 +2394,12 @@ message PerfEventConfig {
   // Available since: perfetto v50.
   repeated uint32 target_cpu = 20;
 
+  // Allow failure of perf_event_open
+  optional bool ignore_open_failure = 21;
+
+  // Filter based on CPUID
```
primiano: can you please explain what's the semantic of the repeated field?
I cannot tell if it's
1) Match any of the following CPU ids
2) cpuid[nth] must match the cpuid of the Nth cpu

(i suspect it's 1. but wuold be nice to confirm and comment)

src/base/cpu_info.cc
```diff
@@ -0,0 +1,186 @@
...
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "perfetto/ext/base/file_utils.h"
+#include "perfetto/ext/base/string_splitter.h"
+#include "perfetto/ext/base/string_utils.h"
+#include "perfetto/ext/base/utils.h"
+#include "src/traced/probes/system_info/cpu_info_features_allowlist.h"
```
primiano: This is a big layering violation as-is.
the fix is easy thoguh. the cpu_info_features_allowlist must be moved in base next to cpu_info.h

include/perfetto/ext/base/cpu_info.h
```diff
@@ -0,0 +1,53 @@
+/*
```
primiano, line 1: putting in src/base is a bit odd. all the other headers are in include/perfetto/ext/base. please put the .h there (together with the allowlist, see the other comment)

src/traced/probes/system_info/system_info_data_source.cc
```diff
@@ -82,105 +49,41 @@ void SystemInfoDataSource::Start() {
...
+        parsed_cpu.variant && parsed_cpu.revision) {
+      auto* identifier = cpu->set_arm_identifier();
+      identifier->set_implementer(parsed_cpu.implementer.value());
+      identifier->set_architecture(parsed_cpu.architecture.value());
+      identifier->set_variant(parsed_cpu.variant.value());
+      identifier->set_part(parsed_cpu.part.value());
+      identifier->set_revision(parsed_cpu.revision.value());
+    } else if (parsed_cpu.implementer || parsed_cpu.architecture ||
+               parsed_cpu.part || parsed_cpu.variant || parsed_cpu.revision) {
+      PERFETTO_DLOG("Arm specific fields not found for cpu%", PRIu32,
```
primiano: same bug here. no comma between %" and PRIu32

primiano (changes requested):

src/base/cpu_info.h
abhmen01: I agree with your comments about the `GetCpuID` function. It is not generic, but specific to the usage in traced_perf where we need to find the `CpuInfo` for given cpu index. We can add the `arm_cpuid` field in `CpuInfo` and use the return value of ReadCpuInfo directly. I find an issue doing so. 

We currently assume the parsed CPU list is dense and ordered (i.e. `cpus[i].cpu_index == i`), which allows direct indexing. However, this is only enforced via `PERFETTO_DCHECK`, so in release builds we could end up with missing or out-of-order CPUs and incorrect lookups. (Code if refactored and not originally written by me)

To make the lookup more robust, I’m considering a couple of options at the caller (**traced_perf**) side. 
	1.	Always use `std::find_if` on cpu index. 
	2.	Build a `std::unordered_map<uint32_t, CpuInfo>` and 
	3.	Detect once whether the vector is dense/ordered, and:
	       - use direct indexing in the fast path
	       - fall back to `std::find_if` otherwise. 

Do you have a preference or any other ideas here?

primiano: Oh I see, so you are worried about the fact that somehow the "processor: " line in /proc/cpuinfo doesn't match with the index in the array? I guess it can happen with offline CPUs actually, so maybe we should even remove that dcheck.
Honestly I'd just do a std::find_if. This code is only run when starting a tracing sessions. Even if have have 256 CPUs... how long can it take to scan 256 elements (which would happen only if you specifcy a cpuid)?
it's still in the sub-ms range so I'd just keep it simple and go with that.

src/traced/probes/system_info/system_info_data_source.cc
abhmen01: Fixed

include/perfetto/ext/base/cpu_info.h
abhmen01: Had this in mind. Thank you for pointing it out. Moved it to include/perfetto/ext/base

src/base/cpu_info.cc
abhmen01: Moved it to include/perfetto/ext/base

protos/perfetto/trace/perfetto_trace.proto
abhmen01: Match any of the following CPU ids. Added the comment.

src/base/cpu_info.h
abhmen01: 1. Removed the previous `GetCpuId` function
2. Added `char arm_cpuid[32] `and computed in the `ParseCpuInfo` function itself. 
3. Added `CpuInfo::GetArmCpuId`
4. Removed the `DCHECKS`.
5. Changed the code in perf_producer.cc to use `std::find `on the `vector<CpuInfo>` to get the arm_cpuid.

include/perfetto/ext/base/cpu_info.h
```diff
@@ -0,0 +1,51 @@
...
+  uint32_t cpu_index = 0;
+  std::optional<uint32_t> implementer;
+  std::optional<uint32_t> architecture;
+  std::optional<uint32_t> variant;
+  std::optional<uint32_t> part;
+  std::optional<uint32_t> revision;
+  uint64_t features = 0;
+  char arm_cpuid[32] = {};
+
+  const char* GetArmCpuId() const { return arm_cpuid; }
```
primiano: why this getter? this doens't do anything other than returning an already available public member

primiano (changes requested):
One final comment left. LGTM after that.

include/perfetto/ext/base/cpu_info.h
abhmen01: It's not needed. Removed.

primiano (approved):

LalitMaganti (approved):

PR #2035, opened 2025-07-03
profiling: symbolizer: adds llvm symbolizer

OmarElfouly:
Adds llvm symbolizer class which using a dynamic library and dlopen to link to LLVM. This was necessary in order to eventually allow #1853 to be implemented without relying on a subprocess.

Statically linking LLVM is not viable because Perfetto uses libc++ while system LLVM is often built with libstdc++, leading to linker errors.

### DIFF
--- a/protos/perfetto/trace/perfetto_trace.proto
+++ b/protos/perfetto/trace/perfetto_trace.proto
@@ -16388,7 +16388,7 @@ message TrackDescriptor {
   //
   // Note: if the parent is set to a track with `thread` or `process` fields
   // set, the track will *not* be nested under the parent in the UI and in the
-  // trace processor data model. Instead, the track will inherit the parent'
+  // trace processor data model. Instead, the track will inherit the parent's
   // thread/process association and will appear as a *sibling* of the parent.
   // This semantic exists for back-compat reasons as the UI used to work this
   // way for years and changing this leads to a lot of traces subtly breaking.
@@ -16396,8 +16396,15 @@ message TrackDescriptor {
   // the parent.
   optional uint64 parent_uuid = 5;
 
-  // Name of the track. Optional but *strongly recommended* to be specified in a
-  // `TrackDescriptor` emitted before any `TrackEvent`s on the same track.
+  // Name of the track.
+  //
+  // Optional but *strongly recommended* to be specified in a `TrackDescriptor`
+  // emitted before any `TrackEvent`s on the same track.
+  //
+  // Note: any name specified here will be *ignored* for the root thread scoped
+  // tracks when `disallow_merging_with_system_tracks` is not set, as in this
+  // case, the name of the track is shared by many different data sources and so
+  // is centrally controlled by trace processor.
   //
   // It's strongly recommended to only emit the name for a track uuid *once*. If
   // a descriptor *has* to be emitted multiple times (e.g. between different
@@ -16416,7 +16423,7 @@ message TrackDescriptor {
   //   * The *last* non-null name in the whole trace according to trace order
   //     will be used.
   //   * If no non-null name is present in the whole trace, the trace processor
-  //     will fall back to other sources to provide a name for the track (e.g.
+  //     may fall back to other sources to provide a name for the track (e.g.
   //     the first event name for slice tracks, the counter name for counter
   //     tracks). This is implementation defined and may change at any time.
   //
@@ -16432,7 +16439,7 @@ message TrackDescriptor {
   //     descriptor with the name was emitted before the first event on a
   //     descendant track (B).
   //   * If no non-null name is present before the event is processed, the trace
-  //     processor will fall back to other sources to provide a name for the
+  //     processor may fall back to other sources to provide a name for the
   //     track (e.g. the first event name for slice tracks, the counter name for
   //     counter tracks). This is implementation defined and may change at any
   //     time.
--- a/protos/perfetto/trace/track_event/track_descriptor.proto
+++ b/protos/perfetto/trace/track_event/track_descriptor.proto
@@ -63,7 +63,7 @@ message TrackDescriptor {
   //
   // Note: if the parent is set to a track with `thread` or `process` fields
   // set, the track will *not* be nested under the parent in the UI and in the
-  // trace processor data model. Instead, the track will inherit the parent'
+  // trace processor data model. Instead, the track will inherit the parent's
   // thread/process association and will appear as a *sibling* of the parent.
   // This semantic exists for back-compat reasons as the UI used to work this
   // way for years and changing this leads to a lot of traces subtly breaking.
@@ -71,8 +71,15 @@ message TrackDescriptor {
   // the parent.
   optional uint64 parent_uuid = 5;
 
-  // Name of the track. Optional but *strongly recommended* to be specified in a
-  // `TrackDescriptor` emitted before any `TrackEvent`s on the same track.
+  // Name of the track.
+  //
+  // Optional but *strongly recommended* to be specified in a `TrackDescriptor`
+  // emitted before any `TrackEvent`s on the same track.
+  //
+  // Note: any name specified here will be *ignored* for the root thread scoped
+  // tracks when `disallow_merging_with_system_tracks` is not set, as in this
+  // case, the name of the track is shared by many different data sources and so
+  // is centrally controlled by trace processor.
   //
   // It's strongly recommended to only emit the name for a track uuid *once*. If
   // a descriptor *has* to be emitted multiple times (e.g. between different
@@ -91,7 +98,7 @@ message TrackDescriptor {
   //   * The *last* non-null name in the whole trace according to trace order
   //     will be used.
   //   * If no non-null name is present in the whole trace, the trace processor
-  //     will fall back to other sources to provide a name for the track (e.g.
+  //     may fall back to other sources to provide a name for the track (e.g.
   //     the first event name for slice tracks, the counter name for counter
   //     tracks). This is implementation defined and may change at any time.
   //
@@ -107,7 +114,7 @@ message TrackDescriptor {
   //     descriptor with the name was emitted before the first event on a
   //     descendant track (B).
   //   * If no non-null name is present before the event is processed, the trace
-  //     processor will fall back to other sources to provide a name for the
+  //     processor may fall back to other sources to provide a name for the
   //     track (e.g. the first event name for slice tracks, the counter name for
   //     counter tracks). This is implementation defined and may change at any
   //     time.
--- a/src/base/flat_hash_map_unittest.cc
+++ b/src/base/flat_hash_map_unittest.cc
@@ -71,7 +71,7 @@ struct Value {
   int id = instances++;
 };
 
-struct Hasher {
+struct KeyHasher {
   size_t operator()(const Key& k) const { return static_cast<size_t>(k.val); }
 };
 
@@ -79,7 +79,7 @@ int Key::instances = 0;
 int Value::instances = 0;
 
 TYPED_TEST(FlatHashMapTest, NonTrivialKeyValues) {
-  FlatHashMap<Key, Value, Hasher, typename TestFixture::Probe> fmap;
+  FlatHashMap<Key, Value, KeyHasher, typename TestFixture::Probe> fmap;
 
   for (int iteration = 0; iteration < 3; iteration++) {
     const int kNum = 10;
@@ -99,7 +99,7 @@ TYPED_TEST(FlatHashMapTest, NonTrivialKeyValues) {
     ASSERT_EQ(Key::instances, kNum - 3);
     ASSERT_EQ(Value::instances, kNum - 3);
 
-    FlatHashMap<Key, Value, Hasher, typename TestFixture::Probe> fmap2(
+    FlatHashMap<Key, Value, KeyHasher, typename TestFixture::Probe> fmap2(
         std::move(fmap));
     ASSERT_EQ(fmap.size(), 0u);
     ASSERT_EQ(fmap2.size(), static_cast<size_t>(kNum - 3));
@@ -169,7 +169,7 @@ TYPED_TEST(FlatHashMapTest, AllTagsAreValid) {
 }
 
 TYPED_TEST(FlatHashMapTest, FillWithTombstones) {
-  FlatHashMap<Key, Value, Hasher, typename TestFixture::Probe> fmap(
+  FlatHashMap<Key, Value, KeyHasher, typename TestFixture::Probe> fmap(
       /*initial_capacity=*/0, /*load_limit_pct=*/100);
 
   for (int rep = 0; rep < 3; rep++) {
@@ -384,7 +384,7 @@ TYPED_TEST(FlatHashMapTest, VsUnorderedMap) {
 
 TYPED_TEST(FlatHashMapTest, Clear) {
   {
-    FlatHashMap<Key, Value, Hasher, typename TestFixture::Probe> fmap;
+    FlatHashMap<Key, Value, KeyHasher, typename TestFixture::Probe> fmap;
     ASSERT_EQ(Key::instances, 0);
     ASSERT_EQ(Value::instances, 0);
 
--- a/src/trace_processor/export_json_unittest.cc
+++ b/src/trace_processor/export_json_unittest.cc
@@ -794,8 +794,11 @@ TEST_F(ExportJsonTest, InstantEvent) {
 
   // Global track.
   TrackEventTracker track_event_tracker(&context_);
-  TrackId track2 = *track_event_tracker.GetDescriptorTrack(
-      TrackEventTracker::kDefaultDescriptorTrackUuid);
+  TrackId track2 =
+      track_event_tracker
+          .GetDescriptorTrack(TrackEventTracker::kDefaultDescriptorTrackUuid,
+                              kNullStringId, std::nullopt)
+          ->track_id();
   context_.storage->mutable_slice_table()->Insert(
       {kTimestamp2, 0, track2, cat_id, name_id, 0, 0, 0});
 
@@ -803,7 +806,9 @@ TEST_F(ExportJsonTest, InstantEvent) {
   TrackEventTracker::DescriptorTrackReservation reservation;
   reservation.parent_uuid = 0;
   track_event_tracker.ReserveDescriptorTrack(1234, reservation);
-  TrackId track3 = *track_event_tracker.GetDescriptorTrack(1234);
+  TrackId track3 =
+      track_event_tracker.GetDescriptorTrack(1234, kNullStringId, std::nullopt)
+          ->track_id();
   context_.storage->mutable_slice_table()->Insert(
       {kTimestamp3, 0, track3, cat_id, name_id, 0, 0, 0});
 
--- a/src/trace_processor/importers/proto/track_event_event_importer.h
+++ b/src/trace_processor/importers/proto/track_event_event_importer.h
@@ -317,42 +317,37 @@ class TrackEventEventImporter {
     //      TrackEvent types), or
     //   b) a default track.
     if (track_uuid_) {
-      std::optional<TrackId> opt_track_id =
-          track_event_tracker_->GetDescriptorTrack(track_uuid_, name_id_,
-                                                   packet_sequence_id_);
-      if (!opt_track_id) {
+      auto opt_resolved = track_event_tracker_->GetDescriptorTrack(
+          track_uuid_, name_id_, packet_sequence_id_);
+      if (!opt_resolved) {
         return base::ErrStatus(
             "track_event_parser: unable to find track matching UUID %" PRIu64,
             track_uuid_);
       }
-      track_id_ = *opt_track_id;
-
-      auto rr = storage_->mutable_track_table()->FindById(track_id_);
-      if (rr && rr->utid()) {
-        utid_ = rr->utid();
-        upid_ = storage_->thread_table()[*utid_].upid();
-      } else if (rr && rr->upid()) {
-        upid_ = rr->upid();
-        if (sequence_state_->pid_and_tid_valid()) {
-          auto pid = static_cast<uint32_t>(sequence_state_->pid());
-          auto tid = static_cast<uint32_t>(sequence_state_->tid());
-          UniqueTid utid_candidate = procs->UpdateThread(tid, pid);
-          if (storage_->thread_table()[utid_candidate].upid() == upid_) {
-            legacy_passthrough_utid_ = utid_candidate;
+      track_id_ = opt_resolved->track_id();
+      switch (opt_resolved->scope()) {
+        case TrackEventTracker::ResolvedDescriptorTrack::Scope::kThread:
+          utid_ = opt_resolved->utid();
+          upid_ = storage_->thread_table()[*utid_].upid();
+          break;
+        case TrackEventTracker::ResolvedDescriptorTrack::Scope::kProcess:
+          upid_ = opt_resolved->upid();
+          if (sequence_state_->pid_and_tid_valid()) {
+            auto pid = static_cast<uint32_t>(sequence_state_->pid());
+            auto tid = static_cast<uint32_t>(sequence_state_->tid());
+            UniqueTid utid_candidate = procs->UpdateThread(tid, pid);
+            if (storage_->thread_table()[utid_candidate].upid() == upid_) {
+              legacy_passthrough_utid_ = utid_candidate;
+            }
           }
-        }
-      } else {
-        if (rr) {
-          StringPool::Id id = rr->name();
-          if (id.is_null()) {
-            rr->set_name(name_id_);
+          break;
+        case TrackEventTracker::ResolvedDescriptorTrack::Scope::kGlobal:
+          if (sequence_state_->pid_and_tid_valid()) {
+            auto pid = static_cast<uint32_t>(sequence_state_->pid());
+            auto tid = static_cast<uint32_t>(sequence_state_->tid());
+            legacy_passthrough_utid_ = procs->UpdateThread(tid, pid);
           }
-        }
-        if (sequence_state_->pid_and_tid_valid()) {
-          auto pid = static_cast<uint32_t>(sequence_state_->pid());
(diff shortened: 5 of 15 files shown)


src/profiling/symbolizer/BUILD.gn
```diff
@@ -58,6 +65,50 @@ if (enable_perfetto_trace_processor) {
...
+    sources = [
+      "llvm_symbolizer.cc",
+      "llvm_symbolizer.h",
+    ]
+    deps = [
+      ":common",
+      ":llvm_symbolizer_c_api",
+      "../../../gn:default_deps",
+    ]
+    public_deps = [ "../../../include/perfetto/ext/base" ]
```
LalitMaganti: This shouldn't be a public dep.

src/profiling/symbolizer/BUILD.gn
```diff
@@ -58,6 +65,50 @@ if (enable_perfetto_trace_processor) {
...
+
+    libs = _llvm_flags_output.libs
+    ldflags = _llvm_flags_output.ldflags
+    cflags = _llvm_flags_output.cflags
+    defines = _llvm_flags_output.defines
+
+    cflags += [ "-fvisibility=hidden" ]  # Not currently necessary, but good for
+                                         # future-proofing.
+  }
+  source_set("llvm_symbolizer_c_api") {
```
LalitMaganti, line 82: Please add a default_deps dependency for this file.

src/profiling/symbolizer/common.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef SRC_PROFILING_SYMBOLIZER_COMMON_H_
+#define SRC_PROFILING_SYMBOLIZER_COMMON_H_
+
+#include <cstdint>
+#include <string>
+
+namespace perfetto {
+namespace profiling {
+
+struct SymbolizedFrame {
```
LalitMaganti: I would suggest we design a batched API not a 1:1. It's going to be very inefficient if we need to do a cross-library function call on every symbol.

src/profiling/symbolizer/common.h
```diff
@@ -0,0 +1,40 @@
...
+  std::string function_name;
+  std::string file_name;
+  uint32_t line = 0;
+};
+
+struct SymbolizationRequest {
```
LalitMaganti: We should really design a batched API.

src/profiling/symbolizer/llvm_symbolizer.cc
```diff
@@ -0,0 +1,147 @@
...
+LlvmSymbolizer& LlvmSymbolizer::operator=(LlvmSymbolizer&& other) noexcept {
+  if (this == &other) {
+    return *this;
+  }
+
+  if (c_api_symbolizer_) {
+    destroy_fn_(c_api_symbolizer_);
+  }
+  if (library_handle_) {
+    dlclose(library_handle_);
```
LalitMaganti: Don't bother with dlclose. It is incredibly flaky and rarely ever works.

src/profiling/symbolizer/llvm_symbolizer_c_api.h
```diff
@@ -0,0 +1,70 @@
...
+// Destroys an instance of the LLVM symbolizer.
+LLVM_SYMBOLIZER_C_API void LlvmSymbolizer_Destroy(LlvmSymbolizer* sym);
+
+// Symbolizes a given address within a given binary/module.
+// The caller is responsible for freeing the result with
+// LlvmSymbolizer_FreeSymbolizationResult.
+LLVM_SYMBOLIZER_C_API SymbolizationResult
+LlvmSymbolizer_Symbolize(LlvmSymbolizer* sym,
+                         const char* binary_path,
+                         uint64_t address);
```
LalitMaganti: This should be a barched API

src/profiling/symbolizer/BUILD.gn
```diff
@@ -58,6 +65,50 @@ if (enable_perfetto_trace_processor) {
   }
 }
 
+if (enable_perfetto_llvm_symbolizer) {
+  _llvm_flags_output =
+      exec_script("../../../tools/get_llvm_flags.py", [], "scope")
```
primiano, line 63: do we really need all those flags? are they realistically ever going to change?
IMHO it's more pragmatic to bake the flags in here, and change them if and whenever they chagne to catch up.
Unless there is a good reason why those flags legitimately need to be polled at runtime
I know about pkgconfig.... but I also know in practice in most cases those flags never change. I'd not add this level of complexity unless there is a strong case why it's needed.
Can't we use the paths we get from the existing linux_find_lvvm.py?

src/profiling/symbolizer/common.h
primiano: +1. also think carefully about the types.
I feel in this design you are spending most time allocating and deallocating std::strings

src/profiling/symbolizer/BUILD.gn
OmarElfouly: I agree as on my machine only a small subset of these were necessary, however my main concern was that since this will be built on different systems and different llvm versions then the following becomes true (I think :) ):

- ldflags becomes necessary to tell the linker where to find the specific llvm installation's lib (in case it is not in a place the linker can find on its own)
- libs becomes necessary as the name of the main llvm lib changes based on the version
- cflags and defines I am not entirely sure about but the LLVM config tool describes them as necessary flags for the compiler, however on my machine the only truly necessary flag appears to be the -isystem flag.

For reference here is the result of running the get_llvm_flags.py tool on my machine:
```
defines = [
  "_GNU_SOURCE",
  "__STDC_CONSTANT_MACROS",
  "__STDC_FORMAT_MACROS",
  "__STDC_LIMIT_MACROS"
]
cflags = [
  "-fno-exceptions",
  "-funwind-tables",
  "-isystem/usr/lib/llvm-19/include",
  "-std=c++17"
]
libs = [
  "LLVM-19"
]
ldflags = [
  "-L/usr/lib/llvm-19/lib"
]
```

src/profiling/symbolizer/BUILD.gn
OmarElfouly: Fixed

src/profiling/symbolizer/BUILD.gn
OmarElfouly: I think I fixed it as long as you meant to add `deps = [ "../../../gn:default_deps" ]` to the llvm_symbolizer_c_api source_set

src/profiling/symbolizer/llvm_symbolizer_unittest.cc
```diff
@@ -0,0 +1,82 @@
...
+    result += ":";
+    result += std::to_string(frame.line);
+    result += ":0\n";
+  }
+  return result;
+}
+
+// Tests symbolization for both a normal function and an inlined function.
+TEST(LlvmSymbolizerTest, Symbolize) {
+  // ust be updated if the binary is recompiled.
```
rasikanavarange: nit: typo "Must"

src/profiling/symbolizer/llvm_symbolizer_unittest.cc
```diff
@@ -0,0 +1,82 @@
...
+
+// Tests that the LlvmSymbolizer can be constructed and destructed. This
+// implicitly tests that the dynamic library can be loaded and the necessary
+// symbols can be resolved.
+TEST(LlvmSymbolizerTest, ConstructDestruct) {
+  LlvmSymbolizer symbolizer;
+}
+
+// Helper to format the symbolized result for comparison.
+std::string FormatFrames(const std::vector<SymbolizedFrame>& frames) {
```
rasikanavarange: Do we need this format? Can we not just test equality for function_name, file_name and line separately?
e.g. ASSERT_EQ("TestFunctionToSymbolize", frame.function_name) etc.

src/profiling/symbolizer/common.h
rasikanavarange: +1

src/profiling/symbolizer/llvm_symbolizer_wrapper.cc
```diff
@@ -0,0 +1,149 @@
...
+  opts.PathStyle =
+      llvm::DILineInfoSpecifier::FileLineInfoKind::AbsoluteFilePath;
+  symbolizer_ = std::make_unique<llvm::symbolize::LLVMSymbolizer>(opts);
+}
+
+LlvmSymbolizerImpl::~LlvmSymbolizerImpl() = default;
+
+SymbolizationResult LlvmSymbolizerImpl::Symbolize(const char* binary_path,
+                                                  uint64_t address) {
+#if LLVM_VERSION_MAJOR >= 11
```
rasikanavarange: Could you leave a comment explaining this?

src/profiling/symbolizer/llvm_symbolizer_unittest.cc
OmarElfouly: Oh yeah using this function only really made sense when I was comparing the output with llvm-symbolizer, however since the unittest doesn't do that I will change it to ASSERT_EQ("TestFunctionToSymbolize", frame.function_name) etc.

src/profiling/symbolizer/common.h
OmarElfouly: Done :)

src/profiling/symbolizer/common.h
OmarElfouly: I have improved the types now but I feel that I might have over-complicated it by adding a RAII wrapper

src/profiling/symbolizer/llvm_symbolizer.cc
OmarElfouly: It's currently been commented out. Should I completely remove it?

src/profiling/symbolizer/llvm_symbolizer_c_api.h
OmarElfouly: Done

src/profiling/symbolizer/llvm_symbolizer.cc
```diff
@@ -0,0 +1,186 @@
...
+    BatchSymbolizationResult c_api_result,
+    decltype(&::LlvmSymbolizer_FreeBatchSymbolizationResult) free_fn)
+    : c_api_result_(c_api_result), free_result_fn_(free_fn) {
+  if (c_api_result_.results == nullptr) {
+    return;
+  }
+  results_.reserve(c_api_result.num_results);
+  for (size_t i = 0; i < c_api_result.num_results; ++i) {
+    const SymbolizationResult& result = c_api_result.results[i];
+    std::vector<LlvmSymbolizedFrame> frames;
```
LalitMaganti: Creating a vector every request is going to tank performance.

src/profiling/symbolizer/llvm_symbolizer.cc
```diff
@@ -0,0 +1,186 @@
...
+    return;
+  }
+}
+
+LlvmSymbolizer::~LlvmSymbolizer() {
+  if (c_api_symbolizer_) {
+    destroy_fn_(c_api_symbolizer_);
+  }
+  // if (library_handle_) {
+  // dlclose(library_handle_);
```
LalitMaganti: PLease remove code, don't comment it out.

src/profiling/symbolizer/llvm_symbolizer.cc
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  if (c_api_symbolizer_) {
+    destroy_fn_(c_api_symbolizer_);
+  }
+  // if (library_handle_) {
+  // dlclose(library_handle_);
+  // }
+
+  library_handle_ = other.library_handle_;
```
LalitMaganti: Instead of this, if you use base::ScopedResource, all of this will be done automatically for you and you won't have to write a manual move constructor.

src/profiling/symbolizer/llvm_symbolizer_c_api.h
```diff
@@ -0,0 +1,82 @@
...
+typedef struct {
+  SymbolizedFrame* frames;
+  size_t num_frames;
+} SymbolizationResult;
+
+// Represents the result of a batch of symbolization operations.
+typedef struct {
+  SymbolizationResult* results;
+  size_t num_results;
+} BatchSymbolizationResult;
```
LalitMaganti, line 89: Instead of creating an array of arrays, I'd suggest flattening into a big vector and then doing offset/size ranges into that. Will be a lot more efficient.

src/profiling/symbolizer/llvm_symbolizer.cc
OmarElfouly: Done

src/profiling/symbolizer/llvm_symbolizer_c_api.h
OmarElfouly: Done :)

src/profiling/symbolizer/llvm_symbolizer.cc
OmarElfouly: Removed 👍

src/profiling/symbolizer/llvm_symbolizer_c_api.h
```diff
@@ -0,0 +1,92 @@
...
+extern "C" {
+#endif
+
+// Opaque handle to the underlying C++ symbolizer object.
+struct LlvmSymbolizer;
+typedef struct LlvmSymbolizer LlvmSymbolizer;
+
+// Represents a single symbolization request.
+typedef struct {
+  const char* binary_path;
```
LalitMaganti, line 37: I'd try to avoid assuming a null terminated string as we cannot always guarantee that. Can we take a pointer + size instead? You can take -1 to mean that the string is null terminated + the caller does not know the size (this is convention with e.g. SQLite. Please document this though).

src/profiling/symbolizer/llvm_symbolizer_c_api.h
```diff
@@ -0,0 +1,92 @@
...
+  const char* function_name;
+  const char* file_name;
+  uint32_t line_number;
+} SymbolizedFrame;
+
+// Represents the result of a single symbolization operation as a range in a
+// flattened array of frames.
+typedef struct {
+  // The offset in the `frames` array of the `BatchSymbolizationResult`.
+  size_t offset;
```
LalitMaganti: Here and below: avoid size_t across ABI boundaries as that can cause very hard to debug issues at it depends on the arch the library was compiled on. Use uint32_t.

src/trace_processor/BUILD.gn
```diff
@@ -267,6 +267,9 @@ if (enable_perfetto_trace_processor_sqlite) {
         "../../protos/perfetto/trace_summary:descriptor",
         "../../protos/third_party/pprof:descriptor",
       ]
+      if (enable_perfetto_llvm_symbolizer) {
+        data_deps += [ "../profiling/symbolizer:llvm_symbolizer_wrapper" ]
```
LalitMaganti, line 270: How will data_deps work on WASM? Is that something we've tested? Are we planning on testing that before landing?

src/profiling/symbolizer/llvm_symbolizer_wrapper.cc
```diff
@@ -0,0 +1,190 @@
...
+            {request.address, llvm::object::SectionedAddress::UndefSection});
+#else
+    llvm::Expected<llvm::DIInliningInfo> res_or_err =
+        symbolizer_->symbolizeInlinedCode(request.binary_path, request.address);
+#endif
+
+    if (!res_or_err) {
+      std::string err_msg;
+      llvm::raw_string_ostream os(err_msg);
+      llvm::logAllUnhandledErrors(res_or_err.takeError(), os,
```
LalitMaganti, line 95: Trace processor is a library and as such, all errors must be properly propogated to the caller. We should not be logging this here.

src/profiling/symbolizer/llvm_symbolizer.h
```diff
@@ -0,0 +1,121 @@
...
+  // pointers and sizes.
+  const LlvmSymbolizedFrame* all_frames_ptr_ = nullptr;
+  size_t num_total_frames_ = 0;
+
+  const SymbolizationResultRange* ranges_ptr_ = nullptr;
+  size_t num_ranges_ = 0;
+};
+
+namespace {
+// A no-op closer function for dlopen handles, as dlclose() is flaky.
```
LalitMaganti: You can make this a static private funciton of the class.

src/profiling/symbolizer/llvm_symbolizer.h
```diff
@@ -0,0 +1,121 @@
...
+// ScopedResource requires a static function pointer for its template.
+inline int NoOpDlclose(void* /*handle*/) {
+  return 0;
+}
+}  // namespace
+
+class LlvmSymbolizer {
+ public:
+  LlvmSymbolizer();
+  ~LlvmSymbolizer();
```
LalitMaganti: The move constructor/copy constructor/destructor can all be deleted now.

src/profiling/symbolizer/llvm_symbolizer.h
```diff
@@ -0,0 +1,121 @@
...
+ private:
+  friend class LlvmSymbolizer;
+
+  SymbolizationResultBatch(
+      BatchSymbolizationResult c_api_result,
+      decltype(&::LlvmSymbolizer_FreeBatchSymbolizationResult) free_fn);
+
+  void Free();
+
+  BatchSymbolizationResult c_api_result_{};
```
LalitMaganti: As before, if you use ScopedFile here, you can avoid the destructor/move constructor.

src/profiling/symbolizer/llvm_symbolizer_c_api.h
OmarElfouly: @LalitMaganti is it still hard to guarantee a null terminated string considering LlvmSymbolizer takes a string initially in its request and then generates a char * using .c_str() for the c api?

https://github.com/google/perfetto/blob/e99879fd33601f0d6a76b387f77e9271268ca229/src/profiling/symbolizer/llvm_symbolizer.cc#L140

I initially chose this approach since the main source for a file name will probably be taken using something like:
```
std::string file_name =
      storage->GetString(storage->file_table().FindById(elf.file_id())->name())
          .ToStdString();
```

LalitMaganti: I'm fully expecting this API to be how we do all symbolization (including on the fly symbolization while parsing the trace) going forward.

I would not hardcode that fact here as it's going to be painful down the line.

OmarElfouly: Alright I will modify it to accept a pointer and size at all points in the code 👍

src/profiling/symbolizer/llvm_symbolizer_c_api.h
OmarElfouly: Done 👍

src/profiling/symbolizer/llvm_symbolizer_c_api.h
OmarElfouly: @LalitMaganti One issue I am having is that the LLVM function symbolizeInlinedCode, actually expects a string (in most LLVM versions) so while I could create a temp string in the case that size is not -1 I feel like that circles back to the issue of spending too much time allocating and deallocating. 

However I also understand that if other parts of the code calling this will only have access to a pointer, then it will just be a question of if the caller should be responsible for creating the std string or if the symbolize function is.

One approach is that since in a batch I believe the majority of binary_path strings will be identical, making callers manage a pool of strings that they reuse (or doing the same in our symbolize function) such that we pay the allocation cost only once per unique binary path, not once per request.

More recent LLVM version do use a StringRef instead of a string which will mean no extra cost, but the above still applies for other versions.

LalitMaganti: Ok given the complexity let's just assume null terminated string in that case but I still suggest passing the size to allow for future optimization using string ref.

OmarElfouly: Done

src/trace_processor/BUILD.gn
rasikanavarange: I believe @OmarElfouly is planning to test this, but this patch probably does not need to block on this imo

src/profiling/symbolizer/llvm_symbolizer.h
OmarElfouly: Done 👍

src/profiling/symbolizer/llvm_symbolizer.h
OmarElfouly: Done :)

src/profiling/symbolizer/llvm_symbolizer.h
OmarElfouly: Done

src/trace_processor/BUILD.gn
LalitMaganti: I'm fine to not block this patch on testing that. Just wanted to check the plan here.

src/profiling/symbolizer/llvm_symbolizer_wrapper.cc
OmarElfouly: Done 👍

src/profiling/symbolizer/llvm_symbolizer_c_api.h
```diff
@@ -0,0 +1,108 @@
...
+// Opaque handle to the underlying C++ symbolizer object.
+struct LlvmSymbolizer;
+typedef struct LlvmSymbolizer LlvmSymbolizer;
+
+// Represents a single symbolization request.
+typedef struct {
+  // Path to the binary file.
+  const char* binary_path;
+  // The length of binary_path. If the caller does not know the size and the
+  // string is null-terminated, this should be set to (uint32_t)-1.
```
LalitMaganti: uint32_t::max() - 1

src/profiling/symbolizer/llvm_symbolizer_c_api.h
```diff
@@ -0,0 +1,108 @@
...
+  // A flat array of all symbolized frames for the entire batch.
+  LlvmSymbolizedFrame* frames;
+  // The total number of frames in the `frames` array.
+  uint32_t total_frames;
+  // An array of `SymbolizationResultRange` structs, each representing a range
+  // in the `frames` array.
+  SymbolizationResultRange* ranges;
+  // The number of ranges, corresponding to the number of original requests.
+  uint32_t num_ranges;
+  SymbolizationError* errors;
```
LalitMaganti, line 86: Need some comments here and below.

LalitMaganti (approved):

gn/perfetto.gni
```diff
@@ -362,6 +362,11 @@ declare_args() {
       is_linux && enable_perfetto_trace_processor &&
       perfetto_build_standalone && !is_perfetto_build_generator
 
+  # Enables the use of the LLVM symbolizer in trace_processor.
+  enable_perfetto_llvm_symbolizer =
```
rasikanavarange: nit: Just a small concern about enabling this by default in these selected builds. Do we want to force users who have not explicitly enabled this to have LLVM on their machines? I wonder if this should be `enable_perfetto_llvm_symbolizer = false` for now?

src/profiling/symbolizer/llvm_symbolizer_c_api.h
```diff
@@ -36,8 +36,11 @@ typedef struct {
   // Path to the binary file.
   const char* binary_path;
   // The length of binary_path. If the caller does not know the size and the
-  // string is null-terminated, this should be set to (uint32_t)-1.
+  // string is null-terminated, this should be set to uint32_t::max() - 1.
```
LalitMaganti: Actually I think i misunderstood your original comment. I think it should just be uint32_t::max() without the -1

gn/perfetto.gni
LalitMaganti: The problem with that is that you are not going to be able to test anything on the CI.

Edit: CI not UI.

src/profiling/symbolizer/llvm_symbolizer_c_api.h
OmarElfouly: Done 👍

gn/perfetto.gni
rasikanavarange: Yes, I guess as long as this new dependency is documented somewhere it should be okay :)

OmarElfouly: I am not sure but I agree it does seem a bit unfair for people who do not care for symbolization.
@LalitMaganti Which option would you prefer?

OmarElfouly: Oops I had not refreshed

LalitMaganti: I mean another option is that you add a CI flag to enable this explicitly on the CI to get coverage there and disable it by default otherwise.

I would prefer that if at all possible.

src/profiling/symbolizer/BUILD.gn
primiano: Ah fair. ack.

gn/perfetto.gni
LalitMaganti: Can we make this false instead and then change the option at https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/.github/workflows/linux-tests.yml;l=34 instead?

OmarElfouly: @LalitMaganti I saw that CI was being passed as an env var so the plan is to use a is_ci = getenv("CI") != ""
Should I continue with this plan or instead use https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/.github/workflows/linux-tests.yml;l=34?

LalitMaganti: I'd prefer changing this to false and change the yml instead of using env varaibles. We don't really use that in GN.

OmarElfouly: @LalitMaganti In that case should I keep the is_ci arg and set is_ci = false in perfetto.gni as true in the linux tests or should I set enable_llvm_symbolizer to be false and enable enable_llvm_symbolizer as an arg to the CI tests that make sense?

LalitMaganti: The latter. Please remove is_ci, we really shouldn't need that.

PR #2036, opened 2025-07-03
ui: Add option to connect to custom consumer socket

pan-:
Testing changes in traced and related probes from the UI requires rebuilding Android or manually replacing system Perfetto binaries, which is cumbersome.

This commit introduces a new option allowing the UI to connect to an alternative tracing service running on the target system, instead of the default Perfetto daemon.

When the feature flag connectToLocalAbstractTraceboxSocket is enabled, the recording plugin connects via a local abstract socket named "@tracebox_traced_consumer". To use this feature, start tracebox and websocket_bridge with the environment variable
PERFETTO_CONSUMER_SOCK_NAME=@tracebox_traced_consumer set.

--- 

Note:

My team regularly updates `traced` and works primarily with Android devices that have root access but a read-only system partition. Historically, we maintained internal code that automatically pushed tracebox to the device and started it as part of the recording process whenever a new binary was available. While this made distributing "complete" versions of Perfetto convenient internally, it is impractical for external sharing or upstreaming.

Having a straightforward method to connect to an alternative tracing service directly from the UI would significantly ease the workflow for testing UI changes dependent on updates to `traced`. Hence this proposed change.

Regarding the implementation in this PR, I'm not entirely satisfied. Initially, I tried following the approach used in the [QueryPage plugin](https://github.com/google/perfetto/blob/main/ui/src/plugins/dev.perfetto.QueryPage/index.ts#L33). Unfortunately, `adb_tracing_session.ts` cannot reference `index.ts` without creating circular dependencies. Passing the flag through multiple function layers down to `createAdbTracingSession` seemed excessive. To avoid this, I introduced `local_flag.ts` to centrally store the local flag without circular dependencies—but this solution feels out of place. Guidance or suggestions on a cleaner approach would be greatly appreciated.

### DIFF
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/adb/adb_tracing_session.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/adb/adb_tracing_session.ts
@@ -18,15 +18,15 @@ import {TracingProtocol} from '../tracing_protocol/tracing_protocol';
 import {errResult, okResult, Result} from '../../../base/result';
 import {exists} from '../../../base/utils';
 import {ConsumerIpcTracingSession} from '../tracing_protocol/consumer_ipc_tracing_session';
-
-export const CONSUMER_SOCKET = '/dev/socket/traced_consumer';
+import {RecordTraceV2Settings} from '../settings';
 
 export async function createAdbTracingSession(
   adbDevice: AdbDevice,
   traceConfig: protos.ITraceConfig,
+  settings: RecordTraceV2Settings,
 ): Promise<Result<ConsumerIpcTracingSession>> {
   const streamStatus = await adbDevice.createStream(
-    `localfilesystem:${CONSUMER_SOCKET}`,
+    settings.getTracedConsumerSocketAddressForAdb(),
   );
   if (!streamStatus.ok) return streamStatus;
   const stream = streamStatus.value;
@@ -37,11 +37,15 @@ export async function createAdbTracingSession(
 
 export async function getAdbTracingServiceState(
   adbDevice: AdbDevice,
+  settings: RecordTraceV2Settings,
 ): Promise<Result<protos.ITracingServiceState>> {
-  const sock = CONSUMER_SOCKET;
-  const status = await adbDevice.createStream(`localfilesystem:${sock}`);
+  const status = await adbDevice.createStream(
+    settings.getTracedConsumerSocketAddressForAdb(),
+  );
   if (!status.ok) {
-    return errResult(`Failed to connect to ${sock}: ${status.error}`);
+    return errResult(
+      `Failed to connect to ${settings.getTracedConsumerSocketAddress()}: ${status.error}`,
+    );
   }
   const stream = status.value;
   using consumerPort = await TracingProtocol.create(stream);
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/index.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/index.ts
@@ -35,11 +35,16 @@ import {RecordingManager} from './recording_manager';
 import {TracedWebsocketTargetProvider} from './traced_over_websocket/traced_websocket_provider';
 import {WebDeviceProxyTargetProvider} from './adb/web_device_proxy/wdp_target_provider';
 import m from 'mithril';
+import {RecordTraceV2Settings} from './settings';
+
 export default class implements PerfettoPlugin {
   static readonly id = 'dev.perfetto.RecordTraceV2';
   private static recordingMgr?: RecordingManager;
 
   static onActivate(app: App) {
+    const settings = new RecordTraceV2Settings();
+    settings.registerSettings(app.settings);
+
     app.sidebar.addMenuItem({
       section: 'trace_files',
       text: 'Record new trace',
@@ -53,7 +58,7 @@ export default class implements PerfettoPlugin {
         return m(RecordPageV2, {
           subpage,
           app,
-          getRecordingManager: () => this.getRecordingManager(app),
+          getRecordingManager: () => this.getRecordingManager(app, settings),
         });
       },
     });
@@ -61,7 +66,7 @@ export default class implements PerfettoPlugin {
       id: 'dev.perfetto.RecordTraceV2.disconnectTarget',
       name: 'Disconnect the current device',
       callback: () => {
-        const recMgr = this.getRecordingManager(app);
+        const recMgr = this.getRecordingManager(app, settings);
         if (recMgr.currentTarget) {
           recMgr.currentTarget.disconnect();
         }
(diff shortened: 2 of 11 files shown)


ui/src/plugins/dev.perfetto.RecordTraceV2/index.ts
```diff
@@ -34,11 +34,19 @@ import {TracedWebsocketTargetProvider} from './traced_over_websocket/traced_webs
 import {savedConfigsPage} from './pages/saved_configs';
 import {WebDeviceProxyTargetProvider} from './adb/web_device_proxy/wdp_target_provider';
 import m from 'mithril';
+import {LOCAL_ABSTRACT_SOCKET_FLAG} from './local_flag';
 export default class implements PerfettoPlugin {
   static readonly id = 'dev.perfetto.RecordTraceV2';
   private static recordingMgr?: RecordingManager;
 
   static onActivate(app: App) {
+    LOCAL_ABSTRACT_SOCKET_FLAG.flag = app.featureFlags.register({
```
LalitMaganti: IMO this should be a user setting not a feature flag. Settings are the right place for user controllable behaviour which is not an implementation detail of the UI.

We used to abuse flags for that but now we have settings, it should live there.

ui/src/plugins/dev.perfetto.RecordTraceV2/adb/adb_tracing_session.ts
```diff
@@ -18,15 +18,37 @@ import {TracingProtocol} from '../tracing_protocol/tracing_protocol';
...
 
-export const CONSUMER_SOCKET = '/dev/socket/traced_consumer';
+function useCustomSocket(): boolean {
+  if (LOCAL_ABSTRACT_SOCKET_FLAG.flag === undefined) {
+    return false;
+  }
+  return LOCAL_ABSTRACT_SOCKET_FLAG.flag.get();
+}
+
+function getConsumerSocketType() {
```
primiano: do we ever need only one of them?
If not, could these two be folded into a single

getConsumerSocketForAdb() which returns either localabstract:traced.... or localfs:/dev/socket/... ?

ui/src/plugins/dev.perfetto.RecordTraceV2/local_flag.ts
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Flag} from '../../public/feature_flag';
+
+// Flag that indicates if a non-default socket should be used to connect to traced.
+// It is present in this file circular dependencies.
+export const LOCAL_ABSTRACT_SOCKET_FLAG: {flag?: Flag} = {};
```
primiano: This is actually quite pragmatic. I'd just move also the registration here for consistency

```
const settings = {
  consumerSocket: ...,
}

function registerRecordingSettings(app: App) {
 app.settings.register(...)
}

getConsumerSockerFlag() {
   // Calling this before registerRecordingSettings() should never
   // happen and should just trigger a crash
   assertExists(settings.consumerSocket). ....
}
```

primiano (commented):
Great overall, i'm very supportive of this. I have a few minor comments, on top of lalit's one.
Apollogies for the flags vs settings confusion. settings are fairly new and we haven't migrated "flags" over.
I just spoke with @stevegolton, what is going to happen at some point soon
- the majorify of the current "flags" will be migrated under settings
- The few things left that are actual "flags" will be moved UI-wise under settings in a section called "Experimental features"

But what you are doing here is a pure setting.

ui/src/plugins/dev.perfetto.RecordTraceV2/local_flag.ts
pan-: I created a `settings` object which can be use to include all the settings of the `RecordTraceV2` plugin. 
It includes the registration as requested.

ui/src/plugins/dev.perfetto.RecordTraceV2/adb/adb_tracing_session.ts
pan-: There are two accessors now: 
- `getTracedConsumerSocketAddressForAdb` which returns the address formated to be used with ADB.
- `getTracedConsumerSocketAddress` which returns the address stored in the settings. For any information printed to the user, it should be better to use this name as it matches what has been set in the setting.

pan-:
Thanks for your feedbacks, I wasn't aware of the Setting API. It is absolutely perfect for this use case. 
It is much better to set the address name to connect to rather than be forced to use a pre-defined one. 

I tidy up the setting registration part. It remains in its own file but it is not specific to the socket address. Other settings of the recording plugin can be added here too.

ui/src/plugins/dev.perfetto.RecordTraceV2/settings.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Return the address value in the setting
+  getTracedConsumerSocketAddress() {
+    if (this.socketAddressValue === undefined) {
+      return RecordTraceV2Settings.socketAddressDescriptor.defaultValue;
+    }
+    return this.socketAddressValue.get();
+  }
+}
+
+export default new RecordTraceV2Settings();
```
stevegolton: Please don't create more globals just to make access more convenient. Global state causes headaches for us. Perhaps you could inject these settings through from RecordingManager.

stevegolton (changes requested):

primiano:
Just checking, if this is ready and you wnat anybody to look at it, please write something.
you are pushing patchsets (which is fine) but cannot tell if you want us to look, or you are iterating.

ui/src/plugins/dev.perfetto.RecordTraceV2/index.ts
primiano: +1

abhmen01:
@primiano Please look at the changes. It is ready for review. I had force-pushed a wrong commit two days back, and it it corrected.

abhmen01:
@primiano There was a previous comment by @stevegolton that has not yet been addressed concerning the introduction of additional global state. I will address this and update the implementation accordingly. I would appreciate it if you could hold off on further review until then.

abhmen01:
Addressed the issue of 'settings' being a global. It is now injected through the RecordingManager. 
Please continue the review.

abhmen01:
Hi @primiano 
This is ready to be reviewed.

stevegolton (approved):
LGTM, thanks!

ui/src/plugins/dev.perfetto.RecordTraceV2/settings.ts
```diff
@@ -0,0 +1,61 @@
...
+  SettingDescriptor,
+  SettingsManager,
+} from '../../public/settings';
+
+// This class holds all the setting offered by the RecordTraceV2 plugin.
+// It is registered at plugin activation time.
+export class RecordTraceV2Settings {
+  static readonly baseId = 'dev.perfetto.RecordTraceV2#';
+  static readonly socketAddressDescriptor: SettingDescriptor<string> = {
+    id: 'dev.perfetto.RecordTraceV2#socketAddressSetting',
```
stevegolton, line 27: Actually, one thing. We typically try not to tie the id of a setting to a particular plugin, which gives us the flexibility to move them around without folks losing their settings. It's likely we'll move this setting in the near future so please could you call it `dev.perfetto.TracedConsumerSocketAddress` or similar.

stevegolton (changes requested):

ui/src/plugins/dev.perfetto.RecordTraceV2/settings.ts
abhmen01: Changed to dev.perfetto.TracedConsumerSocketAddress
Thanks

stevegolton (approved):

primiano (approved):

PR #2040, opened 2025-07-04
Remove layer_id check for frames and CUJs.

devianb:
Earlier implementation matched the layer_id for a frame with that of a CUJ, to determine which frame should be attributed to which CUJ. The per-frame blocking calls metric used this logic eventually. But this causes blocking calls to be missed in certain cases, where there are frames rendered during a CUJ, but for a different layer. Also refactored the code to remove duplicate frames_in_cuj table.

Bug: 429073039
Test: tools/diff_test_trace_processor.py --name-filter ".*android_blocking_calls_cuj_per_frame.*" out/linux_clang_debug/trace_processor_shell


LalitMaganti (approved):

src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
```diff
@@ -54,43 +54,6 @@ ON bc.utid = frame.ui_thread_utid
...
-    MAX(frame.ts, cuj.ts) AS frame_ts,
-    MIN(
-        frame.dur,
-        cuj.ts_end - frame.ts,
-        frame.ts_end - cuj.ts
-    ) AS dur
-FROM android_frames_layers frame
-JOIN android_sysui_jank_cujs cuj
-ON frame.upid = cuj.upid
-   AND frame.layer_id = cuj.layer_id
```
nicomazz, line 77: nit: moving this to the stdlib masks the real change happening in this cl, which is the removal of the layer_id check. I would suggest to do a minimal cl where you only remove the check and update tests in that regard, and then another mechanical cl with no behaviour difference where you just move the table in the stdlib. this would make it easier to debug potential regressions, if they happen

devianb: Done. This CL focuses mainly on layer_id and test changes. Will raise a separate PR for moving the tables.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -187,13 +187,15 @@ WITH
...
+      ON frame.ui_thread_utid = cie.ui_thread AND frame.layer_id IS NOT NULL
     JOIN _sysui_cujs_slices AS cuj
       ON cie.cuj_id = cuj.cuj_id
     -- Check whether the frame_id falls within the begin and end vsync of the cuj.
     -- Also check if the frame start or end timestamp falls within the cuj boundary.
     WHERE
-      (
+      frame_id >= begin_vsync
+      AND frame_id <= end_vsync
+      AND (
```
nicomazz, line 198: the rest of the cl makes sense, but why do we need to add this check as well?

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Accidentally added. Removed now.

PR #2045, opened 2025-07-04
ui: add command to zoom in on selection

LalitMaganti:
Add a new command to zoom in on selection. This is conceptually similar to focusing on events but also works on instant events.

### DIFF
--- a/ui/src/frontend/ui_main.ts
+++ b/ui/src/frontend/ui_main.ts
@@ -223,6 +223,11 @@ export class UiMainPerTrace implements m.ClassComponent {
         defaultHotkey: 'F',
       },
       {
+        id: 'perfetto.ZoomOnInstant',
+        name: 'Zoom in on current selection',
+        callback: () => trace.selection.zoomOnSelection(),
+      },
+      {
         id: 'perfetto.Deselect',
         name: 'Deselect',
         callback: () => {
--- a/ui/src/core/selection_manager.ts
+++ b/ui/src/core/selection_manager.ts
@@ -21,7 +21,7 @@ import {
   TrackEventSelection,
   AreaSelectionTab,
 } from '../public/selection';
-import {TimeSpan} from '../base/time';
+import {Time, TimeSpan} from '../base/time';
 import {raf} from './raf_scheduler';
 import {exists, getOrCreate} from '../base/utils';
 import {TrackManagerImpl} from './track_manager';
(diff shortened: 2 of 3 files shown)


guptaaka:
`Specifically, in the case of instant events, we just pan`

What does pan mean here?

`Unfortunately as we don't have a duration, we don't know "how much" we should zoom.`

Can we zoom in just enough to capture all the neighbouring events?

`a fixed factor of 100x`

What is the factor of 100 relative to? Does this mean we might or might not view all the events that we want to look at?

LalitMaganti:
> What does pan mean here?

Pan = scroll horizontally 

> Can we zoom in just enough to capture all the neighbouring events?

a) it's a bit tricky because of how the code is structured we don't have the information on pixel density in this place to zoom in that much
2) generally zooming so that one pixel is the whole screen is a bit too excessive anyway, even in the case of slices we don't do that. 

> What is the factor of 100 relative to? Does this mean we might or might not view all the events that we want to look at?

It's a 100x the current zoom resolution. Given that screens are ~2000-3000 pixels, you're talking about zooming into 200-300 pixels around the area of interest. I feel that'd a reasonable amount to zoom. 

There's no zoom level you can soon to where you can guarantee to see all events with no overlaps and if you want to zoom more you can always press f again

stevegolton:
I can empathize with the problem and agree that the solution is a pragmatic one, but I fear that the 100x yet another magic number that's likely to confuse. While this solution might work for some, I fear it won't work for everyone and will be difficult to change later if some folks start relying on it.

The problem is there is no such thing as 'real' trace any more. The Perfetto UI has become a generic trace viewer used in a large number of applications, and we have put in a lot of effort to make it less android/chrome specific over the years. An arbitrary zoom scaling factor of 100x might work for some traces but not others.

The other problem is that this behavior is different for slices and instant events now - with a slice you press F once to center it in the window and again to zoom into it, then any subsequent presses have no effect. With instants pressing F will zoom in 100x every time.

If you really want to zoom in a few times on the selected event we could make this a new hotkey for just this, or you could just hit 'W' a few times to zoom in on the cursor position.

Going forward I think we should aim at solving this 'properly' using datasets. In the same way as '<' '>' works by finding the next event in timestamp order, we could work out the nearest events before and after the selected event and work out how much to zoom in so that the individual event is distinguishable amongst it's neighbors at the same depth. This would involve a query, but I think that's fine.

LalitMaganti:
I'm fine with solving this properly but from what I can see that's going to be a non trivial change and I don't want to take you away from other things you're working on. 

Would it better if I made this a command for now without a hokey?

stevegolton:
> Would it better if I made this a command for now without a hokey?

Sure! Just called 'Zoom in 100x on selected instant event' or something (probably more elegant).

LalitMaganti:
Added a new command, PTAL.

ui/src/frontend/ui_main.ts
```diff
@@ -222,6 +222,11 @@ export class UiMainPerTrace implements m.ClassComponent {
         callback: () => trace.selection.scrollToSelection(),
         defaultHotkey: 'F',
       },
+      {
+        id: 'perfetto.ZoomOnInstant',
```
stevegolton: Maybe update the name to reflect that it zooms in on any type of selection.

PR #2059, opened 2025-07-07
tp: ftrace: parse self-describing generic events into the raw table

rsavitski:
The latest scheme for serialising generic (i.e. unknown by perfetto at
compile-time) ftrace events:
* FtraceEvent proto has a range of field ids reserved for
  dynamically-generated extensions.
* FtraceEventBundle proto has a mapping from the field id to a serialised
  proto descriptor for that event (derived from the tracefs /format file).

This patch adds a new tracker to serve as a way for the ftrace_tokenizer
to inform the ftrace_parser of the descriptor map.


src/trace_processor/importers/ftrace/generic_ftrace_tracker.h
```diff
@@ -0,0 +1,98 @@
...
+namespace perfetto::trace_processor {
+
+// The latest scheme for serialising generic (i.e. unknown by perfetto at
+// compile-time) ftrace events:
+// * FtraceEvent proto has a range of field ids reserved for
+//   dynamically-generated extensions.
+// * FtraceEventBundle proto has a mapping from the field id to a serialised
+//   proto descriptor for that event (derived from the tracefs /format file).
+//
+// This tracker is used to carry state from the ftrace tokenizer to the parser:
```
LalitMaganti, line 40: Are we guaranteed that descriptors for the same message will not change during the lifeitme of the tracE? That's an implicit dependency, I would document that here.

LalitMaganti (approved):

src/trace_processor/importers/ftrace/generic_ftrace_tracker.h
rsavitski: Clarified that the mapping is constant for a given trace.

LalitMaganti: Actually how will this work when we are merging data from two different traces?

Is it possible that this tracker lives in the ftrace module rather than in the context?

rsavitski: PTAL, I assume this is what you meant (module member + pointers). Note: I am not clearing the tracker's storage anymore, but the mapping is at most O(distinct ftrace event types) per trace anyway.

LalitMaganti: Looks good thanks!

PR #2060, opened 2025-07-07
[API] Avoid duplicating slow tag

etiennep-chromium:
This CL avoids adding slow tag on "disabled-by-default-*" on a category if already present.


ddiproietto (approved):

etiennep-chromium:
I added a check for kDebugTag as well, PTAnL?

PR #2062, opened 2025-07-07
Disallow swapper thread from being used in tid_override

jahdiel-alvarez:
No writer should be writing events on behalf of swapper thread, via the legacy track event system.

Bug: 425694654


src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -392,6 +392,11 @@ class TrackEventEventImporter {
         if (legacy_event_.has_tid_override())
```
LalitMaganti, line 392: We should do the same thing for pid override.

jahdiel-alvarez: @LalitMaganti  The thing with PID is that it is not necessarily 0 in a multi-machine trace.

jahdiel-alvarez: I meant the PID is not necessarily 0 for swapper in multi-machine traces. But I think it is true that a PID 0 will always be assigned to swapper thread. So I think we can make the change.

LalitMaganti (approved):

PR #2065, opened 2025-07-08
WIP Priority boost implementation plus integration tests.

KirillTim:


include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,126 @@
...
+
+#ifndef INCLUDE_PERFETTO_EXT_BASE_SCOPED_SCHED_BOOST_H_
+#define INCLUDE_PERFETTO_EXT_BASE_SCOPED_SCHED_BOOST_H_
+
+#include "perfetto/ext/base/status_or.h"
+#include "perfetto/ext/base/thread_checker.h"
+
+namespace perfetto::base {
+
+struct SchedOsConfig {
```
primiano: why isn't this just an implementation detail?
please don't expose it as it makes things confusing. It seems sorta identical to  SchedPolicyAndPrio

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,126 @@
...
+  bool operator==(const SchedPolicyAndPrio& other) const {
+    return policy == other.policy && prio == other.prio;
+  }
+
+  bool operator!=(const SchedPolicyAndPrio& other) const {
+    return !(*this == other);
+  }
+
+  Policy policy = Policy::kSchedOther;
+  unsigned int prio = 0;
```
primiano: nit: if you need unsigned, use uint32_t. 
unsigned int is longer to type/read, confusing and ambiguous.

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,126 @@
...
+};
+
+class SchedOsManager {
```
primiano: why this exists? why is it exposed?

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,126 @@
...
+ protected:
+  SchedOsManager() = default;
+
+  SchedOsManager(const SchedOsManager&) = delete;
+  SchedOsManager& operator=(const SchedOsManager&) = delete;
+  SchedOsManager(SchedOsManager&&) = delete;
+  SchedOsManager& operator=(SchedOsManager&&) = delete;
+};
+
+class ThreadMgr {
```
primiano: same quesiton. why this is exists? why isn't an implementation detail in the anonymous namespace?

protos/perfetto/config/priority_boost/BUILD.gn
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2025 The Android Open Source Project
```
primiano, line 1: I don't see the reason of creating a dedicated subdirectory to just host one single .proto file.
please move this one directory up.
You can still have a dedicated target if you really need it

protos/perfetto/config/priority_boost/priority_boost_config.proto
```diff
@@ -0,0 +1,41 @@
...
+  required uint32 priority = 2;
+
+  //  oneof boost_level {
+  //    // From [-20 to 19], larger values correspond to _lower_ scheduling
+  //    // priority. Userspace policy SCHED_OTHER will be used.
+  //    int32 nice_value = 1;
+  //    // From [1 to 99], lager value correspond to _higher_ priority.
+  //    // Realtime policy SCHED_FIFO will be used.
+  //    uint32 realtime_priority = 2;
+  //  }
```
primiano: remove these an dupdate the comment son the above?

src/base/scoped_sched_boost.cc
```diff
@@ -0,0 +1,223 @@
...
+
+#include "perfetto/ext/base/no_destructor.h"
+#include "perfetto/ext/base/status_macros.h"
+
+namespace perfetto::base {
+
+namespace {
+constexpr pid_t kCurrentPid = 0;
+
+SchedOsConfig to_sched_os_config(const SchedPolicyAndPrio& spp) {
```
primiano: why why why 5 lines function called only in one place.
I mentioend this few days ago, i undid it, you re-did it again.
why
please. this stuff hurts readability.

test/priority_boost_integrationtest.cc
```diff
@@ -0,0 +1,190 @@
...
+  ftrace_config.add_ftrace_events("sched_switch");
+  ds_config->set_ftrace_config_raw(ftrace_config.SerializeAsString());
+
+  helper.StartTracing(trace_config);
+  helper.WaitForDataSourceConnected("linux.ftrace");
+
+#if PERFETTO_BUILDFLAG(PERFETTO_ANDROID_BUILD)
+  PERFETTO_DLOG("PERFETTO_ANDROID_BUILD");
+  traced_tid = PidForProcessName("/system/bin/traced");
+  traced_probes_tid = PidForProcessName("/system/bin/traced_probes");
```
primiano: rather than hardcoding the paths could we get them from QueryServiceState? I think that works for both cases

primiano (changes requested):
You reintroduced again a whole lotta complexity, which is what I tried to avoid, even showing you the code in dev/primiano/prioboost
I am not having this. please rewrite this CL without having all this complexity.
We are not wiritng a generalized library for os scheduling. we are writing a tiny class for boosting the priority.
Your test doesn't warrant all those classes. You are making the code more bug prone and harder to read for a test.
If you want to test without having CI capabilities, then add a tiny struct, which has IDENTICAL signatures to sched_getscheduler, so you can mock them in the test.

protos/perfetto/config/priority_boost/BUILD.gn
KirillTim: This is a common pattern in the directory, e.g: 
1. https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/protos/perfetto/config/interceptors/
2. https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/protos/perfetto/config/power/
3. https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/protos/perfetto/config/process_stats/
4. https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/protos/perfetto/config/sys_stats/

Won't it be more consistent to have the dedicated subdirectory?

include/perfetto/ext/base/scoped_sched_boost.h
KirillTim: My bad, the class is now moved to the `scoped_sched_boost.cc` file.

src/base/scoped_sched_boost.cc
KirillTim: Sorry, I've inlined this function now.

protos/perfetto/config/priority_boost/priority_boost_config.proto
KirillTim: Fixed.

include/perfetto/ext/base/scoped_sched_boost.h
KirillTim: This class is used in tests, see `test/priority_boost_integrationtest.cc`

include/perfetto/ext/base/scoped_sched_boost.h
KirillTim: Fixed.

test/priority_boost_integrationtest.cc
KirillTim: We are already using this pattern in android tests, so I'd rather keep it as it is.
Plus, I can't find a way to get a pid/tid from the `QueryServiceState`.

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,104 @@
...
+
+  bool operator!=(const SchedPolicyAndPrio& other) const {
+    return !(*this == other);
+  }
+
+  Policy policy = Policy::kSchedOther;
+  uint32_t prio = 0;
+};
+
+class SchedOsManager {
```
primiano: please add comments explaining that this exist just for being hooked in tests.
Also I'd call this really SchedHooks. This is stateless and doesn't have any state.

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,104 @@
...
+ public:
+  struct SchedOsConfig {
+    int policy;
```
primiano, line 60: do these need to be zero initialized?

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,104 @@
...
+    int rt_prio;
+    int nice;
+  };
+
+  static SchedOsManager* GetInstance();
```
primiano: why this needs to be a singleton?

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,104 @@
...
+  // virtual for testing
+  virtual Status SetSchedConfig(const SchedOsConfig& arg);
+  // virtual for testing
+  virtual StatusOr<SchedOsConfig> GetCurrentSchedConfig() const;
+  virtual ~SchedOsManager() = default;
```
primiano: nit: order, move this to top (ctor and dtors go first)

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,104 @@
...
+
+ protected:
+  SchedOsManager() = default;
```
primiano: Why bother? let people create new instances no? there is no stringent need for this to be a singleton

include/perfetto/ext/base/scoped_sched_boost.h
```diff
@@ -0,0 +1,104 @@
...
+
+  SchedOsManager(const SchedOsManager&) = delete;
+  SchedOsManager& operator=(const SchedOsManager&) = delete;
+  SchedOsManager(SchedOsManager&&) = delete;
```
primiano: why do you bother with deleting copy and assign? it doens't matter, even if you do it doesn't change things.
I only do this when copy/move  can cause some odd behaviour.
In this case you really don't care, this is just visual clutter.

protos/perfetto/config/priority_boost/priority_boost_config.proto
```diff
@@ -0,0 +1,38 @@
...
+// Only supported on Linux and Android the boosted process must have
+// 'CAP_SYS_NICE' capability.
+message PriorityBoostConfig {
+  enum BoostPolicy {
+    // The default policy (e.g., CFS on Linux). Priority range: [0; 20]
+    // priority is interpreted as -(nice), i.e., 1 is slightly higher prio
+    // than default 0, 20 is the highest priority.
+    // Note: this is the opposite semantic of the cmdline nice, and is done for
+    // consistency with POLICY_SCHED_FIFO, so higher number == higher prio.
+    POLICY_SCHED_OTHER = 0;
```
primiano: the =0 value must be POLICY_UNSPECIFIED, see https://protobuf.dev/best-practices/dos-donts/#unspecified-enum

If you don't do it, google3 imports will fail.

This boils down to "you souldn't have semantic difference between 0 and unset"

src/base/scoped_sched_boost.cc
```diff
@@ -0,0 +1,231 @@
...
+
+#include "perfetto/ext/base/no_destructor.h"
+#include "perfetto/ext/base/status_macros.h"
+
+namespace perfetto::base {
+
+namespace {
+constexpr pid_t kCurrentPid = 0;
+
+class ThreadMgr {
```
primiano, line 40: add a comment + the TODO I had was very important: to explain that if we want to support > 1 thread, this needs to be changed.

src/base/scoped_sched_boost.cc
```diff
@@ -0,0 +1,231 @@
...
+  void ResetForTesting(SchedOsManager*);
+
+ private:
+  SchedOsManager* os_manager_;
+  SchedOsManager::SchedOsConfig initial_config_{};
+  std::vector<SchedPolicyAndPrio> prios_;
+};
+
+ThreadMgr& ThreadMgr::GetInstance() {
+  static NoDestructor<ThreadMgr> instance(SchedOsManager::GetInstance());
```
primiano: IMHO NoDestructor should be used only in global variables. I'd be consistent with the rest.

src/base/scoped_sched_boost.cc
```diff
@@ -0,0 +1,231 @@
...
+  }
+}
+
+Status ThreadMgr::Add(SchedPolicyAndPrio spp) {
+  prios_.push_back(spp);
+  return RecalcAndUpdatePrio();
+}
+
+void ThreadMgr::Remove(SchedPolicyAndPrio spp) {
+  prios_.erase(std::remove(prios_.begin(), prios_.end(), spp), prios_.end());
```
primiano: this is a bug, this will remove all the instances of prios_, you want to remove only one

src/base/scoped_sched_boost.cc
```diff
@@ -0,0 +1,231 @@
...
+  // in 'RecalcAndUpdatePrio'.
+  // This for loop makes us ignore all wrongly configured policies and fallback
+  // to the first correct (or the initial one).
+  for (;;) {
+    if (auto res = RecalcAndUpdatePrio(); !res.ok()) {
+      PERFETTO_ELOG("%s", res.c_message());
+    } else {
+      break;
+    }
+  }
```
primiano, line 105: just do this loop in the RecalcAndUpdatePrio

src/base/scoped_sched_boost.cc
```diff
@@ -0,0 +1,231 @@
...
+
+}  // namespace
+
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX) || \
+    PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+
+SchedOsManager* SchedOsManager::GetInstance() {
+  static auto* instance = new SchedOsManager();
+  return instance;
+}
```
primiano: no need for this

src/base/scoped_sched_boost.cc
```diff
@@ -0,0 +1,231 @@
...
+    return ErrStatus("sched_getscheduler failed (errno: %d, %s)", errno,
+                     strerror(errno));
+  }
+  sched_param param{};
+  if (sched_getparam(kCurrentPid, &param) == -1) {
+    return ErrStatus("sched_getparam failed (errno: %d, %s)", errno,
+                     strerror(errno));
+  }
+  int nice = 0;
+  if (param.sched_priority == 0) {
```
primiano, line 183: add a comment as this is a bit subtle. explain that SCHED_OTHER/BATCH make use of the niceness vaue

src/tracing/service/tracing_service_impl.cc
```diff
@@ -934,6 +935,23 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
     }
   }
 
+  std::optional<base::ScopedSchedBoost> priority_boost;
+  if (cfg.has_priority_boost()) {
+    auto sched_policy = CreateSchedPolicyFromConfig(cfg.priority_boost());
+    if (!sched_policy.ok()) {
+      // TODO(ktimofeev): call MaybeLogUploadEvent
```
primiano: you landed the atom right? can you just add this?

protos/perfetto/config/priority_boost/priority_boost_config.proto
KirillTim: Done.

src/base/scoped_sched_boost.cc
KirillTim: Done. I've also updated the `ScopedSchedBoostTest#ScopeEnterExit` to catch this error.

include/perfetto/ext/base/scoped_sched_boost.h
KirillTim: Done.

src/base/scoped_sched_boost.cc
KirillTim: Done.

include/perfetto/ext/base/scoped_sched_boost.h
KirillTim: Not really, we always init all the fields when create an instance of this struct in tests.

PR #2088, opened 2025-07-09
Move duplicate frames_in_cuj table to stdlib.

devianb:


src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -129,6 +129,50 @@ WHERE
 GROUP BY
   cuj_id;
 
+-- Track all frames that overlap with the CUJ slice.
+CREATE PERFETTO VIEW _android_frames_in_cuj AS
```
LalitMaganti, line 133: Fine for now but long term, this sort of composition should happen inside the metric rather than in SQL.

devianb: Ack.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -129,6 +129,50 @@ WHERE
...
+FROM android_frames_layers AS frame
+JOIN _sysui_cuj_instant_events AS cie
+  ON frame.ui_thread_utid = cie.ui_thread AND frame.layer_id IS NOT NULL
+JOIN _sysui_cujs_slices AS cuj
+  ON cie.cuj_id = cuj.cuj_id
+-- Check whether the frame_id falls within the begin and end vsync of the cuj.
+-- Also check if the frame start or end timestamp falls within the cuj boundary.
+WHERE
+  frame_id >= begin_vsync
+  AND frame_id <= end_vsync
```
nicomazz, line 160: just for my undestanding: I assume this cl is on top of the other, so you'll have to merge the base first, then rebase this and merge it? anyway, sgtm

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: I added this check here because the tests need to be modified slightly otherwise. Here in the block from line 161 to 173, we are checking the intersection of frame with the CUJ slice. The frames_in_cuj logic that was earlier defined in the per-frame metrics SQL was a little different, as it was already using the android_sysui_jank_cujs table.

PR #2089, opened 2025-07-09
Extend frame boundary for per-frame metrics.

devianb:
In some cases when a blocking call crosses a frame boundary, the next actual frame does not start till the blocking call ends. With the current metric logic, the part of the blocking call outside the frame boundary is not being considered for metric calculation, which is incorrect.
Fixed the frame boundary logic to not consider the actual frame boundary, but consider the start of the current frame to the start of the next frame as a frame boundary.


src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
```diff
@@ -46,51 +59,14 @@ SELECT
...
 ON bc.utid = frame.ui_thread_utid
    -- The following condition to accommodate blocking call crossing frame boundary. The blocking
    -- call starts in a frame and ends in a frame. It can either be the same frame or a different
    -- frame.
 WHERE (bc.ts >= frame.ts AND bc.ts <= frame.ts_end) -- Blocking call starts within the frame.
     OR (bc.ts_end >= frame.ts AND bc.ts_end <= frame.ts_end); -- Blocking call ends within the frame.
 
--- Table capturing the full and partial frames within a CUJ boundary. This table captures the
--- layer ID associated with the actual frame too.
-DROP TABLE IF EXISTS frames_in_cuj;
```
jernej-google: This table won't exist anymore in the future - since it's replaced with `_android_frames_in_cuj`.

Is there a reason for rename and deprecation of old table? Is it going to break users?

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -129,6 +129,50 @@ WHERE
 GROUP BY
   cuj_id;
 
+-- Track all frames that overlap with the CUJ slice.
+CREATE PERFETTO VIEW _android_frames_in_cuj AS
+SELECT
+  row_number() OVER (PARTITION BY cuj.cuj_id ORDER BY frame.ts) AS frame_idx,
+  count(*) OVER (PARTITION BY cuj.cuj_id) AS frame_cnt,
+  substr(cuj.cuj_slice_name, 3, length(cuj.cuj_slice_name) - 3) AS cuj_name,
```
jernej-google: What are those 3 chars we're cutting away on both ends?

jernej-google (approved):
LGTM, not quite sure why we're renaming the frames_in_cuj table.

src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
LalitMaganti: Any code in the /metrics folder is *not* an API and anyone depending on it was doing so at their own risk.

devianb: The frames_in_cuj table was added  recently by me when creating the per-frame metrics, so it is not being used by anywhere else. Plus removed it keep it in a centralized place(stdlib), from where multiple metrics can use it.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: In SQL the index starts from 1, so we are picking the string from the 3rd character. So 'J<' is being at the beginning. Nothing is being cut from the end.

jernej-google: Based on the table below, shouldn't `cuj.cuj_name` have the name without the prefix?

devianb: Actually the below table android_sysui_jank_cujs is based on this table, not the other way around. I can make the change there.

src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
```diff
@@ -32,6 +32,19 @@ INCLUDE PERFETTO MODULE android.cujs.sysui_cujs;
...
+-- For cases when a blocking call starts within a frame, but does not end before the actual frame
+-- ends, a part of the blocking call can be missed while calculating the metric. To avoid this
+-- issue, the frame boundary is extended, spanning from the start of the current actual frame, till
+-- the start of the next actual frame.
+WITH extended_frame_boundary AS (
+SELECT frame_ts as ts,
+ui_thread_utid,
+frame_id,
+layer_id,
+  COALESCE(LEAD(frame_ts) OVER (PARTITION BY cuj_id ORDER BY frame_id ASC), ts_end) AS ts_end,
```
nicomazz, line 46: let's add a comment to make this easier to understand for the next person reading it. Something like:

```
-- Calculate the end timestamp (ts_end) by taking the start time (frame_ts) of the next frame in the session.
-- For the last frame, fall back to the default ts_end.
```

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
nicomazz: could you extract this logic to a function here, so it's a bit easier to parse?

src/trace_processor/metrics/sql/android/android_blocking_calls_cuj_per_frame_metric.sql
devianb: Done.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Done.

PR #2093, opened 2025-07-09
ui: Card QOL improvements

aMayzner:
Stack cards can be stacked either vertically or horizontally.
Stack cards can be actionable - have reactions on hover.

### DIFF
--- a/ui/src/assets/widgets/card.scss
+++ b/ui/src/assets/widgets/card.scss
@@ -22,6 +22,14 @@ $box-shadow: 0px 2px 6px rgba(0, 0, 0, 0.15);
   box-shadow: $box-shadow;
   border-radius: $border-radius;
   border: solid 1px $pf-colour-thin-border;
+
+  &.pf-actionable {
+    transition: all 200ms ease-in-out;
+    &:hover {
+      transform: translateY(-2px);
+      box-shadow: 0px 4px 12px rgba(0, 0, 0, 0.2);
+    }
+  }
 }
 
 .pf-card-stack {
@@ -40,5 +48,14 @@ $box-shadow: 0px 2px 6px rgba(0, 0, 0, 0.15);
     border-radius: unset;
     border: unset;
     box-shadow: unset;
+  }
+
+  &.pf-card-stack-horizontal {
+    flex-direction: row;
+
+    > .pf-card:not(:last-child) {
+      border-right: 1px solid lightgray;
+      border-bottom: unset;
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


ui/src/plugins/dev.perfetto.WidgetsPage/widgets_page.ts
```diff
@@ -866,13 +866,16 @@ export class WidgetsPage implements m.ClassComponent<{app: App}> {
           multiple Card elements in a vertical stack. Cards placed in this list
           automatically have their borders adjusted to appear as one continuous
           card with thin borders between them.`,
-        renderWidget: () =>
-          m(CardStack, [
-            m(Card, m(Switch, {label: 'Option 1'})),
-            m(Card, m(Switch, {label: 'Option 2'})),
-            m(Card, m(Switch, {label: 'Option 3'})),
+        renderWidget: ({direction, actionable}) =>
```
stevegolton: Can you add an 'actionable' flag to the single card example above?

ui/src/assets/widgets/card.scss
```diff
@@ -22,6 +22,14 @@ $box-shadow: 0px 2px 6px rgba(0, 0, 0, 0.15);
   box-shadow: $box-shadow;
   border-radius: $border-radius;
   border: solid 1px $pf-colour-thin-border;
+
+  &.pf-actionable {
```
stevegolton: I'm not convinced actionable is the best term here, though I could be talked into it. Maybe 'interactive'

ui/src/assets/widgets/card.scss
```diff
@@ -41,4 +49,13 @@ $box-shadow: 0px 2px 6px rgba(0, 0, 0, 0.15);
     border: unset;
     box-shadow: unset;
   }
+
+  &.pf-card-stack-horizontal {
```
stevegolton: BEM name should be `.pf-card-stack--horizontal` and you should take advantage of scss and target it directly with `&--horizontal {...}` here.

ui/src/assets/widgets/card.scss
```diff
@@ -22,6 +22,14 @@ $box-shadow: 0px 2px 6px rgba(0, 0, 0, 0.15);
   box-shadow: $box-shadow;
   border-radius: $border-radius;
   border: solid 1px $pf-colour-thin-border;
+
+  &.pf-actionable {
+    transition: all 200ms ease-in-out;
+    &:hover {
+      transform: translateY(-2px);
```
stevegolton: I'm not convinced about the translate here. Maybe just use the drop shadow.

ui/src/assets/widgets/card.scss
```diff
@@ -41,4 +49,13 @@ $box-shadow: 0px 2px 6px rgba(0, 0, 0, 0.15);
     border: unset;
     box-shadow: unset;
   }
```
stevegolton, line 50: In a stack maybe avoid adding the drop shadow and just tint the background color grey? I think the drop shadow effect looks a little odd when in the stack.

stevegolton (changes requested):
Looks good! Nice idea allowing the stack to be horizontal too!

stevegolton (approved):
Nice!

PR #2109, opened 2025-07-10
ui: Move sqlmodules load out of the traceload hotpath

stevegolton:
This patch moves the SQL modules definition loading to onActivate, so that it's done once on app load rather than having to do it on every trace load. This means we only load the definitions once, and also we also get a head start loading the modules, as we can load in parallel with the rest of the app starting up / trace loading.

This does mean that the definitions are possibly undefined, but everywhere that uses the definitions has now been modified to handle this maybe undefined case.


aMayzner (approved):

LalitMaganti:
Steve: can we land this?

PR #2113, opened 2025-07-10
Modify corner radius field in the surfaceflinger protos

ddiproietto:
Port of internal commit.

**Original commit message:**
Subject: Modify corner radius field in the surfaceflinger protos

Replaces cornerRadius float with cornerRadii and adds
clientDrawnCornerRadii

Bug: 406554726

Test: presubmit; manual

Change-Id: I5b8112de46492b2b17d31706609d0a9275d16a41

---
Original internal commit hash: c62ae29f9fd31dba01efd47d88e10b022393c6a5
Cherry-picked with `-x`.
This PR is part of a stack. Base branch: main


PriyankaSPX (approved):

LalitMaganti (approved):

PriyankaSPX:
i'll open a PR to modify the failing diff test so this PR can be merged after - we really don't need to be testing so many fields and the test shouldn't need changing after every proto change

ddiproietto:
> i'll open a PR to modify the failing diff test so this PR can be merged after - we really don't need to be testing so many fields and the test shouldn't need changing after every proto change

Thank you! I've pushed an updated version of this in the meantime

PR #2128, opened 2025-07-10
UI: add file picker to the Download link in the sidebar

primiano:
Also fix a bug in postmessage_handler, where we ignore 
localOnly: false, giving no option to the embedder to allow downloads.

Bug: b/430934358


stevegolton (approved):

primiano:
> Are we OK with using experimental features?


As long as we have a fallback I think it's okay.

PR #2137, opened 2025-07-11
Improved track organisation in unfold latency plugin

michalbrz:
Instead of pinning all tracks relevant to unfold, we pin only most important ones and rest is grouped below.
I wish we could pin groups as well but currently it's not possible.
This change leaves only 4 tracks pinned and rest is grouped into: Unfold animation, Unfold display switching and Unfold misc.

A few small extra changes:
1. Adding tracks for android.display thread
2. Adding track for PhotonicModulator which is doing the actual display switching. This requires querying content to find track with 'setDisplaySwitching' slice
3. Renaming plugin from 'Pin: Unfold latency tracks' to 'Organize unfold latency tracks'

### DIFF
--- a/ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
+++ b/ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
@@ -13,32 +13,127 @@
 // limitations under the License.
 
 import {Trace} from '../../public/trace';
+import {TrackNode} from '../../public/workspace';
 import {PerfettoPlugin} from '../../public/plugin';
+import StandardGroupsPlugin from '../dev.perfetto.StandardGroups';
+import TraceProcessorTrackPlugin from '../dev.perfetto.TraceProcessorTrack';
+import {NUM} from '../../trace_processor/query_result';
 
 export default class implements PerfettoPlugin {
   static readonly id = 'dev.perfetto.LargeScreensPerf';
+  static readonly dependencies = [
+    StandardGroupsPlugin,
+    TraceProcessorTrackPlugin,
+  ];
+
   async onTraceLoad(ctx: Trace): Promise<void> {
+    this.addFoldedStateTrackToDeviceState(ctx);
+
     ctx.commands.registerCommand({
-      id: 'dev.perfetto.LargeScreensPerf#PinUnfoldLatencyTracks',
-      name: 'Pin: Unfold latency tracks',
-      callback: () => {
-        ctx.workspace.flatTracks.forEach((track) => {
-          if (
-            !!track.name.includes('UnfoldTransition') ||
-            track.name.includes('Screen on blocked') ||
-            track.name.includes('hingeAngle') ||
-            track.name.includes('UnfoldLightRevealOverlayAnimation') ||
-            track.name.startsWith('waitForAllWindowsDrawn') ||
-            track.name.endsWith('UNFOLD_ANIM>') ||
-            track.name.endsWith('UNFOLD>') ||
-            track.name == 'Waiting for KeyguardDrawnCallback#onDrawn' ||
-            track.name == 'FoldedState' ||
-            track.name == 'FoldUpdate'
-          ) {
-            track.pin();
-          }
-        });
+      id: 'dev.perfetto.LargeScreensPerf#UnfoldLatencyTracks',
+      name: 'Organize unfold latency tracks',
+      callback: async () => {
+        this.pinCoreTracks(ctx);
+        this.addUnfoldMiscSection(ctx);
+        await this.addUnfoldDisplaySwitchingSection(ctx);
+        this.addUnfoldAnimationSection(ctx);
       },
     });
+  }
+
+  private addFoldedStateTrackToDeviceState(ctx: Trace) {
+    const foldedStateTrack = ctx.workspace.flatTracks.find(
+      (t) => t.name == 'FoldedState',
+    );
+    if (foldedStateTrack) {
+      const deviceStateGroup = ctx.plugins
+        .getPlugin(StandardGroupsPlugin)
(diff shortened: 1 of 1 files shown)


michalbrz:
@nicomazz @lukaville plz review 🙏

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
```diff
@@ -13,32 +13,101 @@
...
+        this.pinCoreTracks(ctx);
+        this.addUnfoldMiscSection(ctx);
+        await this.addUnfoldDisplaySwitchingSection(ctx);
+        this.addUnfoldAnimationSection(ctx);
       },
     });
   }
+
+  private pinCoreTracks(ctx: Trace) {
+    const titlesToPin = new Set(['FoldedState', 'FoldUpdate']);
```
nicomazz: I think it would be great to pin the folded state even before this plugin is manually triggered, literally all the times, similarly to the "screen on" or other tracks before the thread tracks

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
```diff
@@ -13,32 +13,101 @@
...
+          titlesToPin.has(track.name) ||
+          track.name.includes('hingeAngle') ||
+          track.name.endsWith('UNFOLD>'),
+      )
+      .forEach((track) => track.pin());
+  }
+
+  private addUnfoldMiscSection(ctx: Trace) {
+    // section for tracks that don't fit neatly in other sections and are not so important to be pinned
+    const group = new TrackNode({name: 'Unfold misc'});
```
nicomazz, line 69: can we have all of these as children of a top level unfold group?

Actually.. I wonder if we should make this always enabled: if this only adds an "Folded state" top level track group (that by default in its track shows the folded state e.g. folded/unfolded), and then as it's expanded it shows the other groups, it would be quite useful and probably not an issue to have it always enabled

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
```diff
@@ -13,32 +13,101 @@
...
+    displayTracks
+      // sorting so that "android.display" tracks are next to each other
+      .sort((t1, t2) => t1.name.localeCompare(t2.name))
+      .forEach((t) => group.addChildFirst(t));
+  }
+
+  private async findPhotonicModulatorTrack(
+    ctx: Trace,
+  ): Promise<TrackNode | undefined> {
+    const query = `
```
nicomazz: very nit: rename to say what you're looking for with this query

nicomazz (approved):

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
michalbrz: As discussed offline we'll just add FoldedState track to DeviceState group by default and rest of the tracks will be pinned only after triggering the plugin by the user

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
michalbrz: Explained in https://github.com/google/perfetto/pull/2137#discussion_r2208046878

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
michalbrz: Done, now it's very descriptive

michalbrz:
Default behaviour of adding Folded State
image

After plugin is triggered:
image

After plugin is triggered and groups are expanded:
image

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
```diff
@@ -13,32 +13,124 @@
 // limitations under the License.
 
 import {Trace} from '../../public/trace';
+import {TrackNode} from '../../public/workspace';
 import {PerfettoPlugin} from '../../public/plugin';
+import DeviceStateGroupPlugin from '../dev.perfetto.AndroidLongBatteryTracing';
```
stevegolton: I don't think we should depend on this plugin just for the group.

Can you pull this group out into dev.perfetto.StandardGroups please, and have both the AndroidLongBatteryTracing plugin and this one depend on StandardGroups instead?

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
```diff
@@ -13,32 +13,124 @@
...
+        this.addUnfoldMiscSection(ctx);
+        await this.addUnfoldDisplaySwitchingSection(ctx);
+        this.addUnfoldAnimationSection(ctx);
       },
     });
   }
+
+  private addFoldedStateTrackToDeviceState(ctx: Trace) {
+    const foldedStateTrack = ctx.workspace.flatTracks.find(
+      (t) => t.name == 'FoldedState',
```
stevegolton, line 46: You should probably also depend on `dev.perfetto.TraceProcessorTrack`, which is the plugin that adds this track in the first place, so that we can be sure this track has been given the opportunity to appear.

Or you could create a branch new track based on the track ids in the trace, and avoid having to depend on the other plugin.

Same goes for all of the tracks below.

michalbrz: Added dependency on TraceProcessorTrack 👍

ui/src/plugins/dev.perfetto.LargeScreensPerf/index.ts
michalbrz: Done in https://github.com/google/perfetto/pull/2195

LalitMaganti (approved):

Issue #2140, opened 2025-07-11
Python API: `trace_processor_shell.exe` is not terminated on Windows and becomes an orphan process


seigtm:
## Environment
*   **Operating System:** Windows 11
*   **Python Version:** 3.13.5
*   **Perfetto Python Library Version (pip):** 0.13.1

## Description

When instantiating `perfetto.trace_processor.TraceProcessor` on Windows without providing an explicit `addr`, the library correctly downloads and launches its own `trace_processor_shell.exe` process (daemon). However, the lifecycle of this process is not managed by the Python API itself, so it is not terminated when the Python script finishes.

The only reliable workaround is to manage the process lifecycle manually by starting `trace_processor_shell.exe` in a separate terminal, connecting to it from Python using `TraceProcessor(addr="127.0.0.1:<PORT>", trace=...)` and manually stopping the process after.

**Is it the intended behavior**? **Is there any workaround for it possibly**..?

Thank you in advance.

LalitMaganti:
Thanks for reporting. This is a legit issue and one I was actually aware of (was reported internally) but I was not able to come up with a satisfactory fix. Much of that is because of the difference between how Windows and Linux handle subprocess trees (in Linux, killing a parent also kills recursively all descendents, not so from what I understand on Windows).

> Is it the intended behavior?

It's definitely not the intended behavior. On Mac and Linux, we correctly kill the trace processor subprocess when the TraceProcessor Python object is closed/context manager cleanup/deelted.

>  Is there any workaround for it possibly..?

Ideally we come up with a fix for it. A lot of the complexity comes from the fact that I don't have easy access to a Windows based system (only Mac and Linux systems). 

The key places in the codebase are:
1) The point where trace processor script is killed: https://github.com/google/perfetto/blob/main/python/perfetto/trace_processor/api.py#L306
2) How trace processor exe is invoked in the subprocess: https://github.com/google/perfetto/blob/main/python/perfetto/prebuilts/perfetto_prebuilts.py#L125

I investigated process groups (https://stackoverflow.com/questions/47016723/windows-equivalent-for-spawning-and-killing-separate-process-group-in-python-3) but didn't get far enough to come up with a solution.

Would be very happy if you (or someone reading this bug) proposes how we should fix this. Even better if you can send a PR to do the fix.

seigtm:
@LalitMaganti could you take a look, please?

https://github.com/google/perfetto/pull/2141

LalitMaganti:
[...]

PR #2141, opened 2025-07-11
fix(python): Ensure trace_processor_shell is terminated on Windows

seigtm:
Related issue is: https://github.com/google/perfetto/issues/2140

On Windows, the `trace_processor_shell.exe` subprocess launched by the Python API was not terminated when the `TraceProcessor` object was closed. This resulted in an orphan process that continued to run indefinitely, consuming resources until manually killed.

This was caused by two issues specific to the Windows platform:

1. The `trace_processor_shell` daemonizes itself, creating a new worker process and detaching from the parent. The original PID held by the Python script belonged to the short-lived launcher, not the actual worker.
2. Unlike on POSIX systems, terminating a parent process on Windows does not automatically terminate its children.

This change ensures that the lifecycle of `trace_processor_shell` is correctly managed on all supported platforms.

### DIFF
--- a/python/perfetto/trace_processor/api.py
+++ b/python/perfetto/trace_processor/api.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+import sys
+import signal
 import dataclasses as dc
 from urllib.parse import urlparse
 from typing import List, Optional, Union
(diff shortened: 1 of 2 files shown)


LalitMaganti (approved):

python/perfetto/trace_processor/api.py
```diff
@@ -302,9 +305,16 @@ def __exit__(self, a, b, c):
     return False
 
   def close(self):
-    if hasattr(self, 'subprocess'):
-      self.subprocess.kill()
+    if hasattr(self, 'subprocess') and self.subprocess:
+      # On Windows, we need to kill the whole process group.
+      # On other platforms, killing the parent is enough.
+      if sys.platform == 'win32':
+        os.kill(self.subprocess.pid, signal.CTRL_BREAK_EVENT)
```
LalitMaganti: Instead of using os.kill can we just do self.subprocess.send_signal instead?

seigtm: That's a great point!
Using `send_signal` is indeed more consistent and idiomatic.
I've updated the code

LalitMaganti: Great thanks. LGTM.

LalitMaganti (approved):

PR #2145, opened 2025-07-13
tp: ui: implement tp/ui support for track event merge control

LalitMaganti:
Implements the actual work necessary to correctly parse the protos and
use them to control merge behaviour.

As a bonus, we implement support for minimzing the number of tracks in
the trace processor meaning large traces should be much better
supported.

CHROME_STDLIB_MANUAL_ROLL=trace processor maintainer change

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/tables_views.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/tables_views.sql
@@ -72,7 +72,16 @@ CREATE PERFETTO VIEW track (
   -- expand the args.
   source_arg_set_id ARGSETID,
   -- Machine identifier, non-null for tracks on a remote machine.
-  machine_id LONG
+  machine_id LONG,
+  -- An opaque key indicating that this track belongs to a group of tracks which
+  -- are "conceptually" the same track.
+  --
+  -- Tracks in trace processor don't allow overlapping events to allow for easy
+  -- analysis (i.e. SQL window functions, SPAN JOIN and other similar
+  -- operators). However, in visualization settings (e.g. the UI), the
+  -- distinction doesn't matter and all tracks with the same `track_group_id`
+  -- should be merged together into a single logical "UI track".
+  track_group_id LONG
 ) AS
 SELECT
   id,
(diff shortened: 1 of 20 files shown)


LalitMaganti:
+Primiano/Daniele/Steve FYI.

+Anna: PTAL.

aMayzner (approved):

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/tables_views.sql
```diff
@@ -72,7 +72,16 @@ CREATE PERFETTO VIEW track (
...
+  machine_id LONG,
+  -- An opaque key indicating that this track belongs to a group of tracks which
+  -- are "conceptually" the same track.
+  --
+  -- Tracks in trace processor don't allow overlapping events to allow for easy
+  -- analysis (i.e. SQL window functions, SPAN JOIN and other similar
+  -- operators). However, in visualization settings (e.g. the UI), the
+  -- distinction doesn't matter and all tracks with the same `track_group_id`
+  -- should be merged together into a single logical "UI track".
+  track_group_id LONG
```
aMayzner, line 84: Shouldn't we mention in a name that it's for the UI? Because I guess we don't want people to depend on it outside of the UI

src/trace_processor/tables/track_tables.py
```diff
@@ -67,6 +67,12 @@
             cpp_access=CppAccess.READ_AND_HIGH_PERF_WRITE,
             cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
         ),
+        C(
+            "track_group_id",
+            CppOptional(CppUint32()),
```
aMayzner, line 72: Optional? What is the case it wouldn't be there?

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/tables_views.sql
LalitMaganti: I'm fine with people depending on this otherwise.

src/trace_processor/tables/track_tables.py
LalitMaganti: When it was not part of group at all.

PR #2146, opened 2025-07-13
Use constexpr if instead of SFINAE for WriteIntoTrace.

altimin:
Simplify the checks with the power of C++17 to use single WriteImpl
function with constexpr if instead of SFINAE, which should lead to
better error messages when passing a type which can't be serialised
into a trace.

### DIFF
--- a/include/perfetto/tracing/traced_value.h
+++ b/include/perfetto/tracing/traced_value.h
@@ -324,70 +324,105 @@ class PERFETTO_EXPORT_COMPONENT TracedDictionary {
 
 namespace internal {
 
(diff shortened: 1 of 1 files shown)


altimin:
@ddiproietto, could you take a look at this one as well?

ddiproietto (commented):
Tricky stuff...

I'm not sure this is a net benefit since:

1. It adds more lines of code (+101 -69)
2. It makes compilation slower (at least on my machine, for debug builds with clang from buildtools for src/tracing/traced_value_unittest.cc)

Can you provide an example of better error message?

Do you think it would be possible to reduce compile times?

PR #2147, opened 2025-07-14
tprobes: add "atrace_userspace_only" option

rsavitski:
If set, it prevents "atrace_catgories" from enabling additional ftrace events (besides ftrace/print). Covers both the predefined events and the vendor-supplied extension mapping.


src/traced/probes/ftrace/ftrace_config_muxer.cc
```diff
@@ -195,16 +195,32 @@ std::set<GroupAndName> FtraceConfigMuxer::GetFtraceEvents(
       events.insert(GroupAndName(group, name));
     }
   }
+
   if (RequiresAtrace(request)) {
     InsertEvent("ftrace", "print", &events);
-
+  }
+  if (RequiresAtrace(request) && !request.atrace_userspace_only()) {
```
ddiproietto: For clarity I would remove `RequiresAtrace(request) && `

ddiproietto (approved):
Thanks!

src/traced/probes/ftrace/ftrace_config_muxer.cc
rsavitski: Done.

PR #2149, opened 2025-07-14
ui: make use of newly added track_group_id column

LalitMaganti:
It's the correct thing to group by instead of grouping by name.

### DIFF
--- a/ui/src/plugins/org.kernel.SuspendResumeLatency/index.ts
+++ b/ui/src/plugins/org.kernel.SuspendResumeLatency/index.ts
@@ -36,7 +36,7 @@ export default class implements PerfettoPlugin {
           group_concat(distinct t.id) as trackIds,
           count() as trackCount
         from track t
-        where t.name = "Suspend/Resume Latency"
+        where t.type = 'suspend_resume'
       )
       select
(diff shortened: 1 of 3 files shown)


ui/src/plugins/org.kernel.SuspendResumeLatency/index.ts
```diff
@@ -36,7 +36,7 @@ export default class implements PerfettoPlugin {
           group_concat(distinct t.id) as trackIds,
           count() as trackCount
         from track t
-        where t.name = "Suspend/Resume Latency"
+        where t.type = 'suspend_resume'
```
stevegolton, line 39: While entirely reasonable, this seems unrelated

LalitMaganti: It was part of the overall effort towards "we should not be depending on track names to do things beyond being a display name". I was searching for use of "name" in the codebase and came across this, felt it was worth fixing in the same CL. Let me know if you want it in a different one.

stevegolton: Same CL is fine

PR #2158, opened 2025-07-14
tprobes: rename FtraceProcfs -> Tracefs

rsavitski:
The modern name for /sys/kernel/tracing is "tracefs", while perfetto code is still using a name from the project's origins in 2017. Let's finally move to the upstream name to improve readability.

Generated mostly with:
  sed -i 's/FtraceProcfs/Tracefs/g' $(git grep -l FtraceProcfs)
  sed -i 's/ftrace_procfs/tracefs/g' $(git grep -l ftrace_procfs)
  git mv ...
  tools/format-sources
  tools/gen_all ...


rsavitski:
> Thanks!
> 
> A very nice to have would be to make it an abstract class, so that we don't pull the implementation into test. But that's a cleanup for another day, I guess.

Yeah, one thing at a time.

Issue #941, opened 2024-11-22
[Docs] In the standard library docs, Most outline items on the RHS are truncated


cphlipot1:
In the outline on the RHS of https://perfetto.dev/docs/analysis/stdlib-docs most of the items are truncated with trailing "...", even when there is ample horizontal space in the browser window, impacting readability.

Here is an example of what the page looks like fully maximized on a 4K monitor:
![Image](https://github.com/user-attachments/assets/69748fc2-b86e-47cd-97bd-c443e945db89)

LalitMaganti:
@aMayzner

schartnerc:
Can I create a quick PR for this? @aMayzner

aMayzner:
It would be great, thanks!

LalitMaganti:
Thanks for the fix @schartnerc!

PR #2165, opened 2025-07-15
Docs: Improve Layouting on wide screens

schartnerc:
Based on [#941](https://github.com/google/perfetto/issues/941)

The fractions should only have a minimal layout change on the previous fixed width for MacBooks.
Uses full viewport to display as much of nav, docs and table of contents as possible.


LalitMaganti (approved):

stevegolton (changes requested):
I'm not convinced. This make the site fill the whole screen, which makes the site very **wide** on larger screens, which makes the paragraphs wide and more difficult to read. I'd keep the max width for the content but just bump the available space for the TOC maybe.

schartnerc:
Okay. I understand. I'll increase the size of the toc only on wider screens, so the width of the content on normal screens stays the same.

schartnerc:
@stevegolton @LalitMaganti If you have more suggestions please let know. If you do not like the approach I can also just close the PR.

LalitMaganti:
Tried out the new behaviour. Seems to be reasonable and aligns with what Steve asked for.

Steve is out for the week so adding Ryan and Anna to review.

LalitMaganti (approved):

rsavitski (approved):
Don't know much about css, but can confirm this improves the TOC even on a normal 16:9 screen, as we tend to have long section names.

Thanks.

PR #2169, opened 2025-07-15
Dump the hierarchy path for SF layers.

PriyankaSPX:
So we can avoid running slow recursive queries for trace search. Ensure we handle recursive hierarchies.

Bug: 411363817
Bug: 431939202
Test: tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell --name-filter="SurfaceFlingerLayer|PerfettoTable:winscope"


LalitMaganti:
Fwiw imo this is a bad idea, you'd be better of writing a function/operator or similar to do the on-demand expansion nstead of materializing everything as you're going to waste a lot of memory. OTOH it's up to you as ever.

PriyankaSPX:
> Fwiw imo this is a bad idea, you'd be better of writing a function/operator or similar to do the on-demand expansion nstead of materializing everything as you're going to waste a lot of memory. OTOH it's up to you as ever.

Yeah it's a good point, I'll try and do this as a static table function instead since it's not necessary for general Winscope UI display

src/trace_processor/perfetto_sql/intrinsics/table_functions/winscope_surfaceflinger_hierarchy_paths.cc
```diff
@@ -0,0 +1,179 @@
...
+#include "src/trace_processor/types/trace_processor_context.h"
+
+namespace perfetto::trace_processor {
+
+namespace {
+using SnapshotDecoder = protos::pbzero::LayersSnapshotProto::Decoder;
+using LayersDecoder = protos::pbzero::LayersProto::Decoder;
+using LayerDecoder = protos::pbzero::LayerProto::Decoder;
+using ColumnIndex = tables::SurfaceFlingerLayersSnapshotTable::ColumnIndex;
+
```
LalitMaganti, line 49: Fwiw for the purposes of winscope, as soon as you run this once, you're going to pay the full cost of the string materialization anyway so this is basically no better than what you had before

When I said function what I had in mind was:
a) returning one layer per row
b) only returning for the start points you care about

But your call.

PriyankaSPX: We don't actually need this for the UI. It's part of a separate exploration effort into how we can better automate things in Winscope. So if we're avoiding the full string cost unless the table is created on demand that's be better than what I had before... but I think I get what you mean? If you mean a table with columns like `snapshot_id, layer_id, ancestor` with one row per ancestor. We do care about the path for every single layer though

LalitMaganti: The last bet is what I meant yes. Doing the string join eagerly is both bad for memory and means you have to do substring searches if you want to find "the layer that contains x". It's better to flatten that out to avoid this.

PriyankaSPX: Understood. We will need to populate a table for all layers from all snapshots though, rather than finding the ancestors for a particular layer/snapshot combo. But I can defo flatten it out

PriyankaSPX: done - ptal (just in case i'm still misunderstanding!)

PR #2172, opened 2025-07-16
Add a machine filter to data_source messages in perfetto configs

jahdiel-alvarez:
Allow data source producers to be filtered by the machine they come from in multi-VM traces. The machine name can be set either by setting the PERFETTO_MACHINE_NAME env variable or, in Android, by setting the traced.machine_name system property. If none of these values are set then the kernel's utsname sysname is used as the machine name (e.g Linux in Android).

Bug: 414873776


include/perfetto/ext/base/android_utils.h
```diff
@@ -58,9 +58,15 @@ struct SystemInfo {
...
 };
 
+// Returns the device's utsname information.
+Utsname GetUtsname();
+
 // Returns the device's system information.
 SystemInfo GetSystemInfo();
 
+// Returns the machine name.
+std::string GetMachineName();
```
rsavitski: This is a misnomer in two ways:
* it's not what the reader would expect (utsname->machine)
* it's also not utsname->sysname, due to perfetto-specific overrides.

Can we instead put this inside `struct SystemInfo` (and the corresponding proto) as `perfetto_machine_name`? The struct already has fields for things that aren't strictly derived from utsname, and the "perfetto_" prefix hints at the fact that it's a name that has special meaning to us (and therefore that it might not be just the utsname).

include/perfetto/ext/tracing/core/tracing_service.h
```diff
@@ -302,6 +302,11 @@ class PERFETTO_EXPORT_COMPONENT ConsumerEndpoint {
 };  // class ConsumerEndpoint.
 
 struct PERFETTO_EXPORT_COMPONENT TracingServiceInitOpts {
+  // Name of machine hosting the tracing service (traced). If this option is
+  // empty the tracing service will use the OS system name (see `uname -s`) as
+  // the default value.
+  std::string machine_name;
```
rsavitski: I might be missing something (testability?), but why does this need to be supplied during construction? Shouldn't `traced` always be able to figure out its own machine name on demand by calling the utils function?

src/traced_relay/relay_service.cc
```diff
@@ -187,6 +187,8 @@ void RelayClient::OnServiceDisconnected() {
 void RelayClient::InitRelayRequest() {
   protos::gen::InitRelayRequest request;
 
+  request.set_machine_name(base::GetMachineName());
```
rsavitski: As far as I can tell, the `RelayPort` service is kind of best-effort / optional. It's not directly depended on by the main producer proxying implementation below (i.e. the part that relays connections to the host's `ProducerPort`). The `RelayService::Start` kicks off this RelayPort connection, but it's asynchronous and nothing checks for its success.

If we're adding this `RelayPort` ipc as a precondition for correct machine-filtering, we need to have a clearer dependency between proxied producers and this IPC that is performed by the RelayClient.

So I can see two ways this could go:

  (A) We decide that there will be future dependencies where some producer proxying is dependent on an operation by `RelayClient` --> we need to make the two ipc services linked so that a proxied producer waits until the RelayClient *knows* that the `InitRelay` ipc succeeded.

  (B) Instead of adding onto `RelayPort`, we instead add to the [`SetPeerIdentity`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/ipc/wire_protocol.proto;l=78;drc=189992a04d80da240e34fdee4d2ce6abd60dd8ff) ipc, which is the [canonical step](https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/ipc/host_impl.cc;l=356;drc=23f8392415bf51999d3d69674ce48bf2f43148ea) where, under VSOCK, the host traced derives the machine_id for both RelayPort and ProducerPort connections from the relay (note the two distinct calls to `GenerateSetPeerIdentityRequest()` from relay_service.cc, one per port).

Thoughts? (Note: I'm new to the guts of the relay implementation, so my suggestions might not be optimal either).

protos/perfetto/config/perfetto_config.proto
```diff
@@ -4115,6 +4123,9 @@ message TraceConfig {
     // "bar"] will enable data sources on both "foo" and "bar" (if they exist).
     repeated string producer_name_filter = 2;
     repeated string producer_name_regex_filter = 3;
+
+    // Filter data sources by the source machine.
+    optional MachineFilter machine_filter = 4;
```
rsavitski: nit: I would consider unnesting the options (it's unlikely we'll need more than a couple of options for machine filtering, and it's easier to read the proto without extra container messages).

Separately, there are things that your implementation does (env variable as an override and some "host" magical value for the host(?)) that the comment doesn't cover, but should.

src/tracing/ipc/service/relay_ipc_service.cc
```diff
@@ -67,7 +67,8 @@ void RelayIPCService::InitRelay(const protos::gen::InitRelayRequest& req,
     relay_endpoints_.Insert(client_id, std::move(ep));
   }
 
-  endpoint->CacheSystemInfo(req.system_info().SerializeAsArray());
+  endpoint->CacheSystemInfo(req.machine_name(),
```
rsavitski: nit (other comments are more important): this method name (and surrounding comments) no longer describe the importance of the operation. The significance of machine_name is completely hidden.

include/perfetto/ext/base/android_utils.h
jahdiel-alvarez: Most often utsname->machine is referred to as `Arch`, we do this in system probes parser when parsing the system info (kernel folks chose a bad name). That being said probably a better name for the method could be `GetPerfettoMachineName`.

The issue with passing the machine name through the SystemInfo proto is that it will require decoding the proto on the traced-side, which is something I was trying to avoid. Also, if we are thinking of using a different method from InitRelay to pass the information then we would need to extract the system information twice, the second time just to get a single value.

include/perfetto/ext/tracing/core/tracing_service.h
jahdiel-alvarez: Yes for testability. I am using this to test the filtering behavior when an external machine name is used. We could change it such that `traced` simply calls the utils function and remove the test (it's the same code path regardless).

protos/perfetto/config/perfetto_config.proto
jahdiel-alvarez: Sure, I can un-nest and add a better comment.

cervantesyu: Do we plan to add more ways to filter (e.g. machine_name_regex_filter feels like useful in tracing many machines, but is outside of the scope)? If filter by machine names is good enough then un-nest is a better option.

src/traced_relay/relay_service.cc
cervantesyu: The RelayPort works as an out-of-band channel passing over crucial information that makes proxied producer work correctly. Without info sent over this port, remote machine producers won't work correctly in TP/UI - timestamps of the remote machine producer will be off by a very large offset without clock synchronization.

In this sense, it's more than an optional service, though it's best effort and not guaranteed that it will always succeed.

A) This is more complicated than option B), and creates an explicit precondition before start proxying for producers. I am not sure if blocking proxied producer connections will lead to more problems. Maybe disconnect and propagate the back pressure to producers is a better option. Requiring that the relay port fully initialized is good in a sense that we won't see traces missing clock sync packets that leads to hard-to-debug data.

B) looks simpler to me. In this sense, machine_name can be thought about as an optional, human-friendly "alias" to machine_id. Downside is it's less efficient than option A because machine_name is the same for all producers, but works in the sense that machine_id/machine_name works as a higher-level namespace for producer name. But this being less efficient in data transferred over the socket should be fine compared to the amount of trace data.

include/perfetto/ext/tracing/core/tracing_service.h
rsavitski: right, I would prefer we didn't pollute TracingServiceInitOpts just as a testing shim then. Fine to leave it untested or test it via a different means.

src/traced_relay/relay_service.cc
rsavitski: Then to make things concrete, my position:
* for this option, let's see how messy it looks when the perfetto_machine_name is communicated through `SetPeerIdentity`, as that makes most sense in terms of logical dependencies. I don't think we care about O(10) bytes of serialised proto payload overhead since we're already paying IPC overheads for the message.
* in a future CL, it would be nice to make it clear in a trace that RelayPort was set up correctly, even if it's not a hard dependency for forwarding producers to the host's ProducerPort.

include/perfetto/ext/base/android_utils.h
rsavitski: re: packing it in SystemInfo: let's see how the other conversation about SetPeerIdentity goes first. But let's keep referring to it as perfetto_machine_name regardless.

> The issue with passing the machine name through the SystemInfo proto is that it will require decoding the proto on the traced-side, which is something I was trying to avoid.

Just fyi, this is already fully decoded (and then reserialised) at the time of writing:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/tracing/ipc/service/relay_ipc_service.cc;l=70;drc=d0260a368e65b2be56d179d21fab90976fd494b8

But regardless, I'm not convinced we need to worry about encoding/decoding overheads when we're already in the domain of vsock IPCs other host<->VM tracing overheads.

include/perfetto/ext/tracing/core/tracing_service.h
jahdiel-alvarez: Done

protos/perfetto/config/perfetto_config.proto
jahdiel-alvarez: I think un-nesting should be fine, even if we add `machine_name_regex_filter` I don't think we will be adding too many more machine related filters after that. I did have in mind having an `operating_system_filter` to make it even more convenient to select machines but not in the immediate plans. Even in that case, I think unnesting is still okay.

src/traced_relay/relay_service.cc
jahdiel-alvarez: Updated PR. We are now passing the machine name via SetPeerIdentity.

src/tracing/ipc/service/relay_ipc_service.cc
jahdiel-alvarez: No longer relevant.

protos/perfetto/config/perfetto_config.proto
jahdiel-alvarez: Unnesting done.

src/ipc/host_impl.cc
```diff
@@ -215,6 +216,9 @@ void HostImpl::OnNewIncomingConnection(
   client->id = client_id;
   client->sock = std::move(new_conn);
   client->sock->SetTxTimeout(socket_tx_timeout_ms_);
+  if (client->sock->family() == base::SockFamily::kUnix) {
+    client->machine_name = base::GetPerfettoMachineName();
```
rsavitski: This is potentially adding an env lookup + sysprop lookup + syscall to all ipc connections, even for host-only tracing.

We should either:
* cache this so that we only evaluate the name once per traced restart (plus the name should stay consistent for the lifetime of the process, even if someone plays around with the sysprop, imo).
* not set this at all for unix connections. In other words, make the machine name only relevant to relayed producers, while the host has to be matched by "host".

I'm actually leaning towards the latter even though it's less consistent. The only use-case for a settable localhost name would be what, the ability to get different results when running the same config on linux vs android?

src/tracing/service/tracing_service_impl.h
```diff
@@ -913,6 +916,9 @@ class TracingServiceImpl : public TracingService {
                                              uint64_t trigger_name_hash);
   void StopOnDurationMsExpiry(TracingSessionID);
 
+  // Returns the name of a machine based on the machine's id.
+  std::optional<std::string> GetMachineName(MachineID machine_id);
```
rsavitski: nit: this looks unused?

rsavitski (commented):
I like this version more, thanks. Just posing the question of the need for settable machine names for host.

src/ipc/host_impl.cc
jahdiel-alvarez: Not quite. With the drive towards software-defined vehicles in the automotive industry you end up with systems which have multiple VMs running at once, all interconnected plus the hypervisor. A very common hypervisor is QNX, which isn't Linux based. 

These VMs and hypervisor could be running different OSes. Not only that but each VM is most likely handling different subsystems of the car, meaning that even though they might share the same underlying OS, they might not share all the same data sources. So having a settable localhost name is actually quite benefitial for config sharing across teams. It is very likely that we simply set different machine names via the sysprops in our init.rc files for the different types of VMs we use. Hence, if a config filters based on the custom machine names, it wouldn't actually matter which VM is the host the trace would work regardless.

The caching approach seems like a better alternative for AAOS. We already turn off traced/traced_relay when changing the configuration (traced <-> traced_relay), so its fine that one has to restart the service to change the name.

src/tracing/service/tracing_service_impl.h
jahdiel-alvarez: Oops, missed that one. 

Would be nice if we could check for this sort of thing: unused variables, unused functions, undefined functions in the pre-commit hooks.

src/ipc/host_impl.cc
rsavitski: I assumed that the "host" traced is always on the same type of Linux/Android OS, and therefore didn't need producer filtering beyond host-vs-not. Is that not the case, or were you thinking of config sharing between relayed tracing and fully-inside-VM tracing (where a VM is running its own traced instance)?

Either way, let's figure out a caching approach then.

src/tracing/service/tracing_service_impl.h
rsavitski: I'm a bit surprised the compiler didn't warn on this myself.

jahdiel-alvarez: Done

src/ipc/host_impl.cc
jahdiel-alvarez: Did the caching in HostImpl

include/perfetto/ext/ipc/client_info.h
```diff
@@ -62,13 +67,17 @@ class ClientInfo {
...
+  std::string machine_name() const { return machine_name_; }
+
  private:
   ClientID client_id_ = 0;
   // The following fields are emitted to trace packets and should be kept in
   // sync with perfetto::ClientIdentity.
   uid_t uid_ = kInvalidUid;
   pid_t pid_ = base::kInvalidPid;
   base::MachineID machine_id_ = base::kDefaultMachineID;
+  std::string machine_name_;
```
rsavitski, line 81: nit: the "The following fields..." comment is no longer accurate. Maybe keep this last in field order, but separate by a different comment?

```
 private:
  ClientID client_id_ = 0;
  // The following fields are emitted to trace packets and should be kept in
  // sync with perfetto::ClientIdentity:
  uid_t uid_ = kInvalidUid;
  pid_t pid_ = base::kInvalidPid;
  base::MachineID machine_id_ = base::kDefaultMachineID;
  // ... end of fields shared with ClientIdentity.
  std::string machine_name_;
```

jahdiel-alvarez:
@primiano are you okay with the setting of `machine_name_` within the HostImpl constructors? (see @rsavitski's comment)

include/perfetto/ext/ipc/client_info.h
jahdiel-alvarez: Done

PR #2177, opened 2025-07-16
Generate descriptors from multiple proto sources

JoeNotCharlesGoogle:
protoc only supports one input file with the generate_descriptor option.
  
This works around the lack of support by adding a step to the `perfetto_proto_library` target. It creates a temp file that imports all files in the `sources` list and uses the temp file as input to protoc.

Also updates protos/third_party/chromium/BUILD.gn to generate the descriptor from `chrome_track_event_sources` instead of a single `descriptor_root_source`. This will allow Chrome to add multiple files to `chrome_track_event_sources` in an upcoming patch.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -11908,6 +11908,7 @@ genrule {
 genrule {
     name: "perfetto_protos_third_party_chromium_descriptor",
     srcs: [
+        "out/tmp.gn_utils/gcc_like_host/gen/protos/third_party/chromium/descriptor_root_source.proto",
         "protos/perfetto/trace/track_event/chrome_active_processes.proto",
         "protos/perfetto/trace/track_event/chrome_application_state_info.proto",
         "protos/perfetto/trace/track_event/chrome_compositor_scheduler_state.proto",


Android.bp
```diff
@@ -11908,6 +11908,7 @@ genrule {
 genrule {
     name: "perfetto_protos_third_party_chromium_descriptor",
     srcs: [
+        "out/tmp.gn_utils/gcc_like_host/gen/protos/third_party/chromium/descriptor_root_source.proto",
```
LalitMaganti, line 11911: as you can see, this is not going to work :)

gn/proto_library.gni
```diff
@@ -344,15 +344,56 @@ template("perfetto_proto_library") {
   # Generate the descriptor if the option is set.
   if (defined(invoker.generate_descriptor)) {
     target_name_ = string_replace(target_name, expansion_token, "descriptor")
+
+    if (defined(invoker.descriptor_root_source)) {
```
LalitMaganti, line 348: Unfortunately, it's not as simple as this we build in 4 different build systems so you're going to need to do something which works in all the build systems. At the least you've broken Android here, Bazel/Blaze will almost certainly be broken as well.

JoeNotCharlesGoogle: Is the issue that it's added the tmp .proto file to the Android build list, but didn't add a rule to actually generate that file to the Android build?

LalitMaganti: Well more that the out file you see above is an artifact of building with gn. You have to supress it in `tools/gen_android_bp` and then add a rule to actually generate the amalgamation there. You'll also have to do this with Bazel.

FWIW, taking a step back I really don't think this is a good idea to do. I would suggest adding 1:1 build rules for each root level file and calling it a day or manually maintain this file yourself. It will be a nightmare to do this autogeneration and maintain it in the longer term. Being explicit this much better IMO.

JoeNotCharlesGoogle: I'll just check in a manually generated wrapper.

PR #2179, opened 2025-07-16
Re-add surfaceflinger tables related to jank metrics outside stdlib.

devianb:
These tables are being used in telemtry scripts, and removing them causes breakages.

### DIFF
--- a/src/trace_processor/metrics/sql/android/jank/relevant_threads.sql
+++ b/src/trace_processor/metrics/sql/android/jank/relevant_threads.sql
@@ -57,6 +57,14 @@ DROP TABLE IF EXISTS android_jank_cuj_sf_process;
 CREATE PERFETTO TABLE android_jank_cuj_sf_process AS
 SELECT * FROM _android_sf_process;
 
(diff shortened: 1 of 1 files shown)


src/trace_processor/metrics/sql/android/jank/relevant_threads.sql
```diff
@@ -57,6 +57,14 @@ DROP TABLE IF EXISTS android_jank_cuj_sf_process;
 CREATE PERFETTO TABLE android_jank_cuj_sf_process AS
 SELECT * FROM _android_sf_process;
 
+DROP TABLE IF EXISTS android_jank_cuj_sf_main_thread;
```
LalitMaganti, line 61: Please leave TODOs that these only exist for Google3 backcompat and should be deleted when we migrate the pipelines away from using them.

LalitMaganti (approved):

src/trace_processor/metrics/sql/android/jank/relevant_threads.sql
devianb: Done.

PR #2184, opened 2025-07-16
ui: fix horrible performacne of startup plugin on lots of startups

LalitMaganti:
Instead of doing N + 1 queries, just do a fixed number independent of
number of startups.

Also rewrite code to push down string manipulation into SQL instead of
materializing and doing it in Javascript.


stevegolton (approved):

ui/src/plugins/dev.perfetto.AndroidStartup/optimizations.ts
```diff
@@ -48,141 +32,143 @@ const FILTER_PATTERN = /filter=([^\s]+)/;
...
+    classLoadingTracks.push(childTrack);
   }
 
-  if (startups.length === 0) {
-    // Nothing interesting to report.
-    return undefined;
-  }
+  await trace.engine.query(
+    `
+      CREATE PERFETTO FUNCTION _startup_compilation_state(filter STRING)
```
aMayzner, line 60: Are we sure those shouldn't live in a viz folder? It would be easier to test them and debug them there.

aMayzner (approved):

ui/src/plugins/dev.perfetto.AndroidStartup/optimizations.ts
LalitMaganti: Yes I'm sure.

LalitMaganti:
__

ui/src/plugins/dev.perfetto.AndroidStartup/optimizations.ts
LalitMaganti: If anything it should be in a startup breakdowns module separately. But it is not complex enough to do it that yet.

tikurahul (commented):
Thank you for cleaning this up.

PR #2187, opened 2025-07-17
tp: Handle ChromeThreadDescriptor is_sandboxed_tid in trace processor

jahdiel-alvarez:
TraceProcessor will soon require tids to have unique, non-overlapping lifetimes. Using the is_sandboxed_tid field, TP can now create a unique synthetic tid for track descriptors. This is needed because there is no feasible way to provide the mapping of namespaced tids and the "real" root-level tid from the Chrome browser.

### DIFF
--- a/src/trace_processor/importers/proto/track_event_parser.cc
+++ b/src/trace_processor/importers/proto/track_event_parser.cc
@@ -286,9 +286,15 @@ void TrackEventParser::ParseTrackDescriptor(
   }
 
   if (decoder.has_thread()) {
-    UniqueTid utid = ParseThreadDescriptor(decoder.thread());
     if (decoder.has_chrome_thread()) {
+      protos::pbzero::ChromeThreadDescriptor::Decoder chrome_decoder(
+          decoder.chrome_thread());
+      bool is_sandboxed = chrome_decoder.has_is_sandboxed_tid() &&
+                          chrome_decoder.is_sandboxed_tid();
+      UniqueTid utid = ParseThreadDescriptor(decoder.thread(), is_sandboxed);
       ParseChromeThreadDescriptor(utid, decoder.chrome_thread());
+    } else {
+      ParseThreadDescriptor(decoder.thread(), /*is_sandboxed=*/false);
     }
   } else if (decoder.has_process()) {
     UniquePid upid =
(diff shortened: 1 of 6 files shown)


jahdiel-alvarez:
The new proto field was added in https://github.com/google/perfetto/pull/2168

LalitMaganti (approved):
Please add a test.

jahdiel-alvarez:
Test added

src/trace_processor/importers/proto/track_event_parser.cc
```diff
@@ -363,11 +369,17 @@ void TrackEventParser::ParseChromeProcessDescriptor(
...
   protos::pbzero::ThreadDescriptor::Decoder decoder(thread_descriptor);
-  UniqueTid utid = context_->process_tracker->UpdateThread(
-      static_cast<uint32_t>(decoder.tid()),
-      static_cast<uint32_t>(decoder.pid()));
+  auto pid = static_cast<int64_t>(decoder.pid());
+  auto tid = static_cast<int64_t>(decoder.tid());
+  // If tid is sandboxed then use a unique synthetic tid, to avoid
+  // having concurrent threads with the same tid.
+  if (is_sandboxed) {
+    tid = ProcessTracker::CreateSyntheticTid(tid, pid);
```
betasheet, line 380: Won't this create conflicting tids between parsing and tokenizing this descriptor?

I'm not quite sure what the consequence of that is (we lack some comments about why we UpdateThread() below again), but I wonder if we need to keep track of the tokenization-time translations into synthetic tids somehow?

Also not sure if we need to do something about two different tracks referencing the same thread. I recall that we had this in the past (e.g. one thread emitting events on behalf of another) but I think we got rid of that at some point in Chrome.

src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -197,11 +198,22 @@ ModuleResult TrackEventTokenizer::TokenizeTrackDescriptorPacket(
     }
```
betasheet, line 198: There's a TokenizeThreadDescriptor() call right before, which eventually stores the tids into the sequence state (and they are read from there by parsing code later). I am not sure if keeping these inconsistent with the reservation etc works?

This all feels a bit more subtle than I thought ...

Issue #932, opened 2024-11-13
Is ftrace filter supported in perfetto trace config?


wangjiyang:
Hi Perfetto team.

I am looking for a ftrace filter feature. Background is I am trying to tracing sched:sched_switch ftrace event of some specified processes for a very very long time, eg, serveral days. Enabling sched_switch events for all processes will produce too much  data and may consumes CPU and disk. The overhead is too much for our usecases. So I wonder what's the best practice of tracing limited number of processes?

wangjiyang:
I would use some ebpf features to do such filtering. Eg, calling bpf helpers like `bpf_trace_printk`, and acts as what atrace done to take advantage of ftrace features to link with perfetto. I am still investigating and not sure if `traced_probe` already has this kind of raw ftrace buffers collected. Please share some information if you know. Based on this next step would be parsing this raw buffer and emit track information to trace processor. This approach would be compatible with current ftrace importer, but has some challege because that would need to create a new data source and it depends on ebpf.

wangjiyang:
Seems this is what i needed. Is there any samples?

https://perfetto.dev/docs/reference/trace-config-proto#PerfEvents.Tracepoint

LalitMaganti:
> Eg, calling bpf helpers like bpf_trace_printk, and acts as what atrace done to take advantage of ftrace features to link with perfetto.

If you are doing this, you don't need to do any integration with perf etc. You just need to enable the ftrace events `ftrace/print`, `ftrace/bprint` and `ftrace/bputs` and the data should just show up in Perfetto.

> https://perfetto.dev/docs/reference/trace-config-proto#PerfEvents.Tracepoint

This is totally irrelevant, this is about traced_perf stack sampling. You don't need this, just enable the events above in the normal ftrace data source.

sashwinbalaji:
Available after  #2851

PR #2189, opened 2025-07-17
Implement ftrace pid filtering in traced.

sashwinbalaji:
This change introduces the `pid_filter` field in the `FtraceConfig`
proto, allowing clients to specify which processes should have their
events traced.

The effective filter is computed as the union of all active
data source configurations.

When no pid_filter message is provided,
all pids are traced, which aligns with the legacy behavior.

This is done interacting with two files:
- `set_event_pid`: for specifying the PIDs to be traced, where the
content is a space-separated list.
- `options/event-fork`: for enabling to trace forked child processes.
If enabled and pid filter is not empty, child processes will also be
traced.

Fixes: #932 
Tested on both android and linux by running configs tracking all pids and tracking specific ones separately and together (logs: http://dr/corp/drive/u/0/folders/10cmeNC5w0fKtReyXS2cWaLb6gtdQeOKO)


protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -272,6 +272,47 @@ message FtraceConfig {
   // serialise raw tracing pages that the implementation cannot parse.
   optional bool debug_ftrace_abi = 31;
 
+  // Configuration for filtering ftrace events by Process ID (PID).
```
rsavitski, line 275: General note (see other comments before changing the patch): we shouldn't use "PID" in our own configuration terminology, as it's ambiguous. In fact, I think you got tricked here yourself.

"PID" in the kernel means the thread id (TID). So the "set_event_pid" control file is operating on thread ids, not TGIDs.

We should prefer to use TID/TGID in perfetto configuration as much as possible. PID as TGID is acceptable in some places where the use of userspace terminilogy is obvious from context, but it isn't here.

protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -272,6 +272,47 @@ message FtraceConfig {
...
+
+    // If set to true, any child processes forked from the PIDs specified
+    // in |pids_to_trace| will also be included in the trace.
+    // This interface interacts with |/sys/kernel/tracing/options/event-fork|.
+    optional bool enable_event_fork = 2;
+  }
+
+  // Specifies a filter for which PIDs to trace.
+  //
+  // Behavior with multiple concurrent sessions:
```
rsavitski, line 291: The amount of effort spent on multiplexing this option feels disproportionate given how narrow the option is.

As written, ftrace_config.pid_filter is a hardcoded set of threads per config, so we're in the domain of tracing on a local machine (as opposed to something copy-pastable/shareable, like a process name). I'd argue that making things work transparently for concurrent tracing sessions is less of a concern in that situation, since the user should have more control over their tracing.

Given how extremely niche this option is, I would much rather keep the implemenetation simple and not offer multiplexing at all. So if you set this option, you immediately override set_event_pid for that tracefs instance. (If we want to help avoid footguns, we could reject the config if there's already a perfetto trace running for that instance...)

Note that we have |instance_name| option above for tracefs instances. So if concurrent tracing is important, then I'd argue that using a separate tracefs instance to contain set_event_pid is the correct way to configure this in the first place.

protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -272,6 +272,47 @@ message FtraceConfig {
...
+  // - Result: set_event_pid is set to [10, 11, 12].
+  //
+  // Example 2:
+  // - Session A: pids_to_trace = [10, 11]
+  // - Session B: (no pid_filter)
+  // - Result: set_event_pid is cleared, all PIDs traced.
+  //
+  // Introduced in: perfetto v52.
+  // Supported on: Android 25Q3+.
+  optional PidFilter pid_filter = 35;
```
rsavitski, line 314: TBH, I'm unconvinced we need the option at all:
* The use-case is very narrow and can be achieved by having a script that sets up the tracefs control files outside of perfetto, and then letting perfetto do the recording.
* The option is full of sharp edges, and breaks assumptions across the perfetto stack. For example it breaks |compact_sched| above on the recording side (you'd need to force it to off when using this option). Similarly, the perfetto UI and sched/thread_state tracking assume that we see all scheduling events per cpu (if there are any at all). With this option, all scheduling state outside of the threads covered by set_event_pid appears broken/malformed (we see something switching in and never switching out). Separately, there are other events (rss_state comes to mind) that are relevant for a thread but are not emitted from the thread context. In fact, the only reason the option works for sched* in the first place is due to [special handling hardcoded into the kernel](https://source.corp.google.com/h/kernel/pub/scm/linux/kernel/git/torvalds/linux/+/master:kernel/trace/trace_events.c;l=1076;drc=6956ea9fdcf706dc6201bf7146af188539eac9b1).
* Advertising this at the ftrace_config level is going to mislead people into thinking this "just works", and their time will be wasted because of the aforementioned sharp edges.

There was a similar question on discord the other day, asking about cpu-masking for ftrace (and they figured out how to configure this outside of perfetto themselves). So perhaps instead of a subtle ftrace_config option, we should instead have a docs FAQ entry on how to use niche tracefs features with perfetto? If we do decide to put this into the event config, we need a way of making it clear that it's not general purpose (the equivalent of --i-know-what-i-am-doing).

rsavitski (changes requested):

protos/perfetto/config/ftrace/ftrace_config.proto
LalitMaganti: I don't agree with the idea of we should not land this because it breaks things in our stack. There are a lot of subtle features like that in a lot of places in Perfetto, for power users letting them do what they want is good thing. 

I do agree that gating this a bit more behind a "I know what I'm doing flag" makes sense. I would also consider making some other ones like funcgraph behind this option as well.

rsavitski: Fair points, one vote is enough to sway me towards adding the option. Then my main feedback is: remove multiplexing, make it force compact_sched off, and document more clearly that it is best-effort.

Ashwin/Lalit, thoughts on how to achieve the "I know what I'm doing" meaning?
* per-option comments or even a big banner-style comment saying that all options below it require great care. But people copy-pasting configs will entirely ignore that and then get burned, and we'll be repeatedly asked why it doesn't magically solve process-scoped tracing.
* we could introduce a wrapper proto, e.g. `message Brittle { set_event_pid, set_event_cpu, ... }`. But: it won't be aligned with existing options; is a nuisance to type in textproto; and it'll be difficult to say what does and does not make the cut (for example, we do support multiplexing of function_graph, but not switching to it mid-way through a trace...). Sounds too rigid to be maintainable.
* some option prefix? `brittle_set_event_pid`? That might misrepresent the data itself as unreliable, even though the feature is robust if you know the limitations.
* maybe document not so much the subtlety, as much as the level of the configuration? E.g. call the options `low_level_tid_filter`, `low_level_cpu_filter`?

LalitMaganti: I think a combination of proto documentation making clear these are features with gotchas + having a wrapper proto would be the most clean option. If we support them "properly" some day, we could move out of the proto and deprecate the old option etc. 

I agree it's not consistent with what we have today but it's better to start the precedent now and agree we will move in this direction going forward than continuing to not make this clear or worse not land the feature at all.

protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -272,6 +272,47 @@ message FtraceConfig {
...
+    // A list of Process IDs (PIDs) to include in the trace.
+    // If this list is empty, it implies that all PIDs should be traced for
+    // this specific filter instance.
+    // This interface interacts with |/sys/kernel/tracing/set_event_pid|.
+    repeated int32 pids_to_trace = 1;
+
+    // If set to true, any child processes forked from the PIDs specified
+    // in |pids_to_trace| will also be included in the trace.
+    // This interface interacts with |/sys/kernel/tracing/options/event-fork|.
+    optional bool enable_event_fork = 2;
```
rsavitski, line 286: Something to consider: maybe we want a generic option for setting/clearing/keeping arbitrary "/tracefs/options/...". At least control over "record-cmd" and "record-tgid" is something that's been on a perfetto nice-to-have for a while.

LalitMaganti: Yeah this seems like a good point, one I agree with.

sashwinbalaji:
closing with newer version: go/pgh/2851

PR #2195, opened 2025-07-17
Moving 'Device State' track to StandardGroups

michalbrz:
This is so that 'Device State' can be used by other plugins and not depend on AndroidLongBatteryTracing

### DIFF
--- a/ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
+++ b/ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import {Trace} from '../../public/trace';
+import StandardGroupsPlugin from '../dev.perfetto.StandardGroups';
 import {PerfettoPlugin} from '../../public/plugin';
 import {Engine} from '../../trace_processor/engine';
 import {createQuerySliceTrack} from '../../components/tracks/query_slice_track';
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.AndroidLongBatteryTracing/index.ts
```diff
@@ -1784,6 +1789,11 @@ export default class implements PerfettoPlugin {
     const containedTraces = (ctx.openerPluginArgs?.containedTraces ??
       []) as ContainedTrace[];
 
+    const deviceStateGroup = ctx.plugins
+      .getPlugin(StandardGroupsPlugin)
+      .getOrCreateStandardGroup(ctx.workspace, 'DEVICE_STATE');
+    this.groups.set(this.deviceStateGroupName, deviceStateGroup);
```
michalbrz: this is needed because `addTrack` method from line 844 receives group name as string and then checks if it's present in `this.groups`. If not, it creates new group with that name. `addTrack` (and methods using addTrack) is used in many places and refactoring this would be bigger change. So hopefully this workaround is enough

stevegolton: Sounds good to me!

stevegolton (approved):

aMayzner (approved):

LalitMaganti (approved):

Issue #2188, opened 2025-07-17
fopen(filename, "re") silently fails on windows


darax:
https://github.com/google/perfetto/blob/941c8ad24a062b752fda022b15e863874835123b/src/profiling/deobfuscator.cc#L260

I was trying to deobfuscate a trace I took from a production build of the app with traceconv using the [instructions provided on perfetto.dev](https://perfetto.dev/docs/data-sources/native-heap-profiler#deobfuscation).  The deobfuscation_map file seemed to always be empty. Out of curiosity I cloned the repo and built traceconv myself. {Kudos to the team, the windows build instructions worked perfectly}. I discovered that the fopen call was failing. I modified the call to fopen(filename, "r") and it seems to work now. I'm guessing adding the "e" is intentional since this occurs in a few places in the codebase, and has been so for quite some time. Seems this is a flag on linux. Unfortunately, the published binaries for Windows do not work.

LalitMaganti:
Thanks for finding! Do you mind sending a PR fixing this? Should just be a matter of adding a shim and doing an ifdef for Windows where you omit the flag?

darax:
Will do, just wanted to make sure I wasn't missing something.

PR #2203, opened 2025-07-17
on windows pass only fopen flags it understands

darax:
Fix for https://github.com/google/perfetto/issues/2188


include/perfetto/ext/base/file_utils.h
```diff
@@ -36,9 +36,11 @@ namespace base {
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
 using FileOpenMode = int;
 inline constexpr char kDevNull[] = "NUL";
+#define FOPEN_READ_FLAG "r"
 #else
 using FileOpenMode = mode_t;
 inline constexpr char kDevNull[] = "/dev/null";
+#define FOPEN_READ_FLAG "re"
```
LalitMaganti: Can we use a static constexpr similar to above instead?

PR #2217, opened 2025-07-21
Add debug track for frame with max dur blocking call in per-frame metric

devianb:
For the per-frame metric, apart from pinning the CUJ and blocking calls, it is useful to pin the frame which contains the blocking call with the max duration. For the aggregation case 'max_dur_per_frame_ns', added support to pin the frame during which the blocking call with the max duration occurred in a separate debug track.

Test: Manual trace and metric check
Bug: 394866952

### DIFF
--- a/ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
+++ b/ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
@@ -26,7 +26,7 @@ import {
   addDebugSliceTrack,
   DebugSliceTrackArgs,
 } from '../../../components/tracks/debug_tracks';
-import {LONG, QueryResult} from '../../../trace_processor/query_result';
+import {LONG, QueryResult, Row} from '../../../trace_processor/query_result';
 
 class BlockingCallMetricHandler implements MetricHandler {
   /**
@@ -132,9 +132,9 @@ class BlockingCallMetricHandler implements MetricHandler {
         frame_id
       FROM _blocking_calls_frame_cuj
       WHERE
-        process_name = "${processName}"
-        AND name = "${blockingCallName}"
-        AND cuj_name = "${cuj}"
+        process_name = '${processName}'
+        AND name = '${blockingCallName}'
+        AND cuj_name = '${cuj}'
       -- select frame_id for the metric with the maximum duration.
       ORDER BY dur DESC
       LIMIT 1`);
@@ -145,9 +145,20 @@ class BlockingCallMetricHandler implements MetricHandler {
   ): Promise<
     Pick<DebugSliceTrackArgs, 'data' | 'columns' | 'argColumns' | 'title'>
   > {
-    const row = (
-      await this.getFrameIdWithMaxDurationBlockingCall(ctx, metricData)
-    ).firstRow({frame_id: LONG});
+    let row: Row = {
+      frame_id: null,
+    };
+
+    try {
+      row = (
+        await this.getFrameIdWithMaxDurationBlockingCall(ctx, metricData)
+      ).firstRow({frame_id: LONG});
+    } catch (e) {
+      throw new Error(
+        `${e.message} caused by: No frame found for the given process, CUJ and blocking call.`,
+      );
+    }
+
     // Fetch the ts and dur of the frame corresponding to the above frame_id.
     const frameWithMaxDurBlockingCallQuery = `
       SELECT


devianb:
Old reference CL:
https://android-review.git.corp.google.com/c/platform/external/perfetto/+/3491212

With the stdlib migration almost done, this plugin implementation is modified now based on the new tables.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +112,51 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+    const config = {
+      data: {
+        sqlSource: frameWithMaxDurBlockingCallQuery,
+        columns: ['frame_id', 'ts', 'dur'],
+      },
+      columns: {ts: 'ts', dur: 'dur', name: 'frame_id'},
+      argColumns: ['frame_id', 'ts', 'dur'],
+      title: trackName,
+    };
+    return Promise.resolve(config);
```
LalitMaganti: Just return config directly, you don't need to do Promise.resolve. You can also inline a bunch of variables above (trackName, config etc).

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +112,51 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+          process_name = "${processName}"
+          AND name = "${blockingCallName}"
+          AND cuj_name = "${cuj}"
+          -- select frame_id for the metric with the maximum duration.
+        ORDER BY dur DESC limit 1`);
+    const row = result.firstRow({frame_id: LONG});
+    // Fetch the ts and dur of the frame corresponding to the above frame_id.
+    const frameWithMaxDurBlockingCallQuery = `
+        SELECT
+          cast_string!(frame_id) as frame_id,
```
LalitMaganti: Please fix indentation.

LalitMaganti: Why not rename to string to make your life below easier?

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -59,6 +60,14 @@ class BlockingCallMetricHandler implements MetricHandler {
     this.pinSingleCuj(ctx, metricData);
     const config = this.blockingCallTrackConfig(metricData);
     addDebugSliceTrack({trace: ctx, ...config});
+    // Only trigger adding track for frame when the aggregation is for max duration per frame.
+    if (metricData.aggregation === 'max_dur_per_frame_ns') {
+      this.frameWithMaxDurBlockingCallTrackConfig(ctx, metricData).then(
```
LalitMaganti: Please just await instead of doing .then

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Done.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Sorry did not understand this (difficult to understand what code snippet is the comment related to).

LalitMaganti: Never mind, was a confusing comment and not very useful.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: To use await, I will have to make the addMetricTrack() function async. As it is an interface method, will have to make this change in all the classes that implement this method.

And the block only has one level of 'then' callback (callback hell not too bad).

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Done

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: To use await, I will have to make the addMetricTrack function async. As this is an interface method, will have to make the change in all classes implementing this method.

And the block has just one level of 'then' callback (callback hell not too bad). LMK if this change is really required.

LalitMaganti: In general trying to hide the "asyncness" of the function always leads to long term pain. Maybe it will work fine for now but I can promise doing the work to plumb everything properly will save folks *significant* time in the future when they need to debug this code.

I would strongly suggest you do the plumbing properly now and save yourself the trouble in the future.

devianb: Done.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +112,50 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+
+    return {
+      data: {
+        sqlSource: frameWithMaxDurBlockingCallQuery,
+        columns: ['frame_id', 'ts', 'dur'],
+      },
+      columns: {ts: 'ts', dur: 'dur', name: 'frame_id'},
+      argColumns: ['frame_id', 'ts', 'dur'],
+      title: 'Frame with max duration blocking call',
+    };
```
nicomazz, line 180: nit: consider extracting in a method called "pinFrame(title)" to make this method more clear

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +112,50 @@ class BlockingCallMetricHandler implements MetricHandler {
       title: trackName,
     };
   }
+
+  private async frameWithMaxDurBlockingCallTrackConfig(
+    ctx: Trace,
+    metricData: BlockingCallMetricData,
+  ) {
```
nicomazz: no return type here?

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +112,50 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+        frame_id
+      FROM _blocking_calls_frame_cuj
+      WHERE
+        process_name = "${processName}"
+        AND name = "${blockingCallName}"
+        AND cuj_name = "${cuj}"
+      -- select frame_id for the metric with the maximum duration.
+      ORDER BY dur DESC
+      LIMIT 1`);
+    const row = result.firstRow({frame_id: LONG});
```
nicomazz: does it make sense to check if this was sucessful and give some meaningful error message if not?

nicomazz (approved):

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: The actual pinning happens in the 'addDebugSliceTrack' method. But I moved the query frame with max duration blocking call to a separate function.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Added.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Added a custom Error. Throwing it with a message when no frame_id found. @LalitMaganti Please advise if this is a right way to do it.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +115,79 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+    return ctx.engine.query(`
+      INCLUDE PERFETTO MODULE android.frame_blocking_calls.blocking_calls_aggregation;
+
+      SELECT
+        frame_id
+      FROM _blocking_calls_frame_cuj
+      WHERE
+        process_name = "${processName}"
+        AND name = "${blockingCallName}"
+        AND cuj_name = "${cuj}"
```
LalitMaganti: Please use single quotes.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +115,79 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+    let row: Row = {
+      frame_id: null,
+    };
+
+    try {
+      row = (
+        await this.getFrameIdWithMaxDurationBlockingCall(ctx, metricData)
+      ).firstRow({frame_id: LONG});
+    } catch (e) {
+      if (e instanceof Error) {
```
LalitMaganti: if is not necessary.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +115,79 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+        throw new QueryResultError(
+          'No frame found for the given process, CUJ and blocking call.',
+        );
```
LalitMaganti: I would not bother having a subclass, just throw an error directly with the changed message. Better yet included a "caused by" and print the underlying error as well.

devianb: You mean catch and re-throw an Error with a different message?

LalitMaganti: Yes exactly.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Done

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Removed.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: Done.

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
```diff
@@ -103,6 +115,70 @@ class BlockingCallMetricHandler implements MetricHandler {
...
+      frame_id: null,
+    };
+
+    try {
+      row = (
+        await this.getFrameIdWithMaxDurationBlockingCall(ctx, metricData)
+      ).firstRow({frame_id: LONG});
+    } catch (e) {
+      throw new Error(
+        `${e.message} caused by: No frame found for the given process, CUJ and blocking call.`,
```
nicomazz, line 158: probably you wanted to reverse the order of this string?  like `No frame found for the given process, CUJ and blocking call. Caused by: ...`

nicomazz (approved):

ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/handlers/pinBlockingCall.ts
devianb: No actually. The error message that gets printed right now is this:

Failed assertion caused by: No frame found for the given process, CUJ and blocking call.
...

PR #2219, opened 2025-07-21
traced_perf: Allow CPU 0 to be excluded when offline

jhs2-lee:
Hello Perfetto Maintainers,

I'm work at Samsung to CPU analysis, both hardware and software aspects. 
I have been actively using Perfetto and Tracebox and find them very helpful.

Recently in our team, we collect PMU (Performance Monitoring Unit) data per CPU under various scenarios.
During data collection, we observed that if CPU 0 is offline, the PMU data does not get properly collected.

We often analyze only one type of core while some of the CPUs among multiple cores are offline.
e.g.) online - only middle tier core, offline - others core(big, little, etc...)

This has caused difficulties for us, which led to creating a patch resolve this issue.

I would appreciate it if you could review the patch and consider merge it. 
Please feel free to share any other ideas or suggestions.

Thank you.

### DIFF
--- a/src/profiling/perf/perf_producer.cc
+++ b/src/profiling/perf/perf_producer.cc
@@ -110,8 +110,8 @@ std::vector<uint32_t> CreateCpuMask(const protos::gen::PerfEventConfig& cfg) {
     // check explicit mask from cfg, or allow all by default
     if (!target_cpus.empty() && target_cpus.count(cpu) == 0)
       continue;
-    // consider cpu0 to always be online
-    if (cpu > 0 && !IsCpuOnline(cpu))
+
+    if (!IsCpuOnline(cpu))
       continue;
     ret.push_back(cpu);
   }


LalitMaganti:
Thanks for the patch. Sending to Ryan who is our perf profiling expert.

rsavitski (approved):

LalitMaganti (approved):

jhs2-lee:
The original commit message is as follows:

Before, Perfetto always included CPU 0 as online — even if the system
reported it as offline. This was based on the idea that CPU 0 is often
special in the Linux kernel and usually stays online.

But in some systems, CPU 0 can actually be offline, and forcing it to
appear online can confuse certain analyses. For example, in some power
tests or when analyzing only a subset of CPUs, it's useful to treat CPU 0
just like any other core.

This change lets users exclude CPU 0 if it’s really offline, giving
more accurate results in advanced or custom setups. If CPU 0 is online
(as it usually is), the behavior stays the same.

We're not removing the old assumption — just making it flexible.

Test: Run tools/run_presubmit, Local build, Run tracebox in Android devices
Signed-off-by: JeongHyeon Lee <jhs2.lee@samsung.com>

jhs2-lee:
@LalitMaganti @rsavitski 
Thank you for review and merge it :)

PR #2226, opened 2025-07-21
Replace the test trace for missing_track_names with a simpler one

khokhlov962:
Instead of using a manually edited chrome trace, we generate a small simple trace with no track names to test that missing process/thread names don't break the UI.


khokhlov962:
Friendly ping :)

PR #2243, opened 2025-07-22
tp: add new murmurhash implementation

LalitMaganti:
FNV1A is not cutting it:
* we're spending a lot of time in HashString when parsing traces
* our hashmaps are doing poorly because of collisisons because it's not
  high quality enough
* HLL was failing tests likely because of hash function diffusion not
  being strong enough.

This hash function is murmurhash inspired (it's most heavily based on
implementation in duckdb) and should be a lot better interms of both
performance (as it operates 8 bytes at a time) and quality (as
murmurhash is a much better hash function in quality)

### DIFF
--- a/src/trace_processor/util/murmur_hash.h
+++ b/src/trace_processor/util/murmur_hash.h
@@ -23,6 +23,7 @@
 #include <cstring>
 #include <limits>
 #include <string>
+
 #include "perfetto/public/compiler.h"
 
 // This file implements a 64-bit variant of the MurmurHash algorithm.
--- a/src/trace_processor/util/murmur_hash.h
+++ b/src/trace_processor/util/murmur_hash.h
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#ifndef SRC_TRACE_PROCESSOR_UTIL_MURMURHASH_H_
-#define SRC_TRACE_PROCESSOR_UTIL_MURMURHASH_H_
+#ifndef SRC_TRACE_PROCESSOR_UTIL_MURMUR_HASH_H_
+#define SRC_TRACE_PROCESSOR_UTIL_MURMUR_HASH_H_
 
 #include <cmath>
 #include <cstddef>
@@ -164,4 +164,4 @@ inline uint64_t MurmurHash(const std::string& res) {
 
 }  // namespace perfetto::trace_processor::util
 
-#endif  // SRC_TRACE_PROCESSOR_UTIL_MURMURHASH_H_
+#endif  // SRC_TRACE_PROCESSOR_UTIL_MURMUR_HASH_H_
--- a/src/trace_processor/util/murmur_hash.h
+++ b/src/trace_processor/util/murmur_hash.h
@@ -1,0 +1,163 @@
+/*
+ * Copyright (C) 2019 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


aMayzner (approved):

src/trace_processor/util/murmur_hash.h
```diff
@@ -0,0 +1,168 @@
...
+//   MurmurHash3.
+//
+// - vs. MurmurHash3: It uses the exact same `fmix64` finalization function but
+//   substitutes MurmurHash3's complex, rotation-heavy body loop with a much
+//   simpler and faster one.
+//
+// In summary, it makes a performance-oriented trade-off: a simpler main loop
+// combined with a high-quality final mixing stage.
+//
+// ⚠️ NOTE: This implementation is NOT cryptographically secure. It must not be
```
primiano, line 52: do we really need the emoji in the source code? :)

LalitMaganti: I'll remove this in a followup.

LalitMaganti:
> LGTM given the comment (thanks). Only one quesiton: isn't this going to invalidate / change the UUIDs based on hash? Or are you keeping them with FNV1-a?

What do you mean by UUIDs based on hash? Where do we create UUIDs based on hash of something? I'm not sure I follow.

I ran some tests where I changed the hash used in StringPool to MurmurHash and it had a good ~5-10% improvement in parsing of a large trace.

I anticipate if I was to change base::FlatHashMap to rely on this, I would another decent bump, especially because not only is the hashing faster but also the FlatHashMap should have better Find/Insert performance because of better diffusion of the hash (especially in things like tids).

To that end I was considering making base::Hash -> base::FnvHash and then making this new one base::Hash. Then do a full replace of the codebase to use this. Any old ones which strictly need FNV we could keep in the old one but I don't anticpate many/any of them.

Thoughts on this? Any problems you anticipate?

primiano:
> What do you mean by UUIDs based on hash? Where do we create UUIDs based on hash of something? I'm not sure I follow.

As discussed offline, there's some code in TP which computes a fallback UUID based on hash(first X KB, length) if a UUID is not present. that should stay on FNV-1 to avoid invalidating UUIDs (it's mostly for other trace types that don't have a UUID themselves)

> I ran some tests where I changed the hash used in StringPool to MurmurHash and it had a good ~5-10% improvement in parsing of a large trace.

Nice


> I anticipate if I was to change base::FlatHashMap to rely on this, I would another decent bump, especially because not only is the hashing faster but also the FlatHashMap should have better Find/Insert performance because of better diffusion of the hash (especially in things like tids).

+1 to this, I don't think anybody relies on the hashing used internally by FHM.
Maybe we shoudl put it behind a (Android)flag though and go through that process as, there is a chance that could have some unexpected bad effect, and if it does, we don't want to propagate immediately to all branches.

> To that end I was considering making base::Hash -> base::FnvHash and then making this new one base::Hash. Then do a full replace of the codebase to use this. Any old ones which strictly need FNV we could keep in the old one but I don't anticpate many/any of them.

Hmm that is riskier, imho medium risk, and low gain. 
The risk is that we ended up accidentally depending somewhere on the stability of that specific hash, and changing it might break some ABI. I'd much rather.

- Rename base::Hash -> base::FnvHash
- For now keep aliasing base::Hash to FnvHash
- Introduce base::MurmurHash   (or if you prefer, make base::Hash take a template with the hasher, if possible)
- move manually use cases to base::MurmurHash after audiding them one by one.
- To keep it flagged we could  create some class like base::TransitionHash, which decides which hasher to use behind a flag (so we can roll back the flag in case of problems)

LalitMaganti:
> The risk is that we ended up accidentally depending somewhere on the stability of that specific hash, and changing it might break some ABI. I'd much rather.

Ack I was somewhat worried about this as well. It seems to be used in some places in protozero filtering actually which would definitely break ABI. Also in track_event where it has the potential to.

I like the base::FnvHash, base::MurmurHash plan. But to confirm, you have no problem with me just changing the default hash of base::FlatHashMap to base::MurmurHash right?

primiano:
>But to confirm, you have no problem with me just changing the default hash of base::FlatHashMap to base::MurmurHash right?

correct as long as it's behind a flag (in case causes some weird new pathological case).
We can use the same flag for the migration of other classes later

PR #2248, opened 2025-07-22
Resolve sandboxed tids to synthetic ones at parsing time

betasheet:
By resolving to synthetic tids at parsing time, we pave the way for workarounds to enable backwards-compat for older Chrome Linux traces.

See https://github.com/google/perfetto/pull/2187 and crbug.com/432057400 for context.


src/trace_processor/importers/common/BUILD.gn
```diff
@@ -123,6 +122,32 @@ source_set("parser_types") {
   ]
 }
 
+source_set("args_tracker_hdr") {
```
betasheet: @LalitMaganti I let Roo code fix the dependency issues -- but maybe there's a better "utils" place or similar to move the CreateSyntheticTid helper to?...

LalitMaganti: I don't understand why you need this at at all. Any importer code in importers/proto can depend on code in impoters/common. Why do you need to split this at all instead of just depending on common directly?

betasheet: common depends on packet_sequence_state_generation_hdr currently -- adding common to its dependencies would then create a circular dep.

betasheet: (existing dep: common -> parser_types -> packet_sequence_state_generation_hdr)

LalitMaganti: Please add a new header in the common folder (maybe called synthetic_tid.h) or similar and put the two static functions there.

betasheet:
PTAL :)

src/trace_processor/importers/common/synthetic_tid.h
```diff
@@ -0,0 +1,40 @@
...
+
+// Returns a unique synthetic tid based on the provided |tid| and |pid|.
+// This is useful in systems where the pid is unique but the tid is not.
+// The returned synthetic tid has the following format:
+// [PID(high 32-bits)][TID(low 32-bits)]
+inline int64_t CreateSyntheticTid(int64_t tid, int64_t pid) {
+  return (pid << 32) | tid;
+}
+
+inline bool IsSyntheticTid(int64_t tid) {
```
LalitMaganti, line 35: Please add a comment.

PR #2253, opened 2025-07-22
Templated helpers to decode messages with extensions

JoeNotCharlesGoogle:
GetFieldWithMetadata() and DecodeFieldWithMetadata() will find a non-repeating field based on a generated FieldMetadata struct. These are basically wrappers around ProtoDecoder::FindField() but add static asserts that the FieldMetadata is from the expected message and has the expected type.

CppMessageDecoder is a wrapper around ProtoDecoder that owns the serialized message bytes. It can be used to convert a CppMessageObj from the "cpp" plugin, which does not support extensions at all, to a pbzero decoder.

### DIFF
--- a/include/perfetto/protozero/proto_decoder.h
+++ b/include/perfetto/protozero/proto_decoder.h
@@ -26,7 +26,6 @@
 
 #include "perfetto/base/compiler.h"
 #include "perfetto/base/logging.h"
-#include "perfetto/protozero/cpp_message_obj.h"
 #include "perfetto/protozero/field.h"
 #include "perfetto/protozero/proto_utils.h"
 
@@ -468,37 +467,6 @@ class TypedProtoDecoder : public TypedProtoDecoderBase {
   Field on_stack_storage_[PROTOZERO_DECODER_INITIAL_STACK_CAPACITY];
 };
 
-// A wrapper to convert a "cpp" generated message to a generic "pbzero" decoder.
-// This requires re-serializing the message so has some overhead, but is useful
-// for messages with extension fields that aren't supported by the "cpp"
-// generator.
-template <typename DecoderType>
-class CppMessageDecoder {
- public:
-  static_assert(std::is_base_of_v<ProtoDecoder, DecoderType>,
-                "DecoderType must inherit ProtoDecoder");
-
-  using Decoder = DecoderType;
-
-  explicit CppMessageDecoder(const CppMessageObj& msg)
-      : bytes_(msg.SerializeAsArray()),
-        decoder_(bytes_.data(), bytes_.size()) {}
-
-  ~CppMessageDecoder() = default;
-
-  // Not copyable or movable because the wrapped decoder exposes pointers into
-  // `bytes_`.
-  CppMessageDecoder(const CppMessageDecoder&) = delete;
-  CppMessageDecoder& operator=(const CppMessageDecoder&) = delete;
-
-  Decoder& decoder() { return decoder_; }
-  const Decoder& decoder() const { return decoder_; }
-
- private:
-  std::vector<uint8_t> bytes_;
-  Decoder decoder_;
-};
-
 // Returns the field from `decoder` matching the given `metadata`. Doesn't
 // affect the decoder's read cursor. This is useful for extension fields that
 // have generated FieldMetadata but no generated accessors. Prefer this to
@@ -515,17 +483,6 @@ Field GetFieldWithMetadata(typename MessageType::Decoder& decoder,
       Metadata::kRepetitionType == proto_utils::RepetitionType::kNotRepeated,
       "repeated fields aren't supported");
   return decoder.FindField(metadata.kFieldId);
-}
-
-// Returns the field from `msg` matching the given `metadata`. This is useful
-// for extension fields that have generated FieldMetadata but no generated
-// accessors. However it requires re-serializing `msg`, so to read multiple
-// fields prefer to create one CppMessageDecoder and reuse it.
-template <typename MessageType, typename Metadata>
-Field GetFieldWithMetadata(const CppMessageObj& msg, const Metadata& metadata) {
-  return GetFieldWithMetadata<MessageType, Metadata>(
-      CppMessageDecoder<typename MessageType::Decoder>(msg).decoder(),
-      metadata);
 }
 
 // Returns a nested message decoder for the field from `decoder` matching the
@@ -546,21 +503,6 @@ std::optional<FieldDecoder> DecodeFieldWithMetadata(
     return std::nullopt;
   }
   return FieldDecoder(field.as_bytes());
-}
-
-// Returns a nested message decoder for the field from `msg` matching the given
-// `metadata`, or nullopt if no such field is found. This is useful for
-// extension fields that have generated FieldMetadata but no generated
-// accessors. However it requires re-serializing `msg`, so to read multiple
-// fields prefer to create one CppMessageDecoder and reuse it.
-template <typename MessageType,
-          typename Metadata,
-          typename FieldDecoder = typename Metadata::cpp_field_type::Decoder>
-std::optional<FieldDecoder> DecodeFieldWithMetadata(const CppMessageObj& msg,
-                                                    const Metadata& metadata) {
-  return DecodeFieldWithMetadata<MessageType, Metadata, FieldDecoder>(
-      CppMessageDecoder<typename MessageType::Decoder>(msg).decoder(),
-      metadata);
 }
 
 }  // namespace protozero
--- a/src/protozero/proto_decoder_unittest.cc
+++ b/src/protozero/proto_decoder_unittest.cc
@@ -666,74 +666,10 @@ TEST(ProtoDecoderTest, RepeatedMaxFieldIdStack) {
   }
 }
 
-TEST(ProtoDecoderTest, CppMessageDecoder) {
-  std::vector<uint8_t> proto = CreateExtensionMessage();
-
-  pbgen::LegacyFakeEvent legacy_msg;
-  ASSERT_TRUE(legacy_msg.ParseFromArray(proto.data(), proto.size()));
-  EXPECT_TRUE(legacy_msg.has_base_int());
-  EXPECT_EQ(legacy_msg.base_int(), 100u);
-  EXPECT_TRUE(legacy_msg.has_extension_a());
-  EXPECT_TRUE(legacy_msg.extension_a().has_int_a());
-  EXPECT_EQ(legacy_msg.extension_a().int_a(), 101u);
-  EXPECT_FALSE(legacy_msg.has_extension_b());
-
-  // Convert to protozero LegacyFakeEvent.
-  CppMessageDecoder<pbtest::LegacyFakeEvent::Decoder> legacy_decoder(
-      legacy_msg);
-  EXPECT_TRUE(legacy_decoder.decoder().has_base_int());
-  EXPECT_EQ(legacy_decoder.decoder().base_int(), 100u);
-  EXPECT_TRUE(legacy_decoder.decoder().has_extension_a());
-  pbtest::SystemA::Decoder legacy_decoder_a(
-      legacy_decoder.decoder().extension_a());
-  EXPECT_TRUE(legacy_decoder_a.has_int_a());
-  EXPECT_EQ(legacy_decoder_a.int_a(), 101u);
-  EXPECT_FALSE(legacy_decoder.decoder().has_extension_b());
-
-  // Convert to protozero RealFakeEvent. This should work because it has the
-  // same wire format as LegacyFakeEvent.
-  CppMessageDecoder<pbtest::RealFakeEvent::Decoder> real_decoder(legacy_msg);
-  EXPECT_TRUE(real_decoder.decoder().has_base_int());
-  EXPECT_EQ(real_decoder.decoder().base_int(), 100u);
-
-  // BrowserExtension inherits RealFakeEvent's decoder, so can't be used to
-  // decode extension fields. They should still be available as unknown fields.
-  // TODO(crbug.com/433549151): Implement decoders for extension messages in the
-  // pbzero generator.
-  static_assert(std::is_same_v<pbtest::RealFakeEvent::Decoder,
-                               pbtest::BrowserExtension::Decoder>,
-                "BrowserExtension::Decoder implementation has changed. Update "
-                "the unit test.");
-
-  protozero::Field extension_field_a = real_decoder.decoder().FindField(10);
-  ASSERT_TRUE(extension_field_a.valid());
-  ASSERT_EQ(extension_field_a.type(),
-            proto_utils::ProtoWireType::kLengthDelimited);
-  pbtest::SystemA::Decoder extension_decoder_a(extension_field_a.as_bytes());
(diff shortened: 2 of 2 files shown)


LalitMaganti:
Can you explain how you are planning on using this? Is this something which is going to be used in trace processor? cpp is banned in trace processor, we will definitely not be allowing any use of this there.

JoeNotCharlesGoogle:
CppMessageDecoder is mainly to avoid boilerplate in unit tests. I'll remove it from this patch and put it in the test instead.

GetFieldWithMetadata (not touching cpp at all) will be used in TrackEventParser::ParseTrackDescriptor():

```
   } else if (decoder.has_process()) {
     UniquePid upid =
         ParseProcessDescriptor(packet_timestamp, decoder.process());
-    if (decoder.has_chrome_process()) {
-      ParseChromeProcessDescriptor(upid, decoder.chrome_process());
+    // `chrome_process` is a field in the ChromeTrackDescriptor extension.
+    if (protozero::Field chrome_process =
+            protozero::GetFieldWithMetadata<protos::pbzero::TrackDescriptor>(
+                decoder,
+                protos::pbzero::ChromeTrackDescriptor::kChromeProcess)) {
+      ParseChromeProcessDescriptor(upid, chrome_process.as_bytes());
     }
   }
```

LalitMaganti:
> CppMessageDecoder is mainly to avoid boilerplate in unit tests. I'll remove it from this patch and put it in the test instead.

Yeah that would be good thanks. We don't people to accidentally start depending on it, protozero is used in a *lot* of places, any new function there adds a big burden on us.

> GetFieldWithMetadata (not touching cpp at all) will be used in TrackEventParser::ParseTrackDescriptor():

Conceptually fine with that but performance is a bit of a concern. FindField is an O(n) function. Adding these in several places in the wrong place in TP will tank performance. And it's a very tempting function to use without realising the O(n) cost you are imposing.

JoeNotCharlesGoogle:
Unfortunately I don't see any way to access extension fields without using FindField, since the only O(1) accessors go through the `fields_` array which isn't populated with extension fields.

What do you think of renaming this to `GetExtensionField` to emphasize that it's only for extensions? I started writing it that way but made it more generic when I realized that `GetByMetadata` would work for non-extension fields too, and was a bit safer than FindField. But performance is a good reason not to expose it more widely.

LalitMaganti:
I see two options:
1) You name the function clearly (e.g. with a Slow suffix) to emphasise that the cost of this is high so it can be queried in review. I would also maybe make a presubmit check for the count of this in the codebase (we do this for the UI, I think we can adapt the UI check quite easily to trace processor as well)
2) We create an equivalent of TypedProtoDecoder which works for extensions. Maybe we specify all the extension fields as a tuple when we construct the decoder?

JoeNotCharlesGoogle:
In the meantime I've put my new ChromeProcessDescriptor extension up using FindField directly (https://github.com/google/perfetto/pull/2256).

ddiproietto:
I'm working on something slightly related and I thought I could share some ideas

> Unfortunately I don't see any way to access extension fields without using FindField, since the only O(1) accessors go through the fields_ array which isn't populated with extension fields.

Please correct me if I'm wrong, but I believe they should work. Since https://r.android.com/3055565, we also consider the extensions number when computing the max ids in the fields.

You should be able to use `Decoder::at<FieldNumber>()` to get the field that you're interested in

Is there something I'm missing?

JoeNotCharlesGoogle:
> Please correct me if I'm wrong, but I believe they should work. Since https://r.android.com/3055565, we also consider the extensions number when computing the max ids in the fields.

I think that's only for setters. pbzero generates associated Decoder classes with getters, and those don't include extension fields.

The generated protos/perfetto/trace/track_event/track_event.pbzero.h contains:

```
class TrackEvent_Decoder : public ::protozero::TypedProtoDecoder</*MAX_FIELD_ID=*/54, /*HAS_NONPACKED_REPEATED_FIELDS=*/true> {

...

class TrackEvent : public ::protozero::Message {
 public:
  using Decoder = TrackEvent_Decoder;
```

54 is `uint64 correlation_id_str_iid = 54` in track_event.proto. It definitely doesn't include either the `extensions` declarations in track_event.proto (1000 to 10000) or the actual extensions used in chrome_track_event.proto (1000 to 1076).

The generated protos/third_party/chromium/chrome_track_event.pbzero.h has a ChromeTrackEvent class with no associated Decoder:

```
class ChromeTrackEvent : public ::perfetto::protos::pbzero::TrackEvent {
 public:
  ... setter definitions, but nothing else ...
```

So `ChromeTrackEvent::Decoder` resolves to the inherited `TrackEvent::Decoder` with MAX_FIELD_ID 54.

One option would be to just generate a TypedProtoDecoder for the extension messages, so `ChromeTrackEvent::Decoder` would have MAX_FIELD_ID 1076. It would probably want to include all the inherited TrackEvent getters too, though, which might be awkward to generate. And there would always be ~1000 entries of wasted space in the array, which might not be a huge concern.

I'm hoping not to have to learn how to modify the generators though. :-)

ddiproietto:
Ooohh, I see, your extension field id is 1000, which is greater than 999.

We ignore all the fields (both extensions and non extensions), whose id is greater than 999, exactly for the reason you stated (we don't want to make the structure too big).

https://github.com/google/perfetto/blob/f8dd57464dc2be889646269c48576f2758d80c6c/src/protozero/protoc_plugin/protozero_plugin.cc#L53-L56

https://github.com/google/perfetto/blob/f8dd57464dc2be889646269c48576f2758d80c6c/src/protozero/protoc_plugin/protozero_plugin.cc#L767-L783

Otherwise you could have simply used .at directly in the base decoder (TrackEvent::Decoder).

I found a way to make this a little bit simpler, and use the `at` fast path if the field id is small, something like this:

```
diff --git a/include/perfetto/protozero/proto_decoder.h b/include/perfetto/protozero/proto_decoder.h
index 6fa31b6ff0..4f95dd0344 100644
--- a/include/perfetto/protozero/proto_decoder.h
+++ b/include/perfetto/protozero/proto_decoder.h
@@ -452,6 +452,17 @@ class TypedProtoDecoder : public TypedProtoDecoderBase {
     }
   }
 
+  template <uint32_t FIELD_ID>
+  const Field& GetOrFindFieldById() const {
+    static_assert(FIELD_ID <= MAX_FIELD_ID, "FIELD_ID > MAX_FIELD_ID");
+    if constexpr (FIELD_ID <= MAX_FIELD_ID) {
+      return at<FIELD_ID>();
+    } else {
+      // Slow
+      return FindField(FIELD_ID);
+    }
+  }
+
   TypedProtoDecoder(TypedProtoDecoder&& other) noexcept
       : TypedProtoDecoderBase(std::move(other)) {
     // If the moved-from decoder was using on-stack storage, we need to update
@@ -471,34 +482,34 @@ class TypedProtoDecoder : public TypedProtoDecoderBase {
 // affect the decoder's read cursor. This is useful for extension fields that
 // have generated FieldMetadata but no generated accessors. Prefer this to
 // calling ProtoDecoder::FindField() directly because it has more safety checks.
-template <typename MessageType, typename Metadata>
-Field GetFieldWithMetadata(typename MessageType::Decoder& decoder,
-                           const Metadata& metadata) {
+template <typename Decoder, typename Metadata>
+Field GetFieldWithMetadata(Decoder& decoder,
+                           const Metadata&) {
   static_assert(std::is_base_of_v<proto_utils::FieldMetadataBase, Metadata>,
                 "`metadata` must be a FieldMetadata object");
   static_assert(
-      std::is_base_of_v<MessageType, typename Metadata::message_type>,
+      std::is_base_of_v<typename Decoder::Message, typename Metadata::message_type>,
       "`metadata` must come from a subclass of `decoder`'s message type");
   static_assert(
       Metadata::kRepetitionType == proto_utils::RepetitionType::kNotRepeated,
       "repeated fields aren't supported");
-  return decoder.FindField(metadata.kFieldId);
+  return decoder.template GetOrFindFieldById<Metadata::kFieldId>();
 }
 
 // Returns a nested message decoder for the field from `decoder` matching the
 // given `metadata`, or nullopt if no such field is found. Doesn't affect the
 // decoder's read cursor. This is useful for extension fields that have
 // generated FieldMetadata but no generated accessors.
-template <typename MessageType,
+template <typename Decoder,
           typename Metadata,
           typename FieldDecoder = typename Metadata::cpp_field_type::Decoder>
 std::optional<FieldDecoder> DecodeFieldWithMetadata(
-    typename MessageType::Decoder& decoder,
+    Decoder& decoder,
     const Metadata& metadata) {
   static_assert(
       Metadata::kProtoFieldType == proto_utils::ProtoSchemaType::kMessage,
       "DecodeFieldWithMetadata only supports Message fields");
-  Field field = GetFieldWithMetadata<MessageType, Metadata>(decoder, metadata);
+  Field field = GetFieldWithMetadata<Decoder, Metadata>(decoder, metadata);
   if (!field.valid()) {
     return std::nullopt;
   }
diff --git a/src/protozero/proto_decoder_unittest.cc b/src/protozero/proto_decoder_unittest.cc
index 28be9325d6..faf1e1aceb 100644
--- a/src/protozero/proto_decoder_unittest.cc
+++ b/src/protozero/proto_decoder_unittest.cc
@@ -677,11 +677,11 @@ TEST(ProtoDecoderTest, GetFieldWithMetadata) {
   EXPECT_TRUE(legacy_decoder.has_extension_a());
   EXPECT_FALSE(legacy_decoder.has_extension_b());
 
-  EXPECT_TRUE(protozero::GetFieldWithMetadata<pbtest::LegacyFakeEvent>(
+  EXPECT_TRUE(protozero::GetFieldWithMetadata(
       legacy_decoder, pbtest::LegacyFakeEvent::kBaseInt));
-  EXPECT_TRUE(protozero::GetFieldWithMetadata<pbtest::LegacyFakeEvent>(
+  EXPECT_TRUE(protozero::GetFieldWithMetadata(
       legacy_decoder, pbtest::LegacyFakeEvent::kExtensionA));
-  EXPECT_FALSE(protozero::GetFieldWithMetadata<pbtest::LegacyFakeEvent>(
+  EXPECT_FALSE(protozero::GetFieldWithMetadata(
       legacy_decoder, pbtest::LegacyFakeEvent::kExtensionB));
 
   // Interpret message as RealFakeRevent. `extension_a` and `extension_b` can
@@ -689,11 +689,11 @@ TEST(ProtoDecoderTest, GetFieldWithMetadata) {
   pbtest::RealFakeEvent::Decoder real_decoder(proto.data(), proto.size());
   EXPECT_TRUE(real_decoder.has_base_int());
 
-  EXPECT_TRUE(protozero::GetFieldWithMetadata<pbtest::RealFakeEvent>(
-      real_decoder, pbtest::RealFakeEvent::kBaseInt));
-  EXPECT_TRUE(protozero::GetFieldWithMetadata<pbtest::RealFakeEvent>(
+  EXPECT_TRUE(protozero::GetFieldWithMetadata(real_decoder,
+                                              pbtest::RealFakeEvent::kBaseInt));
+  EXPECT_TRUE(protozero::GetFieldWithMetadata(
       real_decoder, pbtest::BrowserExtension::kExtensionA));
-  EXPECT_FALSE(protozero::GetFieldWithMetadata<pbtest::RealFakeEvent>(
+  EXPECT_FALSE(protozero::GetFieldWithMetadata(
       real_decoder, pbtest::BrowserExtension::kExtensionB));
 }
 
@@ -708,15 +708,15 @@ TEST(ProtoDecoderTest, DecodeFieldWithMetadata) {
   EXPECT_FALSE(legacy_decoder.has_extension_b());
 
   std::optional<pbtest::SystemA::Decoder> legacy_decoder_a =
-      protozero::DecodeFieldWithMetadata<pbtest::LegacyFakeEvent>(
-          legacy_decoder, pbtest::LegacyFakeEvent::kExtensionA);
+      protozero::DecodeFieldWithMetadata(legacy_decoder,
+                                         pbtest::LegacyFakeEvent::kExtensionA);
   ASSERT_TRUE(legacy_decoder_a.has_value());
   EXPECT_TRUE(legacy_decoder_a->has_int_a());
   EXPECT_EQ(legacy_decoder_a->int_a(), 101u);
 
   std::optional<pbtest::SystemB::Decoder> legacy_decoder_b =
-      protozero::DecodeFieldWithMetadata<pbtest::LegacyFakeEvent>(
-          legacy_decoder, pbtest::LegacyFakeEvent::kExtensionB);
+      protozero::DecodeFieldWithMetadata(legacy_decoder,
+                                         pbtest::LegacyFakeEvent::kExtensionB);
   EXPECT_FALSE(legacy_decoder_b.has_value());
 
   // Interpret message as RealFakeRevent. `extension_a` and `extension_b` can
@@ -724,15 +724,15 @@ TEST(ProtoDecoderTest, DecodeFieldWithMetadata) {
   pbtest::RealFakeEvent::Decoder real_decoder(proto.data(), proto.size());
 
   std::optional<pbtest::SystemA::Decoder> extension_decoder_a =
-      protozero::DecodeFieldWithMetadata<pbtest::RealFakeEvent>(
-          real_decoder, pbtest::BrowserExtension::kExtensionA);
+      protozero::DecodeFieldWithMetadata(real_decoder,
+                                         pbtest::BrowserExtension::kExtensionA);
   ASSERT_TRUE(extension_decoder_a.has_value());
   EXPECT_TRUE(extension_decoder_a->has_int_a());
   EXPECT_EQ(extension_decoder_a->int_a(), 101u);
 
   std::optional<pbtest::SystemB::Decoder> extension_decoder_b =
-      protozero::DecodeFieldWithMetadata<pbtest::RealFakeEvent>(
-          real_decoder, pbtest::BrowserExtension::kExtensionB);
+      protozero::DecodeFieldWithMetadata(real_decoder,
+                                         pbtest::BrowserExtension::kExtensionB);
   EXPECT_FALSE(extension_decoder_b.has_value());
 }
 
diff --git a/src/protozero/protoc_plugin/protozero_plugin.cc b/src/protozero/protoc_plugin/protozero_plugin.cc
index f9132edfdc..26262fda03 100644
--- a/src/protozero/protoc_plugin/protozero_plugin.cc
+++ b/src/protozero/protoc_plugin/protozero_plugin.cc
@@ -791,6 +791,9 @@ case $full_class$::$value_name$:
         has_nonpacked_repeated_fields ? "true" : "false");
     stub_h_->Print(" public:\n");
     stub_h_->Indent();
+    stub_h_->Print(
+        "using Message = $name$;\n",
+        "name", GetCppClassName(message));
     stub_h_->Print(
         "$name$(const uint8_t* data, size_t len) "
         ": TypedProtoDecoder(data, len) {}\n",
@@ -935,6 +938,9 @@ case $full_class$::$value_name$:
   }
 
   void GenerateMessageDescriptor(const Descriptor* message) {
+    stub_h_->Print(
+        "class $name$;\n\n",
+        "name", GetCppClassName(message));
     GenerateDecoder(message);
 
     stub_h_->Print(

```

But the fundamental problem remains, it can be a slow operation, if you're querying a lot of fields.

JoeNotCharlesGoogle:
After Lalit's earlier comments about slow operations, I don't like having the automatic fallback. Better to have explicit fast ops and slow ops.

How about adding a MIN_FIELD_ID to TypedProtoDecoder, and subtracting that from each field id to get an index into `fields_`? That way an extension proto defining fields 1000...1050 would allocate 50 slots.

LalitMaganti:
Closing as this is no longer relevant.

PR #2256, opened 2025-07-22
Move ChromeProcessDescriptor to an extension

JoeNotCharlesGoogle:
Adds an extension range for Chrome to TrackDescriptor, and a ChromeTrackDescriptor extension proto in protos/third_party/chromium. This causes protos/third_party/chromium to diverge from Chrome's base/tracing/protos, due to a dependency cycle between Perfetto and Chrome. Chrome will need to import the changes to
protos/perfetto/trace/track_event and update base/tracing/protos in one commit before the next sync from Chrome -> Perfetto.

Also adds an ExtensionsWireCompatibility test to make sure that converting an inline field to an extension field like this is safe.

Now that ChromeProcessDescriptor is part of an extension, it's moved to a different descriptor file which needs to be included when calling `serialize_python_trace` for diff_tests. Every other `serialize_*_trace` variant already takes a list of extension descriptors so this brings Python in line with other functions.


python/generators/diff_tests/runner.py
```diff
@@ -574,6 +579,9 @@ def run_all_tests(
               summary_descriptor,
               metrics_descriptor_paths,
               [chrome_extensions, test_extensions, winscope_extensions],
+              # Some Python tests need ChromeProcessDescriptor from
```
LalitMaganti: I don't understand what you mean by "syntax errors". How can there be syntax errors by adding extensions?

JoeNotCharlesGoogle: Explained in the pull request description (which will be the commit description):

However if I pass the same list of extension descriptors used by other serialize_*_trace functions ([chrome_extensions, test_extensions, winscope_extensions]), the first two tests fail with a different error, and then the whole test suite seems to lock up:

```
$ python3 tools/diff_test_trace_processor.py  out/Default/trace_processor_shell 
[==========] Running 1062 tests.
/usr/local/google/home/joenotcharles/src/perfetto/test/trace_processor/diff_tests/metrics/startup/android_startup.py:85: SyntaxWarning: invalid escape sequence '\{'
  buf='reportFullyDrawn() for \{com.google.android.calendar\}')
/usr/local/google/home/joenotcharles/src/perfetto/test/trace_processor/diff_tests/metrics/startup/android_startup_slow.py:99: SyntaxWarning: invalid escape sequence '\{'
  buf='reportFullyDrawn() for \{com.google.android.calendar\}')
```

LalitMaganti: We should debug this rather than adding a hack like this. This is a non trivial maintenance cost for us. 

Also I'm not even sure chrome has any python tests anyway so this might be redundant.

JoeNotCharlesGoogle: The diff tests have started failing for me with the above error even when I only add `[chrome_extensions]`. And on the bots they're failing with a different error:

```
[       OK ] AndroidParser:android_system_property_counter (ingest: 43.48 ms query: 0.46 ms)
concurrent.futures.process._RemoteTraceback: 
"""
  File "/usr/lib/python3.9/concurrent/futures/process.py", line 243, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/tmp/github-action-runner/_work/perfetto/perfetto/python/generators/diff_tests/runner.py", line 511, in execute
    result, run_str = self.__run(
  File "/tmp/github-action-runner/_work/perfetto/perfetto/python/generators/diff_tests/runner.py", line 397, in __run
    text_format.Merge(self.test.blueprint.trace.contents, proto)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 754, in Merge
    return MergeLines(
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 828, in MergeLines
    return parser.MergeLines(lines, message)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 853, in MergeLines
    self._ParseOrMerge(lines, message)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 877, in _ParseOrMerge
    self._MergeField(tokenizer, message)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 1012, in _MergeField
    merger(tokenizer, message, field)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 1093, in _MergeMessageField
    self._MergeField(tokenizer, sub_message)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 1012, in _MergeField
    merger(tokenizer, message, field)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 1093, in _MergeMessageField
    self._MergeField(tokenizer, sub_message)
  File "/usr/local/lib/python3.9/dist-packages/google/protobuf/text_format.py", line 973, in _MergeField
    raise tokenizer.ParseErrorPreviousToken(
google.protobuf.text_format.ParseError: 28:13 : Message type "perfetto.protos.TrackDescriptor" has no field named "chrome_process".
"""
```

Neither of those errors were happening for me locally yesterday.

Seems to be a missing build dep, so something isn't being regenerated?

protos/perfetto/trace/track_event/track_descriptor.proto
```diff
@@ -336,4 +337,29 @@ message TrackDescriptor {
...
+
+  // Extension range for typed events defined externally.
+  // See docs/design-docs/extensions.md for more details.
+  //
+  // Extension support is work-in-progress, in the future the way to reserve a
+  // subrange for a particular project will be described here and in the design
+  // document linked above.
+  //
+  // Contact perfetto-dev@googlegroups.com if you are interested in a subrange
+  // for your project.
```
LalitMaganti: This comment seems copy pasted and does not apply here.

protos/perfetto/trace/track_event/track_descriptor.proto
```diff
@@ -336,4 +337,29 @@ message TrackDescriptor {
...
+
+  // Extension range reserved for chromium:
+  // https://source.chromium.org/chromium/chromium/src/+/main:base/tracing/protos/chrome_track_event.proto
+  extensions 6, 1000 to 1999;
+  // Extension range reserved for https://b.corp.google.com/issues/301227627.
```
LalitMaganti: This all seems copy pasted and does not apply here.

python/generators/diff_tests/runner.py
LalitMaganti: If you run a clean build locally the problem should show up. You will need to change all diff test textprotos in the repo [1] to switch `chrome_process` field to use the extension mechanism of definining fields in the textproto [2]

[1] https://source.chromium.org/search?q=chrome_process%20f:diff_test%20f:textproto%20f:%5Ethird_party%2Fperfetto&ss=chromium
[2] https://protobuf.dev/reference/protobuf/textformat-spec/#extension

JoeNotCharlesGoogle: Hah, that SyntaxWarning happens on main too. It just flies past too fast to be seen. The real problem is that the tests become slower, presumably due to the larger size of the descriptor pool. I don't know why it got so much slower with only the chrome extension descriptor today.

LalitMaganti: The tests get much slower if there are Python failures. If you are seeing those then that will be the root cause. As soon as you fix the tests, they should go back to normal.

protos/perfetto/trace/track_event/track_descriptor.proto
JoeNotCharlesGoogle: Are you suggesting we only add the Chrome extension range? I thought it was less confusing for `track_descriptor.proto` to have the same assigned extension ranges as `track_event.proto`, even if most projects don't use their extension ranges.

LalitMaganti: Definitely. In fact, *no one* should be extending TrackDescriptor at all. Chrome only extends it for legacy reasons and did this as a "hack". If this was done properly, all the CHrome specific features would be proper upstreamed features of TrackEvent.

In fact I would love instead of what you were doing that you would kill ChromeProcessDescriptor entirely but I know not to ask for too much :)

python/generators/diff_tests/runner.py
JoeNotCharlesGoogle: Fixed, thanks for the tip.

protos/perfetto/trace/track_event/track_descriptor.proto
JoeNotCharlesGoogle: Ack.

protos/perfetto/trace/track_event/track_descriptor.proto
JoeNotCharlesGoogle: Ack.

LalitMaganti (approved):

PR #2257, opened 2025-07-22
stdlib: Add table for power rail metadata

cphlipot1:
The current table contains one row per sample making it non-ideal for looking information about power rail tracks. In addition it does not have any info about the rail subsystem or other metadata.

Add a new table to make it easy to query this info.

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/power_rails.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/power_rails.sql
@@ -79,3 +79,23 @@ SELECT
 FROM counter_leading_intervals!(counter_table) AS c
 JOIN counter_track AS t
   ON c.track_id = t.id;
+
+-- High level metadata about each of the power rails.
+CREATE PERFETTO TABLE android_power_rails_metadata (
+  -- Power rail name. Alias of `counter_track.name`.
+  power_rail_name STRING,
+  -- Raw power rail name from the hardware.
+  raw_power_rail_name STRING,
+  -- Power rail track id. Alias of `counter_track.id`.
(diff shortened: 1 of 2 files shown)


src/trace_processor/perfetto_sql/stdlib/android/power_rails.sql
```diff
@@ -79,3 +79,22 @@ SELECT
 FROM counter_leading_intervals!(counter_table) AS c
 JOIN counter_track AS t
   ON c.track_id = t.id;
+
+-- High level Information about each of the power rails.
+CREATE PERFETTO TABLE android_power_rails_info (
```
LalitMaganti: I would suggest calling this `metadata` instead of `info` to make clearer

LalitMaganti (approved):

cphlipot1:
@LalitMaganti  I now have push access and along with that it looks like I can now directly add reviewers to PRs as well. Should i add a 2nd review to this PR?

rsavitski:
> @LalitMaganti I now have push access and along with that it looks like I can now directly add reviewers to PRs as well. Should i add a 2nd review to this PR?

FYI Lalit is out for a couple of weeks. And we're still figuring out the exact semantics of committership in relation to the "enforce-reviews" github check.

For the next ~week, please keep getting two reviews from perfetto googlers before merging into `main`.

rsavitski:
FYI I think the following will make the formatter stop complaining:
`tools/format-sql-sources --upstream origin/main`

(It's a bit too late in the day for me to start digging as to why you ended up in this state, sorry.)

stevegolton (approved):

Issue #519, opened 2023-06-28
Can not capture traces on arm64 server which has 256 cpus.


Shanrow:
In the host of arm64 server with 256 cpus

[root@xxxxxx perfetto]# sudo out/linux/tracebox -o trace_file.perfetto-trace --txt -c test/configs/scheduling.cfg
[942.317]          service.cc:253 Started traced, listening on @traced-p-103430 @traced-c-103430
[949.864]           probes.cc:104 Starting traced_probes service
[949.864]  probes_producer.cc:340 Connected to the service
[949.867]   perf_producer.cc:1200 Connected to the service
[949.867]    perfetto_cmd.cc:1051 Connected to the Perfetto traced service, TTL: 10s
[949.867] ng_service_impl.cc:1003 Configured tracing session 1, #sources:2, duration:10000 ms, #buffers:1, total buffer size:20480 KB, total sessions:1, uid:0 session name: ""
[949.909]  probes_producer.cc:140 Ftrace setup (target_buf=1)
[949.909]    ftrace_procfs.cc:445 disabled ftrace in /sys/kernel/debug/tracing/
[964.880] ng_service_impl.cc:1970 FlushAndDisableTracing(1) done, success=0
[969.885] ng_service_impl.cc:1670 Timeout while waiting for ACKs for tracing session 1
[969.885]    perfetto_cmd.cc:1213 Wrote 885 bytes into trace_file.perfetto-trace

scheduling.cfg:

buffers {
  size_kb: 20480
  fill_policy: DISCARD
}

data_sources {
  config {
    name: "linux.ftrace"
    target_buffer: 0
    ftrace_config {
      ftrace_events: "sched_switch"
      ftrace_events: "sched_waking"
      ftrace_events: "sched_wakeup_new"

      ftrace_events: "task_newtask"
      ftrace_events: "task_rename"

      ftrace_events: "sched_process_exec"
      ftrace_events: "sched_process_exit"
      ftrace_events: "sched_process_fork"
      ftrace_events: "sched_process_free"
      ftrace_events: "sched_process_hang"
      ftrace_events: "sched_process_wait"
    }
  }
}

data_sources {
  config {
    name: "linux.process_stats"
    target_buffer: 0
  }
}

duration_ms: 10000

Howerver when i startup a linux guest vm with 128 cpus in the host via Qemu and in the guest linux using the same perftool i can 
capture traces normally.

[...]

PR #2260, opened 2025-07-23
ftrace: ClearPerCpuTrace only for offline CPUs

sashwinbalaji:
Clearing `/sys/kernel/tracing/trace` should clear all per-CPU trace data for online CPUs ([src]( https://github.com/torvalds/linux/blob/master/kernel/trace/trace.c#L4933)). For offline CPUs, trace data might still be pending in their `trace_pipe_raw`. To minimize time taken, only clear the trace for these offline CPUs instead of for all.

Flow:
- Read the list of offline CPUs from /sys/devices/system/cpu/offline.
- Parse the comma-separated list of decimal numbers and ranges, representing the CPU list.(src: [show_cpus_attr](https://github.com/torvalds/linux/blob/89be9a83ccf1f88522317ce02f854f30d6115c41/drivers/base/cpu.c#L215), [bitmap_print_to_pagebuf](https://github.com/torvalds/linux/blob/master/lib/bitmap-str.c#L58)).
- Iterate & clear the buffer of only the offline CPUs.

Fixes: #519
Test: Tested locally by manually turning off some cpus on both linux and android

### DIFF
--- a/src/traced/probes/ftrace/tracefs.cc
+++ b/src/traced/probes/ftrace/tracefs.cc
@@ -463,31 +463,31 @@ bool Tracefs::GetOfflineCpus(std::vector<uint32_t>* offline_cpus) {
   // Source: https://docs.kernel.org/admin-guide/cputopology.html
   std::vector<uint32_t> offline_cpus_tmp;
   for (base::StringSplitter ss(offline_cpus_str, ','); ss.Next();) {
-    std::string offline_cpu_range(ss.cur_token(), ss.cur_token_size());
-    auto dash_pos = offline_cpu_range.find('-');
-    if (dash_pos == std::string::npos) {
+    base::StringView offline_cpu_range(ss.cur_token(), ss.cur_token_size());
+    size_t dash_pos = offline_cpu_range.find('-');
+    if (dash_pos == base::StringView::npos) {
       // Single CPU in the format of "%d".
-      std::optional<uint32_t> cpu = base::StringToUInt32(offline_cpu_range);
+      std::optional<uint32_t> cpu = base::StringViewToUInt32(offline_cpu_range);
       if (cpu.has_value()) {
         offline_cpus_tmp.push_back(cpu.value());
       } else {
         PERFETTO_ELOG("Failed to parse single CPU from offline CPU range: %s",
-                      offline_cpu_range.c_str());
+                      offline_cpu_range.data());
         return false;
       }
     } else {
       // Range of CPUs in the format of "%d-%d".
       std::optional<uint32_t> start_cpu =
-          base::StringToUInt32(offline_cpu_range.substr(0, dash_pos));
+          base::StringViewToUInt32(offline_cpu_range.substr(0, dash_pos));
       std::optional<uint32_t> end_cpu =
-          base::StringToUInt32(offline_cpu_range.substr(dash_pos + 1));
+          base::StringViewToUInt32(offline_cpu_range.substr(dash_pos + 1));
       if (start_cpu.has_value() && end_cpu.has_value()) {
         for (auto cpu = start_cpu.value(); cpu <= end_cpu.value(); ++cpu) {
           offline_cpus_tmp.push_back(cpu);
         }
       } else {
         PERFETTO_ELOG("Failed to parse CPU range from offline CPU range: %s",
-                      offline_cpu_range.c_str());
+                      offline_cpu_range.data());
         return false;
       }
     }
--- a/src/traced/probes/ftrace/tracefs.cc
+++ b/src/traced/probes/ftrace/tracefs.cc
@@ -450,6 +450,52 @@ size_t Tracefs::NumberOfCpus() const {
   return num_cpus;
 }
 
+bool Tracefs::GetOfflineCpus(std::vector<uint32_t>* offline_cpus) {
+  std::string offline_cpus_str;
+  if (!ReadFile("/sys/devices/system/cpu/offline", &offline_cpus_str)) {
+    PERFETTO_ELOG("Failed to read offline cpus file");
+    return false;
+  }
+  offline_cpus_str = base::TrimWhitespace(offline_cpus_str);
+
+  // The offline cpus file contains a list of comma-separated CPU ranges.
+  // Each range is either a single CPU or a range of CPUs, e.g. "0-3,5,7-9".
+  // Source: https://docs.kernel.org/admin-guide/cputopology.html
+  std::vector<uint32_t> offline_cpus_tmp;
+  for (base::StringSplitter ss(offline_cpus_str, ','); ss.Next();) {
+    std::string offline_cpu_range(ss.cur_token(), ss.cur_token_size());
+    auto dash_pos = offline_cpu_range.find('-');
+    if (dash_pos == std::string::npos) {
+      // Single CPU in the format of "%d".
+      std::optional<uint32_t> cpu = base::StringToUInt32(offline_cpu_range);
+      if (cpu.has_value()) {
+        offline_cpus_tmp.push_back(cpu.value());
+      } else {
+        PERFETTO_ELOG("Failed to parse single CPU from offline CPU range: %s",
+                      offline_cpu_range.c_str());
+        return false;
+      }
+    } else {
+      // Range of CPUs in the format of "%d-%d".
+      std::optional<uint32_t> start_cpu =
+          base::StringToUInt32(offline_cpu_range.substr(0, dash_pos));
+      std::optional<uint32_t> end_cpu =
+          base::StringToUInt32(offline_cpu_range.substr(dash_pos + 1));
+      if (start_cpu.has_value() && end_cpu.has_value()) {
+        for (auto cpu = start_cpu.value(); cpu <= end_cpu.value(); ++cpu) {
+          offline_cpus_tmp.push_back(cpu);
+        }
+      } else {
+        PERFETTO_ELOG("Failed to parse CPU range from offline CPU range: %s",
+                      offline_cpu_range.c_str());
+        return false;
+      }
+    }
+  }
+  *offline_cpus = std::move(offline_cpus_tmp);
+  return true;
+}
+
 void Tracefs::ClearTrace() {
   std::string path = root_ + "trace";
   PERFETTO_CHECK(ClearFile(path));  // Could not clear.
@@ -460,11 +506,25 @@ void Tracefs::ClearTrace() {
   // In case some of the CPUs were not online, their buffer needs to be
   // cleared manually.
   //
+  // Note: There is a small time gap between when we clear the trace file and
+  // when we get the offline cpus. This introduces a small chance that some CPUs
+  // are missed if they came online in between those two points. This is a
+  // best-effort approach to keep the logic simple.
+  //
   // We cannot use PERFETTO_CHECK as we might get a permission denied error
   // on Android. The permissions to these files are configured in
   // platform/framework/native/cmds/atrace/atrace.rc.
-  for (size_t cpu = 0, num_cpus = NumberOfCpus(); cpu < num_cpus; cpu++) {
-    ClearPerCpuTrace(cpu);
+  std::vector<uint32_t> offline_cpus;
+  if (GetOfflineCpus(&offline_cpus)) {
+    for (const auto& cpu : offline_cpus) {
+      ClearPerCpuTrace(cpu);
+    }
+  } else {
+    // Fallback: if we can't determine which CPUs are offline, clear the buffers
+    // for all possible CPUs.
+    for (size_t cpu = 0, num_cpus = NumberOfCpus(); cpu < num_cpus; cpu++) {
+      ClearPerCpuTrace(cpu);
+    }
   }
 }
 
@@ -655,12 +715,16 @@ bool Tracefs::IsFileReadable(const std::string& path) {
   return access(path.c_str(), R_OK) == 0;
 }
 
+bool Tracefs::ReadFile(const std::string& path, std::string* str) const {
+  return base::ReadFile(path, str);
+}
+
 std::string Tracefs::ReadFileIntoString(const std::string& path) const {
   // You can't seek or stat the tracefs files on Android.
   // The vast majority (884/886) of format files are under 4k.
   std::string str;
   str.reserve(4096);
-  if (!base::ReadFile(path, &str))
+  if (!ReadFile(path, &str))
     return "";
   return str;
 }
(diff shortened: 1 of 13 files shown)


LalitMaganti:
LG but I'll take Ryan take the lead on this one.

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -460,11 +506,25 @@ void Tracefs::ClearTrace() {
...
   // cleared manually.
   //
+  // Note: There is a small time gap between when we clear the trace file and
+  // when we get the offline cpus. This introduces a small chance that some CPUs
+  // are missed if they came online in between those two points. This is a
+  // best-effort approach to keep the logic simple.
+  //
   // We cannot use PERFETTO_CHECK as we might get a permission denied error
   // on Android. The permissions to these files are configured in
   // platform/framework/native/cmds/atrace/atrace.rc.
```
rsavitski: nit: let's drop the highlighted comments, they're not adding much. Edge cases from cpu hotplug during tracing are expected, and the CHECK comment is a bit excessive since the lack of a CHECK is self-documenting in not being a mandatory invariant.

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -450,6 +450,52 @@ size_t Tracefs::NumberOfCpus() const {
   return num_cpus;
 }
 
+bool Tracefs::GetOfflineCpus(std::vector<uint32_t>* offline_cpus) {
```
rsavitski: style nit: fine as is, but I would argue that `std::optional<std::vector<...>>` would've been a clearer return type now that we target c++17.

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -450,6 +450,52 @@ size_t Tracefs::NumberOfCpus() const {
...
+    PERFETTO_ELOG("Failed to read offline cpus file");
+    return false;
+  }
+  offline_cpus_str = base::TrimWhitespace(offline_cpus_str);
+
+  // The offline cpus file contains a list of comma-separated CPU ranges.
+  // Each range is either a single CPU or a range of CPUs, e.g. "0-3,5,7-9".
+  // Source: https://docs.kernel.org/admin-guide/cputopology.html
+  std::vector<uint32_t> offline_cpus_tmp;
+  for (base::StringSplitter ss(offline_cpus_str, ','); ss.Next();) {
```
rsavitski, line 470: I'm fine with this, but will note that Primiano some concerns that we're introducing too much complexity for an edge case, and that we should keep it simple.

One option would be to just check that there's at least one offline cpu core before doing the per-cpu fallback, instead of trying to parse the exact mask. That would move the edge case to "many-core machine + offlined cores + very sensitive to time spent in ClearTrace()", but also avoid whatever potential string parsing errors we're not seeing here :--).

Consider the much simpler to read & maintain callsite in that scenario:
```
  if (!HasOfflineCpus()) {
    return;
  
  for (size_t cpu = 0, num_cpus = NumberOfCpus(); cpu < num_cpus; cpu++) {
    ClearPerCpuTrace(cpu);
  }
```

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -655,12 +715,16 @@ bool Tracefs::IsFileReadable(const std::string& path) {
   return access(path.c_str(), R_OK) == 0;
 }
 
+bool Tracefs::ReadFile(const std::string& path, std::string* str) const {
```
rsavitski, line 729: What's wrong with using ReadFileIntoString for the offlined cores file? I'd prefer not to add any extra shims for the mock-based tests. (Plus I think we should be moving towards a FakeTracefs for majority of tests anyway.)

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: Done

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: agreed, updated to std::optional

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: `What's wrong with using ReadFileIntoString for the offlined cores file?`
We want to know if base::readfile fails so we can invoke the fallback of clearing all CPUs. Currently `ReadFileIntoString` in case of failure returns empty string which is also a possible return value when no cpu is offline.

do you mean `MockTracefs` by FakeTracefs ?

Now the reason for adding the wrapper function `Tracefs::ReadFile` for `base::ReadFile` is purely for testing purposes following the recommendation [go/gmockcook#mocking-free-functions 
](http://go/gmockcook#mocking-free-functions)

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: My reasoning for current approach is: 

1) For no offline cpu: the string should be empty (if not some wider issue in kernel or in base functions(TrimWhitespace/StringSplitter)), so effectively no parsing logic should happen.

2) For some offline cpus(though rare): switching to `just check that there's at least one offline cpu core before doing the per-cpu fallback, instead of trying to parse the exact mask.` would effectively mean no trace for scenario with high NumberOfCpus  and short trace_duration. 
 
From local runs we observe:
```
[242.828]          tracefs.cc:466 Tracefs::ClearTrace started for NumberOfCpus=128
[251.148]          tracefs.cc:474 Tracefs::ClearTrace done
```

```
[416.948]          tracefs.cc:468 Tracefs::ClearTrace started for NumberOfCpus=256
[432.652]          tracefs.cc:475 Tracefs::ClearTrace done
```

Now if no issue with parsing, we improve on this and in case any issue we fallback effectively to the same.

Open to change if you aren't convinced.

src/traced/probes/ftrace/tracefs.cc
rsavitski: Ah, I see your point of needing to distinguish the error case, and wanting to inject that in the test.

It's weird to have a mocking testing shim for both ReadFileIntoString and ReadFile, if the earlier is a minimal wrapper around the latter. Though I can see how it's easier to mock the earlier.

> do you mean MockTracefs by FakeTracefs?
I was talking about a hypothetical fake that we should finally write. A significant chunk of ftrace tests do not assert interactions, and therefore would be easier to maintain with a fake. But let's put that aside for this patch, it's not directly relevant.

src/traced/probes/ftrace/tracefs.cc
rsavitski: > so effectively no parsing logic should happen.

But it's still a bunch of code added to the class, so the ratio of code vs applicability is rather low. Plus the "rare" codepath will be less exercised in practice and therefore more likely to have issues/bitrot.

But to reiterate, I'm personally fine with this patch. In fact we can probably later move the string parsing to src/kernel_utils/, as it comes up in a bunch of procfs/sysfs places.

src/traced/probes/ftrace/tracefs.h
```diff
@@ -152,6 +153,11 @@ class Tracefs {
   // This will match the number of tracing/per_cpu/cpuXX directories.
   size_t virtual NumberOfCpus() const;
 
+  // Parses the list of offline CPUs from "/sys/devices/system/cpu/offline" and
+  // returns them as a vector if successful, or std::nullopt if any failure in
+  // reading or parsing.
+  std::optional<std::vector<uint32_t>> GetOfflineCpus();
```
rsavitski: Can we make this protected since it's effectively a utility helper function that is only here because tracefs is the single caller atm, and the only reason you're exposing it in the interface is for tests?

src/traced/probes/ftrace/tracefs_unittest.cc
```diff
@@ -143,5 +152,39 @@ TEST(TracefsTest, ReadBufferSizeInPages) {
   EXPECT_THAT(ftrace.GetCpuBufferSizeInPages(), 1);
 }
 
+TEST(TracefsTest, GetOfflineCpus) {
```
rsavitski: It'd be useful to have a test showing that the per-cpu buffer clearing is working as expected when calling into ClearTrace(), since that's the logic we actually care about.

src/traced/probes/ftrace/tracefs.cc
primiano: My theory is that here we end up spending a lot of complexity and time just for a rare case.
Can we do instead sysconf(_SC_NPROCESSORS_CONF) - sysconf(_SC_NPROCESSORS_ONLN) ?

It removes a lot of complexity and in practice it will make 90% of the use cases we care about fast

primiano:
Can we just do  sysconf(_SC_NPROCESSORS_CONF) - sysconf(_SC_NPROCESSORS_ONLN) ?
I think a lot of this complexity is not needed in the first place and this CL can get waaaaay simpler.
Whta i'm proposing is:

1. If al CPUs are online (which is true in most cases) just truncate the /trace txt file
2. 2. If any cpu is offline, fall back on today's code which clears the individual buffers.

LalitMaganti:
> Can we just do sysconf(_SC_NPROCESSORS_CONF) - sysconf(_SC_NPROCESSORS_ONLN) ? I think a lot of this complexity is not needed in the first place and this CL can get waaaaay simpler. Whta i'm proposing is:
> 
> 1. If al CPUs are online (which is true in most cases) just truncate the /trace txt file
> 2. 2. If any cpu is offline, fall back on today's code which clears the individual buffers.

I'll be honest, I'm not a fan of doing "simplifications" here which, if we get wrong are a nightmare to debug. You could easily imagine a high cpu server with a core offline for some reason and all we'll be told is "Perfetto does not work" and we'll be scratching our head for years. 

It took us 3 years to figure out the root cause of this bug, the next one will be even worse because it will get more and more niche

Ftrace is our bread and butter. I would argue we should do it right.

primiano:
Okay but at very least we should have a fastpath using sysconf so we if we know upfront that all the CPUs are online we can avoid any looking into /sys at all.

src/traced/probes/ftrace/tracefs.h
sashwinbalaji: After further consideration, I agree that protected is the correct choice to maintain a minimal public API. Using the derived-class pattern for testing, as suggested in http://yaqs/4616483600728064#a5173732018618368

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: Ack, added the fast-path check to exit early if all CPUs are online.

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: Will take it up in a separate cl to update the tests mocking `ReadFileIntoString` to instead rely on `ReadFile` and then make  ReadFileIntoString private and non-virtual

src/traced/probes/ftrace/tracefs_unittest.cc
sashwinbalaji: Ack added test for ClearTrace to verify this

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -450,21 +450,87 @@ size_t Tracefs::NumberOfCpus() const {
...
-  // We cannot use PERFETTO_CHECK as we might get a permission denied error
-  // on Android. The permissions to these files are configured in
-  // platform/framework/native/cmds/atrace/atrace.rc.
-  for (size_t cpu = 0, num_cpus = NumberOfCpus(); cpu < num_cpus; cpu++) {
-    ClearPerCpuTrace(cpu);
+  // in the kernel. So if all cpus are online, no further action needed.
+  if (total_cpu_count == online_cpu_count)
+    return;
+
+  PERFETTO_DLOG(
```
primiano: I'd promote this to a log. this is realistically never going to be hit, but if it's hit might help explaining why we are taking more time.

src/traced/probes/ftrace/tracefs.cc
primiano: imho in a separate CL we should have a single method that returns a StatusOr<string> (or optional<string>) which is mocked, and the other should be a wrapper around it to preserve the existing cases that don't care about failure

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: Done.

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -450,21 +450,87 @@ size_t Tracefs::NumberOfCpus() const {
...
+        PERFETTO_ELOG("Failed to parse CPU range from offline CPU range: %s",
+                      offline_cpu_range.data());
+        return std::nullopt;
+      }
+    }
+  }
+  return offline_cpus;
+}
+
 void Tracefs::ClearTrace() {
```
rsavitski, line 503: LGTM, but given that you're adding an android flag to roll this flag out:

1) Should we invert the submission order (flag definition first), so that you can immediately gate this code behind the flag? Otherwise it'll technically be on by default at first in strict commit order (not a true problem in practice since copybara will merge them at once).

2) How about this code layout instead, accounting for the android flag, it cuts down on the duplication from #2361:
```
void Tracefs::ClearTrace() {
  std::string path = root_ + "trace";
  PERFETTO_CHECK(ClearFile(path));  // must be allowed to clear buffers

  // Truncating the trace file leads to tracing_reset_online_cpus being called
  // in the kernel. In case some of the CPUs were not online, their buffer needs
  // to be cleared separately.
  if (ANDROID_FLAG) {
    size_t total_cpu_count = NumberOfCpus();
    size_t online_cpu_count = NumberOfOnlineCpus();

    if (total_cpu_count == online_cpu_count)
      return;

    PERFETTO_LOG(
        "Since %zu / %zu CPUS are online, clearing buffer for the offline ones "
        "individually.",
        online_cpu_count, total_cpu_count);

    std::optional<std::vector<uint32_t>> offline_cpus = GetOfflineCpus();
    if (offline_cpus.has_value()) {
      for (const auto& cpu : offline_cpus.value()) {
        ClearPerCpuTrace(cpu);
      }
      return;
    }
    // If we couldn't read the offline mask, fallthrough to clearing all per-cpu
    // buffers.
  }

  for (size_t cpu = 0, num_cpus = NumberOfCpus(); cpu < num_cpus; cpu++) {
    ClearPerCpuTrace(cpu);
  }
}
```

sashwinbalaji: For point 2, I think the updated PR already matches the template. #2361

rsavitski: Yup, looks good, I didn't see the update in time.

rsavitski: Having though about the first point a bit more - meh, it's more effort messing around with diffbases than it's probably worth.

Approving both patches for merging, thanks.

rsavitski (approved):

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: Ack, will fix this in a separate cl. resolving to merge the current PR

PR #2265, opened 2025-07-23
Fix uint overflow in proto_trace_tokenizer

### LINKED ISSUE #2264: Unsigned integer overflow in ProtoTraceTokenizer
In the ProtoTraceTokenizer, at https://github.com/google/perfetto/blob/edd70a17a1fdc811de00869e4e6aa6cbef92b58f/src/trace_processor/importers/proto/proto_trace_tokenizer.h#L168 there is a potential overflow in `hdr_size + field_size`. This might cause the subsequent code to read more data out of the reader than what is available.

antosart:
Fixes #2264.

### DIFF
--- a/src/trace_processor/importers/proto/proto_trace_tokenizer.h
+++ b/src/trace_processor/importers/proto/proto_trace_tokenizer.h
@@ -164,8 +164,7 @@ class ProtoTraceTokenizer {
       }
 
       // If there's not enough bytes in the reader, then we cannot do anymore.
-      size_t size_incl_header = hdr_size + field_size;
-      if (size_incl_header > avail) {
+      if (field_size > avail - hdr_size) {
         return base::OkStatus();
       }
 


src/trace_processor/importers/proto/proto_trace_tokenizer.h
```diff
@@ -164,8 +164,9 @@ class ProtoTraceTokenizer {
       }
 
       // If there's not enough bytes in the reader, then we cannot do anymore.
-      size_t size_incl_header = hdr_size + field_size;
-      if (size_incl_header > avail) {
+      uint64_t size_incl_header = hdr_size + field_size;
```
LalitMaganti: Instead of this, I suggest writing the code as follows to avoid the overflow entriely:
```
      // If there's not enough bytes in the reader, then we cannot do anymore.
      if (field_size > avail - hdr_size) {
        return base::OkStatus();
      }
      size_t size_incl_header = hdr_size + field_size;
```

antosart: makes sense, thanks!

LalitMaganti (approved):

PR #2267, opened 2025-07-23
ui: support for hosting Perfetto in other applications

### LINKED ISSUE #2266: ui: rich support for embedding Perfetto UI in other applications
Perfetto UI currently supports an embedded mode via a URL parameter. This suppresses the sidebar and other default behaviours of the UI. However, I am building an application whose integration needs go deeper than this. In particular, I need to be able to

- adjust the way routing works because my application uses the URL hash already for its own purposes
- I need to tweak the content security policy
- I need to customize the keys used by `CacheManager` for caching in the browser, to separate the caches from my application and satisfy other restrictions
- allow file CORS origins
- my application loads up traces into the trace processor before initializing the Perfetto UI, so I need to suppress the dialog that prompts to open a trace already loaded
- handle uncaught errors and promise rejections in my application
- and possibly more

So some kind of API by which I can inject customizations of these Perfetto UI behaviours, before initializing the UI app, would be most helpful to my application.

cdamus:
Define a new embedderContext variable exported from a new embedder.ts module that an embedding application can assign once to inject customization of the Perfetto UI behavior. This includes flags to suppress default capabilities and call-back hooks to override default capabilities, such as:

- adjust routing to work with applications that use hashes already for their own incompatible routing requirements
- customization of the installation of the content security policy
- allow to control cache keys for browser restrictions on caching
- allow file CORS origins for desktop-style applications
- scope key handling to the viewer page so that it doesn't get key events from other UI elements in the host application
- current script is null also in some embedder apps, not just in tests
- suppressing the dialog that prompts to open a trace already loaded in the trace_processor_shell, for apps that manage this a priori
- suppressing registration of error handlers. An embedding application often has its own error handling

Additionally, in support of incorporating the build output into another application's packaging:

- enable generating .d.ts files. Some types had to be exported that previously weren't in order for TypeScript to be able to correctly resolve its generated type definitions
- support loading the trace database from a stream provided by the host application (such as from a remote backend)
- make @types/mithril a dev dependency
- remove unused dependencies from package.json
- explicitly define which location TS `@types` should be read from. Otherwise the tsc compiler possibly finds the types in an embedder application's `node_modules/@types` directory, which can cause clashes esp. between Mithril and React's types for HTML elements
- return showHelp's showModal promise. This allows a host application to track whether the help dialog is (still) showing

Fixes #2266

### DIFF
--- a/docs/concepts/ui-embedder-context-api.md
+++ b/docs/concepts/ui-embedder-context-api.md
@@ -1,0 +1,60 @@
+# EmbedderContext API
+
+This document provides an overview of the `EmbedderContext` API in Perfetto UI, as defined in [`ui/src/core/embedder.ts`](../../ui/src/core/embedder.ts).
+This API allows host (embedding) applications to customize various aspects of Perfetto UI's behavior, especially for advanced integration scenarios.
+
+## Overview
+
+The `EmbedderContext` interface provides a collection of optional hooks, flags, and handlers that a host application can implement to control or override specific Perfetto UI behaviors.
+The primary intended use case is in applications that take care of such concerns as acquiring and managing Perfetto traces but wish to reuse/embed the Perfetto UI for presentation of those traces to the user.
+
+The context object must be set *before* loading the main Perfetto UI module to ensure proper integration.
+
+Host applications can configure the `EmbedderContext` to customize the following aspects of the Perfetto UI app:
+
+- **Custom Error Handling**: Suppress Perfetto UI's internal error handling, deferring to whatever the host application does in that regard.
+- **UI Rendering Control**: Prevent the start-up of the Perfetto UI app from rendering its main interface, allowing the host application can selectively instantiate components.
+- **Content Security Policy Strategy**: Customization of how the CSP is installed in the application, with optional filtering of the rules.
+- **Storage Customization**: Prefix cache storage keys for coexistence with other applications' caches and for certain application frameworks' restrictions.
+- **Routing Hooks**: Intercept or override URL navigation and route handling.
+- **Custom PostMessage Handling**: Process unrecognized messages posted to the window.
+- **Preloaded Trace Handling**: Indicate how Perfetto should deal with traces pre-loaded in the trace processor backend.
+- **App State Restoration**: Delegate Perfetto app state persistence to the host application.
+
+## Setting the Embedder Context
+
+The host application _must_ set its embedder context via the `setEmbedderContext()` function _before_ the Perfetto UI app is initialized.
+This function may be called only once.
+Attempting to replace an established embedder context throws an error.
+
+Since Perfetto UI's main entry point (e.g., `frontend/index.ts`) references the embedder context as it initializes, the safest approach is:
+
+- define your embedder context and call `setEmbedderContext()` in a dedicated module.
+Ideally it will not need to import any other Perfetto UI modules than `embedder.ts` and a few minimal dependencies for types and APIs such as `Router` and `RafScheduler`
+- ensure that this module is imported by your application before any Perfetto source modules
+
+### Example
+
+Suppose you have an application integrating Perfetto:
+
+```typescript
+// perfetto-setup.ts
+import {setEmbedderContext} from 'perfetto/ui/dist/core/embedder';
+
+setEmbedderContext({
+  suppressErrorHandling: true,
+  suppressMainUi: true,
+  cachePrefix: 'https:/',
+  // ...other overrides
+});
+```
+
+Then, in your application entrypoint:
+
+```typescript
+// index.ts
+import './perfetto-setup';        // <-- Import this first!
+import 'perfetto/ui/dist/frontend/index'; // Now load Perfetto UI
+```
+
+This ensures the embedder context is established before Perfetto's frontend initialization.
--- a/src/trace_processor/rpc/httpd.cc
+++ b/src/trace_processor/rpc/httpd.cc
@@ -48,6 +48,7 @@ const char* kDefaultAllowedCORSOrigins[] = {
     "https://ui.perfetto.dev",
     "http://localhost:10000",
     "http://127.0.0.1:10000",
+    "file://",
 };
 
 class Httpd : public base::HttpRequestHandler {
--- a/ui/src/base/http_utils.ts
+++ b/ui/src/base/http_utils.ts
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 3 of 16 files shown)


docs/concepts/ui-embedder-context-api.md
```diff
@@ -0,0 +1,60 @@
+# EmbedderContext API
```
cdamus, line 1: This is a proposal for the embedder support API previously mentioned. Obviously, it works for my application but I welcome any suggestions for further improvement, especially integration points that I may have missed in parts of the Perfetto UI app that I'm not using.

Also, this PR includes changes to certain aspects of the build and development environment that support developing an application against the Perfetto APIs. That can be factored out into a separate PR if preferred.

cdamus:
Hi @stevegolton have you had a chance to evaluate this proposal/prototype for an "embedder context" API? I understand it's not small and I'll be happy to help the process in any way that I can.

cdamus:
> Apologies for the delay in getting back to you on this.

No worries!

> I think part of the problem is that this really does go quite deep into the core of the app and I'm struggling to reason about in which ways it will affect it. It's a bit of a mountain.

Indeed. This is a refactoring of two years' divergence of our fork from the main branch. Taking a step back to look for an approach not based on the 2023 baseline makes a lot of sense to me.

> In general, the approach you've taken is this: start off with the main UI and disable or replace certain bits if the embedder context is provided. I think this is the wrong way of going about it personally. Instead I think we should create an embeddable core which takes configuration options and callbacks for the various bits you might want to plumb in - I'd imagine this could almost be a standalone library - then our index.ts file simply 'embeds' this into a webpage for the sake of hosting on ui.perfetto.dev, and you can do the same and embed it into eclipsesource how you see fit. This is largely the same outcome, but I think it's a lot cleaner as it avoids all the embedderContext related conditionals sprinkled throughout the codebase.

Yes, something more like reworking the `main()` function of the frontend into an object with a customization API would be attractive to me. It would almost certainly make my integration efforts easier if it works for the stand-alone web page case.

> I'm slightly unsure about the future of this PR. I think it might help a lot splitting it up into smaller chunks.

I'm happy to factor out some of these details, yes.

> 1. Additional exports - very happy to land this.
> 2. Proper return types - Promise from openTrackFrom* functions - again very happy to take this
> 3. loadTraceFromStream - sure, makes sense

Yes, this last one is because our application needs to be able to stream trace files via its own mechanism up to the WASM trace processor. Maybe this could be avoided by refactoring the whole `load_trace.ts` module into a suite of pluggable strategies, instead. Then our application could just contribute its own strategy under some private key.

> The rest of the changes are more controversial and I think we need to discuss further.

Absolutely. Does it work better for you to discuss design ideas in #2266, or to review a new proposal for a more top-down design of a main-program API as a new PR?

> I have a plan for the focusable keyboard handlers which I think is quite elegant.

Sounds good!

stevegolton:
> > I think part of the problem is that this really does go quite deep into the core of the app and I'm struggling to reason about in which ways it will affect it. It's a bit of a mountain.
> 
> Indeed. This is a refactoring of two years' divergence of our fork from the main branch. Taking a step back to look for an approach not based on the 2023 baseline makes a lot of sense to me.

Ah yes, it's a moving target. I think in general it's going to be a slow burn I'm afraid. There are a lot of new features we need to add to the UI and I'm going to struggle to justify working on this embeddability over some of the other works thats going on at the moment.

> > In general, the approach you've taken is this: start off with the main UI and disable or replace certain bits if the embedder context is provided. I think this is the wrong way of going about it personally. Instead I think we should create an embeddable core which takes configuration options and callbacks for the various bits you might want to plumb in - I'd imagine this could almost be a standalone library - then our index.ts file simply 'embeds' this into a webpage for the sake of hosting on ui.perfetto.dev, and you can do the same and embed it into eclipsesource how you see fit. This is largely the same outcome, but I think it's a lot cleaner as it avoids all the embedderContext related conditionals sprinkled throughout the codebase.
> 
> Yes, something more like reworking the `main()` function of the frontend into an object with a customization API would be attractive to me. It would almost certainly make my integration efforts easier if it works for the stand-alone web page case.

Excellent, glad we're on the same page!

> > I'm slightly unsure about the future of this PR. I think it might help a lot splitting it up into smaller chunks.
> 
> I'm happy to factor out some of these details, yes.
> 
> > 1. Additional exports - very happy to land this.
> > 2. Proper return types - Promise from openTrackFrom* functions - again very happy to take this
> > 3. loadTraceFromStream - sure, makes sense
> 
> Yes, this last one is because our application needs to be able to stream trace files via its own mechanism up to the WASM trace processor. Maybe this could be avoided by refactoring the whole `load_trace.ts` module into a suite of pluggable strategies, instead. Then our application could just contribute its own strategy under some private key.

Ah, yes, `load_trace.ts` has gotten rather long in the tooth at this point and is due for a cleanup. TBH I'm finding it hard to justify why we don't just normalize this entry point - have it take a blob + some metadata - then where the trace comes from is entirely up to the embedder. There's a lot of non obvious un-tested business logic in this file, a lot of which predates me by quite a few years, so it's a bit of a minefield trying to refactor without causing regressions. There is also another PR #2615 that is about to land that complicates things slightly.

> > The rest of the changes are more controversial and I think we need to discuss further.
> 
> Absolutely. Does it work better for you to discuss design ideas in #2266, or to review a new proposal for a more top-down design of a main-program API as a new PR?

PR or issue works for me. PR is nice as we have some concrete code to look at, but perhaps discussing on the higher level issue make sense.

> > I have a plan for the focusable keyboard handlers which I think is quite elegant.
> 
> Sounds good!

Please see #3028 for the new HotkeyContext. It only handles hotkeys for now (WASD are handled separately, but I think it's a step in the right direction).

cdamus:
Hi @stevegolton Sorry for the long silence. Priorities, eh.

First, I like the new hot-keys implementation! It works well for me in my application.

At your suggestion I have excised several changes from this PR into new PRs:

- #3168 for the TypeScript definitions fixes
- #3169 to return the modal promise for the help modal
- #3170 for return of trace promises from opening a trace, plus support for opening a trace from a stream

This draft PR is rebased to exclude those changes and now has only the proposed — and to be reworked — "embedder context" API. My next step will be to propose a new PR that reworks this as a main entry-point class, tentatively named `Frontend`, which would incorporate much of what the following bits currently do:

- `main()` function in `index.ts`
- the post-message handler, trace-URL handler, RPC-HTTP dialog
- `AppImpl`
- `Router`

Let's see how that looks 😀

cdamus:
Superseded by #3275 .

PR #2275, opened 2025-07-23
perfetto: add support for readonly flags

LalitMaganti:
This CL does the necessary plumbing to have readonly flags usable in
Perfetto.


include/perfetto/ext/base/flags_list.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef INCLUDE_PERFETTO_EXT_BASE_FLAGS_LIST_H_
+#define INCLUDE_PERFETTO_EXT_BASE_FLAGS_LIST_H_
+
+namespace perfetto::base::flags {
+
+enum class NonAndroidPlatformDefault {
+  kFalse = 0,
+  kTrue = 1,
+};
+
```
primiano: Do we really need this? For now I?d support only bool and refactor if we ever want to support somethign else.
I don't think trunk stable supports non-boolean flags today, right?
I like the fact that "NonAndroidPlatformDefault" is spelled out, but you can achieve the same just by doing

constexpr bool NonAndroidPlatformDefault_TRUE = true;
constexpr bool NonAndroidPlatformDefault_FALSE = false;

and everybody else will just copy-paste after we have 1 flag

include/perfetto/ext/base/flags.h
```diff
@@ -0,0 +1,44 @@
...
+#endif
+
+#if PERFETTO_BUILDFLAG(PERFETTO_ANDROID_BUILD) && \
+    PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+#define PERFETTO_FLAGS_DEFINE_FN(name, default_non_android_value) \
+  [[maybe_unused]] constexpr bool name = ::perfetto::flags::name();
+#else
+#define PERFETTO_FLAGS_DEF_GETTER(name, default_non_android_value) \
+  [[maybe_unused]] constexpr bool name =                           \
+      static_cast<bool>(default_non_android_value);
```
primiano: remove the cast (see other comment)

include/perfetto/ext/base/flags.h
```diff
@@ -0,0 +1,44 @@
...
+#endif
+
+namespace perfetto::base::flags {
+
+PERFETTO_READ_ONLY_FLAGS(PERFETTO_FLAGS_DEF_GETTER)
```
primiano, line 56: merge into one file so it's less confusing?

include/perfetto/ext/base/flags_list.h
```diff
@@ -0,0 +1,38 @@
...
+}  // namespace perfetto::base::flags
+
+// The list of all the read-only flags accessible to the Perfetto codebase.
+//
+// The first argument is the name of the flag. Should match 1:1 with the name
+// in `perfetto_flags.aconfig`.
+// The second argument is the default value of the flag in non-Android platform
+// contexts.
+#define PERFETTO_READ_ONLY_FLAGS(X) \
```
primiano: move this to the top of the flags.h file

PR #2276, opened 2025-07-23
perfetto: add flag for using murmurhash as the default hash in FHM

LalitMaganti:


include/perfetto/ext/base/flags_list.h
```diff
@@ -32,7 +32,8 @@ enum class NonAndroidPlatformDefault {
 // in `perfetto_flags.aconfig`.
 // The second argument is the default value of the flag in non-Android platform
 // contexts.
-#define PERFETTO_READ_ONLY_FLAGS(X) \
-  X(test_read_only_flag, NonAndroidPlatformDefault::kFalse)
+#define PERFETTO_READ_ONLY_FLAGS(X)                         \
+  X(test_read_only_flag, NonAndroidPlatformDefault::kFalse) \
+  X(use_murmur_hash_for_flat_hash_map, NonAndroidPlatformDefault::kFalse)
```
primiano: 1) don't we want it to be True here so we start getting coverage in our CI?
2) Rather than having a flag per class that we switch, IMHO we should have:
-  a flag for "transition to murmur hash"
- A class called MaybeMurmurHash (or TrasitionMurmurHash) 
- Switch FHM to use Hasher = base::MaybeMurmurHash

LalitMaganti: Discussed offline. Every "set of classes we migrate" needs to be behind its own flag.

primiano (approved):

PR #2283, opened 2025-07-24
exp: Better manual filters

aMayzner:
- Support typing filters
- Filters editor (one filter can be modified at the time)
- Show added filters as chips
- Refactor of filter.ts file and polishing the filter styles


LalitMaganti:
Just tried this out. While the manual filters work (great!) the + button does not seem to do anything?

PR #2291, opened 2025-07-24
docs: add committer instructions

primiano:
@LalitMaganti  WDYT?


LalitMaganti:
Yup I like it! Describes exactly what I would expect to see. LGTM.

PR #2296, opened 2025-07-24
Workaround to use synthetic tids for older Linux Chrome versions

betasheet:
Older Chrome versions don't set is_sandboxed_tid, so we try to detect traces with data from these versions and apply synthetic tids to all TrackEvent data.

See https://github.com/google/perfetto/pull/2187 and https://crbug.com/432057400 for context.

### DIFF
--- a/src/trace_processor/export_json_unittest.cc
+++ b/src/trace_processor/export_json_unittest.cc
@@ -46,6 +46,7 @@
 #include "src/trace_processor/importers/common/tracks.h"
 #include "src/trace_processor/importers/common/tracks_common.h"
 #include "src/trace_processor/importers/common/tracks_internal.h"
+#include "src/trace_processor/importers/proto/proto_importer_module.h"
 #include "src/trace_processor/importers/proto/track_event_tracker.h"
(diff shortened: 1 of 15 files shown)


LalitMaganti:
Given our recent discussions I'm not sure this is even necessary anymore. AFAIK we are planning on preserving backcompat for TrackEvent eternally as it has the potential to do stuff wrong. Can we wait on this until we figure out if it's necessary or not (the ProfileModule *might* be necessary depending on what we find).

betasheet:
> Given our recent discussions I'm not sure this is even necessary anymore. AFAIK we are planning on preserving backcompat for TrackEvent eternally as it has the potential to do stuff wrong. Can we wait on this until we figure out if it's necessary or not (the ProfileModule might be necessary depending on what we find).

Yes, that was my hope, too. Just wanted to have this ready so you're not blocked if you decide to revert to the original idea. Jahdiel, I'll leave this in your hands while I'm out.

LalitMaganti:
Closing as this is no longer relevant.

PR #2322, opened 2025-07-25
tp: speed up situations where we havve direct paths between clocks

LalitMaganti:
If we have a direct path, there's no point doing the full DFS. We can just do the
direct lookup and find the hash.

On traces with a lot of snapshots, this saves ~50% of the time taken for the
trace.


LalitMaganti:
Subsumed by https://github.com/google/perfetto/pull/2325

PR #2325, opened 2025-07-25
tp: significantly improve load perf of traces with lots of snapshots

LalitMaganti:
Instead of adding a ton of clocks only to immediately return the first, just directly return the first clock if it matches the target

Fixes perf on large chrome traces which have a ton of snapshots. Takes load time on trace from b/433757600 from 27s -> 16s.

Fixes: 433757600


LalitMaganti:
Anna/Ashwin: can I ask you to review and merge this for me next week?

sashwinbalaji (approved):
nice, the only corner case would have been source = target but I guess that is not possible considering the check `PERFETTO_CHECK(src != target);`

PR #2331, opened 2025-07-27
Add line number for task in chrome://tracing

keyou:
When capturing traces via chrome://tracing, Task trace events only include the filename and function name, missing the line number. This PR fixes the issue.

### DIFF
--- a/src/trace_processor/export_json.cc
+++ b/src/trace_processor/export_json.cc
@@ -665,6 +665,7 @@ class JsonExporter {
             if (posted_from.isMember("function_name")) {
               args["src_func"] = posted_from["function_name"];
               args["src_file"] = posted_from["file_name"];
+              args["src_line"] = posted_from["line_number"];
             } else if (posted_from.isMember("file_name")) {
               args["src"] = posted_from["file_name"];
             }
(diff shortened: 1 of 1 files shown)


LalitMaganti:
@khokhlov962 as the most knoledgable of this code. Granting owners from my side.

LalitMaganti (approved):

khokhlov962:
The change LGTM, but note that line numbers are not currently being written into the trace [1]. You'll need to implement it on the chromium side as well to make it work.

[1] https://source.chromium.org/chromium/chromium/src/+/main:base/trace_event/interned_args_helper.cc;l=46;drc=6426c956fff8675e1be78ba3c70fceb7d2651f18

khokhlov962 (approved):

keyou:
> The change LGTM, but note that line numbers are not currently being written into the trace [1]. You'll need to implement it on the chromium side as well to make it work.
> 
> [1] https://source.chromium.org/chromium/chromium/src/+/main:base/trace_event/interned_args_helper.cc;l=46;drc=6426c956fff8675e1be78ba3c70fceb7d2651f18

Yeah, it's in review on https://chromium-review.googlesource.com/c/chromium/src/+/6787472.

PR #2332, opened 2025-07-28
bazel: Export //protos/third_party/chromium:zero

mlcui-corp:
Internally within Google, we build [libchrome], a subset of Chromium's `//base` library, with Bazel. This requires building the protos in Chromium's `//base/tracing/protos`, automatically rolled into Perfetto, as protozero headers, as OSS libchrome [generates the headers](https://crrev.com/c/4852792) after building Perfetto's protozero plugin.

Due to the [one version rule] and other restrictions, we cannot import these protos twice into the internal monorepo. The source of truth would therefore be Perfetto, and so we would need protozero headers for these Chromium protos inside the Perfetto build rules.

Expose `//protos/third_party/chromium:zero` in Bazel to allow building Chromium's `//base/tracing/protos` with Bazel.

Googlers: See b/433847644 for more details.

[libchrome]: https://www.chromium.org/chromium-os/developer-library/guides/infrastructure/libchrome/
[one version rule]: https://opensource.google/documentation/reference/thirdparty/oneversion

### DIFF
--- a/tools/gen_bazel
+++ b/tools/gen_bazel
@@ -164,6 +164,21 @@ proto_groups = {
     }
 }
 
+protozero_groups = {
+    'trace_zero': {
+        'sources': [
+            '//protos/perfetto/trace:minimal_zero',
+            '//protos/perfetto/trace:non_minimal_zero',
+        ],
+        'visibility': ALLOWLIST_PUBLIC_VISIBILITY,
+    },
+    # Required for building Chromium's `//base` with Bazel.
+    'chromium_zero': {
+        'sources': ['//protos/third_party/chromium:zero'],
+        'visibility': ALLOWLIST_PUBLIC_VISIBILITY,
+    }
+}
+
 # Path for the protobuf sources in the standalone build.
 buildtools_protobuf_src = '//buildtools/protobuf/src'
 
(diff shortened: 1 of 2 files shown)


primiano:
Can you please figure out some other way to depend on these protos / make a copy?

THe problem is not the export of the proto. Is having more dependencies on our bazel targets.

I have two main concerns here:

1) Historically bazel has been a source of nightmares for us. To the day there is no active project depending on bazel (gpuinspector.dev used to, and to the best of my knowledge used to be the only one). Our bazel build only exists as a proxy for the internal blaze targets. I am really not looking  forward to get more bazel dependencies.

2) I am not a fan of libchrome honestly and don't want to get involved with being a dependency on it. I have bad memories of libbase since the day i was in chromium. that project takes arbitrary subsets of the chromium codebase (which has never been intended to be used as a library, //base maintainers have clarified that in a number of occasions) and tries to make them work doing weird surgeries; often being in the middle of decisions that chromium upstream isn't committed to maintain.
Supporting chrome itself is already a big burden for our project. I don't want to get involved into supporting a derivative project of chromium which is not blessed by chromium itself.

I'd prefer if you did whatever needed to achieve your goals by copying/forking the repo, rather than pushing the support and mainteinance cost on our project. We have reached the limit of how many weird/unusual build configurations we have to support and I'm not looking to add more. 

We are in a state that every tiny change in perfetto causes backfires everywhere downstream because there is always somebody using that in some weird configuration, and that is hurting our project velocity.

I hope you can understand.

tools/gen_bazel
```diff
@@ -164,6 +164,21 @@ proto_groups = {
     }
 }
 
+protozero_groups = {
```
LalitMaganti, line 167: I'm not a fan of adding this extra ad-hoc branch. Can you instead please modify the proto_groups above instead to keep things consistent.

LalitMaganti:
Largely agree with Primiano but I'm a bit more optimistic on this change as long as we keep it more consistent with the other things we have today

LalitMaganti:
Seems like this is being solved in libchrome with cl/791416787

PR #2335, opened 2025-07-28
Add blocking call with frame and CUJ composition in stdlib.

devianb:
This table is useful for metric calculation as well as ui plugin. To avoid code duplication, this has been added to the stdlib as a hidden table.

Bug: 400855183
Test: tools/diff_test_trace_processor.py --keep-input --name-filter ".*android_blocking_calls_cuj_per_frame.*" out/linux_clang_debug/trace_processor_shell


src/trace_processor/perfetto_sql/stdlib/android/blocking_calls/blocking_calls_aggregation.sql
```diff
@@ -0,0 +1,70 @@
+--
```
LalitMaganti: This should really live in a module called frame_blocking_calls or similar since it's more "blocking calls in a frame" rather than "frames associated with blocking calls".

devianb: Done.

src/trace_processor/perfetto_sql/stdlib/android/frame_blocking_calls/blocking_calls_aggregation.sql
```diff
@@ -0,0 +1,70 @@
...
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- This module primarily performs composition between the blocking calls, frames and CUJs.
+-- This is used for capturing blocking call per frame metrics, and the related plugins.
+INCLUDE PERFETTO MODULE android.critical_blocking_calls;
+
+INCLUDE PERFETTO MODULE android.cujs.sysui_cujs;
+
+-- For cases when a blocking call starts within a frame, but does not end before the actual frame
```
nicomazz, line 22: nit: maybe it's worth fixing the formatting now

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/frame_blocking_calls/blocking_calls_aggregation.sql
devianb: The formatter script adds the new line automatically. Presubmit passes with this formatting. Keeping it as it is.

PR #2167, opened 2025-07-15
ui: refactor CSS to support embedding applications

cdamus:
As described in #2166, these changes are sufficient for my embedding application to isolate its CSS styling from Perfetto UI and to apply its light and dark themes, covering these three subjects:

- broadening the application of the `pf-` class name prefix that seems recently to be the norm to longer extant style classes
- extracting font, color, shadow, and other similar definitions into real (not Sass) CSS variables that can be overridden by an embedding application for consistency with its them
- pull more fonts and colors from the effective CSS styles in the DOM in canvas rendering code, to match the theming supported by the previous item

I have attempted to organize the changes thematically (no pun intended) into separate commits for clarity. Please let me know if you prefer them all squashed or factored out into separate PR. Although, in the latter case, they would constitute a chain of dependencies as each commit builds on the previous.

And, of course, what further changes you would request in this PR or in follow-ups. For example, my application does not use the minimap timeline overview, so that is not changed in this PR.

Fixes #2166

PR #2337, opened 2025-07-28
ui: support multiple loaded traces open simultaneously

### LINKED ISSUE #2336: ui: support instantiation of the UI on multiple loaded traces simultaneously
Perfetto UI currently presents a single loaded trace at any time. When a trace is loaded, any previously open trace is disposed.

I am building an application that runs and collects Perfetto trace data from multiple applications and does so in an iterative process, in which I need to be able to present multiple loaded traces to the user at one time. And, of course, my application wants to use the Perfetto UI that is so well suited to this task.

So, it would be great if the UI could support a mode of operation in which

- the `AppImpl` maintains any number of loaded `TraceImpl`s at any time
- the `UiMain` and/or components thereof (perhaps just individual pages such as the Viewer page and Trace Info page) is parameterized with the `TraceImpl` to present and thus could be instantiated multiple times in the DOM for different traces
- all UI components at all levels of the UI structure have the trace injected into them that they need to present, rather than relying on the currently open trace known via the `AppImpl` singleton

Of course, it would be up to embedder applications such as mine to determine when and where in the DOM to instantiate the UI for any given trace, and also when to unload a trace and dispose of any registrations or other book-keeping associated with it.

cdamus:
This pull request implements a prototype of support for managing multiple open traces in the Perfetto UI. This is based on a new concept of _hierarchical registries_ for trace-scoped registrations of services and the employment of this facility in a select number of the `XyzManager` classes to let traces inherit app-level registrations and add their own.

The PR comprises two commits, one which is simply a wide-ranging refactoring to "push" the `Trace` object down into the leaf levels of UI components that previously "pulled" the currently active trace from the `AppImpl` singleton. This is a prerequisite to being able to instantiate multiple `UiMain` components for multiple traces in separate subtrees of the DOM, but of course I am happy to split it out into its own PR if that would be helpful.

This PR, together with #2267 and #2167 raised previously, seem in my testing so far to be sufficient to allow my application to instantiate any number of instances of the Perfetto UI (especially the trace viewer and info pages) to present the traces run and collected by my application. And in the dark theme that my UX designers prefer.

Fixes #2336

### DIFF
--- a/ui/src/base/registry.ts
+++ b/ui/src/base/registry.ts
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {assertExists, assertFalse} from './logging';
+
 export interface HasKind {
   kind: string;
 }
@@ -26,6 +28,8 @@ export class RegistryError extends Error {
 export class Registry<T> {
   private key: (t: T) => string;
   protected registry: Map<string, T>;
+  private _keyFilter?: (key: string) => boolean;
+  private readonly keyFilter = (key: string) => this._keyFilter?.(key) ?? true;
 
   static kindRegistry<T extends HasKind>(): Registry<T> {
     return new Registry<T>((t) => t.kind);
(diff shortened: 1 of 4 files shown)


docs/design-docs/multiple-open-traces.md
```diff
@@ -0,0 +1,59 @@
+# Support for Multiple Open Traces
```
cdamus, line 1: As in the case of #2267, this is a prototype proposing a way to add support for multiple loaded traces — and multiple instantiation of the UI for those traces — without disrupting the current UI application function. Obviously, it seems to satisfy my application's needs, so I hope this will be a useful starting point for discussion and refinement of what multiple-trace support in the UI can look like.

stevegolton:
Would you mind creating a separate commit for the trace injection work `c5af09e` if it's not too much trouble cherry-picking it onto main? I think everything in that commit is uncontroversial and we can land immediately.

After that I think splitting the PR into:
- subregistries
- multi trace loading

... makes the most sense. Thanks!

cdamus:
Thanks for jumping on this so quickly, @stevegolton !

I've created a separate PR #2353 for the trace injection refactoring and rebased this PR onto that.

The next step to split out the hierarchical registries will be a bit more effort. Are you asking for a commit implementing _just_ the changes to the `Registry` API, or also the various `XyzManager` classes that then make use of this new capability?

cdamus:
Hi @stevegolton I took a stab at pulling out the API changes to the `Registry` class for sub-registries and also the service manager classes that leverage the new API for their own trace-scoped sub-managers into a separate PR. Unfortunately, I ran into some difficulty because with the `TraceImpl` then using the sub-managers, there were more dependencies that I needed to include, especially in the `sidebar.ts` and `ui_main.ts` modules, that seemed clearly to be out of scope for this new isolated PR. So, I'm not sure of the practical utility of a separate PR _just_ for the `Registry` class without the changes in client code to use that new API, but then I got lost in the weeds trying to minimize the scope of those dependencies to include, which perhaps only points to some degree of badness in the design of my originally proposed changes.

All of which is to say that I think I need a bit more clarity on the purpose and scope of this second step of factorizing this PR 😀 Thanks!

cdamus:
Hi, @stevegolton thanks for your work on getting the injection of the `Trace` merged in that bit that we factored out of this PR. Just following up on the rest of it, how should we proceed? As commented earlier, I'm unclear on the matter of factoring out the sub-registries changes, what the scope of that should be. Thanks!

PR #2339, opened 2025-07-28
test: Fix Android in-tree build wrt perfetto_flags_c_lib

ddiproietto:
It as broken since 8a5825d4a6f8("tp: use MurmurHash for base::FlatHashMap behind a flag (#2290)", with errors like:

```
FAILED: out/soong/.intermediates/external/perfetto/test/cts/reporter/CtsPerfettoReporterTestCases/android_x86_x86_64/obj/external/perfetto/test/cts/reporter/reporter_test_cts.o
...external/perfetto/test/cts/reporter/CtsPerfettoReporterTestCases/android_x86_x86_64/obj/external/perfetto/test/cts/reporter/reporter_test_cts.o external/perfetto/test/cts/reporter/reporter_test_cts.cc
In file included from external/perfetto/test/cts/reporter/reporter_test_cts.cc:26:
In file included from external/perfetto/test/test_helper.h:46:
In file included from external/perfetto/src/traced/probes/probes_producer.h:31:
In file included from external/perfetto/src/traced/probes/ftrace/ftrace_controller.h:33:
In file included from external/perfetto/src/traced/probes/ftrace/cpu_reader.h:27:
In file included from external/perfetto/include/perfetto/ext/base/flat_hash_map.h:21:
external/perfetto/include/perfetto/ext/base/flags.h:24:10: fatal error: 'perfetto_flags.h' file not found
   24 | #include <perfetto_flags.h>
      |          ^~~~~~~~~~~~~~~~~~
1 error generated.
```

A nice cleanup would be to list all these libraries into a single `default_cc` that's used by everyone depending on `perfetto_cts_deps`, but that's's work for another day, I guess.

Tested: `cd external/perfetto; mm`

### DIFF
--- a/test/cts/Android.bp
+++ b/test/cts/Android.bp
@@ -28,6 +28,7 @@ cc_test {
         "libprotobuf-cpp-lite",
         "libperfetto_client_experimental",
         "perfetto_cts_deps",
+        "perfetto_flags_c_lib",
         "perfetto_trace_protos",
     ],
     whole_static_libs: [
--- a/test/cts/art_module/Android.bp
+++ b/test/cts/art_module/Android.bp
@@ -22,6 +22,7 @@ cc_test {
     static_libs: [
         "libgmock",
         "perfetto_cts_deps",
+        "perfetto_flags_c_lib",
         "perfetto_trace_protos",
     ],
     whole_static_libs: [
--- a/test/cts/reporter/Android.bp
+++ b/test/cts/reporter/Android.bp
@@ -24,6 +24,7 @@ cc_test {
     "libprotobuf-cpp-lite",
     "libperfetto_client_experimental",
     "perfetto_cts_deps",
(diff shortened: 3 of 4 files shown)


ddiproietto:
> Approving as-is but would highly prefer if we added inline "TODO(): this is needed for perfetto_cts_deps, consider adding a cc_defaults" comments as per your description[1].
>
> Otherwise it'll be 6 months from now, we'll forget the exact details of this failure. And then one of us will have to reverse-engineer /why/ this dep is load-bearing.

I spoke too soon: it's not that simple. The basic problem is that static library dependencies are not propagated transitively (https://b.corp.google.com/issues/169779783). We cannot simply add one default, we have to add two (at least one for cts and one for vts). Ideally tools/gen_android_bp should be taught how to handle some of this.

I removed the comment from the commit message for now.

PR #2343, opened 2025-07-29
ui: Allow state to be persisted when loading a trace using permalink

stevegolton:


docs/visualization/deep-linking-to-perfetto-ui.md
```diff
@@ -62,6 +62,20 @@ print to a URL owned by you that would cause your dashboard to re-open the
 current trace, by re-kicking-off the window.open() process herein described.
 If omitted traces won't be shareable.
 
+`appStateHash` If you want your users to be able to share the UI state (e.g.
+pinned tracks, selections, etc), Perfetto can upload the state to the perfetto
+servers and restore it the next time you load the trace, with a little help from
+you.
+
+In order to get this working, Perfetto needs to be told the hash of the app
```
primiano: I'd rework this a little bit, as it might not be obvious. Maybe something like:

In order to get this working:
- At share time, perfetto will upload the app state (in a perfetto-controlled GCS bucket). the app state is identified by a hash.
- Perfetto can pass the hash back in the share url. this requires you to put a perfettoStateHashPlaceholder (see example below). The palceholder will be replaced with the actual hash at share time.
- When reopening the trace, you are supposed to pass back the actual hash in the URL (example below)

ui/src/frontend/permalink.ts
```diff
@@ -89,23 +91,43 @@ async function createPermalinkInternal(trace: TraceImpl): Promise<string> {
...
+ *
+ * @param trace The trace object to serialize and upload.
+ * @param traceUrl The URL of the trace file, if available. If undefined, only
+ * the app state will be uploaded.
+ * @returns The hash of the uploaded file, which can be used to create a
+ * permalink.
+ */
+export async function createPermalink(
+  trace: TraceImpl,
+  traceUrl: string | undefined,
```
primiano, line 120: we do we need to pass traceUrl? you can retrieve it from `trace`, seems redundant.
also was not passed before

ui/src/public/trace_source.ts
```diff
@@ -14,52 +14,50 @@
...
-  // See PostedTrace (which this interface extends).
+  readonly kind: 'HTTP_RPC';
 }
 
-export interface PostedTrace {
-  buffer: ArrayBuffer;
-  title: string;
-  fileName?: string;
-  url?: string;
+export interface TraceArrayBufferSource {
```
primiano: but now you are decoupling TraceArrayBufferSource and PostedTrace, which really should be in sync. why are you doing that?
You are creating a copy of two interfaces that are supposed to be the same

ui/src/frontend/permalink.ts
stevegolton: If you're referring to `trace.traceInfo.traceUrl`, that's not the same thing.

I split createPermalink() into two parts:
- uploadTraceBlob() which uploads the trace data
- createPermalink() which now just uploads the permalink JSON containing the trace url (usually the one we just uploaded using uploadTraceBlob()), and a trace object to serialize the state from.

I split this function into two so that I could reuse createPermalink() to upload the permalink for array buffer type without having to also upload the trace data.

So to upload the trace + state you now do this:
```ts
const traceUrl = await uploadTraceBlob(trace);
const hash = await createPermalink(trace, traceUrl);
```

If you just want to upload the UI state without the trace, do this:
```ts
const hash = await createPermalink(trace, undefined);
```

I'm open to better solutions if you have suggestions.

ui/src/public/trace_source.ts
stevegolton: This is very much intentional. The two serve different purposes.

- `PostedTrace` is the raw object injected via postmessage. It's the contract with the outside world and should remain fixed for backward compatibility.
- `TraceArrayBufferSource` is the args to the `app.openTrace()` function + defines the shape of `trace.traceInfo.source`. It can change if we need it to.

I separated the two because I needed to add different fields to each one. I needed to add a new `appStateHash` field to the `PostedTrace` so that external applications can inject the hash, but this is not required in `TraceArrayBufferSource` because the state is downloaded and parsed into a `SerializedAppState` object in the `postMessageHandler` before we call `openTrace`. This matches the behavior of when we open a trace from a permalink, where the state is downloaded and decoded before it hits `openTrace()`.

Even before this the two structures had already diverged: `keepApiOpen` is only used in postMessageHandler, not any time after so there is no point in it being in `TraceArrayBufferSource`.

I believe these two interfaces, while structurally similar, are semantically different. We don't **want** them to be kept in sync.

Additionally `openTraceFromBuffer` is called from a bunch of places, not just from the post message handler, so the args are necessarily more generic rather than being specific to posted traces.

Also, the dependency seemed to flow the wrong way.  `PostedTrace` is a concept isolated to postmessage trace loading, which is a high level concept - why should the foundational entities such as `Trace` and `App` have a dependency on that? If anything, the dependency should be the other way.

That said, I'm not willing to die on this hill. We could revert this change and go back to having `TraceArrayBufferSource` extend `PostedTrace`, maybe removing the unused fields using `Omit<PostedTrace, 'keepApiOpen'>` for instance, but that just felt messier to me than a few duplicated lines.

stevegolton:
> I like it overall, but i'm still a bit confused. To be honest one thing that confused me a lot is the fact that you mixed a lot of cosmetic changes (e.g. type->kind ) with the actual behavioral changes.
> 
> Please spit this CL and make it easier to understand what the actual changes are.

I've pushed a new commit with the non-essential refactorings reverted.

docs/visualization/deep-linking-to-perfetto-ui.md
stevegolton: Sounds good - done!

stevegolton:
Pushed a new commit with the docs updated

ui/src/frontend/permalink.ts
primiano: Ahhh I see. ack.

docs/visualization/deep-linking-to-perfetto-ui.md
```diff
@@ -34,33 +34,68 @@ sending a 'PING' message until the opened window replies with a 'PONG'.
...
+*   `url` (optional): A URL for sharing the trace. See the "Sharing" section
+    below.
+*   `appStateHash` (optional): A hash for restoring the UI state when sharing.
+    See the "Sharing" section below.
+
+### Sharing Traces and UI State
+
+The `url` and `appStateHash` properties work together to allow users to share a
+link to a trace that, when opened, restores the UI to the same state (e.g. zoom
+level, selected event).
```
primiano: The main problem here is that if the trace blob is posted from the outside, we cannot retain it an upload on our end (doing so might violate any retention-policy of the original trace source). Instead, we display a URL to your dashboard/frontend, which you have to pass. This url needs to re-kick the postmessage flow to load the trace again.

ui/src/frontend/post_message_handler.ts
```diff
@@ -14,14 +14,47 @@
...
+
+  appStateHash?: string;
+
+  // |uuid| is set only when loading via ?local_cache_key=1234. When set,
+  // this matches global.state.traceUuid, with the exception of the following
+  // time window: When a trace T1 is loaded and the user loads another trace T2,
+  // this |uuid| will be == T2, but the globals.state.traceUuid will be
+  // temporarily == T1 until T2 has been loaded (consistently to what happens
+  // with all other state fields).
+  uuid?: string;
```
primiano: can this be actually passed form the outside? I think this is used only internally?

ui/src/frontend/trace_share_utils.ts
```diff
@@ -26,41 +26,89 @@ export function isShareable(trace: Trace) {
...
-      msg.push(m(CopyableLink, {url: traceUrl}));
-    }
-
-    showModal({
-      title: 'Cannot create permalink from external trace',
-      content: m('div', msg),
-    });
-    return;
-  }
+  // If the trace is not sharable (has been pushed via postMessage()) and
```
primiano: missing comment / cliffhanger

ui/src/frontend/trace_share_utils.ts
```diff
@@ -26,41 +26,89 @@ export function isShareable(trace: Trace) {
...
-      `The trace will be accessible by anybody with the permalink.`,
-  );
-  if (result) {
-    AppImpl.instance.analytics.logEvent('Trace Actions', 'Create permalink');
-    return await createPermalink(trace);
+        if (result) {
+          AppImpl.instance.analytics.logEvent(
+            'Trace Actions',
+            'Create permalink',
+          );
```
primiano: maybe move this into createPermalink? feels like it belongs more there, and also makes htis code a bit tidier

PR #2345, opened 2025-07-29
Add basic evdev event proto

HarryCutts:
This will be used to record input events emitted by the Kernel on Android. Since the evdev protocol is used by all Linux-based OSes, however, the proto is not placed in the Android-specific extensions.

Test: record an input trace:
```sh
    $ echo 'buffers: { size_kb: 100000 fill_policy: RING_BUFFER }
      data_sources: {
          config {
              name: "android.input.inputevent"
              android_input_event_config { mode: TRACE_MODE_TRACE_ALL }
          }
      }
      data_sources: {
          config {
              name: "android.surfaceflinger.layers"
              surfaceflinger_layers_config: {
                  mode: MODE_ACTIVE
                  trace_flags: TRACE_FLAG_INPUT
                  trace_flags: TRACE_FLAG_COMPOSITION
                  trace_flags: TRACE_FLAG_HWC
                  trace_flags: TRACE_FLAG_BUFFERS
                  trace_flags: TRACE_FLAG_VIRTUAL_DISPLAYS
              }
          }
      }
      duration_ms: 20000
      ' | adb shell perfetto -c - --txt \
          -o /data/misc/perfetto-traces/input.perfetto-trace
```
Then use Perfetto's traceconv to convert the trace to text protos,
and check for evdev_event entries.

Bug: 394861376


protos/perfetto/trace/evdev.proto
```diff
@@ -0,0 +1,54 @@
...
+  // Proto version of Linux's struct input_event. The meaning of types and codes
+  // are described in the Linux kernel documentation at
+  // https://www.kernel.org/doc/html/latest/input/event-codes.html.
+  //
+  // Next ID: 5
+  message InputEvent {
+    // The monotonic timestamp at which the event occurred, as reported by the
+    // kernel, in integer nanoseconds. If omitted, assume that it hasn't changed
+    // since the previous event.
+    optional uint64 kernel_timestamp = 1;
```
primiano, line 43: I think I know the answer, but just to double check: you want both the TracePacket timestamp (When the event is seen in userspace) and this (to track the HW timing) right?

primiano (approved):

protos/perfetto/trace/evdev.proto
HarryCutts: Yes, that's right — the `TracePacket` timestamp will store the "read time", when the userspace process picked up the event from the Kernel, while this field will store the time at which the Kernel read it from the device

LalitMaganti: When it comes to parsing all of this, whichever timestamp you want to put in the tables, you should ensure that is what is being passed through the sorted. There is a golden rule in trace processor that data must be inserted into tables in timestamp order and must not be "buffered" to insert later.

primiano: Hmm let's talk about that one second actually

@HarryCutts which of the two timestamps (1: TracePacket.timestamp; 2: this inner timestamp that comes from the kernel) do you plan to use to place the event in the timeline?

If 1, there is no need to do anything special. I assume you'll have the kernel timetamp as an argument when you select the event.

if 2, then what lalit wrote. but then i'd question why don't you make TracePacket.timestamp == kernel timestamp, and put the userspace timestamp in a nested field.

HarryCutts: That's a good point! I had a chat with some other folks on the team and we decided that it makes most sense for the `TracePacket.timestamp` values to reflect the time at which we do the processing (so, the "read time" in this case).

primiano: so, to double-check, you will put slices on the timeline (in the UI) at the time when they are processed, not at the time of the kernel event?

To reiterate: life is easier if the timestamp in TracePacket.timestamp ==. the timestamp that the UI will ultimately show on the timeline (everything else can go in slice details)

HarryCutts: Yes, that's the plan (though the main viewer for this data will be Winscope, as opposed to the normal Perfetto UI)

PR #2346, opened 2025-07-29
CI: install gcloud in sandbox Docker image.

KirillTim:
We later will use the gcloud CLI to access Google Cloud Storage.

### DIFF
--- a/infra/ci/sandbox/Dockerfile
+++ b/infra/ci/sandbox/Dockerfile
@@ -138,6 +138,17 @@ RUN set -ex; \
     rm -rf /var/lib/apt/lists/*; \
     rm -rf /root/.cache/ /usr/share/man/* /usr/share/doc/*;
 
+# Install gcloud
+RUN curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
+    gpg --dearmor -o /usr/share/keyrings/cloud.google.gpg; \
+    echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" \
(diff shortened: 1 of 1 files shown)


infra/ci/sandbox/Dockerfile
```diff
@@ -138,6 +138,17 @@ RUN set -ex; \
...
 
+# Install gcloud
+RUN curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | \
+    gpg --dearmor -o /usr/share/keyrings/cloud.google.gpg; \
+    echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" \
+    | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list; \
+    apt-get update && apt-get install -y google-cloud-cli;
+
+# Create folder for gcloud config
+RUN mkdir -p /tmp/gcloud-config && chmod -R 775 /tmp/gcloud-config;
```
primiano: where does the content for this dir come from? this will be overridden by sandbox_runner.py when it does

```
  # We use the tmpfs mount created by gce-startup-script.sh. The problem is that
  # Docker doesn't allow to both override the tmpfs-size and prevent the
  # "-o noexec".
  tmp_dir = '/tmp/' + SANDBOX_NAME
  subprocess.call(['rm', '-rf', tmp_dir])
  os.makedirs(tmp_dir, exist_ok=True)
  os.chmod(tmp_dir, 0o777)
  cmd += ['-v', '%s:/tmp' % tmp_dir]
```

primiano (approved):

Issue #2249, opened 2025-07-22
[UI] improve UX for power rail visualization


cphlipot1:
I would like to propose a set of changes to improve UX for power rail visualization in the Perfetto UI. In #910 we improved UX for power rail aggregations. this issue aims to capture UX improvements to the presentation of the power rail tracks on the timeline.

current state of UI for reference (test/data/power_rails.pb):

<img width="1704" height="258" alt="Image" src="https://github.com/user-attachments/assets/80f70710-bbc5-496e-8ab5-7065e926121f" />

I'm proposing a plugin to handle this with the following changes to behavior:
1. drop "power." prefix and "_uws" suffix from the track name.
2. give tracks proper units. This unit should change appropriately when switching from "value" or "delta" mode to "rate" mode (eg we should display "uW" as the unit for "rate" mode and not "uWS / S"
3. switch displayed units from uW to mW since in most common cases android phones will be operating in the 1-1000mW range. 
4. Add track groups for rails that share a [subsystem](https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto//protos/perfetto/trace/power/power_rails.proto;l=29) to provide better organization when there are large numbers of tracks.

is there any concerns with any of the above changes? i would be willing to contribute PRs to accomplish the above.

LalitMaganti:
From my perspective no concerns. @stevegolton is out until next Tuesday so please ping him on this thread next week and he can comment as well whether this fits with what he wants to do.

cphlipot1:
> give tracks proper units. This unit should change appropriately when switching from "value" or "delta" mode to "rate" mode (eg we should display "uW" as the unit for "rate" mode and not "uWS / S"

To accomplish this i think we would need to update the plugin apis, as i don't think this can be entirely handled within a new plugin given the current APIs. i've proposed  #2259 to allow plugin authors to override the displayed unit when the track is set to "rate" mode. 

Assuming everyone is going to be ok with the high level changes proposed, i don't think any of the other items would need changes to the plugin APIs themselves and could be entirely handled within the plugins using existing apis.

> switch displayed units from uW to mW since in most common cases android phones will be operating in the 1-1000mW range.

Somewhat related to this, we could also change the energy unit displayed for "value" and "delta" mode to be "mWh" which is sometimes used on techspecs for phones and batteries, instead of "mJ" or "mWS". I could include this as part of the change if you think it makes sense.

cphlipot1:
[...]

PR #2349, opened 2025-07-30
stdlib: add machine_id to android_power_rails_metadata

cphlipot1:
Because we want to use the stdlib in the PowerRails UI plugin, and
we want to ensure this plugin is machine id aware, add the machine
id to the power rail metadata.

Bug: #2249

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/power_rails.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/power_rails.sql
@@ -89,13 +89,16 @@ CREATE PERFETTO TABLE android_power_rails_metadata (
   -- Power rail track id. Alias of `counter_track.id`.
   track_id JOINID(track.id),
   -- Subsystem name that this power rail belongs to.
-  subsystem_name STRING
+  subsystem_name STRING,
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/stdlib/android/power_rails.sql
```diff
@@ -89,13 +89,16 @@ CREATE PERFETTO TABLE android_power_rails_metadata (
   -- Power rail track id. Alias of `counter_track.id`.
   track_id JOINID(track.id),
   -- Subsystem name that this power rail belongs to.
-  subsystem_name STRING
+  subsystem_name STRING,
+  -- Machine id of the device the power rail is associated with.
```
aMayzner: There is no need to mention `id`. Just "The device the power rail is associated with."

aMayzner (approved):

src/trace_processor/perfetto_sql/stdlib/android/power_rails.sql
cphlipot1: done.

LalitMaganti (approved):

PR #2353, opened 2025-07-30
ui: push trace into components that pulled it from AppImpl.instance

cdamus:
Push down into all of the various leaf components/widgets the trace that the UI is presenting, instead of letting them pull it from the singleton AppImpl instance.

Part of #2336.

----

Factored out of #2337 supporting multiple open traces at a time.

### DIFF
--- a/ui/src/plugins/dev.perfetto.SqlModules/index.ts
+++ b/ui/src/plugins/dev.perfetto.SqlModules/index.ts
@@ -20,7 +20,11 @@ import {App} from '../../public/app';
 import {PerfettoPlugin} from '../../public/plugin';
 import {Trace} from '../../public/trace';
 import {SqlModules} from './sql_modules';
-import {SQL_MODULES_DOCS_SCHEMA, SqlModulesDocsSchema, SqlModulesImpl} from './sql_modules_impl';
+import {
+  SQL_MODULES_DOCS_SCHEMA,
+  SqlModulesDocsSchema,
+  SqlModulesImpl,
+} from './sql_modules_impl';
 
 const docs = defer<SqlModulesDocsSchema>();
 
--- a/ui/src/components/details/add_ephemeral_tab.ts
+++ b/ui/src/components/details/add_ephemeral_tab.ts
@@ -17,7 +17,11 @@ import {Trace} from '../../public/trace';
 
 // TODO(primiano): this method probably shouldn't exist at all in favour
 // of some helper in the Trace object).
-export function addEphemeralTab(trace: Trace, uriPrefix: string, tab: Tab): void {
+export function addEphemeralTab(
+  trace: Trace,
+  uriPrefix: string,
+  tab: Tab,
+): void {
   const uri = `${uriPrefix}#${uuidv4()}`;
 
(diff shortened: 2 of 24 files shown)


ui/src/plugins/dev.perfetto.SqlModules/index.ts
```diff
@@ -26,13 +25,8 @@ let globSqlModules: SqlModules | undefined;
...
   static readonly id = 'dev.perfetto.SqlModules';
 
-  static onActivate(_: App): void {
-    // Load the SQL modules JSON file when the plugin when the app starts up,
-    // rather than waiting until trace load.
-    loadJson();
-  }
-
   async onTraceLoad(trace: Trace): Promise<void> {
+    loadJson(trace);
```
stevegolton: Loading the json here and not awaiting will cause races with the current trace implementation. I acknowledge that your subregistry changes in your original PR probably would have addressed this.

I don't think awaiting is the solution, as I recently moved this out of onTraceLoad into onActivate to speed up trace load times.

I'm not sure what the best solution is, I'll have a think and get back to you.

cdamus: Oh, I see what's happening. The problem is that the loading and parsing of the JSON asset is coupled with initializing the `SqlModulesImpl`, which now requires the trace. Commit 9f6a0119247ea4a3656f5d653620b140e48b2c2d separates these two stages and restores the previous flow in the plug-in.

ui/src/plugins/dev.perfetto.SqlModules/index.ts
```diff
@@ -14,35 +14,43 @@
...
+
   static onActivate(_: App): void {
     // Load the SQL modules JSON file when the plugin when the app starts up,
     // rather than waiting until trace load.
-    loadJson();
+    loadJson().then(docs.resolve.bind(docs));
   }
 
   async onTraceLoad(trace: Trace): Promise<void> {
+    docs.then((resolvedDocs) => {
```
cdamus, line 43: @stevegolton Would it be valuable to `await` this promise? Previously, the `globSqlModules` wasn't awaited anywhere and so there was always this pendancy handled at line 48 and the possibility of an `undefined` return at 76. But maybe now the cost of creating the `SqlModulesImpl` once the JSON has finished loading is small enough to make it useful to tighten up the synchronization here?

stevegolton: Great idea, I'm all for this!

martin-fleck-at:
@stevegolton I resolved the merge conflicts an pushed an update.

LalitMaganti:
https://github.com/google/perfetto/actions/runs/17210191383/job/48821932946: looks like there are some integration test failures.

martin-fleck-at:
@LalitMaganti Thank you for following up! I am looking at them right now, hopefully just a minor thing to fix!

martin-fleck-at:
@LalitMaganti CI is happy now ;-) Is there anything else I need to do?

stevegolton:
Thanks very much @martin-fleck-at for landing this!

PR #2354, opened 2025-07-30
Add hypervisor events introduced in android16-6.12

nioffe:
Starting with android16-6.12 kernel the pkvm hypervisor tracing instance was renamed from /sys/kernel/tracing/hyp/events/hyp/* to /sys/kernel/tracing/hypervisor/events/hypervisor/*. This patch adds new events to Perfetto. Since there is a name conflict, the events are special cased in src/tools/ftrace_proto_gen/ftrace_proto_gen.cc

Bug: 428904926


src/traced/probes/ftrace/test/data/synthetic/events/hypervisor/host_hcall/format
```diff
@@ -0,0 +1,9 @@
+name: host_hcall
```
rsavitski: Can me move all of the hypervisor (and hyp/ for that matter, unless something breaks) events into `.../src/traced/probes/ftrace/test/data/hypervisor/events/...`. This is breaking tests through roundabout means because the tests load format files from the /synthetic dir, and there's now a collision on ID.

I assume the script will do the right thing:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/tools/run_ftrace_proto_gen;l=31;drc=7ea83c9fafd2a3af90691607490297046fa71a56

rsavitski (changes requested):

src/traced/probes/ftrace/test/data/synthetic/events/hypervisor/host_hcall/format
nioffe: Done. PTAL

rsavitski (approved):

PR #2356, opened 2025-07-30
Add logic to special case "hypervisor" instance

nioffe:
Starting from android16-6.12 the hypervisor tracing instance was rename from /sys/kernel/tracing/hyp to /sys/kernel/tracing/hypervisor. Add logic to special case new instance in the same way as the hyp one


src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -766,7 +766,6 @@ bool CpuReader::ParseEvent(
   const Event& info = *table->GetEventById(ftrace_event_id);
 
   if (info.size > static_cast<size_t>(end - start)) {
-    PERFETTO_DLOG("Expected event length is beyond end of buffer.");
```
rsavitski: I don't see a reason to remove a DLOG. Please keep it.

src/traced/probes/ftrace/ftrace_controller.cc
```diff
@@ -880,11 +880,11 @@ std::optional<std::string> FtraceController::AbsolutePathForInstance(
 
   // ARM64 pKVM hypervisor tracing emulates an instance, but is not under
   // instances/, we special-case that name for now.
-  if (raw_cfg_name == "hyp") {
-    std::string hyp_path = tracefs_root + "hyp/";
+  if (raw_cfg_name == "hyp" || raw_cfg_name == "hypervisor") {
```
rsavitski, line 883: Can we also add the new special-case to the proto comment here:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/config/ftrace/ftrace_config.proto;l=261;drc=35cc82bb94825ac859e829d0e9c2604a21a496c2

(you'll need to re-run tools/gen_all afterwards)

src/traced/probes/ftrace/cpu_reader.cc
nioffe: That was an accidental removal. DLOG added back

src/traced/probes/ftrace/ftrace_controller.cc
nioffe: Done

PR #2361, opened 2025-07-31
perfetto: Add flag to control offline CPU buffer clearing

sashwinbalaji:
Add flag `ftrace_clear_offline_cpus_only`

This flag(enabled by default) makes it possible to disable the optimization that clears trace buffers for only offline CPUs. 

When the flag is disabled, the system will revert to the previous behavior of clearing the buffers for all CPUs

### DIFF
--- a/src/traced/probes/ftrace/tracefs.cc
+++ b/src/traced/probes/ftrace/tracefs.cc
@@ -25,7 +25,6 @@
 
 #include "perfetto/base/logging.h"
 #include "perfetto/ext/base/file_utils.h"
-#include "perfetto/ext/base/flags.h"
 #include "perfetto/ext/base/string_splitter.h"
 #include "perfetto/ext/base/string_utils.h"
 #include "perfetto/ext/base/utils.h"


include/perfetto/ext/base/flags.h
```diff
@@ -32,9 +32,11 @@ namespace perfetto::base::flags {
 // in `perfetto_flags.aconfig`.
 // The second argument is the default value of the flag in non-Android platform
 // contexts.
-#define PERFETTO_READ_ONLY_FLAGS(X)                       \
-  X(test_read_only_flag, NonAndroidPlatformDefault_FALSE) \
-  X(use_murmur_hash_for_flat_hash_map, NonAndroidPlatformDefault_TRUE)
+#define PERFETTO_READ_ONLY_FLAGS(X)                                       \
+  X(test_read_only_flag, NonAndroidPlatformDefault_FALSE)                 \
+  X(use_murmur_hash_for_flat_hash_map, NonAndroidPlatformDefault_TRUE)    \
+  X(clear_only_offline_cpu_trace_buffers, NonAndroidPlatformDefault_TRUE) \
```
rsavitski: leftover from previous version?

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -503,6 +504,13 @@ void Tracefs::ClearTrace() {
   std::string path = root_ + "trace";
   PERFETTO_CHECK(ClearFile(path));  // Could not clear.
 
+  if (!base::flags::ftrace_clear_offline_cpus_only) {
```
rsavitski: Can we put the "new" logic behind the flag instead, and keep the old code as both the unflagged and the fallback codepath?

Between the two PRs, we're adding more lines of code to this function than I'd expect, there should be a more straightforward code layout.

include/perfetto/ext/base/flags.h
sashwinbalaji: Oops, thanks for catching.

src/traced/probes/ftrace/tracefs.cc
sashwinbalaji: Ack, modified as suggested. The original reason for duplicating the clear all cpu for loop was to keep both the case completely separate.

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -504,34 +505,38 @@ void Tracefs::ClearTrace() {
...
-    for (const auto& cpu : offline_cpus.value()) {
-      ClearPerCpuTrace(cpu);
-    }
-  } else {
-    // Fallback: if we can't determine which CPUs are offline, clear the buffers
-    // for all possible CPUs.
-    for (size_t cpu = 0; cpu < total_cpu_count; cpu++) {
-      ClearPerCpuTrace(cpu);
+
+  if (base::flags::ftrace_clear_offline_cpus_only) {
```
LalitMaganti: if constexpr to avoid even compiling this code.

sashwinbalaji: Done. Thanks for the suggestion!

rsavitski (approved):

PR #2365, opened 2025-07-31
Add "on use" mode for packages list data source.

simonmacm:
This polls for CPU-per-UID use and emits packets only for the UIDs it sees.

We move half of the CPU-per-UID source into a utility class for reuse.

We also separate the package list data source into a list parsing phase and a packet writing phase.

Bug: b/422976206


protos/perfetto/config/android/packages_list_config.proto
```diff
@@ -24,4 +24,10 @@ message PackagesListConfig {
   // If not empty, emit info about only the following list of package names
   // (exact match, no regex). Otherwise, emit info about all packages.
   repeated string package_name_filter = 1;
+
+  // If present, the data source will periodically poll for CPU use by packages
+  // and only emit results for those that it sees. If absent, the data source
+  // will emit results for all packages at startup. The package name filter
+  // applies either way.
+  optional uint32 on_use_poll_ms = 2;
```
ddiproietto: Maybe `only_dump_on_cpu_use_every_ms`? Naming is hard

protos/perfetto/config/android/packages_list_config.proto
```diff
@@ -24,4 +24,10 @@ message PackagesListConfig {
   // If not empty, emit info about only the following list of package names
   // (exact match, no regex). Otherwise, emit info about all packages.
   repeated string package_name_filter = 1;
+
+  // If present, the data source will periodically poll for CPU use by packages
```
ddiproietto: Nit: I feel that for protobuf primitive types, checking for presence is kind of an antipattern. Can we just say non-zero (I believe it's consistent with most other places, unless you have another example in our config, in which case, ignore me).

src/traced/probes/packages_list/packages_list_data_source.cc
```diff
@@ -16,89 +16,201 @@
...
+
 // static
 const ProbesDataSource::Descriptor PackagesListDataSource::descriptor = {
     /*name*/ "android.packages_list",
     /*flags*/ Descriptor::kFlagsNone,
     /*fill_descriptor_func*/ nullptr,
 };
 
-bool ParsePackagesListStream(protos::pbzero::PackagesList* packages_list_packet,
+bool ParsePackagesListStream(base::FlatHashMap<uint64_t, Package>& packages,
```
ddiproietto: There are many packages with the same UID (I had a quick check on my acloud device), I don't think we can use a map like this here.

ddiproietto (commented):
Thanks!

Some preliminary comments.

It looks like a couple of unittests are failing, could you please check those? (I'm not sure why CI hasn't run)

protos/perfetto/config/android/packages_list_config.proto
simonmacm: Switched to `only_write_on_cpu_use_every_ms`.

simonmacm:
> It looks like a couple of unittests are failing, could you please check those? (I'm not sure why CI hasn't run)

Fixed.

src/traced/probes/packages_list/packages_list_data_source.cc
simonmacm: Switched to `std::unordered_multimap`.

protos/perfetto/config/android/packages_list_config.proto
simonmacm: I've done this sort of thing all the time in google3 but if it's not the local pattern then fair enough, changed.

src/traced/probes/packages_list/packages_list_data_source.cc
```diff
@@ -16,89 +16,205 @@
...
+    // The first time we poll the poller it will return details of all UIDs,
+    // even ones whose last activity was days ago, so we wait until a subsequent
+    // time for evidence something has actually run.
+    first_time_ = false;
+    return;
+  }
+
+  std::vector<uint32_t> new_uids;
+  for (auto& time : cpu_times) {
+    if (seen_uids_.find(time.uid) == seen_uids_.end()) {
```
ddiproietto: Nit: I'm not sure this is actually more readable, but, internal guidelines suggest to avoid redundant map lookups:

go/totw/132#counting-collisions

In this case:

```
if(seen_uids_.insert(time.uid).second) {
  new_uids.push_back(time.uid)
}
```

src/traced/probes/packages_list/packages_list_data_source.cc
```diff
@@ -16,89 +16,205 @@
...
+      seen_uids_.insert(time.uid);
+      new_uids.push_back(time.uid);
+    }
+  }
+
+  if (new_uids.size() > 0) {
```
ddiproietto: Nit: `!new_uids.empty()`

src/traced/probes/packages_list/packages_list_data_source.cc
```diff
@@ -16,89 +16,205 @@
...
+  if (only_write_on_cpu_use_every_ms_ == kWriteAllAtStart) {
+    // Flush is no-op. We flush after the first write.
+    callback();
+  } else {
+    writer_->Flush(callback);
+    poller_->Clear();
+  }
+}
+
+void PackagesListDataSource::ClearIncrementalState() {
```
ddiproietto, line 212: I think the only purpose of ClearIncrementalState is to make sure that if all the previous packets emitted were dropped (by ring buffer), the reader would still have a complete picture from this point on.

The only thing I would do here is to clear `seen_uids_`.

Setting first_time_ to true means that we would ignore all the packages seen since the last tick.

Right?

ddiproietto (approved):
Thanks. All minor optional nits, except the comment on ClearIncrementalState.

Have you tested that trace_processor likes this? (I looked at the code and it should)

src/traced/probes/packages_list/packages_list_data_source.cc
simonmacm: Done

src/traced/probes/packages_list/packages_list_data_source.cc
simonmacm: Done

src/traced/probes/packages_list/packages_list_data_source.cc
simonmacm: Setting first_time_ to true is necessary if we reset the poller.

But you're right, we don't need to do that. Done.

simonmacm:
> Have you tested that trace_processor likes this? (I looked at the code and it should)

Yes, it seems to merge multiple packages_list packets happily.

PR #2370, opened 2025-07-31
github: use GCS to store 'buildtools/' cache.

KirillTim:
We use `tar` to create a single cache file from multiple files in `buildtools/`
directory, so we can rely on the atomicity of `gcloud storage cp` command.
This prevents races during concurrent builds, ensuring only one writer can 
successfully update the cache at a time.

When saving and restoring the cache, we use the `--keep-old-files` and
`--exclude-from=` `tar` flags to avoid overwriting files inside
`buildtools/` that are already stored in git.

We store a single, the most recent cache entry for each build type: we
update the cache only when building the main branch.

The content of `buildtools/` is rarely updated, so instead of uploading the
cache after each build, we compare the hashes of its base files to check
if an update is needed.

An alternative was to calculate the checksum of the entire tar content
using
`tar --extract --to-stdout --file="$FILE" | sha256sum | awk '{print $1}'`.
However, this was not feasible as it takes approximately 30 seconds to
compute on CI for the Android build (it contains multiple large binary blobs).


.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,107 @@ inputs:
...
-
-    - name: install-build-deps
-      run: tools/install-build-deps ${{ inputs.install-flags }}
+    # When restoring the buildtools cache, we don't want to overwrite files or folders that are stored in git.
+    - name: Generate list of files to ignore
+      id: files_to_ignore
+      shell: bash
+      run: |
+        # Generate list of files in 'buildtools/' stored in git.
+        FILES_TO_IGNORE_PATH=$(mktemp /tmp/files-to-ignore-XXXXX.txt)
```
primiano, line 53: why isn't this a fixed path like /tmp/buildtools.ignore?
what's the benefit or mktemp? it adds extra complexity and indirection layer (because now you need a var)

.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,107 @@ inputs:
...
+        git ls-tree HEAD:buildtools --name-only > "$FILES_TO_IGNORE_PATH"
+        echo "file_path=$FILES_TO_IGNORE_PATH" >> "$GITHUB_OUTPUT"
+
+    - name: Restore buildtools from cache
+      id: restore_buildtools_cache
+      shell: bash
+      env:
+        GCS_BUCKET: gs://perfetto-ci-artifacts
+        CACHE_PATH: cache/buildtools-${{ steps.cachekey.outputs.flags_key }}/buildtools.tar
+        FILES_TO_IGNORE_PATH: ${{ steps.files_to_ignore.outputs.file_path }}
```
primiano, line 62: just use a /tmp/path.... so many variables here

.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,107 @@ inputs:
...
+        LOCAL_BUILDTOOLS_HASH: ${{ hashFiles('buildtools/**', 'tools/install-build-deps') }}
+        CACHED_BUILDTOOLS_HASH_FILE_PATH: buildtools/buildtools_hash_revision_for_cache_on_ci.txt
+      run: |
+        # Restore content of 'buildtools/' folder from cache.
+        CACHED_TAR_PATH=$(mktemp /tmp/buildtools-cache-XXXXXX.tar)
```
primiano, line 67: ditto. why using mktemp rather than a fixed path?

.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,107 @@ inputs:
...
+    - name: Save buildtools cache (only on main)
       if: github.ref == 'refs/heads/main'
-      uses: actions/cache/save@v4
-      with:
-        path: buildtools
-        # TODO(lalitm): see the comment above about the cache key.
-        key: buildtools-${{ hashFiles('buildtools/BUILD.gn') }}-${{ steps.cachekey.outputs.suffix }}-${{ hashFiles('tools/install-build-deps') }}
+      shell: bash
+      env:
+        GCS_BUCKET: gs://perfetto-ci-artifacts
```
primiano: can this be a global env one level up? feeld weird that we have to set it twice, error prone

.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,107 @@ inputs:
...
+        FILES_TO_IGNORE_PATH: ${{ steps.files_to_ignore.outputs.file_path }}
+        UPDATE_CACHE: ${{steps.restore_buildtools_cache.outputs.update_cache  }}
+      run: |
+        # Save content of 'buildtools/' folder to cache.
+        TAR_PATH=$(mktemp /tmp/buildtools-XXXXXX.tar)
+
+        # '--exclude-from' is used to not cache files or folders that are stored in git.
+        # '--sort=name' is used to make archive more reproducible.
+        tar --create --sort=name --preserve-permissions --exclude-from="$FILES_TO_IGNORE_PATH" --file="$TAR_PATH" --directory=buildtools .
+
```
primiano, line 118: can we avoid the tar if there is no need to reupload? you can jsut look at the hash right?

.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,107 @@ inputs:
...
+    - name: Restore buildtools from cache
+      id: restore_buildtools_cache
+      shell: bash
+      env:
+        GCS_BUCKET: gs://perfetto-ci-artifacts
+        CACHE_PATH: cache/buildtools-${{ steps.cachekey.outputs.flags_key }}/buildtools.tar
+        FILES_TO_IGNORE_PATH: ${{ steps.files_to_ignore.outputs.file_path }}
+        LOCAL_BUILDTOOLS_HASH: ${{ hashFiles('buildtools/**', 'tools/install-build-deps') }}
+        CACHED_BUILDTOOLS_HASH_FILE_PATH: buildtools/buildtools_hash_revision_for_cache_on_ci.txt
```
primiano, line 64: This creates an unversioned file in .git
add it to .gitignore plz in this same cl

.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,107 @@ inputs:
...
+        # Save content of 'buildtools/' folder to cache.
+        TAR_PATH=$(mktemp /tmp/buildtools-XXXXXX.tar)
+
+        # '--exclude-from' is used to not cache files or folders that are stored in git.
+        # '--sort=name' is used to make archive more reproducible.
+        tar --create --sort=name --preserve-permissions --exclude-from="$FILES_TO_IGNORE_PATH" --file="$TAR_PATH" --directory=buildtools .
+        
+        # Content of 'buildtools/' rarely changes, no need to upload the cache it on each build, even if nothing changes.
+        if [[ "$UPDATE_CACHE" == "true" ]]; then
+          gcloud storage cp "$TAR_PATH" "$GCS_BUCKET/$CACHE_PATH"
```
primiano: given you tar without gz I think it would be worth to pass -J to cp, as it compresses on the flight while uplaoding.
Test locally as it might be beneficial also at download time. the whole compression topic is always very confusing and i don't remember the rules, as gcs supports transparent compression using Content-Encoding, but gsutil needs to be told somehow

primiano (commented):
Some comments but thanks for looking into this.
Since you are here, in a separate CL it might be worth also migrating over the other uses of actions/cache/restore@v4/save for ccache

.github/actions/install-build-deps/action.yml
KirillTim: > why isn't this a fixed path like /tmp/buildtools.ignore?

This is a best practice for writing shell scripts :)
We don't want accidentally overwrite already existing file, especially since the code here is a github action, that is re-used by other workflows that may have file with the same name.

I'd really like to keep the files, especially since we later will extract this code to the even more generic "cache" acthon.

.github/actions/install-build-deps/action.yml
KirillTim: See my reply here: https://github.com/google/perfetto/pull/2370#discussion_r2248355920

We don't want to overwrite the users (our own) files by mistake.

.github/actions/install-build-deps/action.yml
KirillTim: I don't think that we need to compress the buildtools cache in-flight.
Both upload and download takes seconds and the biggest files are not much compressible binary blobs (android emulator images). We upload the single tar, so the only option available for us is `--gzip-in-flight-all`, here are some numbers from my cloudtop (it is a gcloud instance, so the number are close to the CI):
```
➜  /tmp/perfetto % tools/install-build-deps --android
➜  /tmp/perfetto % du -sh buildtools 
5.0G    buildtools

➜  /tmp/perfetto % time gcloud storage cp --storage-class=STANDARD buildtools-android.tar gs://perfetto-ci-artifacts/cache/ktimofeev/buildtools-android.tar

Copying file://buildtools-android.tar to gs://perfetto-ci-artifacts/cache/ktimofeev/buildtools-android.tar
  Completed files 32/1 | 4.9GiB/4.9GiB | 455.4MiB/s                                                                                                                                                                                                                                                                

Average throughput: 577.2MiB/s
time 18.32s user 6.33s system 192% cpu 12.802 total

➜  /tmp/perfetto % time gcloud storage cp --gzip-in-flight-all --storage-class=STANDARD buildtools-android.tar gs://perfetto-ci-artifacts/cache/ktimofeev/buildtools-android-faster.tar

Copying file://buildtools-android.tar to gs://perfetto-ci-artifacts/cache/ktimofeev/buildtools-android-faster.tar
  Completed files 32/1 | 4.9GiB/4.9GiB | 469.1MiB/s                                                                                                                                                                                                                                                                

Average throughput: 571.9MiB/s
time 18.25s user 6.63s system 188% cpu 13.190 total
```

Just in case, compressing the entire tar locally take minutes on my 96 cores machine:
```
➜  /tmp/perfetto % time tar --create --gzip --sort=name --preserve-permissions --file=buildtools-android.tar.gz --directory=buildtools .
tar --create --gzip --sort=name --preserve-permissions  --directory=buildtool  183.97s user 8.91s system 103% cpu 3:05.89 total
➜  /tmp/perfetto % du -sh buildtools-android.tar.gz 
1.8G    buildtools-android.tar.gz
```

.github/actions/install-build-deps/action.yml
KirillTim: We can't set an action level variable unfortunately, the best we can do is: c9e9fbb9ea65219d69028367ed2781b07dff5e62
Seems to be good enough for me.

.github/actions/install-build-deps/action.yml
KirillTim: Done.

.github/actions/install-build-deps/action.yml
KirillTim: Done.

.github/actions/install-build-deps/action.yml
KirillTim: See my reply here: https://github.com/google/perfetto/pull/2370#discussion_r2248355920

We don't want to overwrite the users (our own) files by mistake.

.github/actions/install-build-deps/action.yml
```diff
@@ -22,41 +22,125 @@ inputs:
...
   using: "composite"
   steps:
+    - name: Setup Google Cloud config
+      shell: bash
+      run: |
+        # Setup Google Cloud config.
+        # Token is generated in 'infra/ci/worker/sandbox_runner.py'.
+        gcloud config set auth/access_token_file $SVC_TOKEN_PATH
+        # Suppress warning about "parallel composite upload".
+        gcloud config set storage/parallel_composite_upload_enabled True
```
primiano: Just move it to sandbox_entrypoint.sh

KirillTim: https://github.com/google/perfetto/pull/2402

PR #2371, opened 2025-07-31
Fix data type of table columns in bitmap std lib

archisha-baranwal:
Bug: [414346795](b/414346795)


src/trace_processor/perfetto_sql/stdlib/android/bitmaps.sql
```diff
@@ -81,15 +81,15 @@ ORDER BY
 -- category.
 CREATE PERFETTO TABLE android_bitmap_count (
   -- ID of the row in the underlying counter table.
-  id LONG,
+  id ID(counter.id),
   -- Upid of the process.
-  upid LONG,
+  upid ID(process.upid),
```
aMayzner: I'm quite sure this is JOINID(process.id), not ID(process.id). Is this unique and sorted in the table? Here and above

archisha-baranwal: ah, my bad. fixed now.

PR #2374, opened 2025-08-01
ui: Fix content overflow in current selection panel's GridLayoutColumn

KingAlen:
When rendering the args tree with renderArguments, long value texts overflow outside the GridLayoutColumn component. This update adds CSS styling to enable automatic word-level wrapping, ensuring the content stays within the column boundaries.

Before: 
before-gridlayoutcolumn

After:
after-gridlayoutcolumn


KingAlen:
@stevegolton PTAL

Issue #2381, opened 2025-08-01
Optional Diff tests based on configuration


OmarElfouly:
Can we please add some plumbing to diff tests to have optional diff tests based on configuration? Honestly it's high time we have this anyway, it's something we've been lacking for a looooongggg time.

I'd work with Anna to make this happen, she rewrote a lot of the diff test infra so should understand how we should do this.

_Originally posted by @LalitMaganti in https://github.com/google/perfetto/issues/2375#issuecomment-3145328234_

JoeNotCharlesGoogle:
The Chrome diff tests are currently broken because of this. They run a subset of tests that don't include any `Etm` tests, but do include `StdlibSmoke`, which does "INCLUDE *" to make sure the whole stdlib builds. Since the stdlib now includes `etm.sql` unconditionally, and Chrome builds with `enable_perfetto_etm_importer = false`, the StdlibSmoke test fails.

I tried to fix it with

```
if (enable_perfetto_etm_importer) {
  sources += [ "etm.sql" ]
}
```

But the conditional is ignored by the generated Android.bp and BUILD files, so this would break non-GN builds.

LalitMaganti:
> But the conditional is ignored by the generated Android.bp and BUILD files, so this would break non-GN builds.

This won't break non-GN builds. It's WAI that those builds don't have etm code.

PR #2375, opened 2025-08-01
Adds a symbolize function to stdlib

OmarElfouly:
Adds a symbolize function that takes a table with the columns "file_name" "rel_pc" "mapping_id" "address" and then symbolizes each row using llvm_symbolizer  and returns function_name, file_name, line_number, mapping_id, address.

Currently includes mapping_id and address as a way to join back symbolization results to original data.

Follows the pattern of interval_intersect and counter_intervals.

PR #2382, opened 2025-08-01
tp: stdlib: adds symbolize and etm diff tests

OmarElfouly:
This PR adds diff tests for SQL symbolize.

This can only be merged when there is support for optional diff tests based on configuration. #2381

https://github.com/google/perfetto/pull/2375#issuecomment-3145328234


OmarElfouly:
This PR is also responsible for adding back the ETM testsuite with the correct plumbing to avoid #2478

OmarElfouly:
Was already merged into #2574

PR #2386, opened 2025-08-01
tp: cleanup includes and clang-tidy warnings in several importers

LalitMaganti:


LalitMaganti:
> is this the set of warnings from ".clang_tidy" in our repo, or your own [clangd] cfg?

From my own file.

> what was the incantation that worked in the end? My attempts to run clang-tidy in a batch fashion stumble over header filtering.

This is my file (just use clangd VSCode plugin) in :

```
Diagnostics:
  UnusedIncludes: Strict
  MissingIncludes: Strict
  ClangTidy:
    Add: [performance*, modernize*, readability*]
    Remove:
      [
        readability-identifier-length,
        modernize-use-nodiscard,
        modernize-use-trailing-return-type,
        readability-named-parameter,
        readability-implicit-bool-conversion,
        readability-magic-numbers,
        readability-uppercase-literal-suffix,
        bugprone-unchecked-optional-access,
        readability-function-cognitive-complexity,
        readability-braces-around-statements,
        readability-avoid-return-with-void-value,
      ]
Index:
  StandardLibrary: Yes
CompileFlags:
  Add: [-isystem, /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1]
```

rsavitski:
Right. I do have a similar clangd setup as well. Part of my question is whether we should move some of this into .clang_tidy so that people are trying to stick to the same set of checks. But not important atm.

PR #2387, opened 2025-08-01
tp: use kBatteryStatsBlueprint for battery stats sys props

cphlipot1:
The System Property data source can collect battery stats related
system properties. Perfetto can either poll these, or read them once
when the trace starts in order to provide the counter initial values.
In the later case the atrace "power" category can be used to provide
events when these change.

Prior to track typing/blueprinting both data sources wrote to track
of the same name. Now desipte the tracks still having the same name
the they are different tracks, so initial values are placed on entirely
separate tracks.

Switch the system property importing to use the same track as the
systrace parsing so the old behavior is preserved. We also add new
code comments to help ensure these are kept in sync with any future
refractors.

# Before

Two tracks for charging. one with initial state and one with state changes during the trace
image

# After

single track for charging that contains both initial state and later state updates:

image


LalitMaganti:
I think better than a comment would be adding a diff test to make sure they show up on the same track.

cphlipot1:
Added a diff test. I could not find an existing location where this kind of functionality was tested, so created a new test under diff_tests/parser/power. Let me know if you prefer a different location for such a test.

LalitMaganti:
Looks good thanks!

Issue #554, opened 2023-08-10
A lot of MSVC code defines `interface` as `struct`


ddiproietto:
Found in https://github.com/google/perfetto/issues/540

Unfortunately we use `interface` in our protobufs as https://github.com/google/perfetto/blob/aa44034ca2b2af2b32cf7701aa9afdff69189282/protos/perfetto/trace/android/network_trace.proto#L34

We should consider renaming this

PR #2397, opened 2025-08-04
fix mingw builds

simeonschaub:
Upstreaming patches from JuliaPackaging/Yggdrasil#11770. One remaining
issue is that `NetworkPacketEvent_Decoder` in the sdk header defines a
member function with the name `interface`, which conflicts with a mingw
macro, but I wasn't able to figure out from where that definition was
generated

### DIFF
--- a/include/perfetto/base/platform_handle.h
+++ b/include/perfetto/base/platform_handle.h
@@ -29,7 +29,7 @@ namespace base {
 // Windows has two types of "handles", which, in UNIX-land, both map to int:
 // 1. File handles returned by the posix-compatibility API like _open().
 //    These are just int(s) and should stay such, because all the posix-like API
-//    in Windows.h take an int, not a HANDLE.
+//    in windows.h take an int, not a HANDLE.
 // 2. Handles returned by old-school WINAPI like CreateFile, CreateEvent etc.
 //    These are proper HANDLE(s). PlatformHandle should be used here.
 //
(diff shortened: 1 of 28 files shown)


LalitMaganti (approved):

LalitMaganti:
Please run `tools/format-sources`: presubmit is failing because of it: https://github.com/google/perfetto/actions/runs/16720958393/job/47325383179?pr=2397

simeonschaub:
@LalitMaganti Any ideas regarding where `NetworkPacketEvent_Decoder` is defined? We are currently carrying the following patch: https://github.com/JuliaPackaging/Yggdrasil/blob/8f7d9fa972d817f3a6e274229e65e56fa7e42ec5/P/perfetto/bundled/header_mingw.patch

LalitMaganti:
Generated from proto at https://github.com/google/perfetto/blob/21bd92c18580ba83eef30ef2430833f567e76ab6/protos/perfetto/trace/android/network_trace.proto#L34

Changing this might be tricky as I think there are dependencies from code inside the android platform. Will chat with some folks internally about it

simeonschaub:
Ah, right! Just found #554 as well

simeonschaub:
Thanks for the quick merge!

LalitMaganti:
@simeonschaub the interface issue will be fixed by https://github.com/google/perfetto/pull/2423

PR #2402, opened 2025-08-04
github: set up a GCS CLI access token inside Docker.

KirillTim:
We need this to be able to run `gcloud storage cp ...` commands in GitHub actions.


infra/ci/sandbox/sandbox_entrypoint.sh
```diff
@@ -13,19 +13,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eu
+set -eux -o pipefail
```
primiano: -x will create a lot of noise and also leak the token in the logs. i'd avoid it

PR #2403, opened 2025-08-04
Move android_memory_dmabuf_cumulative table to std lib

archisha-baranwal:
Bug: [436475182](b/436475182)

The `dmabuf allocs` track is added via a perfetto plugin which creates the table android_memory_dmabuf_cumulative table to do so.
We have moved this table to std lib to enable its usage for metric generation.


src/trace_processor/perfetto_sql/stdlib/android/memory/dmabuf.sql
```diff
@@ -144,3 +146,35 @@ LEFT JOIN process
   USING (upid)
 ORDER BY
   ts;
+
+-- Provides a timeseries of "dmabuf allocs" counter for each process.
+CREATE PERFETTO TABLE android_dmabuf_allocs_counter_per_process (
```
LalitMaganti: Would suggest making this a new module (dmabuf_counter) to avoid having to run the expensive code above to be able to create this table.

src/trace_processor/perfetto_sql/stdlib/android/memory/dmabuf.sql
```diff
@@ -144,3 +146,35 @@ LEFT JOIN process
   USING (upid)
 ORDER BY
   ts;
+
+-- Provides a timeseries of "dmabuf allocs" counter for each process.
```
LalitMaganti: Please add a note explaining what data source needs to be enabled in the trace to have this table populated following the example of the `android_bitmap_memory` table  (https://perfetto.dev/docs/analysis/stdlib-docs#android-bitmaps)

archisha-baranwal: Done. Added the required ftrace event needed for the new table.

src/trace_processor/perfetto_sql/stdlib/android/memory/dmabuf.sql
archisha-baranwal: Since we are now moving the android_memory_cumulative_dmabuf table to std_lib as per the discussion [here](https://b.corp.google.com/issues/435136511#comment5) I believe it makes sense to keep this table in the same module.

src/trace_processor/perfetto_sql/stdlib/android/memory/dmabuf.sql
```diff
@@ -144,3 +146,24 @@ LEFT JOIN process
   USING (upid)
 ORDER BY
   ts;
+
+-- Provides a timeseries of dmabuf allocations for each process.
+-- To populate this table, tracing must be enabled with the "dmabuf_allocs" ftrace event.
+CREATE PERFETTO TABLE android_memory_cumulative_dmabuf (
+  -- upid of process responsible for the allocation (matches utid)
```
LalitMaganti, line 153: Would suggest exposing the thread name and process name here as well using left joins to make this table more useful to query.

archisha-baranwal: The android_dmabuf_allocs table already has these 2 columns. I have added those in the new table as well.

LalitMaganti (approved):

PR #2405, opened 2025-08-04
Add support for exclusive tracing session.

sashwinbalaji:
This change introduces the concept of "exclusive" tracing sessions, controlled by the `exclusive_mode_priority` field in TraceConfig.

When set to a value > 0, this tracing session will be started in "exclusive mode". This has the following semantics:
- It can only be set by shell or root users.
- A new exclusive session will only be started if its priority is strictly  higher than any other active exclusive session.
- If a new exclusive session is started, all other existing tracing sessions (exclusive or not) are aborted.

We leverage the existing error argument in `OnTracingDisabled` to notify the consumers when their session is terminated by a higher-priority one.

Design doc: [go/perfetto-exclusive](http://go/perfetto-exclusive)

Test: Tested locally on android by running multiple sessions locally (including those with upload) and then starting an exclusive one  

```
[748.625]    perfetto_cmd.cc:1145 Service error: Aborted due to user requested higher-priority (10) exclusive session.
...
[607.597]    perfetto_cmd.cc:1145 Service error: Cannot start non-exclusive session while an exclusive session with priority 10 is active.
```

[All Logs](https://drive.google.com/drive/folders/15Jdi6APKgBJNnWpz-Nh1CthlIjMUViS3?usp=drive_link)

### DIFF
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -783,7 +783,7 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
     if (consumer->uid_ != AID_ROOT && consumer->uid_ != AID_SHELL) {
       return PERFETTO_SVC_ERR(
-          "Exclusive mode can only be requested by root or shell.");
+          "On android, exclusive mode can only be requested by root or shell.");
     }
 #endif
     // Abort all existing sessions.
--- a/src/tracing/service/tracing_service_impl_unittest.cc
+++ b/src/tracing/service/tracing_service_impl_unittest.cc
@@ -6638,21 +6638,21 @@ TEST_F(TracingServiceImplTest, ExclusiveSessionInvalidUser) {
   std::unique_ptr<MockConsumer> unprivileged_consumer = CreateMockConsumer();
   unprivileged_consumer->Connect(svc.get(), 1234 /* uid */);
   unprivileged_consumer->EnableTracing(exclusive_cfg);
-  unprivileged_consumer->WaitForTracingDisabledWithError(
-      StrEq("Exclusive mode can only be requested by root or shell."));
+  unprivileged_consumer->WaitForTracingDisabledWithError(StrEq(
+      "On android, exclusive mode can only be requested by root or shell."));
 
   // A privileged user (root) can start an exclusive session.
   std::unique_ptr<MockConsumer> root_consumer = CreateMockConsumer();
   root_consumer->Connect(svc.get(), AID_ROOT);
   root_consumer->EnableTracing(exclusive_cfg);
-  root_consumer->DisableTracing();
+  root_consumer->FreeBuffers();
   root_consumer->WaitForTracingDisabled();
 
   // A privileged user (shell) can start an exclusive session.
   std::unique_ptr<MockConsumer> shell_consumer = CreateMockConsumer();
   shell_consumer->Connect(svc.get(), AID_SHELL);
   shell_consumer->EnableTracing(exclusive_cfg);
-  shell_consumer->DisableTracing();
+  shell_consumer->FreeBuffers();
   shell_consumer->WaitForTracingDisabled();
 }
 #endif  // OS_ANDROID
@@ -6686,7 +6686,7 @@ TEST_F(TracingServiceImplTest, ExclusiveSessionBlocksNewSessions) {
       StrEq("An exclusive session with priority 1 >= requested priority "
             "1 is already active."));
 
-  exclusive_consumer->DisableTracing();
+  exclusive_consumer->FreeBuffers();
   exclusive_consumer->WaitForTracingDisabled();
 }
 
@@ -6736,7 +6736,7 @@ TEST_F(TracingServiceImplTest, ExclusiveSessionAbortRunningSessions) {
   // The aborted clone session should not have any data.
   EXPECT_THAT(clone_consumer->ReadBuffers(), IsEmpty());
 
-  exclusive_consumer->DisableTracing();
+  exclusive_consumer->FreeBuffers();
   exclusive_consumer->WaitForTracingDisabled();
 }
 
@@ -6764,7 +6764,7 @@ TEST_F(TracingServiceImplTest, ExclusiveSessionHigherPriorityAbortsLower) {
   new_exclusive_consumer->EnableTracing(exclusive_cfg_2);
 
   // Finally, disable the last exclusive session.
-  new_exclusive_consumer->DisableTracing();
+  new_exclusive_consumer->FreeBuffers();
   new_exclusive_consumer->WaitForTracingDisabled();
 }
 


protos/perfetto/config/perfetto_config.proto
```diff
@@ -4776,6 +4776,17 @@ message TraceConfig {
...
+
+  // When set to a value > 0, this tracing session will be started in
+  // "exclusive mode". This has the following semantics:
+  // - It can only be set by shell or root users.
+  // - A new exclusive session will only be started if its priority is strictly
+  //   higher than any other active exclusive session.
+  // - If a new exclusive session is started, all other existing tracing
+  //   sessions (exclusive or not) are aborted.
+  //
+  // Introduced in 25Q3.
```
LalitMaganti: Please also write the Perfetto version.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -761,6 +761,50 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
...
+    const std::string abort_reason =
+        "Aborted due to user requested higher-priority (" +
+        std::to_string(cfg.exclusive_mode_priority()) + ") exclusive session.";
+    for (auto it = tracing_sessions_.begin(); it != tracing_sessions_.end();) {
+      auto next_it = it;
+      ++next_it;
+      const auto session_consumer = it->second.consumer_maybe_null;
+      // FreeBuffers() will complete the teardown of the TracingSession and also
+      // remove it from tracing_sessions_.
+      FreeBuffers(it->first, abort_reason);
```
LalitMaganti, line 799: Q: do we not need to do disable tracing etc? Is just freebuffers sufficient? That's pretty cool!

sashwinbalaji: FreeBuffers should be able to cover the disable tracing part (http://ac/external/perfetto/src/tracing/service/tracing_service_impl.cc?l=2793)

protos/perfetto/config/perfetto_config.proto
```diff
@@ -4776,6 +4776,17 @@ message TraceConfig {
...
   optional PriorityBoostConfig priority_boost = 40;
+
+  // When set to a value > 0, this tracing session will be started in
+  // "exclusive mode". This has the following semantics:
+  // - It can only be set by shell or root users.
+  // - A new exclusive session will only be started if its priority is strictly
+  //   higher than any other active exclusive session.
+  // - If a new exclusive session is started, all other existing tracing
+  //   sessions (exclusive or not) are aborted.
+  //
```
primiano, line 4795: - Any new  non-exclusive session (or exclusive with a lower score) will be rejected as long as the current tracing session is ongoing

protos/perfetto/config/perfetto_config.proto
```diff
@@ -4776,6 +4776,17 @@ message TraceConfig {
...
+  // Introduced in 25Q3.
+  optional uint32 exclusive_mode_priority = 41;
```
primiano: imho "_mode" here doesn't add much and just makes it harder to type. Just exclusive_prio ?

src/tracing/service/tracing_service_impl.cc
```diff
@@ -761,6 +761,50 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
...
   }
 
+  uint32_t current_exclusive_mode_priority = 0;
+  for (const auto& [_, session] : tracing_sessions_) {
+    current_exclusive_mode_priority =
+        std::max(current_exclusive_mode_priority,
+                 session.config.exclusive_mode_priority());
+  }
+  if (cfg.exclusive_mode_priority() == 0 &&
+      current_exclusive_mode_priority > 0) {
```
primiano: shouldn't this be

if (cfg.exclusive_mode_priority() < current_exclusive_mode_priority)?

it will capture also the case of trying to start a session of prio=5 when there is a prio=10 running already

src/tracing/service/tracing_service_impl.cc
```diff
@@ -761,6 +761,50 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
...
+        consumer->uid_ != 2000 /* AID_SHELL */) {
+      return PERFETTO_SVC_ERR(
+          "Exclusive mode can only be requested by root or shell.");
+    }
+    if (current_exclusive_mode_priority >= cfg.exclusive_mode_priority()) {
+      return PERFETTO_SVC_ERR(
+          "An exclusive session with priority %u >= requested priority %u is "
+          "already active.",
+          current_exclusive_mode_priority, cfg.exclusive_mode_priority());
+    }
```
primiano: this can be folded into the initial check, reducing lines of code and thinking

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2782,15 +2828,16 @@ bool TracingServiceImpl::WriteIntoFile(TracingSession* tracing_session,
...
+                                     const std::string& error) {
   PERFETTO_DCHECK_THREAD(thread_checker_);
   PERFETTO_DLOG("Freeing buffers for session %" PRIu64, tsid);
   TracingSession* tracing_session = GetTracingSession(tsid);
   if (!tracing_session) {
     PERFETTO_DLOG("FreeBuffers() failed, invalid session ID %" PRIu64, tsid);
     return;  // TODO(primiano): signal failure?
   }
-  DisableTracing(tsid, /*disable_immediately=*/true);
+  DisableTracing(tsid, /*disable_immediately=*/true, /*error=*/error);
```
primiano: small nit: the /*param_name=*/ is typically useful only in case where the argument is ambiguous (e.g. in the previous case when we pass "true"). In this case, it doesn't add anything that the variable name (error) doesn't say already

protos/perfetto/config/perfetto_config.proto
sashwinbalaji: done

src/tracing/service/tracing_service_impl.cc
sashwinbalaji: agreed.

src/tracing/service/tracing_service_impl.cc
sashwinbalaji: yeah was handling that scenario separately in line 783. fixed the check to cover both the case here itself.

src/tracing/service/tracing_service_impl.cc
sashwinbalaji: yes, updated.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -761,6 +761,50 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
...
+  }
+  if (cfg.exclusive_mode_priority() == 0 &&
+      current_exclusive_mode_priority > 0) {
+    return PERFETTO_SVC_ERR(
+        "Cannot start non-exclusive session while an exclusive session with "
+        "priority %u is active.",
+        current_exclusive_mode_priority);
+  }
+  if (cfg.exclusive_mode_priority() > 0) {  // Exclusive mode.
+    if (consumer->uid_ != 0 /* AID_ROOT */ &&
```
rsavitski: 1) shell UID constant (and more generally, the whole root/shell concept) is android-specific. I would only do the UID checking on android (see other cases where we check UID). We can allow anyone to set this on other platforms (at least for testability).
2) there's a dedicated AID_SHELL constant, though this file doesn't use it consistently, please make all places use the constant unless there are reasons not to.

protos/perfetto/config/perfetto_config.proto
sashwinbalaji: done.

protos/perfetto/config/perfetto_config.proto
sashwinbalaji: done.

src/tracing/service/tracing_service_impl.cc
sashwinbalaji: updated the file and to do so also added AID_ROOT and AID_STATSD to sys_types

src/tracing/service/tracing_service_impl.cc
```diff
@@ -761,6 +763,48 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
...
+    return PERFETTO_SVC_ERR(
+        "An exclusive session with priority %u >= requested priority %u is "
+        "already active.",
+        current_exclusive_prio, cfg.exclusive_prio());
+  }
+
+  if (cfg.exclusive_prio() > 0) {  // Exclusive mode.
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+    if (consumer->uid_ != AID_ROOT && consumer->uid_ != AID_SHELL) {
+      return PERFETTO_SVC_ERR(
```
LalitMaganti, line 785: I add a specifier of "on Android".

LalitMaganti (approved):

PR #2406, opened 2025-08-04
ui: improve rendering for out-of-viewport timestamp hovers

altimin:
Currently the UI draws a vertical line when a user hovers over a timestamp in the details panel, making it easier to understand where the event with relation to the current viewport and the events on the screen.

However, if the hovered timestamp is outside of the viewport, then currently the UI doesn't provide any information on hover (including whether the hovered timestamp is to the left or to the right of the viewport).

This patch fixes that by rendering a label (on either the left or the right side of the notes panel) telling where the timestamp is w.r.t the current viewport and how far.

Preview: Screenshot 2025-08-04 at 16 55 57


stevegolton (approved):
Nice idea! LGTM

PR #2407, opened 2025-08-04
ui: improve markers for selected events.

altimin:
- Draw a marker (small triangle) pointing down for instant events. We currently draw a range for non-instant selected events, draw a marker for selected instant events as well to make it easier to spot them.
- Draw the label in the middle for the intersection of (visible range, event span) for non-instant selections: currently it's always rendered in the middle of the event span, even when it's outside of the viewport.

Instant event marker (after this patch):
Screenshot 2025-08-04 at 17 08 37

Non-instant selected event label when zoomed in (before):
Screenshot 2025-08-04 at 17 10 17

Non-instant selected event label when zoomed in (after this patch):
Screenshot 2025-08-04 at 17 10 42

### DIFF
--- a/ui/src/frontend/viewer_page/time_selection_panel.ts
+++ b/ui/src/frontend/viewer_page/time_selection_panel.ts
@@ -69,8 +69,11 @@ function drawHBar(
   const labelWidth = ctx.measureText(label).width;
 
   // Find a good position for the label:
-  // By default put the label in the middle of the H:
-  let labelXLeft = Math.floor(xWidth / 2 - labelWidth / 2 + xLeft);
+  // By default put the label in the middle of the visible portion of the H.
+  const visibleLeft = Math.max(xLeft, bounds.x);
+  const visibleRight = Math.min(xRight, bounds.x + bounds.width);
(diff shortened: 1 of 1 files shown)


altimin:
@stevegolton — I've been wondering if it would be worth making it more explicit that the current selection continues beyond the viewport (e.g below)? I'm leaning towards "no", as the vertical bars show the boundaries of the event when it's in the viewport, but would be interested in your perspective.

Screenshot 2025-08-04 at 17 11 43

stevegolton:
> @stevegolton — I've been wondering if it would be worth making it more explicit that the current selection continues beyond the viewport (e.g below)? I'm leaning towards "no", as the vertical bars show the boundaries of the event when it's in the viewport, but would be interested in your perspective.
> 
> Screenshot 2025-08-04 at 17 11 43

I wouldn't be against your idea of showing arrows when the slice endpoints are out of bounds. If you wanted to take a stab at it I'd be happy to review.

sashwinbalaji (approved):

rsavitski (approved):

PR #2408, opened 2025-08-04
windows: restore include order for windows.h

rsavitski:
windows.h has to be included before headers like synchapi.h. Restore the ordering before the commit below, introducing extra newlines as separators to prevent autoformatters from reordering it back.

Fixes: 5f14006522 fix mingw builds (#2397)


rsavitski:
Context: chromium roll fails due to windows bots (https://chromium-review.googlesource.com/c/chromium/src/+/6817533?tab=checks).

LalitMaganti:
Broken in Google3 also: cl/790767560

PR #2410, opened 2025-08-04
tp: refactor proto modules to live in a separate struct

LalitMaganti:
This CL does a major refactoring across all of the proto parser and module
to depend on a separate module context instead of the trace processor
context. This is the first step to stopping parser-local state from
using the context entirely for non-processor wide state.

For now, the struct is also stored inside the context but this will
change down the line to live in ProtoTraceReader.

While this change touches a lot of files, in practice, it's just quite
a mechanical change and there should be no behavioural change as a
result of this CL.

### DIFF
--- a/src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
+++ b/src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
@@ -25,6 +25,7 @@
 #include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
 #include "perfetto/ext/base/status_macros.h"
+#include "perfetto/ext/base/status_or.h"
 #include "perfetto/protozero/field.h"
 #include "perfetto/protozero/proto_decoder.h"
 #include "perfetto/protozero/proto_utils.h"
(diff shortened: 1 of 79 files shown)


src/trace_processor/importers/proto/proto_importer_module.h
```diff
@@ -141,11 +148,19 @@ class ProtoImporterModule {
...
-  // Primarily intended for special modules that need to get all TracePacket's,
-  // for example for trace proto content analysis. Most modules need to register
-  // for specific fields using the method above.
-  void RegisterForAllFields(TraceProcessorContext*);
+  void RegisterForField(uint32_t field_id);
+
+  ProtoImporterModuleContext* module_context_;
+};
+
+struct ProtoImporterModuleContext {
```
sashwinbalaji, line 159: nit: This is the crux of the entire change. Can we add some comment explaining the purpose of struct.

src/trace_processor/trace_processor_impl.cc
```diff
@@ -457,6 +456,12 @@ std::pair<int64_t, int64_t> GetTraceTimestampBoundsNs(
 
 TraceProcessorImpl::TraceProcessorImpl(const Config& cfg)
     : TraceProcessorStorageImpl(cfg), config_(cfg) {
+  context_.register_additional_proto_modules = &RegisterAdditionalModules;
+  if (context_.register_additional_proto_modules) {
```
sashwinbalaji, line 460: curious, why is this checked needed ?

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -268,15 +270,17 @@ void FtraceTokenizer::TokenizeFtraceEvent(
           event_id == protos::pbzero::FtraceEvent::kGpuWorkPeriodFieldNumber)) {
     TokenizeFtraceGpuWorkPeriod(cpu, std::move(event), std::move(state));
     return;
-  } else if (PERFETTO_UNLIKELY(event_id ==
```
sashwinbalaji, line 271: why this change ? I mean no harm in it but doesn't seem relevant to the refactoring

src/trace_processor/trace_processor_impl.cc
LalitMaganti: Will be fixed in next CL. This sort of check is common there so I've done it now as well.

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
LalitMaganti: Small stylistic change since I was in same file. Is just a noop.

src/trace_processor/importers/proto/proto_importer_module.h
LalitMaganti: Done

PR #2411, opened 2025-08-04
tp: decouple sorting from TraceProcessorContext

LalitMaganti:
Allows parsing of multiple traces simultaneously without getting clashes between them from using the same parsers/tokenizers etc.


src/trace_processor/types/trace_processor_context.h
```diff
@@ -75,8 +61,12 @@ class TraceFileTracker;
 class TraceReaderRegistry;
 class TraceSorter;
 class TraceStorage;
+class TrackCompressor;
+class TrackCompressor;
```
sashwinbalaji: duplicate

src/trace_processor/types/trace_processor_context.h
```diff
@@ -75,8 +61,12 @@ class TraceFileTracker;
...
 class TrackEventModule;
 class TrackTracker;
+struct ProtoImporterModuleContext;
+struct ProtoImporterModuleContext;
```
sashwinbalaji: duplicate

src/trace_processor/importers/proto/proto_trace_parser_impl.h
```diff
@@ -35,32 +33,25 @@ class TracePacket_Decoder;
 namespace trace_processor {
 
 class PacketSequenceState;
+struct ProtoImporterModuleContext;
 class TraceProcessorContext;
 
-class ProtoTraceParserImpl : public ProtoTraceParser {
+class ProtoTraceParserImpl {
```
sashwinbalaji, line 39: should we just change it everywhere to ProtoTraceParser (maybe a separe cl) ?

src/trace_processor/importers/proto/proto_importer_module.cc
```diff
@@ -59,4 +59,60 @@ void ProtoImporterModule::RegisterForField(uint32_t field_id) {
   module_context_->modules_by_field[field_id].push_back(this);
 }
 
+void ProtoImporterModuleContext::PushFtraceEvent(uint32_t cpu,
```
sashwinbalaji, line 82: should we have a private templated helper method to reduce the duplication ? Maybe not, but depends on if any other scenario will come with per cpu stream

src/trace_processor/sorter/trace_sorter.h
```diff
@@ -380,17 +165,14 @@ class TraceSorter {
     uint64_t chunk_index : BumpAllocator::kChunkIndexAllocIdBits;
     uint64_t chunk_offset : BumpAllocator::kChunkOffsetAllocIdBits;
 
-    // The type of this event. GCC7 does not like bit-field enums (see
-    // https://stackoverflow.com/questions/36005063/gcc-suppress-warning-too-small-to-hold-all-values-of)
-    // so use an uint64_t instead and cast to the enum type.
-    uint64_t event_type : kMaxTypeBits;
+    // Indicates whether the event is a JSON event: may have special rules based
+    // on the event type and duration.
+    uint64_t is_json_event : 1;
```
sashwinbalaji: why uint64_t instead of bool ?

src/trace_processor/importers/proto/proto_trace_parser_impl.h
LalitMaganti: YEs separate CL.

PR #2412, opened 2025-08-05
tp: refactor tests to decouple them from utids

LalitMaganti:
Tests were brittle becuase of use of utid whch is not guaranteed to be
stable, switch them to use tids/timestamps which are stable.

CHROME_STDLIB_MANUAL_ROLL=trace processor maintainer change


LalitMaganti:
@victorvianna as the Chrometto sheriff FYI.

test/trace_processor/diff_tests/stdlib/sched/tests.py
```diff
@@ -168,24 +168,35 @@ def test_sched_previous_runnable_on_thread(self):
         query="""
         INCLUDE PERFETTO MODULE sched.runnable;
 
-        SELECT *
-        FROM sched_previous_runnable_on_thread
-        WHERE prev_wakeup_runnable_id IS NOT NULL
-        ORDER BY id DESC
+        SELECT
+          running.ts AS running_ts,
```
victorvianna, line 172: Why does the set of selected columns change so much in this test and the one below? In the ones above the change was basically utid -> tid

LalitMaganti: Because instead of them being utids, they are ids into sched table so we then need to join to that table. We also emit timestamp because that's what most of the folks in this table actually care about: whether the timestamp matches their expectation.

PR #2420, opened 2025-08-05
Add render_pass_instance_id and use it to color related gpu events with the same color

cwabbott0:
This is based on #1818 but adds a new field to the protocol instead to avoid overloading render_pass_handle. It doesn't handle adding flows yet, but I would want to do that in the future.


src/trace_processor/importers/proto/gpu_event_parser.cc
```diff
@@ -490,6 +491,15 @@ void GpuEventParser::ParseGpuRenderStageEvent(
               }
             }
           }
+
+          if (event.render_pass_instance_id()) {
+            base::StackString<512> id_str("rp:#%" PRIu64,
+                                          event.render_pass_instance_id());
+            inserter->AddArg(context_->storage->InternString("correlation_id"),
```
LalitMaganti: Please intern this in the constructor to prevent repeated internstring on a constant.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/trace_processor_slice_track.ts
```diff
@@ -88,7 +88,9 @@ export async function createTraceProcessorSliceTrack({
       : () => new ThreadSliceDetailsPanel(trace),
     colorizer: (row) => {
       if (row.correlation_id) {
-        return getColorForSlice(row.correlation_id);
+        return getColorForSlice(row.correlation_id, {
+          stripTrailingDigits: false,
```
LalitMaganti, line 92: This needs to be configurable. We really need this to be true for Android slices which also use this codepath.

LalitMaganti: Ahhhh I take it back this is on the correlation_id codepath. makes sense this is false here actually.

LalitMaganti:
Hmmm looks like presubmit failure as you need to run tools/gen_all. Can you please fix?

LalitMaganti (approved):

Issue #2424, opened 2025-08-05
diskstats misses disk


sifangli-labs:
I am getting diskstats with the following cfg - however, I can only see 1 of the two disks (sda) reported in the pftrace.
 
```
name: "linux.sys_stats"
    target_buffer: 0
    sys_stats_config {
      diskstat_period_ms : 1000
```

Here is the summary of the diskstats content - I note the nvme0n1 is missing in the pftrace's diskstats - sda is showing though.
```
$ cat /proc/diskstats 
   7       0 loop0 ...
   7       1 loop1 ...
...
 259       0 nvme0n1 ... 
 259       1 nvme0n1p1 ... 
   8       0 sda ...
```

LalitMaganti:
Please link the trace file. it's very difficult to debug these sort of issues without the trace.

LalitMaganti:
Actually I figured it out, the data is being dropped in trace processor: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/importers/proto/system_probes_parser.cc;l=133;bpv=1;bpt=0

This restriction comes all the way from the initial contribution: https://r.android.com/2249376

I'm not sure what the motivation for this was. Potentially that on Android, we only care about the primary disk? We can probably drop it as long as the semantics of the fields of disk stats are the same between nvme* and sd* disks

sifangli-labs:
Actually the nvme was the primary disk, this is on Linux and we do care about both disk's stats

LalitMaganti:
I'm happy to drop the restriction but the main problem we have right now is that we have a hardcoded value of 2048 sectors per MB (implying sector size of 512). But afaik, nvme devices often use 4K sectors.

Can I ask you to do `/sys/block/<device_name>/queue/logical_block_size` for both disks and see what the values are?

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/importers/proto/system_probes_parser.cc;l=243;bpv=0;bpt=0

sifangli-labs:
both says 512 on in our env

LalitMaganti:
Sent https://github.com/google/perfetto/pull/2425 to fix this. Filed https://github.com/google/perfetto/issues/2427 as a follow-up to track 4k logical sector sizes.

PR #2425, opened 2025-08-05
tp: remove restriction of only showing sda

LalitMaganti:
On Linux, we could well want other disks (NVME for example). restricting
at TP was the wrong move, we should really do something at the UI or on
device levels.

Fixes: https://github.com/google/perfetto/issues/2424


src/trace_processor/importers/proto/system_probes_parser.cc
```diff
@@ -240,12 +236,17 @@ SystemProbesParser::SystemProbesParser(TraceProcessorContext* context)
...
+  // that the disk's logical sector size is 512. This is very commonly true but
+  // enterprise SSDs can have a logical size of 4K.
+  //
+  // Unfortunately the only way we could figure this out is by pollling
+  // `/sys/block/<device_name>/queue/logical_block_size` on device which
+  // requires changing the recording code.
   static constexpr double SECTORS_PER_MB = 2048.0;
   static constexpr double MS_PER_SEC = 1000.0;
   std::string device_name = ds.device_name().ToStdString();
-  if (!IsSupportedDiskStatDevice(device_name)) {
```
sashwinbalaji, line 246: I don't think we can just remove this check. 

Reason being below we maintain perv_xx(http://ac/external/perfetto/src/trace_processor/importers/proto/system_probes_parser.h?l=67-74) with the assumption there is only one disk. Now if multiple disks are allowed we'll mix their values together. (http://ac/external/perfetto/src/trace_processor/importers/proto/system_probes_parser.cc?l=282-312)

LalitMaganti: Nice catch! I've fixed this :)

PR #2429, opened 2025-08-05
tp: add irq index to the end of the num_irq/num_softirq counters

LalitMaganti:


sashwinbalaji (approved):
nit: In the change as well as in other parts of the file we usually create StackString of len 1024 which seems little too large.

LalitMaganti:
> nit: In the change as well as in other parts of the file we usually create StackString of len 1024 which seems little too large.

Yeah agreed but afaik it does not cause any harm and if we want to tackle it should be tackled in a a different CL.

PR #2430, opened 2025-08-05
[SDK] Add NamedTrack::FromPointer

etiennep-chromium:
Similar to Track::FromPointer, a shorthand for 
NamedTrack("name", reinterpret_cast<uintptr_t>(this))
As well as a version of NamedTrack::ThreadScoped that takes a pointer, similar
to Track::ThreadScoped


ddiproietto (approved):
Thanks! If you could add some basic compile tests, so that we don't end up like https://github.com/google/perfetto/pull/1879

etiennep-chromium:
Added tests!

PR #2441, opened 2025-08-06
github: experimental: gemini-cli GH action

LalitMaganti:
Just an trial experiment to see if it's useful. Will disable if not.

For now, is using my personal Gemini API key, we should move it to a proper key if we decide to keep this moving forward.


rsavitski (approved):
LGTM, but can we change the first line of the commit to something like "github: experimental: gemini-cli GH action". Imo it's more descriprive when eyeballing --oneline history.

Issue #2400, opened 2025-08-04
[Feature request][UI] Auto-expand and Pin tracks via URL parameters


pengj:
**Challenge:** 
When a trace is loaded in Perfetto, users often encounter difficulties efficiently navigating the UI, particularly when the trace includes a large number of tracks. Manually locating, expanding, and pinning the most pertinent tracks—such as those for the analyzed application or tracks that are frequently compared—can be tedious and time-consuming.

<img width="755" height="400" alt="Image" src="https://github.com/user-attachments/assets/802c8a7c-b2dd-4587-b0a6-1503c204663e" />

**Proposal:** 
I think a Perfetto UI plugin which parses the URL parameters could help here. The tracked application and pinned tracks could be passed through the URL when opening the trace file. 

Automatically expand or pin tracks that are passed by the URL parameters when a new trace is loaded. This feature would eliminate the need for users to manually search for and expand these key tracks, resulting in a more streamlined workflow.

Are there any concerns with any of the above changes? Let me know if there are other options to handle this. I would be willing to contribute PRs to accomplish the above. 

Thanks!

LalitMaganti:
@stevegolton for thoughts.

Not sure if URL parameters or postMessage is the right answer, not sure what we settled on in the end as the "de-facto" way things are passed to plugins.

There is already a plugin which does saving/restoring of pinned tracks as a command, the easiest thing would be to extend this to also have something which looks at input on url params/post message: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts;l=1

pengj:
@LalitMaganti Is there another way to pass the data to plugins? I can also check that direction. 

Do we have a plugin to automatically expand the track based on URL parameters, or do you think we should also extend the restorePinnedTracks plugin?

There are a few use cases could be benefits with this: the hyperlinks can be created that automatically expand the tracked application traces and pin the performance metrics tracks we common used to analyze the traces.

LalitMaganti:
https://github.com/google/perfetto/blob/000f15c69118661a6a3a5d4ab21dbf900f318b2b/ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/index.ts#L27: this is what one plugin does today. This is not really a fully approved approach and we should do the work to expose it over the plugin API vs directly using the location hash. 

[...]

PR #2443, opened 2025-08-06
ui: Add expand and pin function for the RestorePinnedTrack plugin

pengj:
This PR added the auto-expand and pin function mentioned in this [feature request](https://github.com/google/perfetto/issues/2400) for the RestorePinnedTrack plugin. 

The expand and pin tracked will passed by URL paramters with format: 
```
dev.perfetto.RestorePinnedTrack:expand_tracks=(track-name-1, track-name-2)
dev.perfetto.RestorePinnedTrack:pinned_tracks=(track-name-1, track-name-2)
```

Test with the query: 
http://localhost:10000/#!/viewer?local_cache_key=2fb82c7d-bfaf-6dbe-6e49-6f25e9bdf84a&dev.perfetto.RestorePinnedTrack:expand_tracks=(com.android.vending,com.instagram.android)&dev.perfetto.RestorePinnedTrack:pinned_tracks=(FinskyEvent)

The expand track and pinned track are set correctly. 

Screenshot 2025-08-06 at 18 43 14


ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
```diff
@@ -24,6 +24,22 @@ const SAVED_TRACKS_KEY = `${PLUGIN_ID}#savedPerfettoTracks`;
 
 const RESTORE_COMMAND_ID = `${PLUGIN_ID}#restore`;
 
+const URL_PARAM_EXPAND_TRACKS = 'expand_tracks';
+const URL_PARAM_PINNED_TRACKS = 'pinned_tracks';
+
+
+function getParamValues(queryString: string, paramName: string) {
```
LalitMaganti: As I said on the bug, this code needs to be moved to the core, we should not be propogating more instances of this throughout the codebase. 

We should:
a) make it a part of the schema at https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/public/route_schema.ts;l=1
b) pass it to the pluginmanager at https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/third_party/perfetto/ui/src/frontend/index.ts;l=271?q=enablePlugins&ss=chromium%2Fchromium%2Fsrc
c) plugin manager should parse out args per-plugin and then pass it in onActivate as the second argument at https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/core/plugin_manager.ts;l=110?q=pluginmanagerimpl&ss=chromium%2Fchromium%2Fsrc.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
```diff
@@ -24,6 +24,22 @@ const SAVED_TRACKS_KEY = `${PLUGIN_ID}#savedPerfettoTracks`;
...
+  const regex = new RegExp(`${PLUGIN_ID}:${paramName}=\\(([^)]*)\\)`);
+  const match = queryString.match(regex);
+  if (match && match[1]) {
+    return match[1].split(',').map(s => s.trim()).filter(Boolean);
+  }
+  return [];
+}
+
+let expand_tracks: string[];
```
LalitMaganti: Please use camelCase in keeping with the rest of the UI and also this should be moved to be a static field of the plugin itself.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
```diff
@@ -62,6 +78,8 @@ export default class implements PerfettoPlugin {
       addOrReplaceNamedPinnedTracks(parsed.data);
     });
     document.body.appendChild(input);
+    expand_tracks = getParamValues(location.hash, URL_PARAM_EXPAND_TRACKS);
```
LalitMaganti: They way you've coded this means that even if you use the UI for the new instance of the trace, the same set of tracks will be pinned and the same set of tracks will be expanded. I'm not sure if that's really what we want or this should only happen on the first trace which is loaded.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
```diff
@@ -24,6 +24,22 @@ const SAVED_TRACKS_KEY = `${PLUGIN_ID}#savedPerfettoTracks`;
 
 const RESTORE_COMMAND_ID = `${PLUGIN_ID}#restore`;
 
+const URL_PARAM_EXPAND_TRACKS = 'expand_tracks';
```
LalitMaganti: The id of this plugin needs to change from `dev.perfetto.RestorePinnedTrack` to `dev.perfetto.AutoPinAndExpandTracks` to account for the new scope.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
```diff
@@ -161,6 +179,19 @@ export default class implements PerfettoPlugin {
         assertIsInstance<HTMLInputElement>(files, HTMLInputElement).click();
       },
     });
+
+    // Process URL parameters for auto-expanding groups and pinning tracks
+    this.processUrlParameters();
+  }
+
+  private processUrlParameters(): void {
+    if (expand_tracks.length > 0) {
```
LalitMaganti: Very much not a fan of looking at just a prefix. It's inconsistent with the whole rest of how this class works.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
```diff
@@ -24,6 +24,22 @@ const SAVED_TRACKS_KEY = `${PLUGIN_ID}#savedPerfettoTracks`;
 
 const RESTORE_COMMAND_ID = `${PLUGIN_ID}#restore`;
 
+const URL_PARAM_EXPAND_TRACKS = 'expand_tracks';
+const URL_PARAM_PINNED_TRACKS = 'pinned_tracks';
```
LalitMaganti: These URL parameters here are far too vague. They should be labelled as expand_tracks_with_name_on_startup and pin_tracks_with_name_on_startup to make it clear what they are doing.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
pengj: are you suggest to move this call to `onTraceLoad`?

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
pengj: do you think another PR will be better to handle all these actions?

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
pengj: do I need to change the folder name if I change the ID? if this the case, I feel this is a new plugin now

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
pengj: what would your suggestion here?

LalitMaganti: Well exact string matching is how the rest of the class works so it should work similarly :)

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
LalitMaganti: Yes you should. And yes it essentially would be but that's ok, no one should be depending on this plugin name.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
LalitMaganti: I don't think you can given how the plugin will be passed the arguments by the core UI.

If you wanted it to work the other way, on first onTraceLoad, these parameters should be set to undefined so future trace loads are not affected.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
LalitMaganti: Up to you, don't mind either way. But it will have to happen before this change can land.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
pengj: Got it, once it is expand/pinned, I can reset these values, so it will do it again after trace loaded.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
pengj: you mean use `findMatchingTrack` to find the track, rather than `startsWith`? use length check to track the expand and pin track checking is fine here, right?

LalitMaganti: You cannot use that function because it uses a bunch of other parameters. But the point I was making was it does an exact match of the track name. So if anything, we should do that here as well.

pengj: The reason I am using start with because for App tracking, the main track is not exactly the app package name, it also contains the threads/process running.  Taken the screenshot I put above as example, the track name for app com.android.vending is `com.android.vending 16132`. what's your suggestion to handle this?

LalitMaganti: @stevegolton should we be separating out the tid/pid so that we can do equality matching on track names? 

Alternatively I would suggest making this a regex or glob. Whatever we do, starts with is not the right operator.

ui/src/plugins/dev.perfetto.RestorePinnedTracks/index.ts
pengj: @LalitMaganti  I am not familiar with the route schema and PluginManager, so please bear with me with a few more questions.
a) Do you mean adding something similar to this? (I’m not confident about this structure.)
```
// A comma-separated list of parameters/values want to pass in the plugin, example: pluginParams=plugin-id-1:prameters, plugin-id-2:paramss
pluginParams: z.string().optional().catch(undefined)
```
b) then get the `pluginParams` from the `route.args.pluginParams`, passed to `pluginManager.activatePlugins(overrides, pluginParams)`

c) within the `activatePlugins` function, for each plugin, I will check if the plugin params has the plugin id, then passed the parameters in when call `p.desc.onActivate?.(app, parmas);`?  
In the plugin, do I need to update `static onActivate()`? how to get the passed params?

LalitMaganti: Yes to all your questions.

pengj: @LalitMaganti  for the route parse, in the current implementation, we have
```
const route = Router.parseUrl(window.location.href);
const overrides = (route.args.enablePlugins ?? '').split(',');
```
after I added the pluginParams there, I can't find the passed plugin params as it is not in the location.href, rather than location.hash, any suggesiton here?

LalitMaganti: ~~In that case, for consistency we should use href and we will likely need to break the ad-hoc usage of location the SysUI are doing. I will take care of informing them of this change. You will need to format your URL to place it in the href instead of the hash.~~

See next comment.

LalitMaganti: Actually hold on: the router parses only the hash. So you should not have any problems here? https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/core/router.ts;l=205?q=%20Router.parseUrl&ss=chromium%2Fchromium%2Fsrc

The hash is a subset of href so passing href to the router is correct and WAI.

pengj: > You will need to format your URL to place it in the href instead of the hash.

n00b here, could you give more explaination here? 
I am current use this one to test: 

http://localhost:10000/viewer?local_cache_key=2fb82c7d-bfaf-6dbe-6e49-6f25e9bdf84a&pluginParams=dev.perfetto.AutoPinAndExpandTracks:expand_tracks=(com.android.vending), dev.perfetto.AutoPinAndExpandTracks:pinned_tracks=(FinskyEvent)

what should I change to?

LalitMaganti:
```
http://localhost:10000/viewer?local_cache_key=2fb82c7d-bfaf-6dbe-6e49-6f25e9bdf84a#pluginParams=dev.perfetto.AutoPinAndExpandTracks:expand_tracks=(com.android.vending), dev.perfetto.AutoPinAndExpandTracks:pinned_tracks=(FinskyEvent)
```

pengj: ~I have tried this way, it shows the following error:
{"errno":-2,"code":"ENOENT","syscall":"open","path":"perfetto/out/ui/ui/dist/viewer"}~

@LalitMaganti  I think I found the issue, the herf works here, it is how I passed the argument here

pengj: @LalitMaganti  
the windows.localtion.hash is not a full url here, once I passed the hash in, Router.parseUrl will fail

LalitMaganti: So I just tested out enablePlugins which works a very similar way and:
```
http://localhost:10001/#!/plugins?enablePlugins=dev.perfetto.WidgetsPage
```

causes `dev.perfetto.WidgetsPage` to be correctly enabled.

So In that case, the url should be as follows:

```
http://localhost:10000/#!/viewer?local_cache_key=2fb82c7d-bfaf-6dbe-6e49-6f25e9bdf84a&pluginParams=dev.perfetto.AutoPinAndExpandTracks:expand_tracks=(com.android.vending), dev.perfetto.AutoPinAndExpandTracks:pinned_tracks=(FinskyEvent)
```

ui/src/core/default_plugins.ts
```diff
@@ -59,8 +60,8 @@ export const defaultPlugins = [
   'dev.perfetto.QueryLog',
   'dev.perfetto.QueryPage',
   'dev.perfetto.RecordTraceV2',
-  'dev.perfetto.RestorePinnedTrack',
   'dev.perfetto.Sched',
+  'dev.perfetto.SchedSummary',
```
LalitMaganti, line 64: Please revert.

ui/src/core/plugin_manager.ts
```diff
@@ -106,18 +106,28 @@ export class PluginManagerImpl {
...
-  activatePlugins(enableOverrides: ReadonlyArray<string> = []) {
+  activatePlugins(
+    enableOverrides: ReadonlyArray<string> = [],
+    pluginParams: ReadonlyArray<string> = [],
+  ) {
     const enabledPlugins = this.registry
       .valuesAsArray()
       .filter((p) => p.enableFlag.get() || enableOverrides.includes(p.desc.id));
 
+    console.log(
```
LalitMaganti, line 119: Please remove

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -13,29 +13,47 @@
...
-const PLUGIN_ID = 'dev.perfetto.RestorePinnedTrack';
+const PLUGIN_ID = 'dev.perfetto.AutoPinAndExpandTracks';
 const SAVED_TRACKS_KEY = `${PLUGIN_ID}#savedPerfettoTracks`;
 
 const RESTORE_COMMAND_ID = `${PLUGIN_ID}#restore`;
 
+const URL_PARAM_EXPAND_TRACKS = 'expand_tracks_with_name_on_startup';
+const URL_PARAM_PINNED_TRACKS = 'pin_tracks_with_name_on_startup';
+
+function getParamValues(pluginParams: ReadonlyArray<string>, paramName: string): string[] {
```
LalitMaganti, line 31: Please remove

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -13,29 +13,47 @@
...
-export default class implements PerfettoPlugin {
+export default class AutoPinAndExpandTracks implements PerfettoPlugin {
   static readonly id = PLUGIN_ID;
   private ctx!: Trace;
+  private static expandTracks: string[] = [];
+  private static pinTracks: string[] = [];
 
-  static onActivate() {
+  static onActivate(_app: App, pluginParams: ReadonlyArray<string> = []) {
+    console.log('[RestorePinnedTracks] pluginParams:', pluginParams);
```
LalitMaganti, line 56: Remove

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -62,6 +80,10 @@ export default class implements PerfettoPlugin {
       addOrReplaceNamedPinnedTracks(parsed.data);
     });
     document.body.appendChild(input);
+    AutoPinAndExpandTracks.expandTracks = getParamValues(pluginParams, URL_PARAM_EXPAND_TRACKS);
+    AutoPinAndExpandTracks.pinTracks = getParamValues(pluginParams, URL_PARAM_PINNED_TRACKS);
+    console.log('[RestorePinnedTracks] expandTracks:', AutoPinAndExpandTracks.expandTracks);
```
LalitMaganti, line 85: Remove logs

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -161,6 +183,24 @@ export default class implements PerfettoPlugin {
         assertIsInstance<HTMLInputElement>(files, HTMLInputElement).click();
       },
     });
+
+    // Process URL parameters for auto-expanding groups and pinning tracks
+    this.processUrlParameters();
+  }
+
+  private processUrlParameters(): void {
+    if (AutoPinAndExpandTracks.expandTracks.length > 0) {
```
LalitMaganti, line 192: Please run ui/format-sources. None of this has been formatted.

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -161,6 +183,24 @@ export default class implements PerfettoPlugin {
...
+        this.ctx.workspace.flatTracks.filter((t) => AutoPinAndExpandTracks.expandTracks.some((prefix) => t.name.startsWith(prefix))).forEach((t) => t.expand());
+    }
+    
+
+    if (AutoPinAndExpandTracks.pinTracks.length > 0) {
+       this.ctx.workspace.flatTracks.filter((t) => AutoPinAndExpandTracks.pinTracks.some((prefix) => t.name.startsWith(prefix))).forEach((t) => t.pin());
+    }
+
+    //reset the tracks to process again
```
LalitMaganti, line 201: Please improve the comment explaining why.

ui/src/public/route_schema.ts
```diff
@@ -57,6 +57,9 @@ export const ROUTE_SCHEMA = z
     // A comma-separated list of plugins to enable for the current session.
     enablePlugins: z.string().optional().catch(undefined),
 
+    // A comma-separated list of parameters/values want to pass in the plugin, example: pluginParams=plugin-id-1:prameters, plugin-id-2:prameters
```
LalitMaganti, line 60: Please wrap this better.

LalitMaganti (changes requested):

pengj:
@LalitMaganti  I will close this PR and resubmit a new one. Not sure why my change is not showing up correctly and the CLA check not passed

PR #2444, opened 2025-08-06
tp: update android_special_classes with graph.scan and add heap_graph.class_relationship module

bqlbq:
Update android_special_classes with graph.scan which is much faster than the RECURSIVE statement (milliseconds vs seconds).

Also share the logic in a module so other metrics like java_heap_tags can use it.

b/427894998

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/class_relationship.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/class_relationship.sql
@@ -17,7 +17,7 @@ INCLUDE PERFETTO MODULE graphs.scan;
 
 -- Given a list of classes as ancestor classes, return all the classes that
 -- decend from them.
-CREATE PERFETTO MACRO android_heap_graph_class_ancestor(
+CREATE PERFETTO MACRO android_heap_graph_class_find_descendants(
     -- ancestor class `id`'s from the heap_graph_class table containing a
     -- single column: `id`
     ancestor_class_ids TableOrSubquery
--- a/src/trace_processor/metrics/sql/android/java_heap_histogram.sql
+++ b/src/trace_processor/metrics/sql/android/java_heap_histogram.sql
@@ -36,7 +36,7 @@ WITH
   )
 SELECT
   id as cls_id, ancestor_class_name as category
-FROM android_heap_graph_class_ancestor!(special_classes);
+FROM android_heap_graph_class_find_descendants!(special_classes);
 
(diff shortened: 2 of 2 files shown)


src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/class_relationship.sql
```diff
@@ -0,0 +1,61 @@
...
+-- distributed under the License is distributed ON an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE graphs.scan;
+
+-- Given a list of classes as ancestor classes, return all the classes that
+-- decend from them.
+CREATE PERFETTO MACRO android_heap_graph_class_ancestor(
```
LalitMaganti: Any reason for this to be public?

bqlbq: Not sure about the criteria for `public`. Happy to make it non-public (like `_heap_graph_class_ancestor`?) as long as it does not block me from using it in other trace metrics.

LalitMaganti: Depends what you mean by "other trace metrics". If you mean in Google3, this should be public. If you mean in Perfetto repo, then fine to be underscored.

I try to avoid public APIs as much as possible as they increase our API surface.

bqlbq: The plan is to use it in Google3 `java_heap_tags`.

If the API surface is a big concern then an alternative is to not introducing the module at all but instead have two copies of the same traverse logic, one in java_heap_histogram (Perfetto) and one in java_heap_tags_extraction (Google3).

LalitMaganti: It's fine to be public.

LalitMaganti: I would maybe call this `android_heap_graph_class_find_descendants`. Ancestor makes me thinking it is finding ancestors, it's actually finding recursively descendants.

bqlbq: Good point! Done.

LalitMaganti (approved):

PR #2445, opened 2025-08-06
tp/ui: initial support for kernel trackevent convention

rsavitski:
Define a template for simple kernel slice/counter trace events that are
best visualised on tracks. Perfetto will show such events in a predictable
way, without requiring any changes to perfetto code itself.

Public documentation in a separate patch.

Notable limitations/punts:
* decided not to merge thread-scoped track events with the "main" thread,
  though I could be convinced otherwise.
* only hooking onto "generic ftrace events", so if someone decides to
  generate dedicated protos for such trackevents, we'll need to do an
  additional refactoring to make it easy to call MaybeParseAsTrackEvent().
* android: since this is based on generic events, this is for local
  use-cases only, bytecode and logs will filter all of this out.
* trailing fields could be attached as args to the slices, but not for v1.


src/trace_processor/importers/ftrace/generic_ftrace_tracker.cc
```diff
@@ -20,18 +20,116 @@
...
+
+// Track blueprints for kernel track events.
+//
+// TODO(rsavitski): improve track naming in the UI for cpu & custom scoped
+// tracks, as getTrackName() in dev.perfetto.TraceProcessorTrack won't
+// automatically add those dimensions as a suffix, leading to hard to
+// distinguish tracks in the UI. Alternatively, move these tracks to the ftrace
+// plugin in the UI.
+//
+// clang-format off
```
LalitMaganti: This should not be necessary.

ui/src/plugins/dev.perfetto.StandardGroups/index.ts
```diff
@@ -45,6 +46,7 @@ export default class implements PerfettoPlugin {
     NETWORK: makeGroupNode('Network'),
     DEVICE_STATE: makeGroupNode('Device State'),
     SYSTEM: makeGroupNode('System'),
+    KERNEL: makeGroupNode('Kernel'),
```
LalitMaganti, line 49: I would actually call this LINUX to distinguish from other kernels we now support (e.g. QNX).

src/trace_processor/importers/ftrace/generic_ftrace_tracker.cc
```diff
@@ -20,18 +20,116 @@
...
+
 // We do not expect tracepoints with over 32 fields. It's more likely that the
 // trace is corrupted. See also |kMaxFtraceEventFields| in ftrace_descriptors.h.
 static constexpr uint32_t kMaxAllowedFields = 32;
 
+static constexpr char kScopeFieldNamePrefix[] = "scope_";
+
+// Track blueprints for kernel track events.
+//
+// TODO(rsavitski): improve track naming in the UI for cpu & custom scoped
```
LalitMaganti: Shouldn't we just do this now? I feel like it shouldn't be that much work and would add non-trivial benefits.

LalitMaganti (changes requested):

LalitMaganti:
Also this needs a UI integration test too

src/trace_processor/importers/ftrace/generic_ftrace_tracker.cc
rsavitski: I preferred this over:
* autoformatted layout, since some names end up wrapped differently, breaking the regularity
* autoformatter with tactical "//" suffixes to manually linebreak the couple of offenders

Reverted back to "//" since it is more friendly to refactorings.

src/trace_processor/importers/ftrace/generic_ftrace_tracker.cc
rsavitski: Done.

ui/src/plugins/dev.perfetto.StandardGroups/index.ts
rsavitski: I disagree, the tracks describe some kernel events, the fact that they're currently only derived from ftrace is an implementation detail. In my view it's similar to how we don't suffix scheduling tracks with the OS name atm, even if there could be different OSes in multi-machine tracing.

rsavitski:
> Also this needs a UI integration test too

Done, PTAL.

src/trace_processor/importers/ftrace/generic_ftrace_tracker.cc
```diff
@@ -20,18 +20,107 @@
...
+namespace {
+
 // We do not expect tracepoints with over 32 fields. It's more likely that the
 // trace is corrupted. See also |kMaxFtraceEventFields| in ftrace_descriptors.h.
 static constexpr uint32_t kMaxAllowedFields = 32;
 
+static constexpr char kScopeFieldNamePrefix[] = "scope_";
+
+// Track blueprints for kernel track events.
+constexpr auto kThreadSliceTrackBp = tracks::SliceBlueprint(  //
```
LalitMaganti, line 44: nit: the empty comment at end of line feels off?

rsavitski: It's to hint a linebreak for the autoformatter, as per the resolved comment thread.

PR #2453, opened 2025-08-07
doc: Add exlucsive_prio info

sashwinbalaji:


docs/learning-more/android.md
```diff
@@ -169,3 +170,54 @@ around.
 
 The full reference for the `perfetto` cmdline interface can be found
 [here](/docs/reference/perfetto-cli.md).
+
+## Exclusive Tracing Sessions
+
+Perfetto is designed to support [multiple concurrent tracing sessions](concepts/concurrent-tracing-sessions.md) from different sources (e.g., adb, on-device apps, automated testing). While this works for most data sources, some advanced features cannot be reliably multiplexed and require a guarantee that no other tracing session is active.
```
LalitMaganti: Another reason is to do tracing knowing that other tracing is not interfering. Important for sensitive perf measurements or in situations where you want to minimize other things going on in system.

docs/learning-more/android.md
```diff
@@ -169,3 +170,54 @@ around.
...
+
+To address this, Perfetto offers an "exclusive" mode. When a session is started in exclusive mode, it ensures no other sessions are running, providing a clean tracing environment. This is controlled by the `exclusive_prio` field in the `TraceConfig`.
+
+### When to use exclusive mode
+
+You should use an exclusive session when tracing with data sources that:
+*   Require a pristine state for accurate measurements, free from the potential interference of other concurrent traces.
+*   Have high overhead (e.g., `function_graph` in ftrace).
```
LalitMaganti: It also applies for the ftrace buffer size kb parameter which only applies to the first session that sets it.

PR #2455, opened 2025-08-07
docs: reword ftrace guide + add kernel track event docs

rsavitski:
Firstly, this patch rewrites the existing ftrace guide to be more to the
point and with fewer leaps of reasoning. I'm still not too happy with
the page (some instructions are duplicated too much across our
quickstarts, while other things are namedropped but not obvious), but
believe it's better than before.

Secondly, it adds a new section to the quickstart guide covering a few
example "kernel trackevent" macros for common cases like process-scoped
thread tracks. This includes examples of a declaration, callsite, and
the resulting UI. 

Lastly, it adds a reference page outlining all options for the
well-known fields for this "kernel trackevent" parsing.


docs/getting-started/ftrace.md
```diff
@@ -1,60 +1,81 @@
...
+options][funcgraph-cfg-link] for more details.
+
+This page is split into three sections:
+- [Part A][section-a-link]: covers creating an example kernel module with a
+  static tracepoint, recording the events with perfetto, and viewing the basic
+  event data in the perfetto UI.
+- [Part B][section-b-link] (advanced): covers adding dedicated parsing for the
+  tracepoint by modifying perfetto sources. This lets you take full
+  advantage of turning your events into structured SQL tables and UI tracks.
+- [Part C][section-c-link]: as an alternative to the above, describes a
```
LalitMaganti, line 47: I would suggest swapping part B and part C. I would imagine it's what most people will want and so should appear first in the docs.

LalitMaganti (approved):

docs/getting-started/ftrace.md
rsavitski: I'd prefer to keep it this way for now since the doc is already written to flow this way. Once we're more comfortable with calling this stable (and the ftrace config option becomes the default), I'll swap the order.

PR #2462, opened 2025-08-08
ui: plugin: Enhance PixelMemory to support multiple PID inputs

martinliu628:
- Adds support for analyzing multiple PIDs simultaneously.

- When multiple PIDs are provided, creates an aggregated "Total" track showing their combined memory usage.

- Automatically adds a visual annotation to the timeline to mark the point of maximum memory.

Bug: 437214050
Test: run plugin and verify the result with single/multiple pids

### DIFF
--- a/ui/src/plugins/com.google.PixelMemory/index.ts
+++ b/ui/src/plugins/com.google.PixelMemory/index.ts
@@ -18,6 +18,13 @@ import {PerfettoPlugin} from '../../public/plugin';
 import {Trace} from '../../public/trace';
 import {NUM} from '../../trace_processor/query_result';
 import {randomColor} from '../../components/colorizer';
+import {
+  createView,
+  createVirtualTable,
+  createPerfettoTable,
+  DisposableSqlEntity,
+} from '../../trace_processor/sql_utils';
+import {AsyncDisposableStack} from '../../base/disposable_stack';
 
 const CREATE_BREAKDOWN_TABLE_SQL = `
   DROP TABLE IF EXISTS process_memory_breakdown;
@@ -28,6 +35,8 @@ const CREATE_BREAKDOWN_TABLE_SQL = `
       memory_rss_and_swap_per_process PARTITIONED upid
     );
 `;
+
+const MAX_AGGREGATED_PIDS = 5;
 
 export default class implements PerfettoPlugin {
   static readonly id = 'com.google.PixelMemory';
@@ -59,7 +68,7 @@ export default class implements PerfettoPlugin {
         const maxTs = BigInt(maxRow.ts);
         const maxValueInBytes = maxRow.value;
         const maxValueInKib = (maxValueInBytes / 1024.0).toFixed(2);
-        const noteText = `Max (${noteTarget}): ${maxValueInKib} KiB`;
+        const noteText = `${maxValueInKib} KiB : Max PID (${noteTarget})`;
         const color = randomColor();
 
         ctx.notes.addNote({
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.google.PixelMemory/index.ts
```diff
@@ -41,6 +44,151 @@ export default class implements PerfettoPlugin {
...
+      previousTableName = newJoinedTableName;
+    }
+
+    const finalSelectTable = previousTableName;
+    const sumOfValues = valueColNames
+      .map((col) => `IFNULL(${col}, 0)`)
+      .join(' + ');
+
+    // 3. Materialize the aggregated sum into a standard table
+    await ctx.engine.query(`DROP TABLE IF EXISTS ${aggTableName}`);
```
LalitMaganti: You should use https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/trace_processor/sql_utils.ts;l=170?q=createPerfettoTable&ss=chromium to create this table. These should be PERFETTO TABLEs not plain SQLite tables.

ui/src/plugins/com.google.PixelMemory/index.ts
```diff
@@ -41,6 +44,151 @@ export default class implements PerfettoPlugin {
...
+        FROM process_memory_breakdown
+        WHERE pid = ${pidList[i]};
+      `;
+      await ctx.engine.query(createViewSql);
+    }
+
+    // 2. Iteratively SPAN_OUTER_JOIN the views together.
+    let previousTableName = viewNames[0];
+    for (let i = 1; i < pidList.length; i++) {
+      const newJoinedTableName = `__joined_${runId}_${i}`;
```
LalitMaganti, line 125: a) this is a really terrible way to do this, it will 100% tank in performance for more than a handful of pids. If you plan on sticking with this approach, please add an explicit limit of 5 pids to avoid tanking UI performance if someone accidentally passes a lot of pids
b) AFAIK, you should be able to use https://perfetto.dev/docs/analysis/stdlib-docs#package-intervals to solve this problem properly: it was designed for precisely this case of "I have a table of intervals which I want to intersect which each other".

ui/src/plugins/com.google.PixelMemory/index.ts
```diff
@@ -41,6 +44,151 @@ export default class implements PerfettoPlugin {
...
+  ): Promise<string> {
+    const runId = id.replace(/[#\.]/g, '_');
+    const viewNames = pidList.map((pid) => `__view_${runId}_${pid}`);
+    const valueColNames = pidList.map((pid) => `value_${pid}`);
+    const aggTableName = `__agg_${runId}`;
+
+    // 1. Create a separate VIEW for each PID's memory data.
+    for (let i = 0; i < pidList.length; i++) {
+      const createViewSql = `
+        DROP VIEW IF EXISTS ${viewNames[i]};
```
LalitMaganti: Never drop tables like this in the UI. Also use https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/trace_processor/sql_utils.ts;l=230?q=createPerfettoTable&ss=chromium

ui/src/plugins/com.google.PixelMemory/index.ts
martinliu628: Thanks for reviewing the code. Sorry, I'm pretty bad at SQL and thanks for the pointer. I'll try to figure out how to use that library.

ui/src/plugins/com.google.PixelMemory/index.ts
martinliu628: Thanks, I will learn how to use this.

ui/src/plugins/com.google.PixelMemory/index.ts
martinliu628: Okay. I will try this.

ui/src/plugins/com.google.PixelMemory/index.ts
martinliu628: As discussion, we add 5 pids limit since intervals library has some corner cases that might not work. Thanks.

PR #2468, opened 2025-08-08
[SDK] Add API to control sibling merge behavior on named track

etiennep-chromium:
This PR adds helper functions to control sibling merge behavior on NamedTrack


ddiproietto (commented):
Nice, thanks. How about some tests? At least to make sure that it compiles and doesn't crash (no need to check for the serialization)

etiennep-chromium:
Added test!

etiennep-chromium:
Added tests!

PR #2471, opened 2025-08-08
ui: Add expand and pin function using plugin paramters

pengj:
This PR added the auto-expand and pin function mentioned in this [feature request](https://github.com/google/perfetto/issues/2400) for the AutoPinAndExpandTracks plugin. 

**The changes in this PR:**

* parse the expand and pin tracks using initialPluginRouteArgs
* rename the plugin RestorePinnedTrack to AutoPinAndExpandTracks

The expand and pin tracked will passed by URL paramters with format: 
```
dev.perfetto.AutoPinAndExpandTracks:expand_tracks_with_name_on_startup=(track_name_1--track_name_2)
dev.perfetto.AutoPinAndExpandTracks:pin_tracks_with_name_on_startup=(track_name_1--track_name_2)
```

**Test**

run the cmd: ui/run-all-tests to make sure no failed test

Test it with this URL: http://localhost:10000/#!/viewer?local_cache_key=2fb82c7d-bfaf-6dbe-6e49-6f25e9bdf84a&dev.perfetto.AutoPinAndExpandTracks:expand_tracks_with_name_on_startup=(com.android.vending--com.instagram.android)&dev.perfetto.AutoPinAndExpandTracks:pin_tracks_with_name_on_startup=(FinskyEvent)

Works as expected 

Screenshot 2025-08-08 at 15 43 20


ui/src/frontend/index.ts
```diff
@@ -311,7 +311,8 @@ function onCssLoaded() {
   NON_CORE_PLUGINS.forEach((p) => pluginManager.registerPlugin(p));
   const route = Router.parseUrl(window.location.href);
   const overrides = (route.args.enablePlugins ?? '').split(',');
-  pluginManager.activatePlugins(overrides);
+  const pluginParams = (route.args.pluginParams ?? '').split(',');
```
LalitMaganti: This cannot be so simple. What if the params themselves have commas? Either we need to ban them explicitly or this function needs to be made more robust.

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -13,29 +13,46 @@
...
-const PLUGIN_ID = 'dev.perfetto.RestorePinnedTrack';
+const PLUGIN_ID = 'dev.perfetto.AutoPinAndExpandTracks';
 const SAVED_TRACKS_KEY = `${PLUGIN_ID}#savedPerfettoTracks`;
 
 const RESTORE_COMMAND_ID = `${PLUGIN_ID}#restore`;
 
+const URL_PARAM_EXPAND_TRACKS = 'expand_tracks_with_name_on_startup';
+const URL_PARAM_PINNED_TRACKS = 'pin_tracks_with_name_on_startup';
+
+function getParamValues(pluginParams: ReadonlyArray<string>, paramName: string): string[] {
```
LalitMaganti: Why do we need this function?

LalitMaganti (changes requested):

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -161,6 +181,24 @@ export default class implements PerfettoPlugin {
...
+    if (AutoPinAndExpandTracks.expandTracks.length > 0) {
+        this.ctx.workspace.flatTracks.filter((t) => AutoPinAndExpandTracks.expandTracks.some((prefix) => t.name.startsWith(prefix))).forEach((t) => t.expand());
+    }
+    
+
+    if (AutoPinAndExpandTracks.pinTracks.length > 0) {
+       this.ctx.workspace.flatTracks.filter((t) => AutoPinAndExpandTracks.pinTracks.some((prefix) => t.name.startsWith(prefix))).forEach((t) => t.pin());
+    }
+
+    //Once the traces have been processed, we don’t want to expand or pin the tracks again.
```
LalitMaganti: Please add a comma between // and first word

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -161,6 +181,24 @@ export default class implements PerfettoPlugin {
...
       },
     });
+
+     // Process URL parameters for auto-expanding groups and pinning tracks
+    this.processUrlParameters();
+  }
+
+  private processUrlParameters(): void {
+    if (AutoPinAndExpandTracks.expandTracks.length > 0) {
+        this.ctx.workspace.flatTracks.filter((t) => AutoPinAndExpandTracks.expandTracks.some((prefix) => t.name.startsWith(prefix))).forEach((t) => t.expand());
```
LalitMaganti: Would suggest switching to regex while we wait for Steve.

LalitMaganti: @stevegolton what should we do here? We are using regex instead of exact match because pids/tids are part of track names. Can we separate them or should we just use regexes

LalitMaganti (changes requested):

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
pengj: this is the same issue you mentioned above, what if parmas have commas? we either force the plugin to have a standard way to pass the k/v pair or give the plugin option to parse its own format. Here, we pass the pram as 
```
pluginParams=dev.perfetto.AutoPinAndExpandTracks%3Aexpand_tracks_with_name_on_startup%3D(com.android.vending--com.instagram.android),dev.perfetto.AutoPinAndExpandTracks%3Apin_tracks_with_name_on_startup%3D(FinskyEvent)
```

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
pengj: fogot this one, I have a regex match fix here.

ui/src/frontend/index.ts
pengj: I think we should restrict them within he params themselves, the plugin can decide what's the splitter they want to use within the plugin but not comma?

LalitMaganti: Well maybe suggest making the outer one a less common seprataor (maybe semi-colon?) and the inner one comman. Plugins are read/written by a lot more people so it makes more sense to make the more common separator there. 

Note that we should document whichever separator we do use to ensure that plugins don't use it themselves. It should be part of the plugin API contract documentation.

LalitMaganti: Alternatively we could wrap arguments with "" so we avoid this problem altogether

pengj: > Alternatively we could wrap arguments with "" so we avoid this problem altogether

some more explanation here?

LalitMaganti: You could wrap the plugin arguments with double quotes so that so that if a comma appears inside the quotation marks, you can just ignore it. Common technique to avoid these sorts of problems.

pengj: so, we don't need to change anything here? we just need to document to make sure people aware they need to wrap argument with "" and let plugin handle it? or we still need to parse "" here something like:
`pluginParams=plugin:"", plugin2:""`

LalitMaganti: We will of course need to track " for the purposes of not splitting commas inside the ".

pengj: Could I leave this part to you to add the check? I am not familiar with how we track and check here. You may do it in 5m and I need to check all these basic concepts. 😄

LalitMaganti: Sorry my point was we need to change the the logic here to be a function which splits on commas while *not* splitting on commas which are inside a plugin argument: this requires you to keep track of whether or not you are inside a plugin argument by iterating through character by character and only split on comma if you are not. 

Basically a custom split function.

pengj: something like this:
```
function customSplit(input) {
  const result = [];
  let current = '';
  let inQuotes = false;
  for (let i = 0; i < input.length; i++) {
    const char = input[i];
    if (char === '"') {
      // Toggle inQuotes when encountering a quote
      inQuotes = !inQuotes;
      current += char;
    } else if (char === ',' && !inQuotes) {
      // Only split on commas outside of quotes
      result.push(current.trim());
      current = '';
    } else {
      current += char;
    }
  }
  // Add the last segment
  if (current) {
    result.push(current.trim());
  }
  return result;
}
```
const input = 'plugin-1:"something", plugin-2:"something, else", plugin-3:"another"';
console.log(customSplit(input));
// Output: ['plugin-1:"something"', 'plugin-2:"something, else"', 'plugin-3:"another"']

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
pengj: want to check here, we still need this function, I think we should let plugin to handle how they parse the value

LalitMaganti: I would suggest we provide this to plugins as a default "argument parsing" plugin from the core. Plugins are free to use this or not but we should really avoid them having to add this code unnecessarily.

ui/src/frontend/index.ts
LalitMaganti: Yeah seems about right.

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
pengj: can I move this change to another PR? so I can also remove the change within PinAndroidPerfMetrics https://github.com/google/perfetto/blob/000f15c69118661a6a3a5d4ab21dbf900f318b2b/ui/src/plugins/dev.perfetto.PinAndroidPerfMetrics/index.ts#L26-L43

ui/src/frontend/index.ts
LalitMaganti: Needs some unittests btw.

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
LalitMaganti: I don't mind but you need to write unittests for this code which is somewhat eaiser if it's in the final place.

pengj: yes, it we remove this `argument parsing` to the core class, we only test that core class

ui/src/frontend/index.ts
pengj: the test is under test folder (https://github.com/google/perfetto/tree/main/ui/src/test), should I just create a file and just test this function? as I didn't see we have one test to check this index file

LalitMaganti: Tests should liove in the same folder named the same as the file which contains it and should add a _unittest file.

I would probably put this code inside the PluginManager file as a free function (still called from here though) and then add a plugin_manager_unittest.ts file.

Here's an example of unittest: https://github.com/google/perfetto/blob/main/ui/src/trace_processor/dataset_unittest.ts

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
pengj: move the getParamValues to plugin manager and add unit tests

ui/src/core/default_plugins.ts
```diff
@@ -33,6 +33,7 @@ export const defaultPlugins = [
   'dev.perfetto.AndroidPerf',
   'dev.perfetto.AndroidPerfTraceCounters',
   'dev.perfetto.AndroidStartup',
+  'dev.perfetto.AutoPinAndExpandTracks',
```
LalitMaganti, line 36: 'dev.perfetto.RestorePinnedTrack', below needs to be removed.

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -161,6 +177,35 @@ export default class implements PerfettoPlugin {
...
+
+    // Process URL parameters for auto-expanding groups and pinning tracks
+    this.processUrlParameters();
+  }
+
+  private processUrlParameters(): void {
+    if (AutoPinAndExpandTracks.expandTracks.length > 0) {
+      this.ctx.workspace.flatTracks
+        .filter((t) =>
+          AutoPinAndExpandTracks.expandTracks.some((prefix) =>
```
LalitMaganti: Would materialize the regex creation to avoid n x m regexes being created.

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -161,6 +177,35 @@ export default class implements PerfettoPlugin {
...
+            t.name.match(new RegExp('^' + prefix)),
+          ),
+        )
+        .forEach((t) => t.expand());
+    }
+
+    if (AutoPinAndExpandTracks.pinTracks.length > 0) {
+      this.ctx.workspace.flatTracks
+        .filter((t) =>
+          AutoPinAndExpandTracks.pinTracks.some((prefix) =>
```
LalitMaganti: As above.

LalitMaganti (changes requested):

pengj:
@stevegolton Thanks for the information, this is really helpful. 
Before I update the PR, I want to align on what we want to change here: 
1. use the `initialPluginRouteArgs` to pass/retrieve the URL arg
2. can I still keep the `getParamValues` in the plugin_manager, since it help parse the param values. From the CI error, looks like, I can't import the function from 
`Import not allowed /plugins/dev.perfetto.AutoPinAndExpandTracks/index -> /core/plugin_manager `
3. keep the changes for the AutoPinAndExpandTracks

Are there any other things you think should be changed or that I might have missed?

LalitMaganti (approved):

pengj:
@stevegolton  when you have time, could you have a look the changes? Thanks!

ui/src/plugins/dev.perfetto.AutoPinAndExpandTracks/index.ts
```diff
@@ -13,29 +13,45 @@
...
+
+// Parse the plugin parameters values, which are split by -- and wrapped by ()
+// Examples: (value1--value2)
+function getParamValues(pluginParams: string): string[] {
+  const match = pluginParams.match(/^\((.*)\)$/);
+  if (match) {
+    return match[1].split('--').map((v) => v.trim());
+  }
+  return [];
+}
```
stevegolton: I'm a little bit skeptical about using this format to represent an array of arguments.

- What if `--` needs to appear in a regex?
- What are the parenthesis for?

Let's examine the various 'standard' methods of representing arrays in URL search params:

1. Repeated keys: `?regex=foo&regex=bar`
- Clear and easy to parse out using standard tools `new URLSearchParams(window.location.search).getAll('regex')`.
- Each regex could be percent encoded separately, which would allow commas to be included in the regex itself.
- We must prefix each key with the plugin name which leads to long URLs.
- initialPluginRouteArgs doesn't currently support this IIUC but we could and should support it.

2. Comma separated: `?regex=foo,bar`
- A pseudo standard.
- Shorter URLs as we only need the prefix once.
- I wouldn't feel comfortable parsing commas into arrays automatically for `initialPluginRouteArgs`, as this would prohibit legitimate string values with commas from being passed.
- Doesn't work very well with regexes, we would need to encode the regexes once separately, then URL encode them, which would be a mess.

I'm not 100% sure what the best approach is. I'm leaning towards the former, as this avoids having to double encode the regexes which is messy. You just need to apply standard URL percent encoding. However we would need to make some changes to the way `initialPluginRouteArgs` are parsed for this to be properly conveyed to the user and that's not a super high priority for me right now. If you wanted to have a go at it, I'd be happy to review.

I'm picking this apart because this plugin is has the `dev.perfetto.` prefix which indicates that it's owned and maintained by us, and from what I understand is designed to be a generic way to provide startup args to Perfetto. Once folks start picking it up, this URL format becomes an interface that becomes very difficult to change.

pengj: @stevegolton thanks for the review, I am also learning for the first option. If we think the first option is the way to go, I can update this diff to just support one key: value for now, so we can land this one first. Then we can add the multiple one support in another PR. what do you think? 

The url will be generated like this as a starting point: 
http://localhost:10000/#!/viewer?local_cache_key=2fb82c7d-bfaf-6dbe-6e49-6f25e9bdf84a&dev.perfetto.AutoPinAndExpandTracks:expand_tracks_with_name_on_startup=com.android.vending&dev.perfetto.AutoPinAndExpandTracks:pin_tracks_with_name_on_startup=FinskyEvent

pengj:
@stevegolton updated the parse function to only take one value for each parameters for now. I will check `initialPluginRouteArgs` how to add support mutiple values later.

stevegolton (approved):
Looks good to me, thanks!

Issue #2473, opened 2025-08-08
trace_processor losts Render stages


olehkuznetsov:
[https://ui.perfetto.dev/#!/?s=19a89d2f8ef454b06eb0feb2195b16f9b38c4979](Perfetto file)

From 4 Vertex render stages per submission UI Perfetto shows only 1. They are missing in related SQL table, so, seems as trace_processor issue
<img width="1680" height="657" alt="Image" src="https://github.com/user-attachments/assets/054ba3e2-6e14-4487-95d6-9c3ffab937d3" />

AGI uses older perfetto and shows all 4
<img width="2435" height="679" alt="Image" src="https://github.com/user-attachments/assets/ca4ea0d1-411e-4ea2-96ff-e57752fa93f9" />

LalitMaganti:
Seems like this only worked in the past due to two bugs cancelling out and causing events to look like they parsed correctly (even though can see in the AGI output that the depth computation is busted). We fixed one of the bugs (disallowing overlapping events on a single track) but didn't fix the parsing of render stage events to distribute on multiple tracks using TrackComprsssor

Will fix the parsing of these events.

LalitMaganti:
https://github.com/google/perfetto/pull/2477 fixes.

New state in Perfetto:

<img width="2934" height="1458" alt="Image" src="https://github.com/user-attachments/assets/353e9b82-8144-4d50-b70d-d7b37657580a" />

PR #2477, opened 2025-08-08
tp: Fix overlapping gpu_render_stage slices being dropped

LalitMaganti:
The SliceTracker was silently dropping overlapping gpu_render_stage_event
slices that occurred on the same hardware queue.

This change moves the gpu_render_stage event parsing to use the
TrackCompressor API. To prevent overlaps, TrackCompressor creates the
minimal number of tracks necessary for a given hardware queue, placing
conflicting slices on different tracks. These tracks are associated with
the same hw_queue_id, allowing the UI to compress them into a single
visual track with stacked events. This ensures no slice data is lost.

A new diff test is added to verify that two overlapping events on the
same hw_queue_id are both processed and appear correctly in the
gpu_slice table.

Fixes: https://github.com/google/perfetto/issues/2473

### DIFF
--- a/include/perfetto/ext/base/status_macros.h
+++ b/include/perfetto/ext/base/status_macros.h
@@ -21,11 +21,11 @@
 
 // Evaluates |expr|, which should return a base::Status. If the status is an
 // error status, returns the status from the current function.
-#define RETURN_IF_ERROR(expr)                           \
-  do {                                                  \
-    base::Status status_macro_internal_status = (expr); \
-    if (!status_macro_internal_status.ok())             \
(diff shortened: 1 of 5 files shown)


zzyiwei:
This seems to be the culprit of broken renderstages (only a single set of renderstages tracks are shown instead of per-process). The renderstages are pre-process events. The `hw_queue_iid` interned data is identical across different processes, which is used to link the events within the same process. So the tracks from different `trusted_pid` must be separated.

Edit: I've filed https://github.com/google/perfetto/issues/3030 instead.

PR #2479, opened 2025-08-08
 Revert "tests: Refactor diff_tests scripts (#2418)"

JoeNotCharlesGoogle:
Also revert the dependent patch "tests: Measure longest running tests (#2467)"

This refactor breaks Chrome's `run_diff_tests` target, because the test runner doesn't deal with file paths outside of a standalone Perfetto build. Chrome checks out a copy of Perfetto at `.../chromium/src/third_party/perfetto`, and puts its build output at `.../chromium/src/out/Default`. The new test_executor.py sets `protos_dir` to ".../chromium/src/third_party/perfetto/**out/Default/gen**/protos", but from a Chrome checkout it should be ".../chromium/src/**out/Default/gen**/third_party/perfetto/protos".

As a result the tests all fail with:

```
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/google/home/joenotcharles/src/chromium/src/third_party/perfetto/out/Default/gen/protos/perfetto/trace_processor'
```

Because "trace_processor" is actually located at `/usr/local/google/home/joenotcharles/src/chromium/src/out/Default/gen/third_party/perfetto/protos/perfetto/trace_processor`


JoeNotCharlesGoogle:
Shoot, the revert's not as clean as it looked. Probably easier just to fix the path. Probably need to pass in a value for ROOT_DIR that can be set from BUILD.gn, and use the default from test_executor if it's not set.

JoeNotCharlesGoogle:
> Shoot, the revert's not as clean as it looked. Probably easier just to fix the path. Probably need to pass in a value for ROOT_DIR that can be set from BUILD.gn, and use the default from test_executor if it's not set.

Fixed the error.

I looked into passing in a ROOT_DIR, but it's really confusing because the variable's used for a lot of paths and I'm not sure if they all need to be treated the same.

JoeNotCharlesGoogle:
> > Shoot, the revert's not as clean as it looked. Probably easier just to fix the path. Probably need to pass in a value for ROOT_DIR that can be set from BUILD.gn, and use the default from test_executor if it's not set.
> 
> Fixed the error.
> 
> I looked into passing in a ROOT_DIR, but it's really confusing because the variable's used for a lot of paths and I'm not sure if they all need to be treated the same.

My diagnosis was wrong: the problem is that tests with BinaryProto output have NEVER worked from a chrome checkout, because the path to the descriptors to have always been wrong. Chrome overrides the list of tests using base/tracing/test/trace_processor/diff_tests/include_index.py (to a list that doesn't use BinaryProto) but the refactor ignores the override and tries to run all the tests.

Issue #2481, opened 2025-08-08
Warning when traced_perf is not started


Ecordonnier:
Hi,
I suggest adding a warning when callstack_sampling is used in the perfetto configuration, and traced_perf is not started on the system. I was debugging why this was not working on a vanilla linux system (my perfetto traces did not contain call stacks), and it is really not obvious why the call-stacks are not populated when this happens (I have also not found a verbosity flag in tracebox which would log the issue, maybe I missed it?).

Also, the recording instructions in the documentation does not mention the traced_perf daemon in the linux command line instructions at https://perfetto.dev/docs/getting-started/cpu-profiling , so purely following the instructions does not work.

Thanks!

LalitMaganti:
> I suggest adding a warning when callstack_sampling is used in the perfetto configuration, and traced_perf is not started on the system.

This is not really a workable solution: traced_perf is not special in the Perfetto world and it's *very* normal for data sources to be enabled in the Perfetto config which don't exist (i.e. they may be implemented by some app/system process which is not registered with Perfetto because it's not running  at the moment).

> Also, the recording instructions in the documentation does not mention the traced_perf daemon in the linux command line instructions at https://perfetto.dev/docs/getting-started/cpu-profiling , so purely following the instructions does not work.

This is certainly a bug and should be fixed. @rsavitski can you PTAL?

LalitMaganti:
Actually also reading the code tracebox *should* automatically spawn traced_perf without any user intervention: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/tracebox/tracebox.cc;l=225?q=tracebox%20traced_perf&ss=chromium

It's possible something is not working correctly here.

LalitMaganti:
Yeah actually, if I run `./tracebox --txt -c /tmp/config -o /tmp/trace`, I don't need to spawn traced_perf at all. It just loads for me...

So from everything I can see, the docs are correct and work out of the box.

Ecordonnier:
> Actually also reading the code tracebox _should_ automatically spawn traced_perf without any user intervention: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/tracebox/tracebox.cc;l=225?q=tracebox%20traced_perf&ss=chromium
> 
> It's possible something is not working correctly here.

Not in my case,  since Perfetto was compiled with gcc, and this disables the build of traced_perf as side-effect.

LalitMaganti:
[...]

PR #2483, opened 2025-08-10
tracebox: add elog if traced_perf is compiled out

rsavitski:
Minimal change to help users notice cases like
https://github.com/google/perfetto/issues/2481.

This is in the forking path, so it won't spam the "usage" invocations.
Doing something more complex like showing errors in UI is not worth it.
I'd rather spend time figuring out how to refactor traced_perf to not 
require libunwindstack outside of Android.


rsavitski:
Lalit, WDYT about this as a least-effort patch to make the linked issue easier to debug for users?

LalitMaganti:
Yeah makes sense, it's a good compromise solution. 

Lgtm.

LalitMaganti (approved):

PR #2484, opened 2025-08-11
tracing: Log exclusive session rejection.

sashwinbalaji:
Add atoms:
- kTracedEnableTracingExclusiveSessionRejected
- kTracedEnableTracingExclusiveSessionNotAllowed

### DIFF
--- a/src/android_stats/perfetto_atoms.h
+++ b/src/android_stats/perfetto_atoms.h
@@ -74,8 +74,11 @@ enum class PerfettoStatsdAtom {
   kTracedEnableTracingInvalidTriggerMode = 52,
   kTracedEnableTracingInvalidBrFilename = 54,
   kTracedEnableTracingFailedSessionSemaphoreCheck = 57,
+  kTracedEnablePriorityBoostNoCapability = 60,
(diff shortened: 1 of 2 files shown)


src/android_stats/perfetto_atoms.h
```diff
@@ -74,8 +74,11 @@ enum class PerfettoStatsdAtom {
   kTracedEnableTracingInvalidTriggerMode = 52,
   kTracedEnableTracingInvalidBrFilename = 54,
   kTracedEnableTracingFailedSessionSemaphoreCheck = 57,
+  kTracedEnablePriorityBoostNoCapability = 60,
```
LalitMaganti: This doesn't seem to be used?

sashwinbalaji: yes, initially added it to align with go/atoms.proto, removed it to unblock the PR.

@KirillTim , if not being used shall we also remove this from the internal file ?

LalitMaganti: Was it ever used? Was including it an accident? 

If it was used at any point we cannot remove

PR #2490, opened 2025-08-11
tp: Expose build flags

aMayzner:
Create a __intrinsic_build_flags static table, that exposes the build flags used for building trace processor to SQL layer. 

Create a __intrinsic_modules static table, which informs the user what modules are currently enabled


include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
```diff
@@ -21,36 +21,75 @@
...
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_FORCE_DCHECK_ON() (0)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_FORCE_DCHECK_OFF() (0)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_VERBOSE_LOGS() (1)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_VERSION_GEN() (1)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_TP_PERCENTILE() (1)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_TP_LINENOISE() (0)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_TP_HTTPD() (1)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_TP_JSON() (0)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_TP_INSTRUMENTS() (0)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ANDROID_BUILD() (true)
```
LalitMaganti: Why is this being changed to true/false? For the preprocessor 1/0 is correct

gn/write_buildflag_header.py
```diff
@@ -13,20 +13,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This writes headers for build flags. See the gen_buildflags target in
-# /gn/BUILD.gn for usage.
+# This script generates a C++ header file with build flags.
```
LalitMaganti: This whole file has a much larger diff than I would expect.

include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
```diff
@@ -21,36 +21,75 @@
...
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_WATCHDOG() (PERFETTO_BUILDFLAG_DEFINE_PERFETTO_OS_ANDROID() || PERFETTO_BUILDFLAG_DEFINE_PERFETTO_OS_LINUX_BUT_NOT_QNX())
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_X64_CPU_OPT() (false)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ZLIB() (true)
 // clang-format on
+
+#include <vector>
+
+namespace perfetto::build_config {
+
+inline std::vector<std::pair<const char*, bool>> GetAllBuildFlags() {
```
LalitMaganti: This can be an array instead of a vector

include/perfetto/base/build_configs/bazel/perfetto_build_flags.h
```diff
@@ -21,36 +21,75 @@
...
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_TP_PERCENTILE() (true)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_TRACED_PERF() (false)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_VERBOSE_LOGS() (true)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_VERSION_GEN() (true)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_WATCHDOG() (0)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_X64_CPU_OPT() (false)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ZLIB() (true)
 // clang-format on
+
+#include <vector>
```
LalitMaganti: This file can be included in C code so cannot depend on c++ code

LalitMaganti (changes requested):

gn/write_buildflag_header.py
aMayzner: Just a cleanup, I needed to add a lot of code to resolve the flags, so at this point it was easier to do more OOP changes

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/tables_views.sql
```diff
@@ -1242,3 +1242,50 @@ SELECT
   target_node_id,
   importance
 FROM __intrinsic_memory_snapshot_edge;
+
+CREATE PERFETTO TABLE _modules (
```
LalitMaganti: This should be in C++ not here. That's where it will be in the future (will also be far less verbose in C++)

src/trace_processor/trace_processor_impl.cc
```diff
@@ -376,6 +376,35 @@ void InsertIntoTraceMetricsTable(sqlite3* db, const std::string& metric_name) {
...
+    tables::BuildFlagsTable::Row row;
+    row.name = string_pool->InternString(build_flag.name);
+    row.enabled = static_cast<uint32_t>(build_flag.value);
+    table->Insert(row);
+  }
+}
+
+void InsertIntoModulesTable(tables::ModulesTable* table,
+                            StringPool* string_pool) {
+  tables::ModulesTable::Row etm_row;
```
LalitMaganti: Instead of an enabled column, I would use presence or absence of row to decide whether module is enabled or not. This deals with the future situation where we might not know all the list of enabled modules upfront.

To do this, ifdef the adds of rows into this table with the corresponding build flags.

include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
```diff
@@ -21,6 +21,8 @@
 #define GEN_BUILD_CONFIG_PERFETTO_BUILD_FLAGS_H_
 
 // clang-format off
+#define PERFETTO_BUILDFLAG_COUNT 28
```
LalitMaganti: Instead of hardcoding, you can compute this by doing sizeof(PERFETTO_BUILDFLAGS) / sizeof(struct PerfettoBuildFlag)

include/perfetto/base/build_configs/bazel/perfetto_build_flags.h
```diff
@@ -21,6 +21,8 @@
 #define GEN_BUILD_CONFIG_PERFETTO_BUILD_FLAGS_H_
 
 // clang-format off
+#define PERFETTO_BUILDFLAG_COUNT sizeof(PERFETTO_BUILDFLAGS) / sizeof(struct PerfettoBuildFlag)
```
LalitMaganti: Can you a) make this a static constant variable instead of #define b) place this under the list of items below.

include/perfetto/base/build_configs/bazel/perfetto_build_flags.h
```diff
@@ -52,5 +53,44 @@
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_THREAD_SAFETY_ANNOTATIONS() (0)
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ENABLE_WINSCOPE() (1)
 
+struct PerfettoBuildFlag {
+  const char* name;
+  int value;
+};
+
+static const struct PerfettoBuildFlag PERFETTO_BUILDFLAGS[] = {
```
LalitMaganti: Any reason btw why this is not kPerfettoBuildflags as it's a normal variable?

LalitMaganti (approved):

PR #2498, opened 2025-08-11
tp: etm: rename trace to chunk

OmarElfouly:
In order to avoid using the overloaded trace term, an ETM trace is now an ETM chunk.

However it is important to note in the OSCD (Decoder for ETM) namespace an ETM chunk is an ETM trace.


LalitMaganti (approved):

OmarElfouly:
Ran the currently removed ETM test suite and they passed :)

PR #2500, opened 2025-08-11
etm mem hack for zips

OmarElfouly:
Fixes the current logic for handling zips so that it finalizes EtmTracker after the necessary processing is complete.


src/trace_processor/trace_processor_impl.cc
```diff
@@ -620,12 +620,6 @@ base::Status TraceProcessorImpl::NotifyEndOfFile() {
   // Last opportunity to flush all pending data.
   FlushInternal(false);
 
-#if PERFETTO_BUILDFLAG(PERFETTO_ENABLE_ETM_IMPORTER)
```
LalitMaganti, line 623: We are in the middle of doing v large rewrites of a lot of this code. Can you please just move this down below PerfTracker NotifyEndOfFile calll below and leave a TODO for moving this out instead?

OmarElfouly: Done 👍

PR #2513, opened 2025-08-12
ui: add plugin to display process bindings as seperate tracks

tmkupo:
Add command to display duration of bound services for a given client PID

Bug: 437088506
Test: load traces and check the plugin result.

Welcome to Perfetto!

For more details please see

### DIFF
--- a/ui/src/plugins/com.android.BinderFinder/index.ts
+++ b/ui/src/plugins/com.android.BinderFinder/index.ts
@@ -1,0 +1,98 @@
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.android.BinderFinder/index.ts
```diff
@@ -0,0 +1,98 @@
...
+          data: {
+            sqlSource: `
+              SELECT
+                server_ts AS ts,
+                server_dur AS dur,
+                server_process AS name
+              FROM android_binder_txns
+              WHERE client_pid = ${pid}
+              AND server_package_version_code IS NULL
+              AND server_process NOT LIKE "system_server"
```
LalitMaganti: Not sure why these are LIKE instead of just !=. Far less efficient.

It's also not clear to me why we are excluding these either.

LalitMaganti (commented):
This plugin needs an OWNERS file and an associated owner.

LalitMaganti:
+zezeozue who does a lot of stuff related to binder.

ui/src/plugins/com.android.BinderFinder/index.ts
tmkupo: I'll update to =

Excluding system_server and SurfaceFlinger as they are too noisy

LalitMaganti: This needs a comment and also likely a mention in the command name.

PR #2524, opened 2025-08-12
[tracks] Add merge key as an int

etiennep-chromium:
This PR adds int merge key to avoid privacy concerns around string ids in proto.
It also updates related trace_processor supports and track.h API.


src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -181,6 +181,8 @@ ModuleResult TrackEventTokenizer::TokenizeTrackDescriptorPacket(
         if (track.has_sibling_merge_key()) {
           reservation.sibling_merge_key =
               context_->storage->InternString(track.sibling_merge_key());
+        } else if (track.has_sibling_merge_key_int()) {
```
LalitMaganti, line 184: I think this is overly complex. I would actually just intern the stringified version of sibling_merge_key_int maybe with a prefix like "sibling_merge_key_int:%u" if you want to protect against accidental conflicts.

etiennep-chromium: Fair enough, done.

LalitMaganti (approved):

PR #2526, opened 2025-08-12
Use generated names for unknown Chrome processes and threads

JoeNotCharlesGoogle:
If a new entry is added to the chrome ProcessType (or ThreadType) enum without adding its name to the list in ChromeStringLookup, uses the generated enum name from the proto instead of kNullStringId. This ensures traces from the field that only contain ids will always have distinguishable names, if not always pretty ones.

### DIFF
--- a/src/trace_processor/importers/proto/chrome_string_lookup.cc
+++ b/src/trace_processor/importers/proto/chrome_string_lookup.cc
@@ -30,23 +30,26 @@ namespace trace_processor {
 
 namespace {
 
-// By design, these switches handle unrecognized enum values with a default
-// fallback, so that new entries can be added to
-// protos/third_party/chromium_chrome_enums.proto without immediately updating
-// the list of names.
-#if defined(__GNUC__)  // clang also supports GCC syntax.
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wcovered-switch-default"
-#pragma GCC diagnostic ignored "-Wswitch-enum"
-// -Wswitch includes all warnings from -Wswitch-enum except "enumeration value
-// not explicitly handled in switch".
-#pragma GCC diagnostic error "-Wswitch"
-#endif  // __GNUC__
+constexpr auto kThreadTypeMin =
+    ::perfetto::protos::pbzero::ChromeThreadDescriptor_ThreadType_MIN;
+constexpr auto kThreadTypeMax =
+    ::perfetto::protos::pbzero::ChromeThreadDescriptor_ThreadType_MAX;
+
+// Add Chrome process and thread names to these two functions to get
+// friendly-formatted names in field traces. If an entry is added to
+// chrome_enums.proto without adding it here, the enum value name from the proto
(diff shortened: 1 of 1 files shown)


src/trace_processor/importers/proto/chrome_string_lookup.cc
```diff
@@ -171,17 +173,42 @@ constexpr ThreadName kThreadNames[] = {
...
 
-ChromeStringLookup::ChromeStringLookup(TraceStorage* storage) {
-  for (uint32_t i = 0; i < base::ArraySize(kProcessNames); i++) {
-    chrome_process_name_ids_[kProcessNames[i].type] =
-        kProcessNames[i].name ? storage->InternString(kProcessNames[i].name)
-                              : kNullStringId;
+ChromeStringLookup::ChromeStringLookup(
+    TraceStorage* storage,
+    bool ignore_predefined_names_for_testing) {
+  std::map<chrome_enums::ProcessType, const char*> known_process_names;
```
LalitMaganti: Why do they need to be maps if they're enum keys. Can you not just use an array?

Binary size for small things like this really adds up in the long run...

JoeNotCharlesGoogle: Good point, this code is over-complicated. WDYT of this version that uses a switch?

src/trace_processor/importers/proto/chrome_string_lookup.cc
```diff
@@ -30,158 +30,248 @@ namespace trace_processor {
...
-    {ChromeThreadDescriptor::THREAD_DATABASE, "Database thread"},
-    {ChromeThreadDescriptor::THREAD_PROXYRESOLVER, "Proxy Resolver"},
-    {ChromeThreadDescriptor::THREAD_DEVTOOLSADB, "Chrome_DevToolsADBThread"},
-    {ChromeThreadDescriptor::THREAD_NETWORKCONFIGWATCHER,
-     "NetworkConfigWatcher"},
-    {ChromeThreadDescriptor::THREAD_WASAPI_RENDER, "wasapi_render_thread"},
-    {ChromeThreadDescriptor::THREAD_LOADER_LOCK_SAMPLER, "LoaderLockSampler"},
-    {ChromeThreadDescriptor::THREAD_COMPOSITOR_GPU, "CompositorGpuThread"},
-};
+// By design, these switches handle unrecognized enum values with a default
```
LalitMaganti: Why not cast to a unit and avoid all the compiler warnings? 😅

JoeNotCharlesGoogle: This still gives warnings for int values that aren't actually in the enum (eg. if someone pastes a PROCESS_FOO constant into the thread function by mistake).

LalitMaganti: Sure but this is super ugly hackery we're doing for quite a niche situation. I really don't think it's worth it honestly.

JoeNotCharlesGoogle: Done.

PR #2538, opened 2025-08-13
tp: create special pointer class to handle context forking

LalitMaganti:
Instead of using shared_ptr/unique_ptr, use a new smart pointer which
correctly handles both being generated as the root pointer and being
"forked" into a sub-context.

### DIFF
--- a/src/trace_processor/trace_processor_context.cc
+++ b/src/trace_processor/trace_processor_context.cc
@@ -45,35 +45,37 @@
 #include "src/trace_processor/importers/proto/proto_trace_reader.h"
 #include "src/trace_processor/storage/trace_storage.h"
 #include "src/trace_processor/trace_reader_registry.h"
+#include "src/trace_processor/types/trace_processor_context_ptr.h"
 
 namespace perfetto::trace_processor {
 namespace {
 
+template <typename T>
+using Ptr = TraceProcessorContextPtr<T>;
+
 void InitPerMachineState(TraceProcessorContext* context, uint32_t machine_id) {
   // Per-machine state.
-  context->machine_tracker =
-      std::make_unique<MachineTracker>(context, machine_id);
-  context->process_tracker = std::make_unique<ProcessTracker>(context);
-  context->clock_tracker = std::make_unique<ClockTracker>(context);
-  context->mapping_tracker = std::make_unique<MappingTracker>(context);
-  context->cpu_tracker = std::make_unique<CpuTracker>(context);
+  context->machine_tracker = Ptr<MachineTracker>::MakeRoot(context, machine_id);
+  context->process_tracker = Ptr<ProcessTracker>::MakeRoot(context);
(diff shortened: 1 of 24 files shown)


src/trace_processor/types/trace_processor_context.h
```diff
@@ -124,6 +125,7 @@ class TraceProcessorContext {
 
   GlobalPtr<MetadataTracker> metadata_tracker;
   GlobalPtr<Destructible> content_analyzer;
+  GlobalPtr<Destructible> heap_graph_tracker;  // HeapGraphTracker
```
sashwinbalaji, line 128: saw the todo in the other patch, but out of curiosity what's breaking by keeping it as permachine ?

src/trace_processor/types/trace_processor_context_ptr.h
```diff
@@ -0,0 +1,68 @@
...
+// both at the root level and recursively.
+template <typename T>
+struct TraceProcessorContextPtr {
+ public:
+  TraceProcessorContextPtr() = default;
+  explicit TraceProcessorContextPtr(std::unique_ptr<T> owned)
+      : owned_(std::move(owned)), ptr_(owned_.get()) {}
+  explicit TraceProcessorContextPtr(T* _ptr) : ptr_(_ptr) {}
+
+  // Makes the
```
sashwinbalaji: update

src/trace_processor/trace_processor_context.cc
```diff
@@ -152,24 +155,29 @@ std::optional<MachineId> TraceProcessorContext::machine_id() const {
 }
 
 void TraceProcessorContext::DestroyNonEssential() {
```
sashwinbalaji, line 157: not for this PR and already a todo present, but definitely something we should look into later.

src/trace_processor/types/trace_processor_context.h
```diff
@@ -60,13 +61,13 @@ using MachineId = tables::MachineTable::Id;
 class TraceProcessorContext {
  public:
   template <typename T>
```
sashwinbalaji, line 63: if i understand this correctly, all 3 being aliases, we cannot differentiate between them at compile time. And we use this for advantage in their intialization via `using Ptr = TraceProcessorContextPtr<T>;`. 

should we consider to have some more check for this that mismatch in ptr doesn't happen. maybe a strong typedef ?

src/trace_processor/types/trace_processor_context.h
LalitMaganti: Complex reasons to do with how legacy flamegraph generation works. In practice, once we delete legacy flamegraph, this should also be removed.

src/trace_processor/types/trace_processor_context.h
LalitMaganti: You're correct. Maybe we will make a stronger typing at some point but for now, I think just having these as annotations for both humans and LLMs is decently good for now.

PR #2554, opened 2025-08-14
ui: support larger time units in selection duration.

cphlipot1:
To better support longer durations, add support for printing years
and days if the selected time period is large enough.

image


stevegolton (approved):
Looks like the unit tests are failing, but otherwise looks good!

PR #2555, opened 2025-08-14
Make process_type and thread_type extensions, with the values in chrome_enums.proto.

JoeNotCharlesGoogle:
Deletes the Perfetto-internal copy of ProcessType, making third_party/chromium/chrome_enums.proto the source of truth. To avoid referencing the third_party proto from ChromeProcessDescriptor, which would be a dependency inversion, changes ChromeProcessDescriptor.process_type to an extension field, and adds a new chrome_descriptor_extensions.proto to hold it.

This causes protos/third_party/chromium to diverge from Chrome's base/tracing/protos, due to a dependency cycle between Perfetto and Chrome. Chrome will need to import the changes to
protos/perfetto/trace/track_event and update base/tracing/protos in one commit before the next sync from Chrome -> Perfetto.

Also moves ThreadType to chrome_enums.proto, and turns ChromeThreadDescriptor.thread_type into an extension, so that the manual import only needs to happen once.

This is a simplified version of commit bc8131f33358b2a76100b26c0b29dca33a7a0384 to cut down on the number of changes to perfetto code. It only moves the type enums to third_party/chromium, instead of the entire ChromeProcessDescriptor and ChromeThreadDescriptor messages, because only the enums are expected to change frequently.

(I have an even simpler version at https://github.com/google/perfetto/pull/2633, but it involves manually casting to and from integers. If you prefer that one, I'll withdraw this.)


JoeNotCharlesGoogle:
> Chrome will need to import the changes to
protos/perfetto/trace/track_event and update base/tracing/protos in one commit before the next sync from Chrome -> Perfetto.

FTR the draft commit to do this is https://crrev.com/c/6850891.

JoeNotCharlesGoogle:
Do not submit: When I tried to test import this into Chrome (https://crrev.com/c/6850891) I found out commit 78f26c40f07e5efce4abb1803da160c4dbb0e861 gives build errors due to different compiler flags used from Chrome. To avoid a really complicated deps roll, I'll wait until 78f26c40f07e5efce4abb1803da160c4dbb0e861 is rolled into Chrome and any breakage is sorted out before I land this.

JoeNotCharlesGoogle:
> Do not submit: When I tried to test import this into Chrome (https://crrev.com/c/6850891) I found out commit [78f26c4](https://github.com/google/perfetto/commit/78f26c40f07e5efce4abb1803da160c4dbb0e861) gives build errors due to different compiler flags used from Chrome. To avoid a really complicated deps roll, I'll wait until [78f26c4](https://github.com/google/perfetto/commit/78f26c40f07e5efce4abb1803da160c4dbb0e861) is rolled into Chrome and any breakage is sorted out before I land this.

This is https://github.com/google/perfetto/issues/2634, which is now fixed.

petrcermak (approved):

JoeNotCharlesGoogle:
Went with https://github.com/google/perfetto/pull/2633 intead.

PR #2570, opened 2025-08-14
ui: Shorter summary tracks

stevegolton:
image


ui/src/frontend/viewer_page/track_view.ts
```diff
@@ -59,18 +59,17 @@ import {Popup} from '../../widgets/popup';
 import {Theme} from '../../public/theme';
 
 const TRACK_HEIGHT_MIN_PX = 18;
-const TRACK_HEIGHT_DEFAULT_PX = 30;
 
 function getTrackHeight(node: TrackNode, track?: TrackRenderer) {
   // Headless tracks have an effective height of 0.
   if (node.headless) return 0;
 
   // Expanded summary tracks don't show any data, so make them a little more
```
LalitMaganti, line 67: Can we reduce height even when collapsed?

LalitMaganti (approved):

ui/src/frontend/viewer_page/track_view.ts
stevegolton: When collapsed, summary tracks will show track data if the have it - e.g. the process summary tracks. Thus they'll take the height of whatever the track renderer needs. If they have no track data to show, then they'll collapse to the minimum height.

PR #2575, opened 2025-08-14
tperf: default libunwindstack cache clearing to 5min (previously: none)

rsavitski:
Lower the chance that long configs end up integrating
unbounded amount of cached data in libunwindstack.

Bug: 435725820

### DIFF
--- a/protos/perfetto/trace/perfetto_trace.proto
+++ b/protos/perfetto/trace/perfetto_trace.proto
@@ -2048,7 +2048,7 @@ message PerfEventConfig {
   optional uint32 max_daemon_memory_kb = 13;
 
   //
-  // Uncommon options:
+  // Niche options:
   //
 
   // Timeout for the remote /proc/<pid>/{maps,mem} file descriptors for a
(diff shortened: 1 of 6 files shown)


protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -2061,10 +2061,16 @@ message PerfEventConfig {
   // If unset, an implementation-defined default is used.
   optional uint32 remote_descriptor_timeout_ms = 9;
 
-  // Optional period for clearing state cached by the unwinder. This is a heavy
-  // operation that is only necessary for traces that target a wide set of
-  // processes, and require the memory footprint to be reset periodically.
-  // If unset, the cached state will not be cleared.
+  // Optional period for clearing state cached by the userpsace unwinder. This
```
LalitMaganti: Given the current implementation, is there a way we can turn this off? We don't really look at 0 vs has so how do I hard disabled this?

rsavitski: I was thinking about setting an arbitrarily long value (longer than duration_ms). I could add that as a note in the proto comment, but probably don't want to add extra proto fields for this.

LalitMaganti: Yeah would suggest adding as a note

rsavitski: Done.

PR #2580, opened 2025-08-15
exp: Support sub query nodes

aMayzner:
Make it possible to add child nodes - nodes that apply changes to the source data in a scoped way, without modifying it. 

Currently they only support filtering and aggregation

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/sub_query_node.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/sub_query_node.ts
@@ -27,6 +27,7 @@ import {
   createGroupByProto,
 } from './operations/operation_component';
 import {ColumnInfo, newColumnInfoList} from './column_info';
+import {assertExists} from '../../../base/logging';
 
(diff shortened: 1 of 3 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/sub_query_node.ts
```diff
@@ -0,0 +1,95 @@
...
+  readonly nextNode?: QueryNode;
+  readonly sourceCols: ColumnInfo[];
+  readonly finalCols: ColumnInfo[];
+  readonly state: QueryNodeState;
+
+  constructor(state: QueryNodeState) {
+    this.nodeId = nextNodeId();
+    this.state = state;
+    this.prevNode = state.prevNode;
+    this.sourceCols = this.prevNode!.finalCols;
```
stevegolton, line 47: I understand that `prevNode` being undefined is a developer error, but we usually use `assertExists(this.prevNode)` instead of `!` so we know about it immediately.

stevegolton (approved):

PR #2587, opened 2025-08-15
tp: etm: binary info error fix

OmarElfouly:
As a temporary fix for b/438916722.
ETM registers files with large sizes causing TraceBlobView size to become truncated. This means we can not trust any of the below checks for large files. So a solution is needed that is not too expensive memory wise.

This PR converts these from errors to warnings and adds a TODO to find a memory efficient fix later.

### DIFF
--- a/src/trace_processor/importers/common/registered_file_tracker.cc
+++ b/src/trace_processor/importers/common/registered_file_tracker.cc
@@ -23,17 +23,17 @@
 
 #include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
-#include "perfetto/trace_processor/trace_blob_view.h"
-#include "src/trace_processor/util/elf/binary_info.h"
+#include "perfetto/trace_processor/trace_blob.h"
 #include "src/trace_processor/storage/trace_storage.h"
 #include "src/trace_processor/tables/etm_tables_py.h"
 #include "src/trace_processor/types/trace_processor_context.h"
 #include "src/trace_processor/util/build_id.h"
+#include "src/trace_processor/util/elf/binary_info.h"
 
 namespace perfetto::trace_processor {
 
 base::Status RegisteredFileTracker::AddFile(const std::string& name,
-                                            TraceBlobView data) {
+                                            TraceBlob data) {
   StringId name_id = context_->storage->InternString(name);
   auto* it = files_by_path_.Find(name_id);
   if (it) {
@@ -71,6 +71,11 @@ base::Status RegisteredFileTracker::AddFile(const std::string& name,
     files_by_build_id_.Insert(*build_id, id);
   }
   return base::OkStatus();
+}
+TraceBlob RegisteredFileTracker::GetContent(tables::FileTable::Id id) const {
+  PERFETTO_DCHECK(id.value < file_content_.size());
+  const auto& blob = file_content_[id.value];
+  return TraceBlob::CopyFrom(blob.data(), blob.size());
 }
 
 }  // namespace perfetto::trace_processor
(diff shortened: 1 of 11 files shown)


rasikanavarange:
> Having seen what this hack entails,rReally not a fan of this. This method of just ignoring errors, all this will do is cause errors in other places.
> 
> Is it possible to truncate the ELF file instead to uint32::max? That feels a more reasonable solution.

I think truncating the file might have consequences for ETM instruction decode (@OmarElfouly can confirm), if not then this might be okay?

The root of the issue is that TraceBlobView has the incorrect size for the underlying elf blob. So perhaps another workaround might be if [RegisteredFileTracker::AddFile()](https://github.com/google/perfetto/blob/main/src/trace_processor/importers/common/registered_file_tracker.h#L40) also takes a size_t with the correct size?

I still think it's not ideal and not a permanent solution but it might be better than just ignoring errors. WDYT?

OmarElfouly:
> Having seen what this hack entails,rReally not a fan of this. This method of just ignoring errors, all this will do is cause errors in other places.
> 
> Is it possible to truncate the ELF file instead to uint32::max? That feels a more reasonable solution.

Yeah at the moment it makes sense to me to remove it because when using it with chrome it falsely detects a Corrupted ELF and then fails later in the decode stage to use the ELF.

Maybe a better approach could be to keep the tests for files where size has a true value and otherwise emit a warning that file is too large to check for corrupted elf?

I am also trying to figure out what will go wrong in ETM decode if we reduce size to max

OmarElfouly:
Dependent on #2664 to work properly so must wait

src/trace_processor/importers/common/registered_file_tracker.cc
```diff
@@ -72,5 +72,10 @@ base::Status RegisteredFileTracker::AddFile(const std::string& name,
   }
   return base::OkStatus();
 }
+TraceBlob RegisteredFileTracker::GetContent(tables::FileTable::Id id) const {
```
LalitMaganti: Why do we need to copy, why not return a reference?

OmarElfouly: Thank you! Fixed :)

src/trace_processor/importers/etm/virtual_address_space.cc
```diff
@@ -49,17 +49,18 @@ void VirtualAddressSpace::Builder::AddMapping(
...
+        context_->registered_file_tracker->GetContent(*mmap.file_id());
+    auto file_range = AddressRange::FromStartAndSize(0, blob.size());
     auto required_file_range = AddressRange::FromStartAndSize(
         static_cast<uint64_t>(mapping.exact_offset()), range.size());
 
     PERFETTO_CHECK(file_range.Contains(required_file_range));
 
-    content = content->slice_off(required_file_range.start(),
-                                 required_file_range.length());
+    content = TraceBlob::CopyFrom(blob.data() + required_file_range.start(),
```
LalitMaganti, line 63: Please leave a comment that this is inefficient and we should switch back to using TraceBlobView.

OmarElfouly: @LalitMaganti  Yeah I didn't want to modify TraceBlob and give it a slice_off but using a TraceBlobView still has the issue of size() being wrong.
One work around could be that instead of crashing if TraceBlobView's size is too big we could maybe set a flag that causes TraceBlobView::size() to return blob().size()? Not sure if it is worth it though

LalitMaganti: I think it's fine to have this inefficiency for now, just leave a comment for the future for Rasika to cleanup when she fixes TraceBlobView

OmarElfouly: Done :)

LalitMaganti (approved):

PR #2588, opened 2025-08-15
tp: fix support for JITted frames in simpleperf profiles

LalitMaganti:
We were not correctly handling the addresses for those: they're
different to symbols for normal simpleperf ELF profiles so add special
case for handling them;

### DIFF
--- a/src/trace_processor/importers/common/symbol_tracker.cc
+++ b/src/trace_processor/importers/common/symbol_tracker.cc
@@ -64,13 +64,22 @@ bool SymbolTracker::TrySymbolizeFrame(
     return false;
   }
 
+  int64_t pc = frame.rel_pc();
+
   // Load bias is something we can only determine by looking at the actual elf
   // file. Thus PERF_RECORD_MMAP{2} events do not record it. So we need to
   // potentially do an adjustment here if the load_bias tracked in the mapping
   // table and the one reported by the file are mismatched.
-  uint64_t adj = file->load_bias - static_cast<uint64_t>(mapping.load_bias());
+  pc += static_cast<int64_t>(file->load_bias) - mapping.load_bias();
 
-  auto symbol = file->symbols.Find(static_cast<uint64_t>(frame.rel_pc()) + adj);
+  // If the symbols in the map are absolute, then we need to relativize against
+  // the exact offset and then add to the start of the mapping.
+  if (file->symbols_are_absolute) {
+    pc -= mapping.exact_offset();
+    pc += mapping.start();
+  }
+
+  auto symbol = file->symbols.Find(static_cast<uint64_t>(pc));
   if (symbol == file->symbols.end()) {
     return false;
   }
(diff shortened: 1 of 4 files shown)


src/trace_processor/importers/common/symbol_tracker.cc
```diff
@@ -64,13 +64,22 @@ bool SymbolTracker::TrySymbolizeFrame(
...
   // potentially do an adjustment here if the load_bias tracked in the mapping
   // table and the one reported by the file are mismatched.
-  uint64_t adj = file->load_bias - static_cast<uint64_t>(mapping.load_bias());
+  pc += static_cast<int64_t>(file->load_bias) - mapping.load_bias();
+
+  // If the symbols in the map are absolute, then we need to relativize against
+  // the exact offset and then add to the start of the mapping.
+  if (file->symbols_are_absolute) {
+    pc -= mapping.exact_offset();
+    pc += mapping.start();
```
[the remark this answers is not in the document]
LalitMaganti: This makes the code more complicated not simpler.

src/trace_processor/importers/perf/perf_tracker.cc
```diff
@@ -126,6 +127,10 @@ void PerfTracker::AddSimpleperfFile2(const FileFeature::Decoder& file) {
       return;
   }
 
+  // JIT functions use absolute addresses for their symbols instead of relative
+  // ones.
+  dso.symbols_are_absolute =
+      base::Contains(file.path().ToStdString(), "jit_app_cache");
```
rsavitski: Probably need to also handle `jit_zygote_cache` the same:

https://cs.android.com/android/platform/superproject/+/android-latest-release:system/extras/simpleperf/JITDebugReader.cpp?q=%22%20TempSymFile::Create(std::move(path),%20symfile_option_%20%3D%3D%20SymFileOption::kDropSymFiles);%22&ss=android%2Fplatform%2Fsuperproject

src/trace_processor/importers/common/symbol_tracker.cc
```diff
@@ -64,13 +64,22 @@ bool SymbolTracker::TrySymbolizeFrame(
...
   // Load bias is something we can only determine by looking at the actual elf
   // file. Thus PERF_RECORD_MMAP{2} events do not record it. So we need to
   // potentially do an adjustment here if the load_bias tracked in the mapping
   // table and the one reported by the file are mismatched.
-  uint64_t adj = file->load_bias - static_cast<uint64_t>(mapping.load_bias());
+  pc += static_cast<int64_t>(file->load_bias) - mapping.load_bias();
+
+  // If the symbols in the map are absolute, then we need to relativize against
+  // the exact offset and then add to the start of the mapping.
+  if (file->symbols_are_absolute) {
```
rsavitski, line 80: I don't fully follow this bit (I feel like the perf importer might be using exact_offset for something else than heapprofd/traced_perf, but I /still/ get hopelessly lost every time we start talking about load_bias and mapping offsets).

LGTM if it produces the right results, but could you add a TODO(rsavitski) to double-check if this is confusing "exact_offset for the purposes of llvm RO ELF header mappings" with "pgoff of the mapping".

LalitMaganti: Added the TODO, indeed this does produce the correct results.

src/trace_processor/importers/perf/perf_tracker.cc
LalitMaganti: Done.

PR #2589, opened 2025-08-15
f

LalitMaganti:


src/trace_processor/importers/perf/perf_event_attr.cc
```diff
@@ -149,18 +150,34 @@ PerfEventAttr::PerfEventAttr(TraceProcessorContext* context,
 
 PerfEventAttr::~PerfEventAttr() = default;
 
-PerfCounter& PerfEventAttr::GetOrCreateCounter(uint32_t cpu) {
-  auto it = counters_.find(cpu);
+PerfCounter& PerfEventAttr::GetOrCreateCounter(std::optional<uint32_t> cpu) {
+  if (!cpu) {
+    if (!global_counter_) {
+      base::StringView name(event_name_);
+      TrackId track_id = context_->track_tracker->InternTrack(
```
rsavitski: nit: this is now inconsistent, new inlined code is very similar to CreateCounter, which is instead factored out into a separate method. I'd suggest introducing a CreateGlobalCounter method.

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -209,10 +206,7 @@ export default class implements PerfettoPlugin {
     }
 
     if (perfCountersGroup.hasChildren) {
-      const hardwareGroup = trace.plugins
-        .getPlugin(StandardGroupsPlugin)
-        .getOrCreateStandardGroup(trace.workspace, 'HARDWARE');
-      hardwareGroup.addChildInOrder(perfCountersGroup);
+      trace.workspace.addChildInOrder(perfCountersGroup);
```
rsavitski, line 209: nit: iirc we have docs that mention "HARDWARE" group for the perf counters, could use a quick edit (but don't bother with screenshots).

test/trace_processor/diff_tests/parser/simpleperf/tests.py
```diff
@@ -397,3 +379,31 @@ def test_perf_summary_tree(self):
           8,"[NULL]","","/t1","[NULL]","[NULL]",87,87
           9,"[NULL]","","/elf","[NULL]","[NULL]",64,64
         '''))
+
+  def test_global_counters(self):
+    return DiffTestBlueprint(
+        trace=DataPath('perf_counter.data'),
```
rsavitski, line 389: Not actionable, but I wanted to mention for posterity (maybe turn it into a comment on the test?): interesting side-note is that the "perf record" implementation still decided to open a separate counter on each CPU (and then mapped all of them into a single buffer with ioctls) [1]. It may have had something to do with `inherit` complexity.

But I agree that it makes sense to interpret the samples as one virtual counter.

[1] from perf report --dump --header:
```
0 0x400 [0xc10]: PERF_RECORD_ID_INDEX nr: 96
 ... id: 3646  idx: 0  cpu: 0  tid: 433552
 ... id: 3647  idx: 1  cpu: 1  tid: 433552
 ... id: 3648  idx: 2  cpu: 2  tid: 433552
 ... id: 3649  idx: 3  cpu: 3  tid: 433552
 ... id: 3650  idx: 4  cpu: 4  tid: 433552
 ... id: 3651  idx: 5  cpu: 5  tid: 433552
 ... id: 3652  idx: 6  cpu: 6  tid: 433552
 ... id: 3653  idx: 7  cpu: 7  tid: 433552
 ... id: 3654  idx: 8  cpu: 8  tid: 433552

 <SNIP>
```

src/trace_processor/importers/perf/perf_event_attr.h
```diff
@@ -117,6 +118,7 @@ class PerfEventAttr : public RefCounted {
   std::optional<size_t> id_offset_from_start_;
   std::optional<size_t> id_offset_from_end_;
   size_t sample_id_size_;
+  std::unique_ptr<PerfCounter> global_counter_;
```
rsavitski, line 125: we should add a comment, I'm fairly familiar with perf terminology and "global counter" is still ambiguous (and how it's different to counters_).

Maybe:

```
// Counter not bound to a specific CPU. Might actually be an aggregate from multiple per-cpu counters in cases such as thread-scoped profiling.
std::unique_ptr<PerfCounter> global_counter_;
// Keyed by cpu index. Per-cpu counters.
std::unordered_map<uint32_t, PerfCounter> counters_;
```

rsavitski (changes requested):
Can we please have a longer description in the git commit body. The term "global counter" doesn't mean much to me, and neither did Carlos' use of "synthetic events". 

I'd like the git history to stay useful for code archaeology, especially here since the perf file format is complex and not ours.

LalitMaganti:
> Can we please have a longer description in the git commit body. The term "global counter" doesn't mean much to me, and neither did Carlos' use of "synthetic events".

Changed global to "non-CPU scoped" which I think should be clearer and also added some more commentary in the body of the commit.

PR #2590, opened 2025-08-15
Fix

LalitMaganti:


src/trace_processor/importers/perf/perf_tracker.cc
```diff
@@ -104,23 +104,23 @@ PerfTracker::CreateAuxDataTokenizer(AuxtraceInfoRecord info) {
 void PerfTracker::AddSimpleperfFile2(const FileFeature::Decoder& file) {
   SymbolTracker::Dso dso;
   switch (file.type()) {
-    case DsoType::DSO_KERNEL:
+    case DsoType::DSO_KERNEL: {
       InsertSymbols(file, context_->symbol_tracker->kernel_symbols());
```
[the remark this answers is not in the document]
LalitMaganti: For consistency.

src/trace_processor/importers/perf/perf_tracker.cc
```diff
@@ -104,23 +104,23 @@ PerfTracker::CreateAuxDataTokenizer(AuxtraceInfoRecord info) {
...
       ElfFile::Decoder elf(file.elf_file());
       dso.load_bias = file.min_vaddr() - elf.file_offset_of_min_vaddr();
       break;
     }
-
     case DsoType::DSO_KERNEL_MODULE: {
       KernelModule::Decoder module(file.kernel_module());
       dso.load_bias = file.min_vaddr() - module.memory_offset_of_min_vaddr();
       break;
     }
```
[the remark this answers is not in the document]
LalitMaganti: For consistency.

src/trace_processor/importers/perf/perf_tracker.cc
```diff
@@ -104,23 +104,23 @@ PerfTracker::CreateAuxDataTokenizer(AuxtraceInfoRecord info) {
...
-
-    case DsoType::DSO_DEX_FILE:
+    case DsoType::DSO_DEX_FILE: {
```
rsavitski, line 121: Preliminary question: is the dex_file_offset[1] not necessary? It's odd that it is part of the FILE2 sections (and is nonzero in the uploaded test file), but our decoder doesn't use it. Does it also need handling as if it was load_bias?

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/third_party/simpleperf/record_file.proto;l=62;drc=7068d201f5c0289d8cebe1c4d7a0de271ed5e08f

rsavitski (changes requested):

src/trace_processor/importers/perf/perf_tracker.cc
LalitMaganti: So in theory yes but I don't understand how to use it I'll be honest. It's a repeated field unlike the others which are optionals. Would appreciate some guidance on this if you know how to do so!

rsavitski: I'll go dive into simpleperf sources to see if I can figure it out based on encoding side.

My preliminary guess is that it means "this dex file has been mapped multiple times (piecewise?), and here are the offsets". So it may be purely informational if the symbol addresses are absolute.

rsavitski: Aha, "dex_file_offset" is the offset of the dex /within/ its containing file (since they're ususally stored with zero compression inside apks/jars). So we don't need this.

Though if possible, can we add a comment explaining the field to our `record_file.proto`?

---

For fun, I double-checked with `zipdetails` and the FILE2 dumped as textproto:

for an apk (=zip) with the following:
```
  dex_file: {
    dex_file_offset: [ 0x00000000000038b4, 0x000000000059a264 ]                                                                         
  }
```    

The apk does indeed have the dex files stored at those offsets (see PAYLOAD lines):
```
003888 LOCAL HEADER #14      04034B50 (67324752)
00388C Extract Zip Spec      00 (0) '0.0'
00388D Extract OS            00 (0) 'MS-DOS'
00388E General Purpose Flag  0000 (0)
003890 Compression Method    0000 (0) 'Stored'
003892 Modification Time     02210821 (35719201) 'Thu Jan  1 01:01:02 1981'
003896 CRC                   7AFA6C0E (2063232014)
00389A Compressed Size       00596984 (5859716)
00389E Uncompressed Size     00596984 (5859716)
0038A2 Filename Length       000B (11)
0038A4 Extra Length          0003 (3)
0038A6 Filename              'classes.dex'
0038B1 Null Padding in Extra
#
# INFO: Offset 0x38B1: 0x3 (3) Null Padding Bytes in Extra Field
#
0038B4 PAYLOAD

59A238 LOCAL HEADER #15      04034B50 (67324752)
59A23C Extract Zip Spec      00 (0) '0.0'
59A23D Extract OS            00 (0) 'MS-DOS'
59A23E General Purpose Flag  0000 (0)
59A240 Compression Method    0000 (0) 'Stored'
59A242 Modification Time     02210821 (35719201) 'Thu Jan  1 01:01:02 1981'
59A246 CRC                   B6B42CA3 (3065261219)
59A24A Compressed Size       00060C0C (396300)
59A24E Uncompressed Size     00060C0C (396300)
59A252 Filename Length       000C (12)
59A254 Extra Length          0002 (2)
59A256 Filename              'classes2.dex'
59A262 Null Padding in Extra
#
# INFO: Offset 0x59A262: 0x2 (2) Null Padding Bytes in Extra Field
#
59A264 PAYLOAD
```

LalitMaganti: Added a comment. Thanks for the investigation!

PR #2597, opened 2025-08-18
feat: allow multiple tabs to use the HTTP RPC accelerator + file sharing

WangJialin547:
Added a feature that solves the issue that at most one tab can be using the http + rpc accelerator. Also added file sharing feature that allows multiple different traces to be opened across multiple tabs and devices.
Design document of this change can be found here: 

https://docs.google.com/document/d/16ylk61fJBYUsW6SEbCLTnGeYAqpFhptJmmyQ0eEElcw/edit?usp=sharing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ GRTAGS
 GTAGS
 perf.data*
 TAGS
+/storage*
--- a/README_SHARED_PERFETTO.md
+++ b/README_SHARED_PERFETTO.md
@@ -1,0 +1,44 @@
+## Shared Perfetto Web UI
+### Build
+- Update dependencies
+```
+tools/install-build-deps
+tools/install-build-deps --ui
+```
+- build ui
+```
+ui/build
+ui/run-dev-server
+```
+
+- build processor
+```
+tools/gn args out/ubuntu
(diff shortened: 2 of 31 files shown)


LalitMaganti:
Thanks for the contribution but unfortunately there are a number of problems with this change:
1) it's targeting a release branch of Perfetto instead of main. We never accept contributions on release branches so this is a non starter from the beginning 
2) the number of code changes here is just huge. Compared to what I think this feature needs it's totally unreasonable. 
3) it seems like there are a bunch of unrelated features thrown in here instead of being a single focused patch
4) on recent releases of the Perfetto UI, the native accelerator has become much much less important as the UI is natively able to use much more memory so the need for multiple parallel accelerators has also decreased a lot 

Mainly because of 1) but also because of the other reasons, I will be closing this change. If you're interested in contributing this, please open a focused, minimal PR targeting only what you need to change instead of making so many changes. Also consider 4) carefully whether you still need this.

LalitMaganti:
Also your contribution to allow accessing an arbitrary file on shell over RPC is unfortunately not landable as it's a security hole as it allows reading arbitrary files on the disk which is a big nono. 

At the very least, there needs to be some protection on the shell side about the directories which are "allowlisted" to be accessible over RPC.

PR #2598, opened 2025-08-18
debian: Update packaging scripts and configuration

sashwinbalaji:
Update Perfetto Debian packaging.

This change:
1) Removes the `compat` file in favor of `debhelper-compat` in `debian/control`
2) Adds the `tracebox` binary to the installation
3) Filters out CFLAG `implicit-function-declaration` which is not valid for C++
4) Removes the unbuilt `traced-perf.service` systemd unit.

Test:
- Setup new perfetto repo
- Build: sudo apt-get build-dep . && dpkg-buildpackage -us -uc -b
- Install: sudo apt install ../perfetto_15.0-1_amd64.deb
- Capture trace using perfetto and tracebox
   - Tracebox: sudo tracebox -t 10s -o tracebox.pftrace sched freq
   - Perfetto:
      - sudo systemctl start traced.service
      - sudo systemctl start traced-probes.service
      - sudo perfetto -o perfetto.pftrace -t 10s -b32m sched freq idle (or add user to traced-consumer and run as non-root)

### DIFF
--- a/protos/third_party/simpleperf/record_file.proto
+++ b/protos/third_party/simpleperf/record_file.proto
@@ -59,6 +59,8 @@ message FileFeature {
   repeated Symbol symbol = 4;
 
   message DexFile {
+    // "dex_file_offset" is the offset of the dex /within/ its containing file
+    // (since they're ususally stored with zero compression inside apks/jars).
     repeated uint64 dex_file_offset = 1;
   }
   message ElfFile {
--- a/src/trace_processor/importers/common/symbol_tracker.h
+++ b/src/trace_processor/importers/common/symbol_tracker.h
@@ -32,7 +32,7 @@ class SymbolTracker {
  public:
   // DSO = Dynamic Shared Object.
   struct Dso {
-    uint64_t load_bias;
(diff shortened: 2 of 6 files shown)


primiano:
Exciting to see movement here o/

Out of curiosity how come the removal of the traced_perf svc?

sashwinbalaji:
> Exciting to see movement here o/
> 
> Out of curiosity how come the removal of the traced_perf svc?

Because we are not actually building traced_perf binary. So this service file was for a program that doesn't exist. We only build trace, traced_probes, perfetto and now tracebox. I think this clean up just got missed from aosp/1681266

rsavitski:
> > Exciting to see movement here o/
> > Out of curiosity how come the removal of the traced_perf svc?
> 
> Because we are not actually building traced_perf binary. So this service file was for a program that doesn't exist. We only build trace, traced_probes, perfetto and now tracebox. I think this clean up just got missed from aosp/1681266

yeah, we don't officially support heapprofd/traced_perf in standalone builds since they currently depend on a pinned version of libunwindstack.

PR #2601, opened 2025-08-18
perf proto: Exclusive counting for raw events

auselen:
Using RawEvents allow trace to separate if events are caused by kernel or application.

### DIFF
--- a/protos/perfetto/common/perf_events.proto
+++ b/protos/perfetto/common/perf_events.proto
@@ -158,6 +158,9 @@ message PerfEvents {
     optional uint64 config = 2;
     optional uint64 config1 = 3;
     optional uint64 config2 = 4;
+    optional bool exclude_user = 5;
+    optional bool exclude_kernel = 6;
+    optional bool exclude_hv = 7;
   }
 
   // Subset of clocks that is supported by perf timestamping.
--- a/protos/perfetto/config/perfetto_config.proto
+++ b/protos/perfetto/config/perfetto_config.proto
@@ -2033,6 +2033,9 @@ message PerfEvents {
     optional uint64 config = 2;
     optional uint64 config1 = 3;
     optional uint64 config2 = 4;
+    optional bool exclude_user = 5;
+    optional bool exclude_kernel = 6;
+    optional bool exclude_hv = 7;
   }
 
   // Subset of clocks that is supported by perf timestamping.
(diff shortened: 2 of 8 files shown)


LalitMaganti:
@auselen: was closing this PR intentional?

auselen:
I'm terribly sorry. It wasn't intentional, just my lack of knowledge about working with github workflows.

LalitMaganti:
@rsavitski can you ptal?

auselen:
Why this is useful?
- I'm investigating an app launch behavior; If I see different INSTRUCTIONS RETIRED on different platforms, I would like to know how much kernel is contributing to them.
- I have a test setup to run under different conditions, I would like to make sure core/userland of my test is doing exactly as I expect. Let's say a file write test, running on different environments, I should be able to see the overhead contributed.

rsavitski:
Terribly sorry for not getting back to you on this. I'll have a proper response posted by end of day.

Until then:
* Supporting the event modifiers ":u" and ":k" is a useful feature that we want to support (https://perfwiki.github.io/main/tutorial/#modifiers)
* perfetto config-wise, it is odd to allow modifiers only for raw events. It's perfectly applicable to standard counters as well (though not tracepoints, as those are kernel-only by definition).
* So it's better if this were a configuration option at a larger scope in `Timebase` or `PerfEventConfig`. And at that point it's more human friendly to expose it as an inclusion filter (such as perf's ":u" for "only userspace") instead of an exclusion (the syscall-level description, which would require exclude_kernel + exclude_hypervisor).
* So the question becomes where to put it in our config. I originally thought it could belong at the top-level, like the [`target_cpu` mask](https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/config/profiling/perf_event_config.proto;l=60;drc=714365290796a154d38534f29b99d6e045636bc3), but the kernel allows mixing events with different modifiers in one group (e.g. `perf record -a -e '{cycles:u, instructions:k}:S'`). So on one hand we could allow the same flexibility, or we could simplify and allow one scope for the entire group.
* I'm probably leaning towards exposing this per event (i.e. at the Timebase::Counter and Timebase::RawEvent level), since I can see a use for e.g. "cycles:u + cycles:k" in one group.

rsavitski:
Started on a version with a more general EventModifiers enum in the proto at https://github.com/google/perfetto/pull/3016. Am I overlooking any obvious shortcomings of that approach (besides uglier textproto configs)? Happy to finish off the patch if it fits your needs.

auselen:
@rsavitski, shall I close this one?

rsavitski:
> @rsavitski, shall I close this one?

Yes please. My patch is expected to be reviewed in a day or two.

PR #2603, opened 2025-08-18
Refactor SetProcessMetadata to depend on UniquePids

jahdiel-alvarez:
ProcessTracker::SetProcessMetadata is a complex function with multiple side effects. Simplify the API by accepting UniquePids instead of raw pid values. The handling of UniquePids simplifies the API since it requires the initial process creation to occur outside of the method call.

Bug: 425694654


jahdiel-alvarez:
Will split the method into two in next PR

PR #2608, opened 2025-08-18
tp: workaround weird exact_offset for the kernel text section in perf files

LalitMaganti:


src/trace_processor/importers/perf/perf_tracker.cc
```diff
@@ -144,6 +144,13 @@ void PerfTracker::CreateKernelMemoryMapping(int64_t trace_ts,
       params.memory_range.size() == std::numeric_limits<uint64_t>::max()) {
     return;
   }
+  // Linux perf synthesizes a kernel mapping which maps the kernel image close
+  // to uint64::max() but then *also* claims that the symbols are the same
+  // offset into the kernel ELF file. This does not make any sense. Just zero
+  // out the exact_offset as this is what is happening in practice.
+  if (params.exact_offset == params.memory_range.start()) {
+    params.exact_offset = 0;
+  }
```
rsavitski: We can be way more specific here. How about (with possible notes that we are not handling guest kernels and haven't seen cases where start != exact_offset):
```
  // Linux perf synthesises special MMAP/MMAP2 records for the kernel image. In
  // particular, the KASLR address of _text is stored in the `pgoff` field. This
  // needs special treatment since the kernel ELF is not in fact 0xffffff... in
  // size. See:
  // * https://elixir.bootlin.com/linux/v6.16/source/tools/perf/util/synthetic-events.c#L1156
  // * https://lore.kernel.org/lkml/20201214105457.543111-1-jolsa@kernel.org
  if (params.name == "[kernel.kallsyms]") {
    params.exact_offset = 0;
  }
```

LalitMaganti: Done thanks!

PR #2611, opened 2025-08-18
tp: add support for symbolizing kernel frames in perf.data files

LalitMaganti:


src/profiling/symbolizer/local_symbolizer.cc
```diff
@@ -653,6 +653,43 @@ std::optional<FoundBinary> FindBinaryInRoot(const std::string& root_str,
...
+  }
+  path = base::StackString<1024>("/usr/lib/debug/lib/modules/%s/vmlinux",
+                                 os_release.c_str());
+  binary = FindKernelBinaryAtPath(path.c_str());
+  if (binary) {
+    return binary;
+  }
+  path = base::StackString<1024>("/usr/lib/debug/boot/vmlinux-%s.debug",
+                                 os_release.c_str());
+  binary = FindKernelBinaryAtPath(path.c_str());
```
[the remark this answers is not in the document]
LalitMaganti: Doesn't wokr because printf strings have to be compile time constants and the compiler cannot tell that the `format` string is a constant.

src/profiling/symbolizer/local_symbolizer.cc
```diff
@@ -653,6 +653,43 @@ std::optional<FoundBinary> FindBinaryInRoot(const std::string& root_str,
   return std::nullopt;
 }
 
+std::optional<FoundBinary> FindKernelBinaryAtPath(const char* path) {
+  return IsCorrectFile(path, std::nullopt);
+}
+
+std::optional<FoundBinary> FindKernelBinary(const std::string& os_release) {
```
rsavitski, line 654: Can we write this more compactly? This looks like it should be an array + loop.

Also, this should really point at the source/inspiration for the paths (e.g. a linux to tools/perf/ sources in elixir or github). 

I believe this list is incomplete even relative to linux.perf (there are more fallbacks like /proc/kallsyms and also the use of ~/.debug), so at least documenting where it came from is important.

src/profiling/symbolizer/local_symbolizer.cc
```diff
@@ -801,21 +838,42 @@ std::vector<SymbolizedFrame> LLVMSymbolizerProcess::Symbolize(
...
+    // because llvm-symbolizer expects that we provide absolute addresses unlike
+    // all other files where it expects relative addresses.
+    addr_correction = binary->vaddr;
+  } else if (binary->poffset > 0 && binary_load_bias > load_bias) {
     // On Android 10, there was a bug in libunwindstack that would incorrectly
     // calculate the load_bias, and thus the relative PC. This would end up in
     // frames that made no sense. We can fix this up after the fact if we
     // detect this situation.
+    //
+    // Note that the `binary->poffset > 0` check above accounts for perf.data
```
rsavitski, line 863: Do we need to document the "symbol elf with text segment's poffset zeroed out" surprise? I don't know where the comment best belongs, but it feels like an important finding to write down.

src/profiling/symbolizer/local_symbolizer.cc
LalitMaganti: So the main reason I did it this way is because base::StackString expects the string to be a constant. The compiler cannot see through the fact that the string is a constant in a loop unfortunately.

> Also, this should really point at the source/inspiration for the paths (e.g. a linux to tools/perf/ sources in elixir or github).

Documenting this makes sense, will add a comment.

src/profiling/symbolizer/local_symbolizer.cc
LalitMaganti: Yup will add that in.

src/profiling/symbolizer/local_symbolizer.cc
rsavitski: hmm, took a stab, not sure if better:
```
  auto* rel = os_release.c_str();
  using SS = base::StackString<512>;
  auto path = SS("/boot/vmlinux-%s", rel);
  if (auto binary = FindKernelBinaryAtPath(path.c_str()); binary)
    return binary;

  path = SS("/usr/lib/debug/boot/vmlinux-%s", rel);
  if (auto binary = FindKernelBinaryAtPath(path.c_str()); binary)
    return binary;

  path = SS("/lib/modules/%s/build/vmlinux", rel);
  if (auto binary = FindKernelBinaryAtPath(path.c_str()); binary)
    return binary;

  path = SS("/usr/lib/debug/lib/modules/%s/vmlinux", rel);
  if (auto binary = FindKernelBinaryAtPath(path.c_str()); binary)
    return binary;

  path = SS("/usr/lib/debug/boot/vmlinux-%s.debug", rel);
  if (auto binary = FindKernelBinaryAtPath(path.c_str()); binary)
    return binary;

  return std::nullopt;
```

LalitMaganti:
Thanks for the review! Resolved all comments.

rsavitski (approved):

PR #2613, opened 2025-08-19
ui: move android plugins from dev.perfetto -> com.android

LalitMaganti:


stevegolton (approved):
This is great + much needed, thanks! Are any of the plugins you've renamed not enabled by default - if so folks will have to re-enable the after the rename so we should make this known.

LalitMaganti:
> This is great + much needed, thanks! Are any of the plugins you've renamed not enabled by default - if so folks will have to re-enable the after the rename so we should make this known.

Yes a lot of them. What's the best way to disseminate the information that a reenable will be needed?

PR #2617, opened 2025-08-19
RTMutex: remove RTFutex, use android flag for MaybeRtMutex

primiano:
- Get rid of RtFutex: benchmarks in PR #1648 show that the gain
  of directly using a futex are negligible and don't justify the
  extra complexity.
- Now that we have build time RO flags, use them rather than a
  runtime flag.
- Use dlsym on older versions of android


include/perfetto/ext/base/flags.h
```diff
@@ -35,7 +35,8 @@ namespace perfetto::base::flags {
 #define PERFETTO_READ_ONLY_FLAGS(X)                                    \
   X(test_read_only_flag, NonAndroidPlatformDefault_FALSE)              \
   X(use_murmur_hash_for_flat_hash_map, NonAndroidPlatformDefault_TRUE) \
-  X(ftrace_clear_offline_cpus_only, NonAndroidPlatformDefault_TRUE)
+  X(ftrace_clear_offline_cpus_only, NonAndroidPlatformDefault_TRUE)    \
+  X(use_rt_mutex, NonAndroidPlatformDefault_FALSE)
```
LalitMaganti, line 39: Shouldn't this be true by default on non-Android?

ddiproietto: Let's leave it as false. When we finish testing this on Android we can roll it out everywhere else

ddiproietto (approved):

include/perfetto/ext/base/flags.h
primiano: I'm setting it to true in a different CL, so if it causes problems we can revert indepenently

LalitMaganti (approved):

PR #2618, opened 2025-08-19
tp: fix clashing of track group ids across traces

LalitMaganti:
We should ensure they are put in differnt sets to ensure we don't end up
with multiple traes both assigning group 0


src/trace_processor/types/trace_processor_context.h
```diff
@@ -54,6 +53,7 @@ class StackProfileTracker;
 class SymbolTracker;
 class TraceFileTracker;
 class TraceReaderRegistry;
+struct TrackCompressorGroupIdxState;
```
sashwinbalaji: nit: move it to line 62 to maintain format of class declarations followed by structs.

sashwinbalaji (approved):

PR #2624, opened 2025-08-19
Add url params option for the record android trace script

pengj:
when we use the tools/record_android_trace, it will be good we can pass the url params when open the trace in the browser. This is part of change for this feature rquest ([2400](https://github.com/google/perfetto/issues/2400)) . 

The default value are the same as before, so there is no logic changes.

**Test:**
run the script: ./gen_amalgamated_python_tools to generate the record_android_trace and tested with an Android device trace, all works as expected.

When pass the url-params with cmd: `tools/record_android_trace -c perfetto.config --url-params test1 test2 test3`, the url-params is append to the url correctly in the browser.

### DIFF
--- a/python/tools/record_android_trace.py
+++ b/python/tools/record_android_trace.py
@@ -110,6 +110,9 @@ def setup_arguments():
   help = 'The web address used to open trace files'
   parser.add_argument('--origin', default='https://ui.perfetto.dev', help=help)
 
+  help = 'The URL Params to open the trace file'
+  parser.add_argument('--url-params', default=None, help=help)
+
   help = 'Force the use of the sideloaded binaries rather than system daemons'
   parser.add_argument('--sideload', action='store_true', help=help)
 
@@ -420,7 +423,7 @@ def start_trace(args, print_log=True):
       prt('\n')
       prt('Opening the trace (%s) in the browser' % host_file)
     open_browser = not args.no_open_browser
-    open_trace_in_browser(host_file, open_browser, args.origin)
+    open_trace_in_browser(host_file, open_browser, args.origin, args.url_params)
 
   return host_file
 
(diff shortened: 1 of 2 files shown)


python/tools/record_android_trace.py
```diff
@@ -110,6 +110,9 @@ def setup_arguments():
   help = 'The web address used to open trace files'
   parser.add_argument('--origin', default='https://ui.perfetto.dev', help=help)
 
+  help = 'The URL Params to open the trace file'
```
LalitMaganti: Can we make this a list argument instead of an opaque string?

python/tools/record_android_trace.py
```diff
@@ -464,6 +467,8 @@ def open_trace_in_browser(path, open_browser, origin):
   socketserver.TCPServer.allow_reuse_address = True
   with socketserver.TCPServer(('127.0.0.1', PORT), HttpHandler) as httpd:
     address = f'{origin}/#!/?url=http://127.0.0.1:{PORT}/{fname}&referrer=record_android_trace'
+    if url_params:
```
LalitMaganti, line 470: Once this is a list, the URL parameters should be concatenated together with &

python/tools/record_android_trace.py
pengj: that's my first thinking to put as a list, but we didn't do any validation or prase for the params just concatenate them again within the script, why not just let people pass the url paramters as a string here? or I missed anything here?

LalitMaganti: Well mainly because it's much intuitive to folks to pass as a list as that's what is normal on a command line. The fact that it gets passed as-is is an implementation detail and something I could imagine changing in the future. Whenever you have an API like this, it's best to give yourself maximum flexibility. 

Also it's far easier to have separate arguments for anyone scripting on top of record_android_trace without them needing to handle correct URL encoding

pengj: Good point, will update to a list

LalitMaganti (approved):

stevegolton (approved):

PR #2630, opened 2025-08-19
Move process to parent thread associations to new method

jahdiel-alvarez:
ProcessTracker::StartNewProcess assumes a Linux-based system, where PIDs and TIDs can be used interchangeably. In order to generalize ProcessTracker's implementation, StartNewProcess now accepts an optional upid for the parent process and the new AssociateProcessToParentUtid method is provided specifically for Ftrace to add any pending associations between the new process and its parent.

Bug: 439687339

### DIFF
--- a/src/trace_processor/importers/common/process_tracker.h
+++ b/src/trace_processor/importers/common/process_tracker.h
@@ -113,10 +113,15 @@ class ProcessTracker {
   // Called when a task_newtask without the CLONE_THREAD flag is observed.
   // This force the tracker to start both a new UTID and a new UPID.
   UniquePid StartNewProcess(std::optional<int64_t> timestamp,
-                            std::optional<int64_t> parent_tid,
+                            std::optional<UniquePid> parent_upid,
                             int64_t pid,
                             StringId main_thread_name,
                             ThreadNamePriority priority);
+
+  // Associates a process with its parent thread. Used when the tid that
+  // created the process is known, but not the parent process. Exclusively,
+  // used by Ftrace on new task events where only the parent tid is provided.
+  void AssociateProcessToParentUtid(UniquePid upid, UniqueTid parent_utid);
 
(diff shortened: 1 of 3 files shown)


src/trace_processor/importers/common/process_tracker.h
```diff
@@ -113,11 +113,16 @@ class ProcessTracker {
   // Called when a task_newtask without the CLONE_THREAD flag is observed.
   // This force the tracker to start both a new UTID and a new UPID.
   UniquePid StartNewProcess(std::optional<int64_t> timestamp,
-                            std::optional<int64_t> parent_tid,
+                            std::optional<UniquePid> parent_upid,
                             int64_t pid,
                             StringId main_thread_name,
                             ThreadNamePriority priority);
 
+  // Associates a process with its parent thread. Used when the tid that
```
LalitMaganti, line 121: I would rephrase both the funciton and comment to be more explicit that this funciton should be used when a thread A "spawns/forks" a thread B .

jahdiel-alvarez: We are actually trying to set a pending parent association when thread A spawns process B, but we might not have figured out the process of thread A. So the association is actually between a child **process** and a parent **thread**.

LalitMaganti: OK so please comment that. I would make very clear this is an advanced function which should not be used 99% of the time, right now with its name I could imagine mistaken uses of this.

jahdiel-alvarez: Any method name suggestions?

LalitMaganti: I'd be very explicit. Maybe something like: AddAssociationBetweenCreatedProcessAndParentThread

jahdiel-alvarez: Updated the method name

PR #2633, opened 2025-08-19
Make process_type and thread_type into ints

JoeNotCharlesGoogle:
This removes the ProcessType enum from chrome_process_descriptor.proto and, for symmetry, moves the ThreadType enum from chrome_thread_descriptor.proto to chrome_enums.proto. The enum values are now controlled from Chrome.

The process_type and thread_type field are changed to a int32 instead of naming the enum type. This is wire-compatible according to https://protobuf.dev/programming-guides/proto2/#wire-safe-changes. Specifying the enum type would require importing protos/third_party/chromium/chrome_enums.proto from inside protos/perfetto/trace/track_event, which is a dependency inversion. It would cause duplicate dependency when building Perfetto in Chrome, which would import the same enum from both base/protos/chrome_enums.proto and (indirectly) from third_party/perfetto/protos/third_party/chromium/chrome_enums.proto. 

Unfortunately all test expectations in .textproto format now need to use numeric process and thread types, since the test output doesn't treat the fields as enums.


LalitMaganti (approved):

petrcermak:
I would personally prefer the extension-based approach because of your API design argument. It seems to me that the trade-off is short-term pain (more complicated merge story) vs long-term pain (less clear API and API contract). I think the short-term pain will pay off in the long-term.

petrcermak (approved):

PR #2643, opened 2025-08-20
tp: etm: adds last seen timestamp and cumulative cycles

OmarElfouly:
This PR adds two new columns to __intrinsic_etm_decode_chunk.

- last_seen_timestamp - as the name implies this gives an approximation of the current timestamp for packets that do not have one using the last seen timestamp while accounting for SYNC_MARKERS
- cumulative_cycles - this column keeps track of the reported number of cycles starting from 0 at the start of a chunk.


test/trace_processor/diff_tests/parser/etm/decode_chunk.out
```diff
@@ -1,198 +1,198 @@
...
-223,189,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487550716,523487550728,13
-223,190,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487550728,523487550756,13
-229,191,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487548832,523487548884,13
-229,192,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487920848,523487920864,13
-235,193,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487677904,523487677916,13
-235,194,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487677916,523487677928,13
-235,195,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487677928,523487677936,13
-236,196,"INSTR_RANGE","[NULL]","[NULL]",0,315,"AARCH64",523487921680,523487921696,13
+"chunk_index","element_index","element_type","timestamp","cycle_count","last_seen_timestamp","cumulative_cycles","exception_level","context_id","isa","start_address","end_address","mapping_id"
+12,0,"NO_SYNC","[NULL]","[NULL]","[NULL]","[NULL]","[NULL]","[NULL]","UNKNOWN",-1,-1,"[NULL]"
```
rasikanavarange, line 2: It seems there are no timestamps and cycle counts in this trace. We should add a test that has those columns populates as well so we can ensure it's working correctly

src/trace_processor/perfetto_sql/intrinsics/operators/etm_decode_trace_vtable.cc
```diff
@@ -119,14 +123,94 @@ class EtmDecodeChunkVtable::Cursor
...
+      if (cursor_.Eof()) {
+        if (waiting_for_timestamp_ && !rows_waiting_for_timestamp_.empty()) {
+          flushing_buffer_ = true;
+          buffer_idx_ = 0;
+          return base::OkStatus();
+        }
+        return base::OkStatus();
+      }
+
+      if (waiting_for_timestamp_) {
```
rasikanavarange, line 149: Is it possible to have a case where you have two sync markers then two timestamps? And if so, does this implementation account for that?
```
SYNC_MARKER_A
...
SYNC_MARKER_B
...
TIMESTAMP_A
...
TIMESTAMP_B
...
```

rasikanavarange (commented):
Looks good, just might need a couple updates in a follow-up PR

src/trace_processor/perfetto_sql/intrinsics/operators/etm_decode_trace_vtable.cc
OmarElfouly: Yes there are two edge cases I am aware of ATM:

- Sync and no timestamp - Handled with a limit to buffer size (i believe 30 rows is a good number) after which assume last timestamp for these rows
- Sync and then Sync - I could not find a promise in the docs that this cannot happen. If i wanted accuracy i would have to wait for both timestamps and then assign them based on their timestamp value. But that seems very expensive because timestamps tend to have very similar timings so at the moment it will treat them as a single sync that occurred at the first sync packet. So it would be SYNC_MARKER_A with TIMESTAMP_A and so on.

I am also worried that it might happen that we get:
`SYNC_MARKER_A
...
SYNC_MARKER_B
...
TIMESTAMP_B
...
TIMESTAMP_A
...`

As I have not seen anything in the docs saying it cannot happen.

test/trace_processor/diff_tests/parser/etm/decode_chunk.out
OmarElfouly: Will do!

src/trace_processor/perfetto_sql/intrinsics/operators/etm_decode_trace_vtable.cc
rasikanavarange: > Sync and no timestamp

Is this when it is the last sync in the trace?


> Sync and then Sync

I think we have to assume this can happen based on the intended purpose of a sync_marker, which is to show the correct position of a later timestamp packet.


> I am also worried that it might happen that we get: SYNC_MARKER_A ... SYNC_MARKER_B ... TIMESTAMP_B ... TIMESTAMP_A ...

Sync markers are not associated with timestamps by labels, they are associated by order. A sync marker will refer to the next timestamp, so you can only have SM1, SM2, T1, T2.

OmarElfouly: > Is this when it is the last sync in the trace?

Yes or if data loss happens in just the right way I wanted to avoid an infinite seek for the next TIMESTAMP

> Sync markers are not associated with timestamps by labels, they are associated by order. A sync marker will refer to the next timestamp, so you can only have SM1, SM2, T1, T2.

Yeah my worry is that the order would be bad enough that an older timestamp arrives after a newer one. So SYNCA, SYNCB, TIMEB, TIMEA and then have to sort the TIME packets before assigning them to a SYNC and all the waiting rows.

Do you think that treating two consecutive syncs as one is better or is it worth extra processing for more precision?

rasikanavarange: > if data loss happens

Is perf_aux_truncated or some other data loss warning given when this happens?

> Yeah my worry is that the order would be bad enough

I don't think this is possible as it breaks a lot of guarantees of how the decoder works. The packets arrive as a stream so their order should be chronological. Sync markers are used for delays in timestamps, not because they are out of order. I don't see how it would be possible to finish creating a newer timestamp when the prior one is already delayed.

> Do you think that treating two consecutive syncs as one is better or is it worth extra processing for more precision?

Treating two syncs as one is definitely incorrect logic, but depending on how often it occurs it might not be a big deal - let's get a diff test with real timestamps and see.

Issue #2629, opened 2025-08-19
Handling independent power rail streams


cphlipot1:
## Context:
Currently power rails streams are typically divided into two kinds of packets:
- RailDescriptors
- EnergyData

See: https://perfetto.dev/docs/reference/trace-packet-proto#PowerRails

EnergyData is asociated with the RailDescriptors using a integer rail index, which allows mapping the energy data to a specific rail name.

## Problem

Currently the rail parser is implemented such that the rail index is assumed to be globally unique for the trace. this presents issues when two independent power rail streams are present in the trace that overlap in terms of their indices.

This currently results in undefined behavior at import time, which tends to manifest as a mixture of power railsempty tracks plus some other tracks that contain energy samples from both streams, resulting in nonsensical power data.

## proposed solution

instead of keying on a unique index to map to track ids, instead map unique pairs of the sequence_id and index to track_ids. The assumption here is that each stream would be coming for a different datasource, and thus have a different sequence id.

Namely looks like we would need modifications to both the setter and getter [here](https://github.com/google/perfetto/blob/9fb704869663f154104aca5ba4047610f0039370/src/trace_processor/importers/proto/android_probes_tracker.h#L59-L71
) and the pipe the sequence id to those functions

Is there any concern with the above approach, or other alternatives worth considering?

LalitMaganti:
I'm don't think overloading sequence_id here is a good idea: packet_sequence_id for Perfetto recorded traces directly maps to a thread 1:1. This means if for some reason, there is a e.g. a thread pool of 4 threads which is emitting power rail data, even if the data is coming from the same set of logical rails, it will be split amongst 4 different set of power rail tracks.

Instead, I suggest adding the notion of "session uuid" directly to the power rail protos. Then going forward, whoever is writing these packets can specify a uuid and we can use that as an additional key along with the rail index. For backcompat, if the uuid is not present, it will map to zero and things should work just as it does today. Does that make sense?

LalitMaganti:
While we do this, we should take the opportunity to:
1) Improve the documentation on the power rail protos
2) make sure that we come up with a decent random UUID for the Perfetto native recording data source.

cphlipot1:
>  I suggest adding the notion of "session uuid" directly to the power rail protos. 

Sounds good. i should be able to contribute these changes.

> Improve the documentation on the power rail protos

Is this about improving documentation to accommodate the new functionality, or are there gaps in documentation for the existing functionality you are concerned with

[...]

PR #2645, opened 2025-08-20
 tp: introduce power rails session uuid

cphlipot1:
Make power rails indices scoped only to the given session uuid if sepcifed
to they don't clash with other data sources or syntheticly generated
packets via other methods being produced independently.

Bug: #2629


LalitMaganti:
I think we also need to change the recording side to actually emit this.

cphlipot1:
I was planning on breaking that into a separate PR. Did you want them together?

LalitMaganti:
> I was planning on breaking that into a separate PR. Did you want them together?

No worries, as long as there's a plan for it!

src/trace_processor/importers/proto/android_probes_module.cc
```diff
@@ -222,7 +222,7 @@ ModuleResult AndroidProbesModule::TokenizePacket(
           kPowerBlueprint, tracks::Dimensions(desc.rail_name()),
           tracks::DynamicName(id), args_fn);
     }
-    tracker_->SetPowerRailTrack(desc.index(), track);
```
LalitMaganti, line 225: I think we'll also need to change the blueprint dimensions to include a session id otherwise we will break the expectation there is a unique track per set of dimensions.

test/trace_processor/diff_tests/parser/power/tests_power_rails.py
```diff
@@ -151,6 +151,93 @@ def test_power_rails_well_known_power_rails(self):
...
+          name,
+          ts,
+          value
+        FROM counters
+        WHERE name GLOB "power.*"
+        ORDER BY name, ts;
+        """,
+        out=Csv("""
+        "name","ts","value"
+        "power.CPU_RAIL_uws",1000000,500.000000
```
LalitMaganti, line 235: I think you will find that the session uuid will always be zero here because of the comment above (i.e. there would only be 2 tracks here instead of the three there should be!)

LalitMaganti (approved):

src/trace_processor/importers/proto/android_probes_module.cc
cphlipot1: Sure i can add this, although i'm unclear what the exact impact of doing so is.

test/trace_processor/diff_tests/parser/power/tests_power_rails.py
cphlipot1: I'm not sure i underderstand the issue you are trying to flag. 

> I think you will find that the session uuid will always be zero

session_uuid seemed to map correctly for me when i added logs to PowerRailTrack setter and getter, i was getting non-zero values in this diff test. and i think the output of the diff test is validating that they actually map to different tracks unlike before. Where are you expecting this to be zero?

> there would only be 2 tracks here instead of the three there should be!

this diff test is passing, and i only see it create two power rail tracks in the input data: CPU_RAIL and GPU_RAIL. Where is the 3rd track you're expecting supposed to come from?

If you're expecting 3 here, what would would the diff test expected output be according to your expectations?

src/trace_processor/importers/proto/android_probes_module.cc
LalitMaganti: Well if you don't do this, I think you'll end up back where you started: each different invocation of InternTrack will return the same trackid if the rail names overlap between the two sessions.

test/trace_processor/diff_tests/parser/power/tests_power_rails.py
LalitMaganti: See above: the point I was making was that you will get cross-talk between the two session uuids because InternTrack is only resolving based on the rail name and not the session uuid. So if two sessions emit the same rail name, they'll be folded onto a single track.

src/trace_processor/importers/proto/android_probes_module.cc
cphlipot1: Actually i read though the blueprint code again. if i understand correctly the failure mode happens if we have two rails from two different session uuids that share the same name, because the rail name is the only dimension currently used. Is that correct?

I'm still not clear how that would give us a 3rd track with the diff test trace as you mention below that.

test/trace_processor/diff_tests/parser/power/tests_power_rails.py
cphlipot1: I read your explanation and it makes sense. I've updated the blueprint to include the new dimension.

I'm assuming the comment about expecting a 3rd track for this diff test given the current input trace was a misunderstanding. If not please let me know.

LalitMaganti: Never mind, I did not read the test carefully. You wouldn't get three tracks here.

test/trace_processor/diff_tests/parser/power/tests_power_rails.py
```diff
@@ -151,6 +151,93 @@ def test_power_rails_well_known_power_rails(self):
...
+        packet {
+          power_rails {
+            session_uuid: 200
+            rail_descriptor {
+              index: 0
+              rail_name: "GPU_RAIL"
+              subsys_name: "gpu"
+              sampling_rate: 2000
+            }
+          }
```
LalitMaganti, line 177: We should also test for a situation where two dfiferent sessions at the same rail index have the same rail name and verify there's no cross talk. We can use `extract_arg(dimension_arg_set_id, 'session_uuid')` in SQL for that.

src/trace_processor/importers/proto/android_probes_module.cc
LalitMaganti: > I'm still not clear how that would give us a 3rd track with the diff test trace as you mention below that.

Sorry I was mistaken: see below.

PR #2647, opened 2025-08-20
Parse MEMINFO ETW events in TraceProcessor

seanptmaher:
This change parses MemInfo ETW events in TraceProcessor.

This was verified using `trace_processor_shell`.

### DIFF
--- a/src/trace_processor/importers/etw/etw_parser.cc
+++ b/src/trace_processor/importers/etw/etw_parser.cc
@@ -27,6 +27,8 @@
 #include "src/trace_processor/importers/common/process_tracker.h"
 #include "src/trace_processor/importers/common/sched_event_tracker.h"
 #include "src/trace_processor/importers/common/thread_state_tracker.h"
+#include "src/trace_processor/importers/common/track_tracker.h"
+#include "src/trace_processor/importers/common/tracks.h"
 #include "src/trace_processor/storage/stats.h"
 #include "src/trace_processor/storage/trace_storage.h"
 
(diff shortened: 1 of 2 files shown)


src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -38,10 +41,84 @@ namespace perfetto {
 namespace trace_processor {
 
 namespace {
-
 using protozero::ConstBytes;
 
+static constexpr auto kZeroPageCountBlueprint =
```
LalitMaganti: Given these are not used across functions, can we just put these insnide the file?

src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -38,10 +41,84 @@ namespace perfetto {
...
+    tracks::CounterBlueprint("zero_page_count",
```
LalitMaganti: I would maybe suggest making it a single "etw_meminfo" blueprint and then adding the counter type as a dimension. I think this is more scalable and would better align with how you would want to query with this.

seanptmaher: That's much better, thank you.

src/trace_processor/importers/etw/etw_parser.cc
seanptmaher: done.

LalitMaganti (approved):

LalitMaganti:
I suggest some diff tests to ensure this doesn't break but up to you.

etiennep-chromium (approved):

seanptmaher:
None of ETW seems to have diff-tests. Which isn't ideal.

I think it would be worth adding them, but I don't think that needs to be done here. I'll put it on the stack

PR #2648, opened 2025-08-20
protos: add session_uuid for power rails

cphlipot1:
Add the session uuid for power rails to give the ability for producers
to uniquely mark their packets in order to differentiate them from
the power rail packets produced by others.

Bug: #2629


LalitMaganti (approved):

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -15214,6 +15214,12 @@ message PowerRails {
   }
 
   repeated EnergyData energy_data = 2;
+
+  // A unique session id that can be used to match energy data to sets of
+  // descriptors. The indices used by rail descriptors and energy data packets
+  // are meant to be unique to a given session uuid. When multiple data sources
+  // are running in parallel, each data source should use a unique id.
+  optional uint32 session_uuid = 3;
```
LalitMaganti: Thought about this and yeah I think we should just make this a uint64. There's no downside to making it a u64 in trace size terms as varint encoding takes care of it and it allows it to be a real uuid (i..e random 64-bit number).

LalitMaganti (changes requested):

protos/perfetto/trace/perfetto_trace.proto
cphlipot1: My concerns here was trace size. but as you mentioned due to protobuf encoding this shouldn't make a difference if the number was small.

However i was planning to use the full 32-bits and if i use the same approach of populating all bits in the session_uuid, then this would make difference to trace size. Did you have an idea for UUID generation in the datasource that would only use the lower bits? 

My current UUID generation Impl is here for reference, although i'm still trying to figure out how to test that PR as sideloading doesn't work: https://github.com/google/perfetto/pull/2650/files#diff-fb4cc606269bb0a2c7748ee931bb46f6a650692433b0b4b1d7756cf5ed6b8109R183

I'm fine bumpping to 64-bits though if you'd prefer.

LalitMaganti: Personally given that there is already repeated EnergyData inside the PowerRails proto, I personally feel it's fine to pay an extra 4 bytes.

Though if we usually will use the full bits, we should use fixed64 instead of uint64 as it will be more efficient (8 vs 10 bytes).

cphlipot1: updated to 64-bits. avoiding fixed64 since even if we use the full 64 in the probes data source, which i'm still not sure we want to do, this would leave the option open for people producing synthetic traces to use low bits only if they prefer to optimize for trace size.

LalitMaganti: Sounds good

PR #2651, opened 2025-08-20
tp: Refactored process APIs to not associate main thread by default

jahdiel-alvarez:
Splitted the GetOrCreateProcess API into two variants:
- GetOrCreateProcess : Creates a main thread
- GetOrCreateProcessWithoutMainThread : Doesn't create a main thread

Also, did the same with the StartNewProcess API. These changes are the beginning of the decoupling PID from the main thread's TID in ProcessTracker and now provide a way for process creation without creating default main thread. This should simplify the integration of other non-Linux OSes into Perfetto.

Bug: 439687339

### DIFF
--- a/include/perfetto/ext/base/flags.h
+++ b/include/perfetto/ext/base/flags.h
@@ -36,7 +36,7 @@ namespace perfetto::base::flags {
   X(test_read_only_flag, NonAndroidPlatformDefault_FALSE)              \
   X(use_murmur_hash_for_flat_hash_map, NonAndroidPlatformDefault_TRUE) \
   X(ftrace_clear_offline_cpus_only, NonAndroidPlatformDefault_TRUE)    \
-  X(use_rt_mutex, NonAndroidPlatformDefault_FALSE)
+  X(use_rt_mutex, NonAndroidPlatformDefault_TRUE)
 
 ////////////////////////////////////////////////////////////////////////////////
 //                                                                            //


jahdiel-alvarez:
This is an intermediate commit, which requires updates to UpdateThread to be landed afterwards. Currently, interactions between the new APIs and UpdateThread might lead to unexpected behavior. This is mitigated in the meantime because currently we are always creating a main thread upon process creation. The refactoring of UpdateThread should be uploaded soon.

The PR was sent as-is to keep the code (line count) at a reasonable size for better review.

src/trace_processor/importers/common/process_tracker.h
```diff
@@ -110,14 +110,22 @@ class ProcessTracker {
...
                             ThreadNamePriority priority);
 
+  // Same as StartNewProcess, but also creates a main thread associated with
+  // the process. Called when a task_newtask without the CLONE_THREAD flag is
+  // observed.
+  UniquePid StartNewProcessWithMainThread(std::optional<int64_t> timestamp,
+                                          std::optional<UniquePid> parent_upid,
+                                          int64_t pid,
+                                          StringId process_name,
+                                          ThreadNamePriority priority);
```
LalitMaganti: Thinking a bit more: can we actually invert these (so make StartNewProcessWithMainThread -> StartNewProcess and StartNewProcess -> StartNewProcessWithoutMainThread).

We get a lot of contributions from folks and it's possible that they use the wrong function because it's shorter and I expect StartNewProcessWithMainThread/GetOrCreateProcessWithMainThread to be used more commonly.

Issue #2551, opened 2025-08-13
Implement link sharing for 3P


sifangli-labs:
Hi, here is some background of my request. In the recorded trace I want to highlight a time where the interesting thing happened. I can do it via UI, but I want it to be automatic and when the user opens it, it is already there (Let's assume I already know the 'proper' timestamp to use in this task). 
In Chrome trace I can easily add a global 'instant event' (ph='i') to an existing trace and draw a vertical line to highlight the time. How can I do it to an already recorded perfetto trace? (via a json merge would be nice so I do not need to parse the Perfetto trace myself).
I see the same Chrome event is supported via Perfetto's TrackEvent, so I hope it is not too much to ask for the request.  
Thx.

LalitMaganti:
For a long time, Googlers have had the ability to share traces with the state of the state preserved. Here's an example: https://ui.perfetto.dev/#!/?s=49a3cf8faf432f3fc5dc91e25b248ba5cb1f4c45

But this was not open to external users due to privacy and security reasons.

But @stevegolton is going to be looking into opening this up to external users. Once we have this, should just be a matter of marking a point in the trace, clicking "Share" and everything else will just work.

sifangli-labs:
Sharing is an awesome function, but it does not really replace the need of the asked function since the effort of marking on the trace for the first time could also be saved if the time is already known (in many cases, that is true) before the trace is opened.

LalitMaganti:
So let me break down the options:
1) if you want to have a single "top level track" that you want visible at the top, you can wait for https://github.com/google/perfetto/issues/1018 and then include a synthetic trace with the timestamp that opens at the top of the trace (see note below)
2) If you wanted to actually have an automatic "note" in the UI based on the presence of some event (i.e. similar to note in the trace Is hared above), I think we'd be open to a contribution to a UI plugin which would allow users to write an SQL query and then create notes automatically for every timestamp where that SQL query matches.
3) You can always fork the UI and have your own version which integrates a plugin to do this. Would solve this problem as well without the user needing to type the SQL query e.g. in a setting. 

Note on 1): Just to be clear, global instants with TrackEvent but they work very differently to the chrome://tracing intentionally (global instants there were a mess because everyone decided their event was "important" and spammed it making traces unreadable.).

This is how instants work with Perfetto:

<img width="2330" height="26" alt="Image" src="https://github.com/user-attachments/assets/25b3a418-01ad-4df7-b0cd-d3cb8c11c590" />

[...]

Issue #1342, opened 2025-04-22
FR: allow including UI macros/startup commands in trace files


anakryiko:
It would be amazing to be able to produce Protobuf-based Perfetto trace, in which some of the tracks are actually derived using pre-defined SQL query based on data in another tracks.

E.g., having custom per-thread tracks where track events are tagged (using debug annotations) with associated CPU on which some event happened, it would be nice to have pre-defined per-CPU tracks that combine events across many threads/processes (or whatever custom track scope is) into a "per-CPU" tracks based on debug annotation with CPU number.

Of course, per-CPU is just a one straightforward use case, having full power of SQL at trace emitting stage would be amazingly powerful for producing all sorts of pre-canned views without bloating trace with duplicated data.

LalitMaganti:
Is this not just https://perfetto.dev/docs/analysis/debug-tracks?

anakryiko:
@LalitMaganti if I understand correctly, debug tracks can be only created from UI, manually, is that right? The feature request here is basically to be able to produce such debug tracks programmatically inside the trace file itself. But not through duplicating data, but rather deriving debug tracks with SQL, that will come together with protobuf trace file.

anakryiko:
@LalitMaganti I understand, thanks for entertaining the idea anyways!

Do you have some links for those added settings and "initrc" for the UI? I'd like to learn about them. Thanks!

LalitMaganti:
> Do you have some links for those added settings and "initrc" for the UI? 

Settings were added last week so are incredibly new. There's no link I can point you to beyond https://github.com/google/perfetto/commit/b1fc8d6adbb813b7f98b3eabcd6cf7485567a76b

initrc is not actually implemented (as it had a prerequisite on the settings page) but the idea is basically "a bunch of commands which run on startup and are user configurable" similar to a init.rc file in Linux/Android. The basic premise is that you could have a command which is "run all the queries and create debug tracks for the SQL stored in a given setting" and that would be executed on startup, setting up the UI in a user defined way without needing to write a formal plugin and/or bake anything into the file.

anakryiko:
@LalitMaganti oh, that's great! Yep, that initrc stuff is exactly what I'd like. Any ETAs for that?

stevegolton:
It's definitely on the list. I think we need to work though some unanswered questions first about the syntax, but it's definitely planned.

LalitMaganti:
I'm going to start looking into this in the coming week.

[...]

Issue #1018, opened 2025-02-18
Add support for merging traces


LalitMaganti:
It's become clear that we need some way to have multiple independent traces (collected on either the same device or different devices) visualized and analyzable on a single timeline.

Most of the work for this already exists in the trace processor but it is very hard for anyone who does not understand a lot of the internals of TP to make use of it.

This bug tracks the remaining implementation which involves some small work in TP and then some non-trivial work in the UI to expose this functionality.

LalitMaganti:
Will be looking to this actively in the coming month.

LalitMaganti:
Current state:
1) I'm now pretty happy with the UI part of things, we have a pretty initial draft for "clock synchronisation" between traces.
2) There is some work to be done in trace processor to take the result of what the UI builds for clock sync and actually make use of this while parsing. This will require defining the JSON schema we expect folks to use and making the UI write it as well
3) Work on multiple machines has not started yet but I remain convinced it's not too much work on top of what we already are doing for single machine merging.

sifangli-labs:
@LalitMaganti Thanks for woring on this impritant feature! If the user does not care about multiple machines and the UI function, is there a cmdline util that works to merge two pftraces from the same machine (captured around the same period) into a single pftrace file? 
If so any instructions? 
Thanks!

LalitMaganti:
No sorry you will have to wait until this is done. I'm not giving preview instructions, this is not yet complete.

From my previous comment
> There is some work to be done in trace processor to take the result of what the UI builds for clock sync and actually make use of this while parsing. This will require defining the JSON schema we expect folks to use and making the UI write it as well

LalitMaganti:
For anyone reading this bug: I now feel comfortable enough for people to start using this and reporting bugs. Note that it's still *very* experimental.

Here's how you do it:
1) Switch to the "autopush" channel of the UI: Click "Flags" on the sidebar, then "Release Channel" and switch to "Autopush"
2) Refresh the page
3) Click "Plugins" on the sidebar and enable "dev.perfetto.MultiTraceOpen" 
4) Refresh the page
5) You should then get a new sidebar option called "Open Multiple Trace Files" -> this will allow you to popup a dialog which let's you choose multiple traces and then merge them together on the timeline

There are some limitations on this:
1) The data should be collected on the same machine at the same time. Multiple-machines/same time and comparing different traces collected at different types on same/different machines is *not* supported, They are tracked by child bugs
[...]

PR #2654, opened 2025-08-21
ui: implement startup commands feature

LalitMaganti:
- Add startup commands feature that executes commands after trace load
- Create reusable JsonSettingsEditor widget for JSON-based settings
- Refactor macro settings to use the new widget and shared schemas
- Add command validation with improved error reporting grouped by macro
- Use z.infer to keep TypeScript types in sync with Zod schemas
- Consolidate command invocation schemas in command_manager.ts
- Add proper JSDoc documentation for all new interfaces and functions
- Update macro execution to use command IDs instead of names for consistency

Bug: https://github.com/google/perfetto/issues/1342
Bug: https://github.com/google/perfetto/issues/2551

### DIFF
--- a/ui/src/assets/components/json_settings_editor.scss
+++ b/ui/src/assets/components/json_settings_editor.scss
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-@import "../../assets/theme";
+@import "../theme";
 
-.pf-macro-settings {
-  width: 500px;
+.pf-json-settings-editor {
+  width: 400px;
 
   &__editor-section {
     margin-bottom: 10px;
--- a/ui/src/assets/perfetto.scss
+++ b/ui/src/assets/perfetto.scss
@@ -29,6 +29,7 @@
 // Widgets/components - keep these sorted alphabetically
 @import "components/aggregation_adapter";
 @import "components/data_grid";
+@import "components/json_settings_editor";
 @import "components/query_history";
 @import "components/query_table";
 @import "widgets/anchor";
--- a/ui/src/components/json_settings_editor.ts
+++ b/ui/src/components/json_settings_editor.ts
@@ -1,0 +1,135 @@
+// Copyright (C) 2025 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {z} from 'zod';
+import m from 'mithril';
+import {Editor} from '../widgets/editor';
+import {Callout} from '../widgets/callout';
+import {Intent} from '../widgets/common';
+import {Button} from '../widgets/button';
+import {Setting} from '../public/settings';
+
+export interface JsonSettingsEditorOptions<T> {
+  // Zod schema for validation
+  schema: z.ZodSchema<T>;
+  // Optional validator function for additional business logic validation
+  validator?: (data: T) => string | undefined;
+}
+
+export class JsonSettingsEditor<T> {
+  private textareaValue: string | undefined;
+  private originalValue: string | undefined;
+  private jsonError: string | undefined = undefined;
+  private currentSetting: Setting<T> | undefined;
+
+  constructor(private options: JsonSettingsEditorOptions<T>) {}
+
+  render(setting: Setting<T>): m.Children {
+    this.currentSetting = setting;
+    this.initializeTextValue();
+
+    return m('div', {className: 'pf-json-settings-editor'}, [
+      m('div', {className: 'pf-json-settings-editor__editor-section'}, [
+        m(Editor, {
+          text: this.textareaValue,
+          className: 'pf-json-settings-editor__editor',
+          onUpdate: (text: string) => this.handleUpdate(text),
+          onSave: () => this.handleSave(),
+        }),
+        this.jsonError !== undefined &&
+          m(
+            Callout,
+            {
+              intent: Intent.Danger,
+              className: 'pf-json-settings-editor__error',
+            },
+            `JSON Error: ${this.jsonError}`,
+          ),
+        m('div', {className: 'pf-json-settings-editor__actions'}, [
+          m(Button, {
+            label: 'Save',
+            disabled: this.isSaveDisabled(),
+            onclick: () => this.handleSave(),
+          }),
+        ]),
+      ]),
+    ]);
+  }
+
+  private initializeTextValue(): void {
+    if (this.textareaValue === undefined && this.currentSetting) {
+      const data = this.currentSetting.get();
+      this.originalValue = this.stringifyData(data);
+      this.textareaValue = this.originalValue;
+    }
+  }
+
+  private stringifyData(data: T): string {
+    return JSON.stringify(data, null, 2);
+  }
+
+  private handleUpdate(text: string): void {
+    this.textareaValue = text;
+    this.validateAndSetError(text);
+  }
+
+  private handleSave(): void {
+    if (this.textareaValue === undefined || !this.currentSetting) return;
+    const validatedData = this.validateAndSetError(this.textareaValue);
+    if (validatedData !== undefined) {
+      this.currentSetting.set(validatedData);
+      this.originalValue = this.textareaValue;
+    }
+  }
+
+  private hasUnsavedChanges(): boolean {
+    return this.textareaValue !== this.originalValue;
+  }
+
+  private isSaveDisabled(): boolean {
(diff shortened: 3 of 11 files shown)


anakryiko:
@LalitMaganti I'm sorry if this is a stupid question, but skimming through code I didn't figure this out. Is there a way to add these startup commands into the Perfetto trace itself? Or how is this supposed to be used? I didn't find any Perfetto protobuf changes, so I'm a bit lost. Thank you!

LalitMaganti:
image

1) Go to ui.perfetto.dev, make sure you are on the autopush channel if you're not already
2) Click on settings in sidebar
3) Add the commands you want to run on UI startup there
4) Reload the UI and open a trace and those commands you put there will automatically run when the trace is opened.

Will write some documentation for this once all the features land.

anakryiko:
Ok, so this isn't something that my tool can add into Perfetto trace itself while generating it, right? I can see how this is useful for power users to automate some of their workflow or something, but I was hoping to have something like this as part of the trace itself and let tools record additional UI actions based on trace. Is this something that is in your plans as well?

LalitMaganti:
It's really important to us that the "raw contents of the trace" be separated from "how the UI should be used to visualize it". There are too many problems which come with conflating the two, we saw it the hard way with chrome://tracing.

But maybe we should think about relaxing things slightly so there is *some* way of controlling the UI with commands on startup. I think no matter what we will want some layer of separation between the two; we'll need to figure out exactly what that separation should look like.

This might end up intersecting with https://github.com/google/perfetto/issues/1018: maybe we can inject the UI commands in a separate file which is tarred/zipped together with the raw trace file. That way: 1) we can always "extract" the raw trace at any time very easily 2) users can turn it off if they don't want external tools deciding how the trace should appear (i.e. they want to see the "raw" trace file).

anakryiko:
@LalitMaganti this makes sense. From usability POV it's just important that whatever tools that rely on Perfetto UI for visualization can have some augmentation of UI (e.g., to create additional debug tracks from data in the trace) and that this is possible to do without having to deploy custom UI plug ins.

As you said, tar/zip archive with trace and separately a set of startup commands would work well, provided this archive would be understood natively by Perfetto UI (e.g., through use of `?url=` param, just like it is today with protobuf-based trace).

Thank you for working and thinking about this!

LalitMaganti:
> @LalitMaganti this makes sense. From usability POV it's just important that whatever tools that rely on Perfetto UI for visualization can have some augmentation of UI (e.g., to create additional debug tracks from data in the trace) and that this is possible to do without having to deploy custom UI plug ins.

Sure but we need this to be done in a sustainable way. While we're very happy other people use us in this way, it's also a non-trivial maintenance cost for us to have custom ad-hoc features for other folks. That's why plugins were invented in the first place: to take this burden away from us and shift it onto the people who are writing ese things. 

> e.g., through use of ?url= param, just like it is today with protobuf-based trace

Wait are you using the URL parameter for this? If so, I implemented https://github.com/google/perfetto/pull/2655: you can pass a `startupCommands` URL parameter which will let you run commands on startup. You don't need to modify the trace at all.

> As you said, tar/zip archive with trace and separately a set of startup commands would work well, provided this archive would be understood natively by Perfetto UI

We already understand tar/zip, it's matter of coming up with a schema and file format and doing the plumbing.

anakryiko:
> > @LalitMaganti this makes sense. From usability POV it's just important that whatever tools that rely on Perfetto UI for visualization can have some augmentation of UI (e.g., to create additional debug tracks from data in the trace) and that this is possible to do without having to deploy custom UI plug ins.
> 
> Sure but we need this to be done in a sustainable way. While we're very happy other people use us in this way, it's also a non-trivial maintenance cost for us to have custom ad-hoc features for other folks. That's why plugins were invented in the first place: to take this burden away from us and shift it onto the people who are writing ese things.

Yes, of course, I understand and agree. My whole point was that having something self-contained (like the archive you mentioned) is important to have tools that are completely detached from the Perfetto installation itself, otherwise it gets extremely complicated logistics-wise. Which is why I don't think a plug-in model works great for my needs. 

> 
> > e.g., through use of ?url= param, just like it is today with protobuf-based trace
> 
> Wait are you using the URL parameter for this? If so, I implemented #2655: you can pass a `startupCommands` URL parameter which will let you run commands on startup. You don't need to modify the trace at all.

My users upload traces emitted by my tool (tool has nothing to do with the upload itself), and then download them through `?url=`. But my tool doesn't generate URL and doesn't have control over it (it's just a stand-alone CLI tool emitting a file, nothing more).

> 
> > As you said, tar/zip archive with trace and separately a set of startup commands would work well, provided this archive would be understood natively by Perfetto UI
> 
> We already understand tar/zip, it's matter of coming up with a schema and file format and doing the plumbing.

Oh, cool, good to know. Well, I'll keep watching this space while you work on this, thanks!

LalitMaganti:
> My whole point was that having something self-contained (like the archive you mentioned) is important to have tools that are completely detached from the Perfetto installation itself, otherwise it gets extremely complicated logistics-wise. 

Why does it need to be complicated? The way I see it is very similar to e.g. how the Linux kernel works with respect to development:
1) If you want to keep up to date with the HEAD of the UI, you write some small self contained thing which lives upstream and then you can ask your users who use your tool to enable it as it will be shipped by default with ui.perfetto.dev. Anyone who doesn't use your tool doesn't know or even see it existing: I see this as analogous to how you can enable/disable modules in the kernel.
2) If you don't want to upstream your plugin, you can fork the UI, host a copy yourself on whatever server you want (it's fully client side so any off the shelf hosting provider would work) and there you can enable your plugin by default.

This is not to say that I don't also want to help lower the bar so that many usecases will *not* need plugins. But at the end of the day I don't think it's so bad to have something as a plugin either.

anakryiko:
@LalitMaganti if you want to complete Linux kernel analogy, what I'm asking about is something like BPF: a way to have custom and potentially extremely ad-hoc, short-lived, experimental logic extending some aspects of kernel, that is not a built-in part of kernel, but relies on infrastructure provided by the kernel to do whatever custom thing needs to be done at the moment. It's not worth upstreaming because it's either too custom, or it's still experimental and evolving and with high likelihood will change.

This is not a desire to circumvent upstream process, it's just some small and very non-structured extensions that just don't lay themselves well to "upstreaming" as plugin/driver/part of kernel.

I hope this makes sense.

LalitMaganti:
Yes it does make sense. It's why I said "This is not to say that I don't also want to help lower the bar so that many usecases will not need plugins.": I agree for a lot of ad-hoc cases you will want a more lightweight way for to compose together a few primitives without needing to go to the trouble of writing a plugin.

My thought was that if you're going to the trouble of building an tool which is using Perfetto, building a plugin is maybe not to much work but maybe I'm not grasping the nature of your tool and how experimental/fast moving it is.

In any case, to be clear, I still want to invest in the the above: I'm writing a doc to comprehensively cover how we want to tackle customising the UI on startup: it's a non-trivial problem because of the variety of different ways in which people use the UI and all of these people want to customize thigns in different ways.

Here's the key extract from the doc so you have a sense of the space:
image

LalitMaganti:
Here's an updated table which maybe explains better the situation: 
image

Note specifically the tension between (2) and (4)/(5): people want to customize the UI at different levels and that's what makes things complicated.

anakryiko:
@LalitMaganti thanks a lot for sharing that excerpt from the document! Yes, my case is squarely in (5), but I agree that either "dashboard" (or whatever is linking to the trace) or user might want to adjust the UI to their specific needs. I obviously know nothing about all the internal complexities, but conceptually I think all that can and should coexist, with perhaps just a properly set priorities? E.g., user customization and choices beats whatever dashboard might set up, while dashboard customizations are applied on top of whatever tool "baked in" into the trace (or along with the trace, as we discussed earlier).

I appreciate the complexity of making all this work seamlessly, though. So thanks again for being patient with me and having this discussion. Looking forward to all the planned Perfetto work your team is doing.

LalitMaganti:
> but conceptually I think all that can and should coexist, with perhaps just a properly set priorities? E.g., user customization and choices beats whatever dashboard might set up, while dashboard customizations are applied on top of whatever tool "baked in" into the trace (or along with the trace, as we discussed earlier).

Conceptually yes but the devil is very much in the details. And also every new "entrypoint" we expose is an entrypoint we have to maintain ~forever. So we have to do it thoughtfully.

LalitMaganti:
Discussed within the team: I think we have agreement on doing this and we also have a rough shape how we will do this: will involve a JSON file TARed with the trace describing what commands should be run on startup. 

Will look into implementing this in the coming weeks/months. There are other things I need to finish off before diving into this.

anakryiko:
@LalitMaganti great! And this tar approach will be backwards compatible with older Perfetto UI that doesn't support this? I.e., will it just pick up a traces inside that TAR archive, ignoring that JSON? Or I'd need to make a decision to go with just protobuf trace for older Perfetto installation vs TAR for newer ones?

LalitMaganti:
> And this tar approach will be backwards compatible with older Perfetto UI that doesn't support this? I.e., will it just pick up a traces inside that TAR archive, ignoring that JSON? Or I'd need to make a decision to go with just protobuf trace for older Perfetto installation vs TAR for newer ones?

I think it should be backward compatible. I will check this when I implement this and let you know.

LalitMaganti:
Even if not, the Perfetto UI stable release happens once a month so within a month, everyone will have the new feature. You could just wait until it's available on stable.

PR #2655, opened 2025-08-21
ui: implement URL parameter commands for trace-specific startup commands

LalitMaganti:
- Add 'startupCommands' URL parameter to route schema for JSON-encoded command arrays
- Add parseUrlCommands() utility function to parse and validate URL commands
- Modify trace startup command system to execute URL commands before settings commands
- URL commands have higher priority and are trace-specific vs global settings commands
- Commands are parsed once during trace construction for efficiency

Example URL: #!/viewer?startupCommands=[{"id":"perfetto.CoreCommands#RunQuery","args":[]}]

Bug: https://github.com/google/perfetto/issues/1342

### DIFF
--- a/ui/src/core/command_manager.ts
+++ b/ui/src/core/command_manager.ts
@@ -62,6 +62,27 @@ export function validateCommandInvocations(
   return invalidCommands;
 }
 
+/**
+ * Parses URL commands parameter from route args.
+ * @param commandsParam URL commands parameter (JSON-encoded string)
+ * @returns Parsed commands array or undefined if parsing fails
+ */
+export function parseUrlCommands(
+  commandsParam: string | undefined,
+): CommandInvocation[] | undefined {
+  if (!commandsParam) {
+    return undefined;
+  }
+
+  try {
+    const parsed = JSON.parse(commandsParam);
+    const result = commandInvocationArraySchema.safeParse(parsed);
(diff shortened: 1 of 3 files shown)


LalitMaganti:
@pengj: inspired by your work at https://github.com/google/perfetto/pull/2471, I'm implementing this to be a much more general way to run arbitrary code on UI startup from URL parameters. I think we can remove the ad-hoc code in dev.perfetto.AutoPinAndExpandTracks and instead rely on this mechanism to do what you were trying to do in that CL.

We will likely need to create a new perfetto.PinTracksByNameRegex command which is a more generic version of the current command perfetto.PinTrackByName.

pengj:
@LalitMaganti this approach is neat!  we may need to two commands here: expand and pin the track based on the Name Regex. do you think we should implement these within `AutoPinAndExpandTracks` or within `commands`?

LalitMaganti:
I already implemented PinTracksByRegex in https://github.com/google/perfetto/pull/2670. I can add ExpandTracksByRegex as well.

pengj:
it will be good we can add that support. once both command merged, I think we can remove the expand and pin code in dev.perfetto.AutoPinAndExpandTracks.

LalitMaganti:
https://github.com/google/perfetto/pull/2670 now also adds expand/collapse commands.

PR #2664, opened 2025-08-21
tp: etm: improve etm decode

OmarElfouly:
As a follow up on:
- https://github.com/google/perfetto/pull/2643#discussion_r2288658793
- https://github.com/google/perfetto/pull/2643#discussion_r2288748821

This PR changes a few things:
- Adds a new trace with its own diff test that contains timestamps and cycle counts
- Refactors the decode logic to make it more readable
- In the case where a timestamp was not found for a sync packet it now reports a null last seen timestamp to reflect that in this section of the data the timestamp was essentially lost instead of maintaining the last seen timestamp which at this point would be outdated.
- Makes sure the buffers data is used correctly instead of the current cursors data


test/trace_processor/diff_tests/parser/etm/ts_cc.out
```diff
@@ -0,0 +1,101 @@
+"chunk_index","element_index","element_type","timestamp","cycle_count","last_seen_timestamp","cumulative_cycles","exception_level","context_id","isa","start_address","end_address","mapping_id"
+12,0,"NO_SYNC","[NULL]","[NULL]","[NULL]","[NULL]","[NULL]","[NULL]","UNKNOWN",-1,-1,"[NULL]"
+42,1,"TRACE_ON","[NULL]","[NULL]","[NULL]","[NULL]","[NULL]","[NULL]","UNKNOWN",-1,-1,"[NULL]"
+42,2,"PE_CONTEXT","[NULL]","[NULL]","[NULL]","[NULL]",1,20618,"AARCH64",-1,-1,"[NULL]"
+42,3,"ADDR_NACC","[NULL]","[NULL]","[NULL]","[NULL]",1,20618,"AARCH64",-135284010340,-1,129
```
rasikanavarange, line 5: nit: it doesn't really matter for the purposes of testing the cycle count and timestamp, but did something go wrong with decoding here? Why are there no INSTR_RANGE packets?

test/trace_processor/diff_tests/parser/etm/ts_cc.out
```diff
@@ -0,0 +1,101 @@
...
+127,27,"TIMESTAMP",2652955353049,109,2652955353049,109,1,20618,"AARCH64",-135284010452,-1,"[NULL]"
+133,28,"SYNC_MARKER",2652955353050,130,2652955353050,130,1,20618,"AARCH64",-135284010428,-1,"[NULL]"
+133,29,"TIMESTAMP",2652955353050,130,2652955353050,130,1,20618,"AARCH64",-135284010452,-1,"[NULL]"
+135,30,"SYNC_MARKER",2652955353051,157,2652955353051,157,1,20618,"AARCH64",-135284010428,-1,"[NULL]"
+139,31,"ADDR_NACC","[NULL]","[NULL]",2652955353051,157,1,20618,"AARCH64",-135284002676,-1,129
+145,32,"ADDR_NACC","[NULL]","[NULL]",2652955353051,157,1,20618,"AARCH64",-135288285204,-1,130
+150,33,"ADDR_NACC","[NULL]","[NULL]",2652955353051,157,1,20618,"AARCH64",-135288284500,-1,130
+156,34,"ADDR_NACC","[NULL]","[NULL]",2652955353051,157,1,20618,"AARCH64",-135259263988,-1,0
+159,35,"CYCLE_COUNT","[NULL]",264,2652955353051,264,1,20618,"AARCH64",-135259263988,-1,"[NULL]"
+164,36,"TIMESTAMP",2652955353051,157,2652955353051,421,1,20618,"AARCH64",-135259263988,-1,"[NULL]"
```
rasikanavarange: I don't think TIMESTAMP packets should have a cumulative cycle count associated, since their position in the stream is incorrect (it should be at the SYNC_MARKER).

Having the cumulative cycle count in TIMESTAMP packets produces weird results like in this row, which is misleading.

test/trace_processor/diff_tests/parser/etm/ts_cc.out
OmarElfouly: I beleive the libc included is not the same libc being used when I ran the trace which is probably what lead to these issues.

test/trace_processor/diff_tests/parser/etm/ts_cc.out
OmarElfouly: Oh that makes sense! Changing it now

OmarElfouly: Done :)

rasikanavarange (approved):

PR #2669, opened 2025-08-21
Add a new code kind in v8.proto

thibaudmichaud:
The new code kind is added here:
https://chromium-review.googlesource.com/c/v8/v8/+/6827264

It represents a special function generated by V8, that handles entering and exiting a wasm stack in the context of the wasm stack switching proposal (https://github.com/WebAssembly/stack-switching).

To properly log this new code kind with perfetto, this change adds a corresponding enum value in V8InternalCode.


LalitMaganti (commented):
You'll need to run `tools/gen_all <path to out directory>` to regenerate a bunch of autogenerated file. The CI should fail on this patch as-is.

LalitMaganti (approved):

PR #2693, opened 2025-08-22
[SDK] Expose optional user provided process uuid

etiennep-chromium:
This PR allows client to provide a custom process track uuid on TracingInitArgs

This is used in Chrome to allow the browser process decide on child process uuids, and thus emit events directly
to them without a roundtrip.
https://chromium-review.googlesource.com/c/chromium/src/+/6874371

### DIFF
--- a/include/perfetto/tracing/tracing.h
+++ b/include/perfetto/tracing/tracing.h
@@ -164,7 +164,11 @@ struct TracingInitArgs {
   // Not yet supported for consumer connections currently.
   CreateSocketAsync create_socket_async = nullptr;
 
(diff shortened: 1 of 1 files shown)


include/perfetto/tracing/tracing.h
```diff
@@ -163,6 +164,9 @@ struct TracingInitArgs {
   // Not yet supported for consumer connections currently.
   CreateSocketAsync create_socket_async = nullptr;
 
+  // [Optional] Specifies a custom track uuid for the current current.
```
ddiproietto: Can we expand this comment a little bit?

ddiproietto (approved):
Thanks!

PR #2698, opened 2025-08-25
tp: Refactored thread APIs to not associate main thread by default

jahdiel-alvarez:
The process_tracker thread APIs no longer assume the tid-pid-match-main-thread pattern by default. This decouples PID from main thread TID in trace processor and allows for non-matching TID to be the main thread, as is done in some non-Linux OSes.

Bug: 439687339


src/trace_processor/importers/common/process_tracker.cc
```diff
@@ -306,19 +306,21 @@ UniquePid ProcessTracker::StartNewProcessInternal(
   // we keep them in |tids_| (if they weren't erased by EndThread()), but ignore
   // them in GetThreadOrNull().
 
+  // Same pid is never used concurrently by multiple processes, therefore remove
```
LalitMaganti, line 305: Can you explain why this is safe to do? See comment literally above? Maybe remove the comment if it no longer applies?

LalitMaganti: Shouldn't this maybe be conditional on is_main_thread?

jahdiel-alvarez: I am not sure why this comment was added. @betasheet do you have any insights on this? We need strong invariants to avoid ending up with edge cases which developers depend on. Regardless of the reason the comment was added for, we simply need to fix those cases. Concurrent pids simply will cause too many issues in TP. Btw concurrent TIDs in TP work only because of the assumption of PID uniqueness. So we have to guarantee that.

jahdiel-alvarez: @LalitMaganti this shouldn't be a conditional. This is only relevant if the associate_main_thread is true. If processes are created without associated main threads the operation becomes a NOP.

betasheet: I don't recall, sorry. I presume I was already thinking along the same lines as Jahdiel -- that it would probably make sense to remove all those tids, too. I'd remove the TODO though, this seems to accomplish this now.

jahdiel-alvarez: Removed the TODO

LalitMaganti (approved):

PR #2699, opened 2025-08-25
base: Introduce LockFreeTaskRunner

primiano:
Introduce a new taskrunner class that is designed to replace
and sunset UnixTaskRunner. Thew new class avoids locks and also
avoids any allocations (*) for cases when there are < 512 tasks
outstanding (512 being the size of a slab). It does so by using
two Slabs and a freelist of size one like a graphics double buffer.

(*) It doesn't count for allocations caused by copying around
    tasks' std::function, if the bound task goes beyond the inline
    storage of std::function and requires heap.

Bug: b/394504259

**Stack:**
- [#2697](https://github.com/google/perfetto/pull/2697) (tr_tests)
  - **[This PR] (tr_lockfree)**


primiano:
@ddiproietto @sashwinbalaji this is ready to go now!
See attached doc with a description of the design

include/perfetto/ext/base/lock_free_task_runner.h
```diff
@@ -0,0 +1,273 @@
...
+// 2. Publishing: After writing the task into its reserved slot, the writer
+//    thread atomically sets the corresponding bit in the `tasks_written`
+//    bitmask. This acts as a memory barrier and makes the task visible to
+//    the consumer (main) thread.
+//
+// 3. Consumption: The main thread acquire-reads the `tasks_written` bitmask.
+//    For each bit that is set, it processes the task and then sets the
+//    corresponding bit in its private `tasks_read` bitmask to prevent
+//    reading the same task again.
+struct Slab {
```
sashwinbalaji, line 205: nit: could there be false sharing within the slab ?  
1) for tasks_written multiple writer can be requesting for different BitWord
2) tasks_read(reader) & next_task_slot /tasks_written  (writer) can be on same cache line ?
shall we add padding to address it ?

primiano: > nit: could there be false sharing within the slab ?

WDYM with false sharing?

> for tasks_written multiple writer can be requesting for different BitWord

Yes this can happen, but the bitwords are all atomic, and they do atomic-or operations (unless i missed something somewhere)?

> tasks_read(reader) & next_task_slot /tasks_written (writer) can be on same cache line ?
shall we add padding to address it ?

hah this is actually a very good idea. Didn't think about rearranging things to match (or in this case spread apart) cache lines.

docs/design-docs/lock-free-task-runner.md
```diff
@@ -0,0 +1,340 @@
...
+
+This document focuses only on the design of PostTask() and does not discuss
+PostDelayedTask() or Add/RemoveFileDescriptorWatch() at all. The logic of these
+functions is unchanged from the legacy UnixTaskRunner and they are simply
+implemented by hopping first to the task runner thread, and manipulating the
+delayed task list and FD set on the main thread.
+This involves an extra hop (if called from other threads) vs the legacy
+UnixTaskRunner. However: (1) in practice most calls to PostDelayedTask() and
+Add/RemoveFileDescriptorWatch() happen on the main thread in our codebase; (2)
+they are almost never hot paths.
```
LalitMaganti, line 52: I hope it would be "never" or we have a ticking clock on our hands :)

docs/design-docs/lock-free-task-runner.md
```diff
@@ -0,0 +1,340 @@
...
+A singly-linked list with only a tail pointer suggests that the reader has a
+worst case complexity of O(N), as it has to traverse the whole list to get to
+the first tasks (it must run tasks FIFO). However, in practice we expect to have
+only two slabs ever (and if we have a queue of 10k-100k tasks, walking the list
+is our last problem).
+
+The main compromise of this design is that it scales poorly with a large number
+of tasks, as Run() becomes both slower (to traverse the list) and stack-greedy
+(it uses recursion on the stack to walk the list without using the heap).
+We don't expect a large number of outstanding tasks in Perfetto.
```
LalitMaganti: There are cases this is not true (we might want to fix b/330580374)

include/perfetto/ext/base/lock_free_task_runner.h
```diff
@@ -0,0 +1,273 @@
...
+    // polling it again until the queued task runs. On Windows we can't do that.
+    // Instead we keep track of its state here.
+    bool pending = false;
+#else
+    size_t poll_fd_index;  // Index into |poll_fds_|.
+#endif
+  };
+
+  // Accessed only from the main thread.
+  std::map<PlatformHandle, WatchTask> watch_tasks_;
```
LalitMaganti: is std::map the right structure? Should we use unordered_map?

include/perfetto/ext/base/lock_free_task_runner.h
```diff
@@ -0,0 +1,273 @@
...
+
+  // Stats for testing.
+  std::atomic<size_t> slabs_allocated_{};
+  std::atomic<size_t> slabs_freed_{};
+};
+
+namespace task_runner_internal {
+
+// Returns the index of the refcount_ bucket for the passed Slab pointer.
+static uint32_t HashSlabPtr(Slab* slab) {
```
LalitMaganti, line 177: Suggest using Murmur function for uint64 instead of a new arbitrary hash function. Realised while working with that there are good reasons for using off the shelf hash functions.

include/perfetto/ext/base/lock_free_task_runner.h
LalitMaganti: > shall we add padding to address it ?

+1 on this, you'll get a decent bump if you do this right.

primiano: As discussed offline, I tried, but in reality given that each member is a big array, there is very little sharing as soon as you populate the first 2 entries of the tasks. And i cannot not-share the bitmaps, beucase if I go from 1 bit to 64 bytes... that's a lot of cache misses.

PR #2714, opened 2025-08-26
Add new android.virtualization_framework data source

nioffe:
This PR simply adds the protos and runs the codegen scripts. The code to register new data sources will be added in the follow-up PR.

Bug: 222347044


rsavitski:
I'd like to see the follow-up PR before reviewing the protos. Please put me on reviewer or assignee line once that one's ready.

PR #2715, opened 2025-08-26
ui: Fix crash when loading traces larger than ~2GB

stevegolton:
Attempting to cache a trace larger than around 2GB is currently crashing on when attempting to create an arraybuffer from a file for the sake of caching. Caches can actually take a ReadableStream so we can just take a stream from the file instead. Unfortunately it looks like we have a separate problem in that chrome cannot cache files larger than 2GB anyway.

Since caching is best effort anyway, this patch just makes it so that if we fail to cache, we log it and don't convert the URL to contain the cache UUID, but we ultimately silently ignore the issue and move on.

Fixes: https://buganizer.corp.google.com/issues/435023274


ui/src/core/cache_manager.ts
```diff
@@ -129,11 +129,22 @@ export async function cacheTrace(
...
-  return true;
+  const key = `/_${TRACE_CACHE_NAME}/${traceUuid}`;
+  await cachePut(key, new Response(trace, {headers}));
+
+  // Verify the file was actually cached, large files can silently fail.
+  const cached = await cacheMatch(key);
+  if (cached) {
+    return true;
+  } else {
+    // If we fail to cache the trace, this isn't a critical error. Non
```
LalitMaganti, line 140: We should update the comment in `maybeOpenCachedTrace` to indicate that the trace being over 2GB is another reason this local_cache_key might fail.

stevegolton: `local_cache_key` is never set if this fails

LalitMaganti: Ack.

LalitMaganti (approved):

ui/src/core/cache_manager.ts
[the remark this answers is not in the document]
stevegolton: Done

PR #2728, opened 2025-08-27
exp: Add interval intersect node

aMayzner:
Add support for interval intersect node and support for multiple child nodes


stevegolton (approved):
Looks good, though I did manage to get this error while clicking around quickly on the nodes - not sure it's related to this PR but still.

image

Issue #1734, opened 2025-06-04
Coloring of trace event based on thread time misleading in system trace.


OlivierLi:
<img width="1343" alt="Image" src="https://github.com/user-attachments/assets/436c2779-b771-488f-8bb4-ca787e05db01" />

Please see attached screenshot where the coloring of the ThreadControllerActive shows ~15% light coloring where the system trace even analysis shows it's 99.69% on-cpu.

LalitMaganti:
The two things you are referencing comes from different data sources:
1) the colouring comes from chrome track event which is very coarse grained, just getting the thread time from start and end of the slice from a pretty coarse grained (us precision) clock source 
2) the scheduling info comes from ftrace and is very precise as it's directly what the kernel is telling us. 

Almost certainly the discrepancy is coming from the different sources of data. To confirm we'd need to look at the trace. Likely this should be filed against a chrome bug to see if they can get data from a more precise source.

LalitMaganti:
As stated above, this should be filed against Chrome. Nothing more we can do on our side.

PR #2737, opened 2025-08-27
[TrackEvent] Add thread time subsampling option

etiennep-chromium:
Profiler data shows thread time sampling is the biggest overhead when tracing.
This PR adds an option to subsample thread time.


protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -3848,6 +3848,14 @@ message TrackEventConfig {
   // https://man7.org/linux/man-pages/man3/clock_gettime.3.html
   optional bool enable_thread_time_sampling = 8;
 
+  // When enable_thread_time_sampling is true, and this is specified, thread
+  // time is sampled only if the elapsed wall time >
+  // `thread_time_subsampling_ns`. Otherwise, the elapsed wall time is used
+  // instead of thread time. Effectively, this means thread time will have a
```
LalitMaganti: Can we not do this? I would prefer it to be null rather than lying about it and then it making its way into tables as an authoritative value.

We already get questions like https://github.com/google/perfetto/issues/1734, this will only make things much worse.

etiennep-chromium: If we have a slice A, immediately followed by a slice B:
A.end samples thread time
B.begin 1us later, skips thread time
B.begin samples thread time
Say we set it to null, we won't have the thread time for neither event. The desired behavior is to use thread time sampled in A.end as the approximate thread time at B.begin to show to shaded area in perfetto UI.
We could alternatively emit null here and compute this in trace_processor (that has the additional benefit of saving space in the trace buffer).

But the bigger point is that thread time is very inaccurate already; adding subsampling (to save on roughly 25% overhead of tracing) and using a thread time that's slightly off will make very little difference in (already poor) accuracy.

LalitMaganti: > Say we set it to null, we won't have the thread time for neither event. The desired behavior is to use thread time sampled in A.end as the approximate thread time at B.begin to show to shaded area in perfetto UI.

Sure but can we not emit it into the trace when we use walltime for a slice? Like having it empty is better than pretending it's walltime.

In general when people see stuff in a Perfetto trace they expect it to be in the ballpark of accuracy. If you are saying you are just making up stuff, I really don't like it as we end up having to bear the brunt of questions of "X looks weird, why is that?".

etiennep-chromium: To confirm, are you ok with
1- emit null (empty) here
2- modify trace processor to compute the (approximate) thread time proportion when it makes sense (if the event is bigger than the subsampling period)

LalitMaganti: > 2- modify trace processor to compute the (approximate) thread time proportion when it makes sense (if the event is bigger than the subsampling period)

If you want to estimate it properly, sure. Just don't use wall time as the estimate: if you're going to use walltime, just write NULL.

etiennep-chromium: Discussed offline: I implemented a solution where we emit 0, and 0 is considered nil on slice end.

src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -574,9 +577,15 @@ base::Status TrackEventTokenizer::AddExtraCounterValues(
           "Ignoring TrackEvent with more extra_{double_,}counter_values than "
           "TrackEventData::kMaxNumExtraCounters");
     }
+    auto value = static_cast<double>(*value_it);
```
LalitMaganti: Hmmm this is not the right place to put this. For other counter types, there's no subsampling so dropping counters is wrong. This should be gated on being thread time counter.

etiennep-chromium: Done, now handling in ParseThreadEndEvent

etiennep-chromium:
ddiproietto is OOO -> +primiano

PR #2740, opened 2025-08-28
Collect dmabuf RSS with process stats

tweej:
Dmabuf RSS is available alongside other per-pid files in procfs. Collect
it along with them. This also allows dmabuf RSS to be obtained from the
beginning of a trace if scan_all_processes_on_start is true in
process_stats_config.

Bug: 424646834

Welcome to Perfetto!

For more details please see


tweej:
Screenshot From 2025-08-27 17-54-42

src/traced/probes/ps/process_stats_data_source.cc
```diff
@@ -562,6 +566,16 @@ void ProcessStatsDataSource::WriteAllProcessStats() {
     if (proc_status.empty())
       continue;
 
+    std::string dmabuf_rss = ReadProcPidFile(pid, "dmabuf_rss");
```
LalitMaganti: Given this requires reading a separate file I think this should either be opt-in or opt-out. 

@rsavitski for thoughts.

rsavitski: @tweej I would also prefer if this had a config toggle, should be a straightforward bool like for these options[1]. Afaict this is not in upstream linux, so having it on by default is odd.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/config/process_stats/process_stats_config.proto;l=71-79;drc=a49d2530c8a16af55f023734f63743fca76f41b0

rsavitski (changes requested):

tweej:
Thanks, makes sense. Please give me a few days to get back to this.

PR #2741, opened 2025-08-28
 ui: name perf sample tracks after the counter

rsavitski:
The aim is to better represent what counter is being sampled on. For example there are android lab traces that record binder callstacks, using a tracepoint as the subject of the profiling. Current track names don't reflect that.

Also, make the panel auto-open only if there's exactly one profiled process.

Bug: 441579549


ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -59,19 +59,26 @@ export default class implements PerfettoPlugin {
 
   private async addProcessPerfSamplesTracks(trace: Trace) {
     const pResult = await trace.engine.query(`
-      SELECT DISTINCT upid
+      SELECT DISTINCT upid, pct.name AS cntrName
```
LalitMaganti: Can we guarantee for a given perf_session_id, there will be exactly one timebase `perf_counter_track`?

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -110,30 +117,28 @@ export default class implements PerfettoPlugin {
...
     const tResult = await trace.engine.query(`
-      select distinct
-        utid,
-        tid,
-        thread.name as threadName,
-        upid
-      from perf_sample
-      join thread using (utid)
-      where callsite_id is not null
+      SELECT DISTINCT
```
LalitMaganti, line 157: As I said on the bug, I'm not sure if it's a bug or feature that all callstacks for a given perf counter are merged together on a single timeline. I can see an argument both ways. Would like to discuss that a bit before landing this.

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
rsavitski: one "linux.perf" data source = one sampling group = one leader/timebase counter = one is_timebase per perf_session_id.

EDIT: there can be N per-cpu perf_counter_tracks, but they will all share the same name.

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
rsavitski: Ah yes, IMO the process-level aggregation needs to be split if there are multiple callstack sampling data sources per trace. One aggregate track per perf_session_id. People that want to see as much as possible can do a multi-track selection.

Thoughts?

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
LalitMaganti: Ack sg.

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
LalitMaganti: Yeah multi-track selection is indeed a good counter-argument to that yes. OTOH I worry that for e.g. 30 min stack sampling traces, it will cause a bunch of extra tracks for every thread and process. But maybe those extra tracks actually add value?

LalitMaganti: I thought more about this: I think maybe what we should do is as follows:
1) If there's a single session we just add it in as-is
2) if there's multiple sessions, I'd keep the "merged" view but then *also* add child tracks, one per timebase and show the "breakdown" by the timebase.

I think this is the best compromise.

rsavitski: Did you mean all of these at the same time?
* 1 track: process-level aggregation of *all* samples
* N tracks: process-level aggregation, grouped by timebase
* M tracks: per-thread-per-timebase

I'm instead leaning towards having:
* N tracks: process-level aggregation, grouped by timebase
* M tracks: per-thread-per-timebase

I'll try to remember to post some screenshots once this version of the UI builds.

LalitMaganti: I was actually thinking:
1 track: process-level aggregation of all samples
  -> N tracks: per timebase samples for process
M tracks: per-thread aggregation of all samples
  -> N tracks: per timebase samples for thread

rsavitski: I'm not a huge fan of aggregated tracks in the first place. They duplicate information from the same track group, which perfetto UI doesn't do elsewhere. But since we don't have any better UX atm, I can accept process-level aggregation tracks. Thread-level is too far imo.

My suggestion remains as uploaded, and people can use multi track selection to aggregate multiple timebases at the process/thread level:
7JG9cTNsv5uiCVs

LalitMaganti: Can we please pleas eplease have some better names and descriptions for these tracks? The names are so mysterious as to be completely non-understandable to anyone not an expert in perf.

If I collect a tools/cpu_profile trace, I'm going to be completely lost if I see a track named `process aggregate (cpu-clock)` in the trace. My firt question is going to be "where is my cpu profiling data".

rsavitski: Happy to special-case a simpler name for "cpu-clock" since that's our default for a "just give me a cpu profile" mode. Not sure about the rest. We don't have UX for visually grouping tracks so I'm keeping the thread name + TID prefix for that, and then there isn't that much room to spell out "callstacks sampled on binder:binder_transaction" or similar.

I don't particularly like this either (though I do like it more than HEAD), and am happy to hear suggestions. Or maybe we just need to revisit thread group UX before landing this.

LalitMaganti: > Happy to special-case a simpler name for "cpu-clock" since that's our default for a "just give me a cpu profile" mode. Not sure about the rest. 

SGTM.

>  then there isn't that much room to spell out "callstacks sampled on binder:binder_transaction" or similar.

Even just `Callstacks (clock name)` would be very helpful.

rsavitski: @stevegolton @LalitMaganti:
updated to be closer to what Lalit proposed on Monday - expandable summary tracks, which become headless if there's only one counter with samples for that process/thread.

Small snags:
* no longer clusters visually with the other thread tracks due to chevron icon for expanding
* I don't want to be clever with renaming child tracks based on whether the summary track is headless, so there's a lot of repetition (and waste of space in the name) if the summary track *is* there.

WDYT?

image

LalitMaganti: 1) The different visual look can be fixed by turning off "summary" flag (which is a bit unfortunate) because this is a summary track in the truest sense of the word but removes the visual noise at least).
2) I'm not a huge fan of the repetition but I can live with it.

rsavitski: IIRC the chevron stays even when it's not a summary, only the track shading changes. 

I'd actually prefer to keep the "isSummary" on because it makes area selections more intuitive (selecting the summary selects the children).

LalitMaganti (approved):
Looked at the screenshots from the test. Really like where this ended up. LGTM.

PR #2747, opened 2025-08-28
tracebox: improve default behaviour of tracebox on non-Android

LalitMaganti:
On all OSes today, tracebox defaults to using an abstract socket rather
than a system socket. This makes sense on Android as it's always
expected that the system daemons are running and we don't want to step
on them.

However, on Linux, it's very rare that someone has another daemon
running in the background. Instead, it's almost always the case that
someone is doing a "one-shot" trace collection and in that case we will
very much want to use system sockets to make sure that track_event, gpu
render stages (e.g. mesa), other data sources etc correctly connect ot
this instance of Perfetto.

The fact this doesn't work trips people up all the time: the mesa folk
are still using the tmux script because they though that something was
broken with tracebox: they didn't realise that --system-sockets was
necessary to correctly collect mesa traces.


src/tracebox/tracebox.cc
```diff
@@ -76,19 +79,27 @@ Usage in manual mode:
     printf(" %s", applet.name);
 
   printf(R"(
+Tracebox-specific args
+  --abstract-sockets    : Forces the use of abstract sockets when using
+                          autostart mode. Only has an effect on non-Android
+                          operating systems; Android only supports abstract
+                          sockets in autostart mode. Cannot be used in applet
+                          mode.
+  --system-sockets      : Obsolete flag: supported for backwards compatibility.
```
rsavitski: nit: I would still say what the flag does. Or at least mention that it's the alternative to --abstract-sockets, and now the default.

src/tracebox/tracebox.cc
```diff
@@ -118,16 +129,55 @@ int TraceboxMain(int argc, char** argv) {
     return 1;
   }
 
-  auto* end = std::remove_if(argv, argv + argc, [](char* arg) {
-    return !strcmp(arg, "--system-sockets");
-  });
-  if (end < (argv + argc - 1)) {
+  int64_t system_socket_count =
```
rsavitski: This feels too verbose for what's being done (removing tracebox-specific flags and checking for conflicts). And the argc-- bit could use a named variable.

How about something like:

```cpp
  // Parse out --system-sockets.
  char** new_end;
  new_end = std::remove_if(argv + 1, argv + argc, [&](std::string_view arg) {
    return arg == "--system-sockets";
  });
  int64_t system_socket_count = std::distance(new_end, argv + argc);
  argc -= static_cast<int>(system_socket_count);

  // Parse out --abstract-sockets.
  new_end = std::remove_if(argv + 1, argv + argc, [&](std::string_view arg) {
    return arg == "--abstract-sockets";
  });
  int64_t abstract_socket_count = std::distance(new_end, argv + argc);
  argc -= static_cast<int>(abstract_socket_count);

  if (system_socket_count + abstract_socket_count > 1) {
    PERFETTO_ELOG(
        "Cannot specify more than one of --system-sockets or "
        "--abstract-sockets");
    return 1;
  }
```

That's how I'd think about this operation: we're erasing and counting the flags at the same time, instead of two separate passes.

src/tracebox/tracebox.cc
```diff
@@ -118,16 +129,55 @@ int TraceboxMain(int argc, char** argv) {
...
-  } else {
+  }
+
+  enum {
+    kSystemSocket,
+    kAbstractSocket,
+  } socket_type =
+      PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID) || abstract_socket_count > 0
+          ? kAbstractSocket
+          : kSystemSocket;
```
rsavitski: Do we really need this for a single if-else?

bool use_abstract_sockets = ...;
if (use_abstract_sockets) {
...
} else {
...
}

src/tracebox/tracebox.cc
```diff
@@ -150,6 +200,37 @@ int TraceboxMain(int argc, char** argv) {
...
+    PERFETTO_DCHECK(socket_type == kSystemSocket);
+
+    if (base::FileExists(GetConsumerSocket())) {
+      PERFETTO_ELOG(
+          R"(
+Failed to confirm the consumer and producer system socket were unused. This
+likely indicates that another `tracebox` session or `traced` daemon are running
+in the background.
+
+Either:
```
rsavitski, line 198: (b) is incomprehensible, (a) doesn't really say /why/ I'd want to choose that.

Can we instead say:
* if you want to connect to the existing tracing daemon, use XYZ
* if you want to start this tracebox with its own sockets, use XYZ, but beware...

Although maybe we also need to also call out that multiple concurrent ftrace data source implementations will conflict (they'll fight over the tracefs).

LalitMaganti:
Thanks Ryan. PTAL.

rsavitski (approved):

PR #2750, opened 2025-08-28
tp: add tar writer utility

LalitMaganti:
Useful for uncoming work to "bundle" traces in traceconv.

### DIFF
--- a/src/trace_processor/util/tar_writer.cc
+++ b/src/trace_processor/util/tar_writer.cc
@@ -1,0 +1,223 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


src/trace_processor/util/tar_writer.cc
```diff
@@ -0,0 +1,240 @@
...
+
+  // Set modification time to current time (in octal)
+  snprintf(header.mtime, sizeof(header.mtime), "%011lo",
+           static_cast<unsigned long>(time(nullptr)));
+
+  // Compute and set checksum
+  ComputeChecksum(&header);
+
+  // Write header
+  RETURN_IF_ERROR(WriteHeader(header));
```
sashwinbalaji: shall we consider to dedupe the header creation and write between AddFile and AddFileFromPath ?

LalitMaganti: Good idea. Done.

PR #2751, opened 2025-08-28
dev/lalitm/traceconv 2

LalitMaganti:
- **F**
- **f**


src/traceconv/main.cc
```diff
@@ -180,6 +191,13 @@ int Main(int argc, char** argv) {
       profile_no_annotations = true;
     } else if (strcmp(argv[i], "--full-sort") == 0) {
       full_sort = true;
+    } else if (strcmp(argv[i], "--symbol-paths") == 0) {
+      i++;
+      if (i < argc) {
```
sashwinbalaji: nit: in else elog to indicate wrong input ?

src/traceconv/trace_to_bundle.cc
```diff
@@ -0,0 +1,185 @@
...
+    FROM stack_profile_mapping
+    WHERE build_id != ''
+    ORDER BY name
+  )");
+  while (it.Next()) {
+    std::string name = it.Get(0).AsString();
+    if (!name.empty()) {
+      mapping_names.push_back(name);
+    }
+  }
```
sashwinbalaji: more of a question:
1) why the need for order by ?
2) why not filter empty names in sql itself ?

LalitMaganti: Both excellent points. Done.

src/traceconv/main.cc
```diff
@@ -180,6 +191,16 @@ int Main(int argc, char** argv) {
       profile_no_annotations = true;
     } else if (strcmp(argv[i], "--full-sort") == 0) {
       full_sort = true;
+    } else if (strcmp(argv[i], "--symbol-paths") == 0) {
```
rsavitski: We should be consistent with the rest of the if-else chain in checking i against argc. Either change all to the new style or this to old.

Compare this to:
```
    } else if (i <= argc && strcmp(argv[i], "--pid") == 0) {
      i++;
      pid = StringToUint64OrDie(argv[i]);
    } else if (i <= argc && strcmp(argv[i], "--timestamps") == 0) {
      i++;
```

src/traceconv/trace_to_bundle.cc
```diff
@@ -0,0 +1,179 @@
...
+
+  // Check if this is an Android trace - bundle mode doesn't work for Android
+  // yet.
+  bool is_android;
+  {
+    auto android_check = tp->ExecuteQuery(R"(
+      SELECT COUNT(*)
+      FROM metadata
+      WHERE name = 'android_build_fingerprint'
+        OR (name = 'system_release' AND value LIKE '%android%')
```
rsavitski: question: didn't we ban LIKE? I vaguely thought everything was supposed to use globs, but didn't keep up with when that restriction applies.

src/profiling/symbolizer/local_symbolizer.cc
```diff
@@ -310,59 +310,77 @@ std::optional<BinaryInfo> GetBinaryInfo(const char* fname, size_t size) {
...
-        PERFETTO_DLOG("Ignoring %s, index entry for %s already exists.", fname,
-                      base::ToHex(binary_info->build_id).c_str());
-      }
+  }
+  std::optional<BinaryInfo> binary_info = GetBinaryInfo(fname, size);
+  if (!binary_info) {
+    PERFETTO_DLOG("Failed to extract build id from %s.", fname);
+    return;
+  }
+  auto it = result.emplace(binary_info->build_id, FoundBinary{
```
rsavitski: style nit: probably a good spot to use structured bindings.

rsavitski (approved):
LGTM with *optional* nits (I didn't try it locally myself).

LalitMaganti:
Thanks!

PR #2752, opened 2025-08-28
Add test for empty trace_strings to string pool benchmark

JoeNotCharlesGoogle:
Gives a legible error on a corrupt /tmp/trace_strings file. Without this the benchmark will crash.


JoeNotCharlesGoogle:
> * The added check in `BM_StringPoolIntern` has an inverted condition, causing the benchmark to be skipped when it should run. Please see the inline comment for details.

Already removed.

LalitMaganti (approved):

PR #2753, opened 2025-08-28
stdlib: add ability to measure cycles per thread and cpu

cphlipot1:
Add the ability to measure CPU Cycles for each thread for unique
thread and CPU combos. This is useful when tracking per thread CPU
cycles on heterogeneous architectures where the user may still want
to treat cycles from certain CPUs differently instead of simply summing
them.

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
   USING (id)
 GROUP BY
   utid;
+
+-- Aggregated CPU statistics for each thread per CPU combination.
+CREATE PERFETTO TABLE cpu_cycles_per_thread_per_cpu (
+  -- Thread
+  utid JOINID(thread.id),
+  -- Unique CPU id. Joinable with `cpu.id`.
+  ucpu LONG,
+  -- The number of the CPU. Might not be the same as ucpu in multi machine cases.
+  cpu LONG,
+  -- Sum of CPU millicycles
+  millicycles LONG,
+  -- Sum of CPU megacycles
+  megacycles LONG,
+  -- Total runtime duration
+  runtime LONG,
+  -- Minimum CPU frequency in kHz
+  min_freq LONG,
+  -- Maximum CPU frequency in kHz
+  max_freq LONG,
+  -- Average CPU frequency in kHz
+  avg_freq LONG
+) AS
+SELECT
+  utid,
+  ucpu,
+  cpu,
+  cast_int!(SUM(dur * freq / 1000)) AS millicycles,
+  cast_int!(SUM(dur * freq / 1000) / 1e9) AS megacycles,
+  sum(dur) AS runtime,
+  min(freq) AS min_freq,
+  max(freq) AS max_freq,
+  cast_int!(SUM((dur * freq / 1000)) / SUM(dur / 1000)) AS avg_freq
+FROM _cpu_freq_per_thread
+GROUP BY
+  utid,
+  ucpu;
+
+-- Aggregated CPU statistics for each thread per CPU combination in a provided interval.
+CREATE PERFETTO FUNCTION cpu_cycles_per_thread_per_cpu_in_interval(
+    -- Start of the interval.
+    ts TIMESTAMP,
+    -- Duration of the interval.
+    dur LONG
+)
+RETURNS TABLE (
+  -- Thread with CPU cycles and frequency statistics.
+  utid JOINID(thread.id),
+  -- Unique CPU id. Joinable with `cpu.id`.
+  ucpu LONG,
+  -- The number of the CPU. Might not be the same as ucpu in multi machine cases.
+  cpu LONG,
+  -- Sum of CPU millicycles
+  millicycles LONG,
+  -- Sum of CPU megacycles
+  megacycles LONG,
+  -- Total runtime duration
+  runtime LONG,
+  -- Minimum CPU frequency in kHz
+  min_freq LONG,
+  -- Maximum CPU frequency in kHz
+  max_freq LONG,
+  -- Average CPU frequency in kHz
+  avg_freq LONG
+) AS
+SELECT
+  utid,
+  ucpu,
+  cpu,
+  cast_int!(SUM(ii.dur * freq / 1000)) AS millicycles,
+  cast_int!(SUM(ii.dur * freq / 1000) / 1e9) AS megacycles,
+  sum(ii.dur) AS runtime,
+  min(freq) AS min_freq,
+  max(freq) AS max_freq,
+  cast_int!(SUM((ii.dur * freq / 1000)) / SUM(ii.dur / 1000)) AS avg_freq
+FROM _interval_intersect_single!($ts, $dur, _cpu_freq_per_thread) AS ii
+JOIN _cpu_freq_per_thread AS c
+  USING (id)
+GROUP BY
+  utid,
+  ucpu;
--- a/test/trace_processor/diff_tests/stdlib/linux/cpu.py
+++ b/test/trace_processor/diff_tests/stdlib/linux/cpu.py
@@ -271,6 +271,80 @@ class LinuxCpu(TestSuite):
             92,243675648,0,962397,864000,864000,255157
             """))
 
+  def test_cpu_cycles_per_thread_per_cpu(self):
+    return DiffTestBlueprint(
+        trace=DataPath('android_cpu_eos.pb'),
+        query=("""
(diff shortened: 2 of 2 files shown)


src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
```diff
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
   USING (id)
 GROUP BY
   utid;
+
+-- Aggregated CPU statistics for each thread per CPU combination.
+CREATE PERFETTO TABLE cpu_cycles_per_thread_per_cpu (
```
LalitMaganti: Can we make these the base table and compute the other tables from these? Will be a lot more efficient than every table independently going to the sched table directly.

src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
```diff
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
...
+  -- Unique CPU id. Joinable with `cpu.id`.
+  ucpu LONG,
+  -- The number of the CPU. Might not be the same as ucpu in multi machine cases.
+  cpu LONG,
+  -- Sum of CPU millicycles
+  millicycles LONG,
+  -- Sum of CPU megacycles
+  megacycles LONG,
+  -- Total runtime duration
+  runtime LONG,
```
LalitMaganti: DURATION

src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
```diff
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
   USING (id)
 GROUP BY
   utid;
+
+-- Aggregated CPU statistics for each thread per CPU combination.
+CREATE PERFETTO TABLE cpu_cycles_per_thread_per_cpu (
+  -- Thread
+  utid JOINID(thread.id),
+  -- Unique CPU id. Joinable with `cpu.id`.
+  ucpu LONG,
```
LalitMaganti: JOINID(cpu.id)

src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
```diff
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
...
+GROUP BY
+  utid,
+  ucpu;
+
+-- Aggregated CPU statistics for each thread per CPU combination in a provided interval.
+CREATE PERFETTO FUNCTION cpu_cycles_per_thread_per_cpu_in_interval(
+    -- Start of the interval.
+    ts TIMESTAMP,
+    -- Duration of the interval.
+    dur LONG
```
LalitMaganti: DURATION

src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
```diff
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
...
+  min(freq) AS min_freq,
+  max(freq) AS max_freq,
+  cast_int!(SUM((dur * freq / 1000)) / SUM(dur / 1000)) AS avg_freq
+FROM _cpu_freq_per_thread
+GROUP BY
+  utid,
+  ucpu;
+
+-- Aggregated CPU statistics for each thread per CPU combination in a provided interval.
+CREATE PERFETTO FUNCTION cpu_cycles_per_thread_per_cpu_in_interval(
```
LalitMaganti: I've really come to regret these functions because the performance of them *sucks* when you join them with large number of intervals. Please document that in the function comment.

src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
```diff
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
   USING (id)
 GROUP BY
   utid;
+
+-- Aggregated CPU statistics for each thread per CPU combination.
```
LalitMaganti: We've started asking people to leave a comment saying what data source you need to enable in the trace to have this table populated. Please add that in here.

src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
```diff
@@ -144,3 +144,83 @@ JOIN _cpu_freq_per_thread AS c
...
+  sum(dur) AS runtime,
+  min(freq) AS min_freq,
+  max(freq) AS max_freq,
+  cast_int!(SUM((dur * freq / 1000)) / SUM(dur / 1000)) AS avg_freq
+FROM _cpu_freq_per_thread
+GROUP BY
+  utid,
+  ucpu;
+
+-- Aggregated CPU statistics for each thread per CPU combination in a provided interval.
```
LalitMaganti: We've started asking people to leave a comment saying what data source you need to enable in the trace to have this table populated. Please add that in here.

src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/thread.sql
cphlipot1: So i can definately create **cpu_cycles_per_thread** from this table with no downside, but were you asking about other tables in the other stdlib modules as well?

That would potentially make them depend on thread.sql when they currently don't unless you're also asking i move this table out of the thread.sql to general.sql or something and make it a private backing table for the stuff in all the other modules.

for now i will simply update **cpu_cycles_per_thread** to reference **cpu_cycles_per_thread_per_cpu** unless you would like like something else in addition to that.

LalitMaganti: I would maybe pull this out into a separate thread_cpu module and have both cpu and thread depend on that?

cphlipot1: sure we can do that.

cphlipot1: Actually i just checked. We currently have this code chunk already: https://github.com/google/perfetto/blob/dba5db9ffa1ee3411ea57c809fa8a0b1e549a3bb/src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/general.sql#L148-L163

AFIK all of these tables like cpu_cycles_per_thread_per_cpu, cpu_cycles_per_thread, etc. are pulling from _cpu_freq_per_thread, which is backed by the table _cpu_freq_per_thread_no_id. So if your concern is that we are going to the sched table for each stdlib table we create, i don't think we're doing that with the current verison of the PR unless i'm missing something.

LalitMaganti: > So if your concern is that we are going to the sched table for each stdlib table we create, i don't think we're doing that with the current verison of the PR unless i'm missing something.

My problem is not with the sched table per se but iterating over O(sched) rows. Doing that multiple times is expensive: is someone needs both per-thread/cpu and per-thread breakdowns, it's going to be better if the per-thread/cpu does the first layer of aggregation reducing the row count to O(threads * cpus) << O(sched) and then do an aggregation to O(thread) or O(cpu).

It's a bit of a micro-optimization but on large traces, it's non-trivial speed up.

cphlipot1: I noticed a few issues while working on this, some of which you made me aware of. I put up 3 PRs to address, and will probably base this PR on top of those changes assuming you are ok with them.

Regarding your suggestion to use **cpu_cycles_per_thread_per_cpu** to compute **cpu_cycles_per_thread**, **cpu_cycles_per_thread_per_cpu** is going to be an intermediate aggregation and thus there will be small precision loss in cpu_cycles_per_thread which will mean the expected values for that existing diff test will need to be updated.

Just want to make sure you are ok with that precision loss. If yes, i can go ahead with this change. i will probably loop back to this PR next week.

LalitMaganti: > Just want to make sure you are ok with that precision loss.

I think small precision loss is fine. As-is, from my comment on other change, we are intentionally loosing a little precision to prevent overflows so it's already not absolutely accurate.

cphlipot1:
comments should be addressed now. Also updated base branch to top of my stack of other fixes since this PR now touches existing stdlib implementations instead of just creating new ones.

LalitMaganti (approved):

PR #2756, opened 2025-08-29
stdlib: fix cpu utilization/cycles intermediate truncation

cphlipot1:
some of the stdlib routines are doing intermediate rounding which
causes less accurate final results to be produced. to minimize the
impact of this, we make two changes:
- pull "/ 1000" outside of SUM() aggregations so we aggregate first
 and then do the truncating divide.
- when "/ 1000" is in the numerator and denominator, remove it from
 both.

This does not fix all inaccuracies. as you can see in diff tests
there are still some cases where the inequality
min_freq <= avg_freq <= max_freq does not hold, which is still unexpected.


src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/process.sql
```diff
@@ -102,12 +102,12 @@ CREATE PERFETTO TABLE cpu_cycles_per_process (
 ) AS
 SELECT
   upid,
-  cast_int!(SUM(dur * freq / 1000)) AS millicycles,
-  cast_int!(SUM(dur * freq / 1000) / 1e9) AS megacycles,
+  cast_int!(SUM(dur * freq) / 1000) AS millicycles,
```
LalitMaganti: So actually doing this was a very intentional choice. On large traces, sum(dur * freq) will overflow int64. We should keep it as-is.

cphlipot1: Thanks for clarifying. I was aware this would increase the likelihood of overflow but didn't realize it would impact real-world traces.

I think the is opportunity to improve the logic here for instance dividing by the GCD (greatest common divisor) of all frequencies instead of a hard-coded 1000, to avoid truncation since it is often going to be >= 1000 anyways, but will avoid including any such change for now, at least in this PR.

I have reverted the aspect of the change related to SUM(dur * freq / 1000) , but kept the SUM(dur) /1000 change since SUM(dur) should never overflow before the numerator, and even independent of this expression should not overflow on any practically sized trace.

LalitMaganti: Improving to divide by gcd makes a lot of sense, would be in favour of that!

PR #2762, opened 2025-08-29
tp: perf_text: document clock assumptions

rsavitski:
I'd like us to be pedantic when guessing/assuming clocks going forward.
Here's my interpretation for "3abf62f957 tp: perform trace time
conversion, assuming MONOTONIC for perf text (#2759)"


LalitMaganti:
Thanks for fixing.

FYI though, this code will change once I'm back from conference: I'll need to sit down and properly annotate all places where we assume a clock (and mark something in a TP table) and then flag them in the UI when merging clocks, prompting the user to check that it makes sense.

LalitMaganti (approved):

PR #2764, opened 2025-08-29
tp: StructuredQuery SQL doesn't require cols anymore

aMayzner:
The fact that StructuredQuery required columns was making it very difficult to create Sql structured queries from Explore Page - we didn't know programatically what are the columns that user provided.

Not providing columns will just make you select all columns (`*`).

### DIFF
--- a/protos/perfetto/perfetto_sql/structured_query.proto
+++ b/protos/perfetto/perfetto_sql/structured_query.proto
@@ -113,7 +113,7 @@ message PerfettoSqlStructuredQuery {
     // results with at least one column.
     optional string sql = 1;
 
-    // The name of columns which will be returned by the SQL. Required.
+    // The name of columns which will be returned by the SQL. Optional.
     repeated string column_names = 2;
 
     // A list of dependencies for this SQL query. Optional.
--- a/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
+++ b/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
@@ -338,17 +338,18 @@ base::StatusOr<std::string> GeneratorImpl::SqlSource(
     }
   }
 
+  std::string cols_str = "";
   if (sql.column_names()->size() == 0) {
-    return base::ErrStatus("Sql must specify columns");
+    cols_str = "*";
+  } else {
+    std::vector<std::string> cols;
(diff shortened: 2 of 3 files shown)


protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -113,7 +113,7 @@ message PerfettoSqlStructuredQuery {
     // results with at least one column.
     optional string sql = 1;
 
-    // The name of columns which will be returned by the SQL. Required.
+    // The name of columns which will be returned by the SQL. Optional.
```
LalitMaganti, line 116: If they are specified we should validate they match the returned ones from SQL.

LalitMaganti (dismissed):

protos/perfetto/perfetto_sql/structured_query.proto
aMayzner: To validate we need to execute the preambles, so we can't verify this on the level of structured query generator. The caller can verify this, of course.

LalitMaganti: Sure the caller should verify. Please implement that.

aMayzner: I will, when I actually work on the caller

aMayzner: Done in summary.cc

src/trace_processor/trace_summary/summary.cc
```diff
@@ -428,6 +430,35 @@ base::Status CreateQueriesAndComputeMetrics(
...
+        }
+        if (!std::includes(
+                actual_column_names.begin(), actual_column_names.end(),
+                expected_column_names.begin(), expected_column_names.end())) {
+          std::vector<std::string> expected_vec(expected_column_names.begin(),
+                                                expected_column_names.end());
+          std::vector<std::string> actual_vec(actual_column_names.begin(),
+                                              actual_column_names.end());
+          return base::ErrStatus(
+              "Column names mismatch for metric bundle '%s'. Expected: [%s], "
```
sashwinbalaji: nit: since we are checking for subset only term "mismatch" might confuse user. might be better to say something like not all of expected [%s] in actual [%s]

PR #2765, opened 2025-08-29
tp: AnalyzeStructuredQuery gets columns of the query

aMayzner:
We need to know the resulting columns of the query for Explore Page but we don't want to have to require user to specify them. 

To do it we need to execute preambles and modules, but it is not a particularily bad behavior. We can split it into multiple functions if it starts to be an issue.

### DIFF
--- a/src/trace_processor/trace_processor_impl.cc
+++ b/src/trace_processor/trace_processor_impl.cc
@@ -761,11 +761,11 @@ base::Status TraceProcessorImpl::AnalyzeStructuredQueries(
         auto last_stmt,
         engine_->PrepareSqliteStatement(
             SqlSource::FromTraceProcessorImplementation(analyzed_sq.sql)));
-    auto sqlte_stmt = last_stmt.sqlite_stmt();
-    int col_count = sqlite3_column_count(sqlte_stmt);
+    auto sqlite_stmt = last_stmt.sqlite_stmt();
+    int col_count = sqlite3_column_count(sqlite_stmt);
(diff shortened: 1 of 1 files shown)


src/trace_processor/trace_processor_impl.cc
```diff
@@ -740,6 +740,31 @@ base::Status TraceProcessorImpl::AnalyzeStructuredQueries(
     analyzed_sq.modules = sqg.ComputeReferencedModules();
     analyzed_sq.preambles = sqg.ComputePreambles();
     sqg.AddQuery(sq.ptr, sq.size);
+
+    // Execute modules
```
LalitMaganti, line 744: Can you add some todo here indicating this should be moved to a empty engine.

src/trace_processor/trace_processor_impl.cc
```diff
@@ -740,6 +740,35 @@ base::Status TraceProcessorImpl::AnalyzeStructuredQueries(
...
+    for (const auto& preamble : analyzed_sq.preambles) {
+      engine_->Execute(SqlSource::FromTraceProcessorImplementation(preamble));
+    }
+
+    // Fetch columns
+    ASSIGN_OR_RETURN(
+        auto last_stmt,
+        engine_->PrepareSqliteStatement(
+            SqlSource::FromTraceProcessorImplementation(analyzed_sq.sql)));
+    auto sqlte_stmt = last_stmt.sqlite_stmt();
```
sashwinbalaji: typo?

PR #2771, opened 2025-08-29
Dev/chenghanzou/dev/chenghanzou/perfetto llm sql

chenghanzou-google:
Add a banner to the Query page that points to the Perfetto SQL agent for query generation. This banner is only visible to internal users (i.e. Googlers).

C6o9rxYTDv6LZsg

The hyperlink opens up the agent on a new tab, which includes an example prompt on how to generate queries:

BBFeCCQpvZHbZkC


chenghanzou-google:
Abandoned this CL because the committer was not set up correctly. Will create a new one once fixed.

PR #2772, opened 2025-08-29
Add a banner to the Query page that points to the Perfetto SQL agent for query generation

chenghanzou-google:
Add a banner to the Query page that points to the Perfetto SQL agent for query generation. This banner is only visible to internal users (i.e. Googlers).

C6o9rxYTDv6LZsg

The hyperlink opens up the agent on a new tab, which includes an example prompt on how to generate queries:

BBFeCCQpvZHbZkC

### DIFF
--- a/python/tools/check_imports.py
+++ b/python/tools/check_imports.py
@@ -94,6 +94,9 @@ DEPS_ALLOWLIST = [
     # The record plugin needs access to the wasm .d.ts for trace_config_utils.
     ('/plugins/dev.perfetto.RecordTraceV2/*', '/gen/trace_config_utils'),
 
+    # The query page plugin needs access to globals.isInternalUser.
+    ('/plugins/dev.perfetto.QueryPage/*', '/frontend/globals'),
+
     # Misc legitimate deps.
     ('/frontend/index', ['/gen/*']),
     ('/traceconv/index', '/gen/traceconv'),
--- a/ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
+++ b/ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
@@ -43,6 +43,7 @@ import {MenuItem, PopupMenu} from '../../widgets/menu';
 import {ResizeHandle} from '../../widgets/resize_handle';
 import {Stack, StackAuto} from '../../widgets/stack';
 import {Icon} from '../../widgets/icon';
+import {globals} from '../../frontend/globals';
 
 class CopyHelper {
   private _copied = false;
@@ -75,7 +76,9 @@ export interface QueryPageAttrs {
   readonly editorText: string;
   readonly executedQuery?: string;
   readonly queryResult?: QueryResponse;
+
   onEditorContentUpdate?(content: string): void;
+
   onExecute?(query: string): void;
 }
 
@@ -120,7 +123,10 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
           }),
           m(
             Stack,
-            {orientation: 'horizontal', className: 'pf-query-page__hotkeys'},
+            {
+              orientation: 'horizontal',
+              className: 'pf-query-page__hotkeys',
+            },
             'or press',
             m(HotkeyGlyphs, {hotkey: 'Mod+Enter'}),
           ),
@@ -132,6 +138,31 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
           }),
         ]),
       ]),
+      globals.isInternalUser &&
+        m(
+          Box,
+          m(Callout, {icon: 'star', intent: Intent.None}, [
+            'Try out the ',
+            m(
+              'a',
+              {
+                href: 'http://go/perfetto-sql-agent',
+                target: '_blank',
+              },
+              'Perfetto SQL Agent',
+            ),
+            ' to generate SQL queries and ',
+            m(
+              'a',
+              {
+                href: 'http://go/perfetto-llm-user-guide#report-issues',
+                target: '_blank',
+              },
+              'give feedback',
+            ),
+            '!',
+          ]),
+        ),
       attrs.editorText.includes('"') &&
         m(
           Box,


python/tools/check_imports.py
```diff
@@ -94,6 +94,9 @@
     # The record plugin needs access to the wasm .d.ts for trace_config_utils.
     ('/plugins/dev.perfetto.RecordTraceV2/*', '/gen/trace_config_utils'),
 
+    # The query page plugin needs access to globals.isInternalUser.
+    ('/plugins/dev.perfetto.QueryPage/*', '/frontend/globals'),
```
LalitMaganti: I don't think we want this. +Steve: what's the best way for feature detecting isInternalUser in a plugin?

stevegolton: There's a number of problems here:
- You should not be adding arbitrary exceptions to this file - we are trying to drive the number of one-offs like this down.
- We should not be adding new dependencies on `globals`.
- Philosophically speaking, I'm not sure whether plugins should actually know about `isInternalUser`.

I don't have a better suggestion yet on how to handle this, I'll need to think about it.

stevegolton (commented):
I'm sorry but I'm going to close this PR for the time being. I think this needs to be discussed within the team in more detail.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -132,6 +138,31 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
           }),
         ]),
       ]),
+      globals.isInternalUser &&
+        m(
+          Box,
+          m(Callout, {icon: 'star', intent: Intent.None}, [
+            'Try out the ',
+            m(
+              'a',
```
stevegolton: Please use the `Anchor` widget for links - it's a drop-in replacement for 'a' elements + use the `Icons.ExternalLink` icon to indicate that this is to an external link.

Something like this:
```
m(Anchor, {..., icon: Icons.ExternalLink}, ...),
```

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -132,6 +138,31 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
...
+              {
+                href: 'http://go/perfetto-sql-agent',
+                target: '_blank',
+              },
+              'Perfetto SQL Agent',
+            ),
+            ' to generate SQL queries and ',
+            m(
+              'a',
```
stevegolton: Same as above^^

python/tools/check_imports.py
stevegolton: After https://github.com/google/perfetto/pull/2793 lands you should be able to access isInternalUser on the trace object within query page below and you can remove this exception.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -132,6 +138,31 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
           }),
         ]),
       ]),
+      globals.isInternalUser &&
```
stevegolton: After https://github.com/google/perfetto/pull/2793 lands, globals is now gone but you will be able to access `isInternalUser` on the trace object: e.g. `attrs.trace.isInternalUser`, which would avoid having to import globals and add the exception in the import checker.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -132,6 +138,31 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
...
+        m(
+          Box,
+          m(Callout, {icon: 'star', intent: Intent.None}, [
```
stevegolton: I landed https://github.com/google/perfetto/pull/2783 today that allows configuring a dismiss / close button to dismiss a Callout I think this popup should be dismissable, and its dismissed state stored in localstorage so it doesn't appear again.

If you wanted to, maybe you could add an AI button to the left of the copy button on this page, so that the link to http://go/perfetto-sql-agent is still available even after the popup has been dismissed.

python/tools/check_imports.py
chenghanzou-google: Thank you both! I have reverted the change in this file.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
chenghanzou-google: Done.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
chenghanzou-google: Done.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
chenghanzou-google: Done.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -120,18 +124,68 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
...
+        !this.hidePerfettoSqlAgentBanner &&
+        m(
+          Box,
+          m(
+            Callout,
+            {
+              icon: 'star',
+              dismissable: true,
+              onDismiss: () => {
+                this.hidePerfettoSqlAgentBanner = true;
```
LalitMaganti: This should probably be hidden in LocalStorage as steve suggested.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
chenghanzou-google: Thank you for the suggestions! I have made the following changes:

1. Make the banner dismissable and hide it for the current session, i.e. it will come back after a page/trace load.
2. Add a button next to the copy button, so the user can easily open the agent when the banner is gone.

7mqznaTuQSByzmK

I didn't make the banner permanently gone when dismissed because I thought the button might not be very discoverable due to its style and location, and it would not be as effective as the banner in terms of drawing users' attention to use the agent. With the dismiss button, there is a mechanism for users to hide the banner if they want, so I feel overall the UI has a balance between discoverability and obtrusiveness. WDYT?

cc @LalitMaganti

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
chenghanzou-google: I posted a comment to Steve's thread just now. Let's continue the discussion over there.

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
stevegolton: > I didn't make the banner permanently gone when dismissed because I thought the button might not be very discoverable due to its style and location, and it would not be as effective as the banner in terms of drawing users' attention to use the agent. With the dismiss button, there is a mechanism for users to hide the banner if they want, so I feel overall the UI has a balance between discoverability and obtrusiveness. WDYT?

I would prefer that the banner be permanently gone when dismissed. If the banner re-appears every time I reopen the UI this feels too intrusive to me. I think the dismissable callout + permanent button is more than discoverable enough.

chenghanzou-google: OK, sounds good. Done.

stevegolton (approved):

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -93,18 +98,68 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
...
           }),
         ]),
       ]),
+      this.shouldDisplayPerfettoSqlAgentBanner(attrs) &&
+        m(
+          Box,
+          m(
+            Callout,
+            {
+              icon: 'star',
```
stevegolton: I just updated the material icons pack to the latest version, which includes a bunch more icons, if you wanted to use the [wand_stars](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:wand_stars:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=magic&icon.size=24&icon.color=%23e8eaed) icon or similar here?

chenghanzou-google: Thanks. Done.

LalitMaganti (approved):

PR #2803, opened 2025-09-02
Add the ability to represent fan-out and merge-points in the Perfetto protos.

tikurahul:
WIP: Add the ability to represent fan-out and merge-points in the Perfetto protos.

* More context in the design document at: go/perfetto-context-propagation
* This makes it possible to represent a true fan-out, and a true-fan-in trace event (with multiple parents)

```

                                                              ┌────┐
   ───────────────────────────────────────────────────────────┤time├───────────────────────────────────────────────────────────────▶
                                                              └────┘

         ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
         │                                                                                                             │
         │            ┌────────────────────┐ parent_span_ids: null                                                     │
─────────┤            │  Frontend.Request  │       span_id: 1                                                          │
         │            └────────────────────┘                                                                           │
         │                                                                                                             ├───────────□
         └───┳─────────────▲────────────┳───────────────────────────────────────────────────────────────────▲──────────┘
             ┃             ┃            ┃                                                                   │
             ┃             ┃            ┃                                                                   │
          ┌──▼─────────────┻─┐          ┃                                                                   │
          │Backend.call      │          ┃                                                                   │
          │                  │          ┃                                                                   │
    ──────┤                  ├──────────╋───────────────────────────────────────────────────────────────────┼──────────────▶
          │parent_span_ids: 1│          ┃                                                                   │
          │    span_id: 2    │          ┃                                                                   │
          └──────────────────┘          ┃                                                                   │
                                      ┌─▼───────────────────────────┐            ┌──────────────────────────┴──┐
                                      │Backend.DoSomething          │            │Backend.DoSomething.Finished │
                                      │                             │            │                             │
                                      │                             │            │                             │
                                      │          parent_span_ids: 1 │            │       parent_span_ids: 4,5  │
                                      │              span_id: 3     │            │            span_id: 6       │
                                      └──━━┳────────────────────────┘            └─────────────────────────────┘
                                           ┃                                                    ▲
                                           ┃                                                    │
                                           ┃  ┌──────────────────┐                              │
                                           ┃  │Helper.Call       │                              │
                                           ┃  │                  │                              │
                                           ┣─▶│                  │──────────────────────────────┤
                                           ┃  │parent_span_ids: 3│                              │
                                           ┃  │    span_id: 4    │                              │
                                           ┃  └──────────────────┘                              │
                                           ┃                                                    │
                                           ┃  ┌────────────────────────────────┐                │
                                           ┃  │Helper.Call                     │                │
                                           ┃  │                                │                │
                                           ┗━▶│                                │────────────────┘
                                              │      parent_span_ids: 3        │
                                              │          span_id: 5            │
                                              └────────────────────────────────┘
```

Test: Will add tests, once the shape of the protos are acceptable to the Perfetto team.


protos/perfetto/trace/track_event/track_event.proto
```diff
@@ -221,6 +221,13 @@ message TrackEvent {
   // consumption.
   repeated fixed64 flow_ids = 47;
 
+  // A span is a timed operation in a trace which can be comprised of children.
```
LalitMaganti, line 224: I don't think we want to use the word span here. Slice vs span is very confusing. We should come up with a name for this. Also we should add a note on what type of events this is supported on (since counters/instants also go through here): I guess this can only be specified on slice_begin events (should we support specifying on slice_end, I think no but just checking)

tikurahul: Do you have any thoughts on what a good name might be ? I can make it clearer that this information is only relevant in the context of `slice_begin` events. I don't think we need to do `slice_end`.

LalitMaganti: Sadly no :/ Maybe a brainstorming session with AI is needed to come up with this.

tikurahul: I did a quick brainstorming session and here are some interesting options:

## Alternatives to the word Span that I don't hate:

* `event_id` and `parent_event_ids` if we wanted to keep things generic.
* `continuation_id` and `parent_continuation_ids` is something that the AI did not come up, but something which is frequently used in similar contexts.
* `task_id` and `parent_task_ids`

https://g.co/gemini/share/c6a6c77fb709 has my prompt if you want to brainstorm further. 

WDYT?

LalitMaganti: See http://shortn/_RuBiArJtIe. 

End result is:

> 
> Ah yes, that's a much cleaner design! One fork/join operation per slice, and use nested slices for composition. That keeps the semantics simple and unambiguous.
> 
> So the validation would be:
> 
> A slice can have EITHER fork fields OR join fields, not both
> If you need both operations, create a nested slice
> protobuf
> // For fork operations: this slice's fork identifier
> oneof fork_id_field {
>   uint64 fork_id = XX;
>   uint64 fork_id_iid = XX;
> }
> 
> // For fork operations: the fork that spawned this slice
> oneof parent_fork_field {
>   uint64 parent_fork_id = XX;
>   uint64 parent_fork_id_iid = XX;
> }
> 
> // For join operations: this slice's join identifier
> oneof join_id_field {
>   uint64 join_id = XX;
>   uint64 join_id_iid = XX;
> }
> 
> // For join operations: the joins this slice is waiting for
> repeated uint64 parent_join_ids = XX;
> Example usage:
> 
> Slice A: fork_id=100
>   Slice B: parent_fork_id=100, join_id=200
>   Slice C: parent_fork_id=100, join_id=201
> Slice D: parent_join_ids=[200, 201]
>   Slice E: fork_id=300  // Nested slice for next fork
>     Slice F: parent_fork_id=300
> This enforces clear semantics:
> 
> Each slice has a single, well-defined role in the fork/join pattern
> Complex patterns are explicitly represented through nesting
> The UI can unambiguously interpret each slice's participation
> I think this constraint actually makes the system easier to understand and implement. The naming fork_id, parent_fork_id, join_id, parent_join_ids works well with this model!

tikurahul: I don't seem to have permissions to the full chat. But let me try and understand the proposed design.

PR #2804, opened 2025-09-02
Implement android.virtualization_framework data source

nioffe:
This PR adds new AndroidVirtualizationFrameworkDataSource which implements the corresponding AndroidVirtualziationFrameworkConfig proto

Bug: 222347044


protos/perfetto/config/data_source_config.proto
```diff
@@ -259,6 +260,10 @@ message DataSourceConfig {
   // Data source name: android.cpu_per_uid
   optional CpuPerUidConfig cpu_per_uid_config = 137 [lazy = true];
 
+  // Data source name: android.virtualization_framewoerk
```
rsavitski, line 263: typo

src/traced/service/service.cc
```diff
@@ -207,6 +207,7 @@ int PERFETTO_EXPORT_ENTRYPOINT ServiceMain(int argc, char** argv) {
     PERFETTO_ELOG("Failed to start the traced service");
     return 1;
   }
+  PERFETTO_ELOG("[ioffe] welcome to traced!");
```
rsavitski, line 210: welcome!

src/traced/probes/android_virtualization_framework/android_virtualization_framework_data_source.cc
```diff
@@ -0,0 +1,97 @@
...
+
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+#include <sys/system_properties.h>
+#endif
+
+namespace perfetto {
+
+namespace {
+
+static bool SetAndroidSysProp(const std::string& name,
```
rsavitski, line 33: there should already be a helper for this somewhere in the codebase

protos/perfetto/config/android/android_virtualization_framework_config.proto
```diff
@@ -0,0 +1,35 @@
...
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Custom configuration for the "android.virtualization_framework" data source.
+// Next Id: 2
+message AndroidVirtualizationFrameworkConfig {
+  // Tracing configuration for a single VM managed by the Android Virtualization
+  // Framework.
```
rsavitski, line 25: Similarly, in these comments: the repeated full spelling of AVF is hurting readability (certain ARM manuals come to mind...). And the comments are adding very little information in general. 

Can we instead focus on describing what this does when enabled?

protos/perfetto/config/android/android_virtualization_framework_config.proto
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Custom configuration for the "android.virtualization_framework" data source.
+// Next Id: 2
+message AndroidVirtualizationFrameworkConfig {
```
rsavitski, line 23: General point: while being pedantic is often useful, is there really no shorter name for this than "AndroidVirtualizationFramework"? This also extends to the C++ parts since both types and filenames (e.g. android_virtualization_framework_data_source.h) are so long that they hurt readability.

Is this specific to microdroid? Should we name it after that? Or shorten AVF in specific places? What do humans call AVF when they don't want to pronounce the whole word soup?

protos/perfetto/config/android/android_virtualization_framework_config.proto
```diff
@@ -0,0 +1,35 @@
...
+  // Tracing configuration for a single VM managed by the Android Virtualization
+  // Framework.
+  // Next Id: 2
+  message AndroidVmConfig {
```
rsavitski, line 27: Are we expecting more options here in the near future? Otherwise consider flattening this to a single repeated field of names.

src/traced/probes/android_virtualization_framework/android_virtualization_framework_data_source.cc
```diff
@@ -0,0 +1,97 @@
...
+    if (!SetAndroidSysProp(prop_name, "0")) {
+      PERFETTO_ELOG("Failed to stop traced_relay for VM %s",
+                    vm_config.name_.c_str());
+    }
+  }
+}
+
+void AndroidVirtualizationFrameworkDataSource::Start() {
+  for (const auto& vm_config : vm_configs_) {
+    PERFETTO_LOG("Starting traced_relay for VM %s", vm_config.name_.c_str());
```
rsavitski, line 82: We prefer not to log at the data source activation granularity, I would downgrade this to a PERFETTO_DLOG.

src/traced/probes/android_virtualization_framework/android_virtualization_framework_data_source.cc
```diff
@@ -0,0 +1,97 @@
...
+    std::string prop_name =
+        "persist.avf_vm.traced_relay.enable." + vm_config.name_;
+    if (!SetAndroidSysProp(prop_name, "1")) {
```
rsavitski, line 85: There should be some handling of concurrent tracing sessions (or even multiple data sources in one tracing config) that name the same VM.

Either:
* do the necessary tracking to keep the sysprop set as long as at least one config is requesting the relay.
* fail starting the data source if there's already one active.

I'm a bit unclear on how much is planned for this data source, if it's only ever going to be used to start the relay, then the first option sounds preferable.

src/traced/probes/android_virtualization_framework/android_virtualization_framework_data_source.cc
```diff
@@ -0,0 +1,97 @@
...
+        it->as_bytes());
+    vm_configs_.emplace_back(vm_config.name().ToStdString());
+  }
+}
+
+AndroidVirtualizationFrameworkDataSource::
+    ~AndroidVirtualizationFrameworkDataSource() {
+  for (const auto& vm_config : vm_configs_) {
+    std::string prop_name =
+        "persist.avf_vm.traced_relay.enable." + vm_config.name_;
```
rsavitski, line 72: Do we get much from this being a "persist" sysprop? Consider if traced_probes crashes, this won't be cleaned up even after a reboot.

We do have a flag for cleanup in case of a crash, but I'm not sure if this fits: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/perfetto.rc;l=50;drc=694d0519ffe5284fd0461f7a0476c00335c3061c

rsavitski (changes requested):

PR #2839, opened 2025-09-04
stdlib: Add suspend-aware CPU utilization metrics

gignat-dev:
This change introduces functions to calculate CPU utilization, accounting for potential suspend periods by using the monotonic clock for computation of durations.

Added 'awake_runtime' column to the result of the functions:
	- 'cpu_cycles_in_interval'
	- 'cpu_cycles_per_thread_in_interval'
	- 'cpu_cycles_per_process_in_interval' This new column measures runtime using 'sum(to_monotonic(ii.ts + ii.dur)- to_monotonic(ii.ts))', ensuring that time elapsed during system suspend is not included in the duration.

Introduced new utilization functions:
1. 'cpu_utilization_in_interval': all CPUs utilization over given interval;
2. 'cpu_thread_utilization_in_interval': per thread_name utilization over given interval;
3. 'cpu_process_utilization_in_interval': per process_name utilization over given interval;

These new functions consume the 'awake_runtime' from the modified functions from above. The output is:
	- ['thread_name'|'process_name']: for the thread/process functions
	- 'awake_dur': Total runtime while 'awake' (CPUs not suspended)
	- 'awake_utilization': Percentage of 'awake_dur' over the 'awake' duration of the given interval, normalized by the number of CPUs. Values in [0.0, 100.0]
	- 'awake_unnormalized_utilization': unnormalized percentage. Values in [0.0, 100.0 * <number_of_cpus>]

Added new test cases to
`test/trace_processor/diff_tests/stdlib/linux/cpu.py` for each new functions.
Updated existing test cases for the modified functions to validate the new 'awake_runtime' column.

Test:
tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell --name-filter='.*_utilization_in_interval.*'

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/process.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/process.sql
@@ -132,6 +132,8 @@ RETURNS TABLE (
   megacycles LONG,
   -- Total runtime duration
   runtime LONG,
+  -- Total runtime duration, while 'awake' (CPUs not suspended).
+  awake_runtime LONG,
   -- Minimum CPU frequency in kHz
   min_freq LONG,
   -- Maximum CPU frequency in kHz
@@ -156,6 +158,7 @@ SELECT
   cast_int!(SUM(ii.dur * freq / 1000)) AS millicycles,
   cast_int!(SUM(ii.dur * freq / 1000) / 1e9) AS megacycles,
   sum(ii.dur) AS runtime,
+  sum(to_monotonic(ii.ts + ii.dur) - to_monotonic(ii.ts)) AS awake_runtime,
   min(freq) AS min_freq,
   max(freq) AS max_freq,
   cast_int!(SUM((ii.dur * freq / 1000)) / SUM(ii.dur / 1000)) AS avg_freq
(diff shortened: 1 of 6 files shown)


src/trace_processor/perfetto_sql/stdlib/linux/cpu/utilization/process.sql
```diff
@@ -164,3 +167,49 @@ JOIN threads_counters
   USING (id)
 GROUP BY
   upid;
+
+-- Returns a table with process utilization over a given interval.
```
LalitMaganti, line 174: Please add a note here that this function is only intended for use for calling over a small set of intervals. Since you're here, please also add this to all similar functions.

gignat-dev: Ok, will do that

gignat-dev: Actually: what I think you refer to is the interval to be of small size, right?

LalitMaganti: No I mean if you join this function with another table, the other table cannot have a lot of rows of this function will fall apart in performance.

gignat-dev: Oh
I get it now
Thanks

PR #2841, opened 2025-09-04
traced_probes: support polling Adreno gpu frequency in SysStatsDataSource

rsavitski:
Nothing fancy as with other gpu frequency polling, assume a single gpu and a hardcoded path. Zhu Li helpfully tested on an XR device.

To enable, set "gpufreq_period_ms" in "linux.sys_stats" data source config.

Co-authored-by: Zhu Li <imzhu@google.com>
Co-authored-by: Ryan Savitski <rsavitski@google.com>


src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -406,16 +406,24 @@ std::optional<uint64_t> SysStatsDataSource::ReadAMDGpuFreq() {
 }
 
 void SysStatsDataSource::ReadGpuFrequency(protos::pbzero::SysStats* sys_stats) {
```
sashwinbalaji, line 408: LGTM but as optimisation would it be better if we check which is the right sysfs at first call and then read only that one in subsequent calls to avoid 2 syfs calls?

sashwinbalaji (approved):

src/traced/probes/sys_stats/sys_stats_data_source.cc
rsavitski: Same thought crossed my mind, but for now I'm keeping things super-simple. The polling frequency should be low enough that a couple of failed filepath lookups are negligible.

PR #2842, opened 2025-09-04
Add optional compression of processed traces after computation

david-keyes:
Add ability to enable zlib compression of processed traces on summarization from the TraceSummaryOutputSpec.

This allows consumers to receive a compressed output for those traces that will be stored or transmitted and need to memory consumption is required.


src/trace_processor/trace_summary/summary.cc
```diff
@@ -16,6 +16,7 @@
 
 #include "src/trace_processor/trace_summary/summary.h"
 
+#include <zlib.h>
```
LalitMaganti: zlib is not necessarily available in all configurations of Perfetto. Please ifdef with 
`#if PERFETTO_BUILDFLAG(PERFETTO_ZLIB)`

src/trace_processor/trace_summary/summary.cc
```diff
@@ -577,6 +578,20 @@ base::Status CreateQueriesAndComputeMetrics(
       *output = std::vector<uint8_t>(out.begin(), out.end());
       break;
   }
+
+  if (output_spec.compression == TraceSummaryOutputSpec::Compression::kZlib &&
```
LalitMaganti, line 587: This should be wrapped with `if constexpr (IsGzipSupported())`.

src/trace_processor/trace_summary/summary_integrationtest.cc
```diff
@@ -895,5 +913,46 @@ TEST_F(TraceSummaryTest, TemplateSpecWithValueColumnsAndSpecsError) {
                         "value_column_specs defined"));
 }
 
+TEST_F(TraceSummaryTest, OutputIsCompressed) {
```
LalitMaganti, line 920: THis shouldbe if-deffed as well.

src/trace_processor/trace_summary/summary_integrationtest.cc
```diff
@@ -16,6 +16,7 @@
 
 #include "src/trace_processor/trace_summary/summary.h"
 
+#include <zlib.h>
```
LalitMaganti: Should be ifdefed.

src/trace_processor/trace_summary/summary.cc
LalitMaganti: Please use if constexpr rather than #if. Also the else branch should return an error stating that zlib is not included in trace processor.

src/trace_processor/trace_summary/summary_integrationtest.cc
david-keyes: added

src/trace_processor/trace_summary/summary_integrationtest.cc
david-keyes: added

src/trace_processor/trace_summary/summary.cc
david-keyes: Attempted `if constexpr` which fails, which appears to be related to the functions not existing but the types still needing to work.

Used ifdef, but wrapped inside the compression method check. If it's determined that compression was enabled via output spec but zlib is unavailable, then an error will be returned. 

Please take focus on the test case wrapping which includes if/else def to confirm the behavior.

src/trace_processor/trace_summary/summary.cc
david-keyes: Thanks for the detailed note, applied and updated test case for both scenarios.

LalitMaganti (approved):

PR #2852, opened 2025-09-05
ftrace: Add tracefs options support

sashwinbalaji:


LalitMaganti:
Should this be tied to exclusive mode?

protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -295,6 +295,28 @@ message FtraceConfig {
...
+  //
+  // Introduced in: perfetto v52.
+  // Supported on: Android 25Q3+.
+  message TracefsOption {
+    enum State {
+      STATE_UNSET = 0;
+      STATE_ENABLED = 1;
+      STATE_DISABLED = 2;
+    }
+    // Note: The string must exactly match (including case) the sysfs file name.
```
rsavitski: nit: this is not sysfs, it's tracefs. And I don't think we need to describe it in terms of filepaths, just say that this is the name of the tracefs option as found in tracefs/trace_options (without the "no" prefix).

protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -295,6 +295,28 @@ message FtraceConfig {
...
+  // Options control the tracers or the trace output.
+  // They let the user control the amount of data that is displayed in the
+  // trace and how a tracer or events work (stack traces, timestamps, etc).
+  // Interacts with trace option in /sys/kernel/tracing/options directory.
+  //
+  // Introduced in: perfetto v52.
+  // Supported on: Android 25Q3+.
+  message TracefsOption {
+    enum State {
+      STATE_UNSET = 0;
```
rsavitski: zeroeth option should be an UNKNOWN sentinel value (see the plethora of proto docs on the topic).

Even though we generate our own protozero / protogen encoding & decoding code, we should still try to conform to proto best practices.

src/traced/probes/ftrace/ftrace_config_muxer.cc
```diff
@@ -463,6 +463,45 @@ bool FtraceConfigMuxer::SetupConfig(FtraceConfigId id,
...
 
+  if (!request.tracefs_options().empty()) {
+    base::FlatHashMap<std::string, bool> current_tracefs_options;
+    for (const auto& tracefs_option : request.tracefs_options()) {
+      // Skip unset options.
+      if (tracefs_option.state() == FtraceConfig::TracefsOption::STATE_UNSET) {
+        continue;
+      }
+      // Validate the option name.
+      if (tracefs_option.name().empty() ||
```
rsavitski: I feel like this could be laid out in fewer source lines, for example `tracefs_option.name()` is repeated 9 times, an opportunity for a shorter const reference?

Maybe the std::all_of is better as a named function in the anonymous namespace?

protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -295,6 +295,28 @@ message FtraceConfig {
   // Supported on: Android 25Q3+.
   repeated uint32 tids_to_trace = 35;
 
+  // Options control the tracers or the trace output.
+  // They let the user control the amount of data that is displayed in the
+  // trace and how a tracer or events work (stack traces, timestamps, etc).
+  // Interacts with trace option in /sys/kernel/tracing/options directory.
```
rsavitski: I feel like this comment is misleading for most users, since it's documenting things from the perspective of someone using the textual tracefs (./trace) output.

The majority of options have no bearing on perfetto. Should we list the subset that the users /might/ want to set?

CHANGELOG
```diff
@@ -3,6 +3,7 @@ Unreleased:
     * Added support for advanced single-tenant features in ftrace data source.
       These features can only be used by a single tracing session.
     * Added support for ftrace filtering by TID as an exclusive feature.
+    * Added support for tracefs Options as an exclusive feature.
```
rsavitski: Why the capital O?

src/traced/probes/ftrace/tracefs.cc
```diff
@@ -293,6 +293,20 @@ bool Tracefs::ClearEventTidFilter() {
   return ClearFile(path);
 }
 
+std::optional<bool> Tracefs::GetTracefsOption(const std::string& option) {
+  std::string path = root_ + "options/" + option;
+  auto value = base::TrimWhitespace(ReadFileIntoString(path));
```
rsavitski: nit: bad place to use `auto` since the type is not obvious (there are many string-like types), and the correctness of lines like `value == "1"` depends on the implicit conversions due to this type being std::string.

protos/perfetto/config/ftrace/ftrace_config.proto
```diff
@@ -290,10 +290,43 @@ message FtraceConfig {
...
+  // - record-cmd: To enable mapping the TID to its command name.
+  // - record-tgid: To enable mapping the TID to its Thread Group ID(TGID) .
+  // - Options for function_graph tracer (funcgraph-proc, funcgraph-irqs,
+  // graph-time, etc.)
+  //
+  // Note: This is an exclusive feature, see:
+  // https://perfetto.dev/docs/learning-more/android#exclusive-tracing-sessions.
+  //
+  // Introduced in: perfetto v52.
+  // Supported on: Android 25Q3+.
```
rsavitski:
```suggestion
  // Tracefs options to set directly in the tracefs instance. This is a very
  // niche feature since almost all of the options deal with formatting textual
  // output (the /trace file), which perfetto doesn't use.
  // 
  // The options with a known use-case:
  // * event-fork: when using `tids_to_trace` above, the kernel will
  //   automatically add newly spawned descendant threads to the set of TIDs.
  // 
  // Full list of options is available at
  // https://docs.kernel.org/trace/ftrace.html#trace-options.
  //
  // Note: this is an exclusive feature, see:
  // https://perfetto.dev/docs/learning-more/android#exclusive-tracing-sessions.
  //
  // Introduced in: perfetto v52.
  // Supported on: Android 25Q3+.
```

rsavitski (changes requested):

sashwinbalaji:
go/pgh/2878 rebased on top of main

PR #2853, opened 2025-09-05
tracefs: Add tracing_cpumask support

sashwinbalaji:


LalitMaganti:
Shouldn't this also be exclusive mode gated?

sashwinbalaji:
https://github.com/google/perfetto/pull/2853/files#diff-37b1e14bd902f1a04da838d9413bef53f8958b2e8285d03a0f5e5f6742ad183cR171

should handle the exclusivity

LalitMaganti:
> https://github.com/google/perfetto/pull/2853/files#diff-37b1e14bd902f1a04da838d9413bef53f8958b2e8285d03a0f5e5f6742ad183cR171
> 
> should handle the exclusivity

Ok yes. But there's a bit of a problem in how the docs are written right now: the comments on the proto may not be read by everyone because e.g. they don't show up except if you're reading the actual proto. 

If you're looking at docs.perfetro.dev trace config page, the fact that it's exclusive mode would be missed

Can we add a comment on each exclusive mode option saying it is so and linking to what that means?

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -1330,6 +1330,13 @@ message FtraceConfig {
   }
   repeated TracefsOption tracefs_options = 36;
 
+  // This is a mask that lets the user only trace on specified CPUs.
+  // The format is a hex string representing the CPUs.
```
rsavitski: Please write the documentation you'd like to read as a user. "hex string" isn't really helpful. Give a non-trivial example or link to a kernel reference on this.

For example, I found it surprising that this works: `ffffffff`. But this doesn't: `fffffffff`. The format actually requires us to put commas as delimiters.

CHANGELOG
```diff
@@ -1,9 +1,9 @@
 Unreleased:
   Tracing service and probes:
-    * Added support for advanced single-tenant features in ftrace data source.
+    * Added support for `advanced single-tenant` features in ftrace data source.
       These features can only be used by a single tracing session.
-    * Added support for ftrace filtering by TID as an advanced feature.
-    * Added support for tracefs Options as an advanced feature.
+    * Added support for `tracing_cpumask`, `tracefs Options`, `ftrace filtering
```
rsavitski: None of the backticks make sense? none of this is a code literal.

protos/perfetto/trace/perfetto_trace.proto
sashwinbalaji: Sure, will update it

This is a direct copy paste of kernel doc though: https://www.kernel.org/doc/Documentation/trace/ftrace.txt

rsavitski: And did you find the kernel doc useful? I seem to remember us having to go to the source code to figure out the specifics yesterday :--)

rsavitski: My more general point is: if we're going through the trouble of exposing these super-niche things at our proto configuration level, the least we can do is provide useful-to-humans guidance on how/when to use the options. 
It's better if one of us spends a bit of time figuring out what & how, and then writes it down. Instead of asking every user to understand what exactly a cpumask hex string looks like.

sashwinbalaji: Agreed, will do so.

rsavitski: Thanks, this is much better.

CHANGELOG
```diff
@@ -2,8 +2,8 @@ Unreleased:
   Tracing service and probes:
     * Added support for advanced single-tenant features in ftrace data source.
       These features can only be used by a single tracing session.
-    * Added support for ftrace filtering by TID as an exclusive feature.
-    * Added support for tracefs Options as an exclusive feature.
+    * Added support for tracing_cpumask, tracefs Options, and ftrace filtering
+      by TID as advanced features.
```
rsavitski, line 6: nit: this needs to integrate the earlier changes (exclusive instead of advanced and fixed capitalisation).

sashwinbalaji:
abandoned in favor of 2885

PR #2857, opened 2025-09-05
stdlib: Add support for Wattson generic policy dependencies

wusamuel6:
Adds support so that policies can arbitrarily vote on each other. For
example, policyX and affect policyY and policyY can affect policyX with
no restrictions like X < Y. Prior to this patch, there were constraints
on how policies depended on each other.

As a consequence of generic policy dependencies, there are a few other
changes that need to go hand-in-hand.
  - Indicated both freq and policy in the LUTs instead of derived value
    mapped to the two
  - Generic static power calculations (i.e. static power from policy0,
    policy1, ..., policyN are accounted for) without restrictions

The majority of the functionality in this patch comes in the form of the
_w_dependent_cpus_calc table in pivot.sql, which replaces the entirety
of w_cpu_dependence.sql file. The table operation is described as such:
  1. Unpivot only necessary columns for CPU dependence calculation.
     Unpivot only necessary rows for CPU dependence calculation, as
     defined by the CROSS JOIN table.
  2. For each CPU, find the dependent CPUs with the highest power vote,
     and log the frequency and policy of that voter.
  3. Pivot and re-JOIN the calculated dependent data back into the
     table.

After this patch series, there is an increase in query compute time, shown here:
- WattsonStdlib:wattson_dsu_devfreq_system_state: 5717.91ms -> 5984.51ms
- AndroidMetrics:wattson_atrace_apps_threads_output: 3981.77ms -> 5114.24ms
- WattsonStdlib:wattson_idle_attribution: 3395.53ms -> 3492.97ms
- AndroidMetrics:wattson_app_startup_threads_output: 2745.09ms -> 3539.86ms

Test: tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell --name-filter '.*wattson*' --print-slowest-tests
Bug: 443097308


src/trace_processor/perfetto_sql/stdlib/wattson/cpu/arm_dsu.sql
```diff
@@ -62,3 +66,53 @@ CREATE VIRTUAL TABLE _arm_l3_rates USING SPAN_OUTER_JOIN (
...
+        FROM metadata
+        WHERE
+          name = 'android_guest_soc_model'
+      ) IN (
+        SELECT
+          device
+        FROM _use_devfreq
+      )
+      AND (
+        SELECT
```
LalitMaganti, line 99: Generally more efficient to do exists + `limit 1`.

LalitMaganti (approved):

PR #2859, opened 2025-09-06
ui: Add track settings infrastructure & bulk track settings edit menu

stevegolton:
Currently tracks own their settings completely, which is clean encapsulation but makes bulk operations and global defaults impossible. 

This patch adds infrastructure for 'track settings' where each track can advertise a list of settings that it supports, including how to render the menu items in so that the user can modify those settings, and also getters and settings to get and modify the setting values externally.

This patch is built with bulk setting modification in mind, as each settings has a descriptor associated with it, and multiple tracks can point to the same descriptor. When multiple tracks are selected, the list of common settings supported by all selected tracks is generated (those that point to the same descriptor) are rendered in the bulk track dropdown menu in the header panel of the timeline (see screenshot).

image


LalitMaganti:
I'm not a huge fan of the UX here unfortunately.

1) as a user, I would not really think that the button in the panel there would contain contextual actions for the selection. I would not think to click there for bulk track editing
2) we should have a not somewhere in the menu about "n tracks selected" similar to the style of android when you do multiselect just so it's clear it's operating on selected tracks
3) the menu itself does not say anywhere that these are per track settings. I think folks would be confused just looking at the menu unless they deeply understood how track settings work.

stevegolton:
> I'm not a huge fan of the UX here unfortunately.
> 
> 1. as a user, I would not really think that the button in the panel there would contain contextual actions for the selection. I would not think to click there for bulk track editing
> 2. we should have a not somewhere in the menu about "n tracks selected" similar to the style of android when you do multiselect just so it's clear it's operating on selected tracks
> 3. the menu itself does not say anywhere that these are per track settings. I think folks would be confused just looking at the menu unless they deeply understood how track settings work.

Thanks for the review @LalitMaganti  - these are all entirely valid points.

To address 1 and 2, I've reworked the timeline toolbar to include a 'multiselect chip' when multiple tracks are selected. Clicking it brings up a drop down menu for operating on the tracks in bulk.

To address 3 I've added a title about the settings menu items and also a little callout to make it more obvious.

See the screenshot below:

Let me know what you think. We might need to iterate on this a little.

image

LalitMaganti:
This is so much better thanks a lot for the iterarion! 

Why is the copy/copy and switch not under bulk track settings though? Isn't that also applying to the selected tracks?

stevegolton:
> This is so much better thanks a lot for the iterarion!
> 
> Why is the copy/copy and switch not under bulk track settings though? Isn't that also applying to the selected tracks?

In my head the workspace operations and the track settings are two distinct groups of operations. Perhaps I just put the callout in the wrong place, as it's applicable to the entire menu. Maybe it should go at the very top of the menu above `WORKSPACE` instead?

LalitMaganti:
> Maybe it should go at the very top of the menu above WORKSPACE instead?

Yes I think this is the main issue.

stevegolton:
> > Maybe it should go at the very top of the menu above WORKSPACE instead?
> 
> Yes I think this is the main issue.

Here we go!

image

LalitMaganti:
Thanks this is fantastic!

LalitMaganti (approved):

PR #2860, opened 2025-09-07
ui: Clean up how font styles are injected for screenshot tests

stevegolton:
In order to make screenshot tess more stable, playwright needs to disable font kerning & smoothing. However this is currently done with the search param `?testing=1` which tells the UI it's in testing mode and it applies the styles internally.

The problem is that this is purely testing code, and it would be a lot cleaner to avoid putting these styles into prod. Whatsmore, these styles could also leak out into the parent application when the UI is embedded.

This PR modifies the PerfettoTestHelper so that it injects these styles using playwright's built-in `page.addStyleTag(...)` function, which is a much cleaner approach as it means testing logic is less invasive and stays in the testing code.


stevegolton:
Superseded by #3679

PR #2863, opened 2025-09-08
tp: Allow columns transformation in StructuredQuery

aMayzner:
This change introduces the ability to define new columns in a StructuredQuery using SQL expressions.

A transformation field has been added to the SelectColumn message. This field contains a sql_expression which is used to compute the column's value. The alias field is mandatory when a transformation is used.


protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -294,12 +294,26 @@ message PerfettoSqlStructuredQuery {
...
+
+    message Transformation {
+      // The SQL expression for the transformation. Required.
+      optional string sql_expression = 1;
+      // The name of the module for a function in the expression. Optional.
+      optional string module_name = 2;
+    }
+    // If this is set, this column is a transformation. `column_name` is
+    // ignored and the SQL expression is used instead. The `alias` field
+    // must be set to name the resulting column.
```
LalitMaganti: Any reason for this complexity rather than relaxing column_name to also support being an expression? Keeping this flat would be both simpler to understand and scale a lot better. 

The module_name you might want to make a dependency at the structured query level to require arbitrary dependencies regardless of what it's for.

aMayzner: 1.  Regarding the nested transformation field: I opted for this more explicit structure because it makes the query's intent clearer and simplifies validation. It avoids overloading column_name and provides a dedicated place to add more transformation-specific options in the future. I really don't see using column_name for stuff like that, it would make it less clean what is happening. Plus, it's not really column_name.
2.  For module_name: I can see the pros of top level dependencies, but it's more difficult to read and to figure out is the dependency was really needed or not (when people copy paste the code). I kept it local to the transformation to keep the dependency declaration close to its usage, but I'm open to centralizing it if you really think it's cleaner.

LalitMaganti: 1) can we maybe add a new "column_expression" field which can either be a name or expression and deprecate column_name field. I don't see the reason to have two fields for this to be very honest, there's no distinction in sql and I don't think it's important enough to be distinct here
2) with what you have, the structure is fine but given 1) I think it should be moved to the top level for ease of use.

aMayzner: 1. Wouldn't column_expression be very confusing where the most common usage will continue to be just saying what column you want? column_expression: {ts}?
2. If we agree on 1. that you are proposing I can move it to top level

LalitMaganti: Call it column_name_or_expression if you want to be very clear. But pragmatically speaking column name is just an expression with the identity function. I think that is pretty easy to understand

aMayzner: Top level of the whole structured query or selected column?

LalitMaganti: Do you mean for the module dependency? I think top of the whole structured query personally.

aMayzner: I really would like to couple the module with why do we need the module when I think about it. That's what we do for table and I believe it's a good direction

LalitMaganti: Sure but how are you actually going to enforce it? How do you prevent leakage between expressions?

I think having the false sense of security is worse than being a bit more lax and actually being able to enforce.

aMayzner: We can't force it now, but we might in the future (if we keep it on this level).
I don't get what do you mean by leakage between expressions

LalitMaganti: I don't understand how you could possibly enforce this. 

Leakage between expressions: because expression 0 includes something, expression 1 also can use the included module. It's a problem we have in general with includes, this makes it worse. 

Also I'm a bit worried we are going a bit too much in being difficult for people to write these protos. It's already pretty tricky to write and this is just adding busywork.

aMayzner: Ok, I will make it top level

aMayzner: Do we want to deprecate them from table too?

LalitMaganti: Yeah I'd be ok with that.

protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -50,17 +50,15 @@ message PerfettoSqlStructuredQuery {
...
-    optional string module_name = 2;
-
     // The name of the columns of this table which will be used by this query.
     // Required.
     //
     // Note: specifying this field is *mandatory* at all times. In the future,
     // this may become option for public tables in the standard library.
     repeated string column_names = 3;
+
+    // DEPRECATED: Use PerfettoSqlStructuredQuery.referenced_modules instead.
```
sashwinbalaji, line 60: should we also add [deprecated = true] ?

PR #2865, opened 2025-09-08
base: Reland RTFutex for android

sashwinbalaji:
Bug: 443178555

### DIFF
--- a/gn/perfetto.gni
+++ b/gn/perfetto.gni
@@ -267,16 +267,15 @@ declare_args() {
       perfetto_build_standalone && !is_perfetto_build_generator &&
       defined(use_custom_libcxx) && use_custom_libcxx
 
-  # Enables the use of priority-inheritance mutexes via PTHREAD_PRIO_INHERIT.
+  # Enables the use of priority-inheritance mutexes via PTHREAD_PRIO_INHERIT
+  # or wrapper around PI Futexes (dependant on OS).
   # Note that on Android platform (non-standalone) builds this flag is ignored
-  # and the Android flag "use_rt_mutex" is used instead (perfetto_flags.aconfig)
+  # and Android flags "use_rt_mutex", "use_rt_futex" are used instead (perfetto_flags.aconfig)
   # This is disabled in chromium, because the BPF-sandbox allows PI-futex only
   # when a field-trial is enabled, which is incompatible with a build time flag.
   enable_perfetto_rt_mutex =
-      (!is_wasm &&
-       (((build_with_chromium && is_android) || perfetto_build_standalone) &&
-        (current_cpu == "x64" || current_cpu == "arm64"))) ||
-      is_perfetto_build_generator
+      !is_wasm && (perfetto_build_standalone || is_perfetto_build_generator) &&
+      !build_with_chromium
 
   # This flag is used for the migration of UnixTaskRunner -> LockFreeTaskRunner.
   # It determines whether MaybeLockFreeTaskRunner is backed by UnixTaskRunner
--- a/include/perfetto/ext/base/flags.h
+++ b/include/perfetto/ext/base/flags.h
@@ -33,7 +33,7 @@ namespace perfetto::base::flags {
 // The second argument is the default value of the flag in non-Android platform
 // contexts.
 //
-// Note: For rt_mutex and rt_futex, the source of truth for non android platform
+// Note: For rt_mutex and rt_futex, the source of truth for non-Android platform
 // is in rt_mutex.h
 #define PERFETTO_READ_ONLY_FLAGS(X)                                    \
   X(test_read_only_flag, NonAndroidPlatformDefault_FALSE)              \
--- a/include/perfetto/ext/base/flags.h
+++ b/include/perfetto/ext/base/flags.h
@@ -36,13 +36,12 @@ namespace perfetto::base::flags {
   X(test_read_only_flag, NonAndroidPlatformDefault_FALSE)              \
   X(use_murmur_hash_for_flat_hash_map, NonAndroidPlatformDefault_TRUE) \
   X(ftrace_clear_offline_cpus_only, NonAndroidPlatformDefault_TRUE)    \
-  X(use_rt_mutex, PERFETTO_BUILDFLAG(PERFETTO_ENABLE_RT_MUTEX)         \
-                      ? NonAndroidPlatformDefault_TRUE                 \
-                      : NonAndroidPlatformDefault_FALSE)               \
   X(use_lockfree_taskrunner,                                           \
     PERFETTO_BUILDFLAG(PERFETTO_ENABLE_LOCKFREE_TASKRUNNER)            \
         ? NonAndroidPlatformDefault_TRUE                               \
-        : NonAndroidPlatformDefault_FALSE)
+        : NonAndroidPlatformDefault_FALSE)                             \
+  X(use_rt_mutex, NonAndroidPlatformDefault_FALSE)                     \
+  X(use_rt_futex, NonAndroidPlatformDefault_FALSE)
 
 ////////////////////////////////////////////////////////////////////////////////
 //                                                                            //
--- a/include/perfetto/ext/base/flags.h
+++ b/include/perfetto/ext/base/flags.h
@@ -32,6 +32,9 @@ namespace perfetto::base::flags {
 // in `perfetto_flags.aconfig`.
 // The second argument is the default value of the flag in non-Android platform
 // contexts.
+//
+// Note: For rt_mutex and rt_futex, the source of truth for non android platform
+// is in rt_mutex.h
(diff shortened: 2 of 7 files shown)


primiano:
OH wait can we make a small tweak? 
Let's get rid of the GETTID_IS_FAST and all the tid caching logic. that's too much.

I think what you are doing is right, and doens't require that hackery anymore:

1. On android: use RtFUtex  (if the flag is on). add a comment that we do only android only because gettid is slower on linux/Glibc.
2. On Linux/mac: use RtPosixMutex

Also at this point you can get rid of the initial android flag, as we won't use it on android at all. Just leave the gni variable and PERFETTO_BUILDFLAG(PERFETTO_ENABLE_RT_MUTEX).
We shouldn't have a flag around that has no effect on android, as those flag make sense only on android

gn/perfetto.gni
```diff
@@ -272,9 +272,17 @@ declare_args() {
   # and the Android flag "use_rt_mutex" is used instead (perfetto_flags.aconfig)
   # This is disabled in chromium, because the BPF-sandbox allows PI-futex only
   # when a field-trial is enabled, which is incompatible with a build time flag.
-  enable_perfetto_rt_mutex =
-      !is_wasm && (perfetto_build_standalone || is_perfetto_build_generator) &&
-      !build_with_chromium
+  perfetto_mutex_impl = 0
```
primiano: i'd still keep "rt_" the name, maybe call it perfetto_rt_mutex_mode
because otherwise it (wrongly) suggests this affects any mutex in our codebase.
Also imho "impl" should be the name of a class that implements it.
imho mode makes it more clear that this is not just a boolean (otherwise you'd have called ENABLE_XXX like everything else)

gn/perfetto.gni
```diff
@@ -272,9 +272,17 @@ declare_args() {
...
+  if (!is_wasm && (perfetto_build_standalone || is_perfetto_build_generator) &&
+      !build_with_chromium) {
+    if (is_android) {
+      perfetto_mutex_impl = 2
```
primiano: add a comment explaining what the values mean (either inline or just above ?

include/perfetto/ext/base/rt_mutex.h
```diff
@@ -34,8 +34,13 @@
 #include "perfetto/ext/base/flags.h"
 #include "perfetto/public/compiler.h"
 
-#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID) || \
-    PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX) ||   \
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
```
primiano, line 62: add a comment saying we enable this only on android because relies on gettid() Being fast, which is not the case on Linux/glibc, where the pthread-based implementation is faster (which however we cannot use in android because has been introduced only in later api levels). also link the dlsym bug.
otherwise i might be tempted to redo the same thing in 6 months having forgotten the context

include/perfetto/ext/base/rt_mutex.h
```diff
@@ -50,10 +55,95 @@
 #include <pthread.h>
 #endif
 
+#if PERFETTO_HAS_RT_FUTEX()
+#include <unistd.h>  // For gettid().
```
primiano, line 90: honestly just include these two (unistd and pthread) if !OS_WINDOWS rather than being so over-picky. they are very common headers that every unix-based OS has.
keeping this way is more liekly to create compilation errors in future if somebody adds a dependency elsewhere in this file and then we turn a flag off

src/base/rt_mutex.cc
```diff
@@ -25,10 +25,42 @@
 #include <dlfcn.h>
```
primiano: I think you can remove this now?

perfetto_flags.aconfig
```diff
@@ -35,3 +35,10 @@ flag {
   bug: "439758665"
   is_fixed_read_only: true
 }
+flag {
+  name: "use_rt_futex_for_android"
```
primiano: remove the _for_android

these flags make sense only on android, so for_android is completely redundant.
in your next cl where we sort out the ifdef we can rationalize everything. but this is not the place.

src/base/rt_mutex.cc
sashwinbalaji: Sorry the change wasn't complete (hence the draft status), PTAL now.

Thanks for the review btw as lot of the comments I would have missed.

We cannot remove dlfcn because as we agreed to keep flag enable_rt_mutex, there is a slight chance RTMutex is used on android with api version< 28 so we still need dlsym there.

Tested on android & linux by adding some logs to validate:
```
sashwinbalaji-mac:Desktop sashwinbalaji$ perfetto/tools/record_android_trace --sideload-path tracebox  --config  config_pbtxt -o mock.pftrace
tracebox: 1 file pushed, 0 skipped. 67.5 MB/s (10193384 bytes in 0.144s)
adbd cannot run as root in production builds
Running /data/local/tmp/tracebox --background --txt -o /data/local/tmp/2025-09-15_22-13-a650b5.pftrace -c -
[062.170]           rt_mutex.h:86 RTFutex Created
[062.170]  unix_task_runner.cc:45 PERFETTO_HAS_RT_FUTEX true
[062.170]  unix_task_runner.cc:53 PERFETTO_HAS_POSIX_RT_MUTEX false
[062.170]  unix_task_runner.cc:59 base::flags::use_rt_mutex false
[062.170]  unix_task_runner.cc:63 base::flags::use_rt_futex_for_android true
[062.173]     perfetto_cmd.cc:597 Parsing TraceConfig, 814 bytes


[hi on] sashwinbalaji@ashwinbalaji:~/Desktop/perfetto$ ./out/linux_clang_release/tracebox perfetto --txt -c ~/Desktop/traces/config_pbtxt -o ../mock.pftrace
[183.944]          rt_mutex.cc:88 RtPosixMutex Created
[183.944]  unix_task_runner.cc:47 PERFETTO_HAS_RT_FUTEX false
[183.944]  unix_task_runner.cc:51 PERFETTO_HAS_POSIX_RT_MUTEX true
[183.944]  unix_task_runner.cc:57 base::flags::use_rt_mutex true
[183.944]  unix_task_runner.cc:65 base::flags::use_rt_futex_for_android false
```

primiano: > We cannot remove dlfcn because as we agreed to keep flag enable_rt_mutex, there is a slight chance RTMutex is used on android with api version< 28 so we still need dlsym there.

Oh wait how do we end up using RTMutex on Android <28?

In my mind on android we should only try RTFutex, on linux/mac only the posix one

sashwinbalaji: ahh I had the misunderstanding flag rollout can happen on older android API versions which seems totally illogical. So have removed dlsym section and added #error to disable compile for RTMutex on older API versions

include/perfetto/base/build_configs/bazel/perfetto_build_flags.h
```diff
@@ -52,7 +52,7 @@
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_SYSTEM_CONSUMER() (1)
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_THREAD_SAFETY_ANNOTATIONS() (0)
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ENABLE_WINSCOPE() (1)
-#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ENABLE_RT_MUTEX() (1)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_RT_MUTEX_MODE() (1)
```
primiano: I just realized that if we do this, this will end up using PosixMutex in google3, which will break if somebody builds android in google3 and depends on our sdk (which I think people do).

Sorry for the back and forth.

I think the best way is realy to have only a boolean here, and move the decision between which impl to #ifdefs inside our code

sashwinbalaji: so moved the decision making to rt_mutex.h.

Tested it out on android platform build to be safe, generated perfetto_flags.h: http://gpaste/4883970873491456

include/perfetto/ext/base/flags.h
```diff
@@ -37,6 +37,9 @@ namespace perfetto::base::flags {
   X(use_murmur_hash_for_flat_hash_map, NonAndroidPlatformDefault_TRUE) \
   X(ftrace_clear_offline_cpus_only, NonAndroidPlatformDefault_TRUE)    \
   X(use_rt_mutex, PERFETTO_BUILDFLAG(PERFETTO_ENABLE_RT_MUTEX)         \
```
primiano: I believe now the 2nd argument here is useless.
I would just set it to ..._FALSE and add a comment saying that the source of truth for other platforms is in rt_mutex.h

PR #2871, opened 2025-09-09
WIP support clone operation for 'write_into_file' sessions.

KirillTim:
We pass a FD from the `perfetto_cmd` into the `traced` when doing `CloneSession` RPC call.
In `traced` we:
1. Copy the data already written to `write_into_file` fd to the `clone_session` fd.
2. Flush data for the `write_into_file` session
3. When the data is flushed and cloned, instead of asking customer to read it back we write it in `traced`.

Unit tests:
`WriteIntoFileCloneSession` is a simple test
`CloneSessionPreamblePackets_DEBUG` is an important one (missing asserts though). It should test that our "alternative" clone code flow generates the trace that contains all the expected lifecycle events.

### DIFF
--- a/include/perfetto/ext/tracing/core/null_consumer_endpoint_for_testing.h
+++ b/include/perfetto/ext/tracing/core/null_consumer_endpoint_for_testing.h
@@ -34,7 +34,7 @@ class NullConsumerEndpointForTesting : public ConsumerEndpoint {
   void ChangeTraceConfig(const perfetto::TraceConfig&) override {}
   void StartTracing() override {}
   void DisableTracing() override {}
-  void CloneSession(CloneSessionArgs) override {}
+  void CloneSession(CloneSessionArgs, base::ScopedFile) override {}
   void Flush(uint32_t, FlushCallback, FlushFlags) override {}
   void ReadBuffers() override {}
   void FreeBuffers() override {}
--- a/src/perfetto_cmd/perfetto_cmd.cc
+++ b/src/perfetto_cmd/perfetto_cmd.cc
@@ -1020,6 +1020,14 @@ void PerfettoCmd::OnConnect() {
     return;
   }
 
+  PERFETTO_LOG(
+      "PerfettoCmd::OnConnect, is_clone(): %d, trace_out_stream_: %p, "
+      "trace_out_path_: %s, trace_config_->write_into_file(): %d, "
+      "snapshot_config_.empty(): %d",
+      is_clone(), static_cast<void*>(trace_out_stream_.get()),
+      trace_out_path_.c_str(), trace_config_->write_into_file(),
+      snapshot_config_.empty());
+
   if (is_clone()) {
     task_runner_.PostDelayedTask(std::bind(&PerfettoCmd::OnTimeout, this),
                                  kCloneTimeoutMs);
@@ -1039,7 +1047,14 @@ void PerfettoCmd::OnConnect() {
       args.clone_trigger_boot_time_ns = snapshot_trigger_info_->boot_time_ns;
       args.clone_trigger_delay_ms = snapshot_trigger_info_->trigger_delay_ms;
     }
-    consumer_endpoint_->CloneSession(std::move(args));
+
+    base::ScopedFile optional_fd;
+    if (trace_out_stream_) {
+      optional_fd.reset(dup(fileno(*trace_out_stream_)));
+    }
+
+    PERFETTO_LOG("PerfettoCmd::OnConnect, consumer_endpoint_->CloneSession");
+    consumer_endpoint_->CloneSession(std::move(args), std::move(optional_fd));
     return;
   }
 
@@ -1123,6 +1138,9 @@ void PerfettoCmd::CheckTraceDataTimeout() {
 void PerfettoCmd::OnTraceData(std::vector<TracePacket> packets, bool has_more) {
   trace_data_timeout_armed_ = false;
 
+  // PERFETTO_DLOG("PerfettoCmd::OnTraceData, packets.size(): %lu",
+  //               packets.size());
+
   PERFETTO_CHECK(packet_writer_.has_value());
   if (!packet_writer_->WritePackets(packets)) {
     PERFETTO_ELOG("Failed to write packets");
@@ -1138,6 +1156,7 @@ void PerfettoCmd::OnTracingDisabled(const std::string& error) {
 }
 
 void PerfettoCmd::ReadbackTraceDataAndQuit(const std::string& error) {
+  PERFETTO_LOG("ReadbackTraceDataAndQuit, error: %s", error.c_str());
   if (!error.empty()) {
     // Some of these errors (e.g. unique session name already exists) are soft
     // errors and likely to happen in nominal condition. As such they shouldn't
--- a/src/tracing/ipc/consumer/consumer_ipc_client_impl.cc
+++ b/src/tracing/ipc/consumer/consumer_ipc_client_impl.cc
@@ -464,11 +464,20 @@ void ConsumerIPCClientImpl::SaveTraceForBugreport(
   consumer_port_.SaveTraceForBugreport(req, std::move(async_response));
 }
 
-void ConsumerIPCClientImpl::CloneSession(CloneSessionArgs args) {
+void ConsumerIPCClientImpl::CloneSession(CloneSessionArgs args,
+                                         base::ScopedFile fd) {
   if (!connected_) {
     PERFETTO_DLOG("Cannot CloneSession(), not connected to tracing service");
     return;
   }
+
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  if (fd) {
+    consumer_->OnTracingDisabled(
+        "Passing FDs into CloneSession is not supported on Windows");
+    return;
+  }
+#endif
 
   protos::gen::CloneSessionRequest req;
   if (args.tsid) {
@@ -514,6 +523,6 @@ void ConsumerIPCClientImpl::CloneSession(CloneSessionArgs args) {
               {response->success(), response->error(), uuid});
         }
       });
-  consumer_port_.CloneSession(req, std::move(async_response));
+  consumer_port_.CloneSession(req, std::move(async_response), *fd);
 }
 }  // namespace perfetto
--- a/src/tracing/ipc/service/consumer_ipc_service.cc
+++ b/src/tracing/ipc/service/consumer_ipc_service.cc
@@ -355,7 +355,9 @@ void ConsumerIPCService::CloneSession(
   if (req.has_clone_trigger_delay_ms()) {
     args.clone_trigger_delay_ms = req.clone_trigger_delay_ms();
   }
-  remote_consumer->service_endpoint->CloneSession(std::move(args));
+  base::ScopedFile fd = ipc::Service::TakeReceivedFD();
+  remote_consumer->service_endpoint->CloneSession(std::move(args),
+                                                  std::move(fd));
 }
 
 // Called by the service in response to
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -29,6 +29,7 @@
 
 #if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN) && \
     !PERFETTO_BUILDFLAG(PERFETTO_OS_NACL)
+#include <sys/sendfile.h>
 #include <sys/uio.h>
 #include <sys/utsname.h>
 #include <unistd.h>
@@ -274,6 +275,39 @@ base::ScopedFile CreateTraceFile(const std::string& path, bool overwrite) {
     PERFETTO_PLOG("Failed to create %s", path.c_str());
   }
   return fd;
+}
+
+// TODO(ktimofeev): implement for other OSes
+// TODO(ktimofeev): add fallback to read/write loop if sendfile doesn't work
+// See chromium implementation for details:
+// https://source.chromium.org/chromium/chromium/src/+/main:base/files/file_util.cc;l=190;drc=134eb3609e728c12c60a92656f8bca1c888947c1
+base::Status CopyFile(base::PlatformHandle from_fd,
+                      base::PlatformHandle to_fd) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID) || \
+    PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX)
+  auto maybe_file_size = base::GetFileSize(from_fd);
+  if (!maybe_file_size)
+    return base::ErrStatus("Can't get size of 'from_fd' file.");
+  uint64_t file_size = maybe_file_size.value();
+
+  size_t copied = 0;
+  ssize_t res = 0;
+  do {
+    // Don't specify an offset and the kernel will begin reading/writing to the
+    // current file offsets.
+    res = PERFETTO_EINTR(
+        sendfile(to_fd, from_fd, /*offset=*/nullptr,
+                 /*count=*/static_cast<size_t>(file_size) - copied));
+    if (res <= 0) {
+      break;
+    }
+
+    copied += static_cast<size_t>(res);
+  } while (copied < file_size);
+  return base::OkStatus();
+#else
+  return base::ErrStatus("CopyFile is not implemented for this OS.");
+#endif
 }
 
 bool ShouldLogEvent(const TraceConfig& cfg) {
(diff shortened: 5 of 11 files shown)


include/perfetto/ext/tracing/core/null_consumer_endpoint_for_testing.h
```diff
@@ -34,7 +34,7 @@ class NullConsumerEndpointForTesting : public ConsumerEndpoint {
   void ChangeTraceConfig(const perfetto::TraceConfig&) override {}
   void StartTracing() override {}
   void DisableTracing() override {}
-  void CloneSession(CloneSessionArgs) override {}
+  void CloneSession(CloneSessionArgs, base::ScopedFile) override {}
```
primiano: move this into CloneSessionARgs.
You are defeating the point of me introducing CSA.
That exists to avoid API brekages when doing changes like this. this interface is extended in other repos outside of perfetto (chromium, google3).
if you do this, now you'll require multi-side patches and you'll never get alive out of it.

P.S: When you saw that, didn't you wonder "I wonder why we have CSArgs, maybe I should put it there?". ?? ;-)

src/tracing/ipc/service/consumer_ipc_service.cc
```diff
@@ -355,7 +355,9 @@ void ConsumerIPCService::CloneSession(
   if (req.has_clone_trigger_delay_ms()) {
     args.clone_trigger_delay_ms = req.clone_trigger_delay_ms();
   }
-  remote_consumer->service_endpoint->CloneSession(std::move(args));
+  base::ScopedFile fd = ipc::Service::TakeReceivedFD();
```
primiano: add a comment saying that this is only needef for when write_into_file=true

src/tracing/service/tracing_service_impl.cc
```diff
@@ -276,6 +277,39 @@ base::ScopedFile CreateTraceFile(const std::string& path, bool overwrite) {
...
 
+// TODO(ktimofeev): implement for other OSes
+// TODO(ktimofeev): add fallback to read/write loop if sendfile doesn't work
+// See chromium implementation for details:
+// https://source.chromium.org/chromium/chromium/src/+/main:base/files/file_util.cc;l=190;drc=134eb3609e728c12c60a92656f8bca1c888947c1
+base::Status CopyFile(base::PlatformHandle from_fd,
+                      base::PlatformHandle to_fd) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID) || \
+    PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX)
+  auto maybe_file_size = base::GetFileSize(from_fd);
```
primiano: for readability either you do

1) std::optional<uint64_t> maybe_file_size ... here
2) if (maybe_file_size.has_value()) below

otherwise it's unclear this is a std::optional when reading the code

src/tracing/service/tracing_service_impl.cc
```diff
@@ -276,6 +277,39 @@ base::ScopedFile CreateTraceFile(const std::string& path, bool overwrite) {
...
+    return base::ErrStatus("Can't get size of 'from_fd' file.");
+  uint64_t file_size = maybe_file_size.value();
+
+  size_t copied = 0;
+  ssize_t res = 0;
+  do {
+    // Don't specify an offset and the kernel will begin reading/writing to the
+    // current file offsets.
+    res = PERFETTO_EINTR(
+        sendfile(to_fd, from_fd, /*offset=*/nullptr,
```
primiano: if you use sendfile, you should:

1. store the original seek position
2. lseek(0, SEEK_SET)
3. do the sendfile
4. restore the seek position with lseek(prev, SEEK_SET)


but also I wouln't use sendfile here, because the signature is different on MacOs, and you'd have to work with both. Also windows.

Honestly rather than all this drama, I'd just implement it in file_utils.h/cc and for now just use pread/pwrite

Also I think (i am not sure tbh) you might have to call FlushFile on the source before reading it back. i vaguely remember some problems in som etests that were doing somethign similar before

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2450,22 +2484,13 @@ bool TracingServiceImpl::ReadBuffersIntoFile(TracingSessionID tsid) {
...
-  bool stop_writing_into_file = false;
-  do {
-    std::vector<TracePacket> packets =
-        ReadBuffers(tracing_session, kWriteIntoFileChunkSize, &has_more);
+  const uint64_t max_size = tracing_session->max_file_size_bytes
+                                ? tracing_session->max_file_size_bytes
+                                : std::numeric_limits<size_t>::max();
 
-    stop_writing_into_file = WriteIntoFile(tracing_session, std::move(packets));
-  } while (has_more && !stop_writing_into_file);
```
primiano, line 2469: why are you touching all this?

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2765,15 +2816,10 @@ void TracingServiceImpl::MaybeCompressPackets(
...
+                                       uint64_t* bytes_written_into_file,
+                                       const uint64_t max_size,
                                        std::vector<TracePacket> packets) {
-  if (!tracing_session->write_into_file) {
-    return false;
-  }
-  const uint64_t max_size = tracing_session->max_file_size_bytes
-                                ? tracing_session->max_file_size_bytes
-                                : std::numeric_limits<size_t>::max();
-
```
primiano, line 2777: why are you touching this?

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4097,7 +4142,8 @@ size_t TracingServiceImpl::PurgeExpiredAndCountTriggerInWindow(
 
 base::Status TracingServiceImpl::FlushAndCloneSession(
     ConsumerEndpointImpl* consumer,
-    ConsumerEndpoint::CloneSessionArgs args) {
+    ConsumerEndpoint::CloneSessionArgs args,
```
primiano: did it really not occur to you, after having added this "fd argument" dozen times "hmm maybe I should put it in CSArgs?"

src/tracing/service/tracing_service_impl.h
```diff
@@ -881,6 +884,17 @@ class TracingServiceImpl : public TracingService {
   // happened.
   static bool IsWaitingForTrigger(TracingSession* tracing_session);
 
+  // Reads all the tracing buffers from the `*tracing_session` (but no more than
+  // `max_size` bytes) and writes them into `file`.
+  //
+  // Sets `*bytes_written_into_file` to the amount of bytes being written.
+  //
+  // Returns false in case of error.
+  bool DoReadBuffersIntoFile(TracingSession* tracing_session,
```
primiano: can you explain why we need this further abstraction in the middle?

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4461,6 +4531,17 @@ base::Status TracingServiceImpl::FinishCloneSession(
   cloned_session->final_flush_outcome = final_flush_outcome
                                             ? TraceStats::FINAL_FLUSH_SUCCEEDED
                                             : TraceStats::FINAL_FLUSH_FAILED;
+
+  if (output_file_fd) {
+    uint64_t bytes_written_into_file = 0;
+    DoReadBuffersIntoFile(cloned_session, output_file_fd,
```
primiano: it's not clear to me why here you don't just

1. set the session write_period_ms = 0, which will trigger a stop/flush
2. just call ReadBuffersIntoFile(tsid /*the cloned id*) here

src/tracing/ipc/consumer/consumer_ipc_client_impl.cc
```diff
@@ -470,6 +470,14 @@ void ConsumerIPCClientImpl::CloneSession(CloneSessionArgs args) {
     return;
   }
 
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  if (args.output_file_fd) {
+    consumer_->OnTracingDisabled(
```
primiano: This should be consumer->OnSessionCloned(OnSessionClonedArgs(success=false, error="Passing..."))

Tecnically we should not do OnTracingDisabled if it's NOT in response to a EnableTracing

src/tracing/ipc/consumer/consumer_ipc_client_impl.cc
```diff
@@ -514,6 +522,7 @@ void ConsumerIPCClientImpl::CloneSession(CloneSessionArgs args) {
               {response->success(), response->error(), uuid});
         }
       });
-  consumer_port_.CloneSession(req, std::move(async_response));
+  consumer_port_.CloneSession(req, std::move(async_response),
```
primiano, line 530: Just for future ourselves, add a comment

// |fd| will be closed when this function returns, but it's fine because the
// IPC layer dup()'s it when sending the IPC.

src/tracing/ipc/service/consumer_ipc_service.cc
```diff
@@ -355,6 +355,7 @@ void ConsumerIPCService::CloneSession(
   if (req.has_clone_trigger_delay_ms()) {
     args.clone_trigger_delay_ms = req.clone_trigger_delay_ms();
   }
+  args.output_file_fd = ipc::Service::TakeReceivedFD();
```
primiano, line 360: Add a comment saying that the client sends the file descriptor in all cases, but we use it only  if the sesssion is write_into_file, and close it otherwise.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4140,6 +4142,40 @@ base::Status TracingServiceImpl::FlushAndCloneSession(
     return PERFETTO_SVC_ERR("Not allowed to clone a session from another UID");
   }
```
primiano, line 4143: I would do:

```txt

int write_into_file_fd = session->write_into_file.get();
if (!base::flags::buffer_clone_preserve_read_iter) {
  // In the old behavior we used to ignore the fill.
  // keep doing it.
  write_into_file_fd = kInvalid
}
if (write_into_file_fd != kInvalid) {
    if (!args.output_file_fd) {
      return PERFETTO_SVC_ERR(
          "Failed to clone 'write_into_file' session: a file descriptor is "
          "required to copy existing file");
    }
 base::FlushFile(..);
 CopyFileContents
 ...
}
```

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2473,7 +2474,8 @@ bool TracingServiceImpl::ReadBuffersIntoFile(TracingSessionID tsid) {
     base::FlushFile(tracing_session->write_into_file.get());
     tracing_session->write_into_file.reset();
     tracing_session->write_period_ms = 0;
-    if (tracing_session->state == TracingSession::STARTED)
+    if (tracing_session->state == TracingSession::STARTED ||
+        tracing_session->state == TracingSession::CLONED_READ_ONLY)
```
primiano: I don't think you need this

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4140,6 +4142,40 @@ base::Status TracingServiceImpl::FlushAndCloneSession(
...
+    base::FlushFile(*session->write_into_file);
+    base::Status status =
+        base::CopyFileContents(*session->write_into_file, *args.output_file_fd);
+    if (!status.ok()) {
+      return PERFETTO_SVC_ERR(
+          "Failed to clone 'write_into_file' session: failed to copy existing "
+          "file: %s",
+          status.c_message());
+    }
+  }
```
primiano, line 4174: isn't this missing an else, like

if (original_session.write_into_file) {

  ...
} else {
  args.output_file_fd.reset()?  //// !this
}

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4192,6 +4228,9 @@ base::Status TracingServiceImpl::FlushAndCloneSession(
         args.clone_trigger_producer_name,
         args.clone_trigger_trusted_producer_uid, args.clone_trigger_delay_ms};
   }
+  if (args.output_file_fd) {
```
primiano, line 4228: Linked to the other comment, this should not happen if we didn't need a fd because the original session wasn't clone into file

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4355,7 +4395,8 @@ base::Status TracingServiceImpl::FinishCloneSession(
     bool final_flush_outcome,
     std::optional<TriggerInfo> clone_trigger,
     base::Uuid* new_uuid,
-    int64_t clone_started_timestamp_ns) {
+    int64_t clone_started_timestamp_ns,
+    base::ScopedFile output_file_fd) {
```
primiano, line 4398: naming-wise I feel it's a bit easier to understand if you call this write_into_file_fd, so it's clear this is only valid if the session was a write-into-file itself.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1039,6 +1039,11 @@ void PerfettoCmd::OnConnect() {
       args.clone_trigger_boot_time_ns = snapshot_trigger_info_->boot_time_ns;
       args.clone_trigger_delay_ms = snapshot_trigger_info_->trigger_delay_ms;
     }
+
+    if (trace_out_stream_) {
+      args.output_file_fd = base::ScopedFile(dup(fileno(*trace_out_stream_)));
```
primiano, line 1048: add a comment saying that here we have to apss the file anyways because we dont' know if the session is write_into_file or not. traced will decide ... yada yada

src/tracing/service/tracing_service_impl.cc
KirillTim: Nice catch, thanks!
I've also add a new test `CloneWithFileDescriptorNoWriteIntoFile` for this specific case.

src/tracing/service/tracing_service_impl.cc
KirillTim: I've done this in a slightly different way.

src/tracing/service/tracing_service_impl.cc
KirillTim: For me it seems to be more confusing to have two different things with the same name.
The `write_into_file` unambiguously names the session that have a file it may write to.
The `output_file_fd` names the file, only used for the clone operation. If we rename it to the `write_into_file_fd` it will be harder to figure out that this thing is relevant only for the clone operation.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4140,6 +4141,34 @@ base::Status TracingServiceImpl::FlushAndCloneSession(
...
+    base::Status status =
+        base::CopyFileContents(*session->write_into_file, *args.output_file_fd);
+    if (!status.ok()) {
+      return PERFETTO_SVC_ERR(
+          "Failed to clone 'write_into_file' session: failed to copy existing "
+          "file: %s",
+          status.c_message());
+    }
+  } else {
+    args.output_file_fd.reset();
```
primiano, line 4173: Add a small comment saying:

// The client always sends a FD because when it asks to CloneSession,
// it doesn't know if the session being cloned is WIF or not. If it's
// not we should just ignore the file, the client will readback via IPC
// as usual in that case.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1039,6 +1039,15 @@ void PerfettoCmd::OnConnect() {
       args.clone_trigger_boot_time_ns = snapshot_trigger_info_->boot_time_ns;
       args.clone_trigger_delay_ms = snapshot_trigger_info_->trigger_delay_ms;
     }
+
+    if (trace_out_stream_) {
+      // We always send the file descriptor to the traced, because perfetto_cmd
+      // doesn't know if the session to clone is 'write_into_file' session.
+      // The traced decides weither it writes the cloned buffers to this file
```
primiano: Drop the "the", just "traced"

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1039,6 +1039,15 @@ void PerfettoCmd::OnConnect() {
...
+      // descriptor or send them to the perfetto_cmd.
+      args.output_file_fd = base::ScopedFile(dup(fileno(*trace_out_stream_)));
+    }
+
```
primiano, line 1050: don't we need something below in ReadbackTraceDataAndQuit()?
perfetto_cmd doesn't know if the session cloned was WIF or not, so needs to know somehow it from traced, or it will try to read from IPC?

test/cmdline_integrationtest.cc
```diff
@@ -1130,6 +1131,87 @@ TEST_F(PerfettoCmdlineTest, CloneByName) {
...
+    std::string stderr_str;
+    EXPECT_EQ(0, perfetto_proc.Run(&stderr_str)) << stderr_str;
+  });
+
+  test_helper().WaitForProducerEnabled();
+  auto on_data_written = task_runner_.CreateCheckpoint("data_written");
+  fake_producer->ProduceEventBatch(test_helper().WrapTask(on_data_written));
+  task_runner_.RunUntilCheckpoint("data_written");
+
+  // Wait untill all the data for the 'write_into_file' session is written into
```
primiano: nit: "until"

primiano (commented):
+  // NOTE: if we arrive here through CloneSession()->OnSessionCloned(), the
+  // branch below will be always NOT taken, even if the cloned session was
+  // write_into_file (in which case traced will have done the right thing).
+  // We rely on the fact that issuing a ReadBuffers() on a cloned session will
+  // return only a single TracingServiceEvent::read_tracing_buffers_completed

PR #2876, opened 2025-09-10
ui: Add Network trace section for WiFi debugging

loordoza:
This pull request introduces a new 'Network' section to the UI's trace recorder. This feature is crucial for the ChromeOS WiFi team to easily capture detailed kernel events, which is essential for debugging issues.

The new UI provides users with direct controls to configure tracing for network activity:
- 802.11 Layer Events: By enabling ftrace events from the cfg80211 and mac80211 domains.
- Packet Traffic: Using kernel events for packet transmission and reception (net_dev_xmit, netif_receive_skb), with a dedicated slider to set a separate buffer for high-volume data.
- Custom Events: A text area is included to add any other driver-specific ftrace events for targeted debugging.

This change provides a user-friendly way to capture critical network data, significantly improving our ability to diagnose complex WiFi issues.


ui/src/plugins/dev.perfetto.RecordTraceV2/pages/network.ts
```diff
@@ -0,0 +1,98 @@
...
+    icon: 'wifi',
+    probes: [wifiNetworkTracing()],
+  };
+}
+
+function wifiNetworkTracing(): RecordProbe {
+  const cfgMacEvents = ['cfg80211/*', 'mac80211/*'];
+  const netEvents = ['net/netif_receive_skb', 'net/net_dev_xmit'];
+  const settings = {
+    cfg_mac: new Toggle({
```
LalitMaganti, line 37: I would make clear here that this is a really spammy thing to do, not something you want to enable by default. I can easily imagine people enabling this without being a wlan expert and having data losses/buffer overflows because you are enabling every single event

I'd either ask for you to curate these events or make this an advanced toggle which is clear most people should not enable.

loordoza: Thanks for the feedback. I've updated the descriptions to better warn users about the risks of potential data loss.

I agree that some event categories are a risk for buffer overflow. However, in my experience, the 802.11 layer events are not a significant source of data volume. Even with over 300 different events enabled, they are relatively rare compared to high-frequency events like those related to CPU scheduling or I/O.

PR #2879, opened 2025-09-10
tp: Implement StructuredQuery join

aMayzner:
This commit introduces a `join` source to the StructuredQuery, allowing for structured `INNER` and `LEFT` joins between two queries.

The implementation includes the Protobuf definition for the `Join` message, C++ logic in `structured_query_generator.cc` to generate the corresponding SQL, unit tests for both join types, and updated documentation with an example.


protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -180,6 +180,32 @@ message PerfettoSqlStructuredQuery {
...
+  message Join {
+    // The type of join to perform.
+    enum Type {
+      // Standard SQL INNER JOIN. A row is produced only if the join condition
+      // is met in both the left and right queries.
+      INNER = 0;
+      // Standard SQL LEFT JOIN. All rows from the left query are included,
+      // with matching rows from the right query. If there is no match, the
+      // columns from the right query will be NULL.
+      LEFT = 1;
```
LalitMaganti, line 209: What about outer?

protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -180,6 +180,32 @@ message PerfettoSqlStructuredQuery {
...
+    }
+
+    // The type of join. Required.
+    optional Type type = 1;
+    // The "left" query in the join. Required.
+    optional PerfettoSqlStructuredQuery left_query = 2;
+    // The "right" query in the join. Required.
+    optional PerfettoSqlStructuredQuery right_query = 3;
+    // The name of the column in the left query to join on. Required.
```
LalitMaganti: Are we intentionally not supporting arbitrary join expressions?

protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -180,6 +180,32 @@ message PerfettoSqlStructuredQuery {
...
+      // columns from the right query will be NULL.
+      LEFT = 1;
+    }
+
+    // The type of join. Required.
+    optional Type type = 1;
+    // The "left" query in the join. Required.
+    optional PerfettoSqlStructuredQuery left_query = 2;
+    // The "right" query in the join. Required.
+    optional PerfettoSqlStructuredQuery right_query = 3;
```
LalitMaganti, line 217: What about joining n tables? Do we really anticipate having to do a cascading set of two table joins to join n tables?

docs/analysis/trace-summary.md
```diff
@@ -509,6 +509,46 @@ query: {
 }
 ```
 
+### Joining Tables with `join`
```
LalitMaganti: I'm skeptical of this proto being the right shape. I'd consider making this an underscored feature so we can experiment without this being API.

protos/perfetto/perfetto_sql/structured_query.proto
aMayzner: This is for now how explore page is doing it, so let's do it like that for now. We might change it in the future.

protos/perfetto/perfetto_sql/structured_query.proto
aMayzner: Supported

protos/perfetto/perfetto_sql/structured_query.proto
aMayzner: I'll add it when it's needed/requested

protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -223,6 +268,9 @@ message PerfettoSqlStructuredQuery {
     // Source is an interval intersect operation. See IntervalIntersect
     // documentation for more information.
     IntervalIntersect interval_intersect = 7;
+
+    // DON'T USE. EXPERIMENTAL.
+    ExperimentalJoin join = 12;
```
LalitMaganti: experimental_join. Also use some very high id so this id can be reused.

Issue #2702, opened 2025-08-25
Shared y-axis bounds don't update properly when new traces are loaded


cphlipot1:
It looks like we are not refreshing the y-axis bounds correctly whenever a user loads a new trace in the same perfetto UI tab.

I've reproduced this using the recently added y_axis_share_key, but appears to be unrelated to that specific addition as this also reproduces with other plugins that turn on y-axis sharing by default, like power rails.

## How to reproduce

[large_values_synthetic_counters_shared_y_axis.pftrace.zip](https://github.com/user-attachments/files/21977840/large_values_synthetic_counters_shared_y_axis.pftrace.zip)
[synthetic_counters_shared_y_axis.pftrace.zip](https://github.com/user-attachments/files/21977839/synthetic_counters_shared_y_axis.pftrace.zip)

1. open ui.perfetto.dev 
2. load **synthetic_counters_shared_y_axis.pftrace.zip**
3. observe
4. in the same tab, load **large_values_synthetic_counters_shared_y_axis.pftrace.zip**
5. observe
6. in the same tab load **synthetic_counters_shared_y_axis.pftrace.zip**
7. observe

## Expected outcome:

the y-axis bounds should match the same values as if the traces were loaded in a fresh Perfetto UI browser tab.

After step 2:

<img width="419" height="155" alt="Image" src="https://github.com/user-attachments/assets/3b77a868-1c97-49bc-84fb-108dbe6e6cb5" />

After step 4:

<img width="407" height="177" alt="Image" src="https://github.com/user-attachments/assets/cf091084-ecc7-4d18-aaeb-436068b65cd5" />

After Step 6:

<img width="419" height="155" alt="Image" src="https://github.com/user-attachments/assets/fa38cf0e-6611-4dfc-8bb8-2a5583fb351f" />

## Actual outcome

y-axis bounds never seem to decrease when new traces are loaded. only increase

After step 2 (normal)

<img width="419" height="155" alt="Image" src="https://github.com/user-attachments/assets/5accfd2a-e887-451a-93c5-3311812f3efb" />

After step 4  (bug?) -I'm seeing the top track not refresh its range until I mouse over it. presumably this is an unrelated redraw bug:

<img width="437" height="135" alt="Image" src="https://github.com/user-attachments/assets/1b4a5df8-cd1a-44ce-8073-43e14d028185" />

After step 6 (bug)- y-axis max does not go back down to that from step 2.

[...]

PR #2900, opened 2025-09-11
ui: fix counter range sharing and add scoped object mechanism to trace

LalitMaganti:
This CL fixes the singelton rangesharer class which would cause range
sharing to persist across UI trace loads. To do this, I introduce a new
mechanism to attach aribtrary state to a trace object to avoid coupling
between the core and the components/plugins.

Fixes: https://github.com/google/perfetto/issues/2702


ui/src/public/trace.ts
```diff
@@ -57,6 +57,10 @@ export interface Trace extends App {
   readonly minimap: MinimapManager;
   readonly search: SearchManager;
 
+  // Gets or creates a trace-scoped object by key. Objects created this way
+  // are automatically disposed when the trace is closed.
+  getOrCreateScopedObject<T>(key: string, factory: () => T): T;
+
```
stevegolton: IMO a cleaner approach would be to use a `WeakMap<Trace, T>` in `base_counter_track.ts`. A WeakMap allows you to use an object's referential identity as a key while ensuring the object can still be GCd when no other references exist.

This enables you to effectively add extra properties that are associated with a given trace and persist for the lifetime of that trace without having to change the trace API.

LalitMaganti: Thanks, that's indeed a lot cleaner! I didn't know that was possible. Let me change to do that instead.

LalitMaganti: Done. PTAL.

Issue #734, opened 2024-03-08
cpu_profiler.py uses tempfile  before closing it on window 10 problems


dvdface:
on Window 10, can't push file before close it. it will cause adb push failure
https://raw.githubusercontent.com/google/perfetto/main/tools/cpu_profile

`
def record_trace(config, profile_target):
  """Runs Perfetto with the provided configuration to record a trace.

  Args:
    config: The Perfetto config to be used for tracing/profiling.
    profile_target: The directory where the recorded trace is output.
  """
  NULL = open(os.devnull)
  NO_OUT = {
      'stdout': NULL,
      'stderr': NULL,
  }
  if not release_or_newer('T'):
    sys.exit("This tool requires Android T+ to run.")

  #Push configuration to the device.
  tf = tempfile.NamedTemporaryFile()
  tf.file.write(config.encode('utf-8'))
  tf.file.flush()
  profile_config_path = '/data/misc/perfetto-configs/config-' + UUID
  adb_check_output(['adb', 'push', tf.name, profile_config_path])
  tf.close()
`

should change to 

`
def record_trace(config, profile_target):
  """Runs Perfetto with the provided configuration to record a trace.

  Args:
    config: The Perfetto config to be used for tracing/profiling.
    profile_target: The directory where the recorded trace is output.
  """
  NULL = open(os.devnull)
  NO_OUT = {
      'stdout': NULL,
      'stderr': NULL,
  }
  if not release_or_newer('T'):
    sys.exit("This tool requires Android T+ to run.")

  #Push configuration to the device.
  tf = tempfile.NamedTemporaryFile(delete=False)
  tf.file.write(config.encode('utf-8'))
  tf.file.flush()
  tf.close()
  profile_config_path = '/data/misc/perfetto-configs/config-' + UUID
  adb_check_output(['adb', 'push', tf.name, profile_config_path])
  os.remove(tf.name)
`

primiano:
Can you please send a patch following https://perfetto.dev/docs/contributing/getting-started ?

PR #2912, opened 2025-09-12
tools: fix cpu_profile temp file handling on Windows

LalitMaganti:
On Windows, temp files cannot be accessed by external processes while
still open due to file locking. This caused adb push to fail when
trying to push the config file to the device.

Fix by closing the temp file before adb push and using try/finally
to ensure proper cleanup even if errors occur.

Fixes: https://github.com/google/perfetto/issues/734

### DIFF
--- a/python/tools/cpu_profile.py
+++ b/python/tools/cpu_profile.py
@@ -342,10 +342,12 @@ def record_trace(config, profile_target):
     sys.exit("This tool requires Android T+ to run.")
 
   # Push configuration to the device.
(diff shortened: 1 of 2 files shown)


LalitMaganti:
> there should be a comment on why the code has to be written this way. I foresee a rewrite to these scripts soon (imo we should make these things a plugin/addon onto record_android_trace), and we'll forget about windows again.

Ack. Added a comment.

> do we need a similar edit in tools/heap_profile?

tools/heap_profile seems to work a different way. No change needed there.

Issue #816, opened 2024-06-10
Build on Windows should compile additional security flags


ivberg:
This is related to similar errors using trace_processor_shell.exe under constrained security environment #635

Similar compiler flags are recommended to be used. 

So we were looking to use trace_processor_shell.exe at Microsoft but the binary failed a set of security checks from [binskim](https://github.com/microsoft/binskim). These compile flags would be needed to use the binary.

1. Optional to fix (warning)
##[warning]1. BinSkim Warning BA2024 - File: file:///D:/a/_work/_temp/Microsoft.Performance.Toolkit.Plugins.PerfettoPlugin-1.5.5.ptix/plugin/trace_processor_shell.exe.  
Signature: bb153f488d6c6f10d936daa45314e203f796b8a444582e5b04226f08aec44667
Tool: BinSkim: Rule: BA2024 (EnableSpectreMitigations). https://github.com/microsoft/binskim/blob/main/docs/BinSkimRules.md#rule-BA2024EnableSpectreMitigations
'trace_processor_shell.exe' was compiled with one or more modules that do not enable code generation mitigations for speculative execution side-channel attack (Spectre) vulnerabilities. Spectre attacks can compromise hardware-based isolation, allowing non-privileged users to retrieve potentially sensitive data from the CPU cache. To resolve the issue, provide the /Qspectre switch on the compiler command-line (or /d2guardspecload in cases where your compiler supports this switch and it is not possible to update to a toolset that supports /Qspectre). This warning should be addressed for code that operates on data that crosses a trust boundary and that can affect execution, such as parsing untrusted file inputs or processing query strings of a web request.
The following modules were compiled with a toolset that supports /Qspectre but the switch was not enabled on the command-line:
[...]

PR #2913, opened 2025-09-12
gn: add additional flags for msvc

LalitMaganti:
Fixes: https://github.com/google/perfetto/issues/816

### DIFF
--- a/gn/standalone/toolchain/BUILD.gn
+++ b/gn/standalone/toolchain/BUILD.gn
@@ -466,7 +466,7 @@ toolchain("msvc") {
   tool("cc") {
     precompiled_header_type = "msvc"
     pdbname = "{{target_out_dir}}/{{label_name}}_c.pdb"
-    command = "$cc_wrapper $cc /nologo /showIncludes /FC {{defines}} {{include_dirs}} {{cflags}} {{cflags_c}} ${external_cflags} /c {{source}} /Fo{{output}} /Fd\"$pdbname\" /guard:cf /ZH:SHA_256"
+    command = "$cc_wrapper $cc /nologo /showIncludes /FC {{defines}} {{include_dirs}} {{cflags}} {{cflags_c}} ${external_cflags} /c {{source}} /Fo{{output}} /Fd\"$pdbname\" /guard:cf /ZH:SHA_256 /Qspectre"
     depsformat = "msvc"
     outputs =
         [ "{{source_out_dir}}/{{target_output_name}}.{{source_name_part}}.obj" ]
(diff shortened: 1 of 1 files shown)


gn/standalone/toolchain/BUILD.gn
```diff
@@ -476,7 +476,7 @@ toolchain("msvc") {
   tool("cxx") {
     precompiled_header_type = "msvc"
     pdbname = "{{target_out_dir}}/{{label_name}}_c.pdb"
-    command = "$cc_wrapper $cxx /nologo /showIncludes /FC {{defines}} {{include_dirs}} {{cflags}} {{cflags_cc}} ${external_cflags} /c {{source}} /Fo{{output}} /Fd\"$pdbname\" /guard:cf /ZH:SHA_256"
+    command = "$cc_wrapper $cxx /nologo /showIncludes /FC {{defines}} {{include_dirs}} {{cflags}} {{cflags_cc}} ${external_cflags} /c {{source}} /Fo{{output}} /Fd\"$pdbname\" /guard:cf /ZH:SHA_256 /Qspectre"
```
sashwinbalaji, line 479: should we benchamark before landing the change: https://learn.microsoft.com/en-us/cpp/build/reference/qspectre?view=msvc-170#performance-impact

gn/standalone/toolchain/BUILD.gn
```diff
@@ -505,7 +505,7 @@ toolchain("msvc") {
     pdbname = "${dllname}.pdb"
     rspfile = "${dllname}.rsp"
 
-    command = "$linker /nologo /IMPLIB:$libname ${sys_lib_flags} /DLL /OUT:$dllname /PDB:$pdbname @$rspfile"
+    command = "$linker /nologo /IMPLIB:$libname ${sys_lib_flags} /DLL /guard:cf /DYNAMICBASE /CETCOMPAT /OUT:$dllname /PDB:$pdbname @$rspfile"
```
sashwinbalaji, line 508: It's good for clarity but just saw by default, /DYNAMICBASE is enabled. (https://learn.microsoft.com/en-us/cpp/build/reference/dynamicbase?view=msvc-170#remarks)

gn/standalone/toolchain/BUILD.gn
primiano: I left a comment on the bug. not clear what's the implication of BinSkim being unhappy.
Not sure why we shoud enable bound checks for some code that is NOT a platform, where spectre attacks would do nothing useful.

LalitMaganti: Closing this PR. TP is very perf sensitive, I don't think we want to enable this.

PR #2914, opened 2025-09-12
tp: fix compile on windows by adding noreturn conditionally

LalitMaganti:


src/trace_processor/util/regex.h
```diff
@@ -117,6 +125,7 @@ class Regex {
     }
 #else
     base::ignore_result(s);
+    base::ignore_result(out);
```
primiano, line 128: in future, the best pay to do this is something like 
if (s) PERFETTO_FATAL

the compiler doesn't know that s is always non-null in practice (fi it's not, your code above would crash) and avoids having to deal with the WIN_NORETURN which is a bit awkward

Issue #888, opened 2024-09-16
Page keeps reloading when loading trace from localhost URL


s7tya:
As I mentioned in https://github.com/rust-lang/rustc-perf/pull/1967#issuecomment-2351902875, the page keeps reloading when trying to load a trace from a localhost URL. This only happens when [`isLocalhostTraceUrl`](https://github.com/google/perfetto/blob/56b5e6fdd77f3d86ad60d402197d91bda911b1f4/ui/src/frontend/trace_url_handler.ts#L218) is set to `true`.

Although this is a bit of an edge case because we’re modifying some URLs to allow CSP and embedding Perfetto within a project, I wanted to go ahead and report it anyway.

LalitMaganti:
Just a note to others: I looked at this issue, I think we should add some additional metadata to the open_trace_in_ui flow to distinguish these traces from other POSTs from localhost.

LalitMaganti:
This issue should now be resolved because as of v54 (https://github.com/google/perfetto/discussions/5063), we now support directly opening any publicly accessible trace using url= in the address bar.

See https://perfetto.dev/docs/visualization/deep-linking-to-perfetto-ui#option-1-direct-url-for-public-traces for more information.

Will close this out.

LalitMaganti:
(we also completely overhauled how our CSP worked to relax it and switched to using a service worker based approach instead).

PR #2916, opened 2025-09-12
ui: distinguish localhost urls from open_trace_in_ui from others

LalitMaganti:
Use the referrer to detect this to prevent reload loops from other
localhost places

Fixes: https://github.com/google/perfetto/issues/888


LalitMaganti:
Hmmm actually I think this is going to break people who push traces to UI but don't set referrer. Will leave it as it is and investigate at some point in the future.

PR #2924, opened 2025-09-12
Add OutOfMemoryError doc

ilkos:
Adding an OOME-specific doc (along with a way to get the heap dump without any dependencies, repo checkouts, python etc); in the past it's been useful to ask people to just copy and paste a command to diagnose


docs/case-studies/android-outofmemoryerror.md
```diff
@@ -0,0 +1,57 @@
+# Getting a OutOfMemoryError heap dump on Android
```
LalitMaganti: Can we please link this from the Android memory guide? Feels like it's a tool we should tell people about from there.

ilkos: Done

PR #2925, opened 2025-09-12
Keep first zero delta in counter_leading_intervals().

simonmacm:
We do this for the case where deltas are being displayed as a counter track in the UI. In this case we want to "reset" the counter to zero when it returns to zero (so we don't keep showing a non-zero value), but don't then need a long stream of zeroes after that.

Bug: http://b/422976190


src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
```diff
@@ -450,13 +450,25 @@ struct CounterPerTrackAgg
...
+      // no value.
+      //
+      // We check against two previous counters here because in the common case
+      // where deltas are being displayed as a counter track in the UI, we want
+      // to "reset" the counter to zero when it returns to zero (so we don't
+      // keep showing a non-zero value), but don't then need a long stream of
+      // zeroes after that.
+      std::vector<double> prev_vals = new_rows_track->val;
+      auto size = prev_vals.size();
+      if (size > 1 && std::equal_to<double>()(prev_vals[size - 1], val) &&
```
LalitMaganti: Small optimization: can we do the equals check to previous first and then the size check and then the second-previous check? This will keep it the same conditional behaviour as what we have today.

src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
```diff
@@ -450,13 +450,25 @@ struct CounterPerTrackAgg
...
+      // before had the same value we can safely remove the new one as it adds
+      // no value.
+      //
+      // We check against two previous counters here because in the common case
+      // where deltas are being displayed as a counter track in the UI, we want
+      // to "reset" the counter to zero when it returns to zero (so we don't
+      // keep showing a non-zero value), but don't then need a long stream of
+      // zeroes after that.
+      std::vector<double> prev_vals = new_rows_track->val;
+      auto size = prev_vals.size();
```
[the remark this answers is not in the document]
LalitMaganti: +1 this should be const&

src/trace_processor/perfetto_sql/stdlib/counters/intervals.sql
```diff
@@ -15,7 +15,10 @@
 --
 
 -- For a given counter timeline (e.g. a single counter track), returns
--- intervals of time where the counter has the same value.
+-- intervals of time where the counter has the same value. For every run
```
aMayzner, line 18: Please edit the data example at the bottom of the comment to reflect this change (the comment suggested that there will be two rows emitted)

src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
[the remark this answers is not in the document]
simonmacm: Oops, one day I will stop writing Java in C++. Done.

src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
simonmacm: Done.

src/trace_processor/perfetto_sql/stdlib/counters/intervals.sql
simonmacm: Done.

PR #2929, opened 2025-09-12
stdlib: Share SVG counter y-axis within track groups

zezeozue:
The previous sharing was per SVG which was too broad and doesn't allow groups of counter tracks to share axis e.g CPU frequency tracks and power rail tracks. The CPU freq ones should share an axis and power rails should share an axis.

Test: tools/diff_test_trace_processor.py out/android/trace_processor_shell --name-filter '.*svg.*'


zezeozue:
Seems this introduced a bug investigating

LalitMaganti:
Given the direction Brush has gone, do we want this anymore? Can we close?

PR #2931, opened 2025-09-12
base: Add UnixSocketWatch for inotify-watch creation

primiano:
Adds a class to reconnect immediately when the initial
connection fails, rather than relying on the exponential
backoff (which is still there as a safety net)

### DIFF
--- a/src/base/unix_socket.cc
+++ b/src/base/unix_socket.cc
@@ -1335,7 +1335,7 @@ std::unique_ptr<LinuxFileWatch> WatchUnixSocketCreation(
     TaskRunner* task_runner,
     const char* sock_name,
     std::function<void()> callback) {
-  if (!base::flags::use_unix_socket_inotify)
+  if constexpr (!PERFETTO_FLAGS(USE_UNIX_SOCKET_INOTIFY))
     return nullptr;
 
   if (!sock_name || base::GetSockFamily(sock_name) != base::SockFamily::kUnix ||
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -24,7 +24,9 @@ Unreleased:
     * Updated TraceConfig proto to allow referring to target buffers by name
       rather than index.
   SDK:
-    *
+    * Added inotify support for UNIX sockets: if traced is unavailable or
+      crashes, on Linux and Android we reconnect immediately after the socke
+      has been recreated rather than relying on periodic retries.
 
 
 v53.0 - 2025-11-12:


primiano:
@rsavitski mind taking an initial look (no rush)? i'll do the SDK part later (gotta catch my train)

src/base/unix_socket_unittest.cc
```diff
@@ -1182,6 +1182,67 @@ TEST_F(UnixSocketTest, ShmemSupported) {
...
+  auto watch1 = UnixSocketWatch::WatchUnixSocketCreation(
+      &task_runner_, "127.0.0.1:8080", []() {});
+
+  auto watch2 = UnixSocketWatch::WatchUnixSocketCreation(
+      &task_runner_, "vsock://1:1000", []() {});
+
+  ASSERT_EQ(watch1, nullptr);
+  ASSERT_EQ(watch2, nullptr);
+}
+#endif  // OS_LINUX || OS_ANDROID || OS_MAC
```
sashwinbalaji: typo: if is for linux and android.

sashwinbalaji:
out of curiosity: why not use std::filesystem, is it due to binary size or for being an overkill ?

LalitMaganti:
> out of curiosity: why not use std::filesystem, is it due to binary size or for being an overkill ?

Has lots of problems, look at Google internal style guide.

primiano:
Okay this is good to go now!

src/base/unix_socket.cc
```diff
@@ -1292,5 +1298,98 @@ void UnixSocket::EventListener::OnConnect(UnixSocket*, bool) {}
...
+
+  ScopedFile inotify_fd(inotify_init1(IN_CLOEXEC | IN_NONBLOCK));
+  if (!inotify_fd) {
+    PERFETTO_DLOG("inotify_init() failed");
+    return nullptr;
+  }
+
+  {
+    // The {} block bounds the lifetime of sock_dir.
+    // We need to use strdup because dirname() and basename() on Linux
```
rsavitski: nit: just say "duplicate the string", this particular wording made me ctrl-f for occurences of strdup, just to find none.

src/base/unix_socket.cc
```diff
@@ -1292,5 +1298,98 @@ void UnixSocket::EventListener::OnConnect(UnixSocket*, bool) {}
...
+    // manipulate the source string.
+    base::StackString<255> sock_name_copy("%s", sock_name);
+    const char* sock_dir = dirname(sock_name_copy.mutable_data());
+    if (inotify_add_watch(*inotify_fd, sock_dir, IN_CREATE) < 0) {
```
rsavitski: Could you clarify (both here and in a comment) why we're watching the enclosing directory and not the exact path? Some brittleness gotchas around renames and such (IN_MOVED_TO)?

Also, I suppose this devolves into us inotifying on every new file under `/tmp/` when running on linux? Probably fine.

src/traced/probes/probes_producer.cc
```diff
@@ -415,13 +417,23 @@ void ProbesProducer::OnConnect() {
...
-                                connection_backoff_ms_);
+
+  auto reconnect_task = [weak_this = weak_factory_.GetWeakPtr()] {
+    if (weak_this && weak_this->state_ == kNotConnected)
+      weak_this->Connect();
+  };
+
+  task_runner_->PostDelayedTask(reconnect_task, connection_backoff_ms_);
+  if (!sock_inotify_) {
+    sock_inotify_ = base::UnixSocketWatch::WatchUnixSocketCreation(
```
rsavitski: In which setups would traced_probes have a wildly different lifetime to that of traced? Or is it to increase code coverage? Should we add this to traced_perf, heapprofd and traced_relay later?

src/ipc/client_impl.cc
```diff
@@ -153,21 +155,38 @@ bool ClientImpl::SendFrame(const Frame& frame, int fd) {
 
 void ClientImpl::OnConnect(base::UnixSocket*, bool connected) {
```
rsavitski, line 157: For my sake: it's a bit odd that we try to connect in an "OnConnect" callback. Is there a piece of code I'm overlooking that always fires this once for ipc clients?

src/ipc/client_impl.cc
```diff
@@ -153,21 +155,38 @@ bool ClientImpl::SendFrame(const Frame& frame, int fd) {
...
   if (!connected && socket_retry_) {
```
rsavitski, line 164: nit: a short comment saying that we attempt two reconnect strategies in parallel would help. Otherwise an unfamiliar reader might take a bit of time figuring out what's happening here.

src/base/unix_socket.cc
primiano: > Could you clarify (both here and in a comment) why we're watching the enclosing directory and not the exact path? Some brittleness gotchas around renames and such (IN_MOVED_TO)?

Done

> Also, I suppose this devolves into us inotifying on every new file under /tmp/ when running on linux? Probably fine.

yes. but also soon we'll use more and more /run/perfetto with the debian work Ashwin is doing

src/traced/probes/probes_producer.cc
primiano: > In which setups would traced_probes have a wildly different lifetime to that of traced?

This is mainly to make sure that if the user starts traced_probes independently (before traced) it connects immediately, to avoid hard-to-explain gaps due to the backoff.

> Or is it to increase code coverage? 

Also this

> Should we add this to traced_perf, heapprofd and traced_relay later?

Hah, fair, we should to traced_perf probably.

heapprofd, not worth it. on linux the way it's used is via LD_PRELOAD that should happen always after traced.
traced_relay... it feels it only matters for vm deployments so I'd be tempted to say it's not worth it making it more complicated.

src/ipc/client_impl.cc
primiano: The deal here is that if a connection fails, we still call "OnConnect(connected=false)".
It's a bit weird but was a way for me to force people to think about the disconnection code paths and keep it distinct from "OnDisconnect" (which assumes the connection succeeded in the first place) .
Probably  I should have  created a dedicated OnConnectionFail rather than mixing into OnConnect

src/ipc/client_impl.cc
primiano: Done.

primiano:
OK I addressed everything, but sorry, I just realized that there is 0 reason why this thing is called UnixSocketWatch. There is nothing unix-socket specific.

Let me move this into "FileWatch" in a dedicated file. One more round of patience

primiano:
@rsavitski this is ready now

include/perfetto/ext/base/file_utils.h
```diff
@@ -115,6 +118,39 @@ std::optional<uint64_t> GetFileSize(const std::string& path);
 // Returns the size of the open file |fd|, or nullopt in case of error.
 std::optional<uint64_t> GetFileSize(PlatformHandle fd);
 
+// FileWatch monitors filesystem-linked file creation using inotify.
+// This class uses inotify (on Linux/Android) to watch for the creation of
```
rsavitski, line 146: nit: this line starts off by repeating the previous line. And the comment in general is quite verbose and repetitive. Please do an editing pass. This includes the repetition in the comment on `WatchFileCreation()`.

src/profiling/common/proc_cmdline.cc
```diff
@@ -24,6 +24,7 @@
 #include <unistd.h>
 
 #include "perfetto/ext/base/file_utils.h"
+#include "perfetto/ext/base/utils.h"
```
rsavitski: Please keep this cl focused. This change looks completely unrelated.

include/perfetto/ext/base/file_utils.h
```diff
@@ -115,6 +118,39 @@ std::optional<uint64_t> GetFileSize(const std::string& path);
...
+// special file types).
+// It's only supported on Linux and Android, it's a no-op (returns nullptr) on
+// other platforms.
+//
+// Usage:
+//   auto watch = FileWatch::WatchFileCreation(
+//       task_runner, "/tmp/my_file", []() {
+//         // Called when /tmp/my_file is created
+//       });
+class FileWatch {
```
rsavitski: Maybe rename to `LinuxFileWatch` so that its purpose is clearer when added as a member to existing classes? Right now it looks like:
```
  std::unique_ptr<base::FileWatch> sock_inotify_;
```
Which doesn't suggest that it's a platform-specific (and therefore optional) piece of functionality (even though it comes from base/).

WDYT about something like this instead? Feel free to keep as is if this is looking like a bikeshed.
```
  std::unique_ptr<base::LinuxFileWatch> sock_watch_;
```

src/base/file_utils.cc
```diff
@@ -503,5 +512,119 @@ std::optional<uint64_t> GetFileSize(PlatformHandle fd) {
...
 }
 
+// FileWatch
+
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX_BUT_NOT_QNX) || \
+    PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+
+namespace {
+
+// Implementation class extends FileWatch
```
rsavitski: nit: useless comment since it describes the next line verbatim

src/base/file_utils.cc
```diff
@@ -503,5 +512,119 @@ std::optional<uint64_t> GetFileSize(PlatformHandle fd) {
...
+            esize =
+                static_cast<ssize_t>(sizeof(struct inotify_event) + evt->len);
+            if (evt->len > 0 && (evt->mask & IN_CREATE)) {
+              if (weak_handle->file_base_name_ == evt->name) {
+                weak_handle->task_runner_->PostTask(weak_handle->callback_);
+                return;
+              }
+            }
+          }  // for(evt);
+        });
```
rsavitski, line 623:
```suggestion
    task_runner_->AddFileDescriptorWatch(
        *inotify_fd_, [weak_handle = weak_ptr_factory_.GetWeakPtr()] {
          if (!weak_handle)
            return;
          alignas(inotify_event) char buf[4096];
          ssize_t rsize =
              base::Read(*weak_handle->inotify_fd_, buf, sizeof(buf));
          if (rsize <= 0)
            return;
          for (ssize_t i = 0; i < rsize;) {
            auto* evt = reinterpret_cast<inotify_event*>(buf + i);
            if (evt->len > 0 && (evt->mask & IN_CREATE) &&
                (weak_handle->file_base_name_ == evt->name)) {
              weak_handle->task_runner_->PostTask(weak_handle->callback_);
              return;
            }
            i += static_cast<ssize_t>(sizeof(inotify_event) + evt->len);
          }  // for(evt);
        });
```

src/base/file_utils.cc
```diff
@@ -503,5 +512,119 @@ std::optional<uint64_t> GetFileSize(PlatformHandle fd) {
...
+    if (inotify_add_watch(*inotify_fd, file_dir, IN_CREATE) < 0) {
+      PERFETTO_DLOG("inotify_add_watch(%s) failed", file_dir);
+      return nullptr;
+    }
+  }
+  std::string file_base_name;
+  {
+    base::StackString<255> path_copy("%s", path);
+    file_base_name = basename(path_copy.mutable_data());
+  }
```
rsavitski: I'm looking at the manpages for libgen.h and the basename situation looks like a huge mess - you get either the GNU or the posix implementation depending on the header included, plus this code having to use base::StackString and scoped blocks...

WDYT, is it that difficult for us to implement basename/dirname/split in utils, and sidestep this entire library function? Unless I'm missing something, is it not just a split on the last slash?

rsavitski (changes requested):

include/perfetto/ext/base/file_utils.h
primiano: done

include/perfetto/ext/base/file_utils.h
primiano: Good point. renamed.

src/base/file_utils.cc
primiano: Fair, accepted % moving the increments soon after the evt, to avoid later bitrot.

src/base/file_utils.cc
primiano: Good point. I have splitted that in #3655

PR #2943, opened 2025-09-15
Display perf_samples that doesn't have a callsite_id

haixiaosugoogle:
Perf samples without callsite_ids are emitted when the perf_sample skips unwinding. However, these should still be displayed as we would be able to differentiate processes without perf_samples and processes with skipped perf_samples. It also gives us a more accurate accounting for the amount of work that the process is actually doing.

### DIFF
--- a/ui/src/plugins/dev.perfetto.CpuFreq/index.ts
+++ b/ui/src/plugins/dev.perfetto.CpuFreq/index.ts
@@ -15,7 +15,6 @@
 import m from 'mithril';
 import {PerfettoPlugin} from '../../public/plugin';
 import {Trace} from '../../public/trace';
-import {CPU_FREQ_TRACK_KIND} from '../../public/track_kinds';
 import {TrackNode} from '../../public/workspace';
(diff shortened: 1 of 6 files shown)


rsavitski:
As discussed offline:
* Skipped samples will need an entirely different UI element, or users will be confused as to why there are tracks full of perf samples that show nothing when selected (especially in systemwide sampling).
* I'm personally unconvinced that rendering skipped samples (effectively = thread X was running at time Y) is adding more signal than noise in most traces:
  * If you have scheduling information, then you already know which threads are running when. 
  * If you are profiling a single process, then the fact that we do per-cpu profiling underneath is an implementation detail. Yet with this change, it would make the UI start showing tracks for /most/ processes on the system.

The types of traces where showing skipped samples is useful are when it's the only (loose) proxy for what's running on the device.

To me this still feels better as a UI command (entered on-demand via ctrl-shift-p), e.g. "render all skipped samples". So by default we do not pollute the UI with skipped samples, but people with a specific need to see them can run the command.

Let me get a second opinion within the team and get back to you.

PR #2944, opened 2025-09-15
ui: Remove buttons entirely from track shell when not hovered

stevegolton:
Currently, some track shell buttons appear only on hover - which avoids cluttering the UI visually. However the invisible buttons still take up space in the shell which is could be otherwise used for the title.

When titles are very long it would be nice if we could use that space for the title instead of the buttons, especially if the track has a chip on it too (e.g. main thread), which takes up even more space.

This patch makes it so that the buttons are actually removed (display: none) when the track shell is not hovered, so the space may be used for the title.

The one bit of UX that I think is essential is that the permanent track buttons (those that remain even when the track is not being hovered) remain in the same place, so that they don't move when you go to click them, which would be maddening. In order to preserve this feature while also removing the hover-only buttons, I've had to change the order of the track shell buttons so that the permanent ones are always on the right. This could be a little jarring, but I don't see any other way to do it.

# Screenshots

## More room for long track names:

Before:
image

After:
image

## Note: Reordered buttons:

Before:
image

After:
image

### DIFF
--- a/ui/src/assets/widgets/middle_ellipsis.scss
+++ b/ui/src/assets/widgets/middle_ellipsis.scss
@@ -16,6 +16,7 @@
   display: flex;
   flex-direction: row;
   overflow: hidden;
+  max-width: fit-content;
 
   .pf-middle-ellipsis-left {
     overflow: hidden;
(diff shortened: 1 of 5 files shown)


rsavitski:
Clarifying the state: I gave offline feedback that it'd be better if the underlying text was truncated instead of re-ellipsising, as the latter makes the text jump around quite a lot on mouseover.

stevegolton:
Superseded by https://github.com/google/perfetto/pull/4750

Issue #1256, opened 2025-04-12
Self time in the slice table


Fabian446:
Hello!
Currently, if you for example create a Table slice with all the block with the same name, there is a column with the duration that you can sort by. I understand that this is the wall duration.
Would it also be possible to include the self time of the block too?

LalitMaganti:
There was a discussion about this on the chrometto channel on our Discord about this. I guess it was with you.

Yes it makes sense for us to do this, we probably won't put this in the trace processor slice table itself (for memory reasons) but displaying it in the UI on demand makes sense.

@stevegolton FYI

stevegolton:
This feature has been added in #2946

PR #2946, opened 2025-09-15
ui: Add self-time (aka self-duration) to slice aggregation

stevegolton:
image


LalitMaganti (approved):
Awesome! So cool! LGTM :)

LalitMaganti:
Can we also mark https://github.com/google/perfetto/issues/1256 as fixed?

stevegolton:
Benchmarks seem to about double the time taken to calculate the aggregation (less the time taken to build the initial ii table). I'll take a look tomorrow to see if there are any improvements we can make before landing.

Edit: With the ii table creation it seems to only take about 25% longer. I think this is probably worth it for the extra functionality.

PR #2953, opened 2025-09-15
tp: one more attempt to fix compile on windows

LalitMaganti:
Compiler cannot see that it's not no-return on other branch of if constexpr

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/utils.h
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/utils.h
@@ -18,14 +18,27 @@
 #define SRC_TRACE_PROCESSOR_PERFETTO_SQL_INTRINSICS_FUNCTIONS_UTILS_H_
 
 #include <sqlite3.h>
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
+#include <cstring>
 #include <limits>
+#include <memory>
 #include <string>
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/intrinsics/functions/utils.h
```diff
@@ -448,12 +465,14 @@ struct RegexpExtract : public sqlite::Function<RegexpExtract> {
       return sqlite::result::TransientString(
           ctx, result_sv.data(), static_cast<int>(result_sv.size()));
     } else {
-      PERFETTO_FATAL("Regex not supported");
+      // Always-true branch to avoid spurious no-return warnings.
+      if (ctx) {
+        PERFETTO_FATAL("Regex not supported");
+      }
```
sashwinbalaji, line 471: how about just using macro #if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN) or creating one for regex ?

LalitMaganti: Not sure how this helps things, the compiler will still show warnings in that case?

sashwinbalaji: ahh sorry, misunderstood the problem

sashwinbalaji (approved):

PR #2956, opened 2025-09-16
tools: add experimental script for generating relnotes and move scripts

LalitMaganti:
This CL adds an experimental script containing prompts for generating
the changelog and release notes for Perfetto releases. It also
consolidates all the releasing scripts into a single folder to make them
less scattered in the tools folder.


sashwinbalaji (approved):
is it correctly categorizing the changes ? 

we usually prefix the commit with tp, ui, ftrace, etc. should we add this to input information to make it easier

LalitMaganti:
> is it correctly categorizing the changes ?

Yes it's able to figure it out. Anyway this is an experimental thing, I want to land it and iterate on it for next few releases before we decide to keep or delete it.

PR #2960, opened 2025-09-16
Speed up android.kernel_wakelocks using counter_leading_intervals().

simonmacm:
For a typical long trace this goes from 9.4s to 1.6s.

Bug: http://b/445123061

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/kernel_wakelocks.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/kernel_wakelocks.sql
@@ -32,12 +32,13 @@ WITH
     SELECT
       *
     FROM counter_leading_intervals!((
-        select 
-          c.id,
-          c.ts,
-          c.track_id,
-          c.value
-        from counter c join _kernel_wakelock_track t on t.id = c.track_id
+        SELECT
+          id,
+          ts,
+          track_id,
+          value
+        FROM counter
+        WHERE track_id IN (SELECT id FROM _kernel_wakelock_track)
     ))
   )
 SELECT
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/stdlib/android/kernel_wakelocks.sql
```diff
@@ -15,33 +15,42 @@
...
-    WHERE
-      t.type = 'android_kernel_wakelock'
+      *
+    FROM counter_leading_intervals!((
+        select 
+          c.id,
+          c.ts,
+          c.track_id,
+          c.value
+        from counter c join _kernel_wakelock_track t on t.id = c.track_id
```
LalitMaganti: Writing this as a `WHERE c.track_id IN (select id from _kernel_wakelock_track)` will be more efficient.

src/trace_processor/perfetto_sql/stdlib/android/kernel_wakelocks.sql
```diff
@@ -15,33 +15,42 @@
...
   ts,
   ts AS original_ts,
-  next_ts - ts AS dur,
+  dur,
   name,
   hash(name) AS name_int,
   type,
-  value - lag_value AS held_dur
-FROM raw;
+  next_value - value AS held_dur
```
LalitMaganti, line 51: You're changing the semantic of this. Was that intentionaL If so maybe make a note in the CL description?

src/trace_processor/perfetto_sql/stdlib/android/kernel_wakelocks.sql
simonmacm: Neat, thanks!

src/trace_processor/perfetto_sql/stdlib/android/kernel_wakelocks.sql
simonmacm: Actually not, the (pointless) ts shuffling in line 24 of the old version means they're equivalent - the tests for this code didn't change and I've also manually verified it against a real trace.

PR #2964, opened 2025-09-16
tools: push release branch with --no-verify

LalitMaganti:
Necessary because the amalgamated sources should intentionally not be covered by presubmit checks


sashwinbalaji (approved):
nit: please add PR description on why this is needed

PR #2971, opened 2025-09-16
Add 'base::CopyFileContents' function.

KirillTim:
We doesn't expect `CopyFileContents` to be a bottleneck, so the implementation is the same for all platforms.
In the future it may be useful to add a platform-specific implementations, that uses `sendfile` or even `copy_file_range` APIs.

`CopyFile` have three different ways to fail, so it returns human readable `base::Status` instead of just `bool` or `ssize_t`.


src/base/file_utils.cc
```diff
@@ -188,6 +189,48 @@ ssize_t WriteAll(int fd, const void* buf, size_t count) {
...
+        "Can't get offset in 'fd_in', lseek error: %s (errno: %d)",
+        strerror(errno), errno);
+  }
+
+  auto restore_offset_on_exit = OnScopeExit([fd_in, original_offset] {
+    off_t ret = lseek(fd_in, original_offset, SEEK_SET);
+    // 'lseek' should never fail here, but if it fails we crash, to prevent
+    // possible data loss/overwrite in the 'fd_in'.
+    PERFETTO_CHECK(ret >= 0);
+  });
```
primiano: this is very nice (the use of OnScopeExit), well done!
maybe just make it in one line (CHECK(lseek(...) >= 0)), so if it fails the error is easier to codesearch 

However this should go after the next block (The one that does SEEK_SET), not before.

Imagine this happens:
- the user calls copy file on a read-only file
- the initial lseek(SEEK_CUR) succeeds
- The SEEK_SET below fails, you (try to) return an ErrStatus (which is the right thing)
- But as you return, the OnScopeExit tries to do a SEEK_SET, which will fail, and crashes instead

Also intuitively you should restore the initial offset only after you have successfully manipulated, not before

src/base/file_utils.cc
```diff
@@ -188,6 +189,48 @@ ssize_t WriteAll(int fd, const void* buf, size_t count) {
...
+  if (lseek(fd_in, 0, SEEK_SET) == -1) {
+    return base::ErrStatus(
+        "Can't change the offset in 'fd_in', lseek error: %s (errno: %d)",
+        strerror(errno), errno);
+  }
+
+  // Use bigger buffer when copy files.
+  constexpr size_t kCopyFileBufSize = 32 * 1024;  // 32KB.
+  static_assert(kCopyFileBufSize > kBufSize);
+  char buf[kCopyFileBufSize];
```
primiano: 32KB on the stack is too much, especially for a base funciton that could be called eslewhere in future.
Last time I did somethign similar, it started failing on random places (e.g. b/194510398#comment5).
IIRC google3's fibers have a 64K stack by default.

Allocate this buffer on the heap outide of the loop.

src/base/utils_unittest.cc
```diff
@@ -215,6 +218,122 @@ TEST(UtilsTest, HexDump) {
...
+TEST(UtilsTest, CopyFile) {
+  auto assert_file_content_fn = [&](TempFile& file,
+                                    const std::string& expected) {
+    // Flush content of the original dst FD.
+    ASSERT_TRUE(FlushFile(*file));
+    std::string actual;
+    // Create new FD from the path with zero offset.
+    base::ScopedFile ro_fd = base::OpenFile(file.path(), O_RDONLY);
+    ASSERT_TRUE(ReadFileDescriptor(*ro_fd, &actual));
+    ASSERT_EQ(expected, actual);
```
primiano, line 229: you can just use base's `bool ReadFile(const std::string& path, std::string* out)`

src/base/utils_unittest.cc
```diff
@@ -215,6 +218,122 @@ TEST(UtilsTest, HexDump) {
...
+    ASSERT_EQ(GetFileSize(*ro_fd).value(), expected.size());
```
primiano: this is redundant as the strings can't be equal if the size isn't  equal

src/base/utils_unittest.cc
```diff
@@ -215,6 +218,122 @@ TEST(UtilsTest, HexDump) {
...
+    }
+
+    ScopedFile src = OpenFile("/dev/zero", O_WRONLY);
+    {
+      Status result = CopyFile(*src, *dst);
+      // CopyFile can't read from write only fd.
+      ASSERT_THAT(result.message(), StartsWith("Read failed:"));
+    }
+  }
+#endif
```
primiano: There isn't much point writing a test for something we do NOT support.
if it' unsupported... is unsupported
Remove this

KirillTim:
@primiano 
Apart from code review fixes, 
I've renamed the function to `CopyFileContents`, because `CopyFile` is a macro on Windows.
`UtilsTest.CopyFileContents` now passes on Windows.
Also add an additional test to make sure we can copy `/proc/` files, inspired by: https://source.chromium.org/chromium/chromium/src/+/main:base/files/file_util_unittest.cc;l=5092;drc=e3e576a7d7851b19fae5f947ebab554d2996db8e

src/base/utils_unittest.cc
```diff
@@ -215,6 +218,127 @@ TEST(UtilsTest, HexDump) {
...
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX) || \
+    PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+  // Test CopyFileContents can read special files with zero length that might
+  // actually have contents (such as a seq_file).
+  // This test is important, because 'sendfile' can't read from such files, and
+  // we want to make sure that if/when we implement the platform specific
+  // optimization we won't forget about this special case.
+  {
+    for (auto* const file : {"/proc/meminfo", "/proc/self/cmdline",
+                             "/proc/self/environ", "/proc/self/auxv"}) {
```
primiano, line 318: just pick one file? the more file you pick the more liekly this test is going to fail due to some selinux.
I'd jsut do /proc/self/cmdline as /proc/self is generally safe

primiano (approved):
LGTM with one final comment thanks!

PR #2994, opened 2025-09-18
tp: Introduce view capture extractor.

PriyankaSPX:
To process views in DFS order for visibility computation.

Bug: 438683146
Test: tools/ninja -C out/linux_clang_debug perfetto_unittests && out/linux_clang_debug/perfetto_unittests --gtest_filter=ViewCapture*

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -14464,6 +14464,7 @@ filegroup {
         "src/trace_processor/importers/proto/winscope/surfaceflinger_transactions_parser.cc",
         "src/trace_processor/importers/proto/winscope/viewcapture_args_parser.cc",
         "src/trace_processor/importers/proto/winscope/viewcapture_parser.cc",
+        "src/trace_processor/importers/proto/winscope/viewcapture_views_extractor.cc",
         "src/trace_processor/importers/proto/winscope/winscope_geometry.cc",
         "src/trace_processor/importers/proto/winscope/winscope_module.cc",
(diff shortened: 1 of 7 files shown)


protos/perfetto/trace/android/BUILD.gn
```diff
@@ -112,6 +112,17 @@ perfetto_proto_library("winscope_extensions_@TYPE@") {
   import_dirs = [ "${perfetto_protobuf_src_dir}" ]
 }
 
+# Winscope messages added to TracePacket directly
+perfetto_proto_library("winscope_unittest_@TYPE@") {
```
LalitMaganti, line 116: Don't understand the need for this. Why are we not adding gen to the list of proto generators above?

PriyankaSPX: privacy.proto cannot be generated as it contains defaults which are not supported:

```
message PrivacyFlags {
  optional Destination dest = 1 [default = DEST_UNSET];
   ...
}
```

LalitMaganti: Maybe let's add support for that instead of introducing workarounds like this which won't scale? one day I'm pretty certain with how much you use defaults that you are going to need this anyway?

LalitMaganti (approved):

protos/perfetto/trace/android/BUILD.gn
PriyankaSPX: picking this up on the other PR

PR #2997, opened 2025-09-18
Introduce ViewCapture test utils.

PriyankaSPX:
Bug: 438683146
Test: tools/ninja -C out/linux_clang_debug perfetto_unittests && out/linux_clang_debug/perfetto_unittests

### DIFF
--- a/protos/perfetto/trace/android/BUILD.gn
+++ b/protos/perfetto/trace/android/BUILD.gn
@@ -112,6 +112,17 @@ perfetto_proto_library("winscope_extensions_@TYPE@") {
   import_dirs = [ "${perfetto_protobuf_src_dir}" ]
 }
 
+# Winscope messages added to TracePacket directly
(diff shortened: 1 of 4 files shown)


protos/perfetto/trace/android/BUILD.gn
```diff
@@ -112,6 +112,17 @@ perfetto_proto_library("winscope_extensions_@TYPE@") {
   import_dirs = [ "${perfetto_protobuf_src_dir}" ]
 }
 
+# Winscope messages added to TracePacket directly
+perfetto_proto_library("winscope_unittest_@TYPE@") {
```
LalitMaganti, line 116: As discussed before, still applies here.

PriyankaSPX: turns out we can get around this by creating protos from protozero, don't need the cpp target

PR #2999, opened 2025-09-18
docs: ftrace: actually fix __assign_str macro comment

rsavitski:
Fixes: f8afd4d7c0 docs: clarify syntax for __string usage in ftrace kernel documentation (#2993)
Fixes: ce4a52ec16 docs: fix typo in ftrace.md (#2988)

### DIFF
--- a/docs/getting-started/ftrace.md
+++ b/docs/getting-started/ftrace.md
@@ -504,6 +504,9 @@ and do not need to be specified.
 
 Tracepoint declaration example, named `trk_example/tid_track_example`:
 
+_Note: on kernels older than v6.10, \_\_assign_str requires two arguments, see
+[this patch](https://lore.kernel.org/linux-trace-kernel/20240516133454.681ba6a0@rorschach.local.home/)._
+
 ```h
 // trace/events/trk_example.h
 #undef TRACE_SYSTEM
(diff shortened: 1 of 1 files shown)


docs/getting-started/ftrace.md
```diff
@@ -522,20 +522,12 @@ TRACE_EVENT(tid_track_example,
...
-        // ```
-        //
-        // [1] https://lore.kernel.org/linux-trace-kernel/20240516133454.681ba6a0@rorschach.local.home/
-        __string(slice_name)
+        __string(slice_name, slice_name)
     ),
     TP_fast_assign(
         __entry->track_event_type = track_event_type;
-        __assign_str(slice_name, slice_name);
+        /* on kernels before v6.10: __assign_str(slice_name, slice_name) */
```
LalitMaganti, line 529: The increased commentary above was intentional. Could we keep the context, I think it's helpful to have this.

LalitMaganti (approved):

docs/getting-started/ftrace.md
rsavitski: I absolutely think this is sufficient. Kernel code -> kernel APIs are known to change -> if this fails to build and the comment states that there's a different signature on older kernels, that's enough info to both fix the code and to find out more via "git log" (for the curious). I don't want to pollute the primary subject of the example with a multiline comment for every possible kernel version compat issue.

As a compromise, I could add the extra context above the code block (as an italicised note).

LalitMaganti: I always find links like LKML discussions interesting to see the context for why something changed but I appreciate it distracts from the main flow of the text. Maybe a footnote to bottom of the doc or a note above the codeblock instead as you suggested?

rsavitski: Done.

PR #3000, opened 2025-09-18
tp: Compute visibility for viewcapture views.

PriyankaSPX:
Bug: 438683146
Test: tools/ninja -C out/linux_clang_debug perfetto_unittests && out/linux_clang_debug/perfetto_unittests --gtest_filter=ViewCapture*

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -14465,6 +14465,7 @@ filegroup {
         "src/trace_processor/importers/proto/winscope/viewcapture_args_parser.cc",
         "src/trace_processor/importers/proto/winscope/viewcapture_parser.cc",
         "src/trace_processor/importers/proto/winscope/viewcapture_views_extractor.cc",
+        "src/trace_processor/importers/proto/winscope/viewcapture_visibility_computation.cc",
         "src/trace_processor/importers/proto/winscope/winscope_geometry.cc",
         "src/trace_processor/importers/proto/winscope/winscope_module.cc",
         "src/trace_processor/importers/proto/winscope/winscope_rect_tracker.cc",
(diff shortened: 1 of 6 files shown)


protos/perfetto/trace/android/BUILD.gn
```diff
@@ -112,6 +112,17 @@ perfetto_proto_library("winscope_extensions_@TYPE@") {
   import_dirs = [ "${perfetto_protobuf_src_dir}" ]
 }
 
+# Winscope messages added to TracePacket directly
+perfetto_proto_library("winscope_unittest_@TYPE@") {
```
LalitMaganti, line 116: This seems to duplicate other patch. Shouldn't one depend on the other to make this possible to review properly?

PriyankaSPX: it's the same commit in both patches? i thought you preferred to review in separate patches but this patch depends on the same commit 0720195e7c4a013be074b54df964c7185783f735 in https://github.com/google/perfetto/pull/2994

LalitMaganti: From go/perfetto-github-instructions:

```
Working with stacks of CL
Keep doing the usual thing of having

origin/main
  first_cl
     second_cl
        third_cl

Created with:
git checkout -b second_cl -t first_cl
```

That is, the way we do stacked CLs is to have branches . So if A -> B -> C, then B depends on branch containing A and C depends on branch containing B. AFAIK this is how we've always done things since we moved to Github?

PriyankaSPX: so when the stack is uploaded to github you want commit A -> commit B -> commit C all in one patch/PR ?

LalitMaganti: Each branch becomes a different patch. So yes each later branch has all commits of earlier ones but Github just hides them

PriyankaSPX: don't understand what's duplicated here then.... or how i can get the other commits hidden on this branch. I've already uploaded the patches separately: 

A: https://github.com/google/perfetto/pull/2997

-> B: https://github.com/google/perfetto/pull/2994

-> C: https://github.com/google/perfetto/pull/3000

LalitMaganti: In B, the upstream claims to be main not A:
image

You need to set the upstream branch to A's PR. Same with C and B.

PriyankaSPX: thank you!! i've updated the branches accordingly.

PriyankaSPX: although this seems to be causing a presubmit failure

LalitMaganti (approved):

PR #3003, opened 2025-09-18
stdlib: add default anr durations to android_anrs table

gignat-dev:
Add default anr durations to android_anrs table

When the anr_dur can't be computed from the timer expiration event, we fallback to the default anr durations for common ANR types (default means in AOSP/Pixel). Actual values can vary by OEM or Android version.

Test:tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell --name-filter '.*anr.*'

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/anrs.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/anrs.sql
@@ -57,6 +57,71 @@ SELECT
     ELSE 'UNKNOWN_ANR_TYPE'
   END;
 
+CREATE PERFETTO FUNCTION _get_broadcast_flag(
+    subject STRING
+)
+RETURNS LONG AS
+SELECT
+  CASE
+    WHEN $subject IS NULL OR NOT $subject GLOB 'Broadcast of Intent *flg=*'
+    THEN NULL
+    ELSE unhex(str_split(substr($subject, instr($subject, ' flg=') + 5), ' ', 0))
+  END AS flag;
+
+-- Function to get the default ANR duration in milliseconds based on ANR type.
+-- Note: These are common defaults. Actual timeouts can vary by OEM, Android version, and FG/BG status.
+CREATE PERFETTO FUNCTION _default_anr_dur(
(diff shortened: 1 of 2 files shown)


src/trace_processor/perfetto_sql/stdlib/android/anrs.sql
```diff
@@ -57,6 +57,71 @@ SELECT
...
+RETURNS LONG AS
+SELECT
+  CASE
+    WHEN $subject IS NULL OR NOT $subject GLOB 'Broadcast of Intent *flg=*'
+    THEN NULL
+    ELSE unhex(str_split(substr($subject, instr($subject, ' flg=') + 5), ' ', 0))
+  END AS flag;
+
+-- Function to get the default ANR duration in milliseconds based on ANR type.
+-- Note: These are common defaults. Actual timeouts can vary by OEM, Android version, and FG/BG status.
```
LalitMaganti, line 72: If thisn is the case, I'd like for this to be a different column. We should not be giving these things unless we know with high probabilit hthis is the case. Otherwise people will open bugs and complain 100% guaranteed

gignat-dev: done

src/trace_processor/perfetto_sql/stdlib/android/anrs.sql
```diff
@@ -97,7 +162,9 @@ CREATE PERFETTO VIEW android_anrs (
   -- The standard type of ANR.
   anr_type STRING,
   -- Duration of the ANR, computed from the timer expiration event.
-  anr_dur_ms LONG
+  anr_dur_ms LONG,
+  -- Default duration of the ANR, based on the anr_type (default means in AOSP/Pixel).
```
LalitMaganti, line 166: Please explicitly add a note that other OEMs may have customized these values so this might not be accurate for them.

gignat-dev: done

PR #3009, opened 2025-09-18
Use os.replace instead of os.rename

jessemckenna:
Perfetto's scripts use `os.rename` to replace files with new versions. However, `os.rename` fails on Windows when the destination file exists.

`os.replace` is equivalent to `os.rename` on POSIX, and on Windows allows replacing an existing file (see [docs](https://docs.python.org/3/library/os.html#os.rename)). This change replaces uses of `os.rename` with `os.replace`, because the scripts don't seem like they intend to fail if the file already exists.

This fixes the following error on Windows:
```
PS C:\src\perfetto> python3 tools/gen_all out/default
Running python3 tools\check_include_violations
Running python3 tools\check_proto_comments
Running python3 tools\fix_include_guards
0 files changed
Running python3 tools\gen_merged_protos
Running python3 tools\gen_amalgamated_python_tools
INFO    : Amalgamating python/tools/record_android_trace.py -> tools/record_android_trace
  File "C:\src\perfetto\tools\gen_amalgamated_python_tools", line 123, in <module>
    sys.exit(main())
             ^^^^^^
  File "C:\src\perfetto\tools\gen_amalgamated_python_tools", line 118, in main
    success = success and amalgamate(src, dst, check_only)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\src\perfetto\tools\gen_amalgamated_python_tools", line 106, in amalgamate
    os.rename(dst + '.tmp', dst)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'tools/record_android_trace.tmp' -> 'tools/record_android_trace'
Error: Command: python3 tools\gen_amalgamated_python_tools failed
```

### DIFF
--- a/infra/ci/worker/sandbox_runner.py
+++ b/infra/ci/worker/sandbox_runner.py
@@ -60,7 +60,7 @@ def create_sandbox_token(token_path):
   ]).decode().strip()
   with open(token_path + '.tmp', 'w') as f:
     f.write(sandbox_svc_token)
-  os.rename(token_path + '.tmp', token_path)
+  os.replace(token_path + '.tmp', token_path)
 
 
 def main():
--- a/infra/ci/worker/sandbox_runner.py
+++ b/infra/ci/worker/sandbox_runner.py
@@ -60,7 +60,7 @@ def create_sandbox_token(token_path):
   ]).decode().strip()
   with open(token_path + '.tmp', 'w') as f:
     f.write(sandbox_svc_token)
-  os.replace(token_path + '.tmp', token_path)
+  os.rename(token_path + '.tmp', token_path)
 
 
 def main():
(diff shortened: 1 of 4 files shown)


LalitMaganti (approved):

infra/ci/worker/sandbox_runner.py
```diff
@@ -60,7 +60,7 @@ def create_sandbox_token(token_path):
   ]).decode().strip()
   with open(token_path + '.tmp', 'w') as f:
     f.write(sandbox_svc_token)
-  os.rename(token_path + '.tmp', token_path)
+  os.replace(token_path + '.tmp', token_path)
```
LalitMaganti, line 63: I don't think we should touch this. This is run inside a docker container for Linux and so should never really be run on Windows.

jessemckenna: Done, thank you for reviewing!

PR #3012, opened 2025-09-18
ui: seperating the display and generation of Wattson warnings

runpinglai-creator:
1. Moved the configuration checking logic out of index.ts and into its own dedicated file, warning.ts, considering that there could be more logics to be added for checking user configuraitons in the future.
2. Added the logics to check atrace_categories and predefined cpu_idle/freq_ms.
3. Let CPU track appear all the time so that user would be able to see the warning messages.

Test: Local Perfetto UI Test
Bug: 444535474


wusamuel6 (approved):

runpinglai-creator:
Hi @LalitMaganti @stevegolton ,

All 15 CI test failures are due to UI Screenshots diffs. The expected pngs are missing Wattson (maybe because they are old screenshots?) while the actual pngs will have Wattson elements. Should the test resources be updated or just skip the tests for now?

Example of old (expected) png, there's no Wattson: 
image

Example of new (actual) png, there is Wattson causing the diff:
image

Thanks,
Runping

LalitMaganti:
Please update the screenshot tests. There are instructions on doing that in the UI docs

runpinglai-creator:
Thanks Lalit for the pointer. It works now.

PR #3016, opened 2025-09-18
traced_perf: add event modifiers such as :u and :k

rsavitski:
It's sometimes useful to configure a perf event to only increment while
in a specific context (userspace vs kernel vs hypervisor). This patch
adds "event modifiers" to both the timebase and the followers to achieve
this, modeled after perftool's
https://man7.org/linux/man-pages/man1/perf-list.1.html#EVENT_MODIFIERS.

The modifiers are recorded in perf_sample_defaults and contribute to the
trace_processor track name as a suffix.

Example data source config:
```
  data_sources {
    config {
      name: "linux.perf"
      perf_event_config {
        timebase {
          frequency: 1000
          counter: SW_CPU_CLOCK
          timestamp_clock: PERF_CLOCK_MONOTONIC
        }
        followers {
          counter: HW_CPU_CYCLES
          modifiers: EVENT_MODIFIER_COUNT_USERSPACE
        }
        followers {
          counter: HW_CPU_CYCLES
          modifiers: EVENT_MODIFIER_COUNT_KERNEL
        }
      }
    }
  }
```

Based on a contribution by: Ali Utku Selen


auselen:
I can't see the advantage of this over just changing RawEvent? To me using RawEvent encapsulate the idea this is some advanced setting. As you mentioned this won't be valid for ftrace events, but also confusing for some Counter like "SW_CONTEXT_SWITCHES".

rsavitski:
> I can't see the advantage of this over just changing RawEvent? To me using RawEvent encapsulate the idea this is some advanced setting. As you mentioned this won't be valid for ftrace events, but also confusing for some Counter like "SW_CONTEXT_SWITCHES".

But it's perfectly valid for some `Counter`s, for example "cycles:k" shouldn't require one to reach for the RawEvent encoding imo. And the reason I went with an array of enum modifiers is to be closer to perftool's modifiers (https://man7.org/linux/man-pages/man1/perf-list.1.html#EVENT_MODIFIERS). Partly for possible future extensions, partly to avoid inventing new concepts and nomenclature.

RawEvent is more of an escape hatch (requiring the user to manually find the right PMU and perf_event_attr config values) since we don't do a great job of exposing device-specific counters and PMUs. I'd prefer to require it less in the future, rather than add features specific to it.

PS: perftool also allows idiosyncratic combinations like "context-switch:u".

src/profiling/perf/event_config.cc
```diff
@@ -243,36 +243,79 @@ int32_t ToClockId(protos::gen::PerfEvents::PerfClock pb_enum) {
...
+      case PerfEvents::EVENT_MODIFIER_COUNT_HYPERVISOR:
+        include_hv = true;
+        has_counting_scope = true;
+        break;
+      default:
+        break;
+    }
+  }
+
+  if (has_counting_scope) {
```
auselen, line 275: This shouldn't be needed, you can just set counter values in the block above? This also forces default values above to other settings, even if I only choose one of them.

auselen:
> ... And the reason I went with an array of enum modifiers is to be closer to perftool's modifiers (https://man7.org/linux/man-pages/man1/perf-list.1.html#EVENT_MODIFIERS). Partly for possible future extensions, partly to avoid inventing new concepts and nomenclature.

I guess it is about if you want to handle the complexity via configuration which requires future-proofing and handling details in the implementation or just pass it to perf_event_open. Both solutions would help me with my tasks as long as functionality is there.

src/profiling/perf/event_config.cc
rsavitski: (fwiw I didn't test the code yesterday, it was a sketch for the discussion)

The linking of the three options is intentional, unless I misunderstood your point. We want "if at least one is set, then the non-included options are excluded":
* COUNT_USERSPACE -> exclude_user=0, exclude_kernel=1, exclude_hv=1
* COUNT_USERSPACE + COUNT_KERNEL -> exclude_user=0, exclude_kernel=0, exclude_hv=1
* etc

And the code is written to be compatible with future EventModifier enum values that do not affect the exclude_* fields.

auselen: Ok, got it. I was still thinking about excluding rather than including.

The way I would be using I guess; (which reflects perf tool and not perf_event_open)
```
  followers {
    raw_event {
      type: 4
      config: 17 # 0X0011
    }
    name: "CPU_CYCLES"
  }
  followers {
    raw_event {
      type: 4
      config: 17 # 0X0011
    }
    name: "CPU_CYCLES:u"
    modifier: COUNT_USERSPACE
  }
```

rsavitski: Yeah, exactly. Though I now made the counter naming automatically add suffixes such as :u and :k as you did in https://github.com/google/perfetto/pull/2601.

rsavitski:
> > ... And the reason I went with an array of enum modifiers is to be closer to perftool's modifiers (https://man7.org/linux/man-pages/man1/perf-list.1.html#EVENT_MODIFIERS). Partly for possible future extensions, partly to avoid inventing new concepts and nomenclature.
> 
> I guess it is about if you want to handle the complexity via configuration which requires future-proofing and handling details in the implementation or just pass it to perf_event_open. Both solutions would help me with my tasks as long as functionality is there.

Please check if the latest state of this PR is suitable, thanks.

src/trace_processor/importers/proto/perf_sample_tracker.cc
```diff
@@ -95,61 +95,54 @@ const char* StringifyCounter(int32_t counter) {
...
+    return "unknown";
+  };
 
-  std::vector<StringId> string_ids;
+  std::string name = base_counter_name().ToStdString();
 
-  for (auto it = perf_defaults.followers(); it; ++it) {
-    FollowerEvent::Decoder followers(*it);
-    string_ids.push_back(InternCounterName(perf_defaults, context, followers));
+  // Suffix with event modifiers, if any. Following the perftool convention.
```
auselen, line 127: I'm not against this, but it is also interesting to change/append user provided name.

auselen:
LGTM.

src/trace_processor/importers/proto/perf_sample_tracker.cc
rsavitski: Yeah, definitely considered both versions. Right now the trace parsing collapses identically named counters within a group (so the example from the patch description wasn't working well). 
So the suffixing is doing two things: (a) lets me easily "solve" the name uniqueness, (b) avoids requiring users from having to annotate their counter names whenever they add modifiers.
I can see upsides and downsides both ways, committing to this one until I see a strong reason not to.

src/profiling/perf/event_config.cc
```diff
@@ -243,36 +243,78 @@ int32_t ToClockId(protos::gen::PerfEvents::PerfClock pb_enum) {
...
+    } else {
+      return PerfCounter::BuiltinCounter(
+          name, protos::gen::PerfEvents::PerfEvents::SW_CPU_CLOCK,
+          PERF_TYPE_SOFTWARE, PERF_COUNT_SW_CPU_CLOCK);
+    }
+  };
+
+  std::optional<PerfCounter> counter = build_counter_without_modifiers();
+  if (!counter)
+    return counter;
```
KirillTim: Nit: maybe it is better to explicitly write `return std::nullopt;` instead?

rsavitski: Done.

PR #3021, opened 2025-09-19
Fix test name generation for the BytecodeInterpreterCastTest on Windows.

KirillTim:
On Windows, `to_string(NAN)` returns `"-(ind)"` string.
GTest test names can't contains `"("` or `")"`, so we replace them with `"_"`.

### DIFF
--- a/src/trace_processor/dataframe/impl/bytecode_interpreter_test_utils.h
+++ b/src/trace_processor/dataframe/impl/bytecode_interpreter_test_utils.h
@@ -75,8 +75,13 @@ struct Fetcher : ValueFetcher {
   uint32_t i = 0;
 };
 
-inline std::string FixNegativeAndDecimal(const std::string& str) {
-  return base::ReplaceAll(base::ReplaceAll(str, ".", "_"), "-", "neg_");
+inline std::string FixNegativeAndDecimalAndDouble(const std::string& str) {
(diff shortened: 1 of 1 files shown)


src/trace_processor/dataframe/impl/bytecode_interpreter_test_utils.h
```diff
@@ -75,8 +75,13 @@ struct Fetcher : ValueFetcher {
...
-inline std::string FixNegativeAndDecimal(const std::string& str) {
-  return base::ReplaceAll(base::ReplaceAll(str, ".", "_"), "-", "neg_");
+inline std::string FixNegativeAndDecimalAndDouble(const std::string& str) {
+  std::vector<std::string> replace_with_underscore = {".", "(", ")"};
+  std::string result = str;
+  for (const auto& to_replace : replace_with_underscore) {
+    result = base::ReplaceAll(result, to_replace, "_");
+  }
+  return base::ReplaceAll(result, "-", "neg_");
 }
```
primiano, line 85: at this point it's nicer to have a for loop like:

std::string result = str;
if (results.size() > 0 && result[0] == '-')
  result = "neg_" + result.substr(1)
  for (char& c : result) {
        if (c == '.' || c == '(' || c == ')' || c == '-') {
            c = '_';
        }

PR #3024, opened 2025-09-19
etw_parser: Improve performance of task state lookup

g-aattar:
The `etw_states_map` in `TaskStateToStringId` was being constructed on every function call, which is inefficient.

This patch replaces the `std::map` with a `static constexpr` array (`std::string_view[]`). This ensures the lookup table is initialized at compile time and provides a more efficient array-index lookup.

### DIFF
--- a/src/trace_processor/importers/etw/etw_parser.cc
+++ b/src/trace_processor/importers/etw/etw_parser.cc
@@ -242,7 +242,7 @@ StringId EtwParser::TaskStateToStringId(int64_t task_state_int) {
   const auto state = static_cast<uint8_t>(task_state_int);
 
   // Mapping for the different Etw states with their string description.
-  static const base::StringView etw_states[] = {
+  static constexpr std::string_view etw_states[] = {
       "Initialized",    // 0x00
       "Ready",          // 0x01
       "Running",        // 0x02


src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -241,19 +242,21 @@ void EtwParser::PushSchedSwitch(uint32_t cpu,
...
-      {0x03, "Stand By"},        // STANDBY
-      {0x04, "T"},               // TERMINATED
-      {0x05, "Waiting"},         // WAITING
-      {0x06, "Transition"},      // TRANSITION
-      {0x07, "Deferred Ready"},  // DEFERRED_READY
-  };
-
-  return etw_states_map.find(state) != etw_states_map.end()
-             ? context_->storage->InternString(etw_states_map[state])
+  static base::NoDestructor<std::map<uint8_t, base::StringView>> etw_states_map(
```
LalitMaganti: A map with <10 elements will be so much more efficient as an std::array + a linear scan. Please switch to that, we should avoid using NoDestructor as much as possible.

src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -242,7 +242,7 @@ StringId EtwParser::TaskStateToStringId(int64_t task_state_int) {
   const auto state = static_cast<uint8_t>(task_state_int);
 
   // Mapping for the different Etw states with their string description.
-  static const base::StringView kEtwStates[] = {
+  static const base::StringView etw_states[] = {
```
LalitMaganti: PLease use std:string_view. And make it constexpr.

src/trace_processor/importers/etw/etw_parser.cc
g-aattar: Done

src/trace_processor/importers/etw/etw_parser.cc
g-aattar: Done

PR #3026, opened 2025-09-19
On Linux, extract SoC model from the devicetree base compatible

willmcvicker:
In order to support Wattson on Linux platforms using a Device Tree (DT), we need to set the SystemInfo `android_soc_model` attribute. This can be extracted from the DT base compatible string and then mapped to the Android property "ro.soc.model".

Bug: 446195032
Tested: collect trace on Pixel 6 with Linux and verify Wattson track

### DIFF
--- a/src/base/android_utils.cc
+++ b/src/base/android_utils.cc
@@ -15,6 +15,7 @@
  */
 
 #include "perfetto/ext/base/android_utils.h"
+#include "perfetto/ext/base/file_utils.h"
 
 #include "perfetto/base/build_config.h"
 
@@ -67,6 +68,16 @@ std::string GetAndroidProp(const char* name) {
 
 #endif  // PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
 
+static std::string DeviceCompatibleToSoCModel(std::string& device_compatible) {
+  std::string ret;
+
+  // We only support Pixel 6 (Tensor) on non-Android platforms
+  if (device_compatible.find("google,gs101") != std::string::npos) {
+    ret = std::string("Tensor");
+  }
+  return ret;
+}
+
 Utsname GetUtsname() {
   Utsname utsname_info;
 #if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN) &&  \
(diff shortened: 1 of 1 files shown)


src/base/android_utils.cc
```diff
@@ -67,6 +68,16 @@ std::string GetAndroidProp(const char* name) {
 
 #endif  // PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
 
+static std::string DeviceCompatibleToSoCModel(std::string& device_compatible) {
```
LalitMaganti, line 71: Sorry but this is not an acceptable change to make to such a low level function. If you want something like this you need to plumb it through the Perfetto stack (new proto field, parse it in TP in a new metadata column) and handle it properly in Wattson code. 

Hacking this in like this leads to unmaintainable code in the long term.

src/base/android_utils.cc
```diff
@@ -96,6 +107,16 @@ SystemInfo GetSystemInfo() {
   info.utsname_info = GetUtsname();
   info.page_size = static_cast<uint32_t>(sysconf(_SC_PAGESIZE));
   info.num_cpus = static_cast<uint32_t>(sysconf(_SC_NPROCESSORS_CONF));
+
+  // Use the device compatible string to determine the SoC Model
```
LalitMaganti, line 111: This also needs to be prefixed with an if Linux block surely? Otherwise this is going to lead to problems on macos. 

Also, is this actually available on all Linux kennels? I would anticipate this will be gated on some kernel version? In which case we should not log so aggressively as it's expected that it won't be present.

willmcvicker: Ah, I wasn't considering MacOS here. I can put it in a PERFETTO_OS_LINUX  if-block. This path should be present on all linux devices using a devicetree. I can remove the ELOG so that it doesn't show up on x86 platforms that use ACPI though.

LalitMaganti: Yeah if you're not using device tree (how common is that on desktop/server Linux) then this log is irrelevant and should not bother you.

willmcvicker: Right, x86 systems will use ACPI (not device tree) while pretty much all other architectures like ARM64, RISCV, and MIPS use device tree.

src/base/android_utils.cc
willmcvicker: Thanks makes sense. Thanks!

PR #3031, opened 2025-09-21
trace_processor: add support for print_args(arg_set_id)

altimin:
Add a helper function to print the entire set of args at once as a JSON object,
reusing existing functionality for exporting arg sets to json.

This makes it easier to inspect arg sets at a glance and would enable
simplifying the UI code by removing the logic which constructs arg tree in
javascript.


src/trace_processor/perfetto_sql/intrinsics/functions/args.h
```diff
@@ -0,0 +1,45 @@
...
+  static constexpr char kName[] = "extract_arg";
+  static constexpr int kArgCount = 2;
+
+  using UserData = TraceStorage;
+  static void Step(sqlite3_context* ctx, int, sqlite3_value** argv);
+};
+
+// print_args(arg_set_id) prints the entire arg set as a json object.
+struct PrintArgs : public sqlite::Function<ExtractArg> {
+  static constexpr char kName[] = "print_args";
```
LalitMaganti: Please call this `__intrinsic_args_to_json`. I'm working on a way to expose C++ functions in SQL through the stdlib to allow for this to be self-documenting in https://perfetto.dev/docs/analysis/stdlib-docs

src/trace_processor/perfetto_sql/intrinsics/functions/args.cc
```diff
@@ -0,0 +1,206 @@
...
+#include "src/trace_processor/sqlite/bindings/sqlite_result.h"
+#include "src/trace_processor/sqlite/bindings/sqlite_type.h"
+#include "src/trace_processor/util/args_utils.h"
+
+namespace perfetto::trace_processor {
+namespace {
+// We care about the argument order, so we can't use jsoncpp here.
+void WriteAsJson(const ArgNode& node,
+                 const TraceStorage* storage,
+                 std::ostringstream& ss) {
```
LalitMaganti: Please never use string stream in stdlib. Use StringWriter or some other mechanism, stringstream is banned for binary size reasons.

src/trace_processor/util/args_utils.h
```diff
@@ -0,0 +1,87 @@
...
+#include <vector>
+
+#include "perfetto/base/status.h"
+#include "src/trace_processor/types/variadic.h"
+
+namespace perfetto::trace_processor {
+class TraceStorage;
+
+class ArgSet;
+
```
LalitMaganti, line 32: AddJsonValueToArgs should also be moved to this file since you're doing this.

src/trace_processor/util/args_utils.h
```diff
@@ -0,0 +1,87 @@
...
+class ArgNode {
```
LalitMaganti, line 33: Nothing about this is "arg" specific. Please factor this out into a proper json writer class similar to what I did was jsonparser to make this reusable in other contexts (of which I believe there are many).

LalitMaganti (commented):
Initial high level comments which will require rethinking bigger things: haven't dug into the implementation yet.

src/trace_processor/perfetto_sql/intrinsics/functions/args.cc
altimin: Is there a mechanism which dynamically allocates a string? An arg set can be quite large, and StringWriter only writes into a fixed-size strings, I believe.

LalitMaganti: I guess not. But in that case, prefer creating a mechanism to do this (maybe built on top of string/string_view/stackstring instead of using stringstream).

altimin: Ok, let me do this, then. I'll start with renaming `StringWriter` to `BufferedStringWriter` (as I was confused by the fact) and add `DynamicStringWriter`.

src/trace_processor/util/args_utils.h
altimin: The context here is that at some point I'm hoping to add support for `timestamp` type to args, so that SDK can write timestamps as arguments and the UI can interpret them, so I have been planning to keep this specific to args, so it can be extended in the future (the 90% of the complexity here is anyway `AppendArg`, which does all of the necessary parsing to convert a flat list of args into a tree).

src/trace_processor/perfetto_sql/intrinsics/functions/args.cc
LalitMaganti: I think it should be called "FixedStringWriter": the new class being called DynamicStringWriter sounds good.

src/trace_processor/util/args_utils.h
altimin: Keeping this as args-specific (as the most important method here is AppendArg anyway) and added a reusable json writer in #3135.

src/trace_processor/util/args_utils.h
altimin: I looked into it and I think it's better off in json_utils.h as it's very specific to parsing JSON (e.g. requires jsoncpp, etc).

src/trace_processor/perfetto_sql/intrinsics/functions/args.cc
```diff
@@ -0,0 +1,206 @@
...
+      std::nullopt,
+  }});
+  cursor.SetFilterValueUnchecked(0, arg_set_id);
+  ArgSet arg_set;
+  for (cursor.Execute(); !cursor.Eof(); cursor.Next()) {
+    const auto row_number = cursor.ToRowNumber();
+    const auto row = row_number.ToRowReference(arg_table);
+
+    const auto result =
+        arg_set.AppendArg(storage->GetString(row.key()).ToStdString(),
```
LalitMaganti: You probably don't want to use ToStdString here.

src/trace_processor/perfetto_sql/intrinsics/functions/args.cc
```diff
@@ -0,0 +1,206 @@
...
+                          storage->GetArgValue(row_number.row_number()));
```
LalitMaganti, line 191: Would be much more efficient if you followed the example of extract_arg above. I intentionally moved away from GetArgValue for it.

test/trace_processor/diff_tests/stdlib/prelude/args_functions_tests.py
```diff
@@ -0,0 +1,327 @@
...
+          extract_arg(arg_set_id, 'debug.key3') AS double_val,
+          extract_arg(arg_set_id, 'debug.missing_key') AS missing_val
+        FROM slice;
+        """,
+        out=Csv("""
+        "name","string_val","int_val","double_val","missing_val"
+        "Event","value1",42,3.250000,"[NULL]"
+        """))
+
+  def test___intrinsic_arg_set_to_json_simple(self):
```
LalitMaganti: Remove __intrinisic here and below: the fact it's an intrinisic doesn't matter for tests.

src/trace_processor/perfetto_sql/intrinsics/functions/args.h
```diff
@@ -0,0 +1,45 @@
...
+struct ExtractArg : public sqlite::Function<ExtractArg> {
+  static constexpr char kName[] = "extract_arg";
+  static constexpr int kArgCount = 2;
+
+  using UserData = TraceStorage;
+  static void Step(sqlite3_context* ctx, int, sqlite3_value** argv);
+};
+
+// Prints the entire arg set as a json object.
+struct PrintArgs : public sqlite::Function<ExtractArg> {
```
LalitMaganti: Rename to match function name

src/trace_processor/perfetto_sql/intrinsics/functions/args.cc
altimin: It actually doesn't matter, as StringSplitter needs a std::string anyway, but updated it to take a `string_view` in case  `StringSplitter` gets optimised at some point.

src/trace_processor/perfetto_sql/intrinsics/functions/args.cc
altimin: Chatted offline — I need to store `Variadic` in a tree, so I can't get away from having two switches.

LalitMaganti (approved):

PR #3032, opened 2025-09-21
ui: refactor display of args

altimin:
Leverage the new print_args() function to avoid reconstructing the args tree in JS.

This change doesn't affect UX, just simplifies the code.


altimin:
Checked with Lalit offline — he is happy with Steve's review here, so I'll land this.

PR #3036, opened 2025-09-21
Add DynamicStringWriter

altimin:
Add DynamicStringWriter, a counterpart to FixedStringWriter, which can be used in cases when the string can be arbitrary large.

### DIFF
--- a/include/perfetto/ext/base/dynamic_string_writer.h
+++ b/include/perfetto/ext/base/dynamic_string_writer.h
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2019 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef INCLUDE_PERFETTO_EXT_BASE_DYNAMIC_STRING_WRITER_H_
+#define INCLUDE_PERFETTO_EXT_BASE_DYNAMIC_STRING_WRITER_H_
+
+#include <string.h>
+
+#include "perfetto/base/logging.h"
+#include "perfetto/ext/base/string_utils.h"
+#include "perfetto/ext/base/string_view.h"
+
+namespace perfetto {
+namespace base {
+
+// A helper class which writes formatted data to a string buffer.
+// This is used in the trace processor where we write O(GBs) of strings and
+// sprintf is too slow.
+class DynamicStringWriter {
+ public:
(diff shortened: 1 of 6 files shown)


include/perfetto/ext/base/dynamic_string_writer.h
```diff
@@ -0,0 +1,152 @@
...
+  // number of digits of the integer is less than |padding|.
+  template <char padchar, uint64_t padding>
+  void AppendPaddedUnsignedInt(uint64_t value) {
+    AppendPaddedInt<padchar, padding>(value, false);
+  }
+
+  // Appends a hex integer to the buffer.
+  template <typename IntType>
+  void AppendHexInt(IntType value) {
+    // TODO(lalitm): trying to optimize this is premature given we almost never
```
LalitMaganti: ???

include/perfetto/ext/base/dynamic_string_writer.h
```diff
@@ -0,0 +1,152 @@
...
+    // print hex ints. Reevaluate this in the future if we do print them more.
+    constexpr size_t STACK_BUFFER_SIZE = 64;
+    StackString<STACK_BUFFER_SIZE> buf("%" PRIx64, value);
+    AppendString(buf.string_view());
+  }
+
+  // Appends a double to the buffer.
+  void AppendDouble(double value) {
+    // TODO(lalitm): trying to optimize this is premature given we almost never
```
LalitMaganti: As above

include/perfetto/ext/base/dynamic_string_writer.h
```diff
@@ -0,0 +1,152 @@
...
+    AppendLiteral("false");
+  }
+
+  StringView GetStringView() {
+    return StringView(buffer_.c_str(), buffer_.size());
+  }
+
+ private:
+  template <char padchar, uint64_t padding>
+  void AppendPaddedInt(uint64_t absolute_value, bool negate) {
```
LalitMaganti: Can we use base::StackString or printf and avoid duplicating the logic.

include/perfetto/ext/base/dynamic_string_writer.h
altimin: I copied this one from fixed_string_writer.h and have been meaning to ask you about it — the impl there is as follows, and I wonder if we need that todo there as well:

```
// Appends a hex integer to the buffer.
  template <typename IntType>
  void AppendHexInt(IntType value) {
    // TODO(lalitm): trying to optimize this is premature given we almost never
    // print hex ints. Reevaluate this in the future if we do print them more.
    size_t res =
        base::SprintfTrunc(buffer_ + pos_, size_ - pos_, "%" PRIx64, value);
    PERFETTO_DCHECK(pos_ + res <= size_);
    pos_ += res;
  }

  // Appends a double to the buffer.
  void AppendDouble(double value) {
    // TODO(lalitm): trying to optimize this is premature given we almost never
    // print doubles. Reevaluate this in the future if we do print them more.
    size_t res = base::SprintfTrunc(buffer_ + pos_, size_ - pos_, "%lf", value);
    PERFETTO_DCHECK(pos_ + res <= size_);
    pos_ += res;
  }
```

LalitMaganti: I have no memory of this. In any case, I think we need to rationalize this: we should not just spread todos to more and more of the codebase nor make blind copies like this.

LalitMaganti (approved):

PR #3042, opened 2025-09-22
tp: migrate createdfunction from legacy api to sqlite function api

LalitMaganti:
Remove the LegacySqlFunction abstraction completely: we inline it into
the last remaining one.

In the future, we should really simplify this code (remove memoization,
reduce the layers of indirection) but for now, just do the minimum
amount of work necessary to get rid for LegacySqlFunction


src/trace_processor/perfetto_sql/engine/created_function.cc
```diff
@@ -135,15 +172,19 @@ struct StoredSqlValue {
 
   SqlValue AsSqlValue() {
     if (std::holds_alternative<std::nullptr_t>(data)) {
```
sashwinbalaji: else if as before was better

LalitMaganti: Moved to switch + variant_index/unchecked_get instead.

PR #3044, opened 2025-09-22
ui: Fix disappearing slices by removing an optimization that assumes sorted slices in BaseSliceTrack

stevegolton:
Currently we make the assumption in BaseSliceTrack that slices are ordered by `ts`, which allows us to perform a very specific optimization which avoids rendering anything when the first slice is off the right hand side of the viewport.

However, slices from the mipmap function are ordered by depth first then ts, meaning that slices are not necessarily sorted by ts. In practice, slices at lower depths usually do start before those at higher depths anyway, so this is usually not an issue.

This patch removes the optimization entirely as it seems pretty dubious whether it actually helps all that much.

Fixes: https://buganizer.corp.google.com/issues/446503267

### DIFF
--- a/ui/src/components/tracks/base_slice_track.ts
+++ b/ui/src/components/tracks/base_slice_track.ts
@@ -21,7 +21,6 @@ import {assertExists} from '../../base/logging';
 import {clamp, floatEqual} from '../../base/math_utils';
 import {cropText} from '../../base/string_utils';
 import {Time, time} from '../../base/time';
-import {exists} from '../../base/utils';
 import {uuidv4Sql} from '../../base/uuid';
 import {featureFlags} from '../../core/feature_flags';
 import {raf} from '../../core/raf_scheduler';
@@ -109,15 +108,6 @@ function filterVisibleSlices<S extends Slice>(
   // The last visible slice is simpler, since the slices are sorted
   // by timestamp you can binary search for the last slice such
   // that slice.start <= end.
-
-  // One specific edge case that will come up often is when:
-  // For all slice in slices: slice.startNsQ > end (e.g. all slices are
-  // to the right).
-  // Since the slices are sorted by startS we can check this easily:
-  const maybeFirstSlice: S | undefined = slices[0];
-  if (exists(maybeFirstSlice) && maybeFirstSlice.startNs > end) {
-    return [];
-  }
 
   return slices.filter((slice) => slice.startNs <= end && slice.endNs >= start);
 }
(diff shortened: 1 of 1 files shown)


ui/src/components/tracks/base_slice_track.ts
```diff
@@ -733,6 +733,7 @@ export abstract class BaseSliceTrack<
         ${resolution}
       ) z
       CROSS JOIN (${this.getSqlSource()}) s using (id)
+      ORDER BY tsQ
```
LalitMaganti: For performance reasons, this is a very bad idea and will tank query performance. We should really figure out why the underlying table is returning non-sorted data: the API contract is that it's always sorted.

stevegolton: You're saying the virtual table based on __intrinsic_slice_mipmap should always return sorted results?

LalitMaganti: Yes, there's actually very little chance of implementing anything which is not sorted. The only potential snag is how depth is dealt with. I think it might be sorted on depth then name. I'd have to check the code. But in general we should make it return data in the way you need.

rsavitski (changes requested):
-1 as well, let's investigate the trace (it was a local tracebox on my laptop, and I was messing with perf counters but I don't see how that would affect the scheduling tracks unless it went full on data corruption).

ui/src/components/tracks/base_slice_track.ts
stevegolton: It's sorted by depth then ts. This is a problem for an optimization performed later when culling slices. We could just remove this optimization though, I'm not convinced about how useful it is really.

stevegolton:
> -1 as well, let's investigate the trace (it was a local tracebox on my laptop, and I was messing with perf counters but I don't see how that would affect the scheduling tracks unless it went full on data corruption).

This is definitely a bug in the base slice track renderer.  I'm not quite sure why we don't see the problem more often, but apparently your trace is just unusual enough to provoke it.

LalitMaganti (approved):

rsavitski:
> > -1 as well, let's investigate the trace (it was a local tracebox on my laptop, and I was messing with perf counters but I don't see how that would affect the scheduling tracks unless it went full on data corruption).
> 
> This is definitely a bug in the base slice track renderer. I'm not quite sure why we don't see the problem more often, but apparently your trace is just unusual enough to provoke it.

Right, I originally thought the data is out of order at the table level, but it's a UI issue. LGTM.

PR #3051, opened 2025-09-22
ui: make QueryResult classes easier to be used by Winscope

primiano:
Minor non-functional refactoring: nothing should change
in our codebase, but makes our code easier to be reused
by Winscope.
The goal for Winscope is the following:
- Be able to call directly queryStreaming() passing an object
  that they own which will take the query result raw bytes.
- This object that they own will just forward the raw bytes onto
  a worker they own, where they can decode it.

The rough sketch of their code is something like:
engine.streamingQuery(winscopeWritableRes, sqlQuery)

Where WinscopeWritableRes looks like this:
```

engine.streamingQuery(new ForwardingWritableQueryResult(worker.messagePort), 'SELECT * FROM blah...')

export class ForwardingWritableQueryResult implements WritableQueryResult {
  private _isLastBatch = false;

  constructor(private port: MessagePort) {}

  appendResultBatch(resBytes: Uint8Array): void {
    // 1. Send the raw bytes to the worker that will do the decoding
    // TODO here you'll have to do something a bit more sophisticated to tell
    // the message type. Depends on how your worker works.
    this.port.postMessage(resBytes);

    // 2. We need to do enough decoding just to extract the isLastBatch
    const reader = protobuf.Reader.create(resBytes);
    assertTrue(reader.pos === 0);
    while (reader.pos < reader.len) {
      const tag = reader.uint32();
      switch (tag >>> 3) {
        case 3: // batch
          const batchLen = reader.uint32();
          const batchRaw = resBytes.subarray(reader.pos, reader.pos + batchLen);
          reader.pos += batchLen;
          this._isLastBatch = this.extractIsLastBatch(batchRaw);
          break;

        default:
          reader.skipType(tag & 7);
          break;
      } // switch (tag)
    } // while (pos < end)
  }

  isComplete(): boolean {
    return this._isLastBatch;
  }

  private extractIsLastBatch(batchBytes: Uint8Array) {
    const reader = protobuf.Reader.create(batchBytes);
    assertTrue(reader.pos === 0);
    const end = reader.len;
    let result = false;
    while (reader.pos < end) {
      const tag = reader.uint32();
      switch (tag >>> 3) {
        case 6: // is_last_batch (boolean).
          result = !!reader.bool();
          break;

        default:
          reader.skipType(tag & 7);
          break;
      } // switch(tag)
    } // while (pos < end)
    return result;
  }
}

```


primiano:
@stevegolton WDYT?
I am also tempted to keep streamingQuery private and tell winscope to mark as public in their repo.

PR #3053, opened 2025-09-22
ui: add flag option to enable full sorting mode for trace processor

zytyz:
This adds an option in the UI flags tab to toggle full sort mode for the trace processor. By default, the full sort mode is off.


LalitMaganti:
I'm fine to land this but in general, it's a better idea to properly write traces in the first place rather than work around it with this option :)

ui/src/core/load_trace.ts
```diff
@@ -89,6 +89,13 @@ const FTRACE_DROP_UNTIL_FLAG = featureFlags.register({
     'Drop ftrace events until all per-cpu data streams are known to be valid',
   defaultValue: true,
 });
+const FORCE_FULL_SORT_FLAG = featureFlags.register({
+  id: 'forceFullSort',
+  name: 'Force full sort',
+  description:
+    'Forces the trace processor into performing a full sort ignoring any windowing logic',
+  defaultValue: false,
+});
```
stevegolton, line 98: Just one thing to note, we will move these flags to 'settings' in the near future so you may need to reset them once that happens.

LalitMaganti:
See https://github.com/google/perfetto/pull/3053#discussion_r2371907356. Resolved to allow merging

ui/src/core/load_trace.ts
zytyz: Will do. Thanks for noting that!

Issue #3055, opened 2025-09-22
Perfetto on main no longer shows GPU Counters from Samsung Exynos devices


dfriederich:
When capturing a perfetto trace with GPU counters on a Samsung Exynos, then the GPU counter data including the descriptors is captured.
But with the current main perfetto, I no longer see any data in the UI.
The issue seems to be how the Samsung Exynos GPU driver encodes the clock in the GPU counter descriptor entry, and how perfetto no maps this internally to a negative timestamp, and then drops it.

[gapid.zip](https://github.com/user-attachments/files/22478644/gapid.zip)

The issue seems to be how Samsung Exynos encodes the  gpu_counter_event, more specific how it adds the counter descriptors with a timestamp 0 using a local clock id.
from the trace_convert text output of the attached perfetto file:
```
packet {
  clock_snapshot {
    clocks {
      clock_id: 6
      timestamp: 6480711894224
    }
    clocks {
      clock_id: 81
      timestamp: 6485040353007
    }
  }
  trusted_uid: 2000
  trusted_packet_sequence_id: 2
  trusted_pid: 6132
  previous_packet_dropped: true
}
packet {
  timestamp_clock_id: 81
  timestamp: 0
  gpu_counter_event {
    counter_descriptor {
      specs {
        counter_id: 100
        name: "GPU Busy %"
        description: "Total percentage of time that the GPU was busy"
        numerator_units: PERCENT
        denominator_units: NONE
        select_by_default: true
      }
      specs {
        counter_id: 101
        name: "GPU Busy Cycles"
        description: "Number of cycles that GPU command processor was busy"
        numerator_units: INSTRUCTION
        denominator_units: NONE
        select_by_default: true
      }
    }
  }
  trusted_uid: 2000
  trusted_packet_sequence_id: 2
  trusted_pid: 6132
}
packet {
  timestamp_clock_id: 81
  timestamp: 6485009954335
  gpu_counter_event {
    counters {
      counter_id: 100
      int_value: 0
    }
    counters {
      counter_id: 101
      int_value: 0
    }
  }
  trusted_uid: 2000
  trusted_packet_sequence_id: 2
  trusted_pid: 6132
}
```

note that the event with the counter descriptor has a timestamp of 0 using a clock id of 81.
This maps to a clock 6 timestamp of -4 seconds (-4'328'458'783= 6480711894224-6485040353007)

[...]

PR #3057, opened 2025-09-22
tp: fix parsing of gpu counter descriptors

LalitMaganti:
Turns out some GPU vendors set timestamps on descriptor packets which are broken and are ending up less than zero (because of different clock + clock sync).

Fix this by handling gpu descriptors at tokenization time instead. This ensures they are always handled first.

Fixes: https://github.com/google/perfetto/issues/3055


sashwinbalaji (approved):
do we always want to support this or ask vendors to update ? If planning to support please add a comment explaining the reason else a todo to remove it.

LalitMaganti:
> do we always want to support this or ask vendors to update ? If planning to support please add a comment explaining the reason else a todo to remove it.

This was always how this code should have been written anyway: we always handle packets which don't have (or shouldn't have) timestamps at tokenization time and the remainder at parse time.

PR #3058, opened 2025-09-22
Plumbing for descriptors from Json Amalgamator through UI

hteplitz-Google:
This plumbs the extra parsing descriptors from Json Amalgamator through the Perfetto UI. 
This is a DRAFT PR and will not work until the Trace Processor has plumbing for the descriptors in another PR.


ui/src/core/load_trace.ts
```diff
@@ -540,3 +547,13 @@ async function getTracingMetadataTimeBounds(engine: Engine): Promise<TimeSpan> {
 
   return new TimeSpan(startBound, endBound);
 }
+
+function base64Decode(str: string): Uint8Array {
```
LalitMaganti: Pretty sure this function should already be in base

ui/src/core/app_impl.ts
```diff
@@ -95,6 +95,8 @@ export class AppContext {
   // via is_internal_user.js
   extraSqlPackages: SqlPackage[] = [];
 
+  extraParsingDescriptors: string[] = [];
```
LalitMaganti, line 100: Add a comment here similar to other ones.

ui/src/core/app_impl.ts
```diff
@@ -360,6 +362,7 @@ export class AppImpl implements App {
     // complete trace loading (we don't bother supporting cancellations. If the
     // user is too bothered, they can reload the tab).
     this.appCtx.openTraceAsyncLimiter.schedule(async () => {
+      await this.extraLoadingPromise;
```
LalitMaganti, line 370: Add a comment about why this is needed.

src/base/lock_free_task_runner.cc
stevegolton, line 28: Unrelated change?

ui/src/core/app_impl.ts
```diff
@@ -360,6 +364,10 @@ export class AppImpl implements App {
     // complete trace loading (we don't bother supporting cancellations. If the
     // user is too bothered, they can reload the tab).
     this.appCtx.openTraceAsyncLimiter.schedule(async () => {
+      // Wait for extras parsing descriptors to be loaded
+      // via is_internal_user.js. This prevents a race condition where
+      // trace loading would otherwise begin before this data is available.
+      await this.extraLoadingPromise;
```
stevegolton, line 370: @LalitMaganti 

This is a significant change in where we await the promise. Currently we only await it at traceReady time. 

I mean we already currently have a race when loading the extraSqlModules (that is to say we don't await this promise before loading the extraSqlModules), but at least that happens after the trace has been ingested, which gives us a bit more time.

Here, we have to await before we even start loading the trace, which could have a large impact on trace load times.

Are we OK with this?

ui/src/frontend/is_internal_user_script_loader.ts
```diff
@@ -41,6 +41,8 @@ interface Globals {
   // internal_user script.
   readonly extraSqlPackages: SqlPackage[];
 
+  extraParsingDescriptors: string[];
```
stevegolton: This should be a `ReadonlyArray<string>`. Same in the various getters and setters

ui/src/trace_processor/engine.ts
```diff
@@ -396,12 +396,12 @@ export abstract class EngineBase implements Engine, Disposable {
     args.ingestFtraceInRawTable = ingestFtraceInRawTable;
     args.analyzeTraceProtoContent = analyzeTraceProtoContent;
     args.ftraceDropUntilAllCpusValid = ftraceDropUntilAllCpusValid;
-    args.sortingMode = forceFullSort
-      ? protos.ResetTraceProcessorArgs.SortingMode.FORCE_FULL_SORT
-      : protos.ResetTraceProcessorArgs.SortingMode.DEFAULT_HEURISTICS;
     args.parsingMode = tokenizeOnly
       ? protos.ResetTraceProcessorArgs.ParsingMode.TOKENIZE_ONLY
       : protos.ResetTraceProcessorArgs.ParsingMode.DEFAULT;
+    if (extraParsingDescriptors) {
```
stevegolton: What's the desired logic here? Should `args.extraParsingDescriptors` only be set if `extraParsingDescriptors` is defined, or can we just do the following without the `if`?

```ts
args.extraParsingDescriptors = extraParsingDescriptors;
```

ui/src/trace_processor/engine.ts
```diff
@@ -44,7 +44,7 @@ export interface TraceProcessorConfig {
   ingestFtraceInRawTable: boolean;
   analyzeTraceProtoContent: boolean;
   ftraceDropUntilAllCpusValid: boolean;
-  forceFullSort: boolean;
+  extraParsingDescriptors?: Uint8Array[];
```
stevegolton: This should be a `ReadonlyArray<Uint8Array>`

ui/src/core/app_impl.ts
LalitMaganti: Unfortunately, there's not much we can do. We need this to parse the trace, we cannot really do it retroactively. It sucks but it is what it is.

Best thing we can do is figure out how we can make this fast for everyone. @primiano for ideas on this as well.

stevegolton: Fair enough. If folks complain we might have to do something about it but let's see.

stevegolton (approved):
The UI changes look good! I feel like more C++ changes have crept in though?

ui/src/core/app_impl.ts
```diff
@@ -409,6 +417,14 @@ export class AppImpl implements App {
     return this.appCtx.extraSqlPackages;
   }
 
+  get extraParsingDescriptors(): ReadonlyArray<string> {
+    return this.appCtx.extraParsingDescriptors;
+  }
+
+  set extraParsingDescriptors(value: ReadonlyArray<string>) {
```
LalitMaganti: Because we append, I don't think we need this.

stevegolton (approved):

LalitMaganti (approved):

PR #3070, opened 2025-09-24
Add collect clocks primitive to trace redactor

edgararriagag:
Collect clocks primitive is used to collect clock information from the trace. This is intended to be used in future commits to perform redactor Timeline conversions between different clocks in redacted traces.

### DIFF
--- a/src/trace_redaction/BUILD.gn
+++ b/src/trace_redaction/BUILD.gn
@@ -32,6 +32,8 @@ source_set("trace_redaction") {
     "add_synth_threads_to_process_trees.h",
     "broadphase_packet_filter.cc",
     "broadphase_packet_filter.h",
+    "collect_clocks.cc",
+    "collect_clocks.h",
     "collect_frame_cookies.cc",
     "collect_frame_cookies.h",
     "collect_system_info.cc",
@@ -63,6 +65,8 @@ source_set("trace_redaction") {
     "redact_process_events.h",
     "redact_sched_events.cc",
     "redact_sched_events.h",
+    "redactor_clock_converter.cc",
+    "redactor_clock_converter.h",
     "reduce_threads_in_process_trees.cc",
     "reduce_threads_in_process_trees.h",
     "scrub_process_stats.cc",
@@ -125,6 +129,7 @@ perfetto_unittest_source_set("unittests") {
   testonly = true
   sources = [
     "broadphase_packet_filter_unittest.cc",
+    "collect_clocks_unittest.cc",
     "collect_frame_cookies_unittest.cc",
     "collect_system_info_unittest.cc",
     "collect_timeline_events_unittest.cc",
--- a/src/trace_redaction/collect_clocks.cc
+++ b/src/trace_redaction/collect_clocks.cc
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/trace_redaction/collect_clocks.h"
+
+#include "perfetto/protozero/field.h"
+
+#include "perfetto/base/status.h"
+#include "perfetto/ext/base/status_macros.h"
+#include "src/trace_redaction/trace_redaction_framework.h"
+
+#include "protos/perfetto/trace/clock_snapshot.pbzero.h"
+#include "protos/perfetto/trace/trace_packet_defaults.pbzero.h"
+
+using namespace perfetto::trace_processor;
+
+namespace perfetto::trace_redaction {
+
+base::Status CollectClocks::Collect(
+    const protos::pbzero::TracePacket::Decoder& packet,
+    Context* context) const {
+  context->clock_snapshot.clear();
+  if (packet.has_clock_snapshot()) {
+    protos::pbzero::ClockSnapshot_Decoder snapshot_decoder(
+        packet.clock_snapshot());
+
+    if (snapshot_decoder.has_primary_trace_clock()) {
+      int32_t trace_clock = snapshot_decoder.primary_trace_clock();
+      context->clock_converter.SetPrimaryTraceClock(
+          static_cast<int64_t>(trace_clock));
+    }
+    for (auto clock_it = snapshot_decoder.clocks(); clock_it; clock_it++) {
+      RedactorClockSynchronizer::ClockTimestamp clock_ts(0, 0);
+
+      protos::pbzero::ClockSnapshot_Clock_Decoder clock_decoder(
+          clock_it->as_bytes());
+
+      if (clock_decoder.has_clock_id()) {
+        RedactorClockSynchronizer::Clock clock =
+            RedactorClockSynchronizer::Clock(
+                static_cast<int64_t>(clock_decoder.clock_id()));
+        clock_ts.clock = clock;
+      }
+
+      if (clock_decoder.has_timestamp()) {
+        clock_ts.timestamp = static_cast<int64_t>(clock_decoder.timestamp());
+      }
+      context->clock_snapshot.push_back(clock_ts);
+    }
+
+    RETURN_IF_ERROR(
+        context->clock_converter.AddClockSnapshot(context->clock_snapshot));
+  } else if (packet.has_trace_packet_defaults()) {
+    RETURN_IF_ERROR(
+        OnTracePacketDefaults(packet.trace_packet_defaults(), context));
+  }
+
+  return base::OkStatus();
+}
+
+base::Status CollectClocks::OnTracePacketDefaults(
+    protozero::ConstBytes trace_packet_defaults_bytes,
+    Context* context) const {
+  protos::pbzero::TracePacketDefaults_Decoder trace_packet_defaults_decoder(
+      trace_packet_defaults_bytes);
+  if (trace_packet_defaults_decoder.has_timestamp_clock_id()) {
+    uint32_t perf_clock_id = trace_packet_defaults_decoder.timestamp_clock_id();
+    context->clock_converter.SetPerfTraceClock(perf_clock_id);
+    return base::OkStatus();
+  }
+
+  return base::OkStatus();
+}
+
+}  // namespace perfetto::trace_redaction
--- a/src/trace_redaction/redactor_clock_converter.cc
+++ b/src/trace_redaction/redactor_clock_converter.cc
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/trace_redaction/redactor_clock_converter.h"
+
+namespace perfetto::trace_redaction {
+
+using ClockId = RedactorClockSynchronizer::ClockId;
+
+RedactorClockSynchronizerListenerImpl::RedactorClockSynchronizerListenerImpl()
+    : trace_time_updates_(0) {}
+
+base::Status RedactorClockSynchronizerListenerImpl::OnClockSyncCacheMiss() {
+  return base::OkStatus();
+}
+
+base::Status RedactorClockSynchronizerListenerImpl::OnInvalidClockSnapshot() {
+  PERFETTO_ELOG("Invalid clocks snapshot found during redaction");
+  return base::ErrStatus("Invalid clocks snapshot found during redaction");
+}
+
+base::Status RedactorClockSynchronizerListenerImpl::OnTraceTimeClockIdChanged(
+    ClockId trace_time_clock_id [[maybe_unused]]) {
+  ++trace_time_updates_;
+  if (PERFETTO_UNLIKELY(trace_time_updates_ > 1)) {
+    // We expect the trace time to remain constant for a trace.
+    PERFETTO_ELOG(
+        "Redactor clock conversion trace time unexpectedly changed %d times",
+        trace_time_updates_);
+    return base::ErrStatus(
+        "Redactor clock conversion trace time unexpectedly changed %d times",
+        trace_time_updates_);
+  }
+  return base::OkStatus();
+}
+
+base::Status RedactorClockSynchronizerListenerImpl::OnSetTraceTimeClock(
+    ClockId trace_time_clock_id [[maybe_unused]]) {
+  return base::OkStatus();
+}
+
+bool RedactorClockSynchronizerListenerImpl::IsLocalHost() {
+  // Redactor does not support multi-machine clock conversion
+  return true;
+}
+
+ClockId RedactorClockConverter::GetPrimaryTraceClock() {
+  return primary_trace_clock;
+}
+
+base::Status RedactorClockConverter::SetPrimaryTraceClock(ClockId clock_id) {
+  primary_trace_clock = clock_id;
+  RETURN_IF_ERROR(clock_synchronizer.get()->SetTraceTimeClock(clock_id));
+  return base::OkStatus();
+}
+
+void RedactorClockConverter::SetPerfTraceClock(ClockId clock_id) {
+  perf_clock = clock_id;
+}
+
+ClockId RedactorClockConverter::GetPerfTraceClock() {
+  return perf_clock;
+}
+
+base::Status RedactorClockConverter::AddClockSnapshot(
+    std::vector<RedactorClockSynchronizer::ClockTimestamp>& clock_snapshot) {
+  base::StatusOr<uint32_t> snapshot_id =
+      clock_synchronizer.get()->AddSnapshot(clock_snapshot);
+  RETURN_IF_ERROR(snapshot_id.status());
+  return base::OkStatus();
+}
+
+base::Status RedactorClockConverter::ConvertPerfToTrace(
+    uint64_t perf_ts,
+    uint64_t* out_ts) const {
+  ASSIGN_OR_RETURN(int64_t trace_ts, clock_synchronizer.get()->ToTraceTime(
+                                         static_cast<int64_t>(perf_clock),
+                                         static_cast<int64_t>(perf_ts)));
+  *out_ts = static_cast<uint64_t>(trace_ts);
+  return base::OkStatus();
+}
+
+}  // namespace perfetto::trace_redaction
--- a/src/trace_redaction/redactor_clock_converter.h
+++ b/src/trace_redaction/redactor_clock_converter.h
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_TRACE_REDACTION_REDACTOR_CLOCK_CONVERTER_H_
+#define SRC_TRACE_REDACTION_REDACTOR_CLOCK_CONVERTER_H_
+
+#include "perfetto/ext/base/status_macros.h"
+#include "protos/perfetto/common/builtin_clock.pbzero.h"
+#include "src/trace_processor/util/clock_synchronizer.h"
+
+namespace perfetto::trace_redaction {
+
+class RedactorClockSynchronizerListenerImpl {
+ private:
+  // Number of time that trace time has been updated.
+  uint32_t trace_time_updates_;
+
+ public:
+  RedactorClockSynchronizerListenerImpl();
+
+  base::Status OnClockSyncCacheMiss();
+
+  base::Status OnInvalidClockSnapshot();
+
+  base::Status OnTraceTimeClockIdChanged(
+      perfetto::trace_processor::ClockSynchronizer<
+          RedactorClockSynchronizerListenerImpl>::ClockId trace_time_clock_id);
+
+  base::Status OnSetTraceTimeClock(
+      perfetto::trace_processor::ClockSynchronizer<
+          RedactorClockSynchronizerListenerImpl>::ClockId trace_time_clock_id);
+
+  // Always returns true as redactor only supports local host clock conversion.
+  bool IsLocalHost();
+};
+
+using RedactorClockSynchronizer = perfetto::trace_processor::ClockSynchronizer<
+    RedactorClockSynchronizerListenerImpl>;
+
+/**
+ * This class handles conversions between different clocks for trace redactor.
+ *
+ * This class is a wrapper for trace_processor::ClockSynchronizer with the
+ * addition that it caches clocks required for conversion for different data
+ * sources and it is designed to be used by the trace redactor.
+ *
+ * Any trace packet intends to use the redactor ProcessThreadTimeline and whose
+ * clock won't be the default trace time should use this class to convert
+ * it to the default trace time which is used by ProcessThreadTimeline.
+ */
+class RedactorClockConverter {
+ private:
+  std::unique_ptr<RedactorClockSynchronizer> clock_synchronizer;
+  RedactorClockSynchronizer::ClockId primary_trace_clock;
+  RedactorClockSynchronizer::ClockId perf_clock;
+
+ public:
+  RedactorClockConverter() {
+    // Set the default clocks for the trace
+    primary_trace_clock = protos::pbzero::BuiltinClock::BUILTIN_CLOCK_BOOTTIME;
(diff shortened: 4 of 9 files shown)


src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,99 @@
...
+#define SRC_TRACE_REDACTION_REDACTOR_CLOCK_CONVERTER_H_
+
+#include "perfetto/ext/base/status_macros.h"
+#include "protos/perfetto/common/builtin_clock.pbzero.h"
+#include "src/trace_processor/util/clock_synchronizer.h"
+
+namespace perfetto::trace_redaction {
+
+class RedactorClockSynchronizerListenerImpl {
+ private:
```
LalitMaganti: Please follow the coeestyle elsewhere: public first then private.

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,99 @@
...
+ * it to the default trace time which is used by ProcessThreadTimeline.
+ */
+class RedactorClockConverter {
+ private:
+  std::unique_ptr<RedactorClockSynchronizer> clock_synchronizer;
+  RedactorClockSynchronizer::ClockId primary_trace_clock;
+  RedactorClockSynchronizer::ClockId perf_clock;
+
+ public:
+  RedactorClockConverter() {
```
LalitMaganti: This can go in the cc file

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,99 @@
...
+          RedactorClockSynchronizerListenerImpl>::ClockId trace_time_clock_id);
+
+  // Always returns true as redactor only supports local host clock conversion.
+  bool IsLocalHost();
+};
+
+using RedactorClockSynchronizer = perfetto::trace_processor::ClockSynchronizer<
+    RedactorClockSynchronizerListenerImpl>;
+
+/**
```
LalitMaganti: Use // for comments not /**/

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,88 @@
...
+
+  return base::OkStatus();
+}
+
+base::Status CollectClocks::OnTracePacketDefaults(
+    protozero::ConstBytes trace_packet_defaults_bytes,
+    Context* context) const {
+  protos::pbzero::TracePacketDefaults_Decoder trace_packet_defaults_decoder(
+      trace_packet_defaults_bytes);
+  if (trace_packet_defaults_decoder.has_timestamp_clock_id()) {
```
LalitMaganti: There's no guarantee that default clocks correspond to the perf clock. Just because that's how it's done today, doesn't mean it's going to be the case in the future

To make this more generic and resilient, the correct thing to do is to keep track of the sequence -> clock mapping. You can then look this up for the perf clock.

LalitMaganti (changes requested):
Initial things I spotted. There needs to be some changes to how the clocks are being done: there's an assumption that every packet where a default is specified is a perf clock: this is untrue and will break in the future when we add more databsources. We should design this code to be resilient.

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done

src/trace_redaction/collect_clocks.cc
edgararriagag: Done, ptal and let me know if the new implementation fully covers your request.

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,99 @@
...
+
+    if (snapshot_decoder.has_primary_trace_clock()) {
+      int32_t trace_clock = snapshot_decoder.primary_trace_clock();
+      context->clock_converter.SetPrimaryTraceClock(
+          static_cast<int64_t>(trace_clock));
+    }
+    for (auto clock_it = snapshot_decoder.clocks(); clock_it; clock_it++) {
+      ClockTimestamp clock_ts(0, 0);
+
+      protos::pbzero::ClockSnapshot_Clock_Decoder clock_decoder(
```
LalitMaganti: Surely if both of these are not specified, we should treat it as an error? Would also be nicer if this was factored out into a function in that case to avoid ugly initialization and setting fields.

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,128 @@
...
+// This class is a wrapper for trace_processor::ClockSynchronizer with the
+// addition that it caches clocks required for conversion for different data
+// sources and it is designed to be used by the trace redactor.
+//
+// Any trace packet intends to use the redactor ProcessThreadTimeline and whose
+// clock won't be the default trace time should use this class to convert
+// it to the default trace time which is used by ProcessThreadTimeline.
+class RedactorClockConverter {
+ public:
+  enum ClockType { PERF_CLOCK, UNKNOWN };
```
LalitMaganti: * Use kUnknown etc style for enum valeus
* enum class rather than enum
* The usage of "perf clock" is confusing because this is *not* actually the perf clock in the Linux sense of the word but rather the "clock you are considering as the default for the perf data source"

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,128 @@
...
+  static constexpr int64_t INVALID_CLOCK_ID = -1;
```
LalitMaganti: Anywhere we need this, we can use std::optional as a much better and more type safe way to do the same thing.

edgararriagag: Done

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done all the points. I also renamed to DefaultClockType hopefully this is less ambiguous

src/trace_redaction/collect_clocks.cc
edgararriagag: Done

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,107 @@
...
+#include "protos/perfetto/trace/trace_packet_defaults.pbzero.h"
+
+using namespace perfetto::trace_processor;
+
+namespace perfetto::trace_redaction {
+
+base::Status CollectClocks::Collect(
+    const protos::pbzero::TracePacket::Decoder& packet,
+    Context* context) const {
+  context->clock_snapshot.clear();
```
LalitMaganti: Why is this field in the context? In fact, why is it not just a local? I don't see it being used anywhere else?

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,107 @@
...
+  if (packet.has_clock_snapshot()) {
+    protos::pbzero::ClockSnapshot_Decoder snapshot_decoder(
+        packet.clock_snapshot());
+
+    if (snapshot_decoder.has_primary_trace_clock()) {
+      int32_t trace_clock = snapshot_decoder.primary_trace_clock();
+      context->clock_converter.SetPrimaryTraceClock(
```
LalitMaganti: If this returns an error, shouldn't we fail redaction?

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,107 @@
...
+          static_cast<int64_t>(trace_clock));
+    }
+    for (auto clock_it = snapshot_decoder.clocks(); clock_it; clock_it++) {
+      ClockTimestamp clock_ts(0, 0);
+      RETURN_IF_ERROR(ParseClock(clock_it->as_bytes(), clock_ts));
```
LalitMaganti: I don't think solves my problem? ParseClock should return StatusOr<ClockTimestamp> and specifically, you shouldn't be setting fields (or initializing to 0,0) but directly returning the initialized values.

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,107 @@
...
+      context->clock_snapshot.push_back(clock_ts);
+    }
+
+    RETURN_IF_ERROR(
+        context->clock_converter.AddClockSnapshot(context->clock_snapshot));
+  } else if (packet.has_trace_packet_defaults()) {
+    if (!packet.has_trusted_packet_sequence_id()) {
+      PERFETTO_ELOG("Could not find sequence id for TracePacketDefaults");
```
LalitMaganti: Why ELOG what you're already returning as the error mesage below? That will be ELOGed anyway?

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,107 @@
...
+
+  return base::OkStatus();
+}
+
+base::Status CollectClocks::OnTracePacketDefaults(
+    protozero::ConstBytes trace_packet_defaults_bytes,
+    uint32_t trusted_sequence_id,
+    Context* context) const {
+  protos::pbzero::TracePacketDefaults_Decoder trace_packet_defaults_decoder(
+      trace_packet_defaults_bytes);
```
LalitMaganti: You can shorten this.

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  return base::OkStatus();
+}
+
+base::Status CollectClocks::OnTracePacketDefaults(
+    protozero::ConstBytes trace_packet_defaults_bytes,
+    uint32_t trusted_sequence_id,
+    Context* context) const {
+  protos::pbzero::TracePacketDefaults_Decoder trace_packet_defaults_decoder(
```
LalitMaganti: TracePacketDefaults::Decoder

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,126 @@
...
+// This class is a wrapper for trace_processor::ClockSynchronizer with the
+// addition that it caches clocks required for conversion for different data
+// sources and it is designed to be used by the trace redactor.
+//
+// Any trace packet intends to use the redactor ProcessThreadTimeline and whose
+// clock won't be the default trace time should use this class to convert
+// it to the default trace time which is used by ProcessThreadTimeline.
+class RedactorClockConverter {
+ public:
+  enum class DefaultClockType { kPerfClock, kUnknown };
```
LalitMaganti: My problem here was with `kPerfClock` as a name not `ClockType` :)

I think it's still unclear what exactly "kPerfClock" means. I think it would be clearer if you called it "kPerfDataSourceClock" or similar and added a comment saying this is *not* necessarily the Linux perf clock.

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+   private:
+    std::unordered_map<DefaultClockType, ClockId> clock_type_to_id;
+  };
+
+  std::unique_ptr<RedactorClockSynchronizer> clock_synchronizer;
+  ClockId primary_trace_clock;
+  ClockId perf_clock;
+  std::unordered_map<SequenceId, SequenceClocks> seq_to_default_clocks;
```
LalitMaganti: base::FlatHashMap

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,126 @@
...
+    void SetClock(DefaultClockType clock_type, ClockId clock_id) {
+      clock_type_to_id[clock_type] = clock_id;
+    }
+
+   private:
+    std::unordered_map<DefaultClockType, ClockId> clock_type_to_id;
+  };
+
+  std::unique_ptr<RedactorClockSynchronizer> clock_synchronizer;
+  ClockId primary_trace_clock;
```
LalitMaganti: Should be std::optional

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,126 @@
...
+  ClockId perf_clock;
```
LalitMaganti: Should be std::optional

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,126 @@
...
+      }
+      return clock_type_to_id.at(clock_type);
+    }
+
+    void SetClock(DefaultClockType clock_type, ClockId clock_id) {
+      clock_type_to_id[clock_type] = clock_id;
+    }
+
+   private:
+    std::unordered_map<DefaultClockType, ClockId> clock_type_to_id;
```
LalitMaganti: base::FlatHashMap

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  std::unique_ptr<RedactorClockSynchronizer> clock_synchronizer;
```
LalitMaganti: Why make this a unique_ptr?

src/trace_redaction/trace_redactor.h
```diff
@@ -23,7 +23,7 @@
 #include <vector>
 
 #include "perfetto/base/status.h"
-#include "perfetto/trace_processor/trace_blob_view.h"
+#include "perfetto/trace_processor/trace_processor.h"
```
LalitMaganti: Why is this changing? This should just be trace_blob_view still?

src/trace_redaction/redactor_clock_converter.cc
```diff
@@ -0,0 +1,117 @@
...
+}
+
+bool RedactorClockSynchronizerListenerImpl::IsLocalHost() {
+  // Redactor does not support multi-machine clock conversion
+  return true;
+}
+
+RedactorClockConverter::RedactorClockConverter() {
+  // Set the default clocks for the trace
+  primary_trace_clock = protos::pbzero::BuiltinClock::BUILTIN_CLOCK_BOOTTIME;
```
LalitMaganti: These should be initialized direclty in the fields

src/trace_redaction/collect_clocks.cc
edgararriagag: Yes, good catch. Done

src/trace_redaction/collect_clocks.cc
edgararriagag: , because the function was const I was not allowed to add it as a field, but based on our discussion today, I instead made it a field with mutable to allow bypassing the constness to be able to cache this.

src/trace_redaction/collect_clocks.cc
edgararriagag: You are correct. I had overseen that pattern in the codebase. I went ahead and also modified every other instance I could find where I was passing parameters and now instead return StatusOr<T>, if I missed any, let me know

src/trace_redaction/collect_clocks.cc
edgararriagag: Done, removed.

src/trace_redaction/collect_clocks.cc
edgararriagag: Done

src/trace_redaction/collect_clocks.cc
edgararriagag: Done

src/trace_redaction/redactor_clock_converter.cc
edgararriagag: Did a different refactor for this, I am doing it lazily until we intend to retrieve if it wasn't set, I think thats a better approach

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done. Hopefully its clearer now.

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Modified to not be a unique_ptr

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done

src/trace_redaction/redactor_clock_converter.h
edgararriagag: This actually was lingering and no longer used in newest patchset. Removed

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Done

src/trace_redaction/trace_redactor.h
edgararriagag: Oversight on my end, reverted.

src/trace_redaction/redactor_clock_converter.h
edgararriagag: I actually came back to this, and did several renames and refactors for the functions. I believe the enum is just talking about the data source itself, the functions are what are more involved in clock talk, so, I renamed this to just related to the data source. I believe this way it doesn't even need extra comments to explain

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,138 @@
...
+using ClockTimestamp = RedactorClockSynchronizer::ClockTimestamp;
+
+// This class handles conversions between different clocks for trace redactor.
+//
+// This class is a wrapper for trace_processor::ClockSynchronizer with the
+// addition that it caches clocks required for conversion for different data
+// sources and it is designed to be used by the trace redactor.
+//
+// Any trace packet intends to use the redactor ProcessThreadTimeline and whose
+// clock won't be the default trace time should use this class to convert
```
vaage, line 65: I want to make sure I understand, the data in the timeline will be sanitized, meaning the logic within the timeline does not need to know about time scales. right?

src/trace_redaction/BUILD.gn
```diff
@@ -79,13 +83,15 @@ source_set("trace_redaction") {
     "../../include/perfetto/base",
     "../../include/perfetto/ext/base",
     "../../include/perfetto/protozero:protozero",
-    "../../include/perfetto/trace_processor:storage",
+    "../../include/perfetto/trace_processor",
```
vaage: Why are we including the full target? When we were starting the implementation @LalitMaganti made sure we only included the bare minimum. If I remember correctly, he even refactored some of Trace Processor to make this easier.

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,109 @@
...
+#include "protos/perfetto/trace/trace_packet_defaults.pbzero.h"
+
+using namespace perfetto::trace_processor;
+
+namespace perfetto::trace_redaction {
+
+base::Status CollectClocks::Collect(
+    const protos::pbzero::TracePacket::Decoder& packet,
+    Context* context) const {
+  clock_snapshot_.clear();
```
vaage, line 56: A collect primitive should never modify existing context state, it should only add to the state. A build primitive is responsible for creating a new object using what was collected. This is important because multiple primitives may add to the same context.

No one "owns" part of the context. A prime example of this is the timeline, multiple primitives can add to the timeline. If any one primitive was to say, "never mind, clear it" it will break the whole system.

src/trace_redaction/redactor_clock_converter.h
```diff
@@ -0,0 +1,138 @@
...
+#include "protos/perfetto/common/builtin_clock.pbzero.h"
+#include "src/trace_processor/util/clock_synchronizer.h"
+
+namespace perfetto::trace_redaction {
+
+class RedactorClockSynchronizerListenerImpl {
+ public:
+  RedactorClockSynchronizerListenerImpl();
+
+  base::Status OnClockSyncCacheMiss();
```
vaage, line 31: Why do we have this function? It has no logic in it.

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,109 @@
...
+  if (packet.has_clock_snapshot()) {
```
vaage, line 35: This nesting on this is hard to follow. If you haven't, could try playing with it to make it easier to read (I realize the 80 character width limit doesn't help us here). If you can't, it's okay.

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,109 @@
...
+  if (clock_decoder.has_timestamp()) {
+    clock_ts.timestamp = static_cast<int64_t>(clock_decoder.timestamp());
+  } else {
+    return base::ErrStatus("Could not find clock timestamp in clock snapshot");
+  }
+
+  return clock_ts;
+}
+
+base::Status CollectClocks::OnTracePacketDefaults(
```
vaage, line 107: The question on formatting applies here too.

Could you also add some comments about the clocks/timing here? I don't know anything about them and if I came across this code, I could easily break it without knowing why.

src/trace_redaction/BUILD.gn
edgararriagag: This was an explicit request from Lalit, he wanted just the top level included instead of the detailed one.

src/trace_redaction/redactor_clock_converter.h
edgararriagag: this is an implementation requirement for the ClockSynchronizerListenerImpl classes. That function is used by other implementors. The only other implementor right now is trace processor: https://github.com/google/perfetto/blob/d57eda6f192a4d5f147531d3f63355b5e36c5c78/src/trace_processor/importers/common/clock_tracker.cc#L46 so we have a stubbed version for redactor here because it is called internally.

src/trace_redaction/redactor_clock_converter.h
edgararriagag: Correct

src/trace_redaction/collect_clocks.cc
edgararriagag: IIUC the concern revolves around clock_snapshot_.clear() call. If so, this was added per from a request from @LalitMaganti  (https://github.com/google/perfetto/pull/3070#discussion_r2400059982https://github.com/google/perfetto/pull/3070#discussion_r2400059982) with the goal to cache that local variable just so that we didn't re-create the vector over and over again for performance reasons. We are still maintaining the append-only nature of CollectorPrimitive so the architecture is maintained in that regard.

src/trace_redaction/collect_clocks.cc
edgararriagag: I added some comments here, and I also added a more detailed comment in the `CollectClocks` class about clocks and collection.

src/trace_redaction/collect_clocks.cc
edgararriagag: Per our offline discussion, I refactored this into two different parsing methods to reduce indentation and make intent clearer

src/trace_redaction/BUILD.gn
LalitMaganti: I take back what I said before: this can be changed to just be storage

src/trace_redaction/BUILD.gn
```diff
@@ -79,13 +83,15 @@ source_set("trace_redaction") {
...
     "../../protos/perfetto/common:zero",
+    "../../protos/perfetto/trace:minimal_zero",
     "../../protos/perfetto/trace:non_minimal_zero",
     "../../protos/perfetto/trace/android:zero",
     "../../protos/perfetto/trace/ftrace:zero",
     "../../protos/perfetto/trace/ps:zero",
     "../trace_processor:storage_minimal",
```
LalitMaganti: This is waht I think should be removed

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,139 @@
...
+  protos::pbzero::ClockSnapshot_Clock::Decoder clock_decoder(clock_bytes);
+  if (clock_decoder.has_clock_id()) {
+    RedactorClockSynchronizer::Clock clock = RedactorClockSynchronizer::Clock(
+        static_cast<int64_t>(clock_decoder.clock_id()));
+    clock_ts.clock = clock;
+  } else {
+    return base::ErrStatus("Could not find clock id in clock snapshot");
+  }
+
+  if (clock_decoder.has_timestamp()) {
```
LalitMaganti: Early return pattern would make this read so much better and also avoid the need for variable assignment (you could just return directly).

src/trace_redaction/collect_clocks.cc
```diff
@@ -0,0 +1,139 @@
...
+    // that data source into trace time to be used by the redactor timeline.
+    if (trace_packet_defaults_decoder.has_timestamp_clock_id()) {
+      uint32_t perf_clock_id =
+          trace_packet_defaults_decoder.timestamp_clock_id();
+      context->clock_converter.SetDefaultDataSourceClock(
+          RedactorClockConverter::DataSourceType::kPerfDataSource,
+          perf_clock_id, trusted_sequence_id);
+      return base::OkStatus();
+    } else {
+      return base::ErrStatus(
```
LalitMaganti, line 118: Please use early return.

src/trace_redaction/redactor_clock_converter.cc
```diff
@@ -0,0 +1,144 @@
...
+base::Status RedactorClockSynchronizerListenerImpl::OnInvalidClockSnapshot() {
+  PERFETTO_ELOG("Invalid clocks snapshot found during redaction");
+  return base::ErrStatus("Invalid clocks snapshot found during redaction");
+}
+
+base::Status RedactorClockSynchronizerListenerImpl::OnTraceTimeClockIdChanged(
+    ClockId trace_time_clock_id [[maybe_unused]]) {
+  ++trace_time_updates_;
+  if (PERFETTO_UNLIKELY(trace_time_updates_ > 1)) {
+    // We expect the trace time to remain constant for a trace.
```
LalitMaganti, line 38: Please remove ELOGs when the errors are the same. Errors should be printed anyway.

src/trace_redaction/BUILD.gn
edgararriagag: Done, the code to let this happen is now at https://github.com/google/perfetto/pull/3200 once that other CL lands and I rebase here this diff should go away.

src/trace_redaction/collect_clocks.cc
edgararriagag: Done

src/trace_redaction/redactor_clock_converter.cc
edgararriagag: Done

src/trace_redaction/collect_clocks.cc
edgararriagag: Done

edgararriagag:
Once https://github.com/google/perfetto/pull/3200 lands, I will change the base of this CL to main.

src/trace_redaction/BUILD.gn
LalitMaganti: This should point to util now.

edgararriagag: Done

PR #2927, opened 2025-09-12
Add support for callstack profiling in trace redactor

edgararriagag:
Add support for perf samples in trace redactor which in turn allows for call stack collection of all the hardware and sw counters currently supported with linux.perf collection

PR #3071, opened 2025-09-24
Add prune perf events transform primitive to trace redactor

edgararriagag:
This primitive allows allows redaction of perf events from traces for target process which is a requirement to add support for call stacks in redactor.


LalitMaganti:
Please follow guidance on go/perfetto-github-instructions on correctly stacking PRs. Concretely, this PR should have your other PR as a base.

edgararriagag:
@LalitMaganti this PR should also be ready for review. I modified the base as well

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,122 @@
...
+  auto perf_sample =
+      decoder.FindField(protos::pbzero::TracePacket::kPerfSampleFieldNumber);
+  if (!perf_sample.valid()) {
+    // No perf samples found, skip.
+    return base::OkStatus();
+  }
+
+  protozero::HeapBuffered<protos::pbzero::TracePacket> message;
+
+  auto time_field =
```
LalitMaganti: Given you are basically using ~all the fields of perf sample, I think you'd be better off with using PerfSample::Decoder instead.

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,122 @@
...
+      decoder.FindField(protos::pbzero::TracePacket::kTimestampFieldNumber);
+  PERFETTO_DCHECK(time_field.valid());
+
+  auto trace_packet_clock_id_field = decoder.FindField(
+      protos::pbzero::TracePacket::kTimestampClockIdFieldNumber);
+  int64_t trace_packet_clock_id = -1;
```
LalitMaganti: Please use std::optional instead of -1

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,122 @@
...
+    const Context& context,
+    uint64_t ts,
+    int64_t trace_packet_clock_id,
+    int64_t trusted_packet_sequence_id,
+    protozero::Field& perf_sample_field,
+    protos::pbzero::TracePacket* message) const {
+  protozero::ProtoDecoder decoder(perf_sample_field.as_bytes());
+
+  auto pid = decoder.FindField(protos::pbzero::PerfSample::kPidFieldNumber);
+  PERFETTO_DCHECK(pid.valid());
```
LalitMaganti: Again you are using all the fields, Use a decoder and simplify your life.

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,122 @@
...
+
+  uint64_t trace_ts;
+  ClockId clock_id = trace_packet_clock_id;
+  if (PERFETTO_LIKELY(trace_packet_clock_id == -1)) {
+    // No override provided, so grab the default clock for this sequence id.
+    PERFETTO_DCHECK(trusted_packet_sequence_id != -1);
```
LalitMaganti: Why is this a DCHECK and not a returned error?

edgararriagag: Good catch, made it return errors instead as it is indeed an error.

src/trace_redaction/prune_perf_events.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events.cc
edgararriagag: Done. I modified to use PerfSample::Decoder for most field parsings, for copying the entire message I still used ProtoDecoder (the for loop that appends each non redacted field). I tried looking for an alternative within the decoders and I couldn't find it, if there's a better way, let me know, but hopefully the way I did is what you were expecting.

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,131 @@
...
+
+base::Status PrunePerfEvents::OnPerfSample(
+    const Context& context,
+    uint64_t ts,
+    std::optional<int64_t> trace_packet_clock_id,
+    std::optional<int64_t> trusted_packet_sequence_id,
+    protozero::Field& perf_sample_field,
+    protos::pbzero::TracePacket* message) const {
+  protos::pbzero::PerfSample::Decoder decoder(perf_sample_field.as_bytes());
+  if (!decoder.has_pid()) {
```
LalitMaganti, line 92: Imo this should be an error

edgararriagag: Done

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,130 @@
...
+#include "src/trace_redaction/trace_redaction_framework.h"
+
+#include "protos/perfetto/trace/profiling/profile_packet.pbzero.h"
+
+using namespace perfetto::trace_processor;
+namespace perfetto::trace_redaction {
+
+base::Status PrunePerfEvents::Transform(const Context& context,
+                                        std::string* packet) const {
+  if (packet == nullptr || packet->empty()) {
```
vaage: Please remove. This case is handled by the redactor and should not be checked inside of the primitive.

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,130 @@
...
+  protos::pbzero::TracePacket::Decoder packet_decoder(*packet);
+  if (!packet_decoder.has_perf_sample()) {
+    // No perf samples found, skip.
+    return base::OkStatus();
+  }
+
+  protozero::HeapBuffered<protos::pbzero::TracePacket> message;
+
+  if (!packet_decoder.has_timestamp()) {
+    return base::ErrStatus(
```
vaage: Please follow the general error text used else where. You can look `verify_integrity` for an example.

vaage: Please move the check closer to the use case.

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,130 @@
...
+#include "perfetto/ext/base/status_macros.h"
+#include "src/trace_redaction/proto_util.h"
+#include "src/trace_redaction/trace_redaction_framework.h"
+
+#include "protos/perfetto/trace/profiling/profile_packet.pbzero.h"
+
+using namespace perfetto::trace_processor;
+namespace perfetto::trace_redaction {
+
+base::Status PrunePerfEvents::Transform(const Context& context,
```
vaage, line 33: Where are the tests?

edgararriagag: The tests are the integration tests that were added in the following CL see: https://github.com/google/perfetto/pull/2927/commits/00605d77436bf2af7e0fe2eb38424656f03bfd71#diff-1cdd34bc9a25db5ff947c5c2e7e37ca33143980e91b617a548fdf69bdbbba2e6 I put them on the other CL because they required the clock collection and everything to be put together, I figured unit tests would end up being more fabricated and redudant which is why I skipped for this specific class thus not extremely useful.

vaage: Thanks. That'll work for me.

src/trace_redaction/prune_perf_events.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events.cc
```diff
@@ -0,0 +1,125 @@
...
+    const Context& context,
+    uint64_t ts,
+    std::optional<int64_t> trace_packet_clock_id,
+    std::optional<int64_t> trusted_packet_sequence_id,
+    protozero::Field& perf_sample_field,
+    protos::pbzero::TracePacket* message) const {
+  protos::pbzero::PerfSample::Decoder decoder(perf_sample_field.as_bytes());
+  if (!decoder.has_pid()) {
+    // PID is required to compute belongs-to process relationship
+    return base::ErrStatus("Redactor could not find pid in perf sample.");
```
vaage: Please update all error messages to follow the pattern established in other primitives. The pattern is meant to make it easier to track down the failure. You did this at line 60, please do this for any message.

vaage (changes requested):

src/trace_redaction/prune_perf_events.cc
edgararriagag: Done

vaage (approved):

PR #3075, opened 2025-09-24
Add thread wait reasons to CSwitch events

g-aattar:
This change enhances the ETW CSwitch event parser to include the reason why a thread is entering a waiting state.

Previously, we could see that a thread was blocked, but not the underlying cause. By extracting the old_thread_wait_reason field from the CSwitch event, we can now provide specific details for the blockage (e.g., "UserRequest", "WrQueue", "PageIn").

This reason is recorded in the blocked_reason column in the thread_state table, making it directly visible in the Perfetto UI when inspecting a thread's state. This provides significantly more context for debugging performance issues and understanding thread scheduling behaviour on Windows.


src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -259,5 +283,37 @@ StringId EtwParser::TaskStateToStringId(int64_t task_state_int) {
   return context_->storage->InternString(etw_states[state]);
 }
 
+// Translates a Windows ETW wait reason enum to a string.
+// See: https://learn.microsoft.com/en-us/windows/win32/etw/cswitch
+StringId WaitReasonToStringId(TraceProcessorContext* context, uint8_t reason) {
+  static const std::map<uint8_t, base::StringView> wait_reason_map = {
```
LalitMaganti: Didn't we agree from the previous CL this should be a vector + linear scan? :)

src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -259,5 +283,37 @@ StringId EtwParser::TaskStateToStringId(int64_t task_state_int) {
...
+      {0x1C, "WrPushLock"},       {0x1D, "WrMutex"},
+      {0x1E, "WrQuantumEnd"},     {0x1F, "WrDispatchInt"},
+      {0x20, "WrPreempted"},      {0x21, "WrYieldExecution"},
+      {0x22, "WrFastMutex"},      {0x23, "WrGuardedMutex"},
+      {0x24, "WrRundown"},
+  };
+
+  auto it = wait_reason_map.find(reason);
+  if (it == wait_reason_map.end()) {
+    return context->storage->InternString("Unknown");
```
LalitMaganti: This should be interned in the constructor and stored.

g-aattar: Done

src/trace_processor/importers/etw/etw_parser.cc
g-aattar: Sorry did not mean to send this patch for review at that time, but thanks for the comment. Changed it for the suggested solution.

src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -259,5 +293,54 @@ StringId EtwParser::TaskStateToStringId(int64_t task_state_int) {
...
+      "WrYieldExecution",  // 0x21
+      "WrFastMutex",       // 0x22
+      "WrGuardedMutex",    // 0x23
+      "WrRundown",         // 0x24
+  };
+
+  if (reason >= std::size(wait_reason_map)) {
+    return unknown_wait_reason_id_;
+  }
+  return context_->storage->InternString(wait_reason_map[reason]);
```
LalitMaganti, line 340: Instead of unknown like this actually, can we please reutnr null string id? NULL literally represents unknown in Perfetto so there's no point having a speical "unknown" string instead.

g-aattar: Done

PR #3077, opened 2025-09-24
Fix bug in PerfettoApiTest#NoFlushFlag/System

primiano:
Fix a test. This test has always been buggy as it was missing
a sync fence. It was passing by accident just because the
old task runner was slower.
The new task runner makes the bug visible because runs tasks
faster.
This can be repro locally by setting
enable_perfetto_lockfree_taskrunner=true

Bug: 447118803


LalitMaganti:
Why was this not being caught by our CI? Was it too slow on both old/new?

primiano:
> Why was this not being caught by our CI? Was it too slow on both old/new?

Because I have not enabled the LFTR in our GNI file. I just remembered that I wanted to wait for Gantry to stick, and only later enable everywhere else.

PR #3091, opened 2025-09-25
Rename TraceBuffer -> TraceBufferV1, extract interface

primiano:
Extract a common interface for TraceBufferV1 and V2.
Renames TraceBuffer to TraceBufferV1, in preparation for TBV2

Bug: b/447426810


src/tracing/service/trace_buffer_v1_unittest.cc
```diff
@@ -1935,7 +1935,7 @@ TEST_F(TraceBufferTest, Clone_FragmentsOutOfOrder) {
   {
     // Create a snapshot before the middle 'b' chunk is copied. Only 'a' should
     // be readable at this point.
-    std::unique_ptr<TraceBuffer> snap = trace_buffer()->CloneReadOnly();
+    auto snap = std::unique_ptr<TraceBufferV1>(static_cast<TraceBufferV1*>(trace_buffer()->CloneReadOnly().release()));
```
LalitMaganti: Maybe create a helper function to do this: seems like it happens a bunch and maybe we want to refactor/remove this in the future?

src/tracing/service/trace_buffer.h
```diff
@@ -167,577 +60,79 @@ class TraceBuffer {
...
-    // record.size >= |size| + sizeof(ChunkRecord) (== if no rounding).
-    PERFETTO_DCHECK(size <= ChunkRecord::kMaxSize);
-    PERFETTO_DCHECK(record.size >= sizeof(record));
-    PERFETTO_DCHECK(record.size % sizeof(record) == 0);
-    PERFETTO_DCHECK(record.size >= size + sizeof(record));
-    DcheckIsAlignedAndWithinBounds(wptr);
-
-    // We may be writing to this area for the first time.
-    EnsureCommitted(static_cast<size_t>(wptr + record.size - begin()));
+  // new buffer will be reset, as if no Read() had been called.
```
LalitMaganti, line 103: Performance-wise, is it OK that all these methods are virtual? We good with the cost for all of them?

src/tracing/service/trace_buffer_v1_unittest.cc
primiano: yeah hold on a sec i'm fixing this :)

src/tracing/service/trace_buffer.h
primiano: TBH thy are called quite rarely from TracingService. They are not hot paths.
The implementation (in tracing_buffer_v1.cc) uses directly size_ and used_size_ inside.

ultimately they will become non-virtual again, but I need these to be able to deal with both impl

src/tracing/service/trace_buffer_v1_unittest.cc
primiano: ok now is ready

src/tracing/service/trace_buffer.h
LalitMaganti: Ack.

src/tracing/service/trace_buffer_v1_unittest.cc
LalitMaganti: LGTM thanks!

PR #3092, opened 2025-09-25
TraceBufferV2

primiano:
NOT FOR REVIEW YET
All test pass, but there are a dozen small TODOs
that I need to think about.

Also design doc is really due, as I'm going to write
80% of all this in a month and I don't wish anybody to
ever touch this code.

Bug: b/447426810

**Stack:**
- [#3091](https://github.com/google/perfetto/pull/3091) (tb_extract)
  - **[This PR] (tbv2)**

### DIFF
--- a/include/perfetto/ext/base/circular_queue.h
+++ b/include/perfetto/ext/base/circular_queue.h
@@ -160,6 +160,8 @@ class CircularQueue {
     }
 
    private:
+    friend class CircularQueue<T>;
+
     inline void Add(difference_type offset) {
       pos_ = static_cast<uint64_t>(static_cast<difference_type>(pos_) + offset);
       PERFETTO_DCHECK(pos_ <= queue_->end_);
@@ -172,6 +174,11 @@ class CircularQueue {
     uint32_t generation_;
 #endif
   };
+
+  using ReverseIterator = std::reverse_iterator<Iterator>;
+  using value_type = T;
+  using const_iterator = Iterator;
+  using iterator = Iterator;
 
   explicit CircularQueue(size_t initial_capacity = 1024) {
     Grow(initial_capacity);
@@ -233,6 +240,30 @@ class CircularQueue {
 
   void pop_front() { erase_front(1); }
 
+  void InsertBefore(Iterator pos, T value) {
+    increment_generation();
+    PERFETTO_DCHECK(pos.queue_ == this);
+    PERFETTO_DCHECK(pos.pos_ >= begin_ && pos.pos_ <= end_);
+
+    if (PERFETTO_UNLIKELY(size() >= capacity_))
+      Grow();
+
+    // Move all elements one step forward from end_ - 1 to pos.pos_.
+    for (uint64_t i = end_++; i > pos.pos_; --i) {
+      T* dst = Get(i);
+      T* src = Get(i - 1);
+      new (dst) T(std::move(*src));
+      src->~T();
+    }
+
+    // Place the new value at pos.
+    new (Get(pos.pos_)) T(std::move(value));
+  }
+
+  void InsertAfter(ReverseIterator pos, T value) {
+    InsertBefore(pos.base(), std::move(value));
+  }
+
   void clear() { erase_front(size()); }
 
   void shrink_to_fit() {
@@ -252,6 +283,21 @@ class CircularQueue {
 
   Iterator begin() { return Iterator(this, begin_, generation()); }
   Iterator end() { return Iterator(this, end_, generation()); }
+  Iterator begin() const {
+    return Iterator(const_cast<CircularQueue*>(this), begin_, generation());
+  }
+  Iterator end() const {
+    return Iterator(const_cast<CircularQueue*>(this), end_, generation());
+  }
+
+  ReverseIterator rbegin() { return ReverseIterator(end()); }
+  ReverseIterator rend() { return ReverseIterator(begin()); }
+
+  Iterator Find(const T& value) {
+    auto it = std::find(begin(), end(), value);
+    return it;
+  }
+
   T& front() { return *begin(); }
   T& back() { return *(end() - 1); }
 
--- a/src/tracing/service/trace_buffer_v2.cc
+++ b/src/tracing/service/trace_buffer_v2.cc
@@ -1,0 +1,1150 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/tracing/service/trace_buffer_v2.h"
+
+#include <limits>
+
+#include "perfetto/base/logging.h"
+#include "perfetto/ext/base/utils.h"
+#include "perfetto/ext/tracing/core/client_identity.h"
+#include "perfetto/ext/tracing/core/shared_memory_abi.h"
+#include "perfetto/ext/tracing/core/trace_packet.h"
+#include "perfetto/protozero/proto_utils.h"
+#include "src/base/intrusive_list.h"
+
+// TODO DNS disable TRACE_BUFFER_VERBOSE_LOGGING.
+#define TRACE_BUFFER_VERBOSE_LOGGING() 1  // Set to 1 when debugging unittests.
+
+#if TRACE_BUFFER_VERBOSE_LOGGING()
+#define TRACE_BUFFER_DLOG PERFETTO_DLOG
+#else
+#define TRACE_BUFFER_DLOG(...) base::ignore_result(__VA_ARGS__)
+#endif
+
+using protozero::proto_utils::ParseVarInt;
+namespace proto_utils = ::protozero::proto_utils;
+
+namespace perfetto {
+
+namespace {
+constexpr uint8_t kFirstPacketContinuesFromPrevChunk =
+    SharedMemoryABI::ChunkHeader::kFirstPacketContinuesFromPrevChunk;
+constexpr uint8_t kLastPacketContinuesOnNextChunk =
+    SharedMemoryABI::ChunkHeader::kLastPacketContinuesOnNextChunk;
+constexpr uint8_t kChunkNeedsPatching =
+    SharedMemoryABI::ChunkHeader::kChunkNeedsPatching;
+
+// The only thing that causes incomplete-ness is SMB scraping, when calling
+// CopyChunkUntrusted(...chunk_complete=false). This has nothing to do with
+// patching.
+constexpr uint8_t kChunkIncomplete = 0x80;
+
+// Mask out the flags that don't come from the ABI like kChunkIncomplete.
+constexpr uint8_t kFlagsMask = SharedMemoryABI::ChunkHeader::kFlagsMask;
+
+// Compares two ChunkID(s) in a wrapping 32-bit ID space.
+// Returns:
+//   -1 if a comes before b in modular (wrapping) order.
+//    0 if a == b
+//   +1 if a comes after b in modular order.
+//
+// The key idea is that the order between two distinct IDs is determined by
+// whether the distance from a to b is less than 2^31 (half the range).
+// Many other TCP/IP stacks do the same, e.g.
+// https://github.com/openbsd/src/blob/master/sys/netinet/tcp_seq.h#L43
+int ChunkIdCompare(ChunkID a, ChunkID b) {
+  if (a == b)
+    return 0;
+  return (static_cast<int32_t>(a - b) < 0) ? -1 : 1;
+}
+
+// The number of the latest empty SequenceState(s) to keep around.
+constexpr size_t kKeepLastEmptySeq = 1024;
+
+// The threashold when we start scanning and deleting the oldest sequences.
+const size_t kEmptySequencesGcTreshold = kKeepLastEmptySeq + 128;
+}  // namespace.
+
+namespace internal {
+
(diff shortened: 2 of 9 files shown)


primiano:
@ddiproietto @sashwinbalaji  this is finally ready for review!

docs/design-docs/trace-buffer.md
```diff
@@ -0,0 +1,735 @@
...
+When chunks are written via `CopyChunkUntrusted()` a new `TBChunk` is
+allocated in the buffer's PagedMemory using the usual bump-pointer pattern
+you'd expect from a ring-buffer. Chunks are variable-size, and are stored
+contiguosly with 32-bit alignment.
+
+The offset of the chunk is also appended in the `SequenceState.chunks` list.
+
+After the first wrapping, writing a chunk involves deleting one or more
+existing chunks. The deletion operation `RemoveNextChunksFor()` is as complex
+as a redback, because reconstructs packets being deleted in order, to pass
```
sashwinbalaji: typos e.g. redback maybe use AI for a passthrough

include/perfetto/ext/base/circular_queue.h
```diff
@@ -173,6 +175,11 @@ class CircularQueue {
 #endif
   };
 
+  using ReverseIterator = std::reverse_iterator<Iterator>;
+  using value_type = T;
+  using const_iterator = Iterator;
```
sashwinbalaji, line 180: since "// - It doesn't bother with const-correctness." and it's not being used anywhere remove this ?

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+// When a sequence has 0 chunks we could delete it straight away. However doing
+// do would remove also the last_chunk_id_consumed used to detect data losses.
+// Here we have to balance the tradeoff between:
+// - Bounding memory usage: if we have a lot of writer threads, we can't just
+//   keeping growing the SequenceStates without bounds.
+// - Retaining our ability to detect data losses for sporadic writers.
+// Here the decision is to keep the last 1024 (kKeepLastEmptySeq) empty
+// sequences and delete anything older. We have some hysteresys and start the GC
+// process only after hitting 1024+128 empty sequences. This is to avoid
+// thrashing with a std::sort every time one sequence becomes empty.
```
sashwinbalaji: hmm, how about replacing  empty_sequences_ with a set {age_for_gc, ProducerAndWriterID} .

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+  if (a == b)
+    return 0;
+  return (static_cast<int32_t>(a - b) < 0) ? -1 : 1;
+}
+
+// The number of the latest empty SequenceState(s) to keep around.
+constexpr size_t kKeepLastEmptySeq = 1024;
+
+// The threashold when we start scanning and deleting the oldest sequences.
+const size_t kEmptySequencesGcTreshold = kKeepLastEmptySeq + 128;
```
sashwinbalaji: why not constexpr ?

src/tracing/service/trace_buffer_v2.h
```diff
@@ -0,0 +1,599 @@
...
+
+// +---------------------------------------------------------------------------+
+// | FragIterator                                                              |
+// +---------------------------------------------------------------------------+
+// A tokenized fragment in the buffer. Holds the fragment size & boundaries.
+// This struct is returned by FragIterator when tokenizing the fragments in
+// CopyChunkUntrusted() or reading the buffer.
+// Frag instances are short lived.
+struct Frag {
+  enum FragType : uint8_t {
```
sashwinbalaji, line 199: enum class

src/tracing/service/trace_buffer_v2.h
```diff
@@ -0,0 +1,599 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_TRACING_SERVICE_TRACE_BUFFER_V2_H_
+#define SRC_TRACING_SERVICE_TRACE_BUFFER_V2_H_
+
+#include <stdint.h>
```
sashwinbalaji, line 20: for clarification: any guideline for c (stdint.h) vs c++(cstdint), I see both in our codebase.

rsavitski: driveby: there was a concrete reason for inttypes.h vs cinttypes ([presubmit](https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/tools/run_presubmit;l=648;drc=8dcf87f14175f52c2efb9d2d6f503b9499fc50bf)). For other headers, I don't think we have strict guidance so different people do different things.

docs/design-docs/trace-buffer.md
primiano: Did an AI proofread of the doc. good idea:)I

include/perfetto/ext/base/circular_queue.h
primiano: We don't bother with const correctness, but unfortunately some container users (in the specific case gtest) need to see a const_iterator trait otherwise it fails building

If I remove it:

```
In file included from ../../src/base/circular_queue_unittest.cc:21:
In file included from ../../test/gtest_and_gmock.h:50:
In file included from ../../buildtools/googletest/googlemock/include/gmock/gmock-function-mocker.h:42:
In file included from ../../buildtools/googletest/googlemock/include/gmock/gmock-spec-builders.h:75:
../../buildtools/googletest/googlemock/include/gmock/gmock-matchers.h:3238:28: error: no type named 'const_iterator' in 'perfetto::base::CircularQueue<int>'
 3238 |     typename StlContainer::const_iterator it = stl_container.begin();
 ```

File: src/tracing/service/trace_buffer_v2.cc (hunk 4 above)
Comment (primiano) [reply src/tracing/service/trace_buffer_v2.cc]: Oh just an oversight. fixed.

File: src/tracing/service/trace_buffer_v2.cc (hunk 3 above)
Comment (primiano) [reply src/tracing/service/trace_buffer_v2.cc]: 1) (minor) last time i benchmarked, i found that's it's usually faster to to a std::sort pass (because once you start sorting everything becomes cache hot) rather than using a sorted tree (which adds more operation to every path when we empty a sequence, which is cache-cold). 

2) (bigger) set<> (or any Small/Flat alternative) uses more code. given this is something that we should hit quite rarely, it's not worth bloating the binary size with it.

File: src/tracing/service/trace_buffer_v2.h (hunk 6 above)
Comment (primiano) [reply src/tracing/service/trace_buffer_v2.h:20]: Yeah what Ryan said. I'm honestly fine with both. Just how I got used I guess.

File: src/tracing/service/trace_buffer_v2.h (hunk 5 above)
Comment (primiano) [reply src/tracing/service/trace_buffer_v2.h:199]: I tend to avoid them when there is one/few enums well nested in a small struct, as they create just churn.
If I use enum class, then I need to do everywhere Frag::FragType::kFrag... which is longer and makes the code harder to read.

I'm all up for enum classes when they are in the global/namespace scope, or when a class has many and the values collide.

**Comment by primiano** [comment]
@sashwinbalaji mind taking another look when you have some time. i'm back on this

File: src/tracing/service/trace_buffer_v2.h [hunk 7]
```diff
@@ -0,0 +1,599 @@
...
+// +---------------------------------------------------------------------------+
+// | TBChunk                                                                   |
+// +---------------------------------------------------------------------------+
+// TBChunk is the struct, stored in the trace buffer memory as a result of
+// calling CopyChunkUntrusted from a SMB chunk.
+// TBChunk exists only in the TraceBuffer PagedMemory `data_`, never on the
+// stack or on the heap. It is followed by the fragments and alignment padding.
+// A TBChunk is very similar to a SMB chunk with the following caveats:
+// - The sizeof() both is the same (16 bytes). This is very important to keep
+//   patches offsets consistent.
```
sashwinbalaji: not a caveat per se

src/tracing/service/trace_buffer_v2.h
```diff
@@ -0,0 +1,599 @@
...
+// - The SMB chunk maintains a counter of fragments. TBChunk instead does
+//   byte-based bookkeeping, as that reduces the complexity of the iterators.
+// - The layout of the fields is slightly different, but they both contains
+//   ProducerID, WriterID, ChunkID, fragment counts/sizes and flags.
+//   The SMB chunk layout is an ABI. The TCHunk layout is not: it is an
+//   implementation detail and can be changed.
+// - TBChunk maintains a basic checksum for each chunk (only for debug builds).
```
sashwinbalaji, line 67: nit: put cheksum behind PERFETTO_DCHECK_IS_ON ?

src/tracing/service/trace_buffer_v2.h
```diff
@@ -0,0 +1,599 @@
...
+  // reduce internal fragmentation.
+  // However, doing so requires some careful thinking as that creates the
+  // opportunity for more interesting edge cases, where overwriting a chunk
+  // leaves less than sizeof(TBChunk), leaving no space for a padding header
+  // after it. See TraceBufferV2Test.Overwrite_SizeDiffLessThanChunkHeader.
+  static inline constexpr size_t alignment() { return sizeof(TBChunk); }
+
+  static inline size_t OuterSize(size_t sz) {
+    return base::AlignUp<alignment()>(sizeof(TBChunk) + sz);
+  }
```
sashwinbalaji, line 135: considering the todo in [utils.h ](https://github.com/google/perfetto/blob/main/include/perfetto/ext/base/utils.h#L91-L102) switch to non-templated constepxr version ?

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+  }
+  static_assert(kEmptySequencesGcTreshold >= kKeepLastEmptySeq);
+  if (empty_seqs.size() < kEmptySequencesGcTreshold)
+    return;
+
+  std::sort(empty_seqs.begin(), empty_seqs.end(),
+            [](const SeqIterator& a, const SeqIterator& b) {
+              return a->second.age_for_gc < b->second.age_for_gc;
+            });
+
```
sashwinbalaji, line 1068: std::nth_element as we just want the n old elements and not care about their order.

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+// When a sequence has 0 chunks we could delete it straight away. However doing
+// do would remove also the last_chunk_id_consumed used to detect data losses.
+// Here we have to balance the tradeoff between:
+// - Bounding memory usage: if we have a lot of writer threads, we can't just
+//   keeping growing the SequenceStates without bounds.
+// - Retaining our ability to detect data losses for sporadic writers.
+// Here the decision is to keep the last 1024 (kKeepLastEmptySeq) empty
+// sequences and delete anything older. We have some hysteresys and start the GC
+// process only after hitting 1024+128 empty sequences. This is to avoid
```
sashwinbalaji: putting in one place, some typos / comment suggestions:
- hysteresys -> hysteresis
- 1024+128 -> 1024+128 (kEmptySequencesGcTreshold)
- poisition -> position
- SBM -> SMB

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+    }
+  }
+
+  if (chunk == nullptr) {
+    stats_.set_patches_failed(stats_.patches_failed() + 1);
+    return false;
+  }
+
+  size_t payload_size = chunk->payload_size;
+  PERFETTO_DCHECK(chunk->chunk_id == chunk_id);
```
sashwinbalaji: nit: seems redundant

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+  // violation.
+  if (PERFETTO_UNLIKELY(frag_size_u64 ==
+                        SharedMemoryABI::kPacketSizeDropPacket)) {
+    trace_writer_data_drop_ = true;
+    return std::nullopt;
+  }
+
+  // We don't need to do anything special about padding chunks, because their
+  // payload_avail is always 0. We naturally return nullopt on the check below.
+
```
sashwinbalaji: nit: remove empty line

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+    // ChunkSeqReader(kEraseMode) must delete the chunk once
+    // ReadNextPacketInSeqOrder() returns false.
+    PERFETTO_DCHECK(chunk->is_padding());
+
+    stats_.set_chunks_overwritten(stats_.chunks_overwritten() + 1);
+    stats_.set_bytes_overwritten(stats_.bytes_overwritten() + chunk_outer_size);
+  }
+
+  for (size_t off = wr_; off < clear_end && off < used_size_;) {
```
sashwinbalaji, line 948: comment explaining purpose of the loop aligning with above for loop.

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -0,0 +1,1141 @@
...
+  auto seq_it = sequences_.find(seq_key);
+  bool seq_is_new = false;
+  if (seq_it == sequences_.end()) {
+    TRACE_BUFFER_DLOG("  Added seq %x", seq_key);
+    auto it_and_inserted = sequences_.emplace(
+        seq_key,
+        SequenceState(producer_id_trusted, writer_id, client_identity_trusted));
+    seq_it = it_and_inserted.first;
+    seq_is_new = true;
+  }
```
sashwinbalaji: use try_emplace ?

primiano: ooh good idea. TIL about try_emplace. (old man waving at the sky) "back in my days we didn't have such luxuries" :D

src/tracing/service/trace_buffer_v2.h
primiano: Actually for now I think i prefer turning the DCHECK into a CHECK. Added a comment about it. i'll remove the CHECK later once TBV2 proves itself.
this can spot most bugs.

PR #3096, opened 2025-09-26
ui: Create a single entry point for slice tracks based on DST

stevegolton:
Currently we have two (main) entry points for creating slice tracks, that have very different parameters and behaviors:
1. `new DatasetSliceTrack()`
2. `createQuerySliceTrack()`

> Note: We have a lot more bespoke factory functions which are based on DatasetSliceTrack such as createTraceProcessorTrack(), but they are special purpose dedicated track factories and usually contained within the plugins that use them. They are not general purpose factories to create a track from a query.

This patch normalizes the entry points for creating a slice track from a query.

- Add a default details panel to DatasetSliceTrack if none provided which simply lists the column names and values in the provided schema for the selected slice.
- Relax the requirement to pass an `id` column to `DatasetSliceTrack`. If omitted, an `id` column will be generated using `ROW_NUMBER()`.
- Add two new static factory functions to DatasetSliceTrack:
  - `DatasetSliceTrack.createMaterialized()` - Materializes a perfetto table from the provided dataset - this function is async (returns a promise) as the table is materialized eagerly.
  - `DatasetSliceTrack.create()` - A direct alias for `new DatasetSliceTrack()`, exists purely for symmetry with `createMaterialized()`. Soon we will make the constructor private so `create` must be used instead, but this is out of scope for this patch. This function is synchronous.
- Replace all calls to `createQuerySliceTrack()` with `DatasetSliceTrack.createMaterialized()`, modifying the paramters accordingly. Some of them arguably don't need to materialize a table and could use `DatasetSliceTrack.create()` instead but that decision is out of scope for this patch.
- Remove `createQuerySliceTrack()` entirely.

Migration FAQ: 
- Q: With createQuerySliceTrack, I could provide column mappings for ts, dur, and name. How do I do that with DST?
   A: No one was actually using that feature. If you need to map these columns, do it in the query.
- Q: How do I provide extra columns to be displayed in the details panel?
   A: Add them to the schema. If you don't know or care about the type of these columns, use `UNKOWN`.

### DIFF
--- a/ui/src/components/details/thread_slice_details_tab.ts
+++ b/ui/src/components/details/thread_slice_details_tab.ts
@@ -210,25 +210,6 @@ async function getSliceDetails(
   return getSlice(engine, asSliceSqlId(id));
 }
 
-/**
- * A comprehensive details panel for slices from the `slice` table.
- *
- * This panel provides detailed information about a slice including:
- * - Core slice properties (name, duration, timestamps, thread/process info)
- * - Flow connections (preceding and following flows between slices)
- * - Arguments and custom attributes
- * - Contextual analysis tools and queries
- *
- * Note: This panel includes some domain-specific functionality (like binder
- * transaction analysis and monitor contention tracking) that arguably belongs
- * in specialized panels, but is included here for convenience and
- * discoverability. This makes it a useful "Swiss Army knife" for slice
- * analysis, even if not architecturally pure.
- *
- * The panel dynamically adapts its content based on the slice type and
- * available data, providing relevant contextual options through the "Contextual
- * Options" menu.
- */
 export class ThreadSliceDetailsPanel implements TrackEventDetailsPanel {
   private sliceDetails?: SliceDetails;
   private breakdownByThreadState?: BreakdownByThreadState;
--- a/ui/src/components/tracks/dataset_slice_track.ts
+++ b/ui/src/components/tracks/dataset_slice_track.ts
@@ -54,7 +54,7 @@ export interface InstantStyle {
   render(ctx: CanvasRenderingContext2D, rect: Size2D & Point2D): void;
 }
 
-export interface SliceTrackAttrs<T extends DatasetSchema> {
+export interface DatasetSliceTrackAttrs<T extends DatasetSchema> {
   /**
    * The trace object used by the track for accessing the query engine and other
    * trace-related resources.
@@ -201,7 +201,7 @@ export type RowSchema = {
 type SliceWithRow<T> = Slice & {row: T};
 
 function getDataset<T extends DatasetSchema>(
-  attrs: SliceTrackAttrs<T>,
+  attrs: DatasetSliceTrackAttrs<T>,
 ): SourceDataset<T> {
   const dataset = attrs.dataset;
   return typeof dataset === 'function' ? dataset() : dataset;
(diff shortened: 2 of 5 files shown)


ui/src/components/details/thread_slice_details_tab.ts
```diff
@@ -210,6 +210,25 @@ async function getSliceDetails(
...
+ * A comprehensive details panel for slices from the `slice` table.
+ *
+ * This panel provides detailed information about a slice including:
+ * - Core slice properties (name, duration, timestamps, thread/process info)
+ * - Flow connections (preceding and following flows between slices)
+ * - Arguments and custom attributes
+ * - Contextual analysis tools and queries
+ *
+ * Note: This panel includes some domain-specific functionality (like binder
+ * transaction analysis and monitor contention tracking) that arguably belongs
```
LalitMaganti: In the long term:
1) this class should be renamed
2) we should disaggregate these extra things into separate components which are "injected" in to make this more composable

ui/src/components/tracks/dataset_slice_track.ts
```diff
@@ -178,31 +188,99 @@ export interface DatasetSliceTrackAttrs<T extends DatasetSchema> {
...
-export class DatasetSliceTrack<T extends ROW_SCHEMA> extends BaseSliceTrack<
+export class DatasetSliceTrack<T extends RowSchema> extends BaseSliceTrack<
   SliceWithRow<T>,
   BaseRow & T
 > {
   readonly rootTableName?: string;
 
-  constructor(private readonly attrs: DatasetSliceTrackAttrs<T>) {
+  /**
+   * Factory function to create a DatasetSliceTrack. This is purely an alias for new
```
simonmacm, line 217: Consider making the constructor private to a) push this symmetry out into plugins b) make life easier if this method ever needs to do anything.

simonmacm (changes requested):

ui/src/components/tracks/dataset_slice_track.ts
simonmacm: Not being very used to github I'm not sure how strongly "requested changes" is enforced.

I would mean it in the critique "unresolved" sense of "I want to be sure the CL author has at least read this and clicked something".

LalitMaganti: I think it was stated in the constructor Steve plans to do this in a followup.

simonmacm: Oh, in the description yeah. Missed that.

I was surprised to see new uses of the constructor added to plugins in this CL, that's all.

ui/src/components/details/thread_slice_details_tab.ts
stevegolton: Longish term yes I totally agree. This details panel implementation is ugly. The addition of this comment was actually an error on my part (it's for a future patch). I'll remove it for now.

ui/src/components/tracks/dataset_slice_track.ts
stevegolton: It will be private. New uses of the constructor was a find+replace fumble on my part. I've changed them to use the static factory methods now.

> Not being very used to github I'm not sure how strongly "requested changes" is enforced.

I believe as long as someone resolved the comment then GitHub is happy. AFAICT there's not a lot of difference between changes requested vs just leaving a comment, aside from a simple message of intent to the author.

stevegolton: Huh, OK I tell a lie. AFAICT a 'Requesting changes' review can block the PR from being merged even if it has another 'Approved' review.

simonmacm (approved):

PR #3099, opened 2025-09-26
De-couple jank and latency slices for SysUI CUJs.

devianb:
Due to the existing implementation of jank and latency slices being filtered together, the jank and latency CUJ tables may not have incremental cuj_ids everytime. This can break functionality in traces where both types of CUJs are present, eg. frame pinning plugin.

Bug: 447104599
Test: manual trace test

### DIFF
--- a/src/base/lock_free_task_runner.cc
+++ b/src/base/lock_free_task_runner.cc
@@ -24,8 +24,8 @@
 // Keep the \n before to prevent clang-format reordering.
 #include <synchapi.h>
 #else
-#include <poll.h>
 #include <unistd.h>
+#include <poll.h>
 #endif
 
 #include <thread>
--- a/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
@@ -366,7 +366,7 @@ SELECT
   process.name AS process_name,
   slice.name AS cuj_slice_name,
   -- Extracts "CUJ_NAME" from "L<CUJ_NAME>"
-  substr(slice.name, 3, length(slice.name) - 3) AS cuj_name,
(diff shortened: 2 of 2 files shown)


LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -362,51 +361,24 @@ CREATE PERFETTO TABLE android_sysui_latency_cujs (
...
-      WHERE
-        csm.cuj_id = cuj.cuj_id AND csm.marker_type = 'end'
-    )
-    THEN 'completed'
-    ELSE NULL
-  END AS state
-FROM _sysui_cujs_slices AS cuj
-JOIN _sysui_cuj_instant_events AS cuj_events
-  USING (cuj_id)
+  substr(slice.name, 3, length(slice.name) - 3) AS cuj_name,
```
nicomazz: why not using the macro as before?

src/base/lock_free_task_runner.cc
```diff
@@ -24,8 +24,8 @@
 // Keep the \n before to prevent clang-format reordering.
 #include <synchapi.h>
 #else
-#include <unistd.h>
 #include <poll.h>
+#include <unistd.h>
```
nicomazz, line 27: is this change expected?

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
devianb: Fixed.

src/base/lock_free_task_runner.cc
devianb: Removed.

PR #3121, opened 2025-09-29
ui: update tabs widget functionality

LalitMaganti:
Add support for left/right icons and make some minor styling updates to support this.


ui/src/assets/widgets/tabs.scss
```diff
@@ -54,4 +54,22 @@
...
+    display: flex;
+    align-items: center;
+
+    &--left {
+      margin-right: 4px;
+    }
+
+    &--right {
+      margin-left: 4px;
+      cursor: pointer;
```
stevegolton: This is unnecessary, the containing tab already has cursor: pointer set

ui/src/assets/widgets/tabs.scss
```diff
@@ -54,4 +54,22 @@
...
+
+      &:hover {
+        opacity: 0.7;
+      }
```
stevegolton: Not sure about this UX, was it intentional?

ui/src/assets/widgets/tabs.scss
```diff
@@ -54,4 +54,22 @@
     margin: 0px 4px;
     overflow: hidden;
   }
+
+  &__tab-icon {
+    display: flex;
+    align-items: center;
```
stevegolton: These rules don't actually target anything as we don't have any elements called `pf-tabs__tab-icon`

ui/src/assets/widgets/tabs.scss
```diff
@@ -32,7 +32,7 @@
     white-space: nowrap;
     text-overflow: ellipsis;
     display: flex;
-    align-items: baseline;
+    align-items: center;
```
stevegolton: We generally want to keep align-items baseline to ensure that text in the tabs can be aligned with other text outside the tab strip. While this does pull the icons out of alignment with the text, you can align the icons better by applying align-self: center to the icon elements themselves.

It's a bit fiddly - I don't mind doing this in a followup CL.

PR #3122, opened 2025-09-29
pprof: fix proto annotations and update pprof handling tools

LalitMaganti:
Fix packed field annotations in pprof profile.proto for proper compatibility.
Update protoprofile tool with correct dependencies and sample handling.
Fix pprof builder sample value serialization to use add_value() instead of packed values.

### DIFF
--- a/protos/third_party/pprof/profile.proto
+++ b/protos/third_party/pprof/profile.proto
@@ -90,7 +90,7 @@ message Profile {
   int64 period = 12;
   // Freeform text associated to the profile.
   // Indices into string table.
-  repeated int64 comment = 13;
+  repeated int64 comment = 13 [packed = false];
   // Index into the string table of the type of the preferred sample
   // value. If unset, clients should default to the last sample value.
   int64 default_sample_type = 14;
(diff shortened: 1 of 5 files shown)


protos/third_party/pprof/profile.proto
```diff
@@ -111,14 +111,14 @@ message ValueType {
...
-  repeated uint64 location_id = 1;
+  repeated uint64 location_id = 1 [packed = true];
   // The type and unit of each value is defined by the corresponding
   // entry in Profile.sample_type. All samples must have the same
   // number of values, the same as the length of Profile.sample_type.
   // When aggregating multiple samples into a single sample, the
   // result has a list of values that is the elemntwise sum of the
   // lists of the originals.
-  repeated int64 value = 2;
+  repeated int64 value = 2 [packed = false];
```
sashwinbalaji, line 121: In https://github.com/google/pprof/blob/main/proto/profile.proto#L124, I don't see packed as false.

So curious abt what specific compatibility are we trying to address ?

sashwinbalaji (approved):

LalitMaganti:
Superseded by https://github.com/google/perfetto/pull/3126 which accomplishes the same thing but better.

PR #3124, opened 2025-09-29
ui: add new aggregates page plugin

LalitMaganti:
Add complete aggregates page plugin for displaying pprof aggregate profile data.
Register plugin in default plugins and provide full UI implementation.


ui/src/plugins/dev.perfetto.AggregatesPage/styles.scss
```diff
@@ -0,0 +1,80 @@
...
+  }
+}
+
+// Remove padding from popups containing explanation callouts
+.pf-popup:has(.pf-aggregates-page__page-explanation),
+.pf-popup:has(.pf-aggregates-page__view-explanation) {
+  .pf-popup-content {
+    padding: 0;
+  }
+}
```
stevegolton: These borders on these popups don't look quite right (see screenshot). What benefit does having the callout in there acheive?

Image

ui/src/plugins/dev.perfetto.AggregatesPage/index.ts
```diff
@@ -0,0 +1,252 @@
...
+export default class implements PerfettoPlugin {
+  static readonly id = 'dev.perfetto.AggregatesPage';
+
+  private state: AggregatesPageState = {
+    selectedScope: undefined,
+    availableScopes: [],
+  };
+  private readonly onStateUpdate = (update: AggregatesPageState): void => {
+    this.state = update;
+    m.redraw();
```
stevegolton: I feel like you shouldn't need any of the redraws on this page, but maybe something has gone wrong in one or more of the widgets?

ui/src/plugins/dev.perfetto.AggregatesPage/aggregates_page.ts
```diff
@@ -0,0 +1,398 @@
...
+        ),
+        m(
+          StackFixed,
+          m(
+            Popup,
+            {
+              trigger: m(Button, {
+                label: 'About page',
+                icon: 'help',
+                compact: true,
```
stevegolton: I feel like this button could be full size?

ui/src/plugins/dev.perfetto.AggregatesPage/aggregates_page.ts
```diff
@@ -0,0 +1,398 @@
...
+              onChange: (shouldOpen: boolean) => {
+                this.setExplanationVisible('page', shouldOpen);
+              },
+            },
+            m(
+              Callout,
+              {
+                icon: 'help',
+                dismissible: true,
+                onDismiss: () => this.setExplanationVisible('page', false),
```
stevegolton: The popup open/closed states seem to get out of sync when using the dismiss button for some reason. I.e. if you close the popup using the dismiss button, it takes two clicks to make the popup re-appear. This could well be down to a bug in Popup however.

ui/src/plugins/dev.perfetto.AggregatesPage/aggregates_page.ts
```diff
@@ -0,0 +1,398 @@
...
+        m(
+          StackAuto,
+          m(TabStrip, {
+            className: 'pf-aggregates-page__tabs',
+            tabs: [
+              {
+                key: 'flamegraph',
+                title: 'Flamegraph',
+                rightIcon: m(
+                  Popup,
```
stevegolton: Could this be a tooltip instead?

ui/src/plugins/dev.perfetto.AggregatesPage/aggregates_page.ts
stevegolton, line 1: In general I think the popups are a bit of an assault on the senses when first opening the page.

Also if you click anywhere they both disappear. For example, clicking on the dismiss button of one popup makes the other popup also disappear. I think you might need to control the popups entirely using button onclick callbacks to open/close the popup or you might be able to avoid this by setting the popup `closeOnOutsideClick` and `closeOnEscape` props to false.

stevegolton (approved):
Looks great! Left a few UX related comments, but this is pretty epic in general!

ui/src/plugins/dev.perfetto.PprofProfiles/index.ts
```diff
@@ -0,0 +1,158 @@
...
+      href: '#!/pprof',
+      icon: 'analytics',
+    });
+    trace.onTraceReady.addListener(async () => {
+      const hasAnyTracks = trace.workspaces.all[0].flatTracks.length > 0;
+      // TODO(lalitm): it's really bad that we're unconditionally navigating
+      // to the pprof page: really we should check if the user has not already
+      // set a page and then only navigate if no page is set. However:
+      //  a) no API exists for checking the current page
+      //  b) there is already some code in UI load time which navigates
```
stevegolton, line 58: This is technically not the case any more, but a) still stands. Not a problem for this PR just thought I'd mention it.

stevegolton (approved):
LGTM!

ui/src/plugins/dev.perfetto.PprofProfiles/index.ts
LalitMaganti: I think even after your CL, this is true. Here's why:
1) you go to ui.perfetto.dev and load trace
2) the code in load_trace runs and because you don't have a page, it sets it to viewer page
3) this code runs, see you already navigated to a page and doesn't do anything.

PR #3126, opened 2025-09-29
protozero: add iterator which can deal with both packed/unpacked fields

LalitMaganti:
This CL introduces a new wrapper class for cases where we need to
actually strictly adhere to the proto spec and deal with both packed and
non-packed fields in the same field. While no native Perfetto tools
writes protos like this, it is the case that we cannot control all the
protos we parse.

For example, pprofs have a few repeated fields which are inconsistent in
how their writers write them. To that end, make it possible to parse
both.


include/perfetto/protozero/proto_decoder.h
```diff
@@ -267,6 +272,61 @@ class PackedRepeatedFieldIterator {
...
+
+  // Constructor taking a packed iterator
+  explicit UnifiedRepeatedFieldIterator(PackedIterator iter)
+      : iter_(std::move(iter)) {}
+
+  // Constructs an invalid iterator.
+  UnifiedRepeatedFieldIterator() : iter_(RegularIterator()) {}
+
+  explicit operator bool() const {
+    if (std::holds_alternative<RegularIterator>(iter_)) {
```
primiano: note that you could avoid one branch here, as internally get_if already does a branch (because returns null if you hold it wrong),
so this could be like:

if (auto* r = std::get_if<RegularIterator>(&iter_))
  return static_cast<bool>(*r);
return static_cast<bool>(*std::get_if<PackedIterator>(&iter_));


same below everywhere

or technically you could even just std::visit to keep everything more compact

return std::visit([](auto& it){ return static_cast<bool>(it); }, iter_);

I checked with godbolt with a minified version and emits very miminmal code with only one branch: https://godbolt.org/z/5jzsarKPz

primiano (commented):
LGTM but i'd use std::visit tbh

PR #3130, opened 2025-09-29
ui: fix nullable durations in long battery tracing plugin

LalitMaganti:


ui/src/components/tracks/slice_track.ts
stevegolton, line 198: Why is this change in here? It was part of my other PR #3128

stevegolton, line 228: And this change? Whats going on?

PR #3135, opened 2025-09-29
tp: Add JSONWriter.

altimin:
Add JSON writer and tests in preparation for using it in arg printing logic (where we need tighter control over the order of arguments), as well as preparing to replace jsoncpp in the trace processor.


src/trace_processor/util/json_writer.cc
```diff
@@ -0,0 +1,187 @@
...
+ */
+
+#include "src/trace_processor/util/json_writer.h"
+
+#include <cmath>
+
+#include "perfetto/ext/base/dynamic_string_writer.h"
+#include "perfetto/ext/base/string_utils.h"
+
+namespace perfetto::trace_processor {
```
LalitMaganti: Needs to be namespaced under json::writer if you're going to have a top level "write" function.

src/trace_processor/util/json_writer.h
```diff
@@ -0,0 +1,238 @@
...
+
+// Main entry point for writing JSON.
+// Usage:
+//   std::string json = write([](JsonValueWriter writer) {
+//     std::move(writer).WriteDict([](JsonDictWriter dict) {
+//       dict.AddString("hello", "world");
+//     });
+//   });
+template <typename ValueWriterFn>
+std::string write(ValueWriterFn&& value_writer);
```
LalitMaganti: You need to namespace or rename this function.

src/trace_processor/util/json_writer.h
```diff
@@ -0,0 +1,238 @@
...
+
+  // Primitive values.
+  void AddNull(std::string_view key);
+  void AddBool(std::string_view key, bool value);
+  void AddInt(std::string_view key, int64_t value);
+  void AddUint(std::string_view key, uint64_t value);
+  void AddDouble(std::string_view key, double value);
+  void AddString(std::string_view key, std::string_view value);
+
+  // Writes a nested dictionary. `DictWriterFn` should be a function taking
```
LalitMaganti: I'm a bit concerned about exploding compile times and binary size with such aggressive use of templates. I would suggest maybe making this use std::function as it's rare that we need high-performance JSON writing? Maybe we can have *Optimized functions for those cases where do but most of the time I think the defaults should not need templates.

src/trace_processor/util/json_writer.cc
altimin: Ah, I knew I forgot something!

src/trace_processor/util/json_writer.h
altimin: Added a namespace.

src/trace_processor/util/json_writer.h
altimin: Sure, done — we can revisit this later (I'm interested in doing proper benchmarking at some point).

src/trace_processor/util/json_writer.h
```diff
@@ -0,0 +1,145 @@
...
+class JsonValueWriter;
+
+// Main entry point for writing JSON.
+// Usage:
+//   std::string json = write([](JsonValueWriter writer) {
+//     std::move(writer).WriteDict([](JsonDictWriter dict) {
+//       dict.AddString("hello", "world");
+//     });
+//   });
+std::string write(std::function<void(JsonValueWriter&&)> value_writer);
```
LalitMaganti: Needs to be capitalized.

PR #3146, opened 2025-10-01
ui: Hide a popup when its trigger is no longer visible

stevegolton:
Since popups are rendered out-of-tree, if a trigger becomes invisible due to it or one of its parents having display:none set, then the popup itself doesn't automatically become invisible.

This causes issues when e.g. changing page. Since the previous page is kept in the DOM just with `display:none`, a trigger element with an open popup on that page is not actually removed, just made invisible. However the popup still persists and is left floating awkwardly anchored to (0, 0).

This patch adds a popper.js modifier which detects if the trigger element is visible or not with `element.checkVisibility()`, applying `display:none` to the popup as well.

### DIFF
--- a/ui/src/widgets/popup.ts
+++ b/ui/src/widgets/popup.ts
@@ -284,6 +284,48 @@ export class Popup implements m.ClassComponent<PopupAttrs> {
       matchWidthModifier = [];
     }
 
+    // Custom modifier to hide popup when trigger is not visible. This can be
+    // due to the trigger or one of its ancestors having display:none.
+    const hideOnInvisible: Modifier<'hideOnInvisible', {}> = {
+      name: 'hideOnInvisible',
+      enabled: true,
+      phase: 'main',
+      fn({state}) {
+        const reference = state.elements.reference;
+        if (!(reference instanceof HTMLElement)) {
+          return;
+        }
+
+        // Check if checkVisibility is supported
+        if (typeof reference.checkVisibility === 'function') {
+          const isVisible = reference.checkVisibility();
+
+          if (!isVisible) {
+            // Hide the popper by setting display to none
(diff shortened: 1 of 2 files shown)


ui/src/widgets/popup.ts
```diff
@@ -284,6 +284,50 @@ export class Popup implements m.ClassComponent<PopupAttrs> {
...
 
+    // Custom modifier to hide popup when trigger is not visible. This can be
+    // due to the trigger or one of its ancestors having display:none.
+    const hideOnInvisible: Modifier<'hideOnInvisible', {}> = {
+      name: 'hideOnInvisible',
+      enabled: true,
+      phase: 'main',
+      fn({state}) {
+        const reference = state.elements.reference;
+        const foo = reference instanceof HTMLElement ? reference : undefined;
```
LalitMaganti: nit: name this something other than foo

LalitMaganti (approved):

ui/src/widgets/popup.ts
stevegolton: Doh! Done.

ui/src/widgets/popup.ts
```diff
@@ -284,6 +284,49 @@ export class Popup implements m.ClassComponent<PopupAttrs> {
...
+            rect.top >= 0 &&
+            rect.left >= 0 &&
+            rect.bottom <=
+              (window.innerHeight || document.documentElement.clientHeight) &&
+            rect.right <=
+              (window.innerWidth || document.documentElement.clientWidth) &&
+            window.getComputedStyle(reference).visibility !== 'hidden' &&
+            window.getComputedStyle(reference).display !== 'none';
+
+          console.log('isVisible', isVisible);
```
LalitMaganti: nit: remove?

PR #3156, opened 2025-10-01
Support WM in proto-to-args-with-defaults function.

PriyankaSPX:
Bug: 438681230

Test: tools/ninja -C out/linux_clang_debug perfetto_unittests && out/linux_clang_debug/perfetto_unittests --gtest_filter=WindowManager*

Test: tools/diff_test_trace_processor.py out/linux_clang_debug/trace_processor_shell --name-filter="PerfettoTable:winscope|WindowManager"


src/trace_processor/tables/winscope_tables.py
```diff
@@ -768,7 +768,7 @@
         ),
         C(
             'base64_proto_id',
-            CppUint32(),
+            CppOptional(CppUint32()),
```
rukkal, line 796: Why optional?

src/trace_processor/importers/proto/winscope/windowmanager_parser.cc
```diff
@@ -64,11 +67,12 @@ tables::WindowManagerTable::Id WindowManagerParser::InsertSnapshotRow(
...
   auto inserter = tracker.AddArgsTo(row_id);
   ArgsParser writer(timestamp, inserter, *trace_processor_context->storage);
-  base::Status status =
-      args_parser_.ParseMessage(blob,
-                                *util::winscope_proto_mapping::GetProtoName(
-                                    tables::WindowManagerTable::Name()),
-                                nullptr /* parse all fields */, writer);
+  auto table_name = tables::WindowManagerTable::Name();
+  auto allowed_fields =
+      util::winscope_proto_mapping::GetAllowedFields(table_name);
```
rukkal: Why is this kind of stuff (allowed fields, proto names) factored out in winscope_proto_mapping? To centralize access for the "proto to args with default" parser?

src/trace_processor/importers/proto/winscope/windowmanager_parser.cc
```diff
@@ -34,10 +35,12 @@ WindowManagerParser::WindowManagerParser(WinscopeContext* context)
       args_parser_{*context->trace_processor_context_->descriptor_pool_} {}
 
 void WindowManagerParser::Parse(int64_t timestamp, protozero::ConstBytes blob) {
-  auto snapshot_id = InsertSnapshotRow(timestamp, blob);
+  protos::pbzero::WindowManagerTraceEntry::Decoder entry_decoder(blob);
+  auto service = entry_decoder.window_manager_service();
```
rukkal: Why are we dropping the enclosing WindowContainerRoot message? I assume we are only using the `service` field atm, but doesn't WindowContainerRoot contain anything that might become useful in the future?

src/trace_processor/util/winscope_proto_mapping.h
```diff
@@ -79,11 +114,61 @@ inline base::StatusOr<const char* const> GetProtoName(
...
+  }
+  auto maybe_container_type =
+      GetWindowContainerType(static_table, row, string_pool);
+  if (!maybe_container_type.has_value()) {
+    return std::nullopt;
+  }
+
+  // proto message is WindowContainerChildProto
+  const auto container_type = maybe_container_type.value();
+  if (!std::strcmp(container_type, WINDOW_CONTAINER)) {
```
rukkal: Let's turn this into more idiomatic C++ replacing strcmp with the == operator

just turn the string literals above into std::string_view:
```
constexpr std::string_view ROOT_WINDOW_CONTAINER = "RootWindowContainer";
```
and I think the `==` will work

src/trace_processor/util/winscope_proto_mapping.h
```diff
@@ -79,11 +114,61 @@ inline base::StatusOr<const char* const> GetProtoName(
...
+    return std::vector<uint32_t>({2});
+  }
+  if (!std::strcmp(container_type, DISPLAY_CONTENT)) {
+    auto content = std::vector<uint32_t>({3});
+    return content;
+  }
+  if (!std::strcmp(container_type, DISPLAY_AREA)) {
+    return std::vector<uint32_t>({4});
```
rukkal: nit: just `return {4};` doesn't work?

src/trace_processor/util/winscope_proto_mapping.h
```diff
@@ -79,11 +114,61 @@ inline base::StatusOr<const char* const> GetProtoName(
...
+    return std::nullopt;
+  }
+
+  // proto message is WindowContainerChildProto
+  const auto container_type = maybe_container_type.value();
+  if (!std::strcmp(container_type, WINDOW_CONTAINER)) {
+    return std::vector<uint32_t>({2});
+  }
+  if (!std::strcmp(container_type, DISPLAY_CONTENT)) {
+    auto content = std::vector<uint32_t>({3});
```
rukkal: What about adding comments with the field names?
```
auto content = std::vector<uint32_t>({3}); // field_xyz
```

rukkal (approved):

src/trace_processor/tables/winscope_tables.py
PriyankaSPX: required to be optional for access post finalization

src/trace_processor/importers/proto/winscope/windowmanager_parser.cc
PriyankaSPX: yes exactly, for access in the "proto to args with defaults" function under table functions

src/trace_processor/importers/proto/winscope/windowmanager_parser.cc
PriyankaSPX: the root contains "where" and "elapsedRealtimeNanos" - neither of which is actually used anywhere in winscope or flicker atm

the reason for dropping is that when we build the table with defaults, we can only specify allowed fields for the root message. I don't want to retrieve the args+defaults for all the fields under windowManagerService (since they're already going to be built at the container level), so was planning to only provide the args with root message WindowManagerServiceDumpProto so i can block out the unwanted fields

but perhaps it would be better to prune the root entry itself at this stage and then only add those args.... will play around with this

src/trace_processor/util/winscope_proto_mapping.h
PriyankaSPX: no, it throws `error: chosen constructor is explicit in copy-initialization`

src/trace_processor/util/winscope_proto_mapping.h
PriyankaSPX: done

src/trace_processor/util/winscope_proto_mapping.h
PriyankaSPX: done

src/trace_processor/importers/proto/winscope/windowmanager_parser.cc
PriyankaSPX: changed to prune the root entry

src/trace_processor/perfetto_sql/intrinsics/table_functions/winscope_proto_to_args_with_defaults.cc
```diff
@@ -330,15 +348,18 @@ bool WinscopeProtoToArgsWithDefaults::Cursor::Run(
...
       GetProtoToInternedData(table_name_str, context_->storage.get());
-  base::Status insert_status = InsertRows(
-      *static_table_from_engine, &table_, *proto_name,
-      allowed_fields ? &allowed_fields.value() : nullptr,
-      group_id_col_name ? &group_id_col_name.value() : nullptr,
-      *context_->descriptor_pool_, string_pool_, proto_to_interned_data);
+
+  base::Status insert_status =
+      InsertRows(*static_table_from_engine, &table_, *proto_name,
+                 allowed_fields ? &allowed_fields.value() : nullptr,
```
rukkal, line 359: nit: if it doesn't make things slower, it'd be cleaner to pass a strategy like `const std::function<const std::vector<uint8_t>*(row_id, ...)>& get_allowed_fields`

thus we avoid that logic in InsertRows and it remains WM-agnostic

PriyankaSPX: thanks! will follow up

src/trace_processor/util/winscope_proto_mapping.h
```diff
@@ -31,6 +31,37 @@
...
+constexpr std::string_view DISPLAY_CONTENT = "DisplayContent";
+constexpr std::string_view DISPLAY_AREA = "DisplayArea";
+constexpr std::string_view TASK = "Task";
+constexpr std::string_view TASK_FRAGMENT = "TaskFragment";
+constexpr std::string_view ACTIVITY = "Activity";
+constexpr std::string_view WINDOW_TOKEN = "WindowToken";
+constexpr std::string_view WINDOW_STATE = "WindowState";
+constexpr std::string_view WINDOW_CONTAINER = "WindowContainer";
+
+std::optional<const char*> inline GetWindowContainerType(
```
LalitMaganti: base::StringView

LalitMaganti (approved):

src/trace_processor/util/winscope_proto_mapping.h
PriyankaSPX: if i use std::StringView I don't think I declare these globally within the namespace?

LalitMaganti: Can you make it std::string_view instead then?

PriyankaSPX: ah sorry you meant the function return type? changed that to base::StringView, thanks!

src/trace_processor/util/winscope_proto_mapping.h
```diff
@@ -31,6 +31,37 @@
...
+    return std::nullopt;
+  }
+  auto container_type_id =
+      static_table
+          .GetCellUncheckedLegacy<dataframe::String, dataframe::DenseNull>(
+              col_idx.value(), row);
+  if (!container_type_id.has_value()) {
+    return std::nullopt;
+  }
+  return string_pool->Get(container_type_id.value()).data();
```
LalitMaganti: You probably don't want data here but to create string view directly.

PriyankaSPX: replaced to just return the result of StringPool#Get

PR #3162, opened 2025-10-02
ui: Limit filter chip size

stevegolton:
- Use a common component for filter chip bars and filter chips:
  - GridFilterBar
  - GridFilterChip
- Use in DataGrid and SqlTable for a common and consistent UX.
- Limit filter chip text to 300px (ellipsizing if larger)
- Add the label as a title (browser controlled popup on hover showing
  the full filter text).
- Wrap filters when they overflow (rather than scroll).

image

### DIFF
--- a/ui/src/assets/widgets/chip.scss
+++ b/ui/src/assets/widgets/chip.scss
@@ -20,6 +20,9 @@
     background: color-mix(in srgb, $base-color 15%, transparent);
   }
 
+  display: inline-flex;
+  align-items: baseline;
+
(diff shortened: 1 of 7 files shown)


altimin:
> Wrap filters when they overflow (rather than scroll).

Just got annoyed by this yesterday — thanks for a very prompt fix :).

aMayzner (approved):

LalitMaganti (commented):
Seems like the fix is gone: I only see the test changes?

PR #3165, opened 2025-10-02
More cppcheck performance fixes

tobbi:
This PR fixes many of the performance issues reported by cppcheck.

I have one question:
cppcheck mentions that using std::move() when returning by value prevents some compiler optimizations (copy elision).

Yet, the warning gets specifically mentioned as a code comment in
https://github.com/google/perfetto/blob/main/src/profiling/common/producer_support.cc#L48

Is there a reason for using `std::move` there?

To be fair, I saw `std::move` a couple times throughout the code I was fixing. I fixed some of the instances in this PR.

### DIFF
--- a/src/perfetto_cmd/config.cc
+++ b/src/perfetto_cmd/config.cc
@@ -44,7 +44,7 @@ bool SplitValueAndUnit(const std::string& arg, ValueUnit* out) {
 }
 
 bool ConvertValue(const std::string& arg,
-                  std::vector<UnitMultipler> units,
+                  const std::vector<UnitMultipler>& units,
                   uint64_t* out) {
   if (arg.empty()) {
     *out = 0;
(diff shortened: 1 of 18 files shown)


LalitMaganti:
> Is there a reason for using std::move there?

I think in C++11, copy elision was not guaranteed (we used to be C++11 until ~2022) in all contexts it is in C++17 so there might have been a situation where the compiler was not properly eliding so we tried to force it.

Shouldn't be an issue anymore.

sashwinbalaji (approved):
> Yet, the warning gets specifically mentioned as a code comment in
https://github.com/google/perfetto/blob/main/src/profiling/common/producer_support.cc#L48

my guess regarding it is just that when change was added we were at older version of cpp causing error msg (https://android-review.git.corp.google.com/c/platform/external/perfetto/+/2062290/comment/ac21b232_9b05754c/), maybe now it's safer to update.

Would let the original code author @rsavitski to comment more

tobbi:
<s>Can you please approve again? (Unfortunately, I keep getting build errors locally not caused by this PR).</s>

Never mind. I thought approval is required for CI runs.

tobbi:
Not long before I have to pay Google for excess electricity caused by CI builds.

LalitMaganti:
Thanks for the contribution!

PR #3167, opened 2025-10-02
trace_processor: support f2fs_write_checkpoint

yhli-tw:
Add f2fs_write_checkpoint ftrace to slice for showing it on ui.

Bug: 442974909

Welcome to Perfetto!

For more details please see


src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -523,7 +550,10 @@ FtraceParser::FtraceParser(TraceProcessorContext* context,
       disp_vblank_irq_enable_output_id_arg_name_(
           context_->storage->InternString("output_id")),
       hrtimer_id_(context_->storage->InternString("hrtimer")),
-      local_timer_id_(context_->storage->InternString("IRQ (LocalTimer)")) {
+      local_timer_id_(context_->storage->InternString("IRQ (LocalTimer)")),
+      f2fs_checkpoint_cat_id_(
```
LalitMaganti: Don't set a category for these events.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4383,32 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+
+  base::StringView msg(evt.msg());
+
+  int32_t reason = evt.reason();
+  std::string reason_str = GetF2fsCheckpointReasonString(reason);
+  StringId reason_name_id =
+      context_->storage->InternString(base::StringView(reason_str));
+
+  UniqueTid utid = context_->process_tracker->GetOrCreateThread(pid);
+  TrackId track_id = context_->track_tracker->InternThreadTrack(utid);
```
LalitMaganti: Are we sure this is going to be synchronous event (i.e. can the thread do anything else before you

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4383,32 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+
+  if (msg.find("start block_ops") != base::StringView::npos) {
```
LalitMaganti: I'm very very unhappy with the fact you are doing string matching here. Why is this not an enum which you can match on instead? Surely since you control the event you can change this?

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4383,32 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+void FtraceParser::ParseF2fsWriteCheckpoint(int64_t ts,
+                                            uint32_t pid,
+                                            ConstBytes blob) {
+  protos::pbzero::F2fsWriteCheckpointFtraceEvent::Decoder evt(blob);
+
+  base::StringView msg(evt.msg());
+
+  int32_t reason = evt.reason();
+  std::string reason_str = GetF2fsCheckpointReasonString(reason);
+  StringId reason_name_id =
```
LalitMaganti: Making this the name instead of a constant is a terrible idea: it will make matching on these events from SQL impossible, it will be super mysterious in the UI: you need to think about all the other teams other than your own who will be looking at this event and have no clue what this means.

I would make the event name a constant and put an all the reasons as an array in args.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -262,6 +263,32 @@ std::string GetUfsCmdString(uint32_t ufsopcode, uint32_t gid) {
   return buffer;
 }
 
+std::string GetF2fsCheckpointReasonString(int32_t reason) {
```
LalitMaganti: This is going to have very bad performance and given this is quite a frequent event doing so many allocations is not acceptable.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -262,6 +263,32 @@ std::string GetUfsCmdString(uint32_t ufsopcode, uint32_t gid) {
...
+  if (reason == 0) {
+    return "Unknown";
+  }
+
+  std::vector<std::string> flags;
+  if (reason & 0x00000001)
```
LalitMaganti: All of these strings should be interned in the constructor and passed as an array of arguments to the slice.

yhli-tw: Thank you for your advice. I just submitted a new patch for it.

src/trace_processor/importers/ftrace/ftrace_parser.h
```diff
@@ -442,6 +445,12 @@ class FtraceParser {
   const StringId disp_vblank_irq_enable_output_id_arg_name_;
   const StringId hrtimer_id_;
   const StringId local_timer_id_;
+  const StringId f2fs_checkpoint_name_id_;
+  const StringId f2fs_reason_str_arg_id_;
+  const StringId f2fs_reason_int_arg_id_;
+  const StringId f2fs_dev_arg_id_;
+  const StringId f2fs_checkpoint_unknown_reason_id_;
+  std::vector<StringId> f2fs_checkpoint_reason_ids_;
```
LalitMaganti: This can be an std::array instead of std::vector.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4372,59 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+
+}  // namespace
+
+void FtraceParser::ParseF2fsWriteCheckpoint(int64_t ts,
+                                            uint32_t pid,
+                                            ConstBytes blob) {
+  protos::pbzero::F2fsWriteCheckpointFtraceEvent::Decoder evt(blob);
+
+  uint32_t phase = evt.phase();
+  if (phase == 1) {
```
LalitMaganti: Please add constants for 0 and 1 idnicating what they mean.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4372,59 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+    return;
```
LalitMaganti, line 4416: Why is 1 just ignored? Shouldn't it cause an end to happen?

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4372,59 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+
+  TrackId track_id = context_->track_tracker->InternTrack(
+      kF2fsCheckpointBlueprint, tracks::Dimensions(pid));
+
+  if (phase == 0) {
+    int32_t reason_int = evt.reason();
+    uint64_t dev = evt.dev();
+
+    // End the slice first to prevent any open slice existing.
+    context_->slice_tracker->End(ts, track_id);
```
LalitMaganti, line 4431: Is this necessary anymore? Shouldn't we expect a 0 to be always matched to a 2?

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4372,59 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+          inserter->AddArg(f2fs_dev_arg_id_, Variadic::UnsignedInteger(dev));
+          inserter->AddArg(f2fs_reason_int_arg_id_,
+                           Variadic::Integer(reason_int));
+          if (reason_int == 0) {
+            inserter->AddArg(
+                f2fs_reason_str_arg_id_,
+                Variadic::String(f2fs_checkpoint_unknown_reason_id_));
+          } else {
+            for (size_t i = 0; i < f2fs_checkpoint_reason_ids_.size(); ++i) {
+              if (reason_int & (1 << i)) {
```
LalitMaganti, line 4445: Is it possible to have multiple flags set here? If so you will need to add them as different keys (see [1] for example)

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/util/json_utils.cc;l=169?q=src%2Ftrace_processor%2Futil%2Fjson_utils.cc&ss=chromium

yhli-tw: Yes, it is. However, I thought that one key for multiple flag (e.g.reason_str: "Umount", "Sync") is as my expectation. I don't need different keys here. Is it acceptable?

src/trace_processor/importers/ftrace/ftrace_parser.h
yhli-tw: Thanks! Done.

src/trace_processor/importers/ftrace/ftrace_parser.cc
yhli-tw: Please reference the source code of the trace. https://source.corp.google.com/piper///depot/google3/third_party/appliance_linux/packages/linux/linux_6_1/fs/f2fs/checkpoint.c;l=1662?q=trace_f2fs_write_checkpoint&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental
The "1" state means that the block_operations() function is successfully completed. It's a middle event. However, the block_operations() is related with I/O transfer by UFS HW component. It might have chance to be failed.

src/trace_processor/importers/ftrace/ftrace_parser.cc
yhli-tw: Please reference the source code link I provided, it has chance that 0 doesn't match 2. So, I prefer to keep it to prevent from any open slice. What do you think?

src/trace_processor/importers/ftrace/ftrace_parser.cc
yhli-tw: Thanks! Done.

src/trace_processor/importers/ftrace/ftrace_parser.cc
LalitMaganti: It's required in the trace processor data model that for a single set of args, only one arg with a given key exists. If you need mutliple args for the same key, you need to use the array functions like I menttioned

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4371,64 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
       EventTracker::DmabufRssStat(), ts, static_cast<double>(evt.rss()), utid);
 }
 
+namespace {
+constexpr auto kF2fsCheckpointBlueprint = tracks::SliceBlueprint(
```
LalitMaganti: These can be inlined into the function itself

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4371,62 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+                                            uint32_t pid,
+                                            ConstBytes blob) {
+  enum F2fsCheckpointPhase {
+    kF2fsCpPhaseStart = 0,
+    kF2fsCpPhaseFinishBlockOps = 1,
+    kF2fsCpPhaseFinish = 2,
+  };
+  constexpr auto kF2fsCheckpointBlueprint = tracks::SliceBlueprint(
+      "f2fs_write_checkpoint",
+      tracks::DimensionBlueprints(tracks::UintDimensionBlueprint("pid")),
```
LalitMaganti: This needs to be `tracks::kProcessDimensionBlueprint` not interning pid like this.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4371,62 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+
+  protos::pbzero::F2fsWriteCheckpointFtraceEvent::Decoder evt(blob);
+
+  uint32_t phase = evt.phase();
+  if (phase == kF2fsCpPhaseFinishBlockOps) {
+    return;
+  }
+
+  TrackId track_id = context_->track_tracker->InternTrack(
+      kF2fsCheckpointBlueprint, tracks::Dimensions(pid));
```
LalitMaganti: You need to conver this to a upid with process tracker and intern that. pid is not suitable as a key, it's not globally unique.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -4349,4 +4371,62 @@ void FtraceParser::ParseDmabufRssStat(int64_t ts,
...
+  enum F2fsCheckpointPhase {
+    kF2fsCpPhaseStart = 0,
+    kF2fsCpPhaseFinishBlockOps = 1,
+    kF2fsCpPhaseFinish = 2,
+  };
+  constexpr auto kF2fsCheckpointBlueprint = tracks::SliceBlueprint(
+      "f2fs_write_checkpoint",
+      tracks::DimensionBlueprints(tracks::UintDimensionBlueprint("pid")),
+      tracks::FnNameBlueprint([](uint32_t pid) {
+        return base::StackString<255>("f2fs_ckpt-%u", pid);
```
LalitMaganti: if you need pid in the name, you will need to use dynamic names not FnName.

src/trace_processor/importers/ftrace/ftrace_parser.cc
yhli-tw: Done. Thank you!

src/trace_processor/importers/ftrace/ftrace_parser.cc
yhli-tw: I find that a trace might have multiple f2fs checkpoint kernel threads like the following : https://screenshot.googleplex.com/4u6epUtEXo4Wp6B.
Therefore, I need to add the pid or upid to the track name for distinguishing the different thread.  
Currently, this commit could show the following result for now: https://screenshot.googleplex.com/9bGDpNmQVmbefbs.

LalitMaganti: Sure but as I said you should use dynamic name instead of fnanme. You should not be using pid as a dimension as it's not stable across time in traces.

LalitMaganti (changes requested):
Please see comment. Track name needs to be fixed.

LalitMaganti (approved):

PR #3168, opened 2025-10-02
ui: support TS type definitions for embedders

cdamus:
An application that wishes to embed the Perfetto UI in its own UI may need to invoke various public (and non-public) APIs within Perfetto UI to achieve these ends. And when that embedding application is implemented in TypeScript, it needs TS definitions for those APIs.

This commit ensures consistent generation of TS type definition files, especially correcting missing exports of types that are exposed by other exported signatures.

### DIFF
--- a/ui/tsconfig.base.json
+++ b/ui/tsconfig.base.json
@@ -6,7 +6,7 @@
     "moduleResolution": "node",
     // Lints and checks.
     "allowJs": true,
-    "declaration": false,                  // Generates corresponding '.d.ts' file.
+    "declaration": true,                   // Generates corresponding '.d.ts' file.
     "sourceMap": true,                     // Generates corresponding '.map' file.
     "inlineSources": true,                 // Adds source code to the '.map' file.
     "removeComments": false,               // Do not emit comments to output.
--- a/ui/tsconfig.json
+++ b/ui/tsconfig.json
@@ -20,5 +20,8 @@
     },
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
+    "typeRoots": [
+      "./node_modules/@types"
+    ],
   }
 }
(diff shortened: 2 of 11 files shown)


ui/tsconfig.base.json
```diff
@@ -6,7 +6,7 @@
     "moduleResolution": "node",
     // Lints and checks.
     "allowJs": true,
-    "declaration": false,                  // Generates corresponding '.d.ts' file.
+    "declaration": true,                   // Generates corresponding '.d.ts' file.
```
stevegolton, line 9: Does this have any noticeable impact on the regular build? We could go down the path of swapping out tsconfig files  for different builds, but only if it actually makes a difference.

ui/tsconfig.json
```diff
@@ -20,5 +20,8 @@
     },
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
+    "typeRoots": [
```
stevegolton, line 23: What's the purpose of this? Isn't `node_modules/@types` visible by default?

ui/tsconfig.base.json
cdamus: The only impact is that the `*.d.ts` files are emitted into the `out/ui/ui/tsc/` tree, but of course that's the point.

The distribution JS bundles produced by `rollup` are identical. Which makes sense because they wouldn't be expected to include anything TS-specific like the type definitions.

ui/tsconfig.json
cdamus: The purpose of this is to ensure that when Perfetto UI is incorporated into an embedding application's git repository as a submodule, a local build of Perfetto UI (via `build.js`) doesn't see type definitions in the enclosing project's `node_modules` which, by the structure of git submodules, is in some containing directory. This isolates the Perfetto UI build from potential interference with a parent git module.

Perhaps this should be factored out into a distinct tsconfig for embedders? I agree that it's odd to put something in here that assumes this kind of enclosing project structure.

PR #3169, opened 2025-10-02
ui: return the help modal promise

cdamus:
Applications embedding the Perfetto UI in their own UI, when they show the Help modal via their own triggers, need to know when it is closed.


ui/src/frontend/help_modal.ts
```diff
@@ -28,7 +28,7 @@ import {raf} from '../core/raf_scheduler';
 
 export function toggleHelp() {
```
stevegolton, line 29: This seems like a misnomer, as it doesn't toggle the help at all it just seems to show it, so `showHelp` would probably make more sense. This is how it behaved before, the return of a promise just drives it home. Not going to block on this though.

PR #3170, opened 2025-10-02
ui: return a trace promise from the openTraceFromXyz() APIs

cdamus:
Return promises of the future trace object to callers of the APIs on the App that open traces so that they may act upon the opened trace.

Also, an application that embeds the Perfetto UI may need to hand over a stream of trace data obtained in its own way from a non-URL source and not gathered into a buffer.

### DIFF
--- a/ui/src/core/app_impl.ts
+++ b/ui/src/core/app_impl.ts
@@ -24,6 +24,7 @@ import {PageHandler} from '../public/page';
 import {Raf} from '../public/raf';
 import {RouteArg, RouteArgs} from '../public/route_schema';
 import {Setting, SettingDescriptor, SettingsManager} from '../public/settings';
+import {TraceStream} from '../public/stream';
 import {DurationPrecision, TimestampFormat} from '../public/timeline';
 import {NewEngineMode} from '../trace_processor/engine';
 import {AnalyticsInternal, initAnalytics} from './analytics_impl';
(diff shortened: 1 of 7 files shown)


cdamus:
> Looks good, but do we not also need the corresponding changes in `load_trace.ts` to handle the TRACE_STREAM mode?

Yes. Yes, we do.

cdamus:
Thanks @stevegolton. Commit 0403d6d87e0d6fa2fbfee759d2db358ff406d26c handles the stream source type in loading.

A unit test would have caught this, so I've added some test coverage.

ui/src/core/load_trace_unittest.ts
stevegolton, line 1: This is great and I agree we're really lacking unit testing at this level, but really It's such a large mock I worry about the maintenance burden going forward. Perhaps can you omit this test or put it in another PR so that we can just get the bulk of this PR landed?

cdamus: Dome with commit 6eaf75a6f79ee10d9f3b7f0c3ab1bc952a0687a3. Thanks!

stevegolton (approved):

cdamus:
Oh my, it looks like I missed a detail. Disallowed import:  /public/app -> /core/trace_stream .

I shall have a look what I can do about that.

cdamus:
> Oh my, it looks like I missed a detail. Disallowed import: /public/app -> /core/trace_stream .

Commit 1dee58546741283ce32ea3bae4a56fa8eb766700 addresses this by moving the `TraceStream` type to the public API. It does feel like it should be API; I can imagine that it might be practical, for example, to provide an adapter for ReadableStreams. As a future enhancement 😀  Happy to take alternative suggestions, of course.

stevegolton:
> > Oh my, it looks like I missed a detail. Disallowed import: /public/app -> /core/trace_stream .
> 
> Commit [1dee585](https://github.com/google/perfetto/commit/1dee58546741283ce32ea3bae4a56fa8eb766700) addresses this by moving the `TraceStream` type to the public API. It does feel like it should be API; I can imagine that it might be practical, for example, to provide an adapter for ReadableStreams. As a future enhancement 😀 Happy to take alternative suggestions, of course.

I'd say that makes sense. I've kicked off the CI to check everything.

PR #3174, opened 2025-10-02
tools: allow setting origin/main as a valid parent branch

LalitMaganti:


tools/git_sync_stack.py
```diff
@@ -78,7 +79,6 @@ def main():
 
   default_remote_name = args.remote
   repo_default_branch = 'origin/main'
```
sashwinbalaji, line 81: this can also be updated ? And also https://crsrc.org/c/third_party/perfetto/tools/git_new_branch.py;l=39?q=%22origin%2Fmain%22&ss=chromium%2Fchromium%2Fsrc:third_party%2Fperfetto%2Ftools%2F

LalitMaganti: The concepts are slightly different: this is the branch we are forking for, the other is the branch we want as our upstream. Happens to be origin/main usually but doesn't have to be.

PR #3177, opened 2025-10-02
ui: Adjustable grid column widths

stevegolton:
WIP/DNS: Rebase Grid widget on flex containers rather than <table> to allow more control + add rudimentary column resizing support.


stevegolton:
Deprecated in favor of #3298

PR #3183, opened 2025-10-03
traced_probes: Enable reading GPU frequency from the Intel xe driver.

sammc:
The xe driver exposes GPU clock frequencies under a different path than the i915 driver. Further, it only exposes them at a per-tile/gt granularity, without a single global GPU frequency. Scan for all tiles and GTs reporting frequencies. Generate the potential paths manually rather than performing a directory scan to reduce the permissions required.

Also, refactor the GPU frequency polling to open the files once instead of for each poll to avoid unnecessary work and chatty error logging for each absent potential GPU frequency file.

Bug: 449028300

### DIFF
--- a/src/traced/probes/sys_stats/sys_stats_data_source.cc
+++ b/src/traced/probes/sys_stats/sys_stats_data_source.cc
@@ -190,6 +190,10 @@ SysStatsDataSource::SysStatsDataSource(
   thermal_ticks_ = ticks[8];
   cpuidle_ticks_ = ticks[9];
   gpufreq_ticks_ = ticks[10];
+
+  if (gpufreq_ticks_) {
+    OpenGpuFreqFiles(open_fn);
+  }
 }
 
 void SysStatsDataSource::Start() {
@@ -289,7 +293,13 @@ std::optional<uint64_t> SysStatsDataSource::ReadFileToUInt64(
   if (!fd) {
     return std::nullopt;
   }
-  size_t rsize = ReadFile(&fd, path.c_str());
+  return ReadFileToUInt64(&fd, path.c_str());
+}
+
+std::optional<uint64_t> SysStatsDataSource::ReadFileToUInt64(
+    base::ScopedFile* fd,
+    const char* path) {
+  size_t rsize = ReadFile(fd, path);
   if (!rsize)
     return std::nullopt;
 
@@ -381,13 +391,18 @@ void SysStatsDataSource::ReadCpuIdleStates(
   }
 }
 
-std::optional<uint64_t> SysStatsDataSource::ReadAMDGpuFreq() {
-  std::optional<std::string> amd_gpu_freq =
-      ReadFileToString("/sys/class/drm/card0/device/pp_dpm_sclk");
-  if (!amd_gpu_freq) {
+std::optional<uint64_t> SysStatsDataSource::ReadAmdGpuFreq() {
+  if (!amd_gpufreq_fd_) {
     return std::nullopt;
   }
-  for (base::StringSplitter lines(*amd_gpu_freq, '\n'); lines.Next();) {
+  size_t rsize =
+      ReadFile(&amd_gpufreq_fd_, "/sys/class/drm/card0/device/pp_dpm_sclk");
+  if (!rsize) {
+    return std::nullopt;
+  }
+
+  char* buf = static_cast<char*>(read_buf_.Get());
+  for (base::StringSplitter lines(buf, '\n'); lines.Next();) {
     base::StringView line(lines.cur_token(), lines.cur_token_size());
     // Current frequency indicated with asterisk.
     if (line.EndsWith("*")) {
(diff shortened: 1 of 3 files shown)


LalitMaganti:
Kirill/Ryan: can you please review this?

primiano:
I'll take this (later today)

src/traced/probes/sys_stats/sys_stats_data_source.h
```diff
@@ -103,9 +103,12 @@ class SysStatsDataSource : public ProbesDataSource {
   void ReadThermalZones(protos::pbzero::SysStats* sys_stats);
   void ReadCpuIdleStates(protos::pbzero::SysStats* sys_stats);
   void ReadGpuFrequency(protos::pbzero::SysStats* sys_stats);
-  std::optional<uint64_t> ReadAMDGpuFreq();
+  std::optional<uint64_t> ReadAmdGpuFreq();
+
+  void OpenGpuFreqFiles(OpenFunction open_fn);
 
   size_t ReadFile(base::ScopedFile*, const char* path);
+  std::optional<uint64_t> ReadFileToUInt64(base::ScopedFile*, const char* path);
```
primiano, line 111: can you expand a bit on what's the rationale here?
we have a virtual method above ReadFileToUInt64(const std::string& path) but you are adding a non-virtual here. who is supposed to use what? when?

I think this is a bit confusing as I can see people using one or the other, and perhaps messing up tests. what's the logic here?

Also if this is really a private impl detail it should go into the anonymous namespace in the .cc file instead?

src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -406,25 +421,49 @@ std::optional<uint64_t> SysStatsDataSource::ReadAMDGpuFreq() {
...
-  freq = ReadAMDGpuFreq();
-  if (freq) {
-    sys_stats->add_gpufreq_mhz(*freq);
+void SysStatsDataSource::OpenGpuFreqFiles(OpenFunction open_fn) {
+  for (size_t tile = 0;; ++tile) {
+    bool found = false;
+    for (size_t gt = 0;; ++gt) {
+      base::StackString<256> freq_path(
+          "/sys/class/drm/card0/device/tile%zu/gt%zu/freq0/act_freq", tile, gt);
+      auto fd = open_fn(freq_path.c_str());
```
primiano: can you make this explicitly a base::ScopedFile rather than auto?
There are too many layers here, so it's impossible to tell when reading the code if this is an int (in which case the if(fd) below would be wrong) or a ScopedFile (in which case it's fine)

src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -406,25 +421,49 @@ std::optional<uint64_t> SysStatsDataSource::ReadAMDGpuFreq() {
...
+      } else {
+        break;
+      }
+    }
+    if (!found) {
+      break;
+    }
+  }
+
+  auto fd = open_fn("/sys/class/drm/card0/gt_act_freq_mhz");
```
primiano: ditto (ScopedFile)
This could be just

```
if (base::ScopedFile fd = openfn(...)) {
    intel_gpufreq_fds_.emplace_back(std::move(fd),
                                    "/sys/class/drm/card0/gt_act_freq_mhz")
}
```

primiano: also wrap "/sys/class/drm/card0/gt_act_freq_mhz" in a static local const, so you avoid having it twice here anf below?

src/traced/probes/sys_stats/sys_stats_data_source.h
sammc: There's a split in the way files are read in this class: some open the file once during the constructor and use ReadFile() to read into a reusable buffer stored as a member variable before performing per-file parsing; others open the file each time - these use ReadFileToString() and ReadFileToUInt64(), possibly performing specific parsing when reading a string. The former stub out the initial open, while the latter stub out the ReadFileTo*().

With this change, we end up with a couple cases of open once where the parsing is just a single integer, so can benefit from just the reading and parsing part of ReadFileToUInt64(), but not the opening part, hence the overload to allow just that. It depends on ReadFile(), a member function, so cannot change into a free function.

It's certainly a bit clunky as it is, with this inconsistent mixture. We could certainly rename to be less ambiguous as far as purpose goes, but would still leave behind the lack of clarity of when to use each. Given there are only two callsites for the new ReadFileToUInt64(), we could just inline it for now.

src/traced/probes/sys_stats/sys_stats_data_source.cc
sammc: Done.

src/traced/probes/sys_stats/sys_stats_data_source.cc
sammc: Done.

src/traced/probes/sys_stats/sys_stats_data_source.h
primiano: For now could you just use the ReadFileToUInt64? I feel the whole caching of FDs is making the code a bit of a nightmare to maintain, at least in the current state.
also you are changing the semantic of some of the existing code (AMD, ARdeno) from opening the file every time, to cache the FD. 
Are you sure the sysfs file for those drivers is idempotent and allows caching of the FD?
Maybe for simplicity let's just stick with ReadFileToUInt64(path) even though it opens/closes the FD all the time, until somebody can write a better, more scalable solution, rather than keep adding methods here,

src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -406,25 +420,47 @@ std::optional<uint64_t> SysStatsDataSource::ReadAMDGpuFreq() {
 }
 
 void SysStatsDataSource::ReadGpuFrequency(protos::pbzero::SysStats* sys_stats) {
-  // For Adreno GPUs.
-  auto freq = ReadFileToUInt64("/sys/class/kgsl/kgsl-3d0/devfreq/cur_freq");
-  if (freq) {
+  if (adreno_gpufreq_fd_) {
+    auto freq = ReadFileToUInt64(&adreno_gpufreq_fd_,
+                                 "/sys/class/kgsl/kgsl-3d0/devfreq/cur_freq");
     sys_stats->add_gpufreq_mhz((*freq) / 1'000'000);
```
primiano, line 426: whyd did you ditch the if(freq)? this will crash (or have UB) if parsing fails

src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -406,25 +420,47 @@ std::optional<uint64_t> SysStatsDataSource::ReadAMDGpuFreq() {
...
+}
 
-  // For AMD GPUs.
-  freq = ReadAMDGpuFreq();
-  if (freq) {
-    sys_stats->add_gpufreq_mhz(*freq);
+void SysStatsDataSource::OpenGpuFreqFiles(OpenFunction open_fn) {
+  for (size_t tile = 0;; ++tile) {
+    bool found = false;
+    for (size_t gt = 0;; ++gt) {
```
primiano, line 443: maybe pick a different var name? the chances of some system header doing `#define gt ` are non 0 given all the random platforms we support these days (I have scars from people including <iso646.h>`  in another life)

src/traced/probes/sys_stats/sys_stats_data_source_unittest.cc
```diff
@@ -590,75 +611,23 @@ TEST_F(SysStatsDataSourceTest, CpuIdleStates) {
...
-
   WaitTick(data_source.get());
 
   protos::gen::TracePacket packet = writer_raw_->GetOnlyTracePacket();
   ASSERT_TRUE(packet.has_sys_stats());
   const auto& sys_stats = packet.sys_stats();
-  EXPECT_EQ(sys_stats.gpufreq_mhz_size(), 1);
-  uint32_t adreno_gpufreq = 400;
-  EXPECT_EQ(sys_stats.gpufreq_mhz()[0], adreno_gpufreq);
+  ASSERT_EQ(sys_stats.gpufreq_mhz_size(), 4);
```
primiano, line 626: tbh this test was nicer before as it was trigger the more realistic case where only one cpu is present at any time.

Now this one pretends that you have all the GPUs all the time?

primiano (commented):
Sorry for the delay on this. I completely forgot about this PR until a colleague reminded me

Some comments below.

P.S. I try to respond within 24h to CLS/PRs, if I don't ping me on IM

PR #3185, opened 2025-10-03
ui: Add regex option for Android log tag filtering

### LINKED ISSUE #3072: ui: Improve log filtering flexibility
Hi Perfetto team,

I'd like to suggest a couple of enhancements to the Android log filtering panel to make it more user-friendly and powerful.
Currently, the filtering behavior is as follows:
1. Tag filtering: requires an exact match for the tag.
2. Message filtering: the search is case-sensitive.
I propose the following improvements:
1. Regex support for tag filtering: it would be very helpful to allow filtering tags using regular expressions. For example, a user might want to filter all tags with a certain prefix, like MyApp*. Exact matching is often too restrictive.
2. Case-Insensitive message filtering: making the message search case-insensitive by default would be more practical. It's common to search for log messages without remembering the exact casing, and case-sensitivity can cause users to miss relevant logs.
If you think these suggestions are sensible, I'd be happy to contribute a pull request to implement these features.

loordoza:
Adds a checkbox to allow users to switch between
basic string matching and regex pattern
matching for filtering Android log tags.

Fixes #3072

### DIFF
--- a/ui/src/plugins/com.android.AndroidLog/logs_panel.ts
+++ b/ui/src/plugins/com.android.AndroidLog/logs_panel.ts
@@ -23,6 +23,7 @@ import {AsyncLimiter} from '../../base/async_limiter';
 import {
   escapeQuery,
   escapeSearchQuery,
+  escapeRegexQuery,
 } from '../../trace_processor/query_utils';
 import {Select} from '../../widgets/select';
 import {
@@ -32,6 +33,7 @@ import {
 } from '../../widgets/multiselect';
 import {PopupPosition} from '../../widgets/popup';
 import {Button} from '../../widgets/button';
+import {Checkbox} from '../../widgets/checkbox';
 import {TextInput} from '../../widgets/text_input';
 import {VirtualTable, VirtualTableRow} from '../../widgets/virtual_table';
 import {classNames} from '../../base/classnames';
@@ -44,6 +46,7 @@ const ROW_H = 20;
 export interface LogFilteringCriteria {
   minimumLevel: number;
   tags: string[];
+  isTagRegex: boolean;
   textEntry: string;
   hideNonMatching: boolean;
   machineExcludeList: number[];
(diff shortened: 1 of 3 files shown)


LalitMaganti:
Thanks for the contribution. Tried it out and have some feedback:
1) Having the switch on the left side of the filter feels off. I would expect it to be on the right side
2) I think it would be better if it was a button + popup + list of options situation rather than taking so much space on the filter bar
3) If we're adding tag filtering, don't see why we shouldn't also do logs filtering at the same time.

ui/src/plugins/com.android.AndroidLog/logs_panel.ts
```diff
@@ -346,20 +349,32 @@ export class LogsFilters implements m.ClassComponent<LogsFiltersAttrs> {
...
-            draft.tags.push(tag);
-          });
-        },
-        onTagRemove: (index) => {
-          attrs.store.edit((draft) => {
-            draft.tags.splice(index, 1);
-          });
-        },
-      }),
+      m('div', {style: "display: flex; align-items: center; margin-bottom: 4px; gap: 8px;"},
```
stevegolton: Please don't use inline styles - put them instead with the rest of the styles in `logs_panel.scss`

ui/src/plugins/com.android.AndroidLog/logs_panel.ts
```diff
@@ -346,20 +349,32 @@ export class LogsFilters implements m.ClassComponent<LogsFiltersAttrs> {
...
+        m(Checkbox, {
```
stevegolton: Agree with @LalitMaganti that a checkbox takes up too much space here and that it should go on the right, but perhaps it should be a simple icon-only toggle button instead with [this icon](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:regular_expression:FILL@0;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23e8eaed). You can use the `Button` widget for this and set active to true when toggled on.

Or you could do the dropdown thing that Lalit mentioned, I'll leave it up to you.

ui/src/plugins/com.android.AndroidLog/logs_panel.ts
```diff
@@ -44,6 +46,7 @@ const ROW_H = 20;
 export interface LogFilteringCriteria {
   minimumLevel: number;
   tags: string[];
+  isTagRegex: boolean;
```
stevegolton: This might be missing if loading from a previous generation of state, so maybe make it optional with `?` and handle the fallout accordingly. I think it should just work because undefined is considered falsy.

Ultimately we need to replace this with zod like in the rest of the UI, but this is a job for another time.

stevegolton:
Hey @loordoza, just a gentle reminder - did you manage to take a look at our feedback? We would be quite keen on getting this feature landed :)

loordoza:
Hey, I totally forgot about this! I’ve just pushed the updates based on your feedback - please take a look :)

ui/src/plugins/com.android.AndroidLog/logs_panel.ts
```diff
@@ -381,6 +383,15 @@ export class LogsFilters implements m.ClassComponent<LogsFiltersAttrs> {
           });
         },
       }),
+      m(Button, {
```
stevegolton, line 386: Nit: A `title` here saying 'Use regular expression' would be absolutely ideal, but I'm not going to block on this.

stevegolton:
You're getting stuck on the screenshot tests. Unfortunately non-googlers cannot rebase line at the moment, so I'll do it for you and send you the patch.

stevegolton:
See the attached patch. You might need to rebase on main again as other things in the screenshot might also have changed since.

[fix-screenshots.patch](https://github.com/user-attachments/files/23531107/fix-screenshots.patch)

LalitMaganti:
Merging this as I think this can now land. Thanks for contribution!

PR #3188, opened 2025-10-06
Add pinning/copying of tracks by SQL.

peterthenelson-alphabet:
Background:
https://github.com/google/perfetto/discussions/2847

I followed the guidance from @stevegolton pretty directly. There were a couple
choices I made:
1. I expanded the API for the selection manager to have a batch version of
   resolveSqlEvent. However, I didn't implement an efficient batch resolver, I
   just made it call the internal single-event version in a loop. I don't think
   I understand the existing code well enough to do that myself, but hopefully
   this makes it easier for someone in the future to switch this out w/o needing
   to do any API changes.
2. I added parallel commands to the ...ByRegex ones for pinning and copying to
   another workspace (w/ or w/o ancestors). I didn't add ones for expanding and
   collapsing, since I think those only apply to grouping tracks? Anyhow, let me
   know if I should add parallel ones for those too.


peterthenelson-alphabet:
Did the rebase but have to make a followup commit. Sorry for the notification spam.

peterthenelson-alphabet:
Ok, updated and ready for re-review

ui/src/core/selection_manager.ts
```diff
@@ -228,7 +228,7 @@ export class SelectionManagerImpl implements SelectionManager {
 
   async resolveSqlEvents(
     sqlTableName: string,
-    id: number,
+    ids: number[],
```
stevegolton: Nit: Ideally this should be a ReadonlyArray<number> but I'm not going to block on this.

stevegolton (commented):
Sorry for the delay, I'll take a look at this tomorrow.

ui/src/core/selection_manager.ts
peterthenelson-alphabet: done

ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
```diff
@@ -421,6 +499,20 @@ async function getNameOrPathFromArgOrPrompt(
...
 }
 
+// Helper function to get a sql query from an argument or prompt the user for
+// one. Returns null if the user cancels the prompt.
+async function getQueryFromArgOrPrompt(
+  ctx: Trace,
+  queryArg: unknown,
+  promptText: string,
+): Promise<string | null> {
+  const queryStr =
```
stevegolton, line 509: Nit: We usually prefer `undefined` vs `null` to indicate no return value, but not going to block on this.

stevegolton (approved):
LGTM!

ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
peterthenelson-alphabet: Ack. I'm resolving. I was just being consistent w/the rest of the file (which may need to be updated later).

PR #3189, opened 2025-10-06
WIP: tprobes: generic ftrace events: include group name

rsavitski:


src/traced/probes/ftrace/proto_translation_table.cc
```diff
@@ -632,19 +633,25 @@ const Event* ProtoTranslationTable::CreateGenericEventInternal(
...
-    f->set_type(type);  // e.g. int32, fixed64
-  };
+  auto* event_pb_descriptor =
+      outer_descriptor->BeginNestedMessage<protos::pbzero::DescriptorProto>(
+          GED::kEventDescriptorFieldNumber);
+
+  event_pb_descriptor->set_name(group_and_name.name());
+
+  auto add_pb_desc_field =
+      [&event_pb_descriptor](uint32_t id, const std::string& name,
```
sashwinbalaji: nit: purely from clarity purpose, capture by value as we are not modifying the pointer itself ?

rsavitski: agreed, good spot.

PR #3210, opened 2025-10-08
tp: lower logging level of "missing uuid in hierarchy for track"

sashwinbalaji:


src/trace_processor/importers/proto/track_event_tracker.cc
```diff
@@ -672,7 +672,7 @@ bool TrackEventTracker::IsTrackHierarchyValid(uint64_t uuid) {
     }
     auto* state_ptr = descriptor_tracks_state_.Find(current_uuid);
     if (!state_ptr) {
-      PERFETTO_ELOG("Missing uuid in hierarchy for track %" PRIu64, uuid);
```
LalitMaganti, line 675: Can we make this an error stat and just remove this entirely?

src/trace_processor/storage/stats.h
```diff
@@ -245,6 +245,7 @@ namespace perfetto::trace_processor::stats {
       "TrackEventRangeOfInterest packet, and track event dropping is "         \
       "enabled."),                                                             \
   F(track_event_tokenizer_errors,         kSingle,  kInfo,     kAnalysis, ""), \
+  F(track_hierarchy_missing_uuid,         kSingle,  kError,    kAnalysis, ""), \
```
LalitMaganti: The last comment should be a description explaining:
1) what the error means
2) what sort of issues it can cause
3) how the user can fix it.

See some of the most recent stats below.

LalitMaganti: Friendly ping on this :)

sashwinbalaji: oops, sorry missed it.

PR #3222, opened 2025-10-08
UI: Link official Wattson documentation and update crude_estimate

wusamuel6:
1. Link to official Wattson doc page for more detailed information.
2. Call "estimate" tracks as "crude estimates" tracks if using swapper to get idle state

Bug: 448182021


ui/src/plugins/org.kernel.Wattson/index.ts
```diff
@@ -245,7 +245,17 @@ function createCpuWarnings(
     warningMsg.push(
       m(
         '.pf-wattson-warning',
-        'Perfetto trace configuration is missing below trace_events for Wattson to work. ',
+        'See ',
+        m(
```
LalitMaganti: Please use the Linkify widget or some other LInk widget instead of a direct href.

PR #3226, opened 2025-10-09
redactor: add correct dep on base

LalitMaganti:
We should never depend on base headers, always on base implementation
target.


LalitMaganti:
> In that case should we update all such dependencies ?

Absolutely but maybe that can wait for a followup?

PR #3236, opened 2025-10-09
exp: Modify Columns node details

aMayzner:
Show the details of the modify column node on the node graph


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/modify_columns_node.scss
stevegolton, line 1: Missing header comment

PR #3241, opened 2025-10-09
Flush buffers before periodically write them into file.

KirillTim:
For `write_into_file` sessions, each time we periodically write buffers into file,
we first issue a Flush() to all data source, forcing  them to commit their data into the tracing service.
This helps us to always write latest data to the file.

Add `no_flush_before_write_into_file` configuration option to disable this new behaviour.

### DIFF
--- a/ui/src/plugins/dev.perfetto.WidgetsPage/widgets_page.ts
+++ b/ui/src/plugins/dev.perfetto.WidgetsPage/widgets_page.ts
@@ -383,7 +383,12 @@ function ControlledPopup() {
       return m(
         Popup,
         {
-          trigger: m(Button, {label: `${popupOpen ? 'Close' : 'Open'} Popup`}),
+          trigger: m(Button, {
+            label: `${popupOpen ? 'Close' : 'Open'} Popup`,
+            onclick: () => {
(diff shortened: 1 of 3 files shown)


src/tracing/service/tracing_service_impl.cc
```diff
@@ -2452,35 +2461,58 @@ bool TracingServiceImpl::ReadBuffersIntoFile(TracingSessionID tsid) {
...
-  bool stop_writing_into_file = false;
-  do {
-    std::vector<TracePacket> packets =
-        ReadBuffers(tracing_session, kWriteIntoFileChunkSize, &has_more);
-
-    stop_writing_into_file = WriteIntoFile(tracing_session, std::move(packets));
-  } while (has_more && !stop_writing_into_file);
+  auto do_read_buffers_into_file_fn =
+      [this, tracing_session, tsid](bool async_flush_buffers_before_read) {
+        // ReadBuffers() can allocate memory internally, for filtering. By
```
LalitMaganti, line 2469: You cannot keep a pointer to TracingSession across a PostTask: You need to refetch it using GetTracingSession.

PR #3242, opened 2025-10-09
UI: Assign Gray Color to Anonymized Threads and Optimize Colorizer

g-aattar:
This improves the Perfetto UI by assigning a distinct color to anonymized threads and optimizing the color generation logic for better performance and readability.

Changes:
-  Gray for Anonymized Threads: The colorForThread function now identifies threads named "Anonymized Process" (originating from the ETW parser) and assigns them a gray color. This makes them easy to distinguish on CPU tracks.
- Performance Optimization: A cache (Map) has been added to the colorForTid function. This memoizes results based on the thread ID, avoiding repeated string conversions and hashing in hot render paths.
- Improved Readability: The magic string "Anonymized Process" was extracted into a named constant with a comment indicating its source.
- Testing: Added and updated unit tests to verify the new coloring logic and ensure the caching mechanism works as expected.

### DIFF
--- a/ui/src/components/colorizer.ts
+++ b/ui/src/components/colorizer.ts
@@ -149,15 +149,30 @@ function proceduralColorScheme(seed: string): ColorScheme {
   }
 }
 
+const tidColorCache = new Map<number, ColorScheme>();
+/**
+ * Selects a predictable color scheme from a palette of material design colors,
+ * based on a thread id.
+ */
 export function colorForTid(tid: number): ColorScheme {
-  return materialColorScheme(tid.toString());
+  let colorScheme = tidColorCache.get(tid);
+  if (colorScheme) {
+    return colorScheme;
+  }
+  colorScheme = materialColorScheme(tid.toString());
+  tidColorCache.set(tid, colorScheme);
+  return colorScheme;
 }
+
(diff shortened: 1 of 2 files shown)


ui/src/components/colorizer.ts
```diff
@@ -149,15 +149,30 @@ function proceduralColorScheme(seed: string): ColorScheme {
...
+  let colorScheme = tidColorCache.get(tid);
+  if (colorScheme) {
+    return colorScheme;
+  }
+  colorScheme = materialColorScheme(tid.toString());
+  tidColorCache.set(tid, colorScheme);
+  return colorScheme;
 }
 
+// From src/trace_processor/importers/etw/etw_parser.cc
```
LalitMaganti, line 167: Sorry but this not a hack we can accept in the core UI. Please follow the guidance I gave on the chat on how to expose this properly (new column in thread table, using that column in the sched plugin) rather than hacking  core classes in the UI like this.

stevegolton: Seconded

g-aattar: Will make the appropriate changes. Thanks for the info!

PR #3250, opened 2025-10-10
Remove `std::move` statements for returning values from functions

tobbi:
Removes `std::move` from all functions for returning values to the caller.

All of these issues were reported by cppcheck.


tobbi:
Looks like I'm hitting this:
```
../../src/trace_processor/trace_processor_impl.cc:261:10: error: call to implicitly-deleted copy constructor of 'perfetto::trace_processor::sql_modules::RegisteredPackage'
  return new_package;
         ^~~~~~~~~~~
../../src/trace_processor/util/sql_modules.h:41:46: note: copy constructor of 'RegisteredPackage' is implicitly deleted because field 'modules' has a deleted copy constructor
  base::FlatHashMap<std::string, ModuleFile> modules;
                                             ^
../../include/perfetto/ext/base/flat_hash_map.h:163:3: note: 'FlatHashMap' has been explicitly marked deleted here
  FlatHashMap(const FlatHashMap&) = delete;
  ^
../../include/perfetto/ext/base/status_or.h:49:14: note: passing argument to parameter 'value' here
  StatusOr(T value) : StatusOr(base::OkStatus(), std::move(value)) {}
             ^
1 error generated.
```

I'm assuming adding an inline suppression for cppcheck is the way to go forward here, in order to ignore the warning?

LalitMaganti:
We should also be able to wrap the move with a base::StatusOr which should satisfy both the compiler and cppcheck

tobbi:
> We should also be able to wrap the move with a base::StatusOr which should satisfy both the compiler and cppcheck

I will try to make the necessary modifications.

Issue #3243, opened 2025-10-09
[UI] Allow 'Max duration' input by keyboard


splhack:
<img width="1266" height="192" alt="Image" src="https://github.com/user-attachments/assets/a3e85544-da48-40c9-8d59-930c2bb01e94" />

IIRC, the `h:m:s` text input field used to accept keyboard input, for example, backspace to erase the existing value, put number directly to specify duration. Is it possible to accept such keyboard input?

LalitMaganti:
Thanks for filing! Request makes sense, sent https://github.com/google/perfetto/pull/3251 for this.

PR #3251, opened 2025-10-10
ui: allow setting record time from keyboard

LalitMaganti:
Also remove track_event from Chrome/CrOS: I forgot that they have their
own dedicated section so there's no point enabling it for them.

Fixes: https://github.com/google/perfetto/issues/3243


stevegolton:
It seems that after typing in a value, the value can no longer be adjusted using the slider.

LalitMaganti:
> It seems that after typing in a value, the value can no longer be adjusted using the slider.

Thanks for catching. Fixed.

PR #3255, opened 2025-10-10
Add skipped sample as a separate track

haixiaosugoogle:
Screenshot 2025-10-10 at 16 35 00

### DIFF
--- a/ui/src/plugins/dev.perfetto.LinuxPerf/perf_samples_profile_track.ts
+++ b/ui/src/plugins/dev.perfetto.LinuxPerf/perf_samples_profile_track.ts
@@ -125,11 +125,7 @@ export function createPerfCallsitesTrack(
   });
 }
 
-export function createSkippedCallsitesTrack(
-  trace: Trace,
-  uri: string,
-) {
-
+export function createSkippedCallsitesTrack(trace: Trace, uri: string) {
   return SliceTrack.create({
     trace,
     uri,


ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -28,23 +30,44 @@ import {Flamegraph} from '../../widgets/flamegraph';
...
+export default class LinuxPerf implements PerfettoPlugin {
   static readonly id = 'dev.perfetto.LinuxPerf';
   static readonly dependencies = [
     ProcessThreadGroupsPlugin,
     TraceProcessorTrackPlugin,
   ];
+  private static showSkippedPerfSamples: Setting<boolean>;
+
+  static onActivate(app: App): void {
+    const boolSettingDesc: SettingDescriptor<boolean> = {
```
rsavitski: When we discussed flagging, I meant feature_flags.ts, not settings. 

The settings are toggles for end users that we expect to keep around, while flags are for things that are still in development or where the UX is not finalised (which is the case here).

However I'm also no longer insisting on this being flagged. We can always roll back if the tracks prove surprising.

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -152,6 +175,49 @@ export default class implements PerfettoPlugin {
...
+          upid,
+        },
+        renderer: createSkippedCallsitesTrack(trace, uri, upid),
+      });
+      const group = trace.plugins
+        .getPlugin(ProcessThreadGroupsPlugin)
+        .getGroupForProcess(upid);
+      const summaryTrack = new TrackNode({
+        uri,
+        name: `Process callstacks (skipped)`,
```
rsavitski: Still not a fan of "skipped" because:
1) it's not clear to users what is meant by a skipped sample (is it an error?)
2) it's completely valid to do sampling of counters without callstacks, in which case all samples will appear "skipped".

I'm struggling to suggest a better name. How about "Unclassified perf samples"?

ui/src/plugins/dev.perfetto.LinuxPerf/perf_samples_profile_track.ts
```diff
@@ -125,6 +125,52 @@ export function createPerfCallsitesTrack(
...
 
+export function createSkippedCallsitesTrack(
+  trace: Trace,
+  uri: string,
+  upid?: number,
+  utid?: number,
+  sessionId?: number,
+) {
+  const constraints = [];
+  if (upid !== undefined) {
```
rsavitski: There's quite a bit of copypasted code that is never used. I've started a more minimal implementation [here](https://github.com/google/perfetto/pull/3345/files) (including a different visual styling to avoid confusion with callstacks). Can we similarly cut down the unnecessary code in this patch?

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
haixiaosugoogle: alright, removed flags.

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
haixiaosugoogle: Renamed!

ui/src/plugins/dev.perfetto.LinuxPerf/perf_samples_profile_track.ts
haixiaosugoogle: Yup! I have cut it down to just rendering the track I think, PTAL.

PR #3258, opened 2025-10-10
Update some path changes to docs.

haixiaosugoogle:


stevegolton:
Thanks! I think we have some conflicting changes now as we made similar changes, but this PR has some updates to the imports that I missed, so definitely worth still landing if you don't mind rebasing and sorting out the merge conflict?

PR #3259, opened 2025-10-10
ui: customizable minimum track height

cdamus:
Tracks once occupied more vertical space, which can be more comfortable especially for users with vision and motor (pointer control) difficulties. To accommodate embedding applications and users with accessibility needs, add support for customization of the height of tracks via a new setting.

Besides accessibility concerns, this also relates to the customizability of the UI for embedding applications, such as captured in issues #2166 and #2266. For example, my application that reuses Perfetto UI sets the minimum height of tracks using this new setting (the settings UI is not made available to my users) to match the look and feel of the rest of the host application.

----

A demonstration of the new setting:

https://github.com/user-attachments/assets/77f12343-3f13-441a-bb53-4a413c714fbd

### DIFF
--- a/docs/contributing/ui-plugins.md
+++ b/docs/contributing/ui-plugins.md
@@ -1681,7 +1681,10 @@ interact with the flag's state:
 **Example:**
 
 ```typescript
-import {App, Flag, FlagSettings, PerfettoPlugin, Trace} from '../../public'; // Adjust path as needed
+import {Flag, FlagSettings} from '../../public/featureflag'; // Adjust path as needed
+import {App} from '../../public/app';
+import {PerfettoPlugin} from '../../public/plugin';
+import {Trace} from '../../public/trace';
 
 export default class MyFeatureFlagPlugin implements PerfettoPlugin {
   static readonly id = 'com.example.MyFeatureFlagPlugin';
@@ -1763,7 +1766,10 @@ the descriptor and provides methods to interact with the setting:
 **Example:**
 
 ```typescript
-import {App, Setting, SettingDescriptor} from '../../public'; // Adjust path
+import {Setting, SettingDescriptor} from '../../public/setting'; // Adjust path as needed
+import {App} from '../../public/app';
+import {PerfettoPlugin} from '../../public/plugin';
+import {Trace} from '../../public/trace';
 import {z} from 'zod';
 import m from 'mithril';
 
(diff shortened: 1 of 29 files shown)


ui/src/frontend/index.ts
```diff
@@ -337,6 +337,15 @@ function onCssLoaded() {
     defaultValue: 'light',
   });
 
+  AppImpl.instance.settings.register({
+    id: 'track-height-min-px',
```
stevegolton: While we haven't really been consistent on id naming conventions, we've settled on a standard convention for commands at least, so we should probably start using the same for other entities.

That is to say we put the 'dev.perfetto' prefix in front for 'core' entities and use UpperCamelCase for the rest.

So this settings id should be `dev.perfetto.TrackHeightMinPx`.

cdamus: Yup I can take care of that.

cdamus:
> Looks good to me. I wish we had a type-safe way to drill settings down through multiple layers of mithril components (other than adding those settings to settings to App).

Are you asking me to devise one? I'm happy to try 😀 

AFAICT Mithril has no first-class analogue of React's context provider. If that's even still a thing in React ...  Maybe something along that line ...


> Out of interest @cdamus, when embedding Perfetto, don't you need to configure these settings externally? If so, how do you manage this?

My app uses the private SettingsManagerImpl API to change the setting before ever any trace is opened.

ui/src/frontend/index.ts
cdamus: Done with commit 7dac69b431c77482b3b9bc375da00e12eea85101. Note that I also changed the settings sorting to sort by name because that sorts on what users actually see but also now IDs aren't in order.

stevegolton: Makes sense to me, thanks!

stevegolton:
> Are you asking me to devise one? I'm happy to try 😀
>
> AFAICT Mithril has no first-class analogue of React's context provider. If that's even still a thing in React ... Maybe something along that line ...

I've been thinking about this for a while, and I always come to the conclusion that the least surprising approach is to just drill a service or context object through the layers of components, which is really what we're doing with App/Trace. But, I'm not very imaginative so if you have a better idea I'm all ears! 😀

> My app uses the private SettingsManagerImpl API to change the setting before ever any trace is opened.

Makes sense!

PR #3263, opened 2025-10-11
Fix "Searching before insertion is not necessary"

tobbi:
This fixes one cppcheck performance warning about searching before insertion not being necessary.


tobbi:
There is also this error:
```
src/trace_processor/trace_processor_shell.cc:308:11: performance: Ineffective call of function 'substr' because a prefix of the string is assigned to itself. Use replace() instead. [uselessCallsSubstr]
    sql = sql.substr(0, kPrefix.size()) + "perfetto_export." +
          ^
```

but i don't know how to fix that without making readability worse.

LalitMaganti:
> but i don't know how to fix that without making readability worse.

Maybe make it so that it's assigned to a different variable?

PR #3270, opened 2025-10-13
ui: Add additional simple slice aggregator to fix missing aggregator

stevegolton:
Since #2946, slice tracks must contain a `parent_id` column in order to be considered for slice aggregation, which rules out some of the slice-like plugin tracks that are generated from a query and don't necessarily have a parent_id in their data source or are tracks with no concept of a stack or depth.

This patch relaxes the parent_id constraint, allowing tracks both with and without parent_id to be aggregated.

Fixes https://buganizer.corp.google.com/issues/448068359

### DIFF
--- a/ui/src/plugins/dev.perfetto.TraceProcessorTrack/slice_selection_aggregator.ts
+++ b/ui/src/plugins/dev.perfetto.TraceProcessorTrack/slice_selection_aggregator.ts
@@ -1,4 +1,4 @@
-// Copyright (C) 2025 The Android Open Source Project
+// Copyright (C) 2020 The Android Open Source Project
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


LalitMaganti:
Had a simpler idea while looking at this: can we make the parent_id optional in the slice aggregator and return null if it's not present in the underlying dataset? Having a slice aggregator *and* a simple slice aggregator is quite confusing, I'm not sure anyone will understand the meaning of both of them.

stevegolton:
> Had a simpler idea while looking at this: can we make the parent_id optional in the slice aggregator and return null if it's not present in the underlying dataset? Having a slice aggregator and a simple slice aggregator is quite confusing, I'm not sure anyone will understand the meaning of both of them.

That makes sense, though we need to be careful when to match up the parent ids to ids due to the id namespace issue we were talking about the other day. Definitely doable and I agree it'll be less confusing, so I'll look into this instead.

LalitMaganti:
LGTM but should we change the commit description to match?

ui/src/plugins/dev.perfetto.TraceProcessorTrack/slice_selection_aggregator.ts
```diff
@@ -29,64 +30,152 @@ import {
...
-            GROUP BY parent_id
-          `,
-        });
+        const unionQueries: string[] = [];
+        await using trash = new AsyncDisposableStack();
+
+        if (sliceDatasets.length > 0) {
+          const query = await this.buildSliceQuery(
+            engine,
+            createUnionDataset(sliceDatasets).optimize(),
```
LalitMaganti, line 80: Doesn't this have the problem you mentioned (i.e. different overlapping id spaces)? How do we guard against two different tables having the same id but meaning different things?

stevegolton: TL;DR: It's fragile but it's actually no worse than it was before this PR - and this PR is just so I can fix the bug and get the stable release out.

We are dealing with two sets of tracks here:
- **Tracks with parent_id**: As in the current state of the codebase, these are exclusively trace processor type tracks from the slice table added by the TraceProcessorTrack or TrackEvent plugins. They all share the same id namespace (from the slice table).
- **Track without parent_id**: A lot of plugins add these sorts of tracks (including debug tracks) and their id namespace can be anything.

Before this PR we were simply selecting the former and ignoring the latter, hence the bug. This PR changes this so that both sets of tracks are selected, but we need to make sure to keep the ids separate until after we've matched up the parent ids.

My point is we combine the former set of tracks in a single bucket before matching up parent_ids to ids because we know that they all come from the same namespace.

Yes, this is fragile. If a plugin were to come along and add a track with a parent_id but whose id namespace is not from the slice table then this would break. However, this is no worse than it was before this PR. This change is just to fix the bug so I can get the stable release out, so I'll tackle the issue more completely in a followup.

LalitMaganti: Ack sounds good.

Issue #764, opened 2024-04-12
process_sort_index support or alternative suggestions


nikhilgdev:
Hi,
My team is working on a project that generates a flow of timed events in a system. I am using the trace event format as described here: https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview.
Each of these events is described as a process in the JSON file.

Unfortunately, perfetto UI does not order these processes by start time. It seems to order the processes by 'eventfulness' instead.

As a backup, I figured I would use the process_sort_index metadata item to re-order the processes as needed, however, perfetto UI does not respect that.

I see past issues where implementation of process_sort_index has been shot down. Is there an alternative you could suggest for this? Needing to order the processes chronologically is (I would imagine) a need for many projects adopting this UI.

LalitMaganti:
I don't think there's any suggestion for now. Having arbitrary sorts at the UI level might be something supported at some point but is not a priority. 

Our priority is first features for native Perfetto traces (especially Android system tracing). If our project is used in other ways, we're very happy with that but it's not our focus.

LalitMaganti:
This is now supported in the protobuf trace format: https://perfetto.dev/docs/reference/synthetic-track-event#track-sorting-order

We might also add support for this to the JSON format at some point but it requires signifcantly more work due to interactions with processes: the protobuf one works just on the tracks level, it does not touch processes. 

Closing this out because there is now a semi-reasonable workaround for this.

LalitMaganti:
With https://github.com/google/perfetto/pull/3273 landing, we now support thread_sort_index and process_sort_index keys from JSON. Proto traces are still preferred to be written of course but for those writing JSON traces, it might be a nice quality of life improvement.

nikhilgdev:
This is excellent. Also works for traces generated using Clang since they still generate the trace event file format.
How do we get access to this? I tried to use the canary version of perfetto today with a json trace events file and it didn't respect the process sort index.

LalitMaganti:
We've not pushed to canary yet, it releases once a week. I think it should have released yesterday but didn't. Will check why.

It's available on autopush which you can switch to temporarily if you want to try it at Settings -> Release channel -> Autopush. I would *not* recommend staying on autopush for longer than testing to avoid unnecessarily being broken in the future.

Issue #555, opened 2023-08-11
Cannot control thread order in perfetto 


bartosz-lew-int:
With the following trace event file:

```
{"traceEvents":[
{"name": "process_name", "ph": "M", "pid":0, "args": {"name" : "Process (0)"}},
{"name": "process_sort_index", "ph": "M", "pid":0, "args": {"sort_index" : "0"}},
{"name": "thread_name", "ph": "M", "pid":0, "tid":0, "args": {"name" : "th1"}},
{"name": "thread_name", "ph": "M", "pid":0, "tid":1, "args": {"name" : "th2"}},
{"name": "process_name", "ph": "M", "pid":1, "args": {"name" : "Process (1)"}},
{"name": "process_sort_index", "ph": "M", "pid":1, "args": {"sort_index" : "1"}},
{"name": "thread_name", "ph": "M", "pid":1, "tid":0, "args": {"name" : "th1"}},
{"name": "thread_name", "ph": "M", "pid":1, "tid":1, "args": {"name" : "th2"}},
{"name":"A", "cat":"X", "ph":"X", "ts":1, "dur":3, "pid":0, "tid":0},
{"name":"B", "cat":"Y", "ph":"X", "ts":2, "dur":2, "pid":0, "tid":1},
{"name":"A", "cat":"X", "ph":"X", "ts":1, "dur":3, "pid":1, "tid":0},
{"name":"B", "cat":"Y", "ph":"X", "ts":2, "dur":2, "pid":1, "tid":1}
]
}

```
one obtains the following report

<img width="814" alt="image" src="https://github.com/google/perfetto/assets/110619762/66d4aca9-b867-4469-96ef-63cdebeb1222">

where thread order in the first process follows thread id (or thread name lexicographic ordering) but the second process that contains exactly the same data shows threads in different order.

The expected result would be to have both threads in the same order.

using version: [v35.0-8019b](https://android.googlesource.com/platform/external/perfetto/+/8019b8809dfbec51c8e1ca38e83c8b60da626585/ui) from 
https://ui.perfetto.dev/

LalitMaganti:
This is working as intended: we very intentionally do not support process_sort_index as we made a conscious choice to separate trace rendering from the trace contents. See also our policy on JSON traces [here](https://perfetto.dev/docs/faq#why-does-perfetto-not-support-lt-some-obscure-json-format-feature-gt-).

[...]

Issue #378, opened 2022-11-08
Process sorting is (seemingly) broken


noomorph:
I have a parent and a child process in my timeline, and the parent's lifespan is longer: it starts earlier and ends later.

However, the currently running Perfetto version (as of November 8th, 2022) weirdly sorts my processes, making the child process look like it is more important:

![Screen Shot 2022-11-08 at 12 41 06](https://user-images.githubusercontent.com/1962469/200544243-55289e10-3280-460b-9d56-91dcb7b5b35b.png)

As a workaround, I used `process_sort_index` metadata events to enforce the sorting order, but it seems that there's no support for those events in the source code of Perfetto.

Here's the gist of my stack trace to reproduce the issue:

https://gist.github.com/noomorph/57bb8a80ea44d1ff6fdee6695e9b7afd

I raised a little discussion in Discord, and @chromy helped to discover that the default sorting strategy relies on the number of slices in each process. Perhaps, it makes sense for some projects, but what can I do in my case?

Thanks in advance!

chromy:
Me on Discord:
> @eseckler @altimin @noomorph Hey 🙂 @noomorph and I were talking in ui-dev about tweaking the sorting order. Specifically changing https://cs.android.com/android/platform/superproject/+/master:external/perfetto/ui/src/controller/track_decider.ts;l=1503?q=track_decider&ss=android%2Fplatform%2Fsuperproject:external%2Fperfetto%2Fui%2Fsrc%2F to sort processes by first slice timestamp rather than by slice count. I think that would effect Chrome more than Android since on Android we are more likely to have the sched events so I wanted to know your thoughts 😄

@betasheet @altimin let me know :)

betasheet:
Hm, I think that the chromeProcessRank rank would help ensure that the browser process remains top (although I have a feeling that's not true for system traces). I'd also be a little worried about the behavior in ring buffer traces, because slices/TrackEvent data can start at different points for different threads/processes after the ring buffer wraps, even if all of them were "alive" for the full duration of the trace, and it's rather random which thread emits/retains the first slice then.

altimin:
Re: chromeProcessRank: I think we should switch it to use chrome_processes.sql , which should be more reliable anyway. 

For this specific problem, I don't think that relying on the trace events themselves is a good idea — I'd rather look into populating `parent_upid` column in the `process` table and using that to order the processes.

LalitMaganti:
Folding this into https://github.com/google/perfetto/issues/429

[...]

PR #3273, opened 2025-10-13
tp: ui: add support for thread_sort_index/process_sort_index from JSON

LalitMaganti:
We've relaxed our stance on this for the protobuf format, so there's no reason to leave
people writing JSON traces out in the cold. Respect the standard sort_index key
coming from JSON traces as well.

Fixes: https://github.com/google/perfetto/issues/555
Fixes: https://github.com/google/perfetto/issues/764
Fixes: https://github.com/google/perfetto/issues/378


slai-natanijel:
Great feature, thank you. When can we expect the changes to be reflected in Canary web?

LalitMaganti:
> Great feature, thank you. When can we expect the changes to be reflected in Canary web?

Canary should be released on Wednesday in our normal schedule assuming nothing goes wrong.

balta2ar:
@LalitMaganti 
I'm still very confused. Could you confirm that I'm not missing anything? The Trace Event docs (https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview?tab=t.0) says:
> For the two sort_index items the sort_index argument value specifies the relative position you'd like the item to be displayed. Lower numbers are displayed higher in Trace Viewer. If multiple items all have the **same sort index then they are displayed sorted by name** and, given duplicate names, by id.

So here's my JSON:
```
[
{"cat":"spanner.rwtx","name":"Query","ph":"X","pid":1,"tid":1,"ts":10,"dur":20},
{"cat":"spanner.rwtx","name":"Query","ph":"X","pid":1,"tid":2,"ts":20,"dur":20},
{"cat":"spanner.rwtx","name":"Query","ph":"X","pid":1,"tid":3,"ts":30,"dur":30},
{"name": "thread_name", "ph": "M", "pid": 1, "tid": 1, "args": {"name":"Cherry"}},
{"name": "thread_name", "ph": "M", "pid": 1, "tid": 2, "args": {"name":"Banana"}},
{"name": "thread_name", "ph": "M", "pid": 1, "tid": 3, "args": {"name":"Apple"}},
{"name": "thread_sort_index", "ph": "M", "pid": 1, "tid": 1, "args": {"sort_index":0}},
{"name": "thread_sort_index", "ph": "M", "pid": 1, "tid": 2, "args": {"sort_index":0}},
{"name": "thread_sort_index", "ph": "M", "pid": 1, "tid": 3, "args": {"sort_index":0}},
```
I intentionally set sort_index=0 to all the threads in hope that they get sorted by thread_name, e.g.: Apple 3, Banana 2, Cherry 1. But I see the reverse of that:
image

LalitMaganti:
We're not breaking ties with names but with tids/pids (which is what we've always done). The goal of this change was not to get 1:1 parity with Catapult but simply get to a state where *_sort_index did *something*.

I would need to investigate how difficult it would be to use names without breaking others who rely on tid based sorting. I think we can do it but might be some work.

PR #3276, opened 2025-10-14
stdlib: Replace magic number 255 with formal function _dsu_dep()

runpinglai-creator:
This code change replaces 255, the magic number representing dsu_cpu_dependency, with a function defined in wattson/utils.sql. Code changes from this moment should start referencing _dsu_dep() instead of calling 255.

Bug: 451737922
Test: Perfetto local UI build and test

Welcome to Perfetto!

For more details please see


LalitMaganti:
As an FYI: doing this is *not* zero cost as there's no "function inlining" in SQL. You should make sure this doesn't regress performance.

src/trace_processor/perfetto_sql/stdlib/wattson/curves/utils.sql
```diff
@@ -13,6 +13,8 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
+INCLUDE PERFETTO MODULE wattson.utils;
```
wusamuel6: Please make INCLUDE alphabetical order

src/trace_processor/perfetto_sql/stdlib/wattson/curves/device_l3.sql
```diff
@@ -12,6 +12,8 @@
 -- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
+INCLUDE PERFETTO MODULE wattson.utils;
```
wusamuel6: Please make INCLUDE alphabetical order

runpinglai-creator:
> As an FYI: doing this is _not_ zero cost as there's no "function inlining" in SQL. You should make sure this doesn't regress performance.

@LalitMaganti Thanks for the point. I found that doing so leads to a performance regression. Therefore in latest patch, I replaced `CREATE PERFETTO FUNCTION` with `CREATE PERFETTO MACRO`. Latest patch is passing the performance CI test and query/ingest time didn't slow down.

I noticed that [Perfetto-syntax-guidance ](https://perfetto.dev/docs/analysis/perfetto-sql-syntax) discourages using `PERFETTO MACRO` for simple scalar values. But given the massive references and `PERFETTO FUNCTION` overhead, would my case be ok to use `PERFETTO MACRO`?

LalitMaganti:
Yes it seems fine. Would you mind clarifying that doc (it's in the docs/ folder) saying it's fine to do use macros for this usecase.

runpinglai-creator:
> Yes it seems fine. Would you mind clarifying that doc (it's in the docs/ folder) saying it's fine to do use macros for this usecase.

Of course. Happy to do that. Thanks!

wusamuel6 (approved):

PR #3282, opened 2025-10-14
tp: temporarily work around perf records with zero timestamps

LalitMaganti:
We should come back and do this properly by moving parsing of certain
events to tokenization time but that would be a signifcant undertaking.
Let's just do the essential thing to get this unblocked.

Fixes: b/449973773

### DIFF
--- a/src/trace_processor/importers/perf/perf_data_tokenizer.cc
+++ b/src/trace_processor/importers/perf/perf_data_tokenizer.cc
@@ -358,7 +358,7 @@ base::StatusOr<int64_t> PerfDataTokenizer::ExtractTraceTimestamp(
     return base::ErrStatus("Failed to read time");
   }
 
-  // TODO(lalitm): `*time > 0` is a temporary hack to work around the fact that
+  // TODO(449973773): `*time > 0` is a temporary hack to work around the fact that
   // some perf record types which actually don't have a timestamp. They should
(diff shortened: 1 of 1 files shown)


src/trace_processor/importers/perf/perf_data_tokenizer.cc
```diff
@@ -358,16 +358,20 @@ base::StatusOr<int64_t> PerfDataTokenizer::ExtractTraceTimestamp(
...
+  // trace_ts to `latest_timestamp_`, we don't try and convert a zero timestamp
+  // accidentally, leading to negative timestamps in some clocks.
   base::StatusOr<int64_t> trace_ts =
-      time.has_value()
+      time && *time > 0
           ? context_->clock_tracker->ToTraceTime(record.attr->clock_id(),
                                                  static_cast<int64_t>(*time))
-          : std::min(latest_timestamp_, context_->sorter->max_timestamp());
-
+          : latest_timestamp_;
```
sashwinbalaji, line 371: q: why the min with context_->sorter->max_timestamp() is removed? Is it because it's redundant ?

And also pls add the bug id in todo

LalitMaganti: > q: why the min with context_->sorter->max_timestamp() is removed? Is it because it's redundant ?

I think it was partially redundant but also had the chance of being very wrong.

> And also pls add the bug id in todo

Done

PR #3289, opened 2025-10-14
docs: update SQL Macro usecase

runpinglai-creator:
In PR#3276, we found a MACRO usecase which wasn't listed before. Therefore adding this usecase and explainations of it in the doc.

Basically, it would be ok to use `MACRO` for a constant if it would be referenced massive times and it would be bad to use `FUNCTION` for it which will bring massive performance overhead since SQL `FUNCTION` can't do inlining.

Welcome to Perfetto!

For more details please see

### DIFF
--- a/docs/analysis/perfetto-sql-syntax.md
+++ b/docs/analysis/perfetto-sql-syntax.md
@@ -209,7 +209,7 @@ above.
 If only passing around scalar SQL values, functions are generally preferred
 for their clarity. However, for simple constants used many times in a
 performance-sensitive query, a macro can be more efficient as it avoids the
-potential overhead of massive function calls.
+potential overhead of function calls in a large number.
 
 NOTE: Macros are expanded with a pre-processing step *before* any execution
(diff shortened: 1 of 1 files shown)


docs/analysis/perfetto-sql-syntax.md
```diff
@@ -199,13 +199,18 @@ is inspired by the macros in Rust.
...
 Macros are powerful but also dangerous if used incorrectly, making debugging
 extremely difficult. For this reason, it's recommended that they are used
 sparingly when they are needed and only for the recommended uses described
-above. If only passing around scalar SQL values, use functions as discussed
 above.
 
+If only passing around scalar SQL values, functions are generally preferred
+for their clarity. However, for simple constants used many times in a
+performance-sensitive query, a macro can be more efficient as it avoids the
+potential overhead of massive function calls.
```
LalitMaganti: I would word this as "large number of function calls" instead of "massive"

LalitMaganti (approved):

Issue #3287, opened 2025-10-14
crash in trace_processor_shell when selecting GPU rendertrace slices.


dfriederich:
I'm working on sokatoa, where we use an external trace_processor_shell as the trace processor.
But I can reproduce the same crash with the current main perfetto sources (67c6c7c797bb83ff16edad31eee3cef23b268a46),
by building trace_processor_shell, starting it with -D, when building the perfetto ui, starting it, then load the attached perfetto file 

[sokatoa_gapid.zip](https://github.com/user-attachments/files/22914489/sokatoa_gapid.zip)

trace_processor_shell runs fine initially, but crashes when selecting gpu rendertrace slices in the GPU->Render Stages -> Universal track.

the crash only happens when any slice is selected. Also the crash only happens when trace_processor_shell is running with -D.
When perfetto uses the webassembly based trace processor, the crash is not observed.
The actual crash is due to a nullptr being used with strlen.

console output from trace_processor_shell:
```
[616.800]    proto_decoder.cc:140 Skipping field 102672883 because its id > 16777215                                                                                                                                                                    
[616.800]    proto_decoder.cc:140 Skipping field 102672883 because its id > 16777215                                                                                                                                                                    
[616.800]    proto_decoder.cc:140 Skipping field 102672883 because its id > 16777215                                                                                                                                                                    
Loading trace 0.27 MB (0.0 MB/s)                                                                                                                                                                                                                        
../../buildtools/libcxx/include/string_view:268: assertion __s != nullptr failed: null pointer passed to non-null argument of char_traits<...>::length                                                                                                  
                                                                                                                                                                                                                                                        
------------------ BEGINNING OF CRASH ------------------                                                                                                                                                                                                
Signal: Abort                                                                                                                                                                                                                                           
Fault addr: 0000040700050C30                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                        
Backtrace:                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                        
#00  0x0000770F28C45330                                                                                                                                                                                                                                 
     ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0                                                                                                                                                                                      
                                                                                                                                                                                                                                                        
#01  __GI___pthread_kill                                                                                                                                                                                                                                
     ./nptl/pthread_kill.c:89                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                        
#02  __GI_raise                                                                                                                                                                                                                                         
[...]

PR #3292, opened 2025-10-15
tp: fix crash when writing args

LalitMaganti:
If the string is null, we should treat it like json null

Fixes: https://github.com/google/perfetto/issues/3287


sashwinbalaji (commented):
Same as #3291

LalitMaganti:
Fixed it. Can you please rereview?

PR #3293, opened 2025-10-15
Revert "ui: refactor display of args (#3032)"

altimin:
This reverts commit bbead825da17d13e1353a52525bb26f1007af012.

JSON.parse doesn't properly handle int64, leading to confusion.
Also causes crashes in some corner cases like #3287.


altimin:
@stevegolton we should include this in Canary as well.

PR #3294, opened 2025-10-15
Add metric metadata spec to trace summary

elrakadm:
Add metadata to expand information on dimension columns related to a metric.

For example:  process upid or package name can be expanded to all process metadata without adding all the metadata columns to the metric dimensions 

Bug: b/448858753


LalitMaganti:
> TraceSummary already has a metadata field. Doesn't it help?

No it is not sufficient because you might want to have different "contextual" things per metric. The metadata at the top level is for context of the trace (things in the metadata table etc). This is more about "to understand this metric, here is some additional data you might need". 

We might want to reframe this as "interned dimensions" as in some sense that's really what it is.

PR #3295, opened 2025-10-15
Add metric interned dimension to trace summary

elrakadm:
Add metadata to expand information on dimension columns related to a metric.

For example: process upid or package name can be expanded to all process metadata without adding all the metadata columns to the metric dimensions

Bug: b/448858753


elrakadm:
Messed up local git in https://github.com/google/perfetto/pull/3294 , so created another pull request

LalitMaganti:
See my comments in https://github.com/google/perfetto/pull/3294

elrakadm:
> See my comments in #3294

Just adding for visibility 
[aMayzner](https://github.com/aMayzner) left a comment
TraceSummary already has a metadata field. Doesn't it help?

[LalitMaganti](https://github.com/LalitMaganti) commented [32 minutes ago](https://github.com/google/perfetto/pull/3294#issuecomment-3406446819)
TraceSummary already has a metadata field. Doesn't it help?

No it is not sufficient because you might want to have different "contextual" things per metric. The metadata at the top level is for context of the trace (things in the metadata table etc). This is more about "to understand this metric, here is some additional data you might need".

We might want to reframe this as "interned dimensions" as in some sense that's really what it is.

elrakadm:
> We might want to reframe this as "interned dimensions" as in some sense that's really what it is.

Do you mean rename Metadata to InternedDimension ?

LalitMaganti:
Yeah I think framing in that way is a lot more intuitive and is also clear "this is an option thing you can use to "expand" the metric dimensions if you want but you're not required to do so.

elrakadm:
> Yeah I think framing in that way is a lot more intuitive and is also clear "this is an option thing you can use to "expand" the metric dimensions if you want but you're not required to do so.

Renamed

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -323,6 +323,53 @@ message TraceMetricV2Spec {
...
+  //   }
+  // }
+  // ```
+  //
+  // Next id: 4
+  message InternedDimensionSpec {
+    // Spec for a interned dimension column.
+    //
+    // Next id: 3
+    message InternedDimensionColumnSpec {
```
LalitMaganti: This can be simpler because it's already nested inside the thing

protos/perfetto/trace_summary/v2_metric.proto
LalitMaganti, line 218: Can you move unit/polarity above the "here be dragons" makrer.

LalitMaganti: Expand the note here that you should be very careful using the belwo features as they can have subtle implications and may not do what you expect.

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -323,6 +323,53 @@ message TraceMetricV2Spec {
     NOT_APPLICABLE = 3;
   }
   optional MetricPolarity polarity = 10;
+
+  // Specification for a interned dimension associated with this metric.
+  //
+  // An interned dimension can be used to attach extra information to one of
```
LalitMaganti: I would also say, for a given dimension there can only be *one* interned dimension attached to it.

LalitMaganti: I would reframe how we are pitching it. Instead of "you can attach extra info to a dimension", I would more frame it as "sometime you have a problem where there are a bunch of columns assocaited to a single dimension and repeating many times would blow up the proto size. Using interneddimensions, you can emit only a ID in the key and then "expand" to match.

Note that this means the query *must* return a unique row per dimension (as I don't want more than 1 row per input dimension).

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -533,4 +587,33 @@ message TraceMetricV2Bundle {
...
+  //
+  // Next id: 4
+  message InternedDimensionBundle {
+    // A row of interned dimension, with values for columns specified in
+    // |column_specs|.
+    //
+    // Next id: 2
+    message InternedDimensionRow {
+      // The values in the row.
+      message InternedDimensionValue {
```
LalitMaganti: We should be able to safely reuse Dimension.value.

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -533,4 +587,33 @@ message TraceMetricV2Bundle {
...
+          double double_value = 3;
+          Null null_value = 4;
+        }
+      }
+      repeated InternedDimensionValue interned_dimension_values = 1;
+    }
+    repeated InternedDimensionRow interned_dimension_rows = 2;
+
+    // The spec for this interned dimension bundle.
+    optional TraceMetricV2Spec.InternedDimensionSpec spec = 3;
```
LalitMaganti: Why do we need the spec? We already implicitly know it from the spec which is echoed globally.

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -533,4 +587,33 @@ message TraceMetricV2Bundle {
   //
   // Useful for knowing what the dimension names/value names are.
   repeated TraceMetricV2Spec specs = 3;
+
+  // A bundle of interned dimension rows associated with one of template's
```
LalitMaganti, line 609: I think an example here would really help. I know the strucuture and I'm still very confused.

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -533,4 +587,33 @@ message TraceMetricV2Bundle {
...
+  // interned dimension specs.
+  //
+  // Next id: 4
```
LalitMaganti, line 660: Mark this as advanced and most beginner users do not need to care about this proto.

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -533,4 +587,33 @@ message TraceMetricV2Bundle {
...
+  message InternedDimensionBundle {
+    // A row of interned dimension, with values for columns specified in
+    // |column_specs|.
+    //
+    // Next id: 2
+    message InternedDimensionRow {
```
LalitMaganti, line 666: Don't we need to include the lookup key as well?

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Added an example

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: added a comment at the top that this is an advanced feature

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: I always include the key as the first "interned_dimension_values" value column

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: updated!

elrakadm: Tried to rephrase it better

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Good point removed

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Done

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -323,6 +303,86 @@ message TraceMetricV2Spec {
...
+  // }
+  // ```
+  //
+  // Next id: 4
+  message InternedDimensionSpec {
+    // Spec for a interned dimension column.
+    //
+    // Next id: 3
+    message ColumnSpec {
+      // The name of the column (e.g. name, version_code, etc.).
```
LalitMaganti: Here and below: please say "optional" or "required" to tell people whether or not they need to emit them.

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -533,4 +600,72 @@ message TraceMetricV2Bundle {
   //
   // Useful for knowing what the dimension names/value names are.
   repeated TraceMetricV2Spec specs = 3;
+
+  // !!! THIS IS AN ADVANCED FEATURE !!!
+  // Most beginner users do not need to care about this proto.
```
LalitMaganti: nit: remove beginner

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -323,6 +303,86 @@ message TraceMetricV2Spec {
...
+  // have the same `query`, `dimensions` and `dimension_uniqueness` as well.
+  //
+  // This field exists as an escape hatch for cases where templates for some
+  // reason do not work for you.
+  optional string bundle_id = 7;
+
+  // Specification for a interned dimension associated with this metric.
+  // For a given dimension, there can only be one interned dimension attached to
+  // it.
+  //
```
LalitMaganti, line 336: I would again disclaim this is an advanced feature in case they didn't see the banner above.

protos/perfetto/trace_summary/v2_metric.proto
LalitMaganti: Very much not a fan of that implicitness. I think it's important enough to be a separate field.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -402,6 +447,147 @@ base::Status VerifyBundleHasConsistentSpecs(
...
+    const protos::pbzero::TraceMetricV2Spec::Decoder& spec) {
+  ASSIGN_OR_RETURN(auto dimensions, GetDimensions(spec));
+  std::set<std::string> interned_dimension_keys;
+  for (auto it = spec.interned_dimension_specs(); it; ++it) {
+    protos::pbzero::TraceMetricV2Spec::InternedDimensionSpec::Decoder ms(*it);
+    if (!ms.has_key_column_spec()) {
+      return base::ErrStatus(
+          "key_column_spec must be specified in interned_dimension_specs");
+    }
+    protos::pbzero::TraceMetricV2Spec::InternedDimensionSpec::ColumnSpec::
```
LalitMaganti: Some using declarations would really help readability.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -402,6 +447,147 @@ base::Status VerifyBundleHasConsistentSpecs(
...
+
+base::Status WriteInternedDimensionBundles(
+    TraceProcessor* processor,
+    const protos::pbzero::TraceMetricV2Spec::Decoder& spec,
+    const std::vector<std::string>& interned_dimension_queries,
+    protos::pbzero::TraceMetricV2Bundle* bundle) {
+  RETURN_IF_ERROR(ValidateInternedDimensionSpecs(spec));
+  size_t interned_dimension_idx = 0;
+  for (auto it = spec.interned_dimension_specs(); it; ++it) {
+    protos::pbzero::TraceMetricV2Spec::InternedDimensionSpec::Decoder ms(*it);
```
LalitMaganti: As above.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -402,6 +447,147 @@ base::Status VerifyBundleHasConsistentSpecs(
...
+    protos::pbzero::TraceMetricV2Spec::InternedDimensionSpec::ColumnSpec::
+        Decoder key_col_spec(ms.key_column_spec());
+    std::string key_column_name = key_col_spec.name().ToStdString();
+    auto* interned_dimension_bundle = bundle->add_interned_dimension_bundles();
+
+    std::string sql = interned_dimension_queries[interned_dimension_idx++];
```
LalitMaganti: Why are we making a copy?

src/trace_processor/trace_summary/summary.cc
```diff
@@ -402,6 +447,147 @@ base::Status VerifyBundleHasConsistentSpecs(
...
+    const protos::pbzero::TraceMetricV2Spec::Decoder& spec,
+    const std::vector<std::string>& interned_dimension_queries,
+    protos::pbzero::TraceMetricV2Bundle* bundle) {
+  RETURN_IF_ERROR(ValidateInternedDimensionSpecs(spec));
+  size_t interned_dimension_idx = 0;
+  for (auto it = spec.interned_dimension_specs(); it; ++it) {
+    protos::pbzero::TraceMetricV2Spec::InternedDimensionSpec::Decoder ms(*it);
+    protos::pbzero::TraceMetricV2Spec::InternedDimensionSpec::ColumnSpec::
+        Decoder key_col_spec(ms.key_column_spec());
+    std::string key_column_name = key_col_spec.name().ToStdString();
```
LalitMaganti: You can use base::StringView as you're only using this for logging. Use %.*s and pass the size to make it more

src/trace_processor/trace_summary/summary.cc
```diff
@@ -402,6 +447,147 @@ base::Status VerifyBundleHasConsistentSpecs(
...
+      column_infos.emplace_back(
+          *column_index,
+          static_cast<protos::pbzero::TraceMetricV2Spec::DimensionType>(
+              col_spec.type()));
+    }
+
+    base::FlatHashMap<uint64_t, bool> seen_keys;
+    while (query_it.Next()) {
+      const auto& key_val = query_it.Get(column_infos[0].first);
+      base::FnvHasher key_hasher;
```
LalitMaganti: Please extract this into a custom hasher on flatHahsMap. Right now you're double hashing.

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Added.

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Done

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Done

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: Done

src/trace_processor/trace_summary/summary.cc
elrakadm: done

src/trace_processor/trace_summary/summary.cc
elrakadm: updated it everywhere

src/trace_processor/trace_summary/summary.cc
elrakadm: The same approach is used for dimension uniqueness because sqlvalue doesn't have a hash function, should I implement one ?

src/trace_processor/trace_summary/summary.cc
elrakadm: removed

src/trace_processor/trace_summary/summary.cc
elrakadm: done

src/trace_processor/trace_summary/summary.cc
LalitMaganti: You should write a hasher and pass it as third argument to base::flathashmap. I think that should work though I'm not 100%

LalitMaganti: Ah no it won't work because sqlvalue does not own it's strings. 

Please at least extract this into a function.

elrakadm: Done

protos/perfetto/trace_summary/v2_metric.proto
```diff
@@ -533,4 +602,74 @@ message TraceMetricV2Bundle {
...
+  message InternedDimensionBundle {
+    // A row of interned dimension, with values for columns specified in
+    // |column_specs|.
+    //
+    // Next id: 3
+    message InternedDimensionRow {
+      // The key value in the row.
+      optional TraceMetricV2Bundle.Row.Dimension key_dimension_value = 1;
+      // The data values in the row.
+      repeated TraceMetricV2Bundle.Row.Dimension data_dimension_values = 2;
```
LalitMaganti: nit: I would change this to "interned_dimension_values"

LalitMaganti (approved):

protos/perfetto/trace_summary/v2_metric.proto
elrakadm: done

Issue #1338, opened 2025-04-22
Feature Request: add callstack support for TrackEvent


anakryiko:
It would be extremely useful to have callstacks associated with custom TrackEvents. At least INSTANT kind makes sense, but SLICE_BEGIN/SLICE_END would be great to support as well, if possible (with some ability in UI to decide whether to visualize beginning or ending callstack).

Currently we work around by generating artificial PerfSample events with the exact same thread association and time stamp as TrackEvent, but this duplicates number of tracks in UI and is quite messy and inconvenient.

Is something like this on roadmap? And if not, would it be possible to add it? Thanks!

LalitMaganti:
Can you explain why it's "messy and inconvenient" that they're on different tracks for the visualization? I feel that's a pretty reasonable way for them to be presented given the semantic of call stack Vs event is quite different. There's a lot of features in the UI which assume that something is slice like or call stack like and those are mutually exclusive generally. Putting both of the same track at the same time would require adding a lot of complexity to make things work correctly. 

However in the recording side as a convenience thing, makes total total sense. Honestly the perf sample protos are not stable API but I've long wanted to actually add a proper stable interface for call stack samples in our protos. Patches for doing this welcome.

anakryiko:
@LalitMaganti I have a BPF-based profiling/tracing tool that produces custom Perfetto traces. I track scheduling events and have timer-based callstack sampling, you can think about this as wall-clock profiler, but there is actually more to this (just in case you'll suggest using built-in scheduling protos, it's not quite what I need). And so I usually have a combination of slices representing some thread running on-cpu, with randomly sprinkled instant events representing timer events. And there is a bunch of other stuff that's not too relevant to go into, the main point being that thread track is the focus point, and it is a mix of slices and instants. And some of the instants semantically have stack traces associated with them. Right now I get one extra track just of "PerfEvents" (they are not really perf events, I get stack traces with BPF), and then one extra track for entire process. So this is both visually more overwhelming and inconvenient, if I want to stay within thread track.

Overall, events do have stack traces (callstacks) "attached" to them. But Perfetto currently forces them to be a separate kind of events on their separate tracks. Which seems like an unnecessary restriction.

[...]

PR #3299, opened 2025-10-15
tp: add support for attaching callstacks to track events

LalitMaganti:
This CL implements support for people to attach a callstack directly to a track event.
This is particularly valuable for people converting to Perfetto who may want to
annotate the begin/end event of a slice (or an instant) with that info.

We show the callstack in two ways:
1) The callstack is direclty shown when you click on the slice
2) We have an area selection aggregation where we aggregate all the track event
   callstacks into a flamegraph

Fixes: https://github.com/google/perfetto/issues/1338


anakryiko:
@LalitMaganti thanks a lot for adding this, can't wait till this lands and I can make use of this.

Two questions/requests.

1) For inline callstack frame you have line_number, which is great. But for interned call stack frame it's missing. Can we add line number there as well (and show it, of course, in UI)?

2) Just to make sure, this is not limited only to SLICE_BEGIN and SLICE_END events, INSTANT events also will get their callstack association, is that right? If not, can we please allow INSTANTs to also have a callstack? Having ability to annotate instants with callstack will be super useful. (E.g., think about time slice representing thread running on CPU, and somewhere in the middle of that that thread notifies/wakes up another thread. It's important to be able to show where that notification happened with callstack).

src/trace_processor/importers/proto/track_event_event_importer.h
sashwinbalaji, line 1: maybe separate cl but it has grown too much for a header file

src/trace_processor/importers/proto/stack_profile_sequence_state.cc
```diff
@@ -76,7 +76,7 @@ VirtualMemoryMapping* StackProfileSequenceState::FindOrInsertMapping(
 VirtualMemoryMapping* StackProfileSequenceState::FindOrInsertMappingImpl(
     std::optional<UniquePid> upid,
     uint64_t iid) {
-  if (auto ptr = cached_mappings_.Find({upid, iid}); ptr) {
+  if (auto* ptr = cached_mappings_.Find({upid, iid}); ptr) {
```
sashwinbalaji: nit: this auto is confusing as multiple dereferencing so like other functions VirtualMemoryMapping* would be better?

src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -1437,6 +1447,58 @@ class TrackEventEventImporter {
...
+      protos::pbzero::TrackEvent::Callstack::Decoder callstack(
+          event_.callstack());
+      DummyMemoryMapping* dummy_mapping =
+          parser_->GetOrCreateInlineCallstackDummyMapping();
+
+      std::optional<CallsiteId> callsite_id;
+      uint32_t depth = 0;
+      for (auto frame_it = callstack.frames(); frame_it; ++frame_it, ++depth) {
+        protos::pbzero::TrackEvent::Callstack::Frame::Decoder frame(*frame_it);
+        FrameId frame_id = dummy_mapping->InternDummyFrame(
```
sashwinbalaji, line 1496: I think even for inline we are not using the line_number ?

LalitMaganti:
> For inline callstack frame you have line_number, which is great. But for interned call stack frame it's missing. Can we add line number there as well (and show it, of course, in UI)?

Yes will add that in as well, was an oversight.

> Just to make sure, this is not limited only to SLICE_BEGIN and SLICE_END events, INSTANT events also will get their callstack association, is that right? 

Yes of course instants are supported too.

src/trace_processor/importers/proto/track_event_event_importer.h
LalitMaganti: It's actually intentional it's in a header. It's a header which is only included in a single cc file and exists purely for splitting without causing perf problems becuase inlinining doesn't happen in the right places.

LalitMaganti:
PTAL.

anakryiko:
protobuf side looks great, thanks! I also appreciate you adding a good deal of documentation!

src/trace_processor/importers/proto/stack_profile_sequence_state.cc
```diff
@@ -76,7 +77,8 @@ VirtualMemoryMapping* StackProfileSequenceState::FindOrInsertMapping(
 VirtualMemoryMapping* StackProfileSequenceState::FindOrInsertMappingImpl(
     std::optional<UniquePid> upid,
     uint64_t iid) {
-  if (auto* ptr = cached_mappings_.Find({upid, iid}); ptr) {
+  VirtualMemoryMapping** ptr = cached_mappings_.Find({upid, iid});
```
sashwinbalaji: nit^2: the older initialisation within if was better though with VirtualMemoryMapping**

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -14953,8 +15048,6 @@ message TrackEvent {
   }
 
   optional LegacyEvent legacy_event = 6;
-
-  reserved 51;
```
sashwinbalaji, line 14968: if not reused, let's keep this reserved line.

PR #3301, opened 2025-10-15
tp: procfs process trees: set main thread name if "cmdline_is_comm"

rsavitski:
This came up for traces containing perf samples + procfs trees, but no sched_switch to supply thread names. The result is that the grouped kernel thread tracks end up mostly anonymous (since the procfs scraping is only supplying the process names). Additionally, a subset of those threads does end up with a name due to the trace including "task_rename" ftrace events.

On modern traces, we actually know if the "cmdline" field in the trace was set from /proc/pid/comm, so we can use that information to name the main thread as well.


rsavitski:
> should we also support this for GenericKernelProcessTree?

I don't believe we have a concept of kernel threads in the "generic" (i.e. OS-agnostic) protos.

PR #3302, opened 2025-10-15
tp: add support for custom weighted flamegraph metrics for track event

LalitMaganti:


LalitMaganti:
@anakryiko still WIP but PTAL if you can to make sure proto looks ok.

protos/perfetto/trace/track_event/track_event.proto
```diff
@@ -322,6 +322,37 @@ message TrackEvent {
...
+      string metric_name = 3;
+      // The spec for a metric (see InternedCallstackMetricSpec). Use this when
+      // you want to attach a unit to the weight or you care about trace size
+      // and so want to intern the metric spec.
+      uint64 metric_spec_iid = 4;
+    }
+    // The value of the metric to associate with the callstack.
+    optional double value = 1;
+  }
+  repeated CallstackMetric callstack_metrics = 57;
```
anakryiko, line 354: is this a set of all possible metrics that user will be able to pick for that callstack? I.e., I can have "CPU time" metric set to 5 ms, but also "CPU cycles" set to 1000, and then in the UI I'll be able to select either CPU time or CPU cycles? Or what's the idea?

Generally speaking, it's very useful to have some sort of "default weight" association (e.g., for my case that would be exactly CPU time), but in UI to give users ability to, say, use one of debug annotations as user-provided weight for that call stack.

But let me know how you think about this.

LalitMaganti: I'm confused: this field is available on every event so you just tell us "for this slice, I want to set the metric X to value Y". There's never a "default" in the way you say it.

anakryiko: hm... doesn't this just duplicate DebugAnnotations then? Technically, any integer or float value in DebugAnnotations can serve as such callstack metric, no? Do we need this duplication then?

LalitMaganti: I mean I guess we could do that. But do you not think it would be a bit weird to automatically the flamegraph metrics populate based on that? Not every integer value arg will automatically flamegraph aggregatable. For example, people put timestamps or pointer values in args. Wouldn't it be confusing to see that in the flamegraph when it doesn't make sense?

LalitMaganti: Oh also: how do we deal with nested/repeated fields? Would we support taking an arbitrary key and using it. Just feels very strange imo.

anakryiko: Yes, absolutely, automatically picking a random debug annotation field would be super weird. So in that sense having control over default associated value/weight is best. The `repeated` part of `repeated CallstackMetric` gave me an initial pause, though. But I guess it all depends on how all this is treated in the UI. Would you pick the first CallstackMetric as *the metric* to use initially for flame graph? And then allow users to switch it for their flame graph rendering? How did you think about that?

Here's my slight opposition to the idea of having CallstackMetric set of values in addition to DebugAnnotations. Those callstack matrics could be useful by themselves without callstack in pivot/table views. So I don't see them as callstack-specific. They are useful and usable with callstack, but usually stuff like that is useful on its own as well. So perhaps allowing both CallstackMetric and DebugAnnotations to be used in table/pivot views would be an answer.

But just in the spirit of keeping it simple, I think it would be fine to have (optional, singular) unit-less weight value  associated with the callstack, and if it's not specified, assume 1. But in addition to that, let user pick any of aggregatable (integer or float) field out of DebugAnnotation?

Nested key sounds useful, in principle, but I haven't had any need for that so far. So depending on how much trouble it is to support that, I'd allow it, but wouldn't stress too much about it.

LalitMaganti: > Would you pick the first CallstackMetric as the metric to use initially for flame graph? And then allow users to switch it for their flame graph rendering?

Yes this is basically exactly what I was thinking to do. 

> Here's my slight opposition to the idea of having CallstackMetric set of values in addition to DebugAnnotations...

Yeah I can see this. See below for thoughts

> But just in the spirit of keeping it simple, I think it would be fine to have (optional, singular) unit-less weight value associated with the callstack, and if it's not specified, assume 1. But in addition to that, let user pick any of aggregatable (integer or float) field out of DebugAnnotation?

I'm still skeptical of allowing picking *any" aggregatable thing. And still leaves a problem: how do people specify the unit?

Taking a step back, maybe the problem is more than DebugAnnotation is not giving us enough information to make these decisions. That is, maybe the correct thing to do here is enhance DebugAnnotation with the ability to indicate "this key is aggregatable" or "this is the unit of this value". That way throughout the Perfetto UI (pivot table, breakdown table, flamegraph) we can prioritise showing those to you (I think we'll have the ability to choose other keys but it will be a couple more clicks, hidden in some menu).

anakryiko: > I'm still skeptical of allowing picking *any" aggregatable thing. And still leaves a problem: how do people specify the unit?
> 
> Taking a step back, maybe the problem is more than DebugAnnotation is not giving us enough information to make these decisions. That is, maybe the correct thing to do here is enhance DebugAnnotation with the ability to indicate "this key is aggregatable" or "this is the unit of this value". That way throughout the Perfetto UI (pivot table, breakdown table, flamegraph) we can prioritise showing those to you (I think we'll have the ability to choose other keys but it will be a couple more clicks, hidden in some menu).

That's a great idea, IMO! 

There is still the question of default weight associated with call stack, how do you think about that? Is that a dedicated value, or we somehow should specify default DebugAnnotation key from which weight will be pulled?

It's a bit of a messy problem for sure :)

LalitMaganti: I'm going to reflect a bit more on this before I come to a decision on exactly what I want to do here. May take a bit of time before I come back to this.

LalitMaganti: So I've been thinking on this on and off over the last several months and after discussing with folks in the team about the plans for table viewer, I'm reasonably convinced that we should keep 1:1 consistency with what we do for the table viewer and what we do for flamegraph.

That is, in some sense, a flamegraph is just an alternative visualization of the pivot table representation of data. Given the pivot table lets you sum/min/max/aggregate arbitrary debug arguments even when it might not be so meaningful, I think we should not try and make flamegraph any more special. And if we do decide to be more picky, we'll do the same for both.

Given this, I'll look into rearchitecting this patch to work for any debug arguments in the selection. I also plan on redoing the flaegraph computation code for performance and this will be part of that rework.

Will take me more time to get back to this properly but at least now I have an understanding and a plan.

anakryiko: @LalitMaganti sorry for late reply, I've been out on vacation. I think the above reasoning makes total sense! So I'm looking forward to this be implemented!

On a tangential note, though, I do think that having (optional) weight associated with the stack (so a double/float field next to callstack_iid, not really being part of the stack itself, of course) is a good addition as well and this will serve as an override to a default (and implicit) value of weight 1 used by flamegraph. Is this a reasonable thing to ask? And in UI users will be able to redefine what sort of weight they'd like to use for each sample using debug annotations, of course.

LalitMaganti: Potentially though if we're letting people choose any debug annotation, why is it not just the "default key which should be used?.

anakryiko: Yeah, that would work in most cases (and certainly it's always possible to add dedicated "stack_weight" debug annotation just for this, even though it might be unnecessarily polluting debug section), but imagine that for some TrackEvents the key is one of the built-ins (most probably "dur"). It's not debug annotation, so now you'd need to define in proto how to specify either debug or built-in key. It's just a bit messy.

On the other hand, having a unitless (and totally optional) "weight" field is so temptingly simple a solution.

Up to you, of course. I'd go with weight in addition to allowing user-selectable (but in UI) debug annotation. (and if something like https://github.com/google/perfetto/issues/4229 is ever available, that would hopefully be usable to select stack weight as well). 

BTW, I'm not just making this up as "nice to have", we at Meta do have internal system that allows to do stuff like this, and we actively utilize it specifically when visualizing collections of stack traces. We actually have ability to combine non-stack and stack fields in specific order and they will together form flamegraph view. E.g., process name + stack, when visualized as flamegraph, will group a collection of stacks by process name, and then within each process we'll have nested flamegraph of stack traces. Request name + stack traces is another common combination. It's an extremely useful and powerful capability.

LalitMaganti: Fair enough will think about this. I agree having something simple for stack weight makes sense without the overhead of debug annotations.

> We actually have ability to combine non-stack and stack fields in specific order and they will together form flamegraph view. E.g., process name + stack, when visualized as flamegraph, will group a collection of stacks by process name, and then within each process we'll have nested flamegraph of stack traces. Request name + stack traces is another common combination. It's an extremely useful and powerful capability.

Yeah I totally want to do this in flamegraph but we're going to tie it to pivot table: the UX is a lot easier if *both* support it rather than just flamegraph. If you've ever used Windows Performance Analyze, it allows using entire stack traces as pivots: at some point we want to do that as well. And when we do, that will be the key to also have flamegrpah work in the way you describe.

We would allow process -> stack but also process -> stack -> cpu or stack -> cpu -> debugn annotation. Basically allow any combination in any order with "stack" as a "a bunch of frames concatenated together.

LalitMaganti:
Closed in favor of https://github.com/google/perfetto/pull/6752

protos/perfetto/trace/track_event/track_event.proto
LalitMaganti: Please see https://github.com/google/perfetto/pull/6752 for the weight part of things. Breakdowns will follow at some point.

PR #3303, opened 2025-10-15
Add callstack support to trace redactor v2

edgararriagag:
Add support for perf samples in trace redactor which in turn allows for call stack collection of all the hardware and sw counters currently supported with linux.perf collection


edgararriagag:
This is the same PR from https://github.com/google/perfetto/pull/2927 I just recreated it cleanly to make sure it only shows the current diff

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,133 @@
...
+
+  // Check that all the trace samples in the trace correspond to target process.
+  ASSERT_TRUE(perf_samples_for_target_pid == trace_perf_samples);
+
+  ASSERT_OK(rows.Status());
+}
+
+// TODO (edgararriagag): There is currently a bug in the timeline where some
+// slices are dropped which affects ability of redactor to pass this test. Once
+// b/446994151 is fixed, we should reenable this test.
```
edgararriagag, line 117: @LalitMaganti is it ok if we link buganizer issues in the TODO, or should I leave that part out and just keep the TODO description to reenable this test.

LalitMaganti: Yeah it's fine to link it

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,133 @@
...
+
+  std::unique_ptr<trace_processor::TraceProcessor> trace_processor_unredacted_;
+  std::unique_ptr<trace_processor::TraceProcessor> trace_processor_redacted_;
+};
+
+TEST_F(PrunePerfEventsIntegrationTest, OnlyKeepsTargetProcessPerfSamples) {
+  // This query retrieves the total number of perf samples for target process
+  // in redacted trace
+  auto query =
+      " SELECT COUNT(*) FROM perf_sample "
```
vaage: Were these spaces intentional? I see them in all the queries.

src/trace_redaction/trace_redaction_integration_fixture.cc
```diff
@@ -68,4 +68,26 @@ base::StatusOr<std::string> TraceRedactionIntegrationFixure::ReadRawTrace(
   return base::ErrStatus("Failed to read %s", path.c_str());
 }
 
+base::Status TraceRedactionIntegrationFixure::LoadTrace(
```
vaage: Why did you need another load function? Why didn't the other load functions work?

src/trace_redaction/trace_redaction_integration_fixture.h
```diff
@@ -40,6 +41,12 @@ class TraceRedactionIntegrationFixure {
 
   base::StatusOr<std::string> LoadRedacted() const;
 
+  base::Status LoadTrace(
+      std::string trace_path,
+      trace_processor::TraceProcessor* trace_processor) const;
+
+  std::string GetSourceTrace();
```
vaage: Why do you need this when the test already knows the source trace path? Whenever possible, I would prefer to have one source of truth.

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,133 @@
...
+  ASSERT_TRUE(rows.Next());
+  int64_t trace_perf_samples = rows.Get(0).AsLong();
+
+  // Check that all the trace samples in the trace correspond to target process.
+  ASSERT_TRUE(perf_samples_for_target_pid == trace_perf_samples);
+
+  ASSERT_OK(rows.Status());
+}
+
+// TODO (edgararriagag): There is currently a bug in the timeline where some
```
vaage, line 115: Just to double check, what pids are the slices using? If they are outside of the target process, they're assigned a new pid. If that's the case, its not a bug with the timeline - it's working as expected. It would actually require new functionality.

src/trace_redaction/trace_redactor.cc
```diff
@@ -201,6 +202,12 @@ std::unique_ptr<TraceRedactor> TraceRedactor::CreateInstance(
 
   redactor->emplace_transform<PrunePackageList>();
 
+  {
+    // Remove callstacks that don't belong to target package.
```
vaage: Can you update your command to first example what information needs to be removed/protected, "We need to retain callstack samples so that..." and follow it up with a "we use ScrubProcessStats to <summary of how it works>" and lastly "it needs to come after/before... because..."

This makes sure we don't move primitives around and break things.

src/trace_redaction/trace_redaction_integration_fixture.cc
edgararriagag: It makes integration tests simpler, a lot of them require using the trace processor, the other load functions load raw bytes plus they are private though making them public is not the best utility method because they provide raw bytes, not a ready to query trace, this function adds the trace processor component and gives you the ready to query trace which vastly simplifies tests

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: No, the initial space was unintentional and unnecessary, removed.

src/trace_redaction/trace_redaction_integration_fixture.h
edgararriagag: this is because the field is private field, exposing it allows us to use it for the tests which allows us to read the unredacted trace via:
 "status = LoadTrace(GetSourceTrace(), trace_processor_unredacted_.get());"
previously the tests were just reading the redacted traces, the source trace was just used internally for redaction which is why it was not exposed to implementors.

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: they are using the same pid as the target process, the target process does exist and matches that pid because the trace we use as test/data is controlled to contain that process

src/trace_redaction/trace_redaction_integration_fixture.h
vaage: The test already knows what trace it is using `kTrace`; in what case does a test need to get the trace file from the fixture when it already has the trace file as a constant? No other test ran into the issue you are describing.

src/trace_redaction/trace_redaction_integration_fixture.cc
vaage: Unless it's impossible, please use the existing patterns that are used across the other tests.

src/trace_redaction/trace_redaction_integration_fixture.h
edgararriagag: kTrace is just the filename, when SetSourceTrace is called it is actually altered to a test specific path:
```
void TraceRedactionIntegrationFixure::SetSourceTrace(
    std::string_view source_file) {
  src_trace_ = base::GetTestDataPath(std::string(source_file));
}
```
We store this data and it sounds reasonable to want to query it. If we manually do the same operation to retrieve the test path, then if the base fixture class ever gets refactored and decides to load the trace from a different folder, then our tests would break, this way we make the test less prone to breaking when updating the testing infra. I believe exposing this as a method seems reasonable, unless there is a strong argument against it.

>No other test ran into the issue you are describing.

I believe this is because none of them ever compared baseline vs redacted trace at all, which is a very useful thing to do

src/trace_redaction/trace_redaction_integration_fixture.cc
edgararriagag: It is not possible, I really think we should keep this function because what I am adding here is the ability for other tests to load the unredacted trace as well as the redacted trace with trace processor because I expect more and more integration tests will be proven with TP queries results and comparing them between the two traces just like I did here

src/trace_redaction/trace_redactor.cc
edgararriagag: I added a comment here that explains the dependencies, I think it should be descriptive enough, ptal and let me know if its good

src/trace_redaction/trace_redaction_integration_fixture.h
vaage: Other tests compare the original trace against the redacted trace, that's what `LoadOriginal()` does. I do not see anything here than cannot be done with the existing patterns.

src/trace_redaction/trace_redaction_integration_fixture.cc
vaage: Other tests compare the original trace against the redacted trace, that's what `LoadOriginal()` does. I do not see anything here than cannot be done with the existing patterns.

src/trace_redaction/trace_redaction_integration_fixture.h
edgararriagag: Other tests rely on parsing the proto, they do not load the original with trace processor which is something novel that this test does. Based on our offline discussion. Given this method is currently only used in our test, I moved it to this test.

Once we need to reuse the method, we can then check if we need to bring it to become a common method of the fixture.

Note: I had to modify ReadRawTrace to be protected in order to make this change, but I assume that should be fine.

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+
+constexpr auto kPackageName = "com.example.sampleapp";
+constexpr auto kPid = 25131;
+}  // namespace
+
+class PrunePerfEventsIntegrationTest
+    : public testing::Test,
+      protected TraceRedactionIntegrationFixure {
+ protected:
+  base::StatusOr<std::unique_ptr<trace_processor::TraceProcessor>> LoadTrace(
```
vaage: I must be missing something here, how is you code any different than?

```c++
base::StatusOr<std::unique_ptr<trace_processor::TraceProcessor>> LoadTrace(const std::string& content) {
    std::unique_ptr<trace_processor::TraceProcessor> trace_processor = trace_processor::TraceProcessor::CreateInstance(trace_processor::Config());

    auto read_buffer = std::make_unique<uint8_t[]>(content.size());
    memcpy(read_buffer.get(), content.data(), content.size());

    RETURN_IF_ERROR(trace_processor->Parse(std::move(read_buffer), raw_trace->size()));
    RETURN_IF_ERROR(trace_processor->NotifyEndOfFile());

    return std::move(trace_processor);
}
```

```c++
auto originalTrace = LoadOriginal();
auto originalTraceProcessor = LoadTrace(originalTrace);

auto redactedTrace = LoadRedacted()
auto redactedTraceProcess = LoadTrace(redactedTrace);
```

In general, the project relies on keeping to the same set of simple patterns. This goes with the main code and the tests.

vaage (changes requested):

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: The difference is that LoadOriginal/LoadRedacted only load the trace as a proto, whereas `LoadTrace` loads it as a queryable trace with the trace processor, allowing you to make queries on the result directly.

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+      "JOIN process ON process.upid = thread.upid "
+      "GROUP BY pid "
+      "HAVING pid = " +
+      std::to_string(kPid);
+
+  auto rows = trace_processor_unredacted_->ExecuteQuery(query);
+  ASSERT_TRUE(rows.Next());
+  int64_t unredacted_target_process_samples = rows.Get(0).AsLong();
+
+  rows = trace_processor_redacted_->ExecuteQuery(query);
```
vaage: Don't reuse values.

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+       DISABLED_TargetProcessPerfSamplesMatchesUnredacted) {
+  auto query =
+      "SELECT COUNT(*) FROM perf_sample "
+      "JOIN thread ON thread.utid = perf_sample.utid "
+      "JOIN process ON process.upid = thread.upid "
+      "GROUP BY pid "
+      "HAVING pid = " +
+      std::to_string(kPid);
+
+  auto rows = trace_processor_unredacted_->ExecuteQuery(query);
```
vaage: All the other code uses "original" please use "original" instead of "unredacted".

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+
+  ASSERT_OK(rows.Status());
+}
+
+// TODO (edgararriagag): There is currently a bug in the timeline where some
+// slices are dropped which affects ability of redactor to pass this test. Once
+// b/446994151 is fixed, we should reenable this test.
+TEST_F(PrunePerfEventsIntegrationTest,
+       DISABLED_TargetProcessPerfSamplesMatchesUnredacted) {
+  auto query =
```
vaage: Make `const`.

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+      "SELECT COUNT(*) FROM perf_sample "
+      "JOIN thread ON thread.utid = perf_sample.utid "
+      "JOIN process ON process.upid = thread.upid "
+      "GROUP BY pid "
+      "HAVING pid = " +
+      std::to_string(kPid);
+
+  auto rows = trace_processor_unredacted_->ExecuteQuery(query);
+  ASSERT_TRUE(rows.Next());
+  int64_t unredacted_target_process_samples = rows.Get(0).AsLong();
```
vaage: const

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+
+  rows = trace_processor_redacted_->ExecuteQuery(query);
+  ASSERT_TRUE(rows.Next());
+  int64_t redacted_target_process_samples = rows.Get(0).AsLong();
```
vaage: const

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  std::unique_ptr<trace_processor::TraceProcessor> trace_processor_unredacted_;
+  std::unique_ptr<trace_processor::TraceProcessor> trace_processor_redacted_;
+};
+
+TEST_F(PrunePerfEventsIntegrationTest, OnlyKeepsTargetProcessPerfSamples) {
+  // This query retrieves the total number of perf samples for target process
+  // in redacted trace
+  auto query =
```
vaage: const

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+      std::to_string(kPid);
+
+  auto rows = trace_processor_redacted_->ExecuteQuery(query);
+  ASSERT_TRUE(rows.Next());
+  int64_t perf_samples_for_target_pid = rows.Get(0).AsLong();
+  ASSERT_TRUE(perf_samples_for_target_pid > 0);
+
+  // This query retrieves the total number of perf samples for all processes
+  // in redacted trace
+  query =
```
vaage: Don't reuse variables.

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+      "SELECT COUNT(*) FROM perf_sample "
+      "JOIN thread ON thread.utid = perf_sample.utid "
+      "JOIN process ON process.upid = thread.upid";
+  rows = trace_processor_redacted_->ExecuteQuery(query);
```
vaage: Don't reuse variable name.

src/trace_redaction/prune_perf_events_integrationtest.cc
```diff
@@ -0,0 +1,139 @@
...
+  ASSERT_TRUE(rows.Next());
+  int64_t trace_perf_samples = rows.Get(0).AsLong();
```
vaage: const

src/trace_redaction/prune_perf_events_integrationtest.cc
vaage: What is the query doing? Please make the name more meaningful and add a comment.

src/trace_redaction/prune_perf_events_integrationtest.cc
vaage: What is the query doing? Please make the name more meaningful and add a comment.

src/trace_redaction/prune_perf_events_integrationtest.cc
vaage: What is the query doing? Please make the name more meaningful and add a comment.

src/trace_redaction/trace_redaction_integration_fixture.h
vaage, line 1: Nothing in this file needs to change. You can do this CL without updating the fixture. I know this because I pulled the change and was able to do it myself. There will be some duplicate code between two classes, but that is okay. It is not a common enough pattern that it needs to be addressed in this CL. If it is a big enough problem it MUST be addressed in its own CL so it is easier to find in the git log.

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: I reworked the comment and the variable name to make its intention clearer

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/prune_perf_events_integrationtest.cc
edgararriagag: Done

src/trace_redaction/trace_redaction_integration_fixture.h
edgararriagag: >If it is a big enough problem it MUST be addressed in its own CL so it is easier to find in the git log.

Yes, I believe this must be done. Took your advice and went ahead and refactored it out into this CL https://github.com/google/perfetto/pull/3354

PR #3306, opened 2025-10-16
ui: Allow triggering popups on right click (oncontextmenu)

stevegolton:
- Add attribute `isContextMenu` to Popup-like widgets which shows the popup on right click instead of left click.
- Also add attribute `positionAtCursor` which positions the popup at the cursor position on the trigger instead of on the trigger element itself. This is a more natural positioning when we have large trigger elements, as is more common with right click scenarios.


ui/src/widgets/popup.ts
```diff
@@ -103,6 +103,13 @@ export interface PopupAttrs {
   // content that should not be constrained by a maximum width.
   // Defaults to false.
   fitContent?: boolean;
+  // If true, the popup will appear when the trigger is on right clicked rather
+  // than when it's left clicked.
+  contextMenu?: boolean;
```
aMayzner: Is it a good name for this parameter? Maybe "isContextMenu" or "triggeredonRightClick"

stevegolton: I like `isContextMenu`. I'd avoid using 'right click' terminology as it's not necessarily right click on all platforms, which is why the underlying DOM event that gets fired is called 'contextmenu'.

PR #3308, opened 2025-10-16
exp: Cleanup prevNodes

aMayzner:
Previously all of the nodes had a property of `prevNodes: QueryNode[]`, which was confusing, as most nodes are heavily typed as to what is the number of nodes they can have as parents. This commit makes it explicit by creating 3 query nodes types, Source node (no parent nodes), ModificationNode (one parent) and MultiSourceNode (valid only on 2+ parents)

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/dev/add_columns_node.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/dev/add_columns_node.ts
@@ -28,13 +28,13 @@ import {FilterDefinition} from '../../../../../components/widgets/data_grid/comm
 import {MultiselectInput} from '../../../../../widgets/multiselect_input';
 
 export interface AddColumnsNodeState extends QueryNodeState {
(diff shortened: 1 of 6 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/dev/add_columns_node.ts
```diff
@@ -27,27 +28,31 @@ import {FilterDefinition} from '../../../../../components/widgets/data_grid/comm
...
-  readonly prevNodes?: QueryNode[];
+  readonly prevNode: QueryNode;
   nextNodes: QueryNode[];
   readonly state: AddColumnsNodeState;
 
   constructor(state: AddColumnsNodeState) {
     this.nodeId = nextNodeId();
     this.state = state;
-    this.prevNodes = state.prevNodes;
+    if (state.prevNode === undefined) {
```
stevegolton: Can this not be enforced by the type system?

PR #3314, opened 2025-10-16
ui: Import d3 library

zezeozue:
In preparation to add d3 chart widgets/bigtrace ui


primiano:
What's the plan / what are we going to use?

can we have some consistent charting story? I thought we settled on vega-lite a while ago.
What's the plan going forward? which library?
D3 itself is not a charting library, so i guess yo plan to use something else that uses D3?

zezeozue:
The plan is to use d3 directly. Vega-lite has some limitations for building very interactive charts or custom charts like box plots. If you have chart A and chart B and you select a region of chart A and want to filter chart B there's not a lot of flexibility. 

I think vega-lite was the right choice for the simple bar and histogram charts we added, but I intend to land better charts that use d3 directly, iow, I plan to use d3 directly. The immediate goal is for the bigtrace ui in perfetto (already have some prototypes: https://github.com/google/perfetto/compare/main...dev/zezeozue/perfette), but hopefully it can be usable as first class chart widgets for perfetto ui itself.

PR #3318, opened 2025-10-16
ui: Grid QoL improvements

stevegolton:
Grid QoL improvements:
- Stable, resizable, auto-sizable column widths with sensible defaults.
  - Columns can be user-resized by dragging on the column handle.
  - A sensible initial size is decided as soon as we have some data based on measuring the cells in place.
  - Column widths are stable and persist over various operations such as filtering, sorting, et al.
- Right click context menus in data cells rather than triple dot menu saving valuable real estate and avoiding awkward gaps in the layout.
- Header cell buttons are now removed entirely from the layout when hidden, saving valuable real estate.
- Virtual scrolling and dynamic data loading now supported at the Grid level.
- The old VirtualTable widget has been removed in favor of Grid.
- PivotTable now uses virtual scrolling to improve performance.
- DataGrid now uses virtual scrolling also to improve performance, but also in order to support pivoting in the future as the grouping UX would be very awkward with pagination.


LalitMaganti (approved):

LalitMaganti (approved):
Still LGTM!

PR #3327, opened 2025-10-17
Fix for redactor failure due to missing PID in perf sample

edgararriagag:
Redactor started failing when it encountered this perf sample.

```
packet {
  sequence_flags: 2
  timestamp: 252557484253
  timestamp_clock_id: 6
  perf_sample {
    cpu: 0
    kernel_records_lost: 12
  }
  trusted_uid: 9999
  trusted_packet_sequence_id: 6
  trusted_pid: 11706
}
```

This code modifies redactor so that it skips the sample instead of making redactor bail out and fail.


edgararriagag:
This PR somehow got corrupted, closing it, will reopen a new one

PR #3328, opened 2025-10-17
Fix for redactor failure due to missing PID in perf sample

edgararriagag:
Redactor started failing when it encountered this perf sample.

packet {
  sequence_flags: 2
  timestamp: 252557484253
  timestamp_clock_id: 6
  perf_sample {
    cpu: 0
    kernel_records_lost: 12
  }
  trusted_uid: 9999
  trusted_packet_sequence_id: 6
  trusted_pid: 11706
}
This code modifies redactor so that it skips the sample instead of making redactor bail out and fail.


src/trace_redaction/prune_perf_events.cc
```diff
@@ -90,8 +90,10 @@ base::Status PrunePerfEvents::OnPerfSample(
     protos::pbzero::TracePacket* message) const {
   protos::pbzero::PerfSample::Decoder decoder(perf_sample_field.as_bytes());
   if (!decoder.has_pid()) {
-    // PID is required to compute belongs-to process relationship
-    return base::ErrStatus("PrunePerfEvents: missing field (PerfSample::kPid)");
+    // PID is required to compute belongs-to process relationship, however,
+    // it is possible for samples with data loss to not have a PID.
+    // Skip those samples and continue redaction.
```
rsavitski: Not all PerfSample packets are actual process samples, some packets indicate events relevant to the profiler ([see here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/trace/profiling/profile_packet.proto;l=290-294;drc=189992a04d80da240e34fdee4d2ce6abd60dd8ff))

So:
* messages with a set pid that isn't the target should be filtered out.
* service messages (not attributed to a pid) should preferably be kept, since we wouldn't want to, for example, hide the data loss information from the trace.

I'm not familiar with how trace_extractor typically deals with this situation, and assume that "allowing all packets without a pid through" is not desirable as that leaves the system fail-open.

The alternative would be to keep the redactor updated whenever we add a new non-sample type of PerfSample. For the existing implementation, this means allowing:
* packets with `kernel_record_lost`
* packets with `producer_event`

rsavitski (changes requested):

src/trace_redaction/prune_perf_events.cc
edgararriagag: Done, based on our offline discussion, I went ahead and skipped those two scenarios from being redacted, the rest of them will cause an error

src/trace_redaction/prune_perf_events.cc
```diff
@@ -90,7 +90,15 @@ base::Status PrunePerfEvents::OnPerfSample(
...
   if (!decoder.has_pid()) {
-    // PID is required to compute belongs-to process relationship
+    // PID is required to compute belongs-to process relationship, however,
+    // it is possible for samples to not have a PID such as data loss
+    // and profiler global information packets. So, we should make sure
+    // any service events (non-pid packets) are retained for the proper working
+    // of the profiler. When new service events are added, we should make sure
+    // to include them here so the redactor accounts for them.
+    if (decoder.has_kernel_records_lost() || decoder.has_producer_event()) {
+      return base::OkStatus();
```
rsavitski: I'm not familiar with the redactor code, but isn't there supposed to be a step that actually appends the packet to the redacted trace? Similar to "proto_util::AppendField(...)" below?

edgararriagag: Good catch, yeah, that was an oversight on my end, I had mistakenly made it drop the event. I updated this, it now keeps the sample and I checked with traceconv that the packet is kept.

rsavitski (approved):

Issue #3329, opened 2025-10-19
Add Rust SDK


dreveman:
We've implemented a Rust SDK for Perfetto at Rivos that could be upstreamed to the Perfetto repo if there's interest in having that be part of the main repo. This SDK doesn't have to be part of the upstream repo but we think that would be preferred for the following reasons:

- Rust is an increasingly popular language and having first-class support for it should be a priority.
- These bindings are implemented on top of the C ABI and use a protozero plugin similar to the C plugin to generate native Rust code for efficient encoding of protos. Having these proto bindings be up to date with latest perfetto release would be ideal.
- The "-sys" crate support both the usage of an installed shared perfetto library and "vendored" sources. The latter rely on amalgamated source for the C shared library, which is easy to support as part of the perfetto release process.

I'll crate a pull request with the bindings for more detailed review but here are some of the features:

- Use the C ABI and protos to generate safe bindings.
- Low level "-sys" crate for raw C ABI bindings.
- Higher level "perfetto" crate provides a safe interface, uses common Rust patterns, and is recommended for normal usage.
- "perfetto-derive" crate provides proc-macros to make tracing of function calls even easier. Just tag a function and entry/exit and arguments passed will be traced if a specific track event category is enabled.
- Public API fully documented, including example usage that is checked at compile time. All unsafe Rust usage is also documented.
- Includes examples and a comprehensive set of unit tests.

The pull request will have files with Rivos copyright header and doesn't include CI integration but that could be adjusted if we decide to merge this.

@primiano @LalitMaganti

LalitMaganti:
Hey! Thanks for the context on the bug and sending us the PR! Really exciting stuff but also a whole new area for us to think about when it comes to long term maintenance of this. 

Primiano and will discuss this together and also with the rest of the team and then come back to you with a comprehensive answer on how we want to deal with.

dreveman:
> Hey! Thanks for the context on the bug and sending us the PR! Really exciting stuff but also a whole new area for us to think about when it comes to long term maintenance of this.
> 
> Primiano and will discuss this together and also with the rest of the team and then come back to you with a comprehensive answer on how we want to deal with.

Sounds good. Thanks for considering this.

LalitMaganti:
Update: had an initial discussion with Primiano. We've set up a team meeting on Friday to discuss how we want to handle both this PR *and* future "weighty" external contributions like this in the future.

[...]

PR #3330, opened 2025-10-19
rust_sdk: Add initial version of Rust SDK

dreveman:
This SDK provides Rust bindings for Perfetto on top of the C ABI.

Designed to provide efficient, safe and easy to use bindings for the Rust language. Includes complete documentation with examples and a comprehensive set of unit tests.

Following components accomplish this:

- perfetto-sys crate: Serves as a low-level binding layer to the shared C library.
- perfetto crate: Provides a safe interface and is recommended for normal usage.
- perfetto-derive crate: Convenient proc macros for tracing function calls and automatically capturing all input parameters.
- protozero_rust_plugin: Used to generate rust bindings for a selected set of protos.

Like the public C headers, a complete set of rust protos are not generated by default but are instead limited to a smaller set of protos that can be expanded as needed. The rust macros used to produce bindings also make it easy to include more protos or add custom protos by projects using this SDK.

The prefetto-sys crate has two modes, "vendored" and "non-vendored". The latter is the default and depends on an existing C shared library. The first relies on amalgamated source for the C shared library, which is built and automatically linked when the crate is compiled.

The perfetto crate can take advantage of branch prediction hints when built using a toolchain that supports this. The "intrinsics" feature can be used to enable this.

Fixes: https://github.com/google/perfetto/issues/3329


primiano:
Hah, long time no see. You keep showing up with nice surprises :)

Let me discuss this with the team this Friday.

This has come up a few times, internally and externally.

THe main question I have is what happens if somebody starts using it inside google, then they want to change things, you are out meanwhile etc etc. There are a bunch of ideas floating around. let me have some thinking about the logistics and come back here.

dreveman:
> Hah, long time no see. You keep showing up with nice surprises :)
> 
> Let me discuss this with the team this Friday.
> 
> This has come up a few times, internally and externally.
> 
> THe main question I have is what happens if somebody starts using it inside google, then they want to change things, you are out meanwhile etc etc. There are a bunch of ideas floating around. let me have some thinking about the logistics and come back here.

Makes sense. FYI, I'll be able to provide two more owners from Rivos for this to make sure someone is always around to review changes if needed. Also perfectly fine with me for top-level owners to do it if that's ever needed.

primiano:
@dreveman just to join threads, see the response in https://github.com/google/perfetto/issues/3329, I commented there

dreveman:
> @dreveman just to join threads, see the response in #3329, I commented there

Yep, I saw that. I'll upload an updated version of this PR and respond to your questions shortly.

contrib/rust-sdk/perfetto-sys/build.rs
```diff
@@ -0,0 +1,94 @@
...
+            .to_string()
+    });
+
+    // Vendored bindings can be generated using:
+    if cfg!(feature = "vendored") {
+        let source_file = Path::new("libperfetto_c/perfetto_c.cc");
+        if !source_file.exists() {
+            panic!(
+                "\n\
+                ❌ Missing amalgamated source file: {}.\n\n\
```
primiano, line 34: this is so much AI generated :P (Well done)

gn/perfetto.gni
```diff
@@ -434,6 +434,9 @@ declare_args() {
   perfetto_use_system_sqlite = false
 
   perfetto_use_system_zlib = false
+
+  # Enables the Rust SDK.
+  enable_rust_sdk = false
```
LalitMaganti: nit:` enable_perfetto_rust_sdk`

primiano: minor thing: i'd move it to the upper declare_args block (the one immediately below "Tunable build variables for embedders") and definitely call it enable_perfetto_rust_sdk

Rationale for moving: the only reason for using further declare_args is if you have GN variables depending on other declare-args gn variables, as every declare_args defines an "evaluation scope". (but then I just realized some later contributions happened in random places)

Rationale for naming: this is more important, as there is a chance chromium or v8 or some other GN-based project that embeds us might use the same name and trigger some unwanted behavior in our GN file.

contrib/rust-sdk/OWNERS
```diff
@@ -0,0 +1,3 @@
+dreveman@gmail.com
+jon.bailey45@gmail.com
+ziyi.elisa.tai@gmail.com
```
primiano, line 3: This is good but this file is FYI only for humans

You want to edit also .github/CODEOWNERS (which is what really matters) with something like:

/contrib/rust-sdk @dreveman @otherPeopleGithubHandle

that is what really matters 


Also can you add an entry in CONTRIBUTORS.txt under "Active maintainers" ?

Also as soon as this CL lands, this is definitely enough material for https://perfetto.dev/docs/contributing/become-a-committer 

This would give you push ability to dev/* branches so you can use that rateher than using a forked repo for PRs (it has mostly benefits for running CI jobs without a googler ticking the box)

contrib/rust-sdk/src/protozero/protoc_plugin/BUILD.gn
```diff
@@ -0,0 +1,26 @@
+# Copyright (C) 2025 Rivos Inc.
```
primiano: can I ask you to move the protozero_rust_plugin.cc outside of contrib/, directly under src/protozero?

Here's the rationale:
- One one side I want to give you folks the ability to handle reviews to /contrib/rust-sdk autonomously without requiring any of us (unless you want it explciitly)
- But in order to do that, I need to trust /contrib as "untrusted" (I can expand more in private about what this means)
- In order to do so, soon after this CL land I'm going to add a assert_no_deps(contrib/*) from the root GN file, to make sure that our code never ends up depending on contrib/, only the other way round
- If I do so, your change to the root BUILD.gn file that builds the rust plugin will fail the assert_no_deps

Hence i feel the most pragmatic tradedoff is that we accept that the protozero plugin goes into src/, which means you'll need 2 googlers to review any PR to it (I assume it won't be touched much)

The alternative is that we figure out some extra "enable_contrib" variable that we would keep=false within Google, but you'd set to true in your builds

primiano (approved):
Given you need 2 googlers +2 for these changes i'm going to be the first one. 
@LalitMaganti  who is more rust-excited said he'll look into some more details.
I had a more glancing pass and i'm more on the side of "if this works, good for you. if it doesn't, it's in your interest to fix it :D"

I have few comments, but looks good otherwise

gn/perfetto.gni
dreveman: I removed this as part of moving this plugin into the top level src/ dir. If it's maintained there then I think it's better that it's built when building other plugins. Let me know if know if you want me to add this back.

contrib/rust-sdk/perfetto-sys/build.rs
dreveman: :)

gn/perfetto.gni
dreveman: I removed this as part of moving this plugin into the top level src/ dir. If it's maintained there then I think it's better that it's built when building other plugins. Let me know if know if you want me to add this back.

contrib/rust-sdk/OWNERS
dreveman: Done. Sounds good!

contrib/rust-sdk/src/protozero/protoc_plugin/BUILD.gn
dreveman: Moved it to src/ and removed the conditional compile of it.

gn/perfetto.gni
LalitMaganti: I think having it in the top level src + building with other plugins makes a lot of sense. It's best for code in src to have proper CI coverage.

gn/BUILD.gn
```diff
@@ -232,6 +232,8 @@ protobuf_full_deps_allowlist = [
   "../src/protozero/filtering:filter_util",
   "../tools/*",
   "../src/tools/*",
+  "../contrib/rust-sdk/src/protozero/protoc_plugin:*",
```
primiano, line 235: unneeded now (but also no need to send another PS if this is the only comment), althought it might cause some GN failure

primiano (approved):
Thanks!

gn/BUILD.gn
dreveman: good catch. I'll fix it using a quick follow up or add it to this PR if something beyond this is needed

PR #3331, opened 2025-10-19
ui: Add PerfettoSqlType

altimin:
Add a representation of a PerfettoSQL type in typescript and add an ability to parse it from a raw string.

Use it to generalise generating table definition for the table viewer.

This will be used in the future to teach the table viewer about the PerfettoSQL types and customise the behaviour based on the types.

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/column_info_unittest.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/column_info_unittest.ts
@@ -19,14 +19,18 @@ import {
   newColumnInfo,
   newColumnInfoList,
 } from './column_info';
-import {SqlColumn, SqlType} from '../../dev.perfetto.SqlModules/sql_modules';
+import {SqlColumn} from '../../dev.perfetto.SqlModules/sql_modules';
+import {PerfettoSqlType} from '../../../trace_processor/perfetto_sql_type';
 
 describe('column_info utilities', () => {
-  const stringType: SqlType = {name: 'STRING', shortName: 'string'};
-  const intType: SqlType = {name: 'INTEGER', shortName: 'int'};
-  const timestampType: SqlType = {
-    name: 'TIMESTAMP_NS',
-    shortName: 'timestamp_ns',
+  const stringType: PerfettoSqlType = {
+    kind: 'string',
+  };
+  const intType: PerfettoSqlType = {
+    kind: 'int',
+  };
+  const timestampType: PerfettoSqlType = {
+    kind: 'timestamp',
   };
 
   describe('columnInfoFromSqlColumn', () => {
@@ -39,7 +43,7 @@ describe('column_info utilities', () => {
       const result = columnInfoFromSqlColumn(sqlColumn);
 
       expect(result.name).toBe('id');
-      expect(result.type).toBe('INTEGER');
+      expect(result.type).toBe('INT');
       expect(result.checked).toBe(false);
       expect(result.column).toBe(sqlColumn);
       expect(result.alias).toBeUndefined();
@@ -68,7 +72,7 @@ describe('column_info utilities', () => {
       const result = columnInfoFromSqlColumn(sqlColumn, false);
 
       expect(result.name).toBe('ts');
-      expect(result.type).toBe('TIMESTAMP_NS');
+      expect(result.type).toBe('TIMESTAMP');
       expect(result.checked).toBe(false);
     });
   });
@@ -81,8 +85,7 @@ describe('column_info utilities', () => {
       expect(result.type).toBe('NA');
       expect(result.checked).toBe(false);
       expect(result.column.name).toBe('test_column');
-      expect(result.column.type.name).toBe('NA');
-      expect(result.column.type.shortName).toBe('NA');
+      expect(result.column.type).toBe(undefined);
     });
 
     it('should create ColumnInfo with checked=true when specified', () => {
@@ -114,7 +117,7 @@ describe('column_info utilities', () => {
       const result = newColumnInfo(original);
 
       expect(result.name).toBe('id');
-      expect(result.type).toBe('INTEGER');
+      expect(result.type).toBe('INT');
       expect(result.checked).toBe(false);
       expect(result.column).toBe(original.column);
       expect(result.alias).toBeUndefined();
@@ -132,7 +135,7 @@ describe('column_info utilities', () => {
       const result = newColumnInfo(original);
 
       expect(result.name).toBe('identifier');
-      expect(result.type).toBe('INTEGER');
+      expect(result.type).toBe('INT');
       expect(result.column.name).toBe('id');
       expect(result.alias).toBeUndefined();
     });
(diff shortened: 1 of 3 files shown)


ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -0,0 +1,110 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {errResult, okResult, Result} from '../base/result';
+
+export type PerfettoSqlType =
+  | {
+      kind:
+        | 'int'
+        | 'float'
```
aMayzner: Maybe double, to keep it consistent with the actual name of the type provided by SQL?

ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -0,0 +1,110 @@
...
+  double: 'float',
+  string: 'string',
+  bytes: 'bytes',
+  timestamp: 'timestamp',
+  duration: 'duration',
+  argsetid: 'arg_set_id',
+  id: 'id',
+};
+
+export function parsePerfettoSqlTypeFromString(
```
aMayzner, line 61: Please add unittests

ui/src/trace_processor/perfetto_sql_type.ts
altimin: Done!

ui/src/trace_processor/perfetto_sql_type.ts
altimin: Done.

altimin:
> I'm sorry but we have a few of these query types / query builders floating around the UI now, I think we really need to design a holistic solution that can solve the problem for everywhere, rather than just for one specific place (the SQL table).

In my mind, this is a more fundamental piece than a query builder — it is a direct representation of a PerfettoSQL concept in a typescript (notice that this patch doesn't add new code, but rather generalises the exisiting code from SqlModules.

P. S. I noticed that the https://perfetto.dev/docs/analysis/perfetto-sql-syntax is missing an explicit list of supported types: sent out https://github.com/google/perfetto/pull/3412 to fix that. 

P. P. S. I've update the code to include an explicit link to the documentation.

ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -0,0 +1,112 @@
...
+  int: 'int',
+  bool: 'boolean',
+  float: 'double',
+  double: 'double',
+  string: 'string',
+  bytes: 'bytes',
+  timestamp: 'timestamp',
+  duration: 'duration',
+  argsetid: 'arg_set_id',
+  id: 'id',
```
stevegolton: `id` isn't a simple type according to the types above. I gather that it probably should be though given that we have have `ID` and `ID(table, col)` types. We should expand the PerfettoSqlType above to allow both.

The problem is `Object(SIMPLE_TYPES)[value];` returns `any` so we lose all type checking at this point.

Suggestion: The type system could have picked this up for us if we'd typed the SIMPLE_TYPES object like this:

```ts
const SIMPLE_TYPES: Record<string, SimpleType['kind']> = { ...
```

Where `SimpleType` is defined as:

```ts
type SimpleType = {
  kind:
    | 'int'
    | 'double'
    | 'string'
    | 'boolean'
    | 'bytes'
    | 'timestamp'
    | 'duration'
    | 'arg_set_id';
};
```

and used in `PerfettoSqlType` like:
```ts
export type PerfettoSqlType =
  | SimpleType
  | {
      kind: 'id';
      source?: {
        table: string;
        column: string;
      };
    }
  | {
      kind: 'joinid';
      source: {
        table: string;
        column: string;
      };
    };
```

altimin: Done!

ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -0,0 +1,119 @@
...
+  timestamp: 'timestamp',
+  duration: 'duration',
+  argsetid: 'arg_set_id',
+};
+
+export function parsePerfettoSqlTypeFromString(
+  typeString: string,
+): Result<PerfettoSqlType> {
+  const value = typeString.toLowerCase();
+  const maybeSimpleType = Object(SIMPLE_TYPES)[value];
```
stevegolton: I think you can just omit the Object(...) here as `SIMPLE_TYPES[value]` will return a value of type `SimpleType['kind']` which is what you want. I think wrapping in Object erases the type (converts to any).

ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -0,0 +1,119 @@
...
+      kind: 'joinid';
+      source: {
+        table: string;
+        column: string;
+      };
+    };
+
+type SimpleType = {
+  kind:
+    | 'int'
```
stevegolton, line 38: Don't you need a simple type of 'id' for id columns (not foreign id columns)?

ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -0,0 +1,119 @@
...
+    | 'boolean'
+    | 'string'
+    | 'bytes'
+    | 'timestamp'
+    | 'duration'
+    | 'arg_set_id';
+};
+
+const SIMPLE_TYPES: Record<string, SimpleType['kind']> = {
+  long: 'int',
```
stevegolton, line 49: Same here (we need an id simple type)

ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -0,0 +1,119 @@
...
+// limitations under the License.
+
+import {errResult, okResult, Result} from '../base/result';
+
+// Representation of a PerfettoSQL type:
+// https://perfetto.dev/docs/analysis/perfetto-sql-syntax#types
+export type PerfettoSqlType =
+  | SimpleType
+  | {
+      kind: 'id';
```
stevegolton, line 22: I think this needs to be `foreign_id` or at least something different to `id` right?

ui/src/trace_processor/perfetto_sql_type.ts
altimin: Done.

ui/src/trace_processor/perfetto_sql_type.ts
altimin: There is no distinction between `id` and `foreign id` columns — it's the same id column, which might or might not have additional source. The primary restrictions are the same — each table can have only one `id` column, the values there must be unique and fit into `uint32`. 

Whether this id column refers to a different column (e.g. `slice.id` for the tables that are basically filters on `slice`) is of secondary importance, and I think that in terms of TypeScript types, this is captured well by `type.source` being optional for `id`s.

stevegolton: Ah I see, I didn't realise the source was optional - fair enough.

stevegolton: I still do think that using the kind to differentiate these types would make sense, but I'm not going to block on it, either way works.

PR #3332, opened 2025-10-19
ui: Add perfetto_sql_type to the table viewer columns

altimin:
Add a PerfettoSQLType to the TableColumn and specify the correct types in places where we create table columns.

### DIFF
--- a/ui/src/components/widgets/sql/table/columns.ts
+++ b/ui/src/components/widgets/sql/table/columns.ts
@@ -38,6 +38,10 @@ import {
   renderStandardCell,
 } from './render_cell_utils';
 import {SqlColumn, sqlColumnId, SqlExpression} from './sql_column';
+import {
+  PerfettoSqlType,
+  PerfettoSqlTypes,
+} from '../../../../trace_processor/perfetto_sql_type';
 
 function wrongTypeError(type: string, name: SqlColumn, value: SqlValue) {
   return renderError(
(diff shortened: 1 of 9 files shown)


altimin:
> We should really be relying on SQL annotations in the standard library rather than maintaining these hard coded table definitions.

Yeah, this is something on my radar — the biggest missing pieces are supporting annotations for hiding columns by default (e.g. `slice` has both `cat` and `category`, etc), supporting aliases (e.g. `Open table slice` should open a version of the slice table with threads / processes) and better support for `id` / `joinid` for all stdlib tables.

This is a step towards that — all stdlib tables are going to get their types from documentation for free and this change would unblock further improvements for `id` / `joinid`.

stevegolton:
> Yeah, this is something on my radar — the biggest missing pieces are supporting annotations for hiding columns by default (e.g. `slice` has both `cat` and `category`, etc), supporting aliases (e.g. `Open table slice` should open a version of the slice table with threads / processes) and better support for `id` / `joinid` for all stdlib tables.
> 
> This is a step towards that — all stdlib tables are going to get their types from documentation for free and this change would unblock further improvements for `id` / `joinid`.

That's fair enough, adding hidden columns isn't really something we want to do in the stdlib. Maybe the slice table as you have it could just be an additional stdlib table though?

I can see its out of scope of this PR though, and this is simply normalizing the SQL table definitions with the PerfettoSqlType type so LGTM

stevegolton (approved):

PR #3333, opened 2025-10-19
ui: refactor adding column filters to table viewer

altimin:
Move the "add column filter" menu into a separate component and make it respect
sql type of the column. This allows us to validate the input instead of guessing
its type and restrict the filters according to the column type.

Old filter menu for utid column (all options are present):
Screenshot 2025-10-20 at 01 01 15

New filter menu for utid column:
Screenshot 2025-10-20 at 01 02 21

New filter menu for category column:
Screenshot 2025-10-20 at 01 02 31


ui/src/components/widgets/sql/table/add_column_filter_menu.ts
```diff
@@ -0,0 +1,258 @@
+// Copyright (C) 2024 The Android Open Source Project
```
stevegolton: Nit 2025

PR #3350, opened 2025-10-21
ui: Grid: Reduce flicker when scrolling

stevegolton:
This patch makes the following changes to Grid's virtual scrolling:
- Reduce layzerization overhead by adding `will-change: transform` to the block of rows that moves up and down the slider when scrolling.
- Tuned the grid virtual scrolling overdraw bounds.
- Added scroll velocity compensation to offset the loaded rows a little in the direction of scroll, proportionally to how fast the user is scrolling.

Fixes: https://b.corp.google.com/issues/453940632


LalitMaganti:
Sooooo much better now! LGTM!

LalitMaganti (approved):

PR #3354, opened 2025-10-21
Consolidate common trace processor load function into redactor test fixture

edgararriagag:


edgararriagag:
This CL was created as a request from https://github.com/google/perfetto/pull/3303 to keep the test fixture infra changes independent from the redactor changes.

src/trace_redaction/trace_redaction_integration_fixture.cc
```diff
@@ -68,4 +68,12 @@ base::StatusOr<std::string> TraceRedactionIntegrationFixure::ReadRawTrace(
   return base::ErrStatus("Failed to read %s", path.c_str());
 }
 
+std::string TraceRedactionIntegrationFixure::GetSourceTrace() {
```
vaage: This is not the right way to do this. Look at `trace_redaction/trace_processor_integrationtest.cc` it does exactly what you want to do in your test.

If you copy that code into your test (before a change to the fixture) and then follow it up with a change that creates a shareable function. That establishes precedent for a change - otherwise this change has no reason to exists.

edgararriagag: Done, I modified this CL to only expose the trace processor creation method from trace_processor_integrationtest.cc as a shared function in the redactor fixture

edgararriagag:
the Files changed will show the same files as #3303 but this will look right once that other CL is landed at which point this CL can land.

vaage (approved):

LalitMaganti (approved):

PR #3357, opened 2025-10-22
tp: add new import logs table and prevent negative durations for slices

LalitMaganti:
This CL adds a new tracker and table for storing logs which happened during import.
The idea is that instead of DLOG/ELOG, we should write to this table so that we have
a queriable way to keep track of import errors beyond just the stats table which
is very generic.

Use this new table for the first time to track negative durations in slice tracker
and gpu work period: we've seen a bunch of traces hit these and it's best to track
them properly.

### DIFF
--- a/src/trace_processor/importers/common/import_logs_tracker.cc
+++ b/src/trace_processor/importers/common/import_logs_tracker.cc
@@ -35,7 +35,7 @@ void ImportLogsTracker::RecordImportLog(
     std::optional<int64_t> timestamp,
     std::optional<int64_t> byte_offset,
     std::function<void(ArgsTracker::BoundInserter&)> args_callback) {
-  PERFETTO_DCHECK(stats::kSources[stat_key] == stats::Source::kAnalysis);
+  PERFETTO_CHECK(stats::kSources[stat_key] == stats::Source::kAnalysis);
 
   context_->storage->IncrementStats(stat_key);
 


src/trace_processor/importers/common/import_logs_tracker.cc
```diff
@@ -0,0 +1,89 @@
...
+      severity_info_id_(context_->storage->InternString("info")),
+      severity_data_loss_id_(context_->storage->InternString("data_loss")),
+      severity_error_id_(context_->storage->InternString("error")) {}
+
+void ImportLogsTracker::RecordImportLog(
+    size_t stat_key,
+    std::optional<int64_t> timestamp,
+    std::optional<int64_t> byte_offset,
+    std::function<void(ArgsTracker::BoundInserter&)> args_callback) {
+  PERFETTO_DCHECK(stats::kSources[stat_key] == stats::Source::kAnalysis);
```
rsavitski: nit: do we even really need to maintain this property? I can see this just being an annoyance (discovered only at debug build time) since I don't believe we care about the accuracy of stats::Source in existing places?

LalitMaganti: I would like for this to not accidentally be used for a kTrace stat. We've made that mistake in the past and I'm protecting against that. I'm happy to make this a PERFETTO_CHECK instead if that helps?

rsavitski: Fine either as a CHECK (to let people catch this earlier in dev cycle) or no check at all (since at least personally, I don't attach too much importance to stats::Source).

PR #3362, opened 2025-10-22
Create a plugin to organize nested tracks by name

lukaville:
Adds an action to organize tracks using the ' -> ' separator.

For example, tracks with names:
```
  - Foo -> Bar -> Track1
  - Foo -> Bar -> Track2
  - Foo -> Track3
```

Will be reorganized into:
```
  - Foo
     - Bar
        - Track1
        - Track2
     - Track3
```

Sample:
image


ui/src/plugins/com.android.OrganizeNestedTracks/index.ts
```diff
@@ -0,0 +1,122 @@
...
+export default class implements PerfettoPlugin {
+  static readonly id = 'com.android.OrganizeNestedTracks';
+  static readonly dependencies = [
+    StandardGroupsPlugin,
+    TraceProcessorTrackPlugin,
+  ];
+
+  async onTraceLoad(ctx: Trace): Promise<void> {
+    ctx.commands.registerCommand({
+      id: 'com.android.OrganizeNestedTracks.organize',
```
stevegolton: Not a blocker but I believe the latest guidance for command ids is to omit the plugin name.

In this example it would be `com.android.OrganizeNestedTracks` or similar

ui/src/plugins/com.android.OrganizeNestedTracks/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      name: 'Organize nested tracks',
```
stevegolton: I would prefer a slightly more specific name here, as this command is quite specific. I'm also really starting to think we should introduce command descriptions!

stevegolton (approved):

ui/src/plugins/com.android.OrganizeNestedTracks/index.ts
lukaville: (not relevant anymore as I removed the action)

ui/src/plugins/com.android.OrganizeNestedTracks/index.ts
lukaville: After using this plugin I realized that probably action is not the best way to invoke this as you need to manually invoke it every time.

I updated this plugin to not be a 'default' plugin, and 'organize' the tracks on startup. This way as long as the plugin is enabled the tracks will be grouped automatically.

Also updated the id to 'com.android.CreateTrackHierarchyFromNames' to make it a bit more specific

PR #3364, opened 2025-10-22
Add protozero_descriptor_diff tool

ddiproietto:
It's going to be used by a future commit, as part of the build process, to reduce the size of some descriptors for proto extensions embedded in trace processor.


src/protozero/descriptor_diff/main.cc
```diff
@@ -0,0 +1,153 @@
...
+namespace protozero {
+namespace {
+
+using ::perfetto::base::ReadFile;
+using ::perfetto::base::StatusOr;
+
+const char kUsage[] =
+    R"(Usage: protozero_descriptor_diff [options]
+
+-m --minuend:    Required. Path to a protobuf descriptor (serialized
```
LalitMaganti, line 44: Uhhhhh this is a word I've never heard of ever in the English language. Would it be possible to use a more common word for this? 😃

src/protozero/descriptor_diff/main.cc
```diff
@@ -0,0 +1,153 @@
...
+                 FileDescriptorSet).
+-s --subtrahend: Required. Path to a protobuf descriptor (serialized
```
LalitMaganti, line 46: As above

src/protozero/descriptor_diff/main.cc
```diff
@@ -0,0 +1,153 @@
...
+    return 1;
+  }
+
+  StatusOr<std::string> res = DescriptorDiff(minuend_data, subtrahend_data);
+  if (!res.ok()) {
+    PERFETTO_ELOG("Error diffing: %s", res.status().c_message());
+    return 1;
+  }
+
+  if (perfetto::base::WriteAllHandle(*out_fd, res->data(), res->size()) == -1) {
```
LalitMaganti: I think this will fail on Windows becasue Handle != ScopedFile there.

ddiproietto: Fixed, thanks!

src/protozero/descriptor_diff/main.cc
ddiproietto: It's uncommon, but, at least, unambiguous (which I feel like is more important here).

I'm happy to accept any specific suggestion you have!

LalitMaganti: How about "baseline"

PR #3365, opened 2025-10-22
Reduce size of chrome_track_event descriptor

ddiproietto:
In 011464f7bc56("Prepare to split chrome_track_event.proto into multiple files (#2204)"), `exclude_imports = true` was removed from the proto descriptor target for chrome_track_event.

This unnecessarily increased the build size of trace processor, because the descriptor of the base message perfetto.protos.TrackEvent is included both in the main `trace` descriptor and in the `chrome_track_event` extension descriptor.

With this cl, at build time, after generating the `chrome_track_event` extension descriptor, we remove all the protos already known in the main `trace` descriptor.

This reduces the build size by 25 Kb. To be honest, 25 Kb probably do not justify all this extra complexity in the build, but, in the future we will have more extensions (maybe also in the huge TracePacket proto), so this infrastructure is going to be much more useful.


LalitMaganti (approved):

victorvianna:
protos/third_party/chromium/BUILD.gn LGTM, didn't look at the rest

Issue #3341, opened 2025-10-20
Counter samples continue until the end of track


znoffel:
When using counter events (`"ph":"C"`) in Chrome Trace Event JSON Format, the last counter value automatically extends horizontally until the end of the entire trace timeline, even when the measurement has actually stopped. This happens when opening a JSON trace file in Perfetto UI, but not the legacy Chrome viewer. For example: 

<img width="794" height="200" alt="Image" src="https://github.com/user-attachments/assets/cccbdfd7-a22e-49ea-82b4-5ede41572797" />

Our desired behavior is to indicate that measurement has ended, rather than extending the last value indefinitely. Ideally:
- The counter line stops at the last sample timestamp
- Or shows a visual indication that no more data is available
- Or drops to null/undefined to indicate "no data"

Is there a recommended way to terminate a counter track to prevent this extension? We considered the idea of adding a final counter event with value `0` (but this is inaccurate + the line still continues)
@LalitMaganti

stevegolton:
There is no way to this right now as far as I can see - the counter renderer is fundamentally different. Whatsmore I'm not aware of any way to indicate a 'final' sample in JSON or even the TrackEvent proto.

We *could* put this behind a flag/setting to the counter track renderer in the UI, but I'm not sure it's a high priority for us.

@LalitMaganti or @primiano will probably be able to provide more information into why this is different between the two trace viewers and whether it's something we would consider adding.

LalitMaganti:
I was thinking about the problem a bit as well and thought that maybe it might be worth marking the "last counter sample" for *all* trace types whether JSON or not. I see it as analagous to how we have special handling for "did not complete" slices.

To that end I sent https://github.com/google/perfetto/pull/3367. Here's a summary of the changes:
1) we now have a gradient from the last sample to the right edge of the screen
2) we use a "dashed" line rather than a solid line from the last sample to the right edge of the screen
3) in the details panel, we mark it as -1 duration i.e. "did not complete" instead of zero duration
4) when you hover on the sample, we preserve the dashed line instead of the solid line

Curious to get both your thoughts:
*  @znoffel: whether this is an adequate fix for your problem (i.e. is it clear enough that this is the last sample)
* @stevegolton: whether this is an acceptable UI change in your eyes

stevegolton:
Looks good from my point of view.

LalitMaganti:
[...]

PR #3367, opened 2025-10-22
ui: delineate last counter sample in the trace

LalitMaganti:
1) we now have a gradient from the last sample to the right edge of the screen
2) we use a "dashed" line rather than a solid line from the last sample to the right edge of the screen
3) in the details panel, we mark it as -1 duration i.e. "did not complete" instead of zero duration
4) when you hover on the sample, we preserve the dashed line instead of the solid line

Fixes: https://github.com/google/perfetto/issues/3341


ui/src/components/tracks/base_counter_track.ts
```diff
@@ -831,12 +832,45 @@ export abstract class BaseCounterTrack implements TrackRenderer {
...
+      ctx.beginPath();
+      ctx.moveTo(lastX, lastDrawnY);
+      ctx.lineTo(endPx, lastDrawnY);
+      ctx.lineTo(endPx, zeroY);
+      ctx.lineTo(lastX, zeroY);
+      ctx.closePath();
+      ctx.fill();
+
+      // Dashed line
+      ctx.strokeStyle = `hsl(${hue}, 10%, 71%)`;
```
stevegolton: Maybe just keep the dashed line the same color as the regular line? It's so desaturated it just looks grey.

stevegolton (approved):
Looks good to me % one comment

LalitMaganti:
We got no positive feedback with it so not proceeding

PR #3370, opened 2025-10-22
ui: Improve add debug track menu

stevegolton:
- Fix bug where select boxes can show the incorrect value
- Form validation on required fields when empty, which prohibits invalid values from being passed to the debug track factory.
- Add a `--None--` value to optional selects rather than showing blank, which makes it more obvious that this feature is disabled (i.e. for arg_set_id & pivot)
- Disable submit button if form is invalid
- Add cancel button

image

Fixes: https://b.corp.google.com/issues/453748356

### DIFF
--- a/ui/src/assets/widgets/form.scss
+++ b/ui/src/assets/widgets/form.scss
@@ -26,11 +26,15 @@
     justify-content: right;
     flex-direction: row-reverse;
     align-items: center;
-    gap: 2px;
+    gap: 4px;
   }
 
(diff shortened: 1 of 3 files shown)


aMayzner (approved):
On the screenshot it looks like dur 0. What does it mean?

stevegolton:
> On the screenshot it looks like dur 0. What does it mean?

It means instant slices - for instance if you don't have a dur column in the query

PR #3373, opened 2025-10-22
[Android Logcat] Move tag closer to the message

lukaville:
Changes the order of columns in Android Logs view
to follow the same format as `adb logcat`.

This columns order is also used by many other tools, so imho would be great to match it here as well

Screenshot with the change:

image

Sample `adb logcat` output:

```
10-22 15:49:51.390  9265  9680 E ServiceManager: error in isDeclared
10-22 15:49:51.390  9265  9680 E ServiceManager: java.lang.SecurityException: SELinux denied for service.
10-22 15:49:51.390  9265  9680 E ServiceManager: 	at android.os.Parcel.createExceptionOrNull(Parcel.java:3372)
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,9 @@ Unreleased:
       by default (set it explicitly to false to revert to previous encoding,
       which is now deprecated). This reduces trace size when recording events
       that aren't know by perfetto at compile time.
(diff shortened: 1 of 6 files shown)


lukaville:
oops, sorry, I just noticed from the screenshot that only header has changed, will send another CL with a fix

PR #3383, opened 2025-10-23
tp: Add heterogeneous lookup support to FlatHashMap

LalitMaganti:
Adds heterogeneous lookup capabilities to FlatHashMap, allowing lookups
with types different from the key type when the hasher supports it via
the is_transparent trait.

Key changes:
- Add templated Find() and Erase() overloads for heterogeneous lookup
- Add is_transparent marker to MurmurHash, FnvHash, and AlreadyHashed
- Implement heterogeneous hash functions for compatible types (e.g.,
  string/string_view/const char* combinations, integral types)
- Add benchmarks comparing heterogeneous vs regular lookup performance

Benchmark results show significant performance improvements for
heterogeneous lookups (string_view) vs regular lookups (std::string):
- Heterogeneous lookup: 16.31M lookups/s
- Regular lookup (FlatHashMap): 11.68M lookups/s (1.40x slower)
- Regular lookup (std::unordered_map): 6.14M lookups/s (2.66x slower)

The heterogeneous lookup provides a 39.7% performance improvement over
regular lookups in FlatHashMap by avoiding temporary string allocations
when looking up with string_view keys.

### DIFF
--- a/include/perfetto/ext/base/murmur_hash.h
+++ b/include/perfetto/ext/base/murmur_hash.h
@@ -192,12 +192,19 @@ struct MurmurHash {
     } else if constexpr (std::is_same_v<T, float>) {
       return murmur_internal::MurmurHashMix(
           murmur_internal::NormalizeFloatToInt<float, uint32_t>(value));
+    } else if constexpr (std::is_same_v<T, std::string> ||
+                         std::is_same_v<T, std::string_view> ||
+                         std::is_same_v<T, base::StringView>) {
+      return murmur_internal::MurmurHashBytes(value.data(), value.size());
+    } else if constexpr (std::is_same_v<T, const char*>) {
+      std::string view(value);
+      return murmur_internal::MurmurHashBytes(view.data(), view.size());
     } else {
       return std::hash<T>{}(value);
     }
   }
 
-  // Heterogeneous operator() for integeral types implicitly convertible to T.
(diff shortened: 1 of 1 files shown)


include/perfetto/ext/base/fnv_hash.h
```diff
@@ -121,6 +121,8 @@ class FnvHasher {
 // when using std::hash.
 template <typename T>
 struct FnvHash {
+  using is_transparent = void;
```
primiano: add a comment, it's not obvious why this exists and what it does

include/perfetto/ext/base/hash.h
```diff
@@ -26,7 +26,15 @@ namespace perfetto::base {
 // unordered_map<uint64_t, Value, AlreadyHashed> my_map.
 template <typename T>
 struct AlreadyHashed {
+  using is_transparent = void;
```
primiano: ditto

include/perfetto/ext/base/murmur_hash.h
```diff
@@ -178,6 +179,8 @@ Int NormalizeFloatToInt(Float value) {
 //     to be hashed without needing to override std::hash<T>.
 template <typename T>
 struct MurmurHash {
+  using is_transparent = void;
```
primiano, line 184: ditto

LalitMaganti:
PTAL. New address should address your comments.

PR #3398, opened 2025-10-24
tp: perf.data: assign distinct ids to sampling streams

rsavitski:
The perf importer uses `perf_session_id` in a different way than was intended and used when importing traced_perf packets. The id is supposed to distinguish between different sampling streams inside the file, whereas currently perf.data importer uses it once per container file. This breaks the UI tracks that follow the traced_perf assumption, as they see multiple distinct leader (timebase) counters for the same perf_session_id on the same cpu.

Fix by assigning a distinct perf_session_id to each perf_event_attr in the container. 

Technically there's future work to handle HEADER_GROUP_DESC to re-group events when using leader sampling[1], but it's just a bonus if we wanted to represent the grouping better.

[1] https://man7.org/linux/man-pages/man1/perf-list.1.html#LEADER_SAMPLING

### DIFF
--- a/test/trace_processor/diff_tests/parser/simpleperf/perf_with_add_counter_test.sql
+++ b/test/trace_processor/diff_tests/parser/simpleperf/perf_with_add_counter_test.sql
@@ -16,7 +16,6 @@ WITH
   ),
   named_counter AS (
     SELECT
-      perf_session_id,
       ts,
       cpu,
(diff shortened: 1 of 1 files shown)


rsavitski:
@LalitMaganti I want to rename PerfSession in a followup, any preference on naming, I'm thinking either of:
* PerfInvocation
* PerfContainer

LalitMaganti:
> PerfInvocation

Sounds good to me!

PR #3409, opened 2025-10-25
tp: decouple DeobfuscationTracker from StackProfileTracker

LalitMaganti:
This is important because there is a hidden assumption with
StackProfileTracker that frame names are mutable but this is not true on
perf.data files. Because of this, we were getting out of date data.

Instead, at the end of the trace, go through the frames directly. This
is a much more scalable solution

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -13175,6 +13175,22 @@ filegroup {
     ],
 }
 
+// GN: //src/protozero/descriptor_diff:lib
+filegroup {
+    name: "perfetto_src_protozero_descriptor_diff_lib",
+    srcs: [
+        "src/protozero/descriptor_diff/descriptor_diff.cc",
+    ],
+}
+
+// GN: //src/protozero/descriptor_diff:unittests
+filegroup {
+    name: "perfetto_src_protozero_descriptor_diff_unittests",
+    srcs: [
+        "src/protozero/descriptor_diff/descriptor_diff_unittest.cc",
+    ],
+}
+
 // GN: //src/protozero/filtering:bytecode_common
 filegroup {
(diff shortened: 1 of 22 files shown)


src/trace_processor/importers/proto/deobfuscation_tracker.h
```diff
@@ -0,0 +1,99 @@
...
+  std::vector<TraceBlob> packets_;
+  TraceProcessorContext* context_;
+
+  // Maps (name, package) -> set of FrameIds for deobfuscation
+  base::
+      FlatHashMap<NameInPackage, base::FlatSet<FrameId>, NameInPackage::Hasher>
+          java_frames_for_name_;
+
+  // Frames needing package guessing (temporary during EOF processing)
+  std::unordered_set<FrameId> frames_needing_package_guess_;
```
ilkos, line 94: Does this need to be a member? Can it be internal to the eof processing?

src/trace_processor/importers/proto/deobfuscation_tracker.cc
```diff
@@ -0,0 +1,347 @@
...
+            context_->storage->mutable_stack_profile_frame_table();
+        auto rr = *frames_tbl->FindById(frame_id);
+        rr.set_deobfuscated_name(context_->storage->InternString(
+            base::StringView(merged_deobfuscated)));
+      }
+    }
+  }
+}
+
+void DeobfuscationTracker::ParseDeobfuscationMappingForHeapGraph(
```
ilkos, line 172: Can we rename this since it actually does more than parsing? (afaict actually updates the tables to deobfuscate classes and field names)

PR #3415, opened 2025-10-27
Create android Users table probes

annathedroid:
User list table proto
User list probes parser and unit tests


rsavitski:
Noting that I also approve of the patch as-is (we automerged earlier in the day by mistake).

primiano:
Folks, for posterity, PRs like this should have some some more context, a bug link and a better CL description.

I just stumbled upon this in my inbox I originally thought it was only a trace processor change.
This is a whole data source.
WHat is this for? what's the plan for using this?

PR #3416, opened 2025-10-27
Add support for Android User List in Trace Processor

annathedroid:
This change introduces the ability to parse, store, and query Android user list information from Perfetto traces.

Previously, information about the different Android users (e.g., system, secondary, guest) present on the device was not extracted from the trace. This change adds the necessary logic to process the `AndroidUserList` trace packet.

Key changes include:

*   **New Table:** Added a table `__intrinsic_user_list` to store the user ID (`uid`) and user type (`type`).
*   **Proto Parsing:** Implemented parsing logic within `AndroidProbesModule` to handle the `AndroidUserList` message and populate the new table.
*   **SQL View:** Exposed the data through a new Perfetto SQL view named `android_users`, with columns `uid` and `type`.
*   **Error Handling:** Added `stats::user_list_errors` to track any issues during parsing.
*   **Testing:** Included a new diff test (`test_ftrace_user_list`) using a dedicated trace file (`trace_user_list.pftrace`) to validate the end-to-end functionality of parsing and querying the `android_users` view.

These changes enable trace analysts to easily query and correlate trace events with specific Android users on the system.


test/trace_processor/diff_tests/tables/tests.py
```diff
@@ -587,3 +587,21 @@ def test_machine(self):
...
+
+  # Ftrace user list table
+  def test_ftrace_user_list(self):
+    return DiffTestBlueprint(
+        trace=DataPath('trace_user_list.pftrace'),
+        query="""
+        SELECT
+          uid,
+          type
+        FROM __intrinsic_user_list
```
aMayzner: Use `android_users` table, not intrinsic one

annathedroid:
```suggestion
        FROM android_users
```

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
```diff
@@ -194,6 +194,20 @@ SELECT
   *
 FROM __intrinsic_thread;
 
+-- Android users.
+--
+-- NOTE: sorted by Android user id
+CREATE PERFETTO VIEW android_users (
+  -- Android user id
+  id ID,
```
aMayzner: Please don't call it ID, as it's not actually ID - it is not behaving as a standard primary key, as in, it is not contiguous. Please call it "uid" for now, and maybe you can just expose real ID (hidden under _auto_id column now) as ID to the users table.

annathedroid: No, it's fine, I just made uid, which is always unique. to actually re-allocate released id it require system restart. I thought it would be better but since table is like super small usually up to 8 records I am not concern about optimising it.

aMayzner (commented):
LGTM to me!

primiano:
Removing myself and Ryan, Anna or Lalit (just one of the two) are the best candidates to review this

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
```diff
@@ -194,6 +194,20 @@ SELECT
   *
 FROM __intrinsic_thread;
 
+-- Android users.
```
aMayzner: Please be more verbose with this table. What does it mean 'android users', what data sources have to be enabled to fetch this data?

aMayzner (approved):

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
LalitMaganti: Also sorry but this should not be in the prelude. Please add it to a standard library module.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
```diff
@@ -194,6 +194,20 @@ SELECT
...
+--
+-- NOTE: sorted by Android user id
+CREATE PERFETTO VIEW android_users (
+  -- Android user id
+  uid LONG,
```
LalitMaganti: I think I also asked about this but how does this correspond to uid on the android_process_metadata [1] table.

I think just calling this uid and "android user id" is *highly* confusing for everyone. I would like to see a explanation of this somewhere.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/third_party/perfetto/src/trace_processor/perfetto_sql/stdlib/android/process_metadata.sql;l=78?q=android_process_metadata&ss=chromium

LalitMaganti: On android_process_metada, I also see `user_id` as a column and says it is for "multi-user devices". How does that factor into things?

annathedroid: I can change it to android_user_id because it is exactly that in processes.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
annathedroid: Wait, wouldn't  that mean it wouldn't be populated every time, therefore it can't be the dimension for other traces?

LalitMaganti: The data will still be parsed into __intrinsic_user_list unconditionalyl which is fine but it will not be exposed to users without first doing an `INCLUDE PERFETTO MODULE` statement.

This is desirable: user lists do not meet the "everyone cares about this table" requirement to be in the prelude.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
annathedroid: in `android_process_metadata` 
```
linux uid = user_id * 1000 + uid (android app id)
```
bear in mind `uid` which is shown by `linux ps` is not present in `android_process_metadata` and `android app id` is called `uid`

LalitMaganti: Yeah this is a historical mistake. It's too late to change this now beyond adding some documentation there. We should have done a better job on this when we first designed that table.

annathedroid: we can name it `user_id` to keep the same name in tables

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
annathedroid: Let me double check that I think it falls under the `everyone should about this table` category.

annathedroid: Unless by `not everyone` you mean `only android`?

LalitMaganti: When I say "everyone cares about this table" this means: Chrome engineers, Fuchsia engineers, 10s of teams in Google who convert other tracing data to Perfetto, Linux kernel developers not to mention thousands of other external open source people who use Perfetto in arbitrary ways.

I don't think Android user list meets the bar to be used by all of these folks or even a signifcant amount of these. Even `android_process_metadata`, which I believe is a more important table,  requires `INCLUDE PERFETTO MODULE android.process.process_metadata` first before accessing it.

annathedroid: Great thanks for clarifying this.

src/trace_processor/perfetto_sql/stdlib/android/user_list.sql
```diff
@@ -0,0 +1,29 @@
+--
+-- Copyright 2019 The Android Open Source Project
```
annathedroid:
```suggestion
-- Copyright 2025 The Android Open Source Project
```

PR #3419, opened 2025-10-27
tp: populate WM/SF has_invalid_elapsed_ts columns

PriyankaSPX:
We need to know if the elapsed timestamp value on the proto is 0 (only relevant for converted legacy SF/WM dumps).

Bug: 455539357
Test: tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell --name-filter="WindowManager|SurfaceFlinger|PerfettoTable:winscope"

### DIFF
--- a/src/trace_processor/importers/proto/winscope/surfaceflinger_layers_parser.cc
+++ b/src/trace_processor/importers/proto/winscope/surfaceflinger_layers_parser.cc
@@ -116,6 +116,9 @@ const SnapshotId SurfaceFlingerLayersParser::ParseSnapshot(
   auto* storage = context_->trace_processor_context_->storage.get();
   tables::SurfaceFlingerLayersSnapshotTable::Row snapshot;
   snapshot.ts = timestamp;
+  protos::pbzero::LayersSnapshotProto::Decoder snapshot_decoder(blob);
+  snapshot.has_invalid_elapsed_ts =
+      snapshot_decoder.elapsed_realtime_nanos() == 0;
   snapshot.base64_proto_id = storage->mutable_string_pool()
                                  ->InternString(base::StringView(
                                      base::Base64Encode(blob.data, blob.size)))
--- a/src/trace_processor/perfetto_sql/stdlib/android/winscope/windowmanager.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/winscope/windowmanager.sql
@@ -23,15 +23,18 @@ CREATE PERFETTO VIEW android_windowmanager (
   arg_set_id ARGSETID,
   -- Raw proto message
   base64_proto_id LONG,
-  -- Focused display id for this entry
-  focused_display_id LONG
+  -- Focused display id for this snapshot
+  focused_display_id LONG,
+  -- Indicates whether snapshot was recorded without elapsed timestamp
+  has_invalid_elapsed_ts BOOL
 ) AS
 SELECT
   id,
   ts,
   arg_set_id,
   base64_proto_id,
-  focused_display_id
+  focused_display_id,
+  has_invalid_elapsed_ts
 FROM __intrinsic_windowmanager;
 
 -- Android WindowManager WindowContainer (from android.windowmanager data source).
--- a/src/trace_processor/tables/winscope_tables.py
+++ b/src/trace_processor/tables/winscope_tables.py
@@ -204,16 +204,26 @@ SURFACE_FLINGER_LAYERS_SNAPSHOT_TABLE = Table(
             cpp_access=CppAccess.READ_AND_LOW_PERF_WRITE,
             cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
         ),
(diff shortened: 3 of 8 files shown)


src/trace_processor/importers/proto/winscope/surfaceflinger_layers_parser.cc
```diff
@@ -114,8 +114,10 @@ const SnapshotId SurfaceFlingerLayersParser::ParseSnapshot(
     protozero::ConstBytes blob,
     std::optional<uint32_t> sequence_id) {
   auto* storage = context_->trace_processor_context_->storage.get();
+  protos::pbzero::LayersSnapshotProto::Decoder snapshot_decoder(blob);
+
   tables::SurfaceFlingerLayersSnapshotTable::Row snapshot;
-  snapshot.ts = timestamp;
+  snapshot.ts = snapshot_decoder.elapsed_realtime_nanos() == 0 ? 0 : timestamp;
```
LalitMaganti: Hacking timestamps like this in trace processor tables is not allowed, doubly so because you've marked this as a SORTED column on the table definition. You should be aware removing the sorted flag *will* cause signifcant slowdowns when filtering on this column.

I would suggest having a separate table for situations where you don't have timestamped data.

PriyankaSPX: This is for the specific scenario of converting legacy SF/WM dumps to the perfetto format - there would be exactly one manually constructed WM or SF packet in the trace and no other packets for that trace type. I manually construct the packet with a valid `packet.timestamp` value so it doesn't mess with any syncing/timestamp conversions for other Winscope trace data present, but we need to know that the elapsed ts associated with the Winscope data in that packet should be zero (because it wasn't reported in the legacy data).

The "ts" column will still always be sorted - elapsed_time_nanos is never zero from a collected Perfetto trace.

If this is still not allowed, I could extend the SF/WM tables to include an "override_ts" column?

Also, I thought these tables are basically ours to define and use... for example we have a "ts" column in our Transitions data tables that corresponds to various reported timestamp values in the proto data, not to the timestamp of the packet. Is that also violating what's allowed in TP?

LalitMaganti: I think my problem is that you *are* using these tables in outside of Winscope (e.g. in APC pipelines etc) and so there are expectations on how these should work. It's very confusing when 95% of tables work a certain way but 5% in a cluster work a different way.

> If this is still not allowed, I could extend the SF/WM tables to include an "override_ts" column?

I would prefer this.

> Also, I thought these tables are basically ours to define and use... for example we have a "ts" column in our Transitions data tables that corresponds to various reported timestamp values in the proto data, not to the timestamp of the packet. Is that also violating what's allowed in TP?

But you're right these are your tables, you can do what you wish. But it makes me even more incentivised to segregate this code away from the rest of trace processor.

PriyankaSPX: Noted. Don't think this would break any of those expectations that are already in place for SF/WM, but I recognise that in the event of any issue with setting elapsed_realtime_nanos that will no longer be the case - happy to just add an extra column to identify these dumps, will make a couple changes

PriyankaSPX: done

src/trace_processor/perfetto_sql/stdlib/android/winscope/windowmanager.sql
```diff
@@ -23,15 +23,18 @@ CREATE PERFETTO VIEW android_windowmanager (
   arg_set_id ARGSETID,
   -- Raw proto message
   base64_proto_id LONG,
-  -- Focused display id for this entry
-  focused_display_id LONG
+  -- Focused display id for this snapshot
+  focused_display_id LONG,
+  -- Indicates whether snapshot was recorded without elapsed timestamp
+  has_invalid_elapsed_ts LONG
```
LalitMaganti: BOOL

src/trace_processor/tables/winscope_tables.py
```diff
@@ -204,16 +204,26 @@
             cpp_access=CppAccess.READ_AND_LOW_PERF_WRITE,
             cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
         ),
-        C('sequence_id', CppUint32())
+        C('sequence_id', CppUint32()),
+        C(
+            'has_invalid_elapsed_ts',
+            CppOptional(CppUint32()),
```
LalitMaganti, line 210: CppUint32, I don't think CppOptioonalk is needed.

src/trace_processor/perfetto_sql/stdlib/android/winscope/windowmanager.sql
PriyankaSPX: done

PR #3423, opened 2025-10-27
ui: Cache vdom for individual components of SqlTableTab.

altimin:
With the recent changes to Grid, the vdom now stores the adjusted column widths,
so it's very counterproductive to recreate it.

Refactor SqlTableTab implementation to wrap the child views into Gate, ensuring
that the vdom is preserved when changing views.

Also update the spelling of some variables and switch the implementation to use TabStrip.

### DIFF
--- a/ui/src/components/details/sql_table_tab.ts
+++ b/ui/src/components/details/sql_table_tab.ts
@@ -252,25 +252,28 @@ class SqlTableTab implements Tab {
         buttons: this.getTableButtons(),
         fillHeight: true,
       },
-      (hasFilters || tabs.length > 1) &&
-        m('.pf-sql-table__toolbar', [
-          hasFilters && renderFilters(this.tableState.filters),
-          tabs.length > 1 &&
-            m(TabStrip, {
-              tabs,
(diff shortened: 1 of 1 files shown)


ui/src/components/details/sql_table_tab.ts
```diff
@@ -217,115 +252,42 @@ class LegacySqlTableTab implements Tab {
...
-                  m(MenuItem, {
-                    label: 'Open tab with filters',
-                    onclick: () => {
-                      const newState = this.state.clone();
-                      newState.filters.addFilters(node.getFilters());
-                      addSqlTableTabWithState(this.state.trace, newState);
-                    },
-                  }),
-                ),
+      (hasFilters || tabs.length > 1) &&
```
stevegolton: I think you still need the `pf-sql-table` in order to get the scrolling to work correctly. The grid should scroll so that virtualization works, but right now the entire details shell content is scrolling which results in sluggish performance as the grid can't virtualize properly.

altimin: Thanks for spotting, fixed!

PR #3434, opened 2025-10-28
Emit only interned dimension keys in the metric bundle output

elrakadm:
Emit only interned dimension keys in the metric bundle output

Some metrics have no output but the interned dimension output is the whole list of processes in the trace which is unnecessary for a feature that one of its aims is to reduce the proto output size

### DIFF
--- a/src/trace_processor/trace_summary/summary.cc
+++ b/src/trace_processor/trace_summary/summary.cc
@@ -25,6 +25,7 @@
 #include <string>
 #include <string_view>
 #include <tuple>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -402,7 +403,7 @@ base::Status WriteDimension(
 }
 
 struct InternedDimensionKey {
-  base::StringView key_column_name;
+  std::string key_column_name;
   uint64_t key_hash;
   bool operator==(const InternedDimensionKey& o) const {
     return key_column_name == o.key_column_name && key_hash == o.key_hash;
@@ -562,8 +563,8 @@ base::Status ValidateInternedDimensionSpecs(
 base::Status CheckForDuplicateAndInsert(
     uint64_t hash,
     base::StringView key_column_name,
-    base::FlatHashMap<uint64_t, bool>& seen_keys) {
-  if (!seen_keys.Insert(hash, true).second) {
+    std::unordered_set<uint64_t>& seen_keys) {
+  if (!seen_keys.insert(hash).second) {
     return base::ErrStatus(
         "Duplicate key found in interned dimension bundle with key column "
         "'%.*s'",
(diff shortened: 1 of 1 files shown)


src/trace_processor/trace_summary/summary.cc
```diff
@@ -658,6 +671,18 @@ base::Status CreateQueriesAndComputeMetrics(TraceProcessor* processor,
     const Metric* first = value.front();
     TraceMetricV2Spec::Decoder first_spec(first->spec);
 
+    base::FlatHashMap<std::string, base::FlatHashMap<uint64_t, bool>>
```
LalitMaganti: The amount of allocations you are going to have with nested structures like this is tremendous. Can we please do something more efficient?

elrakadm: Removed nested map

src/trace_processor/trace_summary/summary.cc
```diff
@@ -658,6 +686,18 @@ base::Status CreateQueriesAndComputeMetrics(TraceProcessor* processor,
...
+    base::FlatHashMap<base::StringView, bool> interned_dim_key_cols;
+    for (auto ms_it = first_spec.interned_dimension_specs(); ms_it; ++ms_it) {
+      InternedDimensionSpec::Decoder ms_decoder(*ms_it);
+      interned_dim_key_cols.Insert(InternedDimensionSpec::ColumnSpec::Decoder(
+                                       ms_decoder.key_column_spec())
+                                       .name()
+                                       .ToStdStringView(),
+                                   true);
+    }
+    base::FlatHashMap<InternedDimensionKey, bool, InternedDimensionKey::Hasher>
```
LalitMaganti: If it's a hashmap of bool, you should just use unordered_set. Also I'm not sure I understand why we have InternedDimensionKey. What is InternedDimensionHash. I don't understand why you need to store both the key and hash.

src/trace_processor/trace_summary/summary.cc
```diff
@@ -658,6 +686,18 @@ base::Status CreateQueriesAndComputeMetrics(TraceProcessor* processor,
     const Metric* first = value.front();
     TraceMetricV2Spec::Decoder first_spec(first->spec);
 
+    base::FlatHashMap<base::StringView, bool> interned_dim_key_cols;
```
LalitMaganti: Very very suspicious when you store base::StringView in a hashmap. It's rarely correct and has the potential for nasty use after frees.

src/trace_processor/trace_summary/summary.cc
elrakadm: I updated to use unordered_set in other places as well.

LalitMaganti (approved):

PR #3442, opened 2025-10-29
tp: migrate all of trace processor to use new hasher API

LalitMaganti:
Don't define custom hashers anymore unless it's really required. Define
PerfettoHashValue and use Murmurhash.

All the instances of passing base::MurmurHash explicitly can be dropped
after flathashmap uses it by default without a flag


src/trace_processor/importers/common/address_range.h
```diff
@@ -112,6 +112,11 @@ class AddressRange {
   // this range.
   constexpr bool empty() const { return length() == 0; }
 
+  template <typename H>
```
primiano, line 115: out of curiosity how did this one work before?

PR #3457, opened 2025-10-30
Move generic tables from sysui_cujs module to a separate module.

devianb:
As part of the stdlib migration, we plan to move the
android_jank_cuj table to stdlib. As a first step, move the generic
tables to a separate module, which will contain CUJ jank information
for all relevant processes.

Bug: 456326892
Test: tools/diff_test_trace_processor.py --keep-input --name-filter ".*sysui_jank_cujs.*" out/linux_clang_debug/trace_processor_shell


nicomazz (approved):
approving assuming no content change in the script (other than a small ranaming)

PR #3458, opened 2025-10-30
Add android_jank_cujs table to stdlib.

devianb:
This table is a stdlib version of the android_jank_cuj table present at
src/trace_processor/metrics/sql/android/jank/cujs.sql
This table will read CUJ slices from a common place, to keep the cuj_id
logic unified.

Bug: 456326892

### DIFF
--- a/src/trace_processor/importers/common/import_logs_tracker.cc
+++ b/src/trace_processor/importers/common/import_logs_tracker.cc
@@ -22,9 +22,7 @@
 #include <utility>
 
 #include "perfetto/base/logging.h"
-#include "perfetto/ext/base/string_view.h"
 #include "src/trace_processor/importers/common/args_tracker.h"
-#include "src/trace_processor/storage/stats.h"
 #include "src/trace_processor/storage/trace_storage.h"
 #include "src/trace_processor/tables/metadata_tables_py.h"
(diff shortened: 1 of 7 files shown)


src/trace_processor/perfetto_sql/stdlib/android/cujs/cujs_base.sql
```diff
@@ -257,3 +254,79 @@ WHERE
...
+  -- layer id associated with the actual frame.
+  layer_id LONG,
+  -- vysnc id of the first frame that falls within the CUJ boundary.
+  begin_vsync LONG,
+  -- vysnc id of the last frame that falls within the CUJ boundary.
+  end_vsync LONG
+) AS
+SELECT
+  cuj.*,
+  _extract_cuj_name_from_slice(cuj.cuj_slice_name) AS cuj_name,
```
elrakadm, line 297: I think this function should be part of the base module, and not in sysu_cujs modules as this will be a circular dependency ?

devianb: It is in the base module. It was moved here in this PR:
https://github.com/google/perfetto/pull/3457

PR #3467, opened 2025-10-30
Add initial FreeBSD support

erikarn:
This diff set adds initial FreeBSD support to building perfetto.

It's not yet complete! notably:

* there's a lot of work to do to to get the base modules all updated to at least compile/run on FreeBSD;
* the linux specific sources will run but they'll throw a lot of errors because, well linux isn't freebsd.

The goal of this pull request is to get all the components building and running so the tracing providers can then be written!

### DIFF
--- a/gn/standalone/BUILD.gn
+++ b/gn/standalone/BUILD.gn
@@ -107,6 +107,16 @@ config("extra_warnings") {
       "-Wno-switch-default",
     ]
 
+    # This is required for at least upstream and FreeBSD clang 19.
+    # See https://github.com/llvm/llvm-project/issues/76375 for more
+    # information.
+    if (is_freebsd) {
+      cflags += [
+        "-Wno-c23-extensions",
+        "-Wno-c++20-extensions",
+      ]
+    }
+
     if (perfetto_thread_safety_annotations) {
       cflags += [
         "-Wthread-safety",


erikarn:
(there's a lot more to do here, stay tuned)

LalitMaganti:
Given this is still WIP, I will make this CL a draft to avoid confusing people. When you're ready for review, let us know and we can take a look!

erikarn:
> Given this is still WIP, I will make this CL a draft to avoid confusing people. When you're ready for review, let us know and we can take a look!

Thanks! Do you know anyone who can help me figure out why the bzl/gn rules I have here so far don't include -lpthread -lrt on them like I thought I had specified?

eg;

```
adrian@francine:~/git/github/erikarn/perfetto % python3 tools/ninja -C out/freebsd perfetto traced trace_processor_shell traced_relay traced_probes
ninja: Entering directory `out/freebsd'
[6/898] link ./ipc_plugin
FAILED: [code=1] ipc_plugin 
/usr/bin/c++     @./ipc_plugin.rsp  -ldl -o ./ipc_plugin
ld: error: undefined symbol: pthread_create
>>> referenced by pthread.h:182 (/usr/include/c++/v1/__thread/support/pthread.h:182)
>>>               obj/src/base/base.thread_task_runner.o:(std::__1::__libcpp_thread_create[abi:sn190107](pthread**, void* (*)(void*), void*))
c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

i've had to manually add them to every .gn file just to verify that my (much bigger diff set) compiles and runs but I don't want to put them in here! I'd rather figure out the build issue!

Thanks!

LalitMaganti:
GN and Bazel are two different build systems: GN is our source of truth and we regenerate our Bazel files based on that.

In your change, I only see you've added -lrt/-lpthread to bazel rules *not* GN. The GN rules for these on linux can be found at https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/gn/standalone/BUILD.gn;l=347?q=gn%2Fstandalone%2FBUILD.gn&ss=chromium

Likely you need to do something similar for `if (is_freebsd)`.

erikarn:
> GN and Bazel are two different build systems: GN is our source of truth and we regenerate our Bazel files based on that.
> 
> In your change, I only see you've added -lrt/-lpthread to bazel rules _not_ GN. The GN rules for these on linux can be found at https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/gn/standalone/BUILD.gn;l=347?q=gn%2Fstandalone%2FBUILD.gn&ss=chromium
> 
> Likely you need to do something similar for `if (is_freebsd)`.

AHA brb

erikarn:
(almost there with this round)

erikarn:
ok, that's the initial diffset to get it to build.

There's a few things that popped out of this - the first one is "thread_t is 64 bit on freebsd", hence why I've done all of those recasts. What should we do for this?

LalitMaganti:
> There's a few things that popped out of this - the first one is "thread_t is 64 bit on freebsd", hence why I've done all of those recasts. What should we do for this?

I think the fallout of turning tids/pids throughout the stack into 64-bit is too large to justify given the tertiary nature of FreeBSD to the core aims of the project. Given that, I would ask that we find some way to turn it into a 32-bit number (maybe via an invertible hash function or just dropping bits etc).

erikarn:
> > There's a few things that popped out of this - the first one is "thread_t is 64 bit on freebsd", hence why I've done all of those recasts. What should we do for this?
> 
> I think the fallout of turning tids/pids throughout the stack into 64-bit is too large to justify given the tertiary nature of FreeBSD to the core aims of the project. Given that, I would ask that we find some way to turn it into a 32-bit number (maybe via an invertible hash function or just dropping bits etc).

Yeah, I didn't change it to a 64 bit value in the metarecord entry; i did that locally as an experiment and although it worked out fine I didn't want to add another 4 bytes to the record.

These are all mostly typecasts, is there a simpler way? Heck, maybe just a macro to hide the gory details?

LalitMaganti:
> These are all mostly typecasts, is there a simpler way? Heck, maybe just a macro to hide the gory details?

Is there a reason we cannot just modify base::GetThreadId with an ifdef? It seems like almost all your problems are coming from the output of that function. You can also modify PlatformThreadHandle as well.

Would centralise all of your casts instead of spreading throughout the codebase.

erikarn:
> > These are all mostly typecasts, is there a simpler way? Heck, maybe just a macro to hide the gory details?
> 
> Is there a reason we cannot just modify base::GetThreadId with an ifdef? It seems like almost all your problems are coming from the output of that function. You can also modify PlatformThreadHandle as well.
> 
> Would centralise all of your casts instead of spreading throughout the codebase.

good idea, let me go try that!

erikarn:
> GetThreadId

Oh lord, I didn't define a PlatformThreadId, so it's using pthread_t as it. Oops! Let me go try defining one for FreeBSD.

erikarn:
ok, all of those casts are gone, great suggestion!

erikarn:
ok, i think this is initial review time! thanks!

include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
LalitMaganti, line 17: Just to confirm, you did autogenerate these?

include/perfetto/base/time.h
```diff
@@ -242,7 +242,11 @@ inline TimeNanos GetWallTimeNs() {
 }
 
 inline TimeNanos GetWallTimeRawNs() {
+#if (PERFETTO_BUILDFLAG(PERFETTO_OS_FREEBSD))
+  return GetTimeInternalNs(CLOCK_MONOTONIC);
```
LalitMaganti, line 249: Please add a comment.

src/tracing/platform_posix.cc
```diff
@@ -117,6 +118,8 @@ std::string PlatformPosix::GetCurrentProcessName() {
   return cmdline.substr(0, cmdline.find('\0'));
 #elif PERFETTO_BUILDFLAG(PERFETTO_OS_APPLE)
```
LalitMaganti: Can we not jsut have a || here?

tools/run_buildtools_binary.py
```diff
@@ -52,6 +54,8 @@ def run_buildtools_binary(args):
   exe_path = os.path.join(ROOT_DIR, 'third_party', cmd, cmd) + ext
   if not os.path.exists(exe_path):
     exe_path = os.path.join(ROOT_DIR, 'buildtools', os_dir, cmd) + ext
+  if not os.path.exists(exe_path) and sys_name == "freebsd":
```
LalitMaganti: Very much not a fan of this. This whole script is predicated on the fact that we should use the stuff in buildtools. It's highly confusing it works this way on all platforms except for FreeBSD. Instead I would ask that whatever is calling this is refactored to longer do so on FreeBSD: there should always be a way to automatically use the binaries on PATH if you use the "system mode" of compiling Perfetto.

I can help guide if you tell me what compile error you are facing.

erikarn: This is what happens when I remove it:

```
adrian@francine:~/git/github/erikarn/perfetto % tools/gn args out/android       
  File "/home/adrian/git/github/erikarn/perfetto/tools/gn", line 24, in <module>
    run_buildtools_binary(['gn'] + sys.argv[1:])
  File "/home/adrian/git/github/erikarn/perfetto/tools/run_buildtools_binary.py", line 65, in run_buildtools_binary
    os.execl(exe_path, os.path.basename(exe_path), *args)
  File "<frozen os>", line 542, in execl
FileNotFoundError: [Errno 2] No such file or directory
```

There's no "buildtools/freebsd64/gn", I'm using the system installed ninja/gn stuff.

Was it supposed to be pulled down via the pre-builts or something?

include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
erikarn: oh whoops. :-) No, do i just run that script?

tools/run_buildtools_binary.py
LalitMaganti: Why not just do `gn args out/android`? 

> Was it supposed to be pulled down via the pre-builts or something?

On supported OSes yes but as you've found in this patch, FreeBSD is not one of them :) And I'm not sure we want to start supporting it to that level either as it will mean we will need to source gn/ninja/other devtools prebuilts on every update of them.

It should just be a matter of not using `tools/` wrapper scripts: though we recommend that for hermetic builds for people building on platforms we support, there's nothing stopping using a system gn or ninja or anything else.

include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
LalitMaganti: Yes.

tools/run_buildtools_binary.py
erikarn: oh wait a sec..

LalitMaganti: You can look at how this all works at `tools/install-build-deps`

erikarn: sorry I fat fingered, I normally do this, same error with the correct target directory

```
adrian@francine:~/git/github/erikarn/perfetto % tools/gn args out/freebsd       
exe_path: /home/adrian/git/github/erikarn/perfetto/buildtools/freebsd64/gn
  File "/home/adrian/git/github/erikarn/perfetto/tools/gn", line 24, in <module>
    run_buildtools_binary(['gn'] + sys.argv[1:])
  File "/home/adrian/git/github/erikarn/perfetto/tools/run_buildtools_binary.py", line 66, in run_buildtools_binary
    os.execl(exe_path, os.path.basename(exe_path), *args)
  File "<frozen os>", line 542, in execl
FileNotFoundError: [Errno 2] No such file or directory
adrian@francine:~/git/github/erikarn/perfetto % 
```

erikarn: Aha, ok, yes without calling 'tools/gn' : 

```
$ gn args out/freebsd
```

that works fine. Ok, let me drop this diff from the stack.

include/perfetto/base/time.h
erikarn: Done; described the situation in more detail.

src/tracing/platform_posix.cc
erikarn: Fixed!

include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
erikarn: Ok, what's the source I should feed it?

```
adrian@francine:~/git/github/erikarn/perfetto % ./gn/write_buildflag_header.py 
usage: write_buildflag_header.py [-h] [--rsp RSP] --out OUT
```

primiano:
Oh hey this looks very exciting! I'll look in more details later today, but I skimmed through that and looks nice.

About thread_t, as you found out we should use PlatformThreadId as that's is suppose to hide bitness and such.

One question: which build system do you need? If possible I'd suggest to use gn+ninja, and ignore bazel.

Also overall, could you share what do you plan to use/port? Are you interestd in the tracing sdk only? Are you interested in the tracing daemons (but then bsd has no ftrace, doesn't have most /proc and /sys interfaces ...). Just curious

Also this change makes me realize we should at some point start introducing  things like
PERFETTO_OS_UNIX. Somebody recently added support for QNX, and we have apple support as well.... which is getting interesting. would be nice to rationalize some of these a bit better.

It's a bit tricky because QNX is a BSD that pretends very hard to be Linux, and MacOS is a pure BSD which likes to differ in small subtle ways.

But in any case I'd do that as a separate set of PRs, not in this one.

erikarn:
Hi!

* I'm using gn; I just updated bazel for completeness;
* I plan on starting with the tracing SDK and just saying "hey now we can
build and run the basic perfetto things";
* and then once this is upstream I can err, "twist" peoples arms to build
out freebsd versions of the tracing modules; which does lead to a new
question!
* Yeah we don't TECHNICALLY have an ftrace analogue yet (people have been
using dtrace); but with upstream perfetto support I'd like to use this as a
motivation to flesh out our tracing mechanism a bit more;

Yeah I noticed that this would be easier with "PLATFORM_FEATURE_xxx" flags
(eg "yes use signal.h", "yes support posix", "yes do mmap") but as you
said, we can do that in different PRs after some discussion!

Finally the fun question - once this lands, I'd like to figure out how we
would build the tracing providers based on the target OS.

Eg - right now the system / process providers build, but obviously don't
get far because FreeBSD doesn't have a procfs. (but we do have a linux
procfs, in a cut down fashion, in /compat/linux/procfs :-) ).
I'd like to build a sample "bsd.process_stats" and "bsd.sys_stats" module
which provides the same functionality, but obviously it won't compile on
non-FreeBSD.
Once this lands I'd like to have a chat on Discord about how to get a
couple of these sample providers written and then only built if the target
platform is FreeBSD.

Thanks!

LalitMaganti:
I would suggest that we don't try and put the logic into traced_probes but instead build a different binary, exclusively which works on freebsd. I think so little of what we have is relevant. 

You can even maintain it out of our tree: that's what QNX does.

erikarn:
On Fri, 31 Oct 2025 at 07:27, Lalit Maganti ***@***.***>
wrote:

> *LalitMaganti* left a comment (google/perfetto#3467)
> <https://github.com/google/perfetto/pull/3467#issuecomment-3473338174>
>
> I would suggest that we don't try and put the logic into traced_probes but
> instead build a different binary, exclusively which works on freebsd. I
> think so little of what we have is relevant.
>
> You can even maintain it out of our tree: that's what QNX does.
>

oh ho ho ho, that sounds wonderful. Please share where they're doing this
so I can see!



-adrian

LalitMaganti:
I don't think what they're doing is open source unfortunately. But from what I understand theyre basically just using the SDK to make a bunch of QNX specific data sources.

gn/standalone/toolchain/BUILD.gn
```diff
@@ -120,6 +120,8 @@ declare_args() {
     _target_triplet = "i686-apple-darwin"
   } else if (target_os == "mac" && target_cpu == "arm64") {
     _target_triplet = "aarch64-apple-darwin"
+  } else if (target_os == "linux" && target_cpu == "x86_64") {
```
primiano: shouldn't this be target_os == "freebsd"?
Or you plan to cross-build FOR linux FROM freebsd?
it's a bit unclear

erikarn:
On Fri, 31 Oct 2025 at 07:35, Lalit Maganti ***@***.***>
wrote:

> *LalitMaganti* left a comment (google/perfetto#3467)
> <https://github.com/google/perfetto/pull/3467#issuecomment-3473366518>
>
> I don't think what they're doing is open source unfortunately. But from
> what I understand theyre basically just using the SDK to make a bunch of
> QNX specific data sources.
>

Dang, ok. Well, let's talk about this in discord, I'd like to at least get
/a/ demo compiling of one example provider!

gn/standalone/android.gni
```diff
@@ -21,6 +21,8 @@ declare_args() {
     android_host = "linux-x86_64"
   } else if (host_os == "mac") {
     android_host = "darwin-x86_64"
+  } else if (host_os == "freebsd") {
+    android_host = "freebsd-x86_64"
```
primiano: can you actually build FOR android from freebsd?
What is this for?

include/perfetto/base/build_config.h
```diff
@@ -61,6 +62,7 @@
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_OS_MAC() 1
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_OS_IOS() 0
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_OS_APPLE_TVOS() 0
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_OS_FREEBSD() 0
```
primiano: I think this one should go above, immediately below line 50 (that defines QNX = 0).
Otherwise you define only in this branch, which is the else part of OS_TV

include/perfetto/base/thread_utils.h
```diff
@@ -76,6 +76,11 @@ using PlatformThreadId = pid_t;
 inline PlatformThreadId GetThreadId() {
   return reinterpret_cast<int32_t>(pthread_self());
 }
+#elif PERFETTO_BUILDFLAG(PERFETTO_OS_FREEBSD)
+using PlatformThreadId = uint64_t;
+inline PlatformThreadId GetThreadId() {
+  return reinterpret_cast<uint64_t>(pthread_self());
```
primiano: shouldn't this be [pthread_getthreadid_np](https://man.freebsd.org/cgi/man.cgi?format=html&query=pthread_getthreadid_np) ?

we use pthread_self only in cases when there is no alternative, but pthread_self is usually a pointer to some pthread struct, not a nice number you can print in logs etc

include/perfetto/base/time.h
```diff
@@ -242,7 +242,14 @@ inline TimeNanos GetWallTimeNs() {
 }
 
 inline TimeNanos GetWallTimeRawNs() {
+#if (PERFETTO_BUILDFLAG(PERFETTO_OS_FREEBSD))
+  // Note: CLOCK_MONOTONIC_RAW is a Linux extension.
+  // FreeBSD doesn't implement it and its CLOCK_MONOTONIC_FAST
+  // doesnt implement the same semantics as CLOCK_MONOTONIC_RAW.
+  return GetTimeInternalNs(CLOCK_MONOTONIC);
+#else
```
primiano, line 250: On a tangential note, I'm surprised CLOCK_BOOTTIME works in freebsd? I thought that's also a Linux-ism and you needed CLOCK_UPTIME?

src/base/debug_crash_stack_trace.cc
```diff
@@ -225,7 +225,7 @@ void SignalHandler(int sig_num, siginfo_t* info, void* /*ucontext*/) {
 // In order to retrigger it, we have to queue a new signal by calling
 // kill() ourselves.  The special case (si_pid == 0 && sig == SIGABRT) is
 // due to the kernel sending a SIGABRT from a user request via SysRQ.
-#if PERFETTO_BUILDFLAG(PERFETTO_OS_APPLE)
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_APPLE) || PERFETTO_BUILDFLAG(PERFETTO_OS_FREEBSD)
```
primiano: nit: make sure you run `tools/format-sources`
This code is not clang-formatted

gn/standalone/toolchain/BUILD.gn
erikarn: fixed; and stranger things have happened (eg I do build openwrt on FreeBSD)

erikarn: Honestly though, I'm happy dropping this and the gn/standalone/android.gni entries if it makes things easier to land in the short term. I can tackle cross compiling later if I need to!

include/perfetto/base/thread_utils.h
erikarn: Lemme convert it to that; I'm used to having to write portable code and forgetting pthread_getthread_np() is there!

src/base/debug_crash_stack_trace.cc
erikarn:
```
clang-format: Formatting 23 files
Cannot find $third_party/clang-format/clang-format
Run tools/install-build-deps
Run tools/format-cpp-sources to fix
```

I'll fix this manually and then i'll see if I can figure out how to get this to run manually.

include/perfetto/base/time.h
erikarn: yeah, I'm shocked too :-) I don't even know if it is implemented 100% the same as linux CLOCK_BOOTTIME

From freebsd's manpage on clock_gettime:

```
STANDARDS
     The clock_gettime(), clock_settime(), and clock_getres() system calls
     conform to IEEE Std 1003.1-2008 (“POSIX.1”).  The clock IDs
     CLOCK_BOOTTIME, CLOCK_MONOTONIC_FAST, CLOCK_MONOTONIC_PRECISE,
     CLOCK_REALTIME_FAST, CLOCK_REALTIME_PRECISE, CLOCK_SECOND, CLOCK_TAI,
     CLOCK_UPTIME, CLOCK_UPTIME_FAST, and CLOCK_UPTIME_PRECISE are FreeBSD
     extensions to the POSIX interface.```

and

```
     Finally, CLOCK_BOOTTIME is an alias for CLOCK_MONOTONIC for compatibility
     with other systems and is unrelated to the kern.boottime sysctl(8).
```

Comment (erikarn) [reply include/perfetto/base/time.h:250]: I'm half tempted to put another #if FREEBSD here around CLOCK_BOOTTIME and include a comment about it.

File: include/perfetto/base/build_config.h (hunk 6 above)
Comment (erikarn) [reply include/perfetto/base/build_config.h]: done!

**Comment by erikarn** [comment]
ok, whats next?

I haven't yet got /everything/ building (I'm only building "perfetto traced trace_processor_shell traced_relay traced_probes" right now) but this is enough to kick the ball along.

**Comment by erikarn** [comment]
ok, this updated stack /almost/ allows everything to build on freebsd (ninja -C out/freebsd) .

I'm getting varadic macro empty arg issues which i'm hacking around locally to explore what the right ways to fix them are, eg

```
xample_shlib_track_event.o
../../examples/shared_lib/example_shlib_track_event.c:34:1: error: passing no argument for the '...' parameter of a variadic macro is a C23 extension [-Werror,-Wc23-extensions]
   34 | PERFETTO_TE_CATEGORIES_DEFINE(EXAMPLE_CATEGORIES)
      | ^
../../include/perfetto/public/te_category_macros.h:151:9: note: expanded from macro 'PERFETTO_TE_CATEGORIES_DEFINE'
  151 |   MACRO(PERFETTO_I_TE_CAT_DEFINE_CAT)
      |         ^
../../include/perfetto/public/te_category_macros.h:77:9: note: macro 'PERFETTO_I_TE_CAT_DESCRIPTION_GET' defined here
   77 | #define PERFETTO_I_TE_CAT_DESCRIPTION_GET(D, ...) D
      |         ^
```

I have these in a few files so far:

```
+#pragma clang diagnostic ignored "-Wc++20-extensions"
```

or

```
+#pragma clang diagnostic ignored "-Wc23-extensions"
```

So I'll keep tinkering and report back. Thanks!

**Comment by primiano** [comment]
If most targets build on freebsd, should we start landing this and addressing the rest in dedicated PRs? otherwise the cost of keeping this PR rebased will start going high very quickly (and also the cost of revieweing it)

**Comment by primiano** [comment]
> +#pragma clang diagnostic ignored "-Wc23-extensions"

rather than that I'd add a -Wno-c23-extensions in gn under  'if (is_freebsd)" (look for `Wno-c++98-compat-pedantic`, somewhere next to that)

**Review by LalitMaganti — APPROVED** [review]

File: gn/standalone/BUILD.gn [hunk 10]
```diff
@@ -107,6 +107,14 @@ config("extra_warnings") {
       "-Wno-switch-default",
     ]
 
+    # This is required for at least upstream and FreeBSD clang 19.
```
primiano, line 110: can you wrap this into if (is_freebsd)

I don't want to accidentally use c++20 extension on other platforms unless this flag is needed there

erikarn: done, thanks!

include/perfetto/base/build_configs/android_tree/perfetto_build_flags.h
primiano: You should not invoke that manually.

tools/gen_all out/xxx (where out/xx is a standard linux-target GN dir) will take care of this

primiano: I checked for you by checking out this PR, merging and running gen_all, and produces 0 diff. this is good.

please let's merge this PR ASAP. it got to the limit of how GitHub scales with many patchsets :D

PR #3474, opened 2025-10-31
Bump kMaxWriterID to 32K

etiennep-chromium:
Running Jetstream2 workload creates too many threads, which causes data corruption when hitting kMaxWriterID.
This PR bumps kMaxWriterID as a quick workaround so we can get actionable data.

Bug: chromium:450431265


primiano:
Hah now WriterIDsAllocation  test is failing on android 32 bit because it fails to create 32k TraceWriters
Maybe you need to update the test and limit the max trace writers it tried to create based on bitness. I don't have a better suggesiton,

etiennep-chromium:
> Maybe you need to update the test and limit the max trace writers it tried to create based on bitness

Do you mean have kMaxWriterID based on bitness, or only the max for the purpose of this test?
If I don't actually make it to kMaxWriterID, I also need to update the expectation on the last CreateTraceWriter().
The test also takes forever on my local machine (64bit mac)

primiano (approved):

PR #3484, opened 2025-10-31
tp: Add Experimental UNION, ORDER BY and LIMIT and OFFSET to sq

aMayzner:
Adds support in Structured Query for:
- LIMIT and OFFSET
- ORDER BY

And experimental feature:
- UNION

### DIFF
--- a/protos/perfetto/perfetto_sql/structured_query.proto
+++ b/protos/perfetto/perfetto_sql/structured_query.proto
@@ -423,6 +423,8 @@ message PerfettoSqlStructuredQuery {
     }
 
     // The list of ordering specifications. At least one is required.
+    // The order is significant: the first spec is the primary sort key,
+    // subsequent specs are used to break ties.
     repeated OrderingSpec ordering_specs = 1;
   }
   optional OrderBy order_by = 14;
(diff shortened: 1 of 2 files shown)


protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -374,4 +403,36 @@ message PerfettoSqlStructuredQuery {
     optional string column_name = 1 [deprecated = true];
   }
   repeated SelectColumn select_columns = 10;
+
+  // Represents an ORDER BY operation in SQL. Optional.
+  message OrderBy {
```
LalitMaganti, line 408: At this point we need to define an evaluation order to these operators. Probably would suggest having a terse comment and then pointing to SQL standard for detaisl.

LalitMaganti (approved):

protos/perfetto/perfetto_sql/structured_query.proto
LalitMaganti: I meant in betwen the operators (i.e. order by, limit, join etc). Not inside the order by list.

PR #3490, opened 2025-11-02
ui: add support for casting columns to the table viewer

altimin:
Allow the user to change the type of the column explicitly by
casting the column to one of the types in PerfettoSQL.

Each column can be cast to any primitive PerfettoSQL type (int, double, string, boolean, timestamp, duration) and 
select id types (slice id, utid, upid). If a column already has a given type, it can't be cast to it again. If the column
is already as cast column, a further cast would apply to the original column (so cast<int>(cast<double>(foo)) simplifies to
cast<int>(foo)).

Known limitations
- Only a small number of id types are supported for now (slice id, utid, upid).
- Cast relies on sqlite cast, which is too permissive for casts from string to int / double.

Screenshot 2025-11-02 at 16 00 37

Screenshot 2025-11-02 at 16 01 53

### DIFF
--- a/ui/src/trace_processor/perfetto_sql_type.ts
+++ b/ui/src/trace_processor/perfetto_sql_type.ts
@@ -129,6 +129,7 @@ export function parsePerfettoSqlTypeFromString(args: {
     });
   }
   if (value === 'id') {
+    // The plain `ID` are resolved into `ID($current_table.$current_column)`.
     return okResult({
       kind: 'id',
       source: {


ui/src/trace_processor/perfetto_sql_type.ts
```diff
@@ -45,6 +30,52 @@ type SimpleType = {
     | 'arg_set_id';
 };
 
+type PerfettoSqlIdType = {
+  kind: 'id' | 'joinid';
+  source: {
```
stevegolton, line 35: Doesn't this need to be optional for the id type?

altimin: I've been playing around with this and discovered that it's much more convenient to resolve the `id` columns into `id($current_table.$current_column)` during parse time — otherwise for pretty much operations with them (e.g. figuring if a given type is "slice_id") we'd need to explicitly pass table / column names to all of the places where we would want to use it, and resolving it at parse time is much easier as we already have table and column names there.

Added a comment to that effect.

PR #3504, opened 2025-11-03
exp: Simplify execution

aMayzner:
Backend (C++):
  - Added column validation for UNION queries - ensures all queries have matching column sets
  - Changed UNION SQL generation to use local WITH clauses to avoid CTE name conflicts
  - Added comprehensive tests for UNION validation scenarios

  Frontend (TypeScript):
  - Auto-execute toggle: New switch in DataExplorer lets users control when queries run
  - Manual "Run Query" button: Appears when auto-execute is off
  - Simplified node interface: Removed onExecute parameter from nodeSpecificModify()
  - Centralized execution: All query execution now happens through DataExplorer's onExecute callback
  - Better validation feedback: Shows validation errors before attempting execution

  Affected nodes:
  - SqlSourceNode: No longer tries to execute on Cmd+Enter
  - IntervalIntersectNode: Removed inline Run button
  - UnionNode: Removed inline Run button

### DIFF
--- a/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
+++ b/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
@@ -23,6 +23,7 @@
 #include <cstring>
 #include <memory>
 #include <optional>
+#include <set>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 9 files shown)


src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -575,35 +576,104 @@ base::StatusOr<std::string> GeneratorImpl::Join(
   return sql;
 }
 
+// Helper function to validate that all queries in a UNION have matching columns
+base::Status ValidateUnionColumns(
+    const std::vector<std::vector<std::string>>& query_columns) {
+  if (query_columns.empty() || query_columns[0].empty()) {
+    return base::OkStatus();
+  }
+
```
LalitMaganti, line 585: Would suggest using set_intersection or similar for this.

stevegolton (approved):

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
aMayzner: I think we want exact equality, not set_intersection. I want them to have exactly the same columns (maybe in different order

LalitMaganti: Then why not set equals?

aMayzner: I mean, I am already doing this... Just with not equal at the bottom

PR #3507, opened 2025-11-03
Add function to parse AndroidReportConfig from the trace file

KirillTim:
On Android, when uploading traces after reboot, we need to know where to upload each trace.
This information is a part of a `TraceConfig`, so we parse it back from the trace file.

Perfetto SDK users can emit the package that looks like a `TraceConfig`, so we check 
`trusted_uid` of the package to make sure the `AndroidReportConfig` is legit.

### DIFF
--- a/src/perfetto_cmd/perfetto_cmd_android.cc
+++ b/src/perfetto_cmd/perfetto_cmd_android.cc
@@ -38,7 +38,8 @@
 namespace perfetto {
 namespace {
 
-// traced runs as 'user nobody' (AID_NOBODY) that is 9999.
+// traced runs as 'user nobody' (AID_NOBODY), defined here:
+// https://cs.android.com/android/platform/superproject/+/android-latest-release:system/core/libcutils/include/private/android_filesystem_config.h;l=203;drc=f5b540e2b7b9b325d99486d49c0ac57bdd0c5344
 // We only trust packages written by traced.
 static constexpr int32_t kTrustedUid = 9999;
 


src/perfetto_cmd/perfetto_cmd.h
```diff
@@ -40,6 +40,8 @@
 #include "src/android_stats/perfetto_atoms.h"
 #include "src/perfetto_cmd/packet_writer.h"
 
+#include "protos/perfetto/config/trace_config.gen.h"
```
LalitMaganti: Can we avoid adding pb headers here by forward declaring

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -20,19 +20,24 @@
...
 #include "src/android_internal/lazy_library_loader.h"
 #include "src/android_internal/tracing_service_proxy.h"
 
+#include "protos/perfetto/trace/trace.pbzero.h"
+#include "protos/perfetto/trace/trace_packet.pbzero.h"
+
 namespace perfetto {
 namespace {
 
+static constexpr int32_t kTrustedUid = 9999;
```
LalitMaganti, line 44: Add a comment linking to the source of truth.

LalitMaganti (changes requested):

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -20,19 +20,28 @@
...
+#include "protos/perfetto/config/trace_config.gen.h"
+
+#include "protos/perfetto/trace/trace.pbzero.h"
+#include "protos/perfetto/trace/trace_packet.pbzero.h"
+
 namespace perfetto {
 namespace {
 
+// traced runs as 'user nobody' (AID_NOBODY) that is 9999.
+// We only trust packages written by traced.
```
LalitMaganti, line 43: Can you please actually link to the file [1]

[1] https://cs.android.com/android/platform/superproject/+/android-latest-release:system/core/libcutils/include/private/android_filesystem_config.h;l=203

LalitMaganti (approved):

PR #3512, opened 2025-11-03
CI: add Rust SDK checks

dreveman:
Only FYI and don't prevent CI from being green.

Fixes: https://github.com/google/perfetto/issues/3497

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -118,6 +118,13 @@ jobs:
     needs: analyze
     uses: ./.github/workflows/repo-checks.yml
 
+  rust-sdk:
+    needs: analyze
+    if:
+      needs.analyze.outputs.UI_ONLY_CHANGE == '0' &&
+      needs.analyze.outputs.TRIVIAL_CHANGE == '0'
+    uses: ./.github/workflows/rust-sdk-tests.yml
+
   # This final step joins the results of the nested workflows and succeeds if
   # all the non-skipped workflows have suceeded. This job is used by the
   # "require CI to pass" GitHub ruleset.
@@ -130,6 +137,7 @@ jobs:
       - bazel
       - fuzzer
       - repo-checks
+      - rust-sdk
     if: always()
     runs-on: ubuntu-latest
     steps:
@@ -144,9 +152,17 @@ jobs:
               exit 1
             fi
           }
+          warn() {
+            job="$1"
+            result="$2"
+            if [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
+              echo "::warning::$job failed with result: $result"
+            fi
+          }
           check linux ${{ needs.linux.result }}
           check android ${{ needs.android.result }}
           check ui ${{ needs.ui.result }}
           check bazel ${{ needs.bazel.result }}
           check fuzzer ${{ needs.fuzzer.result }}
           check repo-checks ${{ needs.repo-checks.result }}
+          warn rust-sdk ${{ needs.rust-sdk.result }}
--- a/.github/workflows/rust-sdk-tests.yml
+++ b/.github/workflows/rust-sdk-tests.yml
@@ -1,0 +1,63 @@
+# Copyright (C) 2025 The Android Open Source Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This workflow is triggered by analyze.yml
+name: Perfetto Rust SDK CI
+
+on:
+  workflow_call:
+
+permissions:
+  pull-requests: read
+
+env:
+  RUSTUP_HOME: "${{ github.workspace }}/buildtools/rustup"
+  CARGO_HOME: "${{ github.workspace }}/.cargo"
+
+jobs:
+  rust-sdk:
+    runs-on: self-hosted
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - uses: ./.github/actions/install-build-deps
+        with:
+          install-flags: --rust
+
+      # Fetch the upstream branch as well, so we can diff and see the list of
+      # changed files (unless this is a post-submit test).
+      - name: Fetch upstream branch
+        if: ${{ github.base_ref != '' }}
+        shell: bash
+        run: git fetch origin ${{ github.base_ref }} --depth=1
+
+      - name: Check format
+        # base_ref is "" in post-submit, skip the format check in postsubmit
+        # as there is nothing do diff against.
+        if: ${{ github.base_ref != '' }}
+        shell: bash
+        run: tools/format-rust-sources --upstream=origin/${{ github.base_ref }} --check
+
+      - name: Generate amalgamated shared library sources
+        shell: bash
+        run: tools/gen_amalgamated --gn_args "is_debug=false is_clang=true use_custom_libcxx=false enable_perfetto_ipc=true perfetto_enable_git_rev_version_header=true is_perfetto_build_generator=true enable_perfetto_zlib=false" --output contrib/rust-sdk/perfetto-sys/libperfetto_c/perfetto_c //src/shared_lib:libperfetto_c
+
+      - name: Build and run tests
+        run: .cargo/bin/cargo test --manifest-path=contrib/rust-sdk/Cargo.toml --features=intrinsics
+
+      - name: Run clippy
+        run: .cargo/bin/cargo clippy --manifest-path=contrib/rust-sdk/Cargo.toml --all-targets --features=intrinsics -- -D warnings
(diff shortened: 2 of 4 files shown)


.github/workflows/rust-sdk-tests.yml
```diff
@@ -0,0 +1,82 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This workflow is triggered by analyze.yml
+name: Perfetto Rust SDK CI
+on:
+  workflow_call:
+env:
+  NIGHTLY_VERSION: nightly-2025-01-03
```
LalitMaganti: This feels like quite an old nightly version. Is that intentional?

.github/workflows/rust-sdk-tests.yml
```diff
@@ -0,0 +1,82 @@
...
+            features: ""
+          - name: nightly
+            toolchain: $NIGHTLY_VERSION
+            features: $NIGHTLY_FEATURES
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Install Rust (${{ matrix.name }})
+        uses: dtolnay/rust-toolchain@master
```
LalitMaganti: I would feel more comfortable if this was pinned given it's not a "mainstream" dependency to guard against supply chain attacks.

.github/workflows/rust-sdk-tests.yml
```diff
@@ -0,0 +1,82 @@
...
+    name: Rust format (nightly)
+    runs-on: self-hosted
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Install Rust (nightly)
+        uses: dtolnay/rust-toolchain@master
```
LalitMaganti: As above.

.github/workflows/rust-sdk-tests.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          toolchain: $NIGHTLY_VERSION
+          components: rustfmt
+
+      - name: Check format of source
+        run: cargo fmt --manifest-path=contrib/rust-sdk/Cargo.toml -- --check
```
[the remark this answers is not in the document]
LalitMaganti: > Actions job or workflow does not limit the permissions of the GITHUB_TOKEN. Consider setting an explicit permissions block, using the following as a minimal starting point: {contents: read}

+1 please limit the scope.

.github/workflows/rust-sdk-tests.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Build (${{ matrix.name }})
+        run: cargo build --manifest-path=contrib/rust-sdk/Cargo.toml --verbose ${{ matrix.features }}
+
+      - name: Run tests (${{ matrix.name }})
+        run: cargo test --manifest-path=contrib/rust-sdk/Cargo.toml --verbose ${{ matrix.features }}
+
+      - name: Run clippy (${{ matrix.name }})
+        run: cargo clippy --manifest-path=contrib/rust-sdk/Cargo.toml --all-targets --all-features -- -D warnings
+
+  fmt:
```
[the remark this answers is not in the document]
LalitMaganti: > Actions job or workflow does not limit the permissions of the GITHUB_TOKEN. Consider setting an explicit permissions block, using the following as a minimal starting point: {contents: read}

+1 please limit the scope

.github/workflows/rust-sdk-tests.yml
dreveman: Yes, but somewhat random. I think it needs to be pinned to avoid correct format of code to change over time. This happens to be what we currently use internally but I'm happy to pick a more recent version if you prefer?

.github/workflows/rust-sdk-tests.yml
dreveman: Makes sense. Done.

.github/workflows/rust-sdk-tests.yml
dreveman: Done.

.github/workflows/analyze.yml
```diff
@@ -118,6 +118,10 @@ jobs:
     needs: analyze
     uses: ./.github/workflows/repo-checks.yml
 
+  rust-sdk:
+    needs: analyze
+    uses: ./.github/workflows/rust-sdk-tests.yml
```
dreveman, line 126: @LalitMaganti I'll restore the `if` statement here. Just trying to figure out how to run the step once to confirm it's working but not sure removing the `if` helped..

.github/workflows/rust-sdk-tests.yml
LalitMaganti: I'm happy to have this version to stay in sync with what you need but maybe add a comment explaining where this coms from (because it feels a bit random otherwise!)

dreveman: Done

LalitMaganti:
LGTM from my side but since this is a new Action, would ask Primiano also to comment on this just to double check I haven't missed anything from the security perspective.

dreveman:
> LGTM from my side but since this is a new Action, would ask Primiano also to comment on this just to double check I haven't missed anything from the security perspective.

Will do. No rush do merge this. But I uploaded https://github.com/google/perfetto/pull/3513 in case you want to get a release out and have crates published before. As I've unfortunately not been able to claim the names without the `-sdk` suffix at this time.

.github/workflows/analyze.yml
dreveman: Restored this for review. Not sure exactly what I need to do to have this step actually run at this time.

LalitMaganti:
Not particularly in a rush to get the release out. I think we can wait a few more days. Honestly renaming the crates sounds like a good idea anyway though given the situation. I'll LGTM that but let's wait a few more days to hear back.

primiano (commented):
Security-wise it seems alright

My only question is: how long does this take? Would it be possible / worth using caching for the installation of the toolchain like we do for our other prebuilts?

A bit lukewarm on the nightly format-check. That bot is doomed to become red and get ignored. Maybe it's better if for now this is only done locally.

You could (ask AI to) update our tools/format-sources, and create a tools/code_format_rust.py which runs only if the toolchain is installed.

Also separately we should probably have a --rust option to tools/install-build-deps which installs the relevant binaries to build rust (can do in a separate PR)

dreveman:
> Security-wise it seems alright
> 
> My only question is: how long does this take? Would it be possible / worth using caching for the installation of the toolchain like we do for our other prebuilts?

I don't know about installation of the toolchain but building and running it only takes a couple of minutes locally. I'll look at caching things once I know how long it takes in CI. Having trouble just testing it right now. Do I need the change to be approved before it runs more CI jobs? 

> 
> A bit lukewarm on the nightly format-check. That bot is doomed to become red and get ignored. Maybe it's better if for now this is only done locally.

It uses a pinned version of nightly to ensure that the correct format doesn't change over time. I added some comments about this but can also remove the test for now if you prefer that?

> 
> You could (ask AI to) update our tools/format-sources, and create a tools/code_format_rust.py which runs only if the toolchain is installed.
> 
> Also separately we should probably have a --rust option to tools/install-build-deps which installs the relevant binaries to build rust (can do in a separate PR)

Yes, didn't know if you'd be OK with that but I think it would be a good idea to have install-build-deps be able to pull down the specific rust toolchains needed to build and test the rust SDK. Updating format-sources makes sense too. I'll work on both of these tasks as follow up changes.

dreveman:
hm, looks like I'm getting this error: https://github.com/google/perfetto/actions/runs/19079295703

```
[Invalid workflow file: .github/workflows/analyze.yml#L121](https://github.com/google/perfetto/actions/runs/19079295703/workflow)
The workflow is not valid. .github/workflows/analyze.yml (Line: 121, Col: 3): Error calling workflow 'google/perfetto/.github/workflows/rust-sdk-tests.yml@6de8b6d108f4610b5ff74c2d050ab87a0b472dd9'. The workflow is requesting 'contents: read', but is only allowed 'contents: none'.
```

primiano:
> hm, looks like I'm getting this error: https://github.com/google/perfetto/actions/runs/19079295703
> 
> ```
> [Invalid workflow file: .github/workflows/analyze.yml#L121](https://github.com/google/perfetto/actions/runs/19079295703/workflow)
> The workflow is not valid. .github/workflows/analyze.yml (Line: 121, Col: 3): Error calling workflow 'google/perfetto/.github/workflows/rust-sdk-tests.yml@6de8b6d108f4610b5ff74c2d050ab87a0b472dd9'. The workflow is requesting 'contents: read', but is only allowed 'contents: none'.
> ```

Hmm

1) Any idea why do you need contents: read? what happens if you remove it altogether?
2) If you end up needing it, my theory is that needs to be granted also to the "caller" workflow (analyze.yml). I guess github is preventing that a called workflow can end up acquiring extra privileges that the caller didn't want to grant

primiano:
I kicked off a rerun of the jobs to see if it works.
The comment from the security bot is interesting, it suggests we should use contents:read... but than it's what broke.

Out of curiosity did you try adding contents:read to both analyze.yml and the rust workflow?

dreveman:
`pull-requests: read` seems to be working but this is running into some other confusing bindgen issues that I've never seen before and I need to figure out how to reproduce locally. I'm going to rebase this on `https://github.com/google/perfetto/pull/3644` first and then investigate it more.

dreveman:
@LalitMaganti @primiano This is ready to be merged. Rust-SDK CI job takes 3m ~30s to run. With some follow up caching I think we can make it a lot faster but this is hopefully good enough to be merged. PTAL.

primiano (approved):

Issue #3446, opened 2025-10-29
Publish Rust SDK crates on crates.io


dreveman:
This is currently blocked on `perfetto` and `perfetto-sys` crate names being taken. I've sent a request to be added as owner to those crates to the current owner (@MarijnS95). If not resolved by end of week I suggest that we use the following crate names on crate.io temporarily:

- perfetto-sdk
- perfetto-sdk-sys
- perfetto-sdk-derive

or alternatively use a google- prefix instead of adding sdk

LalitMaganti:
> or alternatively use a google- prefix instead of adding sdk

Given it's not maintained by Google I think this would not be preferred. I think adding -sdk makes sense. Especially as it's possible we add other things in the future (i.e. manually writing the protobuf format etc).

dreveman:
@MarijnS95 Sorry for being impatient. It was unclear if I had the right email and you had even seen this. No huge rush to respond. If you can help us out before end of week then that would be amazing but otherwise we'll just proceed as described above.

dreveman:
@LalitMaganti @primiano,

Crates are now published to crates.io:

[perfetto-sdk-sys](https://crates.io/crates/perfetto-sdk-sys)
[perfetto-sdk](https://crates.io/crates/perfetto-sdk)
[perfetto-sdk-derive](https://crates.io/crates/perfetto-sdk-derive)
[perfetto-sdk-protos-gpu](https://crates.io/crates/perfetto-sdk-protos-gpu)

Docs have also been generated for all crates:

[perfetto-sdk-sys](https://docs.rs/perfetto-sdk-sys/latest/perfetto_sdk_sys)
[perfetto-sdk](https://docs.rs/perfetto-sdk/latest/perfetto_sdk)
[perfetto-sdk-derive](https://docs.rs/perfetto-sdk-derive/latest/perfetto_sdk_derive)
[perfetto-sdk-protos-gpu](https://docs.rs/perfetto-sdk-protos-gpu/latest/perfetto_sdk_protos_gpu)

And I put a minimal example [here](https://github.com/dreveman/hello-perfetto-rust-sdk) that uses these crates.

primiano:
wohoo wow! 🥳 

thanks a lot for pushing for this!

PR #3513, opened 2025-11-03
rust-sdk: rename crates for publishing on crates.io

dreveman:
`perfetto*` are unfortunatly taken by empty crates. This adds a
`-sdk` suffix to the name in order make the crate names unique.

Issue: https://github.com/google/perfetto/issues/3446


dreveman:
@LalitMaganti I haven't heard back from the owner (@MarijnS95) of the empty perfetto crates on crates.io. Is it time to switch to these new names so the crates can be published? Or do we have a few more days before you want to cut a release?

LalitMaganti:
I suggest we merge this change. In all honesty having the -sdk suffix makes sense and better disambiguates things anyway (e.g. if some day I wanted to ship rust bindings to trace processor, I could use perfetto-trace-processor).

PR #3527, opened 2025-11-04
Return the entire `TraceConfig` when parsing the trace file

KirillTim:
When uploading traces, we actually also need a trace `uuid`, not only the 
`AndroidReportConfig`. We now return the full `TraceConfig` to extract all
the necessary information from it later.

Also change function signature: we should always check if the trace is empty
before `mmap`-ing it.

This is a follow-up to bdbc36a798c7333fbb98a9ae411ff86328f94532.

### DIFF
--- a/src/perfetto_cmd/perfetto_cmd_android.cc
+++ b/src/perfetto_cmd/perfetto_cmd_android.cc
@@ -235,9 +235,10 @@ std::optional<TraceConfig> PerfettoCmd::ParseTraceConfigFromMmapedTrace(
       continue;
 
     TraceConfig trace_config;
-    trace_config.ParseFromArray(trace_config_field.data(),
-                                trace_config_field.size());
-    return trace_config;
+    if (trace_config.ParseFromArray(trace_config_field.data(),
+                                    trace_config_field.size())) {
+      return trace_config;
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -235,17 +234,10 @@ PerfettoCmd::ParseAndroidReportConfigFromTrace(const std::string& file_path) {
     if (uid_value != kTrustedUid)
       continue;
 
-    // We already have a dependency on a 'gen::TraceConfig', so we use it here
-    // to parse the full config packet, instead of adding a dependency on a
-    // 'pbzero::TraceConfig' and using the one more nested
-    // 'protozero::ProtoDecoder' to directly access 'AndroidReportConfig'.
-    protos::gen::TraceConfig trace_config;
+    TraceConfig trace_config;
     trace_config.ParseFromArray(trace_config_field.data(),
```
LalitMaganti: Afaik this returns a Boolean which needs to be checked indicating parse failure/success.

KirillTim: Ouch, it indeed returns boolean, thanks!

PR #3528, opened 2025-11-04
Dev/chengxli/cpu uptime calculator

azaleax:
Adding CPU uptime calculation to the stdlib.

Adding a table and macro which calculates the theoretical CPU uptime cost attributed to each thread/process.
It works by assigning the "cost" (idle duration required) to enter deeper C-states (C2, C3, etc.) to the thread that kept the CPU active during that time, preventing the C-state entry.


src/trace_processor/perfetto_sql/stdlib/android/cpu/cpu_uptime.sql
```diff
@@ -0,0 +1,195 @@
...
+  *,
+  row_number() OVER (ORDER BY ts) AS group_id
+FROM step1;
+
+CREATE VIRTUAL TABLE _android_active_sched_joined USING SPAN_JOIN (_cpu_active_for_span PARTITIONED cpu, _sched_processes_for_span PARTITIONED cpu);
+
+-- Table which groups scheduling information along with the CPU idle state
+-- information. This is meant to be used in conjunciton with the android_cpu_uptime_cost
+-- calculation macro.
+CREATE PERFETTO TABLE android_active_sched_joined (
```
LalitMaganti, line 60: Really don't like this table name. Please be more descriptive and consistent with the style of the stdlib instead of just calling it "joined"

src/trace_processor/perfetto_sql/stdlib/android/cpu/cpu_uptime.sql
```diff
@@ -0,0 +1,195 @@
...
+-- * Active_dur: Duration where the CPU is continuously active.
+-- * Sched1_dur: Scheduled duration of the first process after the CPU becomes active.
+-- * SchedN_dur: Scheduled duration of the Nth process in the active window.
+-- * C2_cost: Idle duration needed to enter C2 state.
+-- * C3_cost: Idle duration needed to enter C3 state.
+--
+-- Calculation:
+-- * Uptime1 (First Process): Sched1_dur + C2_cost + C3_cost
+-- * UptimeN (Subsequent Processes, N > 1): SchedN_dur
+CREATE PERFETTO MACRO android_cpu_uptime_cost(
```
LalitMaganti, line 127: Both uptime and cost are a bit vague here. Uptime means to me device uptime (i.e. time since boot). Now usually I think that word also includes time in suspend (tho I'm not 100% on that). 

Can we use better words?

src/trace_processor/perfetto_sql/stdlib/android/cpu/cpu_uptime.sql
```diff
@@ -0,0 +1,195 @@
+--
```
LalitMaganti, line 1: All of this code needs tests.

PR #3529, opened 2025-11-04
FreeBSD: Small quality of life improvements

primiano:
- Use llvm-ar rather than ar, as freebsd's ar doesn't support response files
- Use "env bash" as freebsd's has bash under /usr/local/bin
- Fallback on /usr/local/bin when invoking gn/ninja as we don't have
  prebuilts for freebsd and likely never will, as that's the job of ports.
- use monolithic binaries as freebsd doesn't seem to pick up automatically the .so path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 .venv
 .vscode
 *.code-workspace
+*.core
 *.iml
 *.LOCAL.md
 *.pyc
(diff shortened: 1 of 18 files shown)


primiano:
@erikarn FYI

erikarn:
neat, thanks!

PR #3536, opened 2025-11-05
Improve `base::ListFilesRecursive`.

KirillTim:
Follow symlinks, return absolute paths, add tests.

### DIFF
--- a/python/tools/check_ratchet.py
+++ b/python/tools/check_ratchet.py
@@ -36,7 +36,7 @@ import dataclasses
 
 from dataclasses import dataclass
 
-EXPECTED_ANY_COUNT = 37
+EXPECTED_ANY_COUNT = 35
 EXPECTED_RUN_METRIC_COUNT = 4
 
 ROOT_DIR = os.path.dirname(
--- a/ui/src/assets/widgets/form.scss
+++ b/ui/src/assets/widgets/form.scss
@@ -15,26 +15,72 @@
 @import "../theme";
 
 .pf-form {
-  display: grid;
-  grid-template-columns: auto auto;
-  gap: 8px;
+  display: flex;
+  flex-direction: column;
 
-  .pf-form-button-bar {
-    grid-column: span 2;
-    margin-top: 6px;
+  &__button-bar {
     display: flex;
     justify-content: right;
     flex-direction: row-reverse;
-    align-items: center;
+    align-items: baseline;
     gap: 4px;
   }
 
-  .pf-form-label {
-    font-weight: bolder;
+  &__label {
+    margin-bottom: 6px;
+    font-size: smaller;
+    font-weight: 500;
+    color: var(--pf-color-text-muted);
   }
 
-  select:invalid {
-    outline: 2px auto var(--pf-color-danger);
+  &__section {
+    border: 1px solid var(--pf-color-border);
+    display: flex;
+    flex-direction: column;
+    border-radius: 4px;
+    padding-inline: 8px;
+    padding-top: 16px;
+    position: relative;
+    margin-block: 8px;
+
+    &-label {
+      position: absolute;
+      font-weight: 500;
+      text-transform: uppercase;
+      font-size: 0.7em;
+      margin-bottom: 8px;
+      top: -0.5em;
+      left: 4px;
+      background: var(--pf-color-background);
+      color: var(--pf-color-text-muted);
+      padding: 0 2px;
+    }
+  }
+
+  // Styles for common input elements
+  .pf-checkbox,
+  .pf-switch {
+    margin-bottom: 6px;
+  }
+
+  .pf-select,
+  .pf-text-input {
+    margin-bottom: 8px;
+  }
+
+  .pf-select:invalid {
+    border-bottom-color: var(--pf-color-danger);
+    &:active,
+    &:focus {
+      box-shadow: 0 1px 0 0 var(--pf-color-danger);
+    }
+  }
+
+  .pf-text-input:has(input:invalid) {
+    border-bottom-color: var(--pf-color-danger);
+    &:has(input:focus) {
+      box-shadow: 0 1px 0 0 var(--pf-color-danger);
+    }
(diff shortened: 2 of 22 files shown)


src/base/file_utils.cc
```diff
@@ -354,12 +359,15 @@ base::Status ListFilesRecursive(const std::string& dir_path,
   std::deque<std::string> dir_queue;
   dir_queue.push_back(root_dir_path);
 
+#if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  // Used to avoid infinite loops with circular symlinks.
+  base::FlatSet<std::pair<dev_t, ino_t>> visited_dirs;
```
LalitMaganti, line 364: Why not unordered_set?

src/base/file_utils.cc
```diff
@@ -400,12 +408,25 @@ base::Status ListFilesRecursive(const std::string& dir_path,
...
       struct stat dirstat;
       std::string full_path = cur_dir + dirent->d_name;
-      PERFETTO_CHECK(stat(full_path.c_str(), &dirstat) == 0);
+
+      if (stat(full_path.c_str(), &dirstat) != 0) {
+        // skip broken symlinks.
+        if (errno == ENOENT) {
+          continue;
+        }
+        PERFETTO_FATAL("stat %s", full_path.c_str());
```
LalitMaganti, line 417: Leftover? This should return an error no?

src/trace_processor/trace_processor_shell.cc
```diff
@@ -1441,26 +1440,26 @@ base::Status IncludeSqlPackage(TraceProcessor* trace_processor,
...
-  for (const auto& path : paths) {
-    if (base::GetFileExtension(path) != ".sql") {
+  for (const auto& file_path : paths) {
+    if (base::GetFileExtension(file_path) != ".sql") {
       continue;
     }
-
-    std::string path_no_extension = path.substr(0, path.rfind('.'));
-    if (path_no_extension.find('.') != std::string_view::npos) {
+    std::string file_name = BaseName(file_path);
```
LalitMaganti, line 1447: This is *not* waht this function used to do. This only returns the filename. What it used to do before was the path relative to the root.

This is not the same.

src/base/file_utils.cc
KirillTim: I decided to keep this consistent with the previous implementation, where we had just
`PERFETTO_CHECK(stat(full_path.c_str(), &dirstat) == 0);`.

The broken symlink seems to be an only reasonable `stat` error to ignore and not to crash.

src/base/file_utils.cc
KirillTim: Daniele once told me, that for the small set sizes it is better to use the `FlatSet`.
We can expect the set size to be small :)

LalitMaganti: Sure but this is one of those situations if someone *does* accidentally point TP at a big directory then instead of O(nlogn), you'll be O(n). For small data sets it's marginal but it doesn't matter too much.

Do less harm IMO. There's no "memory" reason to use FlatSet (which is the main reason to use them).

KirillTim: Got it, will change to `std::unordered_set`.

src/base/file_utils.cc
```diff
@@ -400,12 +408,25 @@ base::Status ListFilesRecursive(const std::string& dir_path,
...
+      }
+
       if (S_ISDIR(dirstat.st_mode)) {
+        auto dir_id = std::make_pair(dirstat.st_dev, dirstat.st_ino);
+        if (visited_dirs.find(dir_id) != visited_dirs.end()) {
+          continue;
+        }
```
primiano: This one makes me a little bit nervous. if for any reason (some weird filesystem, or selinux permissions) we read 0,0 here, we can end up ignoring legitimate dirs.

Could the check be done on the dir that we push ?

primiano:
My personal take on this PR:

To begin with, it feels unnecessary. I was hoping to see a PR to solve the reboot problem, instead this one is more of an architectural embellishment. Why are they not done in the opposite order (fix the problem first, embellish the code afterwards?)

The other problem I have, is that I'm generally not a fan of adding functionality that is NOT needed. It leads to brittle codebases. Nobody is going to use "follow symlinks" feature anytime soon. And I strongly believe a codebase should avoid having code that has not been used in production. BEcause one day somebody will find that code, believe that has been battle tested like the rest, and they will find out they are the first users, and get surprises.

It doesn't matter how many tests you add. This is system level programming. You cannot know what goes wrong, until you use it for real. Selinux can bite us. F2FS or some weird FS can bite us.  You don't know what goes wrong until you use it for real with a concrete use case. But you don't have any, because we don't plan to use symlinks ( unless i'm missing something). So you'd be adding code that nobody will exercise. Which weakens the codebase overall.

Having said this, I'm not generally against making things better architecturally, however, I have a few concerns here.

1. I am not convinced that changing the semantic of the function to "return absolute paths" is actually a good change. going from relative->absolute is easy, it's just a string concatenation. Going from absolute->relative is harder, as now you need a basename (and then need to deal with all the possible combination of trailing slashes)

2. In the process of this, you broke one client in trace processor that actually needed relative paths

3. The follow symlink feature makes me very nervous. At very least it should be an option people opt-in, not a given. The history of computer security is full bugs relates in a way to symlinks. Symlinks are to security what regex are to memory.

4. The way the "avoid loops" is implemented makes makes me nervous. If for any weird reason that stat call fails partially and returns (0,0) you will start ignoring directories, and that bug would be extremely hard to debug. I am not confident enough to say "I am 100% sure st_ino/dev will be always populated". I've been bitten enough by weirdnesses of syscalls and android to be cautious.

Overall, if I balance everything, this is my take on this CL: it introduces a number of non-trivial risks to the codebase, increasing its fragility. OTOH doesn't give any big value. It's a nice-to-have...with many new potential surprises attached to it.

As such, I am not in favour of this change.

PR #3539, opened 2025-11-05
shared_lib: Set shmem_page_size_hint_kb to 4.

dreveman:
Matches the PerfettoProducerBackendInitArgsSetShmemSizeHintKb documentation where callers are told to that a multiple of 4KB must be used.

No actual change in behavior with a release build of the tracing service as a 0 value will be adjusted to become 4KB but it removes an "Invalid configured SMB sizes" message when using a debug build of the tracing service.

### DIFF
--- a/src/shared_lib/producer.cc
+++ b/src/shared_lib/producer.cc
@@ -62,6 +62,7 @@ void PerfettoProducerInProcessInit(
   perfetto::TracingInitArgs args;
   args.backends = perfetto::kInProcessBackend;
   args.shmem_size_hint_kb = backend_args->shmem_size_hint_kb;
+  args.shmem_page_size_hint_kb = 4;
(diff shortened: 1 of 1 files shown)


dreveman:
@ddiproietto Far from critical but the debug logs from the service were confusing to me and had me think that the `shmem_size_hint_kb` API wasn't working at first glance. It works fine without this. Just always producing a warning message.

primiano:
@ddiproietto left the team last week, I'm happy to look at these changes for the time being.

Yes that log message is annoying.
My theory is that rather than doing it in here , we should do somewhere inside in the impl (still on the SDK side), maybe TracingMuxerImpl::InitializeInstance or something inner (on the run now, but can look more later)

dreveman:
Closing as not critical. This was just removing a confusing log message

PR #3541, opened 2025-11-05
tprobes: add crash keys for data source & session counts

rsavitski:
There is evidence that droidfood tracing is instantiating so many concurrent data sources, that traced_probes ends up quite starved in the traced_probes<->traced shared memory capacity, as each TraceWriter might be holding onto a 32KB chunk (out of 1MB).

Add crash keys to assist debugging/tracking on android. I decided to log from Setup & Stop, but could be convinced to change it to count /started/ data sources instead.

Bug: 458052719

### DIFF
--- a/src/traced/probes/probes_producer.cc
+++ b/src/traced/probes/probes_producer.cc
@@ -22,6 +22,7 @@
 #include <string>
 
 #include "perfetto/base/logging.h"
+#include "perfetto/ext/base/crash_keys.h"
 #include "perfetto/ext/base/utils.h"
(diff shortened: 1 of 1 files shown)


sashwinbalaji (approved):
would capturing the config names also help in debugging ?

rsavitski:
> would capturing the config names also help in debugging ?

iirc the producers don't know the name of the session, but regardless the crash keys are limited (32 per process), so it's not an appropriate mechanism for logging repeated structures.

PR #3542, opened 2025-11-05
tprobes: ask for 2MB shmem when connecting to traced, instead of 1MB

rsavitski:
There is evidence that droidfood tracing is instantiating so many concurrent data sources, that traced_probes ends up quite starved in the traced_probes<->traced shared memory capacity, as each TraceWriter might be holding onto a 32KB chunk (out of 1MB).

Double the requested shmem size to give us some headroom. I'd argue that the difference between 1MB and 2MB is negligible given how tracing is used (e.g. kernel buffers are typically on the order of 8MB per cpu on android).

Bug: 458052719


primiano (approved):
LGTM. this will be caught by crystalball, we can discuss with them once we get there. i think it's fine

Issue #3537, opened 2025-11-05
Clarification requested about the max size of a trace we can load in perfetto?


colleeneb:
Can I get clarification about the size of a trace we can load in perfetto?

https://perfetto.dev/docs/visualization/large-traces says that we can "take full advantage of the RAM of your machine" if we load via `trace_processor`. However, when I tried this I hit the error:
```
./trace_processor --httpd ./out.pftrace
[176.253]    trace_blob_view.h:56 PERFETTO_CHECK(blob.size() <= std::numeric_limits<uint32_t>::max()) (errno: 0, Undefined error: 0)

-----BEGIN PERFETTO PRE-CRASH LOG-----
[176.253]    trace_blob_view.h:56 PERFETTO_CHECK(blob.size() <= std::numeric_limits<uint32_t>::max()) (errno: 0, Undefined error: 0)
```
which seems like it restricts the size of max uint32, around 4GB. My system has 16 GB RAM so I was hoping I would be able to load something larger than 4GB. 

Could someone clarify if 4 GB is the limit indeed or am I missing something? Thanks a lot! 

(Tagging @TApplencourt since he's interested as well)

LalitMaganti:
Hmmm I've had a few reports about this now so I think there might have been some regression on this. The trace processor should support loading traces as big as you have memory for.

Will investigate.

TApplencourt:
The check seem to be introduced by https://github.com/google/perfetto/pull/2587 (if that help)

LalitMaganti:
https://github.com/google/perfetto/pull/3543 will fix this issue! Thanks for the report!

TApplencourt:
Thanks a lot! Amazing as usual <3

colleeneb:
Thank you! Super fast!

PR #3543, opened 2025-11-05
tp: fix crash when loading traces >4GB

LalitMaganti:
Turns out that https://github.com/google/perfetto/pull/2587 could
have just have turned uint32_t -> size_t. My fears in the original
bug (b/438916722) about memory increases was unfounded: because of
struct padding TraceBlobView is *already* taking 24 bytes so this
change is actually a noop.

Fixes: https://github.com/google/perfetto/issues/3537

### DIFF
--- a/include/perfetto/trace_processor/trace_blob_view.h
+++ b/include/perfetto/trace_processor/trace_blob_view.h
@@ -17,18 +17,17 @@
 #ifndef INCLUDE_PERFETTO_TRACE_PROCESSOR_TRACE_BLOB_VIEW_H_
 #define INCLUDE_PERFETTO_TRACE_PROCESSOR_TRACE_BLOB_VIEW_H_
 
-#include <stddef.h>
-#include <stdint.h>
-
+#include <cstddef>
+#include <cstdint>
 #include <limits>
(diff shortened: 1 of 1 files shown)


sashwinbalaji (approved):

include/perfetto/trace_processor/trace_blob_view.h
```diff
@@ -51,39 +50,26 @@ class alignas(8) TraceBlobView {
...
-  TraceBlobView(RefPtr<TraceBlob> blob, size_t offset, uint32_t length)
-      : blob_(std::move(blob)), data_(blob_->data() + offset), length_(length) {
-    PERFETTO_DCHECK(offset + length_ <= blob_->size());
+                         size_t length = kWholeBlob)
+      : TraceBlobView(RefPtr<TraceBlob>(new TraceBlob(std::move(blob))),
+                      offset,
+                      length) {}
+
+  TraceBlobView(RefPtr<TraceBlob> blob, size_t offset, size_t length) {
+    // Note: we cannot use field initialization as C++ does not define the
```
sashwinbalaji, line 59: interesting

PR #3553, opened 2025-11-05
rust-sdk: fix typo in rust bindings causing patch errors

dreveman:
`>=` should be `<=` in PbMsgSizeField::patch. This could cause packets large enough to be chunked up from being handled correctly.

Issue: https://github.com/google/perfetto/issues/3329


dreveman:
@zytyz please take a look. thanks for helping track this down and writing the test for it!

zytyz:
lgtm :)

PR #3557, opened 2025-11-06
ui: NodeGraph: Add node context menus

stevegolton:
Add the ability to add context menu items to a node using the `contextMenuItems` parameter. This list should take a list of `m(MenuItem)` vnodes.

E.g.
```
contextMenuItems: [
    m(MenuItem, {
      label: 'Delete',
      icon: 'delete',
    }),
  ]
```

With no title bar:

image

With title bar:

image


aMayzner:
I thought that context menu appearing on hover/select as a floating button would be easier. With this implementation it would be difficult to keep it so that everything inside always renders prettily with the menu

PR #3562, opened 2025-11-06
tp: Interval intersect didn't handle partition by string columns

aMayzner:
Interval intersect with string partition columns was producing nonsensical garbage output instead of the actual string values. For example, querying with _interval_intersect!((s, s), (process_name, thread_name)) would return gibberish like "����U" instead of actual process/thread names like "mdss_fb0".

This was a dangling pointer bug for string partition columns, non-string partition columns (integers, etc.) worked correctly.

Root Cause

In src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc, the IntervalTreeIntervalsAgg::Step() function was storing SqlValue objects containing raw string pointers from SQLite (sqlite3_value_text()). These pointers are only valid during the current SQLite function call, as documented in basic_types.h:

  // This string will be owned by the iterator that returned it and is valid
  // as long until the subsequent call to Next().

When the partition values were accessed later in interval_intersect.cc, the pointers pointed to invalid/freed memory, resulting in garbage output.

The Fix

  - Modified RegisterTypeBuilderFunctions() to accept StringPool* pool parameter
  - Added UserData struct to PartitionedTable to carry the StringPool pointer
  - String values are interned during Step() before being stored in partition keys

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
@@ -310,6 +310,7 @@ struct IntervalTreeIntervalsAgg
   static constexpr char kName[] = "__intrinsic_interval_tree_intervals_agg";
   static constexpr int kArgCount = -1;
   static constexpr int kMinArgCount = 3;
+
   struct AggCtx : sqlite::AggregateContext<AggCtx> {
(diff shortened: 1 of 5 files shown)


LalitMaganti (changes requested):
Please rethink this patch fundamentally. We definitely don't want to buffer all strings in all rows in a deque.

PR #3568, opened 2025-11-06
rust-sdk: more cargo fixes for publishing

dreveman:
s/profiling/development-tools::profiling/ as top level profiling category doesn't exist.

Issue: https://github.com/google/perfetto/issues/3446


dreveman:
sorry about these things coming in one by one. publishing dry runs apparently doesn't catch these so not until I do the real thing that I catch these errors..

LalitMaganti (approved):

PR #3572, opened 2025-11-06
docs: Add docs about traced and traced_probes

sashwinbalaji:


docs/reference/traced_probes.md
```diff
@@ -0,0 +1,292 @@
...
+
+This separation of concerns is a key part of Perfetto's design. `traced` is the
+central manager, while `traced_probes` is a specialized data provider. This
+decoupled architecture allows for multiple, independent producers and consumers
+to interact with the tracing system simultaneously without interfering with each
+other.
+
+![traced_probes and traced](/docs/images/platform-tracing.png)
+
+## Data Sources
```
LalitMaganti, line 45: I would strongly suggest we separate out the android data source from the Linux ones to make it very clear.

LalitMaganti: Also we should really give a snippet of the config as an example so people don't have to go searching for it.

LalitMaganti: Maybe I would do one per section of data source even as well as one overall indicating the general structure.

docs/reference/traced.md
```diff
@@ -0,0 +1,100 @@
+# TRACED(8)
+
+## NAME
+
+traced - The Perfetto Tracing Service
+
+## DESCRIPTION
+
+`traced` is the central daemon in Perfetto's
+[service-based architecture](/docs/concepts/service-model.md). It acts as the
```
LalitMaganti, line 10: This link will not be meaningful in debian

docs/reference/traced.md
```diff
@@ -0,0 +1,100 @@
+# TRACED(8)
```
LalitMaganti, line 1: Should we link to these from the TOC?

docs/reference/traced.md
```diff
@@ -0,0 +1,100 @@
...
+*   `--version`: Prints the version number and exits.
+*   `--set-socket-permissions
+    <prod_group>:<prod_mode>:<cons_group>:<cons_mode>`: Sets the group ownership
+    and permission mode for the producer and consumer sockets. This is important
+    for controlling which users and processes can connect to `traced`.
+*   `--enable-relay-endpoint`: Enables an endpoint for multi-machine tracing via
+    `traced_relay`.
+
+## Built-in Producer
+
```
LalitMaganti, line 90: It's not just for these reasons: we also dump clock snapshots, stats of number of prodcuers etc.

docs/reference/traced.md
sashwinbalaji: for debian updated the rule to extract them out.

docs/toc.md
```diff
@@ -136,6 +136,8 @@
     - [Command Line Reference](#)
 
       - [perfetto cmdline](reference/perfetto-cli.md)
```
LalitMaganti: perfetto_cmd

docs/reference/traced_probes.md
```diff
@@ -0,0 +1,577 @@
...
+        name: "linux.ftrace"
+        ftrace_config {
+          ftrace_events: "sched/sched_switch"
+          ftrace_events: "power/cpu_idle"
+          ftrace_events: "sched/sched_waking"
+        }
+      }
+    }
+    ```
+*   **Implementation Details**:
```
LalitMaganti: I think implementation details are not so important honestly.

LalitMaganti: Here and below.

docs/reference/traced_probes.md
```diff
@@ -0,0 +1,577 @@
...
+      }
+    }
+    ```
+*   **Implementation Details**: Connects to the Android StatsD service via
+    Binder to subscribe to push and pull atoms. It uses
+    `CreateStatsdShellConfig()` to generate the subscription configuration.
+*   **Configuration**: `StatsdTracingConfig` allows specifying `pull_config`
+    (for pull atoms with frequency and packages) and `push_atom_id` (for push
+    atoms).
+
```
LalitMaganti: I feel like both of these should be before list of data sources.

LalitMaganti (approved):

PR #3575, opened 2025-11-07
debian: Remove debian/ folder

sashwinbalaji:
To have one single source of truth for debian related files, delete the folder and add debian.txt to point readers to its new home.


sashwinbalaji:
fyi: @ribalda

PR #3577, opened 2025-11-07
tp: Fix CTE name collision in structured query generator

aMayzner:
## Summary
Fixes a bug in the structured query generator where Common Table Expression (CTE) names could collide when explicit query IDs matched auto-generated index-based names, resulting in duplicate CTE definitions and invalid SQL.

## Problem
The generator creates CTE table names using two sources:
1. **Explicit IDs**: When a query proto has `id="X"`, it generates `sq_X` or `shared_sq_X`
2. **Index-based**: When no ID is provided, it uses the query's position in the state vector (e.g., query at index 2 → `sq_2`)

These two naming schemes could collide. For example:
- A nested query at index 2 would get `table_name="sq_2"`
- An inner_query with `id="2"` would also get `table_name="sq_2"`
- This produced SQL with duplicate `WITH sq_2 AS (...)` clauses, causing SQL errors

## Solution
Added a collision avoidance mechanism that tracks all used table names and appends numeric suffixes when collisions are detected:

1. **New tracking set**: Added `std::set<std::string> used_table_names_` to `GeneratorImpl` class to track all CTE names across a single query generation
2. **Deduplication logic**: Modified `QueryState` constructor to check if a proposed name already exists, and if so, append a suffix (`_0`, `_1`, `_2`, etc.) until a unique name is found
3. **Thread-safe design**: The set is passed by reference through all `QueryState` constructions, ensuring consistent tracking within a single `Generate()` call

## Implementation Details

  **Core changes** (structured_query_generator.cc):
  ```cpp
  // Constructor now takes reference to used names set
  QueryState(QueryType _type, protozero::ConstBytes _bytes, size_t index,
             std::optional<size_t> parent_idx,
             std::set<std::string>& used_table_names)

  // Collision detection and resolution
  std::string original_name = table_name;
  size_t suffix = 0;
  while (used_table_names.count(table_name) > 0) {
    table_name = original_name + "_" + std::to_string(suffix);
    suffix++;
  }
  used_table_names.insert(table_name);
```

  All three QueryState construction sites updated:
  - Root queries (line 230)
  - Shared/referenced queries (line 888)
  - Nested sources (line 894)

  Test Coverage

  Updated existing test:
  - OrderByWithInnerQuerySimpleSlices: Updated to expect sq_0_0 instead of sq_0 for the collision case, with
  clear documentation of why the collision occurs

  Added four new regression tests:

  1. IntervalIntersectNoDuplicateCteNames: Directly reproduces the original bug scenario with id="2" colliding
  with index 2, verifies exactly one sq_2 AS appears and collision is resolved
  2. NestedQueriesWithIdCollisions: Tests multiple nesting levels where root (index 0) and inner query (id="0")
   would collide
  3. ComplexIntervalIntersectWithAggregationNoIds: Stress-tests auto-generated names with complex queries
  involving interval intersect, filters, aggregation, and ordering
  4. DeeplyNestedQueriesNoIds: Tests 3+ levels of nesting to ensure uniqueness is maintained across many CTEs

  Examples

  Before (broken):
  WITH sq_2 AS (
    SELECT * FROM thread_or_process_slice WHERE dur >= 0 LIMIT 10
  ),
  sq_2 AS (  -- DUPLICATE! Invalid SQL
    SELECT * FROM thread_or_process_slice WHERE dur >= 0 LIMIT 10
  )
  SELECT * FROM _interval_intersect!(...)

  After (fixed):
  WITH sq_2 AS (
    SELECT * FROM thread_or_process_slice WHERE dur >= 0 LIMIT 10
  ),
  sq_2_0 AS (  -- Collision resolved with suffix
    SELECT * FROM thread_or_process_slice WHERE dur >= 0 LIMIT 10
  )
  SELECT * FROM _interval_intersect!(...)

  Performance Impact

  - Minimal: O(log n) lookup per query for collision detection
  - The set is scoped to a single Generate() call, so no cross-query overhead
  - Typical queries have few CTEs, making the performance impact negligible

  Backward Compatibility

  - Queries without collisions generate identical SQL (no suffixes added)
  - Only collision cases get renamed, which is correct behavior vs. the previous broken state
  - No API changes

  ---
  Test plan

  - All 109 existing structured query generator tests pass
  - Added 4 new regression tests covering collision scenarios
  - Manually verified collision resolution produces valid, unique CTE names

### DIFF
--- a/src/trace_processor/perfetto_sql/generator/structured_query_generator_unittest.cc
+++ b/src/trace_processor/perfetto_sql/generator/structured_query_generator_unittest.cc
@@ -3668,4 +3668,87 @@ TEST(StructuredQueryGeneratorTest, DeeplyNestedQueriesNoIds) {
   EXPECT_NE(third_as, std::string::npos) << "Expected at least 3 nested CTEs";
 }
 
+// Test that string IDs (non-numeric) are used directly in table names
+TEST(StructuredQueryGeneratorTest, StringIdInTableName) {
+  StructuredQueryGenerator gen;
+  // Test a query with a string ID like "foo"
+  auto proto = ToProto(R"(
+    id: "foo"
+    table {
+      table_name: "slice"
+    }
+    filters {
+      column_name: "dur"
+      op: GREATER_THAN
+      int64_rhs: 1000
+    }
+  )");
+  auto ret = gen.Generate(proto.data(), proto.size());
+  ASSERT_OK_AND_ASSIGN(std::string res, ret);
+
+  // The table name should be "sq_foo"
+  // Even though this is the root query, it creates a CTE named sq_foo
+  ASSERT_THAT(res, EqualsIgnoringWhitespace(R"(
+    WITH sq_foo AS (
+      SELECT * FROM slice WHERE dur > 1000
+    )
+    SELECT * FROM sq_foo
+  )"));
+}
+
+// Test nested queries with string IDs
+TEST(StructuredQueryGeneratorTest, NestedQueriesWithStringIds) {
+  StructuredQueryGenerator gen;
+  auto proto = ToProto(R"(
+    id: "outer"
+    inner_query {
+      id: "inner"
+      table {
+        table_name: "test_table"
+      }
+    }
+  )");
+  auto ret = gen.Generate(proto.data(), proto.size());
+  ASSERT_OK_AND_ASSIGN(std::string res, ret);
+
+  // The inner query should have the string-based name "sq_inner"
+  // The outer query is root, so it doesn't create its own CTE
+  ASSERT_THAT(res, EqualsIgnoringWhitespace(R"(
+    WITH sq_inner AS (
+      SELECT * FROM test_table
+    )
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/trace_processor/perfetto_sql/generator/structured_query_generator_unittest.cc [hunk 1]
```diff
@@ -2126,8 +2126,11 @@ TEST(StructuredQueryGeneratorTest, OrderByWithInnerQuerySimpleSlices) {
   )");
   auto ret = gen.Generate(proto.data(), proto.size());
   ASSERT_OK_AND_ASSIGN(std::string res, ret);
+  // Note: The inner_query has id="0" which would collide with the root query's
+  // auto-generated name (sq_0), so the collision avoidance renames it to
+  // sq_0_0.
   ASSERT_THAT(res, EqualsIgnoringWhitespace(R"(
-    WITH sq_0 AS (
```
LalitMaganti, line 2130: Can we drop the inner layer of conflict?

LalitMaganti (approved):

PR #3588, opened 2025-11-07
exp: Introduce Filtering Node and migrate all filtering logic

aMayzner:
1. New FilterNode Architecture
  - Creates a standalone FilterNode class that encapsulates filtering logic
  - Filters are now first-class nodes in the query graph rather than properties on other nodes
  - Registered in core_nodes.ts:165 as a new node type

  2. Smart Filter Management (explore_page.ts:327)
  - handleFilterAdd() intelligently reuses existing FilterNodes when possible:
   - If source is already a FilterNode → adds filter to it
   - If source has one child FilterNode → adds filter to child
   - Otherwise → creates new FilterNode after source
  - Changes onRemoveFilter → onFilterAdd callback pattern

  3. Serialization Updates (json_handler.ts)
  - Removes filters property from MergeNode and UnionNode serialization
  - Adds FilterNode serialization/deserialization support
  - Extensive test updates (34 test modifications) to reflect new architecture

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/explore_page.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/explore_page.ts
@@ -19,9 +19,11 @@ import {Builder} from './query_builder/builder';
 import {
   QueryNode,
   QueryNodeState,
+  NodeType,
   addConnection,
   removeConnection,
 } from './query_node';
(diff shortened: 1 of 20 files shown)


stevegolton (approved):
Nice nice nice!

PR #3611, opened 2025-11-10
perfetto_cmd: add --notify-fd flag

dreveman:
This can be used to wait for all data sources to be started without running perfetto command in background mode.

Useful when you want to capture the output of the command or make sure that it terminates when the calling process terminates.

### DIFF
--- a/src/perfetto_cmd/perfetto_cmd.cc
+++ b/src/perfetto_cmd/perfetto_cmd.cc
@@ -145,6 +145,12 @@ Usage: %s
                              data sources to be started before exiting. Exit
                              code is zero if a successful acknowledgement is
                              received, non-zero otherwise (error or timeout).
+  --notify-fd FD           : Like --background-wait, but instead of daemonizing
+                             and waiting for data sources to be started before
+                             exiting, writes status and closes FD after waiting
+                             for data sources to be started. Writes a zero
+                             byte if a successful acknowledgement is received,
+                             non-zero otherwise (error or timeout).
   --clone TSID             : Creates a read-only clone of an existing tracing
                              session, identified by its ID (see --query).
   --clone-by-name NAME     : Creates a read-only clone of an existing tracing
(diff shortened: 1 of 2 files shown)


dreveman:
Either option is fine with me but using an existing pattern is always good so prefer option 2 that uses the systemd semantics. I guess a bit more complicated to use from a script but I think that's fine.

I'll update this PR later today for option 2 unless you think of a better approach before then.

primiano:
Hold on a sec, let me think a bit more about this before I ask you to do one more CL :)

the problem with option2 is that if you don't use systemd, synchronizing with a socket is more annoying (try to script that).

What is your e2e use case here out of curiosity?

LalitMaganti:
Personally, I like a lot more option 1). I think it would come in handy for other people who want to do non-trivial daemonization/integration in other tools.

primiano:
Okay had some more thoughts.

the NOTIFY_SOCKET is more a solution for traced & traced_probes that are actual daemons [1]. It would not help much on the cmdline client as:

1) Nobody is going to bother creating a systemd file for it
2) It would be unusable on Android

I'd settle with this, if it works for you:

--notify-wait=/path/to/file.pid

- If the file doesn't exist it creates it, otherwise just opens it and truncates it
- When we NotifyBgProcessPipe, just write getpid() into that file and close it.

This is more consistent with the "echo pid" of --background-wait.
This allows people to easily kill prematurely the session

Any other solution I could think of ended up scope creeping and creating some sort of new logging format... better to avoid.

[1] @sashwinbalaji is actually taking care of the debianization of them as we speak

dreveman:
Ok, I can change it to `--notify-wait` as you described but not sure writing the pid makes sense without using `--background`. Using `--background` is part of what I want to avoid.

In my use-case I'm spawning the perfetto command from a python script and I'd like to accomplish three things:
1. Easily capture stdout and stderr. Not running it in the background makes that trivial.
2. Reliably kill the perfetto command when the python script terminates or is killed. Having it be a child process and not run in the background accomplishes that.
3. Have the python script be able to wait for all data sources to be started.

So I can change this to `--notify-wait` and still write the pid even if the caller should already be aware of that in the non-background case. And I guess I write 0 or nothing in case of an error. Does that sound good?

dreveman:
Great. Yes, I was just about to post a comment about how it would be best to keep the FD as it's awkward to pass a file to a forked child process and somehow have it already be open to ensure that I'd be notified when it was closed.

FYI, I found a few optional cleanups to this but I'll upload those as a separate PR.

PR #3624, opened 2025-11-11
Use machine table to identify machine ids in Android logs

jahdiel-alvarez:
When ftrace isn't set as a data source during a multi-VM tracing session, the CPU table won't be populated. This means that if Android logs were indeed set as a data source, then the AndroidLog panel won't show the machine id column and filter button, even though there are logs from multiple machines, because it uses the CPU table to identify the machines involved in the table.

Instead we use the machine table which stores all the machines involved in the multi-VM trace.

Bug: 459819089

### DIFF
--- a/ui/src/plugins/com.android.AndroidLog/index.ts
+++ b/ui/src/plugins/com.android.AndroidLog/index.ts
@@ -47,16 +47,14 @@ interface AndroidLogPluginState {
 
 async function getMachineIds(engine: Engine): Promise<number[]> {
   // A machine might not provide Android logs, even if configured to do so.
-  // Hence, the |process| table might have ids not present in the logs. Given this
+  // Hence, the |machine| table might have ids not present in the logs. Given this
   // is highly unlikely and going through all logs is expensive, we will get
-  // the ids from |process|, even if filter shows ids not present in logs.
-  const result = await engine.query(
-    `SELECT DISTINCT(machine_id) FROM process ORDER BY machine_id`,
-  );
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.android.AndroidLog/index.ts
```diff
@@ -47,11 +47,11 @@ interface AndroidLogPluginState {
...
 async function getMachineIds(engine: Engine): Promise<number[]> {
   // A machine might not provide Android logs, even if configured to do so.
-  // Hence, the |cpu| table might have ids not present in the logs. Given this
+  // Hence, the |process| table might have ids not present in the logs. Given this
   // is highly unlikely and going through all logs is expensive, we will get
-  // the ids from |cpu|, even if filter shows ids not present in logs.
+  // the ids from |process|, even if filter shows ids not present in logs.
   const result = await engine.query(
-    `SELECT DISTINCT(machine_id) FROM cpu ORDER BY machine_id`,
+    `SELECT DISTINCT(machine_id) FROM process ORDER BY machine_id`,
```
LalitMaganti: Can we please query the machine table instead?

jahdiel-alvarez: I swear I had checked for the machine table and hadn't found it. I was only seeing the info in metadata. Let me fix that.

jahdiel-alvarez: Done

PR #3625, opened 2025-11-11
perfetto_cmd: minor code cleanups after adding `--notify-fd` flag

dreveman:
Add missing comment and use a more appropriate name for BgProcessStatus enum now that it's used without a background process.

### DIFF
--- a/src/perfetto_cmd/perfetto_cmd.h
+++ b/src/perfetto_cmd/perfetto_cmd.h
@@ -109,13 +109,20 @@ class PerfettoCmd : public Consumer {
 
   void ReadbackTraceDataAndQuit(const std::string& error);
 
-  enum BgProcessStatus : char {
-    kBackgroundOk = 0,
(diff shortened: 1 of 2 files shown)


src/perfetto_cmd/perfetto_cmd.h
```diff
@@ -109,21 +109,28 @@ class PerfettoCmd : public Consumer {
...
 
-  void Notify(BgProcessStatus status);
+  // Used to implement the --notify-fd flag.
+  //
+  // Signals client by writing to FD (if there is one) that data sources has
+  // started (or failed to start).
+  //
+  // Only the first time this function is called is significant. Further calls
+  // will have no effect.
+  void NotifyFd(WaitStatus status);
```
primiano, line 125: Thank you :) I wanted to suggest this (+"Fd") in the other CL, but then I felt we went too much in circles already and was not big enough .

PR #3627, opened 2025-11-11
rust-sdk: bump perfetto-sdk-sys version number

dreveman:
This is in preparation for a v53 release and publishing a crate that includes amalgamated shared library sources that matches that release.


dreveman:
not really critical to do this before releases but I figured I'd do it now so all I have to do after the release tag is added is run `crate publish`..

PR #3643, opened 2025-11-12
ui: Move all omnibox related logic into omnibox.ts

stevegolton:
Currently a lot of the omnibox business logic is contained in the `UIMain` component. The `Omnibox` component in `omnibox.ts` is a purely a presentational widget.

The issue is that `omnibox.ts` is where one would expect to find all the omnibox logic, both business and presentational.

This patch moves all the business logic into a new component called `Omnibox`, and the original `Omnibox` presenational widget has been renamed to `OmniboxWidget`. Both classes appear in the same file - `omnibox.ts`.

This is a pure architectural change and contains 0 functional changes.


primiano (approved):
awesome

PR #3650, opened 2025-11-13
install-build-deps: --rust-toolchain -> --rust

primiano:


LalitMaganti (approved):

tools/install-build-deps
```diff
@@ -862,7 +862,7 @@ def Main():
       choices=['arm64', 'x64'],
       help='Override the autodetected build CPU architecture')
   parser.add_argument(
-      '--rust-toolchain',
+      '--rust',
```
dreveman: you'll need to change `args.rust_toolchain` to `args.rust` below too

primiano: D'oh! Done
This is the classic "I can certainly make this 2 line change without trying it out" :)

PR #3653, opened 2025-11-13
Add fwtp_perfetto_counter ftrace event

estaats-google:
This ftrace event reports counter events in Pixel firmware.

This ftrace event is only on Pixel kernels and is not upstream.

Bug: 386832373
Test: Verified FWTP counter tracepoints may be collected with Perfetto
 and displayed with traceconv.


LalitMaganti (approved):

LalitMaganti:
FYI you might be interested in knowing about https://perfetto.dev/docs/getting-started/ftrace#part-c-simple-slice-counter-visualisations-without-modifying-perfetto-code-kernel-track-events-. It's a way to emit ftrace events in a form Perfetto understands without needing to land code in trace processor.

I think it won't work for this usecase however because you have custom timestamps rather than using the ftrace one which kernel track events do not support.

estaats-google:
> FYI you might be interested in knowing about https://perfetto.dev/docs/getting-started/ftrace#part-c-simple-slice-counter-visualisations-without-modifying-perfetto-code-kernel-track-events-. It's a way to emit ftrace events in a form Perfetto understands without needing to land code in trace processor.
> 
> I think it won't work for this usecase however because you have custom timestamps rather than using the ftrace one which kernel track events do not support.

Thanks for the link. You are correct about the timestamps. We want the timestamp reported by Perfetto to be the timestamp of the tracepoint when it was recorded in the firmware, not when it was read from the firmware by the kernel module and logged to ftrace. I'll have a follow up change that adds timestamp conversion.

PR #3655, opened 2025-11-13
base: add Basename/Dirname

primiano:
Splitting this from PR #2931


include/perfetto/ext/base/file_utils.h
```diff
@@ -96,6 +95,30 @@ bool FileExists(const std::string& path);
 // extension.
 std::string GetFileExtension(const std::string& filename);
 
+// Returns the basename component of a path (the final component after the last
+// directory separator). Behaves like man 2 basename, but works with both '/'
```
rsavitski, line 99: nit: it's not a syscall, so "man 2" is incorrect, but I'm sending this off to autosubmit since it's not worth even the reupload.

PR #3664, opened 2025-11-13
tp: refactor trace_processor_shell to extract library target

LalitMaganti:
In Google3, we want a version of shell which is tightly integrated with
the rest of Google infra (e.g. the SQL modules, the internal filesystems
etc). But we don't want to maintain a separate shell.

The best way around this is to extract a thin wrapper interface for
shell which allows restricted customization points of shell without
needing to duplicate lots of code

### DIFF
--- a/src/trace_processor/trace_processor_shell.cc
+++ b/src/trace_processor/trace_processor_shell.cc
@@ -2180,7 +2180,7 @@ base::Status TraceProcessorShell::Run(int argc, char** argv) {
   return base::OkStatus();
 }
 
-TraceProcessorShell_PlatformInterface::~TraceProcessorShell_PlatformInterface() =
-    default;
+TraceProcessorShell_PlatformInterface::
+    ~TraceProcessorShell_PlatformInterface() = default;
 
 }  // namespace perfetto::trace_processor


include/perfetto/ext/trace_processor/trace_processor_shell.h
```diff
@@ -0,0 +1,81 @@
...
+#include "perfetto/base/status.h"
+#include "perfetto/trace_processor/basic_types.h"
+#include "perfetto/trace_processor/trace_processor.h"
+
+namespace perfetto::trace_processor {
+
+// Class for implementing a trace processor shell.
+//
+// Only visible for embedders who want very fine grained control of how shell
+// integrates with other systems.
```
primiano: Add a comment saying something like: "this is mainly for google internal builds where we wrap TP shell with custom gFile handlers"
so it's clearer what this is for

src/trace_processor/BUILD.gn
```diff
@@ -258,7 +259,27 @@ if (enable_perfetto_trace_processor_sqlite) {
...
     }
+    public_deps =
+        [ "../../include/perfetto/ext/trace_processor:trace_processor_shell" ]
     sources = [ "trace_processor_shell.cc" ]
+  }
+
+  # Library target for the trace processor shell.
+  # Allows embedders (e.g. Google3) to customize the shell executable's
+  # interaction with the environment.
+  static_library("trace_processor_shell_lib") {
```
primiano: I think the _src traget should be folded into this, so:

shell.exe depdns only on shell_lib
shell_lib depends on lib

src/trace_processor/trace_processor_shell_main.cc
```diff
@@ -0,0 +1,64 @@
...
+#include <string>
+
+#include "perfetto/ext/trace_processor/trace_processor_shell.h"
+#include "perfetto/trace_processor/basic_types.h"
+#include "perfetto/trace_processor/trace_processor.h"
+#include "src/trace_processor/read_trace_internal.h"
+
+namespace perfetto::trace_processor {
+
+class DefaultPlatformInterface : public TraceProcessorShell::PlatformInterface {
```
primiano: move this in the "shell_lib" and expose a TraceProcessorShell::CreateDefaultPlatform() static method or similar?

LalitMaganti:
PTAL: fixed all comments.

include/perfetto/ext/trace_processor/trace_processor_shell.h
```diff
@@ -0,0 +1,87 @@
...
+
+#ifndef INCLUDE_PERFETTO_EXT_TRACE_PROCESSOR_TRACE_PROCESSOR_SHELL_H_
+#define INCLUDE_PERFETTO_EXT_TRACE_PROCESSOR_TRACE_PROCESSOR_SHELL_H_
+
+#include <cstddef>
+#include <functional>
+#include <memory>
+#include <string>
+#include <string_view>
+#include "perfetto/base/status.h"
```
primiano, line 26: nit: \n between < and " headers

include/perfetto/ext/trace_processor/trace_processor_shell.h
```diff
@@ -0,0 +1,87 @@
...
+// Class for implementing a trace processor shell.
+//
+// Only visible for embedders who want very fine grained control of how shell
+// integrates with other systems. This is mainly for Google internal builds
+// where we wrap trace_processor_shell with custom gfile handlers, integration
+// with internal packages etc.
+class TraceProcessorShell {
+ public:
+  // Abstract class for platform specific operations.
+  class PlatformInterface {
```
primiano: A suggestion: I got bitten over time by nested interfaces. (mainly because they can't be fwd declared and other similar ergonomic things)

What I usually do these days is:

1. Define it outside. 
class TraceProcessorShell_PlatformInterface {
  ...
}

2. In class TraceProcessorShell { 
  using PlatformInterface = TraceProcessorShell_PlatformInterface;


You can even move it to the bototm by first fwd declaring, like

class TraceProcessorShell_PlatformInterface;

class TraceProcessorSHell {
...
}

...
class TraceProcessorShell_PlatformInterface{
...
}

PR #3665, opened 2025-11-13
tp: improve documentation of TraceProcessor class

LalitMaganti:
Push down the functions in TraceProcesssorStorage to TraceProcessor
for better comments and allow TraceProcessor to be self contained.


LalitMaganti:
Discussed offline, the current approach I think is the way to go as we need backcompat and C++ overloading rules means we have to dfine it twice.

PR #3668, opened 2025-11-14
Replace `text-weight: bolder` with `500` to improve text clarity

stevegolton:
Currently we use `font-weight: bolder` in a few places to make some text bolder relative to the parent. This is fine but the actual implementation of bolder is uses banding to determine which level is chosen. See: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-weight#meaning_of_relative_weights. In short, the only outputs of bolder are 400, 700, and 900.

Since we only have a handful of Roboto weights available, the boldest of which being 500, it makes sense to just hard code these fonts to weight 500 instead of using bolder.

### DIFF
--- a/ui/src/assets/components/pivot_table.scss
+++ b/ui/src/assets/components/pivot_table.scss
@@ -16,7 +16,7 @@
 
 .pf-pivot-table {
   &__total-values {
-    font-weight: bolder;
+    font-weight: 500;
   }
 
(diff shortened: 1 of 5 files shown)


stevegolton:
Since https://github.com/google/perfetto/pull/3997 we no longer need this as all weights are now rendered correctly with variable weight fonts.

PR #3672, opened 2025-11-14
exp: Interval intersect provides full final columns set

aMayzner:
Interval intersect provides full set of child columns as the result, not only the columns from the first source


stevegolton (approved):
Nice!

PR #3687, opened 2025-11-14
exp: Widgets page stub

aMayzner:
This commit refactors the ModifyColumnsNode component to extract reusable UI widgets and improve code organization. 

So far the only widgets are from the modifyColumnsNode and the code is messy, but I wanted to use it as a starting point for other changes. All of the other nodes can only either reuse widgets from here, or add new ones.

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.scss
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.scss
@@ -38,7 +38,7 @@
 }
 
 // Section widget
-.pf-section {
+.pf-exp-section {
   padding: 1rem;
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.scss
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+.pf-exp-button-group {
+  display: flex;
+  gap: 8px;
+}
+
+// Section widget
+.pf-section {
```
stevegolton: This rule is too generic - it collides with our Section widget which is why the integration tests are failing.

stevegolton (approved):

PR #3688, opened 2025-11-16
tp: fix binary size regression in Chrome due to import logs

LalitMaganti:
Turns out adding a dep on stats arrays is a bad idea as it increases
binary sizes unnecessarily. Instead, just store the key and lookup the
mapping to severity/name in SQL.

### DIFF
--- a/src/trace_processor/importers/common/import_logs_tracker.cc
+++ b/src/trace_processor/importers/common/import_logs_tracker.cc
@@ -22,9 +22,7 @@
 #include <utility>
 
 #include "perfetto/base/logging.h"
-#include "perfetto/ext/base/string_view.h"
 #include "src/trace_processor/importers/common/args_tracker.h"
-#include "src/trace_processor/storage/stats.h"
 #include "src/trace_processor/storage/trace_storage.h"
 #include "src/trace_processor/tables/metadata_tables_py.h"
 #include "src/trace_processor/types/trace_processor_context.h"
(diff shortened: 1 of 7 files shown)


src/trace_processor/importers/common/import_logs_tracker.cc
```diff
@@ -33,28 +31,20 @@ namespace perfetto::trace_processor {
...
-      severity_data_loss_id_(context_->storage->InternString("data_loss")),
-      severity_error_id_(context_->storage->InternString("error")) {}
+    : context_(context), trace_id_(trace_id) {}
 
 void ImportLogsTracker::RecordImportLog(
     size_t stat_key,
     std::optional<int64_t> timestamp,
     std::optional<int64_t> byte_offset,
     std::function<void(ArgsTracker::BoundInserter&)> args_callback) {
-  PERFETTO_CHECK(stats::kSources[stat_key] == stats::Source::kAnalysis);
```
sashwinbalaji, line 47: why is this check removed ?

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
```diff
@@ -324,3 +324,42 @@ SELECT
   tag,
   msg
 FROM __intrinsic_android_logs;
+
+-- Materialized mapping from stat key to severity and name.
```
sashwinbalaji, line 328: nit: mentioning the need for materiazling would be helpful. (i believe this is done because stats is virtual so to avoid re-calculating everytime right  ?)

sashwinbalaji (approved):

src/trace_processor/importers/common/import_logs_tracker.cc
LalitMaganti: Because we cannot lookup severity anymore without the array needing to be included, bloating binary size.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
LalitMaganti: No it's to avoid the distinct as the below is a view!

src/trace_processor/importers/common/import_logs_tracker.cc
sashwinbalaji: 🤦

PR #3694, opened 2025-11-17
ui: Hide material design icons until font load

primiano:
Hides the font with a max-width=1em until the font
is actually loaded (waits max 10s)

Bug: b/458318781

### DIFF
--- a/ui/src/base/utils.ts
+++ b/ui/src/base/utils.ts
@@ -85,7 +85,7 @@ export function createProxy<T extends object>(
 }
 
 // A promisified variant setTimeout().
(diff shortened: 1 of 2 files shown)


ui/src/base/utils.ts
```diff
@@ -83,3 +83,10 @@ export function createProxy<T extends object>(
     },
   }) as T;
 }
+
+// A promisified variant setTimeout().
+export function timeoutPromise(ms: number): Promise<void> {
```
stevegolton: `sleep()`?

:P

primiano: Done

PR #3695, opened 2025-11-17
ui: Show icons only when sidebar collapsed

aMayzner:
## Summary

This change converts the sidebar from a hide/show toggle to a collapsed/expanded state. When collapsed, the sidebar displays icons only instead of completely disappearing, providing persistent access to navigation while maximizing screen space for trace content.

  ## Changes

  - Replace `visible/hidden` state  change to `collapsed/expanded` throughout the codebase 
  - Implemented collapsed state styling that shows icons-only at 60px width with centered layout
  - Added section icons to all sidebar sections (navigation, current trace, examples, settings, support, convert)
  - Added tooltips to all sidebar items when collapsed for discoverability
  - Updated example traces to use platform-specific icons (Android/Web) instead of generic description icon
  - Removed topbar styling adjustment that was specific to hidden sidebar state


LalitMaganti:
I have some feedback on this change:
1) I don't think it makes sense for the header titles to be clickable in collapsed mode
2) I don't think it makes sense to have *all* icons visible in collapsed mode. I think we should have some opinionated defaults of just what is needed, mainly around working with the existing trace
3) We need to still support a way we can *fully* hide the sidebar otherwise I'm 100% sure we'll get complaints.

PR #3699, opened 2025-11-17
ui: gzip XOR no-cache in build_all_channels.py

primiano:
The previous PR (#3689) missed -z. Turns out that transparent
gzip still happens at the GFE level, however -j does NOT store
data gzip-encoded in the bucket.
I tested this on a staging instance and seems to still work.

Bug: b/458318781

### DIFF
--- a/ui/release/build_all_channels.py
+++ b/ui/release/build_all_channels.py
@@ -134,15 +134,15 @@ def main():
   print('===================================================================')
   print('Uploading to gs://%s' % BUCKET_NAME)
   print('===================================================================')
-  # Do NOT set cache-headers here, it messes up with GCS' handling of
-  # transparent gzip content-encoding. Cache-control headers are set instead by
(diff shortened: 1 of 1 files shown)


primiano:
I will wait until tomorrow before deploying. I tested it, but better deploying when I am around.

LalitMaganti (approved):

ui/release/build_all_channels.py
```diff
@@ -137,7 +137,7 @@ def main():
   # Do NOT set cache-headers here, it messes up with GCS' handling of
   # transparent gzip content-encoding. Cache-control headers are set instead by
   # the GAE instance (see /infra/ui.perfetto.dev/appengine/main.py).
-  cp_cmd = ['gsutil', '-m', 'cp', '-j', 'html,js,json,css,wasm,map']
```
LalitMaganti, line 140: Leaving comment to prevent submit.

PR #3701, opened 2025-11-18
perfetto: add support for amalgamating C SDK

LalitMaganti:
This CL adds support for generating a single header + cc file for the
C SDK. This requires making sure that only C code is visible in the
headers: note that the implementation is still a C++ file because it
still relies on the C++ SDK under the hood: only the API is C clean.

I've tested this and made sure it compiles with an example program.

Fixes: https://github.com/google/perfetto/pull/3701

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -223,39 +223,6 @@ if (is_android && (perfetto_build_standalone || perfetto_build_with_android)) {
   all_targets += [ "test/vts:perfetto_vts_deps" ]
 }
 
-group("all") {
-  testonly = true  # allow to build also test targets
-  deps = all_targets
-}
-
-# This target is used when running ninja without any argument (by default would
-# build all reachable targets). This is mainly used to prevent the UI being
-# built when running ninja -C out/xxx.
-# This has effect only in standalone builds, no effect on chromium builds.
-# Chromium's "all" target depends on our "all" target above. However chromium's
-# "default" target depends on any target that we cause to be discovered by
-# depending on other GN files.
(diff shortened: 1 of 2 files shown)


BUILD.gn
```diff
@@ -302,6 +302,19 @@ if (!build_with_chromium) {
     sources = [ "include/perfetto/tracing.h" ]
     assert_no_deps = [ "gn:protobuf_lite" ]
   }
+
+  # Static library for the C SDK, used for amalgamation.
+  static_library("libperfetto_c_sdk") {
```
primiano: out of curisoity why do you need this?
can't you reuse the src/shared_lib:shared_lib directly in the amalgamator?

primiano (approved):

LalitMaganti:
@dreveman you might be interested in this change for the Rust SDK.

dreveman:
Great! Thanks for the heads up. I'll upload a PR later today that makes the Rust SDK use this instead of the current more complicated instructions for producing amalgamated source for the -sys crate.

Is the plan to checkin source for this in release branches in a similar way to the C++ sdk?

LalitMaganti:
We're actually planning on moving away from checking the SDKs (even the C++ one)  into the tree in favour of making it a release artifact on GitHub and Cloud Storage. 

I'll post an rfc about our release changes later before the next release.

dreveman:
Ok, sounds good. That's how to works for the published Rust SDK crates already but I'll keep an eye out for the rfc in case some adjustments are needed to align with that.

PR #3707, opened 2025-11-18
ui: fix removing selection when popup is open

LalitMaganti:
This CL fixes both mouse and 'Escape' keyboard events when a popup is open to prevent it from removing the selection in the UI.

### DIFF
--- a/ui/src/plugins/dev.perfetto.WidgetsPage/demos/popup_demo.ts
+++ b/ui/src/plugins/dev.perfetto.WidgetsPage/demos/popup_demo.ts
@@ -14,7 +14,11 @@
 
 import m from 'mithril';
 import {Button, ButtonBar} from '../../../widgets/button';
-import {OutsideClickHandling, Popup, PopupPosition} from '../../../widgets/popup';
+import {
+  OutsideClickHandling,
+  Popup,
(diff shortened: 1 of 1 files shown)


ui/src/widgets/popup.ts
```diff
@@ -483,6 +483,8 @@ export class Popup implements m.ClassComponent<PopupAttrs> {
 
   private handleDocMouseDown = (e: Event) => {
     if (this.closeOnOutsideClick && !this.eventInPopupOrTrigger(e)) {
+      e.stopPropagation();
```
stevegolton: I don't think there's any point in doing both stopPropagation() and preventDefault(). Seems like a shotgun approach. I would like to see some justification for how the popup interacts with the HotkeyContext.

LalitMaganti:
> This doesn't seem to work.

Hmmm so I tested with a different popup (debug track popup) and that closed correctly. But you're right. Will look into it.

LalitMaganti:
OK this should be good now. It took me best part of 4 hours to get everything working but got there in the end.

LalitMaganti:
We're pausing on this change there is not consensus on what the right behaviour should be. Putting it back to draft.

Issue #3705, opened 2025-11-18
packet timestamp_clock_id default does not match trace proto documentation


cphlipot1:
https://github.com/google/perfetto/blame/db2bc10622b44d1ed994a6401f97be00e9de04ae/src/trace_processor/importers/proto/proto_trace_reader.cc#L349 is setting the timestamp clock id to zero when not present in the packet and no default is set.

However, in https://github.com/google/perfetto/blob/db2bc10622b44d1ed994a6401f97be00e9de04ae/protos/perfetto/trace/trace_packet.proto#L128-L130 it says packet should default to clock boot time if not set

Currently it looks like the implementation and the docs do not match.

I discovered this while implementing #3703, where if i do not set `timestamp_clock_id: 6` in each packet than the test would fail. however based on my reading of the docs this should default to 6 (CLOCK_BOOTTIME) and setting it explicitly should not be necessary.

I would suggest either updating docs or changing the code to match existing docs, assuming there is no user error in this case.

LalitMaganti:
I think @jahdiel-alvarez discovered this as well and I think our agreement was we should change the code to match the docs as making it match the trace time clock is probably more implicit than we'd like.

The only concern is that it's possible that folks were relying on the timestamps implicitly matching the trace clock: I think we need to shore up our handling of clocks in general further in the multi-trace merging case (including allowing for "trace scoped" clocks).

cphlipot1:
> The only concern is that it's possible that folks were relying on the timestamps implicitly matching the trace clock

this was my concern as well. 

Out of curiosity, I did a quick test with changing that from zero to BUILTIN_CLOCK_BOOTIME, and it broke the `test_sched_remote_clock_sync_vcpu0` diff test.

As far as i can tell all timestamps are the same and it only failed due to utid column changing, meaning we probably same some threads sooner or later than without this change.

doing an inspection of multi_machine_trace.pftrace it appears there are process_tree packets which have no trace packet default set, and no timestamp_clock_id set, and this change probably causes them to be processed sooner or later impacting which threads get assigned which utids.

I'm unclear if that difference is benign or not, but it at least clear that real non-synthetic traces are impacted by code changes to this code path, which i did not expect until i ran this experiment.

LalitMaganti:
FWIW real multi-machine traces clock handling of unset packets is buggy and that's exactly why @jahdiel-alvarez was asking me about this. I think in practice for Perfetto generated traces things are already a bit broken.

[...]

PR #3709, opened 2025-11-18
tp: fix default timestamp_clock_id

cphlipot1:
in the trace packet proto it says the if there is no trace packet
defaults that specify the default clock id for the squence and the
packet itself doesn't specify it, then the timestamp should be
CLOCK_BOOTIME. however currently it is set to zero instead.

Update the implementation to reflect the docs

BUG: #3705


cphlipot1:
I see screenshot tests are failing. 

From looking at the failing test, the "actual" screenshots appear to be improvements over "expected"

I see:
- processes inheriting name properly instead of the name of its threads (eg. zygote_64 instead of main, cameraserver instead of binder:25867_2)
- selection bar at top indicates we are zoomed into a larger percentage of the trace, probably because in "expected" version some event was placed far in the past before trace start, and this change fixes that due to proper clock conversion.

That said i think those are stored in gcs if i'm not mistaken and i don't think i have permission to upload new screenshots there. 

What is the best way to proceed here?

LalitMaganti:
Once Jahadiel is happy with the change, I can fix the screenshot tests.

jahdiel-alvarez:
@LalitMaganti do we want to change this for all traces? This will disrupt any data sources relying on the implicit trace clock. We had discussed about doing this only for traces coming from secondary machines in multi-machine traces.

LalitMaganti:
Personally, I'm happy to have the change and deal with the fallout of it if necessary. I think it's somewhat unlikely anyone is relying on this and, if they are, we can figure out what to do then.

LalitMaganti (approved):

LalitMaganti:
Updated the screenshot test, LGTM!

cphlipot1:
> Updated the screenshot test, LGTM!

Thanks. it looks like one screenshot test is still failing though for some reason. Again new version looks better to me than old version.

PR #3711, opened 2025-11-18
tp: small polish fixes for flamegraph

LalitMaganti:
* keep the color of the text consistent based on the base color.
  Changing based on the secondary color felt weird and buggy
* Make sure to hide the popup when it goes off screen. Until now,
  you could make the popup do really weird things like croll off screen
  behind the content. Instead hide it properly.

### DIFF
--- a/ui/src/widgets/flamegraph.ts
+++ b/ui/src/widgets/flamegraph.ts
@@ -233,6 +233,7 @@ export class Flamegraph implements m.ClassComponent<FlamegraphAttrs> {
 
   private canvasWidth = 0;
   private labelCharWidth = 0;
+  private canvasRect?: Rect2D;
 
   constructor({attrs}: m.Vnode<FlamegraphAttrs, {}>) {
     this.attrs = attrs;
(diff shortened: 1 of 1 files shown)


ui/src/widgets/flamegraph.ts
```diff
@@ -563,6 +568,29 @@ export class Flamegraph implements m.ClassComponent<FlamegraphAttrs> {
     }
   }
 
+  private isPopupAnchorVisible(): boolean {
```
stevegolton, line 569: I'm not going to block on this but I do wonder whether a more elegant solution to this would be to mount the popup to an OverlayContext inside the scrolling content.

This should "just work" if you wrap the flamegraph content with an OverlayContext widget - make sure it's inside the scrolling container element - though I'm not sure how the popup it's going to behave in regards to positioning, so it may need more work.

Not for now, this is more of a note for the future.

LalitMaganti: Two problems:
1) Scrolling container is inside the virtualcanvas so I would need some way to inject an element there
2) Turns out this logic is far too unnecessarily complex, we already have the canvas rect returned to us by the overlay canvas so we can just use that!

PR #3714, opened 2025-11-18
multi-machine tracing: stop matching DSes across machines by default

rsavitski:
This is a breaking change to the default data source matching behaviour
across remote producers (connected over traced_relay). Before, we
matched across machines, now we match only the host unless the config
sets the new "trace_all_machines" setting to true. In either case,
DataSource.machine_name_filter can be used to set an explicit set of
machines to match.

The current default is surprising because perfetto has historically been
single-machine, so existing users and configs do not expect to suddenly
match more data if a VM with traced_relay is added to their device.

Note that I am keeping the "builtin producer" behaviour of emitting
system_info *and* remote_clock_sync packets for connected traced_relays,
so the trace will reflect that there's a secondary machine even if none
of its producers matched the trace config.

Bug: 460163124
Tested: `atest CtsPerfettoTestCases` on a frankel with a microdroid VM
        running traced_relay + traced_probes

### DIFF
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -3174,6 +3174,7 @@ TracingServiceImpl::DataSourceInstance* TracingServiceImpl::SetupDataSource(
     const auto& cfg_names = cfg_data_source.machine_name_filter();
     bool filter_match =
         NameMatchesFilter(producer->machine_name_, cfg_names, {});
+    // Special case: "host" also always matches the host machine.
     bool host_match =
         is_host_machine && NameMatchesFilter("host", cfg_names, {});
     if (!filter_match && !host_match) {
@@ -3182,15 +3183,14 @@ TracingServiceImpl::DataSourceInstance* TracingServiceImpl::SetupDataSource(
                     producer->machine_name_.c_str());
       return nullptr;
     }
-  } else {
+  } else if (!tracing_session->config.trace_all_machines() &&
+             !is_host_machine) {
     // Default matching behaviour starting from perfetto v54: match only host,
     // unless the config sets a top level flag.
-    if (!tracing_session->config.trace_all_machines() && !is_host_machine) {
-      PERFETTO_DLOG("Data source: %s is filtered out for remote machine: %s",
-                    cfg_data_source.config().name().c_str(),
-                    producer->machine_name_.c_str());
-      return nullptr;
-    }
+    PERFETTO_DLOG("Data source: %s is filtered out for remote machine: %s",
+                  cfg_data_source.config().name().c_str(),
+                  producer->machine_name_.c_str());
+    return nullptr;
   }
 
   // Check producer name filter.


protos/perfetto/config/trace_config.proto
```diff
@@ -779,4 +779,17 @@ message TraceConfig {
...
+  // sources only match the host machine. In either case, an explicit
+  // |DataSource.machine_name_filter| takes priority.
+  //
+  // NB: perfetto versions before v54 do not have this option and match across
+  // machines by default. To be compatible across this version boundary, either
+  // set this field to true, or set an explicit machine_name_filter on all data
+  // sources.
+  //
+  // Introduced in: perfetto v54.
+  optional bool machine_filter_default = 43;
```
primiano: can we call it `bool trace_all_machines` which is a bit easier to understand? (but still retain the current semantic)

src/tracing/service/tracing_service_impl.cc
```diff
@@ -234,8 +234,7 @@ std::tuple<size_t /*shm_size*/, size_t /*page_size*/> EnsureValidShmSizes(
 bool NameMatchesFilter(const std::string& name,
                        const std::vector<std::string>& name_filter,
                        const std::vector<std::string>& name_regex_filter) {
-  bool filter_is_set = !name_filter.empty() || !name_regex_filter.empty();
-  if (!filter_is_set)
+  if (name_filter.empty() && name_regex_filter.empty())
     return true;
```
primiano: Isn't this going to match all machines anyways if you don't specify any filter at all?

rsavitski: This was a no-op change (an application of De Morgan's). NameMatchesFilter() returned true for an empty set of filters. The existing machine filter codepath reused the function, and checked for non-zero filter size at the callsite.

However, I agree that the two callsites should be more similar.

rsavitski:
PTAL

src/tracing/service/tracing_service_impl.cc
```diff
@@ -3166,27 +3165,45 @@ TracingServiceImpl::DataSourceInstance* TracingServiceImpl::SetupDataSource(
...
-    auto cfg_machine_names = cfg_data_source.machine_name_filter();
-    if (!NameMatchesFilter(producer->machine_name_, cfg_machine_names, {}) &&
-        !(machine_id == kDefaultMachineID &&
-          NameMatchesFilter("host", cfg_machine_names, {}))) {
+    // Explicit set of machines to match.
+    const auto& cfg_names = cfg_data_source.machine_name_filter();
+    bool filter_match =
+        NameMatchesFilter(producer->machine_name_, cfg_names, {});
+    bool host_match =
+        is_host_machine && NameMatchesFilter("host", cfg_names, {});
```
primiano, line 3179: Minor thing, but I find this code a bit easier to reason about if you write it as:

```
static const std::string& host = "host";
const std::string& machine_name = is_host_machine ? host : producer->machine_name_;
bool filter_match = NameMatchesFilter(machine_name, cfg_names, {});
if (!filter_match) ...
```

src/tracing/service/tracing_service_impl.cc
```diff
@@ -3166,27 +3165,45 @@ TracingServiceImpl::DataSourceInstance* TracingServiceImpl::SetupDataSource(
...
+    if (!filter_match && !host_match) {
       PERFETTO_DLOG("Data source: %s is filtered out for machine: %s",
                     cfg_data_source.config().name().c_str(),
                     producer->machine_name_.c_str());
       return nullptr;
     }
+  } else {
+    // Default matching behaviour starting from perfetto v54: match only host,
+    // unless the config sets a top level flag.
+    if (!tracing_session->config.trace_all_machines() && !is_host_machine) {
```
primiano: this could be an 

} else if (!tracing_session->config.trace_all_machines() && !is_host_machine) {
  ...
}

to reduce indentation an clutter

(I purely trasnlated a `else { if (x) {}}` into `else if (X) {}`

rsavitski: ah yeah, makes perfect sense.

src/tracing/service/tracing_service_impl.cc
rsavitski: Added a comment to clarify that the host can match in two ways: by name, as well as the magical "host" string. So we need two checks.

PR #3715, opened 2025-11-18
ui: plumb EngineProxy tag in the error

primiano:
- Remove a bunch of unnecessary plumbing by NOT exposing
  the tag argument to the Engine interface (and hence EngineProxy)
  Only Engine, which gets called by many EngineProxy(ies), needs
  to take that argument.
- Properly pass the argument in QueryErrorInfo and report it
  in the crash reports.

Now errors look like this:
```
UI: http://localhost:10000/v53.0-db2bc1062

  File "stdin" line 2 col 11
    SELECT RUN_METRIC(
    ^
  Metric file "android/network_activity_template.sql" line 41 col 5
        fdsfsd,
        ^
no such column: fdsfsd
EngineTag: com.android.AndroidNetwork   <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
Query:

          SELECT RUN_METRIC(
            'android/network_activity_template.sql',
            'view_name', 'android_network_activity_1763482763691',
            'group_by',  'fdsfsd',
            'filter',    'TRUE',
            'idle_ns',   '10e9',
            'quant_ns',  '3e9'
          );
 -  (Error: Traceback (most recent call last):)
 - WasmEngineProxy.query (frontend_bundle.js:217597:14)
 - async EngineProxy.query (frontend_bundle.js:217716:17)
 - async Object.callback (frontend_bundle.js:45604:18)
```

Bug: b/456448612

### DIFF
--- a/ui/src/trace_processor/engine.ts
+++ b/ui/src/trace_processor/engine.ts
@@ -685,9 +685,9 @@ export abstract class EngineBase implements Engine, Disposable {
 
 // Lightweight engine proxy which annotates all queries with a tag
 export class EngineProxy implements Engine, Disposable {
-  readonly tag: string;
   private engine: EngineBase;
   private disposed = false;
+  private tag: string;
 
   get queryLog() {
     return this.engine.queryLog;


ui/src/trace_processor/engine.ts
```diff
@@ -687,8 +685,8 @@ export abstract class EngineBase implements Engine, Disposable {
 
 // Lightweight engine proxy which annotates all queries with a tag
 export class EngineProxy implements Engine, Disposable {
+  readonly tag: string;
```
stevegolton: Out of curiosity, who needs to see this tag externally? No one should really know they're using an EngineProxy, they should just get given an Engine.

primiano: oh you are absolutely right. I was using it differntly in another attempt, until I figured it didn't need to get exposed in the Engine interface (but then forgot to remove it here).

PR #3717, opened 2025-11-18
Add basic test to validate multi-device power rail traces.

cphlipot1:
Add a basic test to validate functionality power rail traces in the context of a multi-machine trace


cphlipot1:
looks like this is a dupe of #3703

PR #3447, opened 2025-10-29
Rename rust-sdk owners file

sashwinbalaji:
Internal android copybara is failing with below error, so to fix it rename it as CODEOWNERS

```
the domain of the code owner email 'dreveman@gmail.com' in '/contrib/rust-sdk/OWNERS' is not allowed for code owners.
```

PR #3723, opened 2025-11-18
perfetto: switch to distributed github owner files

LalitMaganti:
This CL implements the flow described in [1]. It makes it possible to
maintain owners files without needing to touch the central CODEOWNERS
file.

A PR with update is sent every Monday, Wednesday and Friday.

[1] https://github.com/google/perfetto/pull/3447#issuecomment-3462769450

CHROME_STDLIB_MANUAL_ROLL=intentional change, copybara change in cl/833999390

### DIFF
--- a/OWNERS.github
+++ b/OWNERS.github
@@ -1,0 +1,20 @@
+# This file define code owners for GitHub pull requests
+# NOTE: The last matching rule is the one that defines the owner.
+
+# General repo-wide owners
+* @google/perfetto-team
+
+# Files/Directories that don't require a strict owner check.
+/Android.bp
+/Android.bp.extras
+/BUILD
+/BUILD.*
+/docs/**/*
+/examples/**/*
+/protos/perfetto/trace/**/*
+/protos/perfetto/config/**/*
+/protos/perfetto/metrics/**/*
+/ui/src/plugins/**/*
+/ui/src/test/**/*
+/tools/**/*
+/test/**/*
--- a/test/OWNERS.github
+++ b/test/OWNERS.github
@@ -1,0 +1,4 @@
+# Chromium-specific test data files are rolled from the chromium repo.
+# The chromium owners should be able to approve the rolls.
+
+data/chrome/**.sha256 @google/chromium-stdlib-owners


contrib/rust-sdk/OWNERS.github
```diff
@@ -1,8 +1,8 @@
...
 # that enforces stricter validation on OWNERS files, such as requiring internal
-# email domains. 
+# email domains.
+#
+# Syntax is the same as .github/CODEOWNERS but paths are relative to this directory.
 
-dreveman@gmail.com
-jon.bailey45@gmail.com
-ziyi.elisa.tai@gmail.com
+/ @dreveman @jon-rv @zytyz
```
primiano: could this be `./`  or `.` to be a bit less ambiguous?

OWNERS.github
```diff
@@ -0,0 +1,20 @@
+# This file define code owners for GitHub pull requests
```
primiano, line 1: do we need this file? imho there is just risk of confusing people thinking this is the actual codeowners.

Also I was expecting that our root codeowners becomes /CODEOWNES.github and have the whole .github/CODEOWNERS autogenerated?

src/trace_processor/perfetto_sql/stdlib/chrome/OWNERS.github
```diff
@@ -0,0 +1,4 @@
+# Chromium-specific parts of stdlib is rolled from the chromium repo.
+# The chromium owners should be able to approve the rolls.
+
+**/* @google/chromium-stdlib-owners
```
primiano: why the rust is / and this is **/*

we should pick one or people will be confused

tools/gen_github_codeowners
```diff
@@ -0,0 +1,93 @@
...
+          continue
+
+        parts = line.split()
+        if len(parts) < 2:
+          continue
+
+        pattern = parts[0]
+        owners = ' '.join(parts[1:])
+
+        # Convert relative pattern to absolute path
```
primiano, line 83: can you plz hard crash if somebody tries to put '../' ?
that could be an OWNERS attack model if somebody first gains ownership in contrib/bla, and then tries to escape and get OWNERS on something else outside

PR #3726, opened 2025-11-19
tracebox: Explicit daemon management

sashwinbalaji:
- Introduce `tracebox ctl` for explicit daemon lifecycle management
- Default behavior now required daemons to be running 
- `--autodaemonize` flag preserves old auto-spawn behavior
- `--system-sockets` deprecated with a warning printed on usage

Fixes: b/458318005
Fixes: https://github.com/google/perfetto/discussions/3540


docs/reference/tracebox.md
```diff
@@ -4,97 +4,392 @@
...
+- Use system-wide Perfetto daemons for SDK tracing on Android
+
+### Ftrace issues
+
+```bash
+# Reset ftrace state manually:
+sudo traced_probes --reset-ftrace
+```
+
+## FILES
```
LalitMaganti: IMO this is too much detail for a man page.

docs/reference/tracebox.md
```diff
@@ -4,97 +4,392 @@
...
+
+### Linux/macOS
+
+**Default:** Requires daemons to be running.
+- Use `tracebox ctl start` to start persistent daemons
+- Or use `--autodaemonize` for self-contained mode
+
+### Windows
+
+**Default:** Always spawns daemons with system sockets.
```
LalitMaganti: Why are we doing different things on Windows? We should just be consistent no? At least to the extent of saying on Windows --autodaemonize must be specified.

It sucks for everyone if you have inconsistent behaviour like this.

docs/reference/tracebox.md
```diff
@@ -4,97 +4,392 @@
 
 tracebox - all-in-one binary for Perfetto tracing services
```
LalitMaganti, line 5: I think there needs to be a pass on this file to prune unnecessary infiormation Right now there is too much implementation detail here.

Put yourself in the shoes of a random Linux developer: do you really need everything you've written here? Things like socket paths?

include/perfetto/base/build_configs/bazel/perfetto_build_flags.h
```diff
@@ -54,6 +54,7 @@
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ENABLE_WINSCOPE() (1)
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ENABLE_RT_MUTEX() (1)
 #define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ENABLE_LOCKFREE_TASKRUNNER() (1)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_SYSTEMD() (1)
```
LalitMaganti: This shouldn't be 1?

gn/perfetto.gni
```diff
@@ -353,6 +353,10 @@ declare_args() {
       (perfetto_build_standalone || perfetto_build_with_android) &&
       !is_perfetto_build_generator
 
+  # Enables installation of systemd unit files.
+  # Only for standalone Linux builds where users build Perfetto themselves.
+  enable_perfetto_systemd = perfetto_build_standalone && is_linux
```
LalitMaganti: I think you need a !build_generator here.

src/tracebox/tracebox.cc
```diff
@@ -54,103 +53,91 @@ const Applet g_applets[]{
...
-  Applets:)");
+#if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  printf(R"(
+  tracebox ctl start                      # Start daemons
+  tracebox -t 10s -o trace.pftrace sched  # Capture trace
+  tracebox ctl stop                       # Stop daemons (optional)
+)");
+  PrintTraceboxCtlUsage();
+  printf(R"(
+ADVANCED: --autodaemonize (NOT RECOMMENDED)
```
LalitMaganti: I don't want to call this "advanced" as I think it's legit people want this if they just are looking at ftrace. I would present them as two equal options but with preference to daemonization.

src/tracebox/tracebox.cc
```diff
@@ -227,27 +197,105 @@ int TraceboxMain(int argc, char** argv) {
...
+  }
+
+  // Handle --autodaemonize and --system-sockets flags by removing them from
+  // argv.
+  bool autodaemonize = false;
+
+  for (int i = 1; i < argc;) {
+    if (strcmp(argv[i], "--autodaemonize") == 0) {
+      autodaemonize = true;
+      // Remove from argv.
```
LalitMaganti: This is just a memmove no?

src/tracebox/tracebox.cc
```diff
@@ -227,27 +197,105 @@ int TraceboxMain(int argc, char** argv) {
...
+        argv[j] = argv[j + 1];
+      }
+      argc--;
+      // Don't increment i, check the same position again.
+    } else if (strcmp(argv[i], "--system-sockets") == 0) {
+      PERFETTO_ELOG(
+          "Warning: --system-sockets is deprecated. System sockets are now the "
+          "default.");
+      // Remove from argv.
+      for (int j = i; j < argc - 1; ++j) {
```
LalitMaganti: As above

src/tracebox/tracebox.cc
```diff
@@ -227,27 +197,105 @@ int TraceboxMain(int argc, char** argv) {
...
+    } else {
+      ++i;
+    }
+  }
+
+  if (autodaemonize)
+    return RunWithAutodaemonize(argc, argv);
+
+#if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  // Require daemons to be running (started via 'tracebox ctl start').
```
LalitMaganti: It's imperitive here we explain the changes made in this CL. Otherwise as a user I'm going to be *very very very* confused. Please make sure to document this thoroughly: we shouldn't take breaking changes lightly.

src/tracebox/tracebox.cc
```diff
@@ -227,27 +197,105 @@ int TraceboxMain(int argc, char** argv) {
...
+  PerfettoCmd perfetto_cmd;
+  auto opt_res = perfetto_cmd.ParseCmdlineAndMaybeDaemonize(argc, argv);
+  if (opt_res.has_value()) {
+    if (*opt_res != 0)
+      PrintUsage();
+    return *opt_res;
+  }
+
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  // Windows: Spawn daemons with system sockets (no daemon-running check).
```
LalitMaganti: I don't like this. I would suggest we align Windows with other platforms.

CHANGELOG
```diff
@@ -9,6 +9,11 @@ Unreleased:
     *
   SDK:
     *
+  Tools:
```
LalitMaganti, line 22: We need to document the breaking change here and call it out.

sashwinbalaji:
Thanks for the review:
1) for communication plan was to send out mail(+announcement) will look into https://perfetto.dev/docs/analysis/perfetto-sql-backcompat
2) regarding doc it's in pipeline as a separate cl as there are multiple files using it: https://source.chromium.org/search?q=tracebox%20&sq=&ss=chromium%2Fchromium%2Fsrc:third_party%2Fperfetto%2Fdocs%2F
3) Switching to draft until windows mode is aligned, tracebiox.md rereviewed and try out --background option instead of reinventing the wheel

src/tracebox/tracebox.cc
sashwinbalaji: so for windows:
1) autodaemonize was not supported before and also now it fails with same error: The autostart mode is not supported on this platform as now (will ensure doc matches this)
2) the problem is for system sockets which is the new default behavior (same as giving system sockets), I didn't add getservicesockets check and simply followed as before, will update to check for "127.0.0.1:32278" and "127.0.0.1:32279" aligning with [default_socket](https://crsrc.org/c/third_party/perfetto/src/tracing/ipc/default_socket.cc;l=76)

```
  auto* end = std::remove_if(argv, argv + argc, [](char* arg) {
    return !strcmp(arg, "--system-sockets");
  });
  if (bool system_sockets = end == (argv + argc - 1); system_sockets) {
    argc--;
  } else {
    auto pid_str = std::to_string(static_cast<uint64_t>(base::GetProcessId()));
    PERFETTO_FATAL("The autostart mode is not supported on this platform");
  }

  PerfettoCmd perfetto_cmd;
  auto opt_res = perfetto_cmd.ParseCmdlineAndMaybeDaemonize(argc, argv);
  if (opt_res.has_value()) {
    if (*opt_res != 0) {
      PrintTraceboxUsage();
    }
    return *opt_res;
  }

  std::string self_path = base::GetCurExecutablePath();
  base::Subprocess traced({self_path, "traced"});
  traced.Start();

  base::Subprocess traced_probes(
      {self_path, "traced_probes", "--reset-ftrace"});
  traced_probes.args.posix_proc_group_id = traced.pid();
  traced_probes.Start();

  PERFETTO_ELOG(
      "Unsupported: linux.perf data source support (traced_perf) "
      "compiled-out.");

  perfetto_cmd.ConnectToServiceRunAndMaybeNotify();
  return 0;
}
```

LalitMaganti:
Overall lgtm but I don't quite understand the systemd changes. Can we maybe pull that out into a separate cl? The rest looks really good!

src/tracebox/tracebox.cc
```diff
@@ -54,70 +53,50 @@ const Applet g_applets[]{
...
+
+MODE 1: Daemon mode (Recommended)
+  Background daemons are started once and shared across multiple tracing sessions.
+  This supports SDKs (track_event), reduces latency and is generally more robust.
+
+  > tracebox ctl start
+  > tracebox -t 10s -o trace.pftrace sched
+  > tracebox ctl stop
+
+MODE 2: Autodaemonize mode
```
primiano, line 73: I'd call this

MODE 2: --autodaemonize mode (legacy behavior, discouraged)

src/tracebox/tracebox.cc
```diff
@@ -250,6 +229,85 @@ int TraceboxMain(int argc, char** argv) {
...
+    return 1;
+  }
+
+  bool autodaemonize = false;
+  bool use_system_sockets = false;
+  for (int i = 1; i < argc;) {
+    if (strcmp(argv[i], "--autodaemonize") == 0) {
+      autodaemonize = true;
+      memmove(static_cast<void*>(&argv[i]), static_cast<void*>(&argv[i + 1]),
+              sizeof(char*) * static_cast<size_t>(argc - i - 1));
```
primiano, line 255: I think you can do just:

std::move(argv + i + 1, argv + argc, argv + i);

src/tracebox/tracebox.cc
```diff
@@ -250,6 +229,85 @@ int TraceboxMain(int argc, char** argv) {
...
+        use_system_sockets = true;
+      }
+      ++i;
+    }
+  }
+
+  if (autodaemonize) {
+    if (use_system_sockets) {
+      PERFETTO_ELOG(
+          "Warning: --system-sockets with --autodaemonize is supported but "
```
primiano, line 268: i'd say "supported for legacy reasons"

src/tracebox/tracebox.cc
```diff
@@ -250,6 +229,85 @@ int TraceboxMain(int argc, char** argv) {
...
+          "not recommended. Prefer `tracebox ctl start` for persistent system "
+          "sockets.");
+    }
+    return RunAutodaemonize(argc, argv);
+  }
+
+  if (use_system_sockets) {
+    PERFETTO_FATAL(
```
primiano, line 276: Fatal seems a bit aggressive given that this is just the new default?
I would

1) Move this message in the for loop above. there i sno need to set a variable, just to emit a lot here:
2) SOften this into a PERFETTO_ELOG ?

src/tracebox/tracebox.cc
```diff
@@ -54,70 +53,50 @@ const Applet g_applets[]{
...
+void PrintTraceboxUsage() {
   printf(R"(Welcome to Perfetto tracing!
 
 Tracebox is a bundle containing all the tracing services and the perfetto
-cmdline client in one binary. It can be used either to spawn manually the
-various subprocess or in "autostart" mode, which will take care of starting
-and tearing down the services for you.
+cmdline client in one binary. It can be used in two modes:
+
+MODE 1: Daemon mode (Recommended)
```
primiano, line 65: Add a (Default behavior, reccomended)

src/tracebox/tracebox_ctl.h
```diff
@@ -0,0 +1,45 @@
...
+
+#ifndef SRC_TRACEBOX_TRACEBOX_CTL_H_
+#define SRC_TRACEBOX_TRACEBOX_CTL_H_
+
+#include <string>
+
+namespace perfetto {
+
+// Holds the socket paths for the tracing service IPC endpoints.
+struct ServiceSockets {
```
primiano, line 25: given you are in the "perfetto" namespace I'd call this CtlServiceSockets, to make it less confusing (When autocompletion kicks in)

src/tracebox/tracebox_ctl.h
```diff
@@ -0,0 +1,45 @@
...
+    return !producer_socket.empty() && !consumer_socket.empty();
+  }
+  std::string ToString() const {
+    return "Producer Socket: " + producer_socket +
+           ", Consumer Socket: " + consumer_socket;
+  }
+};
+
+void PrintTraceboxCtlUsage();
+ServiceSockets GetRunningSockets();
```
primiano, line 38: same here about global namespace.

This stuff should be either wrapped in:
A) `namespace tracebox_ctl {`
B) be static methods of  `class TraceboxCtl {`

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+#if PERFETTO_BUILDFLAG(PERFETTO_TRACED_PERF)
+    "traced_perf",
+#endif
+};
+
+bool CanConnectToSocket(const std::string& path) {
+  base::SockFamily family = base::GetSockFamily(path.c_str());
+  auto sock =
+      base::UnixSocketRaw::CreateMayFail(family, base::SockType::kStream);
+  return sock && sock.Connect(path);
```
primiano, line 62: is this enough on linux and mac without having to look for the OnConnect?
 I'm not surprised if it's the case, just checking (because in reality unix socket connections is syncrhonous even on async socket because the kernel knows immediately, there is no "networking")

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+}
+
+#if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+
+std::string GetPidFilePath(const char* daemon_name) {
+  return base::GetSysTempDir() + "/" + daemon_name + ".pid";
+}
+
+bool WritePidToFile(const std::string& path, pid_t pid) {
+  base::ScopedFile fd =
```
primiano, line 72: Can you please write into a path + ".tmp" file, and then do a rename() to remove the suffix.

THis can cause very subtle bugs in bash scripts of people,  because if they are unlucky they might see an empty pid file (if they manage to get between the open(O_TRUNC) and the write)

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+  return written >= 0 && static_cast<size_t>(written) == content.size();
+}
+
+pid_t ReadPidFromFile(const std::string& path) {
+  std::string pid_str;
+  if (!base::ReadFile(path, &pid_str))
+    return 0;
+  pid_str = base::TrimWhitespace(pid_str);
+  if (pid_str.empty())
+    return 0;
```
primiano, line 87: I think this case (empty) is already covered below by StringTOInt64 (check, but I think it fails if it's empty)

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+pid_t StartDaemon(const std::string& tracebox_path,
+                  const std::string& daemon_name) {
+  std::vector<std::string> cmd = {tracebox_path, daemon_name, "--background"};
+
+  base::Subprocess daemon;
+  daemon.args.exec_cmd = cmd;
+  daemon.args.stdout_mode = base::Subprocess::OutputMode::kBuffer;
+  daemon.Start();
+
+  constexpr uint32_t kDaemonStartTimeoutMs = 1000;
```
primiano, line 101: make it 5000. some emulators are sloooooooww

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+  if (!daemon.Wait(kDaemonStartTimeoutMs)) {
```
primiano, line 102: Add a comment says that "--background causes the tracebox applet to daemonize and print the PID in stdout" 
so the code below becomes less mysterious for people who stumble on it without knowing how stuff works

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+    printf("Error: Daemon %s failed to start after %u ms\n",
```
primiano, line 103: `fprintf(stderr,` ...

Same below

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+    return 0;
+  }
+
+  std::string output = daemon.output();
+  auto pid = base::StringToInt64(base::TrimWhitespace(output));
+  if (!pid.has_value()) {
+    printf("Error: Failed to parse daemon PID from output: %s\n",
+           output.c_str());
+    return 0;
+  }
```
primiano, line 120: This is so rare (should really never happen) that doesn't warrant so many lines.
I'd just go PERFETTO_CHECK(pid.has_value())

We own the code that prints the value. if it doesn't, we have a bug, in a way or another.

The otehr cases instead can legitimately fail outside of our control, so the (f)printf is approrpriate

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+}
+
+#endif  // !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+
+int CtlStop() {
+  printf("Stopping daemons...\n");
+
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  ServiceSockets sockets = GetRunningSockets();
+  if (!sockets.IsValid()) {
```
primiano, line 131: Do you actually need to see if the socket is valid?
In my mind you should just open the pid file, and kill that (And wait that the daemon deletes the pid file)

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+    return 0;
+  }
+  printf("Daemons are running. Use Task Manager to stop them.\n");
+  return 1;
+#else
+  int daemons_stopped = 0, daemons_stored = 0;
+  for (const char* daemon : kDaemons) {
+    std::string pid_path = GetPidFilePath(daemon);
+    if (!base::FileExists(pid_path)) {
+      PERFETTO_LOG("Not found pid file for %s at %s", daemon, pid_path.c_str());
```
primiano, line 142: This might add noise if some daemon quits fo rlegitimate reasons.
I'd just print out the ones that you kill.
the fact that nothing is found will be evident when you print "Stopped 0 processes at the end"

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+      continue;
+    }
+    daemons_stored++;
+    pid_t pid = ReadPidFromFile(pid_path);
+    if (pid && kill(pid, SIGTERM) == 0) {
+      daemons_stopped++;
+      printf("%s terminated (PID: %d)\n", daemon, pid);
+      if (remove(pid_path.c_str()) != 0) {
```
primiano, line 150: shouldn't the remove be done by the process?
Otherwise you can you tell when a process has *actually* ended?
There is an expectation that "ctl stop" waits for the termination, so you can call start > stop > start and it should work.

I think the way this could work is: the daemon removes the pid when it quits, and here you poll for the deletion of the pid

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+  return daemons_stored == 0;
+#endif
+}
+
+int CtlStart() {
+  constexpr uint32_t kDaemonStartWaitUs = 1000 * 1000;  // 1s
+  ServiceSockets sockets = GetRunningSockets();
+  if (sockets.IsValid()) {
+    printf("Status: Daemons are already running with %s\n",
+           sockets.ToString().c_str());
```
primiano, line 173: I'd add something more helpful like ". Maybe the daemons have been started via systemd/android init or other means?"

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+  }
+  return daemons_stored == 0;
+#endif
+}
+
+int CtlStart() {
+  constexpr uint32_t kDaemonStartWaitUs = 1000 * 1000;  // 1s
+  ServiceSockets sockets = GetRunningSockets();
+  if (sockets.IsValid()) {
+    printf("Status: Daemons are already running with %s\n",
```
primiano, line 172: s/with/on/

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+    std::string pid_path = GetPidFilePath(daemon);
+    if (!WritePidToFile(pid_path, pid)) {
+      printf("Error: Failed to write PID file for %s\n", daemon);
+      CtlStop();
+      return 1;
+    }
+    printf("%s started (PID: %d)\n", daemon, pid);
+  }
+
+  base::SleepMicroseconds(kDaemonStartWaitUs);
```
primiano, line 220: imho this needs to be a poll of some sort. 1 s is too low on an emulator, and if you go higher, it's annoying to wait that long.
Poll every 250ms for at most 5s?

src/tracebox/tracebox_ctl.cc
```diff
@@ -0,0 +1,306 @@
...
+           started_sockets.ToString().c_str());
+    CtlStop();
+    return 1;
+  }
+  return 0;
+#endif
+}
+
+int CtlStatus() {
+  ServiceSockets sockets = GetRunningSockets();
```
primiano, line 233: I'd be a bit more pedantic here. I'd check both for (1) GetRunningSockets(); (2) the existance of pid files (at least just for traced)

If 1 && 2: you are happy
if 1 && not(2) you say "The daemon seems running but the pid file doesn't exist. looks like something else started the daemon (systemd? initrc?). Use lsof ... (figure out the command) for details"

Issue #3725, opened 2025-11-19
Change release process to make C/C++ SDK a release artifact


LalitMaganti:
Instead of tools/gen_amalagamated happening on the release branch, we should make it instead happen as part of generating Github zips. Specifically, I think we should have a perfetto-c-sdk-src.zip and perfetto-cpp-sdk.zip as artifacts. We should then remove the notion of "release branches" altogether as we will instead have a long lived "canary" and "stable" branches with release tags cut from the stable branch. Exact details of this later bit will follow either as an RFC or potentially as a sequence of changes to our docs and script.

This will require changing our release process docs and release script to match the new state of the world

PR #3727, opened 2025-11-19
github: make SDK generation part of Github release process

LalitMaganti:
This CL makes the amalgamated C/C++ SDK part of the Github zip
generation process rather than happening on the release branch. This is
a necessary step to changing how Perfetto releases work where we have
long lived canary/stable branches.

Fixes: https://github.com/google/perfetto/issues/3725

### DIFF
--- a/docs/contributing/sdk-releasing.md
+++ b/docs/contributing/sdk-releasing.md
@@ -101,7 +101,8 @@ git push origin vX.Y
    - [docs/instrumentation/tracing-sdk.md](/docs/instrumentation/tracing-sdk.md)
    - [examples/sdk/README.md](/examples/sdk/README.md)
 
-6. Send an email with the CHANGELOG to perfetto-dev@ (internal) and to the
+6. Send an email with the CHANGELOG to perfetto-dev@ with perfetto-announce@
+   bcc'ed (internal - make sure you have permissions before doing so) and to the
    [public perfetto-dev](https://groups.google.com/forum/#!forum/perfetto-dev).
 
 ## Creating a GitHub release with prebuilts and SDK sources
--- a/examples/sdk/README.md
+++ b/examples/sdk/README.md
@@ -13,17 +13,12 @@ Dependencies:
 ## Building
 
 First, download the SDK source files from the latest Perfetto release:
+1) Visit https://github.com/google/perfetto/releases/latest
+2) Download perfetto-cpp-sdk-src.zip and extract the files to sdk/perfetto
+   directory
 
-```bash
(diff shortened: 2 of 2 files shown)


docs/contributing/sdk-releasing.md
```diff
@@ -118,36 +96,53 @@ git tag -a -m "Perfetto vX.Y" vX.Y
...
-  - "Attach binaries" -> Attach the ten .zip files from the previous step.
+  - "Attach binaries" -> Attach all twelve .zip files from the previous step
+    (10 prebuilt binaries + 2 SDK source zips).
 
-10. Run `tools/roll-prebuilts vX.Y`. It will update the SHA256 into the various
+7. Run `tools/roll-prebuilts vX.Y`. It will update the SHA256 into the various
    scripts under `tools/`. Upload a CL with the changes.
 
-11. Phew, you're done!
+8. Send an email with the CHANGELOG to perfetto-dev@ (internal) and to the
```
sashwinbalaji, line 146: add perfetto-announce as cc (also make sure you have write permissions before doing so)

tools/release/package-github-release-artifacts
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env python3
+# Copyright (C) 2021 The Android Open Source Project
```
sashwinbalaji, line 2: 2025

examples/sdk/README.md
```diff
@@ -12,24 +12,30 @@ Dependencies:
...
 
-First, check out the latest Perfetto release:
+First, download the SDK source files from the latest Perfetto release:
 
 ```bash
+# Download the C++ SDK source zip from GitHub releases
+# Visit https://github.com/google/perfetto/releases/latest
+# Download perfetto-cpp-sdk-src.zip and extract it to sdk/ directory
+
+# Alternatively, if building from a release tag:
```
primiano: IMHO this is a bit misleading because when I see code, i assume that's the primary way. i didn't put too much attention to the "alternatively".

I'd just say here "Alternatively use tools/gen_amalgamated to generate it from a release tag"

tools/release/package-github-release-artifacts
primiano: > 2025

Technically 2021 is correct as he is renaming a file, and it should retain the original copyright year.

sashwinbalaji:
> > 2025
> 
> Technically 2021 is correct as he is renaming a file, and it should retain the original copyright year.

agreed, but since git is considering it as a new file rather than moved so suggested that. maybe better would be to change the semantic from added to moved. but it's a small nit.

PR #3731, opened 2025-11-19
ui: Resolve JavaScript stacktraces in the browser

stevegolton:
Decode JavaScript stacktraces in the browser to TypeScript sources in the crash dialog and when reporting crash reports to buganizer.

Before:
```
 - Object.callback (frontend_bundle.js:95363:24)
 - CommandManagerImpl.runCommand (frontend_bundle.js:216751:27)
 - onSubmit (frontend_bundle.js:338941:27)
 - HTMLInputElement.onkeydown (frontend_bundle.js:339162:34)
 - EventDict.handleEvent (frontend_bundle.js:10769:56)
```

After:
```
 - Object.callback (src/plugins/dev.perfetto.Chaos/index.ts:28:14)
 - CommandManagerImpl.runCommand (src/core/command_manager.ts:98:20)
 - onSubmit (src/frontend/omnibox.ts:179:17)
 - HTMLInputElement.onkeydown (src/frontend/omnibox.ts:523:20)
 - EventDict.handleEvent@node_modules/.pnpm/mithril (2.2.2/node_modules/mithril/render/render.js:890:54)
```

This consists of two parts:

1. Build system

A new rollup plugin injects a snippet of code at the end of each bundle which injects a stripped down source map for that bundle into a registry called `__SOURCEMAPS` in the global namespace.

```js
self.__SOURCEMAPS = (self.__SOURCEMAPS || {})['<bundle-file-name>']=<source-map-content>
```

Note: Bundles loaded into workers (e.g. engine and traceconv) have a different `self` object to the frontend bundle, so this registry is not for them. The reason this is a registry rather than a single variable is if we ever need multiple bundles in the main UI thread in the future I want to avoid this biting us.

A new dev dependency has been added on `source-map` which is used to strip down the source map before embedded it.

2. UI

The UI can now use this sourcemap registry to look up source mappings for each line in a stack trace. This translated stack is then displayed in the crash dialog and used for analytics.

**How to Test:**
1. **Trigger a crash:** Run the command 'Add debug slice track' and enter an invalid query.
2. **Check the crash dialog**: It should show typescript filenames and line numbers.

### DIFF
--- a/ui/src/base/source_map_utils.ts
+++ b/ui/src/base/source_map_utils.ts
@@ -56,14 +56,14 @@ const VLQ_CONTINUATION_BIT = VLQ_BASE; // 32
 const sourceMapCache = new Map<string, ProcessedSourceMap>();
 
 // Extend the global interface to include our custom property
-interface PerfettoGlobal {
+interface WithSourcemaps {
   __SOURCEMAPS?: Record<string, SourceMap>;
 }
 
 // Get embedded source map for a specific bundle file (synchronous)
 function getEmbeddedSourceMap(bundleFileName: string): SourceMap | null {
   // Use 'self' for both window and worker compatibility
-  const global = self as unknown as PerfettoGlobal;
+  const global = self as unknown as WithSourcemaps;
   const registry = global.__SOURCEMAPS;
   if (!registry) return null;
 


ui/config/rollup.config.js
```diff
@@ -19,10 +19,88 @@ const path = require('path');
...
+          
+          // Remove sourcesContent to reduce size
+          delete minimalMap.sourcesContent;
+          // Remove names array (should be empty anyway since we didn't add names)
+          delete minimalMap.names;
+
+          // Embed the minimal map at the end of the bundle using a registry
+          // Use 'self' instead of 'window' for worker compatibility
+          // Each bundle registers its map with its filename as the key
+          chunk.code += `\n;(self.__SOURCEMAPS=self.__SOURCEMAPS||{})['${chunk.fileName}']=${JSON.stringify(minimalMap)};`;
```
primiano, line 90: so one thing: we need to be able to symbolize all the bundles from the main thread, as, IIRC; the way crash reporting works is that crashes in the engine get forwarded over postMessage to the frontend, and the frontend does evertything

ui/src/base/logging.ts
```diff
@@ -119,6 +120,10 @@ export function reportError(err: ErrorEvent | PromiseRejectionEvent | {}) {
     errMsg = `${err}`;
   }
 
+  if ('stack' in errorObj && typeof errorObj.stack === 'string') {
```
primiano: two things here:

1) as discussed offline we might want to do this only in the error dialog
2) if you want to do here, maybe better to do it below after we destructure the stacktrace into name+locatio?

ui/src/base/source_map_utils.ts
```diff
@@ -0,0 +1,344 @@
...
+const VLQ_BASE = 1 << VLQ_BASE_SHIFT; // 32
+const VLQ_BASE_MASK = VLQ_BASE - 1; // 31
+const VLQ_CONTINUATION_BIT = VLQ_BASE; // 32
+
+const sourceMapCache = new Map<string, ProcessedSourceMap>();
+
+// Extend the global interface to include our custom property
+interface PerfettoGlobal {
+  __SOURCEMAPS?: Record<string, SourceMap>;
+}
```
primiano, line 61: I think you can just do

declare global {
  var __SOURCEMAPS ?: Record<string, SourceMap>;
}

and this avoid the need of "self as unknown as Perfettoglobal"... as you can just do self.__SOURCEMAPS

stevegolton: The problem is these declares are leaky - all typescript files will be able to use __SOURCEMAPS, but we have no reason to use it in any file other than this one. I don't really want other folks (including plugins) to start depending on it.

ui/src/base/logging.ts
stevegolton: I'll move it to the error dialog path.

ui/config/rollup.config.js
stevegolton: I was hoping we could kick this can down the road, especially as 99.9% of crashes are going to come from frontend_bundle.js.

stevegolton: Discussed offline, going to punt this for now.

primiano: sorry I had a brainfart. you are right, engine_bundle.js is boring TS wise, 99% of the interesting things there are in wasm which is already symbolized

primiano: I'm going to answer my own question:

> Out of curiosity, if we were to build a dumb dict of col:line>line how bit it would be after compression?

Too big, like 40 MB before compression

stevegolton:
> LG % comment on moving this. TBH i am a bit worried about all the home-brewed logic to decode the source map. Out of curiosity, if we were to build a dumb dict of col:line>line how bit it would be after compression?

Discussed offline. The source map encoding is actually very compact. That does require a more complex parser/decoder, but it's worth it for the reduced bundle size.

PR #3734, opened 2025-11-19
tp: Add stack tables to stdlib

aMayzner:
## Summary

Add SQL-based stack table implementation to stdlib that computes `stack_id` and `parent_stack_id` o demand. This prepares for removing these columns from the C++ slice table.

  ## Changes

  - **`slices/stack.sql`**: New module with `_slice_with_stack_id` view and stack-based query functions
    - `_slice_with_stack_id`: Computes stack IDs by hashing concatenated ancestor slice names
    - `_ancestor_slice_by_stack()`: Returns all slices with a given stack_id plus their ancestors
    - `_descendant_slice_by_stack()`: Returns all slices with a given stack_id plus their descendants

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/slices/stack.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/slices/stack.sql
@@ -1,0 +1,210 @@
+--
+-- Copyright 2025 The Android Open Source Project
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     https://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE slices.hierarchy;
+
+-- View that provides stack_id and parent_stack_id for all slices by computing
+-- them on-demand.
+--
+-- Note: This is currently for internal use only and computes stack hashes
+-- on-demand, which may be slower than the C++ implementation.
+CREATE PERFETTO VIEW _slice_with_stack_id (
+  -- Slice id.
+  id ID(slice.id),
+  -- Alias of `slice.ts`.
+  ts TIMESTAMP,
+  -- Alias of `slice.dur`.
+  dur DURATION,
+  -- Alias of `slice.track_id`.
+  track_id JOINID(track.id),
+  -- Alias of `slice.category`.
+  category STRING,
+  -- Alias of `slice.name`.
+  name STRING,
+  -- Alias of `slice.depth`.
+  depth LONG,
+  -- Alias of `slice.parent_id`.
+  parent_id JOINID(slice.id),
+  -- Alias of `slice.arg_set_id`.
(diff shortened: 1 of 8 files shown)


src/trace_processor/perfetto_sql/stdlib/slices/stack.sql
```diff
@@ -0,0 +1,198 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE slices.hierarchy;
+
+-- View that provides stack_id and parent_stack_id for all slices by computing
+-- them on-demand.
+--
+-- Note: This is currently for internal use only and computes stack hashes
```
LalitMaganti, line 21: Why internal use only? Isn't this what people will use for migration purposes?

src/trace_processor/perfetto_sql/stdlib/slices/stack.sql
```diff
@@ -0,0 +1,198 @@
...
+  -- Alias of `slice.thread_dur`.
+  thread_dur LONG,
+  -- Alias of `slice.thread_instruction_count`.
+  thread_instruction_count LONG,
+  -- Alias of `slice.thread_instruction_delta`.
+  thread_instruction_delta LONG,
+  -- A unique identifier obtained from the names and categories of all slices
+  -- in this stack. Computed on-demand.
+  stack_id LONG,
+  -- The stack_id for the parent of this slice. 0 if there is no parent.
```
LalitMaganti: Why zero? Why not null? Didn't it used to be!

src/trace_processor/perfetto_sql/stdlib/slices/stack.sql
aMayzner: It will be exposed in the next commit in which we would actually use it and remove the stack_id. In this commit only there is no point for anyone to use it.

src/trace_processor/perfetto_sql/stdlib/slices/stack.sql
```diff
@@ -53,22 +53,28 @@ CREATE PERFETTO VIEW _slice_with_stack_id (
...
-slice_stack_hashes AS (
-  SELECT
-    id,
-    hash(stack_str) AS stack_hash
-  FROM slice_stack_strings
-)
+WITH
+  slice_stack_strings AS (
+    SELECT
+      s.id,
```
LalitMaganti, line 59: Wouldn't it be more efficient to has names and then hash the result instead of doing this?

test/trace_processor/diff_tests/parser/simpleperf/clocks_align_test.sql
```diff
@@ -37,7 +39,7 @@ WITH
         ) AS stack
     FROM span
   )
-SELECT ts, stack FROM tmp WHERE parent_stack_id = 0 ORDER BY TS ASC;
+SELECT ts, stack FROM tmp WHERE parent_stack_id IS NULL ORDER BY TS ASC;
```
LalitMaganti: Wait sorry so in the past parent_stack_id *was* 0 for root slices?

aMayzner: It was 0, but shouldn't it be NULL now, when there is no parent?

LalitMaganti: We should be consistent with what we had before. We don't want two breaking changes to forc eon users at same time.

Issue #3550, opened 2025-11-05
Pressing "f" does not zoom in far enough


saagarjha:
On my trace (attached) if I press "f" to ~~pay respects~~ zoom in it doesn't actually zoom in very far. This means I have to manually pinch and scroll around which is kind of a pain.

[trace.json.zip](https://github.com/user-attachments/files/23374800/trace.json.zip)

https://github.com/user-attachments/assets/2eb2fb86-566b-4312-9746-2571900a52ff

LalitMaganti:
Thanks for filing! Ahhh so I misunderstood this feature when we were discussing: this is about area selection not single selection. Makes sense.

LalitMaganti:
Fixed by https://github.com/google/perfetto/pull/3761. 'F' now consistently zooms and pans no matter where you are in the trace!

PR #3743, opened 2025-11-19
Add a command to zoom and fill the viewport 

ivan-chong:
This is similar to focus, except zooms the selection to fill the viewport. 
- Created a command with the hotkey (Shift+F) for zoom and fill
- Refactored panToTimestamp into centerOnTimestamp to facilitate centering on timestamps which are in the current timeline view

Bug: #3550


LalitMaganti:
Hmmmmm so I don't think we want this feature actually. We're actually changing how focus works in an upcoming change in the fixit this week.

BTW in the future, please ask us whether a given feature makes sense or take one of the bugs which we've already vetted and opened in the github issues!

ivan-chong:
Ah ok got it, thanks for the heads up - closing the PR now.

PR #3749, opened 2025-11-19
tools: gen_c_protos to take arg for extra protos

sashwinbalaji:
We add it as extra protos instead of overriding the existing ones to ensure default ones are always generated and updated. 

Bug: b/462161293

### DIFF
--- a/tools/gen_c_protos
+++ b/tools/gen_c_protos
@@ -22,10 +22,7 @@ import subprocess
 import sys
 import tempfile
 
-CATEGORY_PROD = 'prod'
-CATEGORY_TEST = 'test'
-
-PROD_DEFAULTS = {
+PERFETTO_DEFAULTS = {
     'guard_strip_prefix': 'PROTOS_PERFETTO_',
     'guard_add_prefix': 'INCLUDE_PERFETTO_PUBLIC_PROTOS_',
     'path_strip_prefix': 'protos/perfetto',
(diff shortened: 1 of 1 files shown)


tools/gen_c_protos
```diff
@@ -149,17 +152,49 @@ def include_path_for(filename):
...
+      epilog="""Example workflow:
+  # Generate custom protos
+  tools/gen_c_protos out/linux_clang_release \\
+      --proto protos/perfetto/trace/android/your_custom.proto \\
+      --proto protos/perfetto/common/your_common.proto
+""")
+
+  parser.add_argument(
+      'OUT', help='Build output directory (e.g., out/linux_clang_release)')
+  parser.add_argument(
```
LalitMaganti, line 174: I would suggest a different approach: *ether* it works for inline protos *or* it works for standalone. I think it's never the case you want both.

sashwinbalaji: sure, I was split between either. Though kept the test generation as they silently happen in background just to ensure if something breaks we get to know.

tools/gen_c_protos
```diff
@@ -149,17 +152,49 @@ def include_path_for(filename):
...
+      '--check-only',
+      action='store_true',
+      help='Verify generated files match existing without writing')
+
   args = parser.parse_args()
+
   out = args.OUT
 
+  if args.proto:
+    SOURCE_FILES[CATEGORY_PROD]['files'] = args.proto
```
LalitMaganti: Honestly I'm really not a huge fan of rerplacing the data here. Can we instead refactor the code to have it passed in and making the decision of which source files we want to process instead of changing global varaibles?

sashwinbalaji: sure, what's your thought on running test proto irrespective of flag ?

LalitMaganti: I don't think we should do anything for situations where people run with flag.

tools/gen_c_protos
```diff
@@ -149,17 +153,65 @@ def include_path_for(filename):
...
+      action='store_true',
+      help='Verify generated files match existing without writing')
+
   args = parser.parse_args()
+
   out = args.OUT
 
+  if args.proto:
+    source_files = {
+        CATEGORY_PROD: {
```
LalitMaganti: Shouldn't this be category_custom?

tools/gen_c_protos
```diff
@@ -149,17 +153,65 @@ def include_path_for(filename):
...
+            'files': args.proto,
+            **PROD_DEFAULTS,
```
LalitMaganti: As above?

PR #3756, opened 2025-11-20
ui: Switch to SVG for the sched diagram in order to support theme colors

stevegolton:
Using SVG means we can apply CSS styles and colors from the theme.

Before:
image

After:
image


ui/src/assets/widgets/nodegraph.scss
stevegolton, line 1: Note: The changes in this file are due to a poorly scoped `svg` rule that was applying styles to all SVGs.

Issue #154, opened 2021-07-21
[UI][Feature Request] Snap to region boundaries


arghdos:
New to perfetto, so please redirect me if this question / issue is answered elsewhere.
One feature I find very useful in chrome://tracing that I haven't been able to figure out how to do in perfetto is to snap to boundaries, e.g.:

![image](https://user-images.githubusercontent.com/6463881/126547387-0aaf1c07-ce8d-4ee6-84cf-a8975e7aab09.png)

This lets the user easily measure gaps between events / regions / what have you.
In the [version of the UI I've been using](https://github.com/google/perfetto/tree/e5293f1cf1800d5776f4bfe96d1ec0aeec30cd46/ui), I haven't been able to figure out how to achieve the equivalent:

![image](https://user-images.githubusercontent.com/6463881/126546590-0dd5326d-64d2-419c-a826-6a6b5e3852ba.png)

You can get an approximate duration by dragging the cursor between two points, but a snap-to-boundary would be much more convenient.

primiano:
I agree on the convenience.
This is a nice-but-not-critical feature and there are too many, more important features on top so will take a while until we get here (but it's a good bug/proposal).

Tangentially related, you can get a region around a slice if you select a slice and press 'm'.
But that isn't quite the same of snapping to boundaries (e.g. doesn't work for gaps)

userkang:
I'm looking for the same requirement. Is there a specified support period available?

yohaigevim:
Agree with @skyreflectedinmirrors, adding a feature similar to Chrome Tracing's Timing Mode would improve Perfetto's ability to measure time intervals across multiple slices or gaps. Current tools like area selection and the 'm' shortcut are useful but lack the precision needed for measuring over multiple events. This functionality would make time analysis more straightforward and effective

yohaigevim:
Hope I’m not nagging, but do you have a rough estimate on the timeframe for this feature?

LalitMaganti:
There are no estimates for this. It's not currently on our list of things we are working on.

LalitMaganti:
https://github.com/google/perfetto/pull/3760 implements this feature.

PR #3760, opened 2025-11-20
ui: add api for snapping to items in tracks and implement for slices

LalitMaganti:
This CL implements a new "snapping" idea to the track API: this is a new
fully optional API for tracks to implement which allows the user when
doing area selections to "snap" to the boundary (or even some internal
timestamp) of some item in the track.

The main motivating reason for this is slice boundaries: people often
want to measure the "inter slice" duration or "the duration from start
of slice A to end of slice B": in these situations if they wanted more
precise measurements, they'd have to zoom in on that region.

Instead, with "snapping" it goes to the nearest slice when the user's
cursor is "close" to the slice.

Snapping occurs in the following situations:
* end of an area selection is close to a slice
* start or end drag handles are moved close a slice
* importantly it does *not* work for area selection start: I personally
think snapping here would feel weird to the user.

Also importantly we *only* snap to any slices on the track the user's
mouse is currently over: this makes it both so that it's more efficient
and that it's easier for the user to understand what's going on.

As for how it is visually implemented:
- we draw a vertical "dashed line" for the full length of the UI
  whenever we snap to a slice.
- hypothetically we could also draw something on the slice itsemf but in
  practice I haven't found that to be necessary: maybe something to add
  in the future if it becomes necessary

Fixes: https://github.com/google/perfetto/issues/154

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,7 +8,10 @@ Unreleased:
   Trace Processor:
     *
   UI:
-    *
+    * Added snap-to-boundaries feature for precise time selection. When
+      dragging selection handles or creating area selections, the cursor
+      automatically snaps to nearby slice boundaries to enable exact
+      measurement of slice durations. Hold Alt to temporarily disable snapping.
   SDK:
     *
 


LalitMaganti:
@stevegolton this is just FYI, Primaino has said he's happy to review this bu ccing you on it as well.

PR #3761, opened 2025-11-20
ui: improve focus feature consistency and behavior

LalitMaganti:
Fixes inconsistent focus behavior across selection types by:
- Removing janky dual-stage zoom (no more "press twice" behavior)
- Making all selections zoom to 80% viewport fill (was 20%)
- Adding aggressive zoom for instant events (99.8% zoom in)
- Supporting area selections (scrolls to top track)
- Supporting note selections (horizontal zoom only, no track scroll)
- Removing redundant ZoomOnSelection command

The F key now consistently pans horizontally, zooms, and scrolls
vertically (when applicable) for all selection types.

Bug: https://github.com/google/perfetto/issues/3550


LalitMaganti:
All the screenshot failures look like strict improvements to me. Will fix them once we're happy with the shape of this CL.

stevegolton:
> Removing janky dual-stage zoom (no more "press twice" behavior)

The current implementation of this is bad, but this feature has been around for so long that I fear folks might complain if it's removed. I do not, however, have any evidence for this, I could be wrong and no-one likes it.

> Making all selections zoom to 80% viewport fill (was 20%)

I'm unsure about this one for slices. For area selections yes, but I'm not sure that filling the screen with a slice is necessarily useful? Especially on an ultrawide monitor. Feel free to disagree.

> Adding aggressive zoom for instant events (99.8% zoom in)

Don't love this but unable to put my finger on why not. Something to do with how the zoom compounds every time you press F.

> Supporting area selections (scrolls to top track)
> Supporting note selections (horizontal zoom only, no track scroll)
> Removing redundant ZoomOnSelection command

Nice!

LalitMaganti:
> The current implementation of this is bad, but this feature has been around for so long that I fear folks might complain if it's removed. I do not, however, have any evidence for this, I could be wrong and no-one likes it.

I think people might complain but honestly I think on the balance of probabilities, more people will be happy as a result of this change than people will be annoyed. And I don't think this is one of those annoyances where people will be *too* unhappy.

> I'm unsure about this one for slices. For area selections yes, but I'm not sure that filling the screen with a slice is necessarily useful? Especially on an ultrawide monitor. Feel free to disagree.

I would argue yes because often you click focus because you want to look at the *children* of the slice (i.e. you select a parent and want to look at the child stacks)>

I could be convinced to turn this to maybe 60/70% to keep more context but I think the previous way was not zooming enough.

> Don't love this but unable to put my finger on why not. Something to do with how the zoom compounds every time you press F.

Isn't that what is expected to happen with instants? By definition they are a point so "focus" on them means really to zoom in them.

PR #3762, opened 2025-11-20
perfetto: make proto generation optional in gen_all

LalitMaganti:
Today, touching a .proto requires running gen_all to regenerate
checked-in sources (.gen.cc/h, .pcz.h, descriptors etc), causing
churn in code reviews.

This change makes proto generation opt-in via --protos flag, with
a daily bot that syncs generated code automatically. This allows
.proto and generated code to temporarily diverge.

Changes:
- tools/gen_all now skips proto generation by default
- Use --protos flag to explicitly generate proto files
- Added sync-protos GitHub Action that runs daily at 2AM UTC
- Bot closes old PRs and creates fresh ones with latest changes
- CI no longer checks proto sync (gen_all --check-only)

Developers can still run gen_all --protos manually when the .proto
definition affects checked-in code needed in the same CL.

### DIFF
--- a/.github/workflows/sync-protos.yml
+++ b/.github/workflows/sync-protos.yml
@@ -49,10 +49,10 @@ jobs:
         id: check_changes
         run: |
           # Check if there are any changes in generated proto files
-          if git diff --quiet; then
-            echo "has_changes=false" >> $GITHUB_OUTPUT
-          else
+          if [ -n "$(git status --porcelain)" ]; then
             echo "has_changes=true" >> $GITHUB_OUTPUT
+          else
+            echo "has_changes=false" >> $GITHUB_OUTPUT
           fi
 
       - name: Close existing proto sync PRs


.github/workflows/sync-protos.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Generate proto files
+        run: |
+          tools/gen_all --protos out/proto_sync
+
+      - name: Check for changes
+        id: check_changes
+        run: |
+          # Check if there are any changes in generated proto files
+          if git diff --quiet; then
```
primiano: git diff won't pick new files which are not versioned already
This should be

```
if [ -n "$(git status --porcelain)" ]; then
   echo "has_changes=true" >> $GITHUB_OUTPUT
else
 echo "has_changes=false" >> $GITHUB_OUTPUT
fi
```

PR #3763, opened 2025-11-20
android_sdk: Update PerfettoTrackEventBuilder for the new addFlow api

nicomazz:
This commit updates PerfettoTrackEventBuilder to use the new addFlow and addTerminatingFlow APIs, mainly needed to support recomposition tracing. Previously, only a single flow id was supported from the java api.

Bug: 460413832


LalitMaganti:
Kirill is the right person to review. Removing myself.

PR #3764, opened 2025-11-20
ui: Fix jittery area selection box rect when zoomed in

stevegolton:
Clamp the selection ends of the area selection rect to the bounds of the viewport before rendering to avoid 32bit float precision issues with super wide rects.

Fixes: https://b.corp.google.com/issues/462389449


LalitMaganti:
You should file a fixit bug and take this for yourself :)

PR #3767, opened 2025-11-20
base: switch Android flags to use #define instead of constexpr

LalitMaganti:
Makes it possible to use in ifdefs etc instead of just ifconstexpr
(though I would always prefer to still use ifconstexpr to get as wide
compiler coverage as possible).

### DIFF
--- a/src/tracing/service/trace_buffer_v1.cc
+++ b/src/tracing/service/trace_buffer_v1.cc
@@ -940,7 +940,7 @@ TraceBufferV1::TraceBufferV1(CloneCtor, const TraceBufferV1& src)
   //    old behavior.
   // 2. To support properly cloning of write_into_file sessions (b/382209797).
   index_ = ChunkMap(src.index_);
-  if (!PERFETTO_FLAGS(BUFFER_CLONE_PRESERVE_READ_ITER)) {
+  if constexpr (!PERFETTO_FLAGS(BUFFER_CLONE_PRESERVE_READ_ITER)) {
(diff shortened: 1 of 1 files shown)


src/tracing/core/shared_memory_arbiter_impl.cc
```diff
@@ -421,7 +421,7 @@ void SharedMemoryArbiterImpl::UpdateCommitDataRequest(
         FlushPendingCommitDataRequests();
       } else {
         bool should_post_immediate_flush = true;
-        if (base::flags::sma_prevent_duplicate_immediate_flushes) {
```
sashwinbalaji, line 424: should we switch to #if move things to compile time itself?

LalitMaganti: We should really add constexpr

src/tracing/service/trace_buffer_v1.cc
```diff
@@ -940,7 +940,7 @@ TraceBufferV1::TraceBufferV1(CloneCtor, const TraceBufferV1& src)
   //    old behavior.
   // 2. To support properly cloning of write_into_file sessions (b/382209797).
   index_ = ChunkMap(src.index_);
-  if (!base::flags::buffer_clone_preserve_read_iter) {
+  if (!PERFETTO_FLAGS(BUFFER_CLONE_PRESERVE_READ_ITER)) {
```
sashwinbalaji: this one also

PR #3768, opened 2025-11-20
Fix protozero_descriptor_diff deps

bungeman:
The `deps` on the binary need to match the `binary_label`.

The mismatch appears to be causing Chromium's `win11-arm64-blink-rel` trybot to fail. This is a x86_64 host building for arm64. The build rules currently state to use the host binary for the action but then only depend on the target binary being built.


bungeman:
Note that I have no permissions here, so after enough approvals I suspect someone else will need to land this.

Also, I'm not sure why `ui` is failing, looks like an infra issue? I don't seem to be able to re-run failing checks myself, so I'm not sure if this is persistent.

PR #3775, opened 2025-11-20
DynamicString avoid dangling pointer

sashwinbalaji:
Delete constructor for rvalue std::string so at compile time itself we caqn detect if a temporary one is used.

Bug: b/417982839
Bug: #3776 3776

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -13,7 +13,8 @@ Unreleased:
       automatically snaps to nearby slice boundaries to enable exact
       measurement of slice durations. Hold Alt to temporarily disable snapping.
   SDK:
-    *
+    * Breaking change: `DynamicString` no longer accepts temporary `std::string`
+      arguments to prevent use-after-free bugs.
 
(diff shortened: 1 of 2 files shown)


include/perfetto/tracing/string_helpers.h
```diff
@@ -55,6 +55,8 @@ class PERFETTO_EXPORT_COMPONENT DynamicString {
   }
   DynamicString(const char* str, size_t len) : value(str), length(len) {}
   constexpr DynamicString() : value(nullptr), length(0) {}
+  // Disallow construction from temporary strings to avoid use-after-free.
+  DynamicString(std::string&&) = delete;
```
LalitMaganti, line 60: Did you test compile actually fails in this case?

sashwinbalaji: yes, also can be seen from failing tests: https://github.com/google/perfetto/actions/runs/19554216601/job/55992888178

Issue #1000, opened 2025-02-01
feature: relax requirements for emitting clock snapshots


dreveman:
[ClockSnapshot](https://android.googlesource.com/platform/external/perfetto/+/refs/heads/main/protos/perfetto/trace/clock_snapshot.proto) packet currently [must be emitted before the custom clock is referred to by any TracePacket written by the same TraceWriter](https://android.googlesource.com/platform/external/perfetto/+/refs/heads/main/docs/concepts/clock-sync.md)

Removing this requirement makes custom clocks easier to use and enables the use-case described below.

### Cycle accurate simulator

Cycle accurate simulators advance one or more cycle counters internally when executing workloads. The cycle number is technically the clock that is used to determine performance and what the user of the simulator cares about. This cycle counter doesn't map to the trace time in a predetermined way. The simulator might take a long time to simulate something that takes only a cycle on real hardware, and a short time to simulate something that is really expensive on real hardware. Being able to visualize simulator work based on the cycle count allows perfetto to be used instead of alternative tools to visualize workloads running on simulators. It's also allowing the work to be connected to host side events, which makes it much easier to understand the overall flow of an application.

We need to map events that uses the cycle count to the trace clock. Custom clocks works well except for the requirement that clock snapshots needs to be emitted before track events referring to the clock. The problem is that we don't know how what the unit multiplier for the cycle clock is until the interval that we want to map to trace time has completed. Being able to emit clock snapshots when the interval ends solves this problem.

### Alternatives

1. Manually collect all events in the app and emit them at the end of the "interval". This puts a lot of burden on the application and can result in a large increase in memory usage as we can't flush when desirable and significantly impact application performance.
2. Simply use nanos as cycles and require the user to zoom in and be aware of this assumption.

LalitMaganti:
I've commented on the patch but basically we need to
1) to keep track of only the packets with the missing clock id
2) inform the sorter that you need to do a full sort

dreveman:
> I've commented on the patch but basically we need to
> 
> 1. to keep track of only the packets with the missing clock id
> 2. inform the sorter that you need to do a full sort

Yes, I saw your comment. I looked at the sorting code in more detail yesterday so I understand that better now and you're comment and suggestion makes sense. I'll update the patch as suggested asap.

dreveman:
Fixed with this commit: https://android.googlesource.com/platform/external/perfetto/+/ae42fce6c2d7d328bb32188fcc027f385f8b4864

[...]

PR #3778, opened 2025-11-21
tp: only allow missing clocks when full sort is enabled

dreveman:
This allows us to continue to generate errors for missing timestamps in the default case shen partial sorting is used and defer the errors until EOF when full sorting is used.

Also don't flush and force event extraction before EOF as that prevents deferred packages from being processed.

### DIFF
--- a/include/perfetto/trace_processor/trace_processor_storage.h
+++ b/include/perfetto/trace_processor/trace_processor_storage.h
@@ -45,6 +45,9 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessorStorage {
   // See comment on TraceProcessor::Flush.
   virtual void Flush() = 0;
 
+  // See comment on TraceProcessor::ProcessEndOfFileDeferredPackets.
+  virtual base::Status ProcessEndOfFileDeferredPackets() = 0;
+
   // See comment on TraceProcessor::NotifyEndOfFile.
   virtual base::Status NotifyEndOfFile() = 0;
 };
--- a/src/trace_processor/forwarding_trace_parser.cc
+++ b/src/trace_processor/forwarding_trace_parser.cc
@@ -146,6 +146,13 @@ base::Status ForwardingTraceParser::Parse(TraceBlobView blob) {
   return reader_->Parse(std::move(blob));
 }
 
+base::Status ForwardingTraceParser::ProcessEndOfFileDeferredPackets() {
+  if (reader_) {
+    RETURN_IF_ERROR(reader_->ProcessEndOfFileDeferredPackets());
+  }
+  return base::OkStatus();
+}
+
 base::Status ForwardingTraceParser::NotifyEndOfFile() {
   if (reader_) {
     RETURN_IF_ERROR(reader_->NotifyEndOfFile());
--- a/src/trace_processor/forwarding_trace_parser.h
+++ b/src/trace_processor/forwarding_trace_parser.h
@@ -38,6 +38,7 @@ class ForwardingTraceParser : public ChunkedTraceReader {
 
   // ChunkedTraceReader implementation
   base::Status Parse(TraceBlobView) override;
+  [[nodiscard]] base::Status ProcessEndOfFileDeferredPackets() override;
   [[nodiscard]] base::Status NotifyEndOfFile() override;
 
   TraceType trace_type() const { return trace_type_; }
--- a/src/trace_processor/importers/archive/gzip_trace_parser.cc
+++ b/src/trace_processor/importers/archive/gzip_trace_parser.cc
@@ -125,6 +125,10 @@ base::Status GzipTraceParser::ParseUnowned(const uint8_t* data, size_t size) {
   }
 }
 
+base::Status GzipTraceParser::ProcessEndOfFileDeferredPackets() {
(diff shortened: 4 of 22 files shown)


LalitMaganti:
Nope sorry this change is definitely *not* the right fix.

please tell me the problem not provide a solution :)

LalitMaganti:
The reaosn for the above: Perfetto supports lots of other trace types, this will break many of them. It will also likely break many sub-category of proto traces as well

dreveman:
@LalitMaganti with improved support for full sort I think limiting eof deferred packets makes sense but there's other parts of this that I'm less sure about. e.g. I need to replace the `last_blob.offset_in_blob <= tbv.offset()` check with a fallback. It's easy to see why removing the early flushes are needed but not clear if that could cause some other harm.

LalitMaganti:
The whole eof deferred packets was not fully implemented and should just be removed from the codebase. It was only done partially and has a bunch of edge cases. I would really want to just pull that out.

dreveman:
Being able to resolve clocks late is critical for our simulator workloads and we'd have to look elsewhere if perfetto can't support that. Seems like it's close to supporting it but maybe that's just us not understanding some details? Maybe there's some alternative approach we could take to solve this problem instead of abusing the existing clock system?

LalitMaganti:
Can you explain more on your problem? Why do you "need to resolve clocks late"? The more I understand, the better I can come up with some solution.

I'm not opposed to supporting your usecase but I need to understand it better. This fix is not in the right direction.

One potential fix is to have a field in the TracePacket which tells us "this sequence requires late clock resolution" which means Perfetto can know to not do windowed sort (i.e. it will fall back to full sorting always).

dreveman:
> Can you explain more on your problem? Why do you "need to resolve clocks late"? The more I understand, the better I can come up with some solution.
> 
> I'm not opposed to supporting your usecase but I need to understand it better. This fix is not in the right direction.
> 
> One potential fix is to have a field in the TracePacket which tells us "this sequence requires late clock resolution" which means Perfetto can know to not do windowed sort (i.e. it will fall back to full sorting always).

sorry, I should have linked to the original issue: https://github.com/google/perfetto/issues/1000

originally we released deferred packets when clock snapshots came in and that allowed them to be released as needed instead of at eof and that avoided all the changes in this PR. maybe we can go back to that?

LalitMaganti:
> sorry, I should have linked to the original issue: https://github.com/google/perfetto/issues/1000

Ahhhhh I remember this now. So you were the one who added the original code! right sorry I didn't put that together I thought it was someone else 😓 

I started looking into separating flushes into multiple stages as I suggested in https://android-review.git.corp.google.com/c/platform/external/perfetto/+/3505773 but I thought it wasn't worth because it wasn't followed up on. I can resurrect that work if necessary.

I also really think is that we allow marking events as "processed late" though which will turn on a speical mode of processing where we defer packets for processing.

dreveman:
Yes, sorry for not following up on this until now. It's been less urgent for us as we've been using this patch internally without any issues since. I can continue the work on "separating flushes into multiple stages" if you think that's a decent approach? 

I'm not against marking packets as "processed late" but the flushing issue would still have to be resolved, right? Or are you thinking we we have another event to indicate that "processed late" packets can be processed?

LalitMaganti:
>  I can continue the work on "separating flushes into multiple stages" if you think that's a decent approach?

Yes that would be ideal.

> I'm not against marking packets as "processed late" but the flushing issue would still have to be resolved, right?

I think the idea would be that these packets which are "processed late" would *not* be pushed through the queue until NotifyEndOfFile.

dreveman:
@LalitMaganti Sorry for the delay with this. Is the latest version of this change something like what you had in mind? I wasn't sure about the name of the new function or if I missed some place where it'd make sense to implement it beyond what I'm doing now. I implemented in enough places to make the UI and trace_processor_shell work for me.

include/perfetto/trace_processor/trace_processor_storage.h
```diff
@@ -45,6 +45,9 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessorStorage {
   // See comment on TraceProcessor::Flush.
   virtual void Flush() = 0;
 
+  // See comment on TraceProcessor::ProcessEndOfFileDeferredPackets.
+  virtual base::Status ProcessEndOfFileDeferredPackets() = 0;
```
LalitMaganti, line 49: This should definitely *not* be a public function. Also this is just what the semantic of Flush should be...

dreveman:
> Hey sorry no this is not at all in the direction I wanted. As I said with just Flush and NotifyEndOfFile there should be a way to do this.
> 
> Happy to take a look at this myself. Sorry, have a lot of opinions on the correct way this should be done because this is has signifcant impact to how lots of thigns in TP work.

Ok, no problem. I'm not very familiar with this code so I'll just wait for you to take a look or give me some more directions for how it should be done

LalitMaganti:
I started work on this at https://github.com/google/perfetto/pull/4182. Once I test this, it should be all that's needed for your traces to work. If not, I'd like to see an example of a trace so I can look into this more.

LalitMaganti:
https://github.com/google/perfetto/pull/4666 should be an alternative way to fix this issue.

LalitMaganti:
Closing as this should now be fixed.

PR #3780, opened 2025-11-21
ui: introduce a new set of presets and signifcantly improve record page

LalitMaganti:
This CL adds the idea of "presets" to the recording page for Android and
Linux. This basically is "preset saved configs": configs people can
start recording traces on the UI without needing to go and manually
configure knobs.

While doing this, I also realised we should really merge the cmdline
instructions, saved configs and record pages into one page which
efficiently lays out all the information: there was no need for that to
spread across three pages.

So implemented a nice merged page using all the great card/grid widgets
similar in style to the overview page. The idea now is that you
literally don't need to leave the overview page to record a trace:
everything can be done from that one page.

### DIFF
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/presets.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/presets.ts
@@ -63,7 +63,15 @@ const ATRACE_BATTERY = [
   'wm',
 ];
 
-const ATRACE_THERMAL = ['freq', 'idle', 'power', 'sched', 'thermal', 'workq'];
+const ATRACE_THERMAL = [
+  'freq',
+  'idle',
+  'power',
+  'sched',
+  'thermal',
+  'thermal_tj',
+  'workq',
+];
 
 const LOGCAT_DEFAULT = [
   protos.AndroidLogId.LID_DEFAULT,
@@ -144,6 +152,7 @@ export const ANDROID_PRESETS: Preset[] = [
       fileWritePeriodMs: 2500,
       compression: false,
       probes: {
+        cpu_sched: {settings: {}},
         atrace: {
           settings: {
             categories: ATRACE_THERMAL,


LalitMaganti:
@stevegolton just adding you as an FYI: Primiano should be able to review this since it just touches recording page.

primiano:
Some comment purely based on opening the UI link without looking at code:

1. I love 90% of it
2. I am not a huge fan of the cmdline instructions being moved as a "transport", for the following reasons:
 - Is disorienting to existing users with little reason. A bunch of people will be confused as it seems the cmdline has disappeared (it took me a minute or so to figure out that it wasn't a bug and was moved to transport)
 -  I think the cmdline case is common enough that warrants its own section. the overview page where you added it is getting too cluttered 
 - There is a UX problem introduced by this: it adds one more click to regenerate the cmdline instructions: if you open a probe page (Say memory and you toggle heap profile) and want to generate new cmdline, you need to go back to the overwiew page... but  as you do the WebUsb target is selected again by default (because "targets" stick only after they are connected)


I like how you rearranged user presets, much nicer.
But I'd move back cmdline instructions where it was

LalitMaganti:
> I like how you rearranged user presets, much nicer.
> But I'd move back cmdline instructions where it was

Done in the latest version: PTAL.

stevegolton:
No complaints from me, looks great!

ui/src/plugins/dev.perfetto.RecordTraceV2/pages/target_selection_page.ts
```diff
@@ -119,15 +228,192 @@ class TransportSelector implements m.ClassComponent<RecMgrAttrs> {
...
+      Card,
+      {
+        className: `pf-preset-card${extraClass}${isSelected ? ' pf-preset-card--selected' : ''}`,
+        onclick,
+        tabindex: 0,
+      },
+      m(Icon, {icon}),
+      m('.pf-preset-card__title', title),
+      m('.pf-preset-card__subtitle', subtitle),
+      isSelected &&
```
rsavitski: driveby nit: it's odd that this element invents its own styling for "currently selected", while being nested in the record page which uses inverted backgrounds to mark the selected option.

LalitMaganti: Good catch. I've made it consistent.

ui/src/plugins/dev.perfetto.RecordTraceV2/pages/target_selection_page.ts
```diff
@@ -53,40 +71,131 @@ export function targetSelectionPage(recMgr: RecordingManager): RecordSubpage {
...
-      await recMgr.setProvider(prov);
-      if (state.target.targetId === undefined) return;
-      for (const target of await recMgr.listTargets()) {
-        if (target.id === state.target.targetId) {
-          await recMgr.setTarget(target);
+      if (prov !== undefined) {
+        await recMgr.setProvider(prov);
+      }
+
+      // Restore target selection
```
primiano, line 95: As discussed offline, rather than doing all this reverse matching, i'd remember the preset name, and re apply it.
If the preset name is the "custom" one then you reload the saved probes rather than reloading the preset as if the user did "blank" and then clicked on the quick start config

ui/src/plugins/dev.perfetto.RecordTraceV2/config/config_manager.ts
```diff
@@ -40,6 +40,8 @@ export class ConfigManager {
   private enabledProbes = new Map<string, boolean>();
   private indirectlyEnabledProbes = new Map<string, Set<string>>();
 
+  constructor(private readonly onProbeChange?: () => void) {}
```
primiano: rather than exposing an onChange callback, which feels a bit of a hack can we just expose a generation counter, (incremented on every change), so you can use that in recording_manager.ts without injecting callbacks, just by checking the gen

primiano:
Ah sory i forgot one thing, you can undo the chnage in the instructions_page.ts
I tried and this still works

```
diff --git a/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/instructions_page.ts b/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/instructions_page.ts
index 79643ce96d..ac9ae89133 100644
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/instructions_page.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/instructions_page.ts
@@ -43,17 +43,8 @@ class InstructionsPage implements m.ClassComponent<RecMgrAttrs> {
   private docsLink?: string;

   constructor({attrs}: m.CVnode<RecMgrAttrs>) {
-    this.updateConfig(attrs.recMgr);
-  }
-
-  onupdate({attrs}: m.CVnode<RecMgrAttrs>) {
-    // Regenerate config when the component updates
-    this.updateConfig(attrs.recMgr);
-  }
-
-  private updateConfig(recMgr: RecordingManager) {
     // Generate the config PBTX (text proto format).
-    const cfg = recMgr.genTraceConfig();
+    const cfg = attrs.recMgr.genTraceConfig();
     const cfgBytes = protos.TraceConfig.encode(cfg).finish().slice();
     traceConfigToTxt(cfgBytes).then((txt) => {
       this.configTxt = txt;
@@ -61,7 +52,7 @@ class InstructionsPage implements m.ClassComponent<RecMgrAttrs> {
     });

     // Generate platform-specific commands.
-    switch (recMgr.currentPlatform) {
+    switch (attrs.recMgr.currentPlatform) {
       case 'ANDROID':
         this.cmdline =
           'cat config.pbtx | adb shell perfetto' +
```

LalitMaganti:
OK we should be good for another round :)

primiano (approved):

ui/src/plugins/dev.perfetto.RecordTraceV2/presets.ts
```diff
@@ -0,0 +1,222 @@
...
+      mode: 'STOP_WHEN_FULL',
+      bufSizeKb: 64 * 1024,
+      durationMs: 30000,
+      maxFileSizeMb: 500,
+      fileWritePeriodMs: 2500,
+      compression: false,
+      probes: {
+        atrace: {
+          settings: {
+            categories: ATRACE_THERMAL,
```
sashwinbalaji, line 158: catgeory: thermal_tj (if pixel need cpm but too much overhead)
need to get atraces thermal category for: http://ac/hardware/google/pixel/thermal/Thermal.cpp
cpu_sched (as scheduling policy is ties to thermal state)

PR #3786, opened 2025-11-21
TP: remove accidental dep on protobuf_full from TP unittests

primiano:
Short version:
we ended up with an accidental dependency between
protozero_to_json_unittest.cc and protobuf_full.
@KirillTim found this while address b/447162132.
This dependency snuck in because:
- All unittests are built together
- some tools/ unittests legitimately need proto_full.

The only dependency chain is:
- One test in protozero_to_json_unittest depends on a metric.pb.h
  (which by itself does NOT warrant a pblite dependency)
- That specific metric is using a protobuf custom option
  for `[(unit) = "ms_biggerIsBetter"];`
- Unfortunately protobuf custom options end up requiring
  a dep on protobuf_full.
- So I changed the code to use pbzero.
- As part of this I needed to add "zero" to the list of
  generated metrics proto.

CONTROVERSIAL PART:
I think that was the only thing that was depending on metrics:lite.
I tried removing the "lite" targets from protos/metrics and everything
still worked.
I am bit unsure on why those "lite" targets existed in the first place.
But I worry that either chromium or google3 depend on them.
So I am doubtful on whether I should:
A) Keep lite and add "zero"
B) Replace "lite" with "zero"

I am going to do A in this PR and try B in a separate PR (#3787).

Bug: b/462688465

### DIFF
--- a/protos/perfetto/metrics/BUILD.gn
+++ b/protos/perfetto/metrics/BUILD.gn
@@ -16,7 +16,10 @@ import("../../../gn/perfetto.gni")
 import("../../../gn/proto_library.gni")
 
 perfetto_proto_library("@TYPE@") {
-  proto_generators = [ "lite" ]
+  proto_generators = [
+    "lite",
+    "zero",
+  ]
   deps = [
     "android:@TYPE@",
     "common:@TYPE@",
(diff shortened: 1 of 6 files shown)


primiano:
@KirillTim FYI

LalitMaganti (approved):
Afaik lite exists for G3. But lgtm.

primiano:
> Afaik lite exists for G3. But lgtm.

@KirillTim just made a great point:
removing "lite" (which I am doing in #3787) doesn't seem to cause any regeneration of BUILD files, so this shouldn't possibly affect any non-GN dependency.

PR #3788, opened 2025-11-21
ui: Invert record page images in dark mode

stevegolton:
The images on the record page don't work well with dark backgrounds, making the text difficult to read. This patch inverts these images and shifts the hues by 180 in dark mode to make them readable.

Before:
image

After:
image

Note: This works well for images that are moderately saturated but can tend to blow out images that were not originally particularly saturated - see the following screenshot:

image


primiano:
this was a very tough call. the team has spoken in favour of this vs the other bg alternative!
well done with the hue rotation trick

PR #3791, opened 2025-11-21
ui: Reorder sidebar navigation

aMayzner:
## Summary

Reorganizes the sidebar navigation to improve usability and reduce visual clutter. The main changes consolidate trace-related actions into a single "Trace Files" section, collapse less-frequently-used sections by default, and add platform-specific icons for example traces.

## Changes

- Merged "Navigation" and "Example Traces" sections into a unified "Trace Files" section
- Collapsed "Convert trace" and "Support" sections by default to reduce visual clutter
- Added platform-specific icons (Android, Web) for example traces to improve recognition
- Removed duplicate "Info and Stats" menu - now only available via overview
- Moved "Widgets" page to the "Settings" section

## Notes

The sidebar now presents a cleaner, more focused interface with trace-related actions grouped together at the top, making the most common workflows (opening and recording traces) more discoverable.


stevegolton:
image

The example traces feel a bit too prominent now. Maybe out of scope for this PR but perhaps they would work better in a dropdown menu - especially if we add more in the future.

- Open trace file
- Record trace
- Open example trace...
  - Android trace
  - Chrome trace

At the very least I think they should go below 'Record trace'

I'm also not a huge fan of the section title 'Trace files' - maybe 'Open trace' or 'New trace'?

ui/src/frontend/sidebar.ts
```diff
@@ -445,7 +445,14 @@ export class Sidebar implements m.ClassComponent {
     // Don't render empty sections.
     if (menuItems.length === 0) return undefined;
 
-    const expanded = getOrCreate(this._sectionExpanded, sectionId, () => true);
+    // Collapse 'convert_trace' and 'support' sections by default
+    const defaultCollapsed =
```
primiano: given that we have a switch statement a few lines above, can we move that decision there by setting a var in the `case 'support'` or `case 'convert_trace'` rather than having a secondary list here?

ui/src/plugins/dev.perfetto.RecordTraceV2/index.ts
```diff
@@ -42,11 +42,11 @@ export default class implements PerfettoPlugin {
 
   static onActivate(app: App) {
     app.sidebar.addMenuItem({
-      section: 'navigation',
+      section: 'trace_files',
```
primiano, line 44: can you make this go on the top, before Open trace?
This is my rationale: record is very frequently used and should be at the top-ish (1st or 2nd position), while example traces shoudl be below (still within the group)

but then i guess your argument is that for consistency "open trace file" should be adjacent to "open example trace", whicih also makes sense.

so the only resolution for me is moving record new trace in 1st position, so:

1. record
2. open file
3. example android
4. example chrome

primiano:
> At the very least I think they should go below 'Record trace'

Yeah I had the same feedback. Imho the right order is
1. record
2. open
3.  example

I'd not over complicate it for now with submenu.

> I'm also not a huge fan of the section title 'Trace files' - maybe 'Open trace' or 'New trace'?

Don't have strong opinions, I'm fine with this. Open trace is fine, as in a way "record" will give you a trace to open.

primiano (approved):
Thanks a lot !
LGTM. Up to you and Steve if you want to rename "Trace Files" to "Open trace", i'm fine either way

LalitMaganti:
Personally I'm not a fan of putting record above open. It's muscle memory at this point for me to go to the top page. Can we please put it below?

aMayzner:
It seems like we have to decide between:
- Record is the most common use (from the landing page) for most people so it should be first:
image
- Openinig traces it the most obvious behaviour of the UI - you load a trace and it allows you to explore the trace, also it has always been at the top, so let's keep it at the top and put Record to the bottom (as it is a separate behaviour than opening traces)
image

@LalitMaganti, @primiano, @stevegolton, WDYT? 

I personally would put Record at the bottom, to have it visually separated and not break muscle memory.

stevegolton:
I think it should be a 3rd option like this (in order of importance/frequency of use):
- Open trace file
- Record new trace
- Open android example
- Open chrome example

primiano:
> Personally I'm not a fan of putting record above open. It's muscle memory at this point for me to go to the top page. Can we please put it below?

Your muscle memory will need to be retrained anyways because we are moving the whole section.

I am fine with @stevegolton proposal. The only thing I'm not fine with is record at the bottom. So pick one between:

A) Record, Open file, Open example traces

B) Open file, Record, Open example traces

Both work for me. Pick one with Steve

LalitMaganti:
Please let's do B) :)

aMayzner:
Ok, will do B

Issue #3463, opened 2025-10-30
[Doc] How to generate "summary/overview" of track with childrens?


TApplencourt:
Hi,

First sorry to not use the correct terminology.

In the android example trace, some tracks (for example `com.google.camera...`) have a "summary track" (reflecting somehow their children activity)

<img width="1280" height="218" alt="Image" src="https://github.com/user-attachments/assets/0f9d6c6a-da88-4325-a5d4-f9a18397d09a" />

When you click on this track, you see all their children particular track.

<img width="1280" height="218" alt="Image" src="https://github.com/user-attachments/assets/5670a24e-3214-4edc-9f75-1592a058e671" />

 I was checking the doc on how to create this `summary / children overview` or the parent track (aka picture 1)  using protobuff format, but I was not able to find any information. By default, the UI just generate a "black" track (as for example the "FtraceEvent").

Sorry for the dumb question,  

Thanks!

stevegolton:
Unfortunately we don't really have a free-form way to control what is displayed in the track summary from the protobuf TrackEvent format. The summaries are derived from other trace datasources in the trace and can take two different forms depending on what information is available:

1. A histogram showing a count of how many process threads are running simultaneously - you can see an example of this in the Chrome example trace.
2. A mini scheduling overview for that process based on the scheduling information gleaned from ftrace events - this is what you can see on the Android example trace. The summary track is split up into 8 mini tracks - one per CPU - and a miniature slice is rendered any time a thread within that process is running on a given CPU. It's similar to the CPU tracks above but just for one process.

I'll acknowledge this is a limitation, but I'm also not aware of any plans to make this more configurable. What were you hoping to show on these summaries in an ideal world?

Edit: Typo

stevegolton:
If you're looking to have a way to group tracks while retaining control of what is displayed on the the group tracks, you can arbitrarily nest tracks using the TrackEvent `parent_uuid` field like so: https://perfetto.dev/docs/getting-started/converting#track-hierarchies-for-waterfall-trace-views

This will let you control track nesting arbitrarily and also allow you to display slices on the parent tracks as well as the child tracks.

TApplencourt:
Thanks a lot for the quick reply! 

>This will let you control track nesting arbitrarily and also allow you to display slices on the parent tracks as well as the child tracks.

Yes, and thanks a lot for that! It's really nice feature, and we use it abundantly! 

[...]

PR #3793, opened 2025-11-21
ui: add summary track for slice tracks and merge with scheduling summary

LalitMaganti:
This CL unifies the process scheduling summary track and the slice summary track together into a single summary track type. The goal of this track is to always show some summary of either scheduling or slice like data, whichever is available.

Schedulling is preferred if available and will use the same "one track per CPU" type which it does today. For example:
image

If not available, we also get the first 8 non-summary slice tracks (more precisely, slice tracks with datasets associated to them) and show a representation of the "depth 0" slices active on that track:

image

The implementation of this summary is not ideal in several ways:
1) The choice of track only happens when a depth column exists (i.e. slices without a depth don't work)
2) The binding of tracks to the dataset happens at non-optimal time due to how the renderer <-> track node relationship works.

But having something is significantly better than nothing so we're going to land this now and follow up with improvements later.

Fixes: https://github.com/google/perfetto/issues/3463


ui/src/plugins/dev.perfetto.ProcessSummary/index.ts
```diff
@@ -25,41 +25,53 @@ import ThreadPlugin from '../dev.perfetto.Thread';
...
+    // We need to do this because for slice tracks, other plugins might be
+    // adding tracks in onTraceLoad which we won't see here. So we wait for
+    // traceready to bind the tracks.
+    ctx.onTraceReady.addListener(async () => {
+      for (const [uri, track] of this.trackUrisToResolve) {
+        const node = ctx.defaultWorkspace.tracks.getTrackByUri(uri);
+        if (!node) {
+          continue;
+        }
+        track.fetchDatasetsFromSliceTracks(node);
```
stevegolton: While this works, it has some significant downsides:

1. Latch on first call - Whatever children exist when the plugin calls this, that's what you get forever.
2. Hidden timing contract - Caller must call this before onCreate, but nothing enforces that.
3. No access later - The track doesn't have the TrackNode after this, so it can't re-query children even if it wanted to.

While I think accepting 1 is pragmatic for now (no live updates), rather than having the plugin call fetchDatasetsFromSliceTracks manually, let's add trackNode to the TrackContext passed to lifecycle hooks, and latch the list of child tracks onCreate() when we create the mipmap table. This way:
- Plugins don't need to know about this orchestration so they cannot get the timing wrong.
- The track gets live access to the list of tracks when it needs them so we can implement live updates on top of this if we want to at a later date.

This approach is actually simpler than your approach in this PR, I think it'll basically involve:
- Add `trackNode: TrackNode` to the TrackContext interface.
- Inject the node `trackNode: node` in TrackView.
- Use `ctx.trackNode` to glean the list of child tracks if mode === 'slice' in side onCreate.

LalitMaganti: Thanks for the suggestion! Please see the latest state whcih should align with your request.

PR #3795, opened 2025-11-21
ui: improve performance of async slice depth computation

LalitMaganti:
Do it in one shot at the start of the trace instead of once per track.
Also this allows for signifcantly better creation of union datasets as
we're now able to correctly optimize as the strings are now constants.

### DIFF
--- a/ui/src/plugins/dev.perfetto.TraceProcessorTrack/trace_processor_slice_track.ts
+++ b/ui/src/plugins/dev.perfetto.TraceProcessorTrack/trace_processor_slice_track.ts
@@ -146,7 +146,7 @@ async function getDataset(
           select
             id,
             layout_depth as depth,
-            ${Math.min(...trackIds)}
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.TraceProcessorTrack/trace_processor_slice_track.ts
```diff
@@ -125,21 +133,25 @@ async function getDataset(engine: Engine, trackIds: ReadonlyArray<number>) {
...
-        id,
-        layout_depth as depth
-      from experimental_slice_layout('${trackIds.join(',')}')
-    `);
+    if (depthTableName === undefined) {
+      // Create the depth table if not provided by caller
+      await createPerfettoTable({
+        name: tableName,
+        engine,
+        as: `
```
stevegolton, line 145: This table is missing a minTrackId column, which is required by the dataset below

stevegolton (approved):

PR #3803, opened 2025-11-21
CI: run perfetto unit tests with `--gtest_shuffle`

KirillTim:
This is just a good practice to do so.

### DIFF
--- a/src/base/file_utils.cc
+++ b/src/base/file_utils.cc
@@ -190,8 +190,6 @@ ssize_t WriteAll(int fd, const void* buf, size_t count) {
 }
 
 base::Status CopyFileContents(int fd_in, int fd_out) {
-  // TRIGGER CI BUILD!!! OLOLOLO!
-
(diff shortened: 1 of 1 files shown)


src/base/file_utils.cc
```diff
@@ -190,6 +190,8 @@ ssize_t WriteAll(int fd, const void* buf, size_t count) {
 }
 
 base::Status CopyFileContents(int fd_in, int fd_out) {
+  // TRIGGER CI BUILD!!! OLOLOLO!
```
sashwinbalaji: do we want to keep this ?

KirillTim: Of course not :)
I'll remove this after we get the results from the CI. 

Without the code change the CI won't trigger the build and test workflow

primiano: Usually I add a dot somewhere just in case I forget :)

primiano: I mean in a MD file, like I just did here https://github.com/google/perfetto/pull/3804 :)

PR #3806, opened 2025-11-21
traceconv profile: auto-detect profile type

rsavitski:
Right now, "traceconv profile" means allocator profiles, while perf
profiles require an explicit flag ("traceconv profile --perf"). This is
cumbersome to most users.

Make the tools automatically detect between three types of profiles:
* allocator (heapprofd)
* perf
* android java heap graphs

If the trace contains more than a single type of profile, the tool asks
to be reinvoked with an explicit flag.

Bug: 463307876

### DIFF
--- a/include/perfetto/profiling/pprof_builder.h
+++ b/include/perfetto/profiling/pprof_builder.h
@@ -47,11 +47,7 @@ struct SerializedProfile {
   std::string heap_name;
 };
 
-enum class ConversionMode : uint64_t {
-  kHeapProfile,
-  kPerfProfile,
-  kJavaHeapProfile
-};
+enum class ConversionMode { kHeapProfile, kPerfProfile, kJavaHeapProfile };
 
 enum class ConversionFlags : uint64_t {
   kNone = 0,


docs/quickstart/traceconv.md
```diff
@@ -37,35 +37,6 @@ chmod +x traceconv
...
-
-However if you are using callstack sampling like traced_perf then use the
-following instead:
-
-`~/traceconv profile [input proto file] --perf`
-
-Note for `--perf` the output is one pprof file per process sampled in the trace.
-You can use pprof to merge them together if desired.
-
-You can also specify an output directory for the profiles using `--output-dir`:
```
sashwinbalaji, line 65: I would prefer to document output-dir as otherwise this would result in a run failure and then user has to rerun with modification

src/traceconv/main.cc
```diff
@@ -311,18 +309,20 @@ int Main(int argc, char** argv) {
...
-    return perf_profile
-               ? TraceToPerfProfile(input_stream, output_stream, pid,
-                                    timestamps, !profile_no_annotations,
-                                    output_dir)
-               : TraceToHeapProfile(input_stream, output_stream, pid,
-                                    timestamps, !profile_no_annotations,
-                                    output_dir);
+    if (positional_args.size() > 2) {
+      PERFETTO_ELOG(
+          "-o is not supported for profile, use --output-dir instead");
```
sashwinbalaji: how about saying output-file instead of -o as there is no -o option ?

src/traceconv/main.cc
```diff
@@ -311,18 +309,20 @@ int Main(int argc, char** argv) {
...
+      return Usage(argv[0]);
+    }
+    return TraceToProfile(input_stream, pid, timestamps,
+                          !profile_no_annotations, output_dir, profile_type);
   }
 
   if (format == "java_heap_profile") {
```
sashwinbalaji, line 322: print a log asking user to switch to profile --java-heap so hopefully sometime in future we can remove this

include/perfetto/profiling/pprof_builder.h
```diff
@@ -47,7 +47,11 @@ struct SerializedProfile {
   std::string heap_name;
 };
 
-enum class ConversionMode { kHeapProfile, kPerfProfile, kJavaHeapProfile };
+enum class ConversionMode : uint64_t {
```
sashwinbalaji: nit^nit: ordering

src/traceconv/trace_to_profile.cc
```diff
@@ -112,34 +110,104 @@ std::string GetDestinationDirectory(
...
+  int64_t count = alloc_present + perf_present + graph_present;
+  if (count == 0) {
+    PERFETTO_LOG("No profiles found.");
+    return std::nullopt;
+  } else if (count > 1) {
+    PERFETTO_ELOG(
+        "More than one type of profile found in the trace, pass an explicit "
+        "disambiguation flag (--alloc / --perf / --java-heap).");
+    return std::nullopt;
+  }
```
sashwinbalaji, line 144: As user, I would prefer to know what all type of profiles were present

docs/quickstart/traceconv.md
rsavitski: --output-dir is documented in the CLI usage, which is the authoritative source imo.

I don't think this docs/ page should do much more than highlighting the existence of traceconv (and a couple of its primary modes), and how to download traceconv. The rest is covered by CLI usage.

include/perfetto/profiling/pprof_builder.h
rsavitski: Sorry, what do you mean? Reordering enum options to be alphabetic?

src/traceconv/main.cc
rsavitski: Oops, done.

src/traceconv/main.cc
rsavitski: I personally think it's fine to keep as is, it's not a lot of cognitive load on either devs or users to have this option around.

src/traceconv/trace_to_profile.cc
rsavitski: Done.

include/perfetto/profiling/pprof_builder.h
sashwinbalaji: yes, it was about being alphabetical

rsavitski: Right, I don't think that reordering to be alphabetic is a strict benefit (right now the options are ordered chronologically, which is also a valid ordering). And actually, dropping this file from the patch since the edit was superfluous in the first place.

sashwinbalaji: hence the nit^nit :)

PR #3810, opened 2025-11-21
Fix for BTP queries not providing trace filename inside metadata

edgararriagag:
This patch fixes BTP queries when using local file resolver to contain filename within the query which is especially helpful when trying to pin point which traces are linked to certain outliers among other things.

This is how a BTP query looks before the fix:
```
                 ts           name dur_ms                
0   606266286752528  activityStart    774
1   606879008114196  activityStart    239
```

And this is how it looks after the fix:
```
                 ts           name dur_ms                                              _path
0   606266286752528  activityStart    774  profile_trigger-type-1_2025-11-18-19-25-09.per...
1   606879008114196  activityStart    239  profile_trigger-type-1_2025-11-18-19-35-21.per...
```

### DIFF
--- a/python/test/api_integrationtest.py
+++ b/python/test/api_integrationtest.py
@@ -214,9 +214,16 @@ class TestApi(unittest.TestCase):
         self.assertEqual(row.dur, dur_result[num])
 
   def test_simple_resolver(self):
+    sample_path = example_android_trace_path()
     dur = [178646, 178646, 178646, 178646]
     source = ['generator', 'path', 'path_resolver', 'file']
-    expected = pd.DataFrame(list(zip(dur, source)), columns=['dur', 'source'])
+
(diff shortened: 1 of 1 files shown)


python/perfetto/trace_uri_resolver/path.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, path: str):
   def resolve(self) -> TraceUriResolver.Result:
     return [
         TraceUriResolver.Result(
-            trace=util.file_generator(self.path), metadata=dict())
+            trace=util.file_generator(self.path), metadata={'trace': self.path})
```
LalitMaganti: As discussed, the key for this direciton should be _path.

edgararriagag: Done

PR #3814, opened 2025-11-24
ProtoVM integration

rukkal:


src/tracing/service/tracing_service_impl.h
```diff
@@ -367,7 +376,7 @@ class TracingServiceImpl : public TracingService {
   std::map<ProducerID, ProducerEndpointImpl*> producers_;
   std::map<RelayClientID, RelayEndpointImpl*> relay_clients_;
   std::map<TracingSessionID, TracingSession> tracing_sessions_;
-  std::map<BufferID, std::unique_ptr<TraceBuffer>> buffers_;
+  std::map<BufferID, BufferAndProtoVms> buffers_;
```
rukkal: @primiano I merged buffer and ProtoVM instances for now. It seemed a natural thing to do, given that a ProtoVM is tied to a single buffer, they share the same lifetime and they must be cloned together. We could almost say that a ProtoVM belongs to a buffer.

primiano:
As discussed offline:

- Let's split a first PR just to integrate protovm in traced. Split the provenance part in a different PR. Split TP
- Makek ProtoVM an impl detail of TraceBufferV2
- Move the MaybeSetupProtoVM into TSImpl::SetupDataSource
- Add a new virtual method to TraceBuffer called GetMemoryUsageBytes(), which in TBV1 is == size, in TBv2 == size + SUM(protovms)

rukkal:
> As discussed offline:
> 
> * Let's split a first PR just to integrate protovm in traced. Split the provenance part in a different PR. Split TP
> * Makek ProtoVM an impl detail of TraceBufferV2
> * Move the MaybeSetupProtoVM into TSImpl::SetupDataSource
> * Add a new virtual method to TraceBuffer called GetMemoryUsageBytes(), which in TBV1 is == size, in TBv2 == size + SUM(protovms)

Done and the PR is ready for another review pass

protos/perfetto/trace/trace_packet.proto
```diff
@@ -113,8 +114,23 @@ package perfetto.protos;
 // See the [Buffers and Dataflow](/docs/concepts/buffers.md) doc for details.
 //
 // Next reserved id: 14 (up to 15).
-// Next id: 125.
+// Next id: 126.
 message TracePacket {
+  // Encapsulates the state and configuration of the ProtoVM instances running
```
primiano, line 119: can you move this below , so that the most frequent used fields are lef ton the top?

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -16,11 +16,18 @@
 
 #include "src/tracing/service/trace_buffer_v2.h"
 
+#include <algorithm>
+#include <memory>
+
+#include "perfetto/ext/base/murmur_hash.h"
+#include "src/protovm/vm.h"
+
```
primiano: remove this empty \n (and sort al lthe includes in the group)

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -916,14 +924,21 @@ void TraceBufferV2::DeleteNextChunksFor(size_t bytes_to_clear) {
...
-      // TODO(keanmariotti): if we have a ProtoVM here we should pass a non-null
-      // TracePacket. But if we don't we should pass nullptr, as AddSlice() on
-      // TracePacket has significant overhead (due to hitting the allocator)
-      TracePacket* trace_packet = nullptr;
-      if (!csr.ReadNextPacketInSeqOrder(trace_packet))
-        break;
+      // If we have a ProtoVM, passes a non-null TracePacket to read back the
+      // overwritten packet. But if we don't have a ProtoVM passes a nullptr to
+      // reduce overhead.
+      if (!protovms_.empty()) {
```
primiano, line 930: maybe this reads a bit nicer if you do;

```
// if we have protovm, we need to readout each packet and
// pass to protovm. if not, we still need to do reads, but
// without having to accumulated data in a packet.
TracePacket* maybe_packet = nullptr;
if (!protovms_.empty()) {
  overwritten_packet_.Clear();
  maybe_packet = &overwritten_packet_;
}
if (!csr.ReadNextPacketInSeqOrder(&maybe_packet)) {
    break;
} else if(maybe_packet) {
  MaybeProcessOverwrittenPacketWithProtoVm(maybe_packet, ...)
}
```

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1107,6 +1130,25 @@ TraceBufferV2::TraceBufferV2(CloneCtor, const TraceBufferV2& src)
 
   // Finally copy over the SequenceState map.
   sequences_ = src.sequences_;
+
+  for (const auto& vm : src.protovms_) {
+    auto vm_cloned = vm.CloneReadOnly();
+    if (!vm_cloned.instance) {
+      PERFETTO_ELOG("Failed to clone ProtoVM");
+      break;
```
primiano, line 1139: maybe protovms_.clear() as well ?

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1107,6 +1130,25 @@ TraceBufferV2::TraceBufferV2(CloneCtor, const TraceBufferV2& src)
...
+    }
+    protovms_.push_back(std::move(vm_cloned));
+  }
+}
+
+TraceBufferV2::Vm TraceBufferV2::Vm::CloneReadOnly() const {
+  Vm cloned_vm;
+  cloned_vm.data_source_name = data_source_name;
```
primiano, line 1147: shludn't the Vm copy-ctor eal with this automatically?

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1141,4 +1183,125 @@ void TraceBufferV2::DumpForTesting() {
...
 
+TraceBufferV2::Vm::Vm() = default;
+TraceBufferV2::Vm::~Vm() = default;
+TraceBufferV2::Vm::Vm(Vm&&) noexcept = default;
+
+void TraceBufferV2::MaybeSetUpProtoVm(const std::string& data_source_name,
+                                      const std::string& program_bytes,
+                                      uint32_t memory_limit_kb,
+                                      ProducerID producer_id) {
+  Vm* vm;
```
primiano: = nullptr

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1141,4 +1183,125 @@ void TraceBufferV2::DumpForTesting() {
...
+  // Re-use existing ProtoVM instance, if any.
+  uint64_t program_hash = base::MurmurHashValue(program_bytes);
+  auto vm_it =
+      std::find_if(protovms_.begin(), protovms_.end(), [&](const Vm& vm) {
+        return vm.data_source_name == data_source_name &&
+               vm.program_hash == program_hash;
+      });
+  if (vm_it != protovms_.end()) {
+    vm = std::addressof(*vm_it);
```
primiano: out of curiosity why not just &(*vm_it)?

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1141,4 +1183,125 @@ void TraceBufferV2::DumpForTesting() {
...
+  }
+  // Otherwise instantiate new ProtoVM
+  else {
```
primiano: nit can you make this
```
} else {
    // Otherwise instantiate new ProtoVM
```

as it's a bit confusing when looking at it

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1141,4 +1183,125 @@ void TraceBufferV2::DumpForTesting() {
...
+
+    protovms_.push_back(std::move(new_vm));
+    vm = &protovms_.back();
+  }
+  // Update ProtoVM's producer IDs, if needed.
+  bool is_producer_already_inserted =
+      std::find(vm->producers.cbegin(), vm->producers.cend(), producer_id) !=
+      vm->producers.cend();
+  if (!is_producer_already_inserted) {
+    vm->producers.push_back(producer_id);
```
primiano: just make producers a base::SmallSet so you can avoid the checks here

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1141,4 +1183,125 @@ void TraceBufferV2::DumpForTesting() {
...
+  }
+}
+
+namespace {
+void StitchPacket(const TracePacket& packet, std::vector<uint8_t>* stitched) {
```
primiano: I feel this should be just a method in TracePacket?
I think you can add it as a standard metod in TracePacket, and then make the existing GetRawBytesForTesting call the new method you introduce

or anyways if you want this in the anon namespace, move it on the very top (after includes, before the non-anon-namespace stuff)

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -1141,4 +1183,125 @@ void TraceBufferV2::DumpForTesting() {
...
+  */
+  for (auto& vm : protovms_) {
+    bool producer_matches =
+        std::find(vm.producers.cbegin(), vm.producers.cend(), producer) !=
+        vm.producers.cend();
+    if (!producer_matches) {
+      continue;
+    }
+    if (protovm_patch_.empty()) {
+      // TODO(primiano): can we expect packet.slices().size() == 1 quite often?
```
primiano: yeah I think you could have that codepath here, as in various cases we might just need one slice. but let's do in another cl

src/tracing/service/trace_buffer_v2.h
```diff
@@ -373,6 +378,20 @@ class TraceBufferV2 : public TraceBuffer {
...
+  // Represents a ProtoVM instance and some metadata
+  struct Vm {
+    Vm();
+    ~Vm();
+    Vm(Vm&&) noexcept;
+    Vm CloneReadOnly() const;
+
+    std::unique_ptr<protovm::Vm> instance;
+    std::string data_source_name;
+    uint64_t program_hash;
```
primiano: zero-initialize this (And below), so you don't get surprises if somebody forgets about it

src/tracing/service/trace_buffer_v2.h
```diff
@@ -428,6 +447,13 @@ class TraceBufferV2 : public TraceBuffer {
                              size_t patches_size,
                              bool other_patches_pending) override;
 
+  void MaybeSetUpProtoVm(const std::string& data_source_name,
```
primiano, line 451: nit: move these metods below, so all the ones that deals with readback are together

src/tracing/service/trace_buffer_v2.h
```diff
@@ -596,6 +624,22 @@ class TraceBufferV2 : public TraceBuffer {
   // bugs in the producers. This is for tests that feed malicious inputs and
   // hence mimic a buggy producer.
   bool suppress_client_dchecks_for_testing_ = false;
+
```
primiano, line 628: add a link to go/perfetto-protovm so people can get context

src/tracing/service/tracing_service_impl.cc
```diff
@@ -3088,6 +3093,51 @@ void TracingServiceImpl::FreeBuffers(TracingSessionID tsid,
...
 
+void TracingServiceImpl::MaybeSetUpProtoVm(
+    const DataSourceConfig& ds_config,
+    const RegisteredDataSource& ds_instance,
+    BufferID buffer_id) {
+  if (!ds_config.has_protovm_config() &&
+      !ds_instance.descriptor.has_protovm_program()) {
+    return;  // Data source has no ProtoVM
+  }
+  if (!ds_instance.descriptor.has_protovm_program()) {
```
primiano, line 3113: can you add a todo for some future CL. we should report these errors in the trace itself (in some protovm_error packet). so we can surface these in the UI.
otherwise nobody will ever see those

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4274,6 +4322,48 @@ void TracingServiceImpl::MaybeEmitRemoteClockSync(
...
 }
 
+void TracingServiceImpl::MaybeEmitProtoVmInstances(
+    TracingSession* tracing_session,
+    std::vector<TracePacket>* packets) {
+  if (tracing_session->did_emit_protovm_instances_)
+    return;
+
+  bool has_proto_vms = false;
+  protozero::HeapBuffered<protos::pbzero::TracePacket> packet;
```
primiano: I would do this logic slighgly differently:

- get rid of the has_proto_vms boolean
- wrap |packet| in std::optional<HeapBuffered...
- Have pbzero::Protovms* packet_vms = nullptr;

then you have a labmda of the form

auto create_packet_vms = [&] {
  if (!packet_vms) {
    packet.reset(HeapBuffered...);
  }
  return packet_vms;
};

then below you use the nullity of packet_vms to check in the
if (packet_vms)

protos/perfetto/trace/trace_packet.proto
rukkal: I could move it just a little bit below, before the `oneof data {`. Alas, I can't move it into the oneof.

src/tracing/service/trace_buffer_v2.cc
rukkal: Done

src/tracing/service/trace_buffer_v2.cc
rukkal: yeah better

src/tracing/service/trace_buffer_v2.cc
rukkal: My idea was to have a CloneReadOnly() function to be more explicit than a copy-ctor. Mostly because it is not a pure copy. We are not making a copy of the `std::unique_ptr<protovm::Vm> instance`, we are making a read-only clone.

src/tracing/service/trace_buffer_v2.cc
rukkal: Oh.. much nicer! So we also enforce that the linear search remains small. I don't like too much the current error handling though (just a PERFETTO_ELOG). Should this be an atom?

src/tracing/service/trace_buffer_v2.cc
rukkal: I made it a `TracePacket::GetRawBytes(std::string* out_data)` method

src/tracing/service/trace_buffer_v2.h
rukkal: This one is over here because it's public. Then we have MaybeProcessOverwrittenPacketWithProtoVm() farther below because it's private. They are close to each other in the .cc file though.

src/tracing/service/tracing_service_impl.cc
rukkal: Here my idea was that a developer would see the logs, while debugging ProtoVM program and config because the tracing is not working as intended.

TODO added anyway.

src/tracing/service/tracing_service_impl.cc
rukkal: Done

src/tracing/core/trace_packet.cc
```diff
@@ -64,15 +70,18 @@ std::tuple<char*, size_t> TracePacket::GetProtoPreamble() {
...
-  data.resize(size());
-  size_t pos = 0;
-  for (const Slice& slice : slices()) {
-    PERFETTO_CHECK(pos + slice.size <= data.size());
-    memcpy(&data[pos], slice.start, slice.size);
+void TracePacket::GetRawBytes(std::string* out_data) const {
+  out_data->resize(size());
+  uint64_t pos = 0;
+  for (const auto& slice : slices()) {
+    std::memcpy(out_data->data() + pos, slice.start, slice.size);
```
primiano: why std::memcpy, we just use memcpy everywhere else? not a big deal but feels a bit inconsistent.
also why removing the check,
if you are worried about perf you can make it faster by caching the size() on the first line and checking against that.

src/tracing/service/trace_buffer_v2.h
```diff
@@ -373,6 +379,20 @@ class TraceBufferV2 : public TraceBuffer {
...
+    Vm();
+    ~Vm();
+    Vm(Vm&&) noexcept;
+    Vm CloneReadOnly() const;
+
+    std::unique_ptr<protovm::Vm> instance;
+    std::string data_source_name;
+    uint64_t program_hash = 0;
+    uint32_t memory_limit_kb = 0;
+    SmallSet<ProducerID, 8> producers;
```
primiano: err sorry I meant FlatSet not smallSet. smallset is just a O(N) lookup in a vector, my bad

rukkal: Changed to FlatSet

src/tracing/service/trace_buffer_v2.cc
rukkal: Sg. Added a TODO for now.

src/tracing/service/trace_buffer_v2.cc
rukkal: Just out of habit when writing templated code that has to be robust to weird types that overload operator& 😅

I replaced it with `&(*vm_it)`, if that improves readability

src/tracing/core/trace_packet.cc
rukkal: > why std::memcpy, we just use memcpy everywhere else?

changed to memcpy

> why removing the check

It felt a bit redundant and expensive as TracePacket::size() is extensively unit tested. Anyway I added it back.

rukkal:
@primiano I pushed some final polishing and I'll start preparing a follow up PR to fix some TODOs.

Just double checking that you saw this one:
```
  // TODO(primiano): something we should discuss. Currently, through this config
  // we control the memory limit of a single ProtoVM instance. However, the
  // total memory usage is unbounded because producers are passing N programs,
  // where each could have a difference version and the service would
  // instantiate N ProtoVMs.
  //
  // I'd say clean solution would be specifying programs in the tracing config,
  // thus we control from there the total number of instances and memory usage.
  //
  // But afaiu that's not possible, so what about setting a hard limit for the
  // number of ProtoVM instances? I guess we'll never have more than two program
  // versions, i.e. more than two ProtoVM instances for a given data source.
```

protos/perfetto/config/data_source_config.proto
```diff
@@ -154,6 +155,25 @@ message DataSourceConfig {
 
   optional PriorityBoostConfig priority_boost = 10;
 
+  // If specified, the data source requires the tracing service to process
+  // overwritten packets (patches) using a ProtoVM instance with this config.
+  // The ProtoVM program is specified at data source registration time through
+  // the descriptor.
+
+  // TODO(primiano): something we should discuss. Currently, through this config
```
primiano: I would word the comment as this:

// TODO(primiano): today when we enable a protovm_config, we actually accept
// that if N producers advertise M different versions of the same program we
// will create M instances of that program.
// To overcome this, we could move the program to the TraceConfig. But doing so
// would bloat the trace config size too big and create problems to statsd.
// Once the config store (b/482305876) exists we should move protovm programs
// onto that.

rukkal: Done

PR #3817, opened 2025-11-24
ui: Fetch distinct values for equality

aMayzner:
## Summary

Adds support for filtering data grid columns by selecting from distinct values. The feature adds "Filter equal to..." and "Filter not equal to..." menu items that fetch and display up to 100 distinct values from the current dataset. The menu includes a search box for columns with many values and respects existing filters/sorting.

  ## Changes

  - Added optional `getDistinctValues()` method to `DataGridDataSource` interface
  - Implemented `getDistinctValues()` in both `SQLDataSource` and `InMemoryDataSource`
  - Created `DistinctValuesFilterMenu` component with search, loading states, and value caching
  - Extracted `compareSqlValues()` utility function for consistent sorting across datagrid codebase
  - Added caching layer to avoid re-fetching values for the same query state

### DIFF
--- a/ui/src/components/widgets/data_grid/data_grid.ts
+++ b/ui/src/components/widgets/data_grid/data_grid.ts
@@ -319,17 +319,20 @@ export interface DataGridApi {
 const MAX_DISTINCT_VALUES = 100;
 const MAX_DISPLAY_STRING_LENGTH = 50;
 const SEARCH_BOX_THRESHOLD = 10;
+const DISTINCT_VALUES_MENU_MAX_HEIGHT = 300;
+
+type DistinctValuesResult = SqlValue[] | 'too_many' | 'error';
+
+interface DistinctValuesMenuState {
+  values?: DistinctValuesResult;
+  loading: boolean;
+  searchText: string;
+}
 
 interface DistinctValuesFilterMenuAttrs {
   readonly column: ColumnDefinition;
   readonly dataSource: DataGridDataSource;
-  readonly filters: ReadonlyArray<DataGridFilter>;
-  readonly sorting: Sorting;
   readonly onFilterAdd: OnFilterAdd;
-  readonly getCachedValues: () => SqlValue[] | 'too_many' | 'error' | undefined;
-  readonly setCachedValues: (values: SqlValue[] | 'too_many' | 'error') => void;
-  readonly fetchValues: () => Promise<SqlValue[] | 'too_many' | 'error'>;
-  readonly formatValue: (value: SqlValue) => string;
   readonly operator: '=' | '!=';
   readonly label: string;
 }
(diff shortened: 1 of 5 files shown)


ui/src/components/widgets/data_grid/data_grid.ts
```diff
@@ -307,6 +315,174 @@ export interface DataGridApi {
...
+{
+  private distinctValues?: SqlValue[] | 'too_many' | 'error';
+  private loading = false;
+  private searchText = '';
+
+  private getDistinctValuesArray(): SqlValue[] | undefined {
+    return Array.isArray(this.distinctValues) ? this.distinctValues : undefined;
+  }
+
+  private async onPopupMount(attrs: DistinctValuesFilterMenuAttrs) {
```
stevegolton: Probably don't need to use onPopupMount for this - maybe just do it in oncreate/oninit? Also you probably shouldn't call m.redraw() in this as it's called from a lifecycle hook.

Generally you only need to use onpopupMount if you need the underlying popup's DOM element for some reason.

ui/src/components/widgets/data_grid/data_grid.ts
```diff
@@ -307,6 +315,174 @@ export interface DataGridApi {
   ): Promise<string>;
 }
 
+// Configuration constants
+const MAX_DISTINCT_VALUES = 100;
+const MAX_DISPLAY_STRING_LENGTH = 50;
+const SEARCH_BOX_THRESHOLD = 10;
+
+interface DistinctValuesFilterMenuAttrs {
```
stevegolton, line 332: This is subjective but this component feels a bit too interested in its parent state. Maybe it would be better off as a render() function inside the main DataGrid component so you're having to do a lot of prop drilling.

ui/src/components/widgets/data_grid/data_grid.ts
```diff
@@ -334,6 +510,8 @@ export class DataGrid implements m.ClassComponent<DataGridAttrs> {
   private currentDataSource?: DataGridDataSource;
   private currentColumns?: ReadonlyArray<ColumnDefinition>;
   private currentValueFormatter?: ValueFormatter;
+  // Cache for distinct values per column
+  private distinctValuesCache = new Map<string, CachedDistinctValues>();
```
stevegolton: This cache should live in the data source. DataGrid is supposed to be able to handle the datasource changing underneath it every frame, so it shouldn't cache anything at all. It is merely the presentation/coordination layer.

As a compelling example, imagine I switch datasource to a completely different one, the cache will still contain values from the previous datasource.

ui/src/components/widgets/data_grid/data_grid.ts
```diff
@@ -307,6 +315,174 @@ export interface DataGridApi {
...
+  readonly onFilterAdd: OnFilterAdd;
+  readonly getCachedValues: () => SqlValue[] | 'too_many' | 'error' | undefined;
+  readonly setCachedValues: (values: SqlValue[] | 'too_many' | 'error') => void;
+  readonly fetchValues: () => Promise<SqlValue[] | 'too_many' | 'error'>;
+  readonly formatValue: (value: SqlValue) => string;
+  readonly operator: '=' | '!=';
+  readonly label: string;
+}
+
+class DistinctValuesFilterMenu
```
stevegolton, line 340: This filter menu should be scrollable, possibly fixed height to avoid flickering when things load

PR #3821, opened 2025-11-24
rust-sdk: update amalgamated source instructions

dreveman:
Bump perfetto-sdk-sys crate version to publish crate that uses the new instructions for amalgamated source.


dreveman:
FYI, latest version also updates .github/workflows/rust-sdk-tests.yml to match the new instructions

PR #3829, opened 2025-11-24
Android: Enhance privacy for sensitive user types

annathedroid:
**Description**

This change implements a default configuration in Perfetto to enhance privacy for sensitive Android user types, such as Private Space.

The solution is to introduce an allow list for user type collection. Any user type not explicitly on this list will be categorized as `android.os.usertype.FILTERED` in the trace data. This ensures that package information associated with these sensitive user types is not trivially linkable to the profile.

**Tasks Implemented:**

1.  Check if user type is not listed on allowlist and replace user type with `android.os.usertype.FILTERED`
2. Add unit tests to check if categorisation by config filter works as expected works as expected.

**Test**

Verified the categorisation by using a trace configuration that only allowed `android.os.usertype.system.HEADLESS`. The trace output confirmed that other users were correctly categorized as `android.os.usertype.FILTERED`:

```
$ SELECT * FROM android_user_list

android_user_id                     type
----------------------------------- -----------------------------------
                                  0 android.os.usertype.system.HEADLESS
                                 10 android.os.usertype.FILTERED
                                 11 android.os.usertype.FILTERED

```

Bug: 462380293
Flag: EXEMPT BUGFIX
Test: tools/run_android_test --system-adb out/android perfetto_unittests
Change-Id: I4df2a058e67a2f50a17d69a9231b56536fa91bfe

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/graph/graph.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/graph/graph.ts
@@ -313,8 +313,6 @@ function ensureNodeLayouts(
           accentBar: true,
           content: m(NodeBox, {
             node: qnode,
-            onDuplicateNode: attrs.onDuplicateNode,
-            onDeleteNode: attrs.onDeleteNode,
             onAddOperationNode: attrs.onAddOperationNode,
           }),
           // Don't include 'next' here - we want placement for just this node
@@ -386,6 +384,22 @@ function getNextDockedNode(
   return undefined;
 }
 
+function buildNodeContextMenuItems(
+  qnode: QueryNode,
+  attrs: GraphAttrs,
+): m.Children {
+  return [
+    m(MenuItem, {
+      label: 'Duplicate',
+      onclick: () => attrs.onDuplicateNode(qnode),
+    }),
+    m(MenuItem, {
+      label: 'Delete',
+      onclick: () => attrs.onDeleteNode(qnode),
+    }),
+  ];
+}
+
 function createNodeConfig(
   qnode: QueryNode,
   attrs: GraphAttrs,
@@ -406,10 +420,9 @@ function createNodeConfig(
     canDockTop: !noTopPort,
     hue: getNodeHue(qnode),
     accentBar: true,
+    contextMenuItems: buildNodeContextMenuItems(qnode, attrs),
     content: m(NodeBox, {
       node: qnode,
-      onDuplicateNode: attrs.onDuplicateNode,
-      onDeleteNode: attrs.onDeleteNode,
       onAddOperationNode: attrs.onAddOperationNode,
     }),
     next: getNextDockedNode(qnode, attrs),
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/graph/node_box.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/graph/node_box.ts
@@ -15,9 +15,7 @@
 import m from 'mithril';
 
 import {classNames} from '../../../../base/classnames';
-import {Icons} from '../../../../base/semantic_icons';
-import {Button} from '../../../../widgets/button';
-import {MenuItem, PopupMenu} from '../../../../widgets/menu';
+import {PopupMenu} from '../../../../widgets/menu';
 import {QueryNode, singleNodeOperation, NodeType} from '../../query_node';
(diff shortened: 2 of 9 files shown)


src/traced/probes/user_list/user_list_data_source.cc
```diff
@@ -45,11 +46,34 @@ UserListDataSource::UserListDataSource(const DataSourceConfig& ds_config,
...
+
+  bool config_has_filters = false;
   for (auto type = cfg.user_type_filter(); type; ++type) {
     user_type_filter_.emplace((*type).ToStdString());
+    config_has_filters = true;
+  }
+
+  if (!config_has_filters) {
+    PERFETTO_ILOG("Applying default user type filters for 'user' build.");
+    const std::vector<std::string> default_filters = {
```
aMayzner: This list for sure shouldn't just live here (the enums and globals should be at the top of the file), but I am not sure whether it should exist anywhere. Maybe it would be better to just always expect filters to be provided in the proto? I don't like this idea of "there will be filtering, when you specify it or not". What I think might be a compromise is having a list of enum values in the proto, that will be used for filtering if the "filter default" or sth like that flag is set to true

annathedroid: Moved to protovalidators

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -879,8 +879,34 @@ message SurfaceFlingerTransactionsConfig {
...
 message AndroidUserListConfig {
-  // If not empty, emit info about only the following list of user types
-  // (exact match, no regex). Otherwise, emit info about all users.
+  // An allowlist of user type strings, used to control the granularity of
+  // user type information emitted in the trace. Exact, case-sensitive string
+  // matching is used.
+  //
+  // Any user type read from the device that is NOT present in the
+  // effective allowlist will have its type reported as
+  // "android.os.usertype.UNKNOWN".
```
LalitMaganti: Instead of unknown, I would expect to see "filtered" indicating it was filtered on device

src/traced/probes/user_list/user_list_data_source.cc
```diff
@@ -45,11 +46,14 @@ UserListDataSource::UserListDataSource(const DataSourceConfig& ds_config,
                                        std::unique_ptr<TraceWriter> writer)
     : ProbesDataSource(session_id, &descriptor), writer_(std::move(writer)) {
   AndroidUserListConfig::Decoder cfg(ds_config.user_list_config_raw());
+
   for (auto type = cfg.user_type_filter(); type; ++type) {
     user_type_filter_.emplace((*type).ToStdString());
   }
 }
 
+// ... rest of the file remains the same ...
```
LalitMaganti: Remove this comment?

protos/perfetto/trace/perfetto_trace.proto
annathedroid: So actually no change? Then we put null as a user type when joining with processes or heap stacks?

annathedroid: I just wonder, if we don't replace protected user types with unknown, then it could change perspective in performance analysis. We know each user decreases performance, and if we don't indicate the existence of those users it might be harder to analyse the traces.

protos/perfetto/config/perfetto_config.proto
```diff
@@ -879,8 +879,34 @@ message SurfaceFlingerTransactionsConfig {
...
+  //            user_list_config {
+  //              # Only report these specific types, others become UNKNOWN.
+  //              user_type_filter: "android.os.usertype.full.SYSTEM"
+  //              user_type_filter: "android.os.usertype.system.HEADLESS"
+  //            }
+  //        }
+  //    }
+  //
+  //
+  // Note: This field does not support regular expressions.
```
annathedroid, line 909:
```suggestion
```

protos/perfetto/trace/perfetto_trace.proto
LalitMaganti: I mean to say I would expect to see the string "android.os.usertype.FILTERED" rather than "android.os.usertype.UNKNOWN". Unknown means we didn't know (i.e. there was some parsing error) while filtered makes clear we *did* know we just didn't include it because of the filtering option specified in the config.

annathedroid: Yeah FILTERED works too, thanks!

LalitMaganti (approved):

PR #3831, opened 2025-11-24
ui: decouple pan + zoom from panning in "scrollTo"

LalitMaganti:
This CL separates out "focusing" (which means to pan + zoom) from
"panning" (which means just pan, no zoom)

The motivation for this change:
1) Before https://github.com/google/perfetto/pull/3761, focus used to work in a very janky way where pressing F once would cause pan to center the slice and pressing again would zoom in
2) After https://github.com/google/perfetto/pull/3761, F consistently pans and zooms into the slice which is selected. But it accidentally also changed the behaviour of a bunch of other places which used focus to *just* pan.
3) After this change, we change things so that *only* F pans and zooms and all other places just pan which is what we expect most of the time. We can also change other places if it becomes necessary to opt into this new behaviour.


primiano:
I didn't look at the CL but... how are we going to explain this to people? it's too many commands for too many similar cases?

The only thing that I can remotely think it could work it's something like "press F to pan only, press FF to pan and zoom".. but this requires more complex logic all over.

If we don't have a well thought solution I feel we are making it worse for everybody by trying to keep everybody happy

LalitMaganti:
> I didn't look at the CL but... how are we going to explain this to people? it's too many commands for too many similar cases?

You should really see the CL :) There are no new commands being introduced here, it's just a change in what behavior behind the scenes for already existing commands.

primiano:
It would be nice if I could tell from the commit description ;)

primiano:
I mean I see from the code that now you have an explicit 'pan' vs 'focus' control... I can't tell what happens in which case without looking deeply, and would be nice if the PR description told what was going on

altimin:
This patch LGTM.

Thinking about this a bit more and playing with the changes, I think that we should have a hotkey for "pan". The main use case I have in mind is pretty much whenever you select a slice without clicking on it (or look around after having selected a slice) — the selected slice often serves as a "anchor" point, where you want to go back to — and if you do, you would want to keep the zoom level the same.

Supporting it through either `Shift + F` or `P` should address these use cases.

LalitMaganti:
> I mean I see from the code that now you have an explicit 'pan' vs 'focus' control... I can't tell what happens in which case without looking deeply, and would be nice if the PR description told what was going on

Thanks for feedback. Done.

primiano:
Oh I see now it's more clear, the history of what happened was unclear to me

> focus used to work in a very janky way where pressing F once would cause pan to center the slice and pressing again would zoom in

you say "janky way" but my theory (I hope to be wrong) is that some people liked and might want it back.
No action needed unless people complain (But if they do, I'd understand them tbh)

PR #3839, opened 2025-11-25
perfetto: static_libs -> whole_static_libs for cc_static_library

LalitMaganti:
This is necessary to unblock the roll into Android. The root cause is
b/169779783 (i.e. soong does not propogate transitive deps of
static_library targets). The suggested workaround is whole_static_libs
which is what we've done for all targets consistently in this change.

Bug: 169779783

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -727,7 +727,7 @@ cc_library_shared {
             ],
         },
         host: {
-            static_libs: [
+            whole_static_libs: [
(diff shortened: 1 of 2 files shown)


sashwinbalaji (approved):
from the bug seems like a necessary temp fix, but curious how much bloat this will cause and how did we check no undefined behavior due to ODR ?

PR #3841, opened 2025-11-25
Added producer-side SMB scraping for remote producers

jahdiel-alvarez:
Producers using shmem emulation won't be SMB scraped by the tracing service since the buffer isn't shared. To handle this scenario such producers after receiving a flush request from the tracing service will self-scrape their buffers, will mark the chunks with uncommitted data as complete and will send the trace data via a commit data request back to the tracing service.

Also, the tracing service has been refactored such that it always emits a flush request for producers using shmem emulation and it no longer does SMB scraping on service-side for such producers (those using shmem emulation).

Lastly, the tracing service was fixed such that it now properly sets the use_shmem_emulation on the shmem_abi when setting up the shared memory for producers using shmem emulation.

Bug: 456746462
Bug: 462494442

### DIFF
--- a/src/tracing/core/shared_memory_arbiter_impl.cc
+++ b/src/tracing/core/shared_memory_arbiter_impl.cc
@@ -917,7 +917,7 @@ void SharedMemoryArbiterImpl::ScrapeEmulatedSharedMemoryBuffer(
 
   CommitDataRequest commit_req;
   ForEachScrapableChunk(&shmem_abi_, [&](SharedMemoryABI::Chunk* chunk,
-                                         bool chunk_complete, auto) {
+                                         bool chunk_complete, auto, auto) {
     const auto writer = buffer_for_writers.find(chunk->writer_id());
     if (writer == buffer_for_writers.end())
       return;
--- a/src/tracing/core/shared_memory_arbiter_impl.h
+++ b/src/tracing/core/shared_memory_arbiter_impl.h
@@ -164,12 +164,12 @@ class SharedMemoryArbiterImpl : public SharedMemoryArbiter {
 
   // F is lambda with signature:
   // void(SharedMemoryABI::Chunk*, bool chunk_complete,
-  //      std::pair<uint16_t, uint8_t> packetCountAndFlags)
+  //      uint16_t packet_count, uint8_t packet_flags)
   template <typename F>
   static inline void ForEachScrapableChunk(SharedMemoryABI* shmem_abi,
                                            F handle_chunk_function) {
     static_assert(std::is_invocable_v<F, SharedMemoryABI::Chunk*, bool,
-                                      std::pair<uint16_t, uint8_t>>);
+                                      uint16_t, uint8_t>);
     // num_pages() is immutable after the SMB is initialized and cannot be
     // changed even by a producer even if malicious.
     for (size_t page_idx = 0; page_idx < shmem_abi->num_pages(); page_idx++) {
@@ -198,13 +198,15 @@ class SharedMemoryArbiterImpl : public SharedMemoryArbiter {
         SharedMemoryABI::Chunk chunk =
             shmem_abi->GetChunkUnchecked(page_idx, header_bitmap, chunk_idx);
 
+        uint16_t packet_count;
+        uint8_t packet_flags;
         // GetPacketCountAndFlags has acquire_load semantics.
-        auto packet_count_and_flags = chunk.GetPacketCountAndFlags();
+        std::tie(packet_count, packet_flags) = chunk.GetPacketCountAndFlags();
 
         // It only makes sense to copy an incomplete chunk if there's at least
         // one full packet available. (The producer may not have completed the
         // last packet in it yet, so we need at least 2.)
-        if (!chunk_complete && packet_count_and_flags.first < 2)
+        if (!chunk_complete && packet_count < 2)
           continue;
 
         // At this point, it is safe to access the remaining header fields of
@@ -212,7 +214,8 @@ class SharedMemoryArbiterImpl : public SharedMemoryArbiter {
         // kChunkFree into kChunkBeingWritten state, the header should be
         // written completely once the packet count increased above 1 (it was
         // reset to 0 by the service when the chunk was freed).
-        handle_chunk_function(&chunk, chunk_complete, packet_count_and_flags);
+        handle_chunk_function(&chunk, chunk_complete, packet_count,
+                              packet_flags);
       }
     }
   }
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -2273,7 +2273,7 @@ void TracingServiceImpl::ScrapeSharedMemoryBuffers(
   SharedMemoryArbiterImpl::ForEachScrapableChunk(
       &producer->shmem_abi_,
       [&](SharedMemoryABI::Chunk* chunk, bool chunk_complete,
-          std::pair<uint16_t, uint8_t> packet_count_and_flags) {
+          uint16_t packet_count, uint8_t packet_flags) {
         WriterID writer_id = chunk->writer_id();
         std::optional<BufferID> target_buffer_id =
             producer->buffer_id_for_writer(writer_id);
@@ -2290,16 +2290,12 @@ void TracingServiceImpl::ScrapeSharedMemoryBuffers(
         if (!target_buffer_belongs_to_session)
           return;
 
-        uint16_t packet_count;
-        uint8_t flags;
-        std::tie(packet_count, flags) = packet_count_and_flags;
-
         uint32_t chunk_id =
             chunk->header()->chunk_id.load(std::memory_order_relaxed);
 
         CopyProducerPageIntoLogBuffer(
             producer->id_, producer->client_identity_, writer_id, chunk_id,
-            *target_buffer_id, packet_count, flags, chunk_complete,
+            *target_buffer_id, packet_count, packet_flags, chunk_complete,
             chunk->payload_begin(), chunk->payload_size());
       });
 }


include/perfetto/ext/tracing/core/shared_memory_abi.h
```diff
@@ -470,6 +470,7 @@ class SharedMemoryABI {
   size_t page_size() const { return page_size_; }
   size_t num_pages() const { return num_pages_; }
   bool is_valid() { return num_pages() > 0; }
+  bool is_shmem_emulated() { return use_shmem_emulation_; }
```
primiano: can you plz call it `use_shmem_emulation()` ? When doing codesearches is quite nice when the names of private variables and accessor match, as you can see in one go how things are plumbed rather than following renames.

src/tracing/core/shared_memory_arbiter_impl.cc
```diff
@@ -908,6 +908,64 @@ void SharedMemoryArbiterImpl::NotifyFlushComplete(FlushRequestID req_id) {
   }
 }
 
+void SharedMemoryArbiterImpl::ScrapeEmulatedSharedMemoryBuffer(
+    const std::map<WriterID, BufferID>& buffer_for_writers) {
+  if (!use_shmem_emulation_)
```
primiano: I think this should be a PERFETTO_CHECK(). we should never get here without shmem emulation, if we do it's  a bug and we should fail fast, not hide it

src/tracing/core/shared_memory_arbiter_impl.cc
```diff
@@ -908,6 +908,64 @@ void SharedMemoryArbiterImpl::NotifyFlushComplete(FlushRequestID req_id) {
...
+    return;
+
+  for (size_t page_idx = 0; page_idx < shmem_abi_.num_pages(); page_idx++) {
```
primiano: So most of this code is identical to the on in TracingServiceImpl::ScrapeSharedMemoryBuffers... but also cannot be 100% identical.

WDYT if we factored out an inline function in shared_memory_arbiter(_impl.h) along the lines of

```
// F is a lambda with signature void(WriterID, SharedMemoryABI::Chunk*)
template<typename F>
void ForEachScrapableChunk(F lambda) {
  ...
}
```

which essentially moves all the code up to chunk.writer_id(),

so then both TSImpl and your code can diverge there and do

ForEachScrapableChunk([&] (WriterID writer_id, Chunk* chunk)) {
   const auto writer = buffer_for_writers.find(chunk.writer_id());
      if (writer == buffer_for_writers.end())
        continue;

      BufferID target_buffer_id = writer->second;
      PatchList ignored;

      ReturnCompletedChunk(std::move(chunk), target_buffer_id, &ignored);
});

src/tracing/ipc/producer/producer_ipc_client_impl.cc
```diff
@@ -399,6 +399,28 @@ void ProducerIPCClientImpl::OnServiceRequest(
...
         static_cast<size_t>(cmd.flush().data_source_ids().size()), flags);
+
+    // NB: For producers using SMB emulation, the actual value of
+    // ProducerSMBScrapingMode::kDefault in the service-side is unknown on the
+    // producer-side. Therefore we only do producer-side SMB scraping if the
+    // scraping mode is explicitly enabled.
+    if (use_shmem_emulation_ &&
+        smb_scraping_mode_ ==
+            TracingService::ProducerSMBScrapingMode::kEnabled) {
+      // Producer-side SMB scraping should occur after the flush is complete.
```
primiano: Hmm I am a bit conflicted here. I went back and forth ~30 mins while editing this comment and reasoning about it. Please double-check below and don't blindly accept it. this is tricky


Here's my reasoning in summary, and then details:

- I see you are trying to stay consistent with the same order in which TracingServiceImpl does things

- But TracingServiceImpl has the benefit of being able to handle the sequence of async events linearly (because ScrapeSharedMemoryBuffers is synchronous within TSImpl::CompleteFlush)

- BUt you cannot guarantee the same linearity. SO there is a chance that your code might be racy, and by the time you commit the scraped data, the producer has already acknowledged the flush and the service might have moved on.

- As such I think you should make the scraping syncrhonously (i.e. NOT posted) but move it in ProducerIPCClientImpl::NotifyFlushComplete. Because that's the equivalent linearization that TSImpl does, on the client side.


WDYT?

(some further notes below just out of interest. You have no ideas I many times i rewrote this comment XD)

scraping has been designed to be idempotent. As in, because it happens under the hoods, you can first scrape, and then the producer can do the real commit, and TraceBuffer is designed to deal with it. i.e. it allows "recommit" and checks that you are only "growing" payload and flags. (*)

BUT THEN, I looked at the code in TracingServiceImpl, and realized that scraping has been implemented only AFTER NotifyFlushDoneForProducer. And here I realize you are trying to be consistent with it.

I think in practice both will work. If you commit, and then scrape, the data should be gone already (and even if you accidentally commit twice, the TraceBuffer should be able to handle that)

If you look at the code in trace_buffer_v1.cc that says
```
   // Check whether we have already copied the same chunk previously. This may
  // happen if the service scrapes chunks in a potentially incomplete state
  // before receiving commit requests for them from the producer. Note that the
  // service may scrape and thus override chunks in arbitrary order since the
  // chunks aren't ordered in the SMB.
  ```

File: src/tracing/ipc/producer/producer_ipc_client_impl.h [hunk 5]
```diff
@@ -124,6 +124,13 @@ class ProducerIPCClientImpl : public TracingService::ProducerEndpoint,
   // to |ipc_channel_| and (de)serializes method invocations over the wire.
   std::unique_ptr<protos::gen::ProducerPortProxy> producer_port_;
 
+  // Maps registered TraceWriter IDs to their target buffers as registered
+  // by the producer. When using shmem emulation, the producer internally
+  // needs the mapping of writer to buffer in order to do producer-side
+  // SMB scraping as it cannot rely on the map stored in the tracing
+  // service.
+  std::map<WriterID, BufferID> writers_;
```
primiano: call this writers_for_scraping_ so it's easier to reason about them?

src/tracing/service/tracing_service_impl.cc
```diff
@@ -1997,9 +2000,19 @@ void TracingServiceImpl::Flush(TracingSessionID tsid,
...
       continue;
-    data_source_instances[producer_id].push_back(ds_inst.instance_id);
+    }
+    ProducerEndpointImpl* producer = GetProducer(producer_id);
+    if (producer->smb_scraping_enabled_ && producer->IsShmemEmulated()) {
+      // Remote producers (connected via traced_relay) don't share buffer memory
+      // with the tracing service, therefore we NEED to trigger a flush request
+      // so that the producer can self-scrape their emulated SMB. Trigger the
+      // flush by inserting an empty vector if the producer not in the map
+      // already.
```
primiano: since you are here and you put the effort of writing a RFC, link it. 
 "See [RFC-0010](https://github.com/google/perfetto/blob/rfcs/0010-smb-scraping-remote-producers.md)"

src/tracing/service/tracing_service_impl.cc
```diff
@@ -1997,9 +2000,19 @@ void TracingServiceImpl::Flush(TracingSessionID tsid,
...
+      data_source_instances[producer_id];
```
primiano: this looks like a bug (I know it's not, I mean, if you just look at the code without thinking too much).
Do data_source_instances.emplace(producer_id)  instead as it looks more explicit and deliberate?

src/tracing/service/tracing_service_impl.cc
```diff
@@ -3307,9 +3320,16 @@ TracingServiceImpl::DataSourceInstance* TracingServiceImpl::SetupDataSource(
     // client to go away.
     PERFETTO_DLOG("Creating SMB of %zu KB for producer \"%s\"", shm_size / 1024,
                   producer->name_.c_str());
+    // TODO(jahdiel): We are creating shared memory buffers on the tracing
+    // service side even if the producer is using shmem emulation, which is
+    // wasting memory on the tracing service's machine.
```
primiano: TBH it might NOT be the case in practice.
If we page-fault only on reading (i.e. if the service never tries to change the smb) in reality it will cost 0 physical memory:
- THe SMB is allocated using mmap() (well base::SHaredMemory -> PagedMemory -> mmap) and not malloc
- If you pagefault read-only on MAP_ANON memory, on Linux it will map it to the zero page (a read-only physical page filled with zeros), which is one for the whole system
- Actual memory usage happens only if you try to write, when copy-on-write kicks in.

Now, I would have to audit the code but I believe that TSIMpl in the case of an unused SMB will only page-fault in read-only. It will try to read the page headers, read 0x00000000 (Which mwans "page not formatted") and move on.

So, all in all, I think we can live with this

cervantesyu: That's correct. It's mmap()ed but the tracing service doesn't read or write it. Chunk header and content are both serialized over the socket. The kernel won't back the (non-)SMB with physical memory.

src/tracing/core/shared_memory_arbiter_impl.cc
jahdiel-alvarez: Yes, this makes sense. I didn't deduplicate from the first go since I wanted to get the initial review first without being too disruptive in the code.

src/tracing/service/tracing_service_impl.cc
jahdiel-alvarez: Done, used `try_emplace`.

src/tracing/service/tracing_service_impl.cc
jahdiel-alvarez: Thanks for the in-depth explanation. I added a comment in the code instead to make this subtlety more explicit and avoid others from make the same incorrect assumption I did.

src/tracing/ipc/producer/producer_ipc_client_impl.cc
jahdiel-alvarez: @primiano thanks for taking the time to seriously analyze the code and look for the best way forward.

There are two points which I think are important and we should consider:

> If you commit, and then scrape, the data should be gone already (and even if you accidentally commit twice, the TraceBuffer should be able to handle that)

Committing and scraping aren't equal. When we flush pending commit data requests, we only [flush chunks which are marked completed](https://github.com/google/perfetto/blob/955549147a0b60cae1b10566921d1bea83888182/src/tracing/core/shared_memory_arbiter_impl.cc#L606-L635). This is why, while investigating the issue with my test CUJ, even if I forced a flush from the producer-side the issue still remained. The chunks weren't emitted because basically while the CUJ ran the trace data was never enough to fill the initial chunk and hence it was never marked completed. Scraping fixes this by explicitly marking the uncommitted chunks as completed while iterating through the SMB.

> As such I think you should make the scraping syncrhonously (i.e. NOT posted) but move it in ProducerIPCClientImpl::NotifyFlushComplete. Because that's the equivalent linearization that TSImpl does, on the client side.

Your analysis of the issue makes sense to me. My only concern is that `ProducerIPCClientImpl::NotifyFlushComplete` is called depending on the producer's implementation (we have many implementations) and so scraping would be dependent on the policy the producer decides to use when having to flush its internal data sources. For example if the producer never notifies the flush completed (maybe it was taking too long), then we would also skip scraping, whose main purpose is to address this exact scenario (scrape any commits that were missed during flushing). In the tracing service side, if we hit OnFlushTimeout we [log which data sources were slow](https://github.com/google/perfetto/blob/955549147a0b60cae1b10566921d1bea83888182/src/tracing/service/tracing_service_impl.cc#L2111-L2112) and we proceed to scrape them. Initializing the scraping here does have the benefit that it isn't directly attached to the producer's flush policy and if given enough time would scrape regardless of it. 

IMO this will always be racy because we are on the producer-side and we aren't synchronizing with the tracing service. This is why my initial proposal was to add a new RPC method to the protocol so that the tracing service could explicitly synchronize on the request to the producer and we would have more control of when remote producers scraped their SMB. Apart from that approach I think we still end up with two suboptimal scenarios: 1) Race between flush complete being sent and SMB scraping on producer-side or 2) NotifyFlushComplete never being called by producer and scraping never happening. In both the uncommitted chunks never reach the final trace buffer. If you think (2) is much less likely then yes moving the scraping to `ProducerIPCClientImpl::NotifyFlushComplete` would make more sense.

WDYT?

primiano: So I had some thoughts.

The problem with introducing a new IPC method is not the IPC method itself. is the fact that it requires a lot of extra complexity in tracing_service_impl.cc, precisely in functions that are already uber-complicated. Even if we retro-fitted scraping in the existing flush (for a while i was going to propose this), if you think about about what it takes, you need to add an extra async step to the flush (which actually needs to be in 2 places, both in FlushAndCloneSession and FlushAndDisableTracing) then you need another timeout, another failure mode... it goes up very quickly. If there was a simple (mainteinance-wise) way to add this to TracingServiceImpl I would consider it, but I'm not keen to add extra complexity to something already very complex and load-bearing.
For this reason I'd much rather do this as a more self-contained change in the client-side. (unless you can prove otherwise, but I spent the evening trying to think about it, and I can't come up with some self-contained change in TSImpl) 

I'm still on the NotifyFlushComplete camp, for two reasons:

### Case 1. the producer never acknolwedges NotifyFlushComplete.

- If this happens, this is a bug. We should fix it regardless of VMs as it delays flush for all tracing sessions.
- In practice I think everybody uses the SDK (aka goes via TracingMuxerImpl). To the best of my knowledge the only producers (perfetto clients) that implement the lowest level of the tracing protocol (i.e. Producer/Consumer interfaces, and hence have to deal with NotifyFlushComplete) are only two:
- 1) Perfetto's own damons (traced_probes etc). which always call NotifyFlushComplete
- 2) Chrome (Which uses its own IPC implementation)
- Nobody else should even depend on those include/perfetto/ext/ headers because [include/README.md](https://github.com/google/perfetto/blob/main/include/README.md) says clearly it's NOT an api. Also that's the most complex ever way to use perfetto.

### Case 2: the producer acknowledges too late NotifyFlushComplete.

This can legitimately happen in theory. But if it happens, we are doomed to lose data anyways (even if there was no no-flush data source). So does it matter what we lose more or less?
In practice, to the best of my knowledge, outside of perfetto's own processes, and some very special sysui data sources, very few project do anything big on flush. any other implementation will just invoke trace_writer.Flush(). and that does the same thing (actually simpler) that you would do in  the self-scraping code. So if that times out, you would time out as well.

For this reason, I think i'm convinced on NotifyFlushComplete. It is not perfect, but its failure mode is more predictable (fails in the cases where stndard flushes would have failed already) rather than being race-driven (which is more subtle and harder to debug). Instead we have logging  in the trace for flush timeouts, which makes the "flush timeout failure mode" easier to reason about.

jahdiel-alvarez: So actually even if we move the scraping to NotifyFlushComplete, we still need to post it if the call is made from ProducerIpcClientImpl. `ProducerIpcClientImpl::NotifyFlushComplete` simply calls `SharedMemoryArbiterImpl::NotifyFlushComplete` which post to the `task_runner_`. Hence, if we want to make sure scraping occurs after the flush, then we would either need to post the scraping or move the scraping into `SharedMemoryArbiterImpl::NotifyFlushComplete`.

WDYT?

primiano: > So actually even if we move the scraping to NotifyFlushComplete, we still need to post it if the call is made from ProducerIpcClientImpl. `ProducerIpcClientImpl::NotifyFlushComplete` simply calls `SharedMemoryArbiterImpl::NotifyFlushComplete` which post to the `task_runner_`. 

Let's assume you have a producer that handles flush (similary to what ProbesProducer::Flush does)
The following will happen:

1. The Producer gets the flush request.
2. It will itnernally broadcast the request to all its data sources
3. Every data source will (optionally) do some work to get their latest data (e.g. the Ftrace data source will read the latest contents). then they will call TraceWriter.Flush (this is the important bit that everybody does)
4. When they call TraceWriter.Flush(), that will write the outstanding data in the SMB (well in the VM case it will be a non-shared SMB, as it's emulated) and then enqueue the commit in the commit_data_req_
5. Then when All TraceWriter.Flush() have been invoked, the Producer calls NotifyFlushComplete

Now as you say, internally IPcClientImpl::NotifyFlushComplete invokes SharedMemoryArbiterImpl::NotifyFlushComplete which does a PostTask. But if you look at that task, it operates on the `commit_data_req_` (Which is where all the commits get staged, before the posted task sends the IPC)


At this point if you do the scraping, you will calling ScrapeEmulatedSharedMemoryBuffer which then will call ReturnCompletedChunk which then will call UpdateCommitDataRequest, which will append more stuff to `commit_data_req_`. Which is what you want actually


So the subtle part is this:
- Yes there is a postTask
- But the state of "what that PostTask will do" is not "in flight" (it's not bound to the post task arguments) but it's a local member variable (commit_data_req_)
- So you can "append" stuff to that commit_data_req_ even before that task is executed

Let's make an example with numbers

1. During the Flush a TraceWriter commits [Chunk1, Chunk2]
2. At this point commit_data_req_ contans [C1, C2] and the Task is posted
3. You are sitll running the current task
4. Now you do scraping, and you find Chunk3. You append it to commit_data_req_, whichi now contains [C1, C2, C3]
5. When the task runs, it will now commit in order C1, C2, C3, which is what you want

Essentially  only the act of sending the IPC is posted (Because in the general case we might not be ion the correct thread to send IPCs). But the data that will be committed is a mutex-protected queue, and that is the linearization point you care about

jahdiel-alvarez: Yes, this makes sense to me. Thanks for the discussion and review.

I have updated the PR to follow this approach.

src/tracing/core/shared_memory_arbiter_impl.h
```diff
@@ -162,6 +162,61 @@ class SharedMemoryArbiterImpl : public SharedMemoryArbiter {
     return default_page_layout;
   }
 
+  // F is lambda with signature:
+  // void(SharedMemoryABI::Chunk*, bool chunk_complete,
+  //      std::pair<uint16_t, uint8_t> packetCountAndFlags)
```
primiano: minor thing but you could pass here as uint16_t packet_count, uint8_t packet_flags, rather than passing a pair and re-destructuring here.
I'd just do it in the lambda.

src/tracing/core/shared_memory_arbiter_impl.cc
```diff
@@ -908,6 +908,23 @@ void SharedMemoryArbiterImpl::NotifyFlushComplete(FlushRequestID req_id) {
...
+  ForEachScrapableChunk(
+      &shmem_abi_, [&](SharedMemoryABI::Chunk* chunk, auto, auto) {
+        const auto writer = buffer_for_writers.find(chunk->writer_id());
+        if (writer == buffer_for_writers.end())
+          return;
+
+        BufferID target_buffer_id = writer->second;
+        PatchList ignored;
+
+        ReturnCompletedChunk(std::move(*chunk), target_buffer_id, &ignored);
```
primiano: Hmm why are we returning the Chunk as completed? This is what worries me here:

1. Somewhere in the same VM you have a TraceWriter which owns this chunk of SMB (well the TW thinks it's shared, it doesn't know it's not)
2. Now when you get a flush, you return the completed chunk.
3. This is fine from a service viewpoint, but from the client viewpoint this will trigger bugs, i believe, because the the TraceWriter it's still convinced it owns the chunk, and it will keep writing onto that (And eventually it will call ReturnCompletedChunk itself).

I think here you should only send over IPC a copy of the chunk to the service, without marking it as complete in the local SMB. the TW needs to be able to keep writing into that, unless i'm missing something

jahdiel-alvarez: Thanks for providing the patch to update the IPC such that the service is aware whether the chunk is complete or not.

PR #3844, opened 2025-11-25
exp: Widgets page

aMayzner:
## Summary

This PR introduces reusable widgets for the Explore Page query builder, consolidating repeated patterns across multiple node components into a shared widget library. The refactor improves code maintainability, ensures consistent UI styling, and adds validation for duplicate column names.

  **Key changes:**
- Updated `widgets.ts` with reusable components: `ColumnNameRow`, `CardWithHeader`, `ButtonGroup`, `Section`, `ListItem`, `ActionButtons`, `EmptyState`, `FormRow`, `LabeledControl`, and `DraggableItem`
- Added corresponding `widgets.scss` with consolidated styles using the `.pf-exp-*` prefix
- Refactored some query builder nodes to use the shared widgets (this change should be done gradually)
- Added duplicate column name validation in AddColumnsNode that dynamically disables the "Apply" button when duplicates exist
- Set AddColumnsNode to autoexecute
- Extended `ModalButton.disabled` to accept a function for dynamic evaluation on each render
- Added `onRemove` capability to `ListItem` for deleting computed columns

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.ts
@@ -202,24 +202,26 @@ export class ActionButtons implements m.ClassComponent<ActionButtonsAttrs> {
 }
 
 // Widget for a labeled form row with input
+// The children are placed inside the label for proper accessibility
 export interface FormRowAttrs {
   label: string;
 }
 
 export class FormRow implements m.ClassComponent<FormRowAttrs> {
   view({attrs, children}: m.CVnode<FormRowAttrs>) {
-    return m('.pf-exp-form-row', m('label', attrs.label), children);
+    return m('label.pf-exp-form-row', m('span', attrs.label), children);
   }
 }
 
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.ts
```diff
@@ -124,3 +120,140 @@ export class Section implements m.ClassComponent<SectionAttrs> {
...
+}
+
+// Widget for inline label with any control (input, select, multiselect, etc.)
+export interface LabeledControlAttrs {
+  label: string;
+}
+
+export class LabeledControl implements m.ClassComponent<LabeledControlAttrs> {
+  view({attrs, children}: m.CVnode<LabeledControlAttrs>) {
+    return m('.pf-exp-labeled-control', m('label', attrs.label), children);
```
stevegolton: Nit: You usually want to put the input (children, presumably in this case) inside the label element, then that label is bound to that form and the browser gives it special treatment.

Same as above.

See https://github.com/google/perfetto/blob/main/ui/src/widgets/form.ts#L133C14-L133C23

PR #3850, opened 2025-11-25
WIP: perfetto reboot traces

KirillTim:
Tested:
atest CtsPerfettoReporterRebootTestCases
atest CtsPerfettoReporterTestCases

### DIFF
--- a/src/perfetto_cmd/perfetto_cmd.cc
+++ b/src/perfetto_cmd/perfetto_cmd.cc
@@ -196,8 +196,9 @@ Statsd-specific and other Android-only flags:
   --config-uid          : UID of app which registered the config.
   --subscription-id     : ID of the subscription that triggered this trace.
   --upload              : Upload trace.
-  --upload-after-reboot : Used by Android to upload the traces saved to a 
-                          special directory after the system reboot.
+  --upload-after-reboot : Reports to Android framework all the traces marked
+                          as `persist_trace_after_reboot=true` which haven't
+                          been cleanly terminated.
   --dropbox        TAG  : DEPRECATED: Use --upload instead
                           TAG should always be set to 'perfetto'.
   --save-for-bugreport  : If a trace with bugreport_score > 0 is running, it


protos/perfetto/config/trace_config.proto
```diff
@@ -792,4 +792,15 @@ message TraceConfig {
   //
   // Introduced in: perfetto v54.
   optional bool trace_all_machines = 43;
+
+  // Android-only. Not for general use. If set, saves the trace into a special
+  // directory, to be uploaded by the android after the system reboot.
```
primiano: Just mention what the "special" directory is :)
No need to menion "uploaded" here as you say below about --upload.

Also you need to remind people to also set a unique_session_id as that becomes the name of the file

src/android_internal/tracing_service_proxy.cc
```diff
@@ -58,6 +62,14 @@ bool ReportTrace(const char* reporter_package_name,
   // possible.
   android::base::unique_fd fd(owned_trace_fd);
 
+  // 'ReportTrace' may be called before system boot is completed (e.g., when
```
primiano: Do we actually need this?

From b/461765543 looks like waitForService will wait for the service to be up, so you don' tneed to wait for boot (or any precondition) as the wait below will do the right thing anyways

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -590,8 +613,9 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
 
   bool parsed = false;
   bool cfg_could_be_txt = false;
-  const bool will_trace_or_trigger =
-      !is_attach() && !query_service_ && !clone_all_bugreport_traces_;
+  const bool will_trace_or_trigger = !is_attach() && !query_service_ &&
+                                     !clone_all_bugreport_traces_ &&
+                                     !upload_after_reboot_flag_;
```
primiano: why do you propagate this flag all the way down here? 
In my if you see this flag you should early out and do

"return UploadAfterReboot()"; early before doing any trace config parsing or anything else

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -119,6 +119,9 @@ void ArgsAppend(std::string* str, const std::string& arg) {
 }
 }  // namespace
 
+const char* kAndroidPersistentStateDir =
```
primiano, line 122: this should go in perfetto_cmd_android.cc (together with the function that handles it)

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -768,6 +792,38 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
...
+      PERFETTO_ELOG(
+          "TraceConfig's 'unique_session_name' must be set when using "
+          "'persist_trace_after_reboot'.");
+      return 1;
+    }
+    // PerfettoCmd assumes 'trace_out_path_' is the active output file owned by
+    // this process. When perfetto_cmd stops it automatically removes the file
+    // if '!trace_out_path_.empty()' and 'bytes_written_ == 0'.
+    // We use a separate field for the persistent trace file to prevent it from
+    // being automatically removed when perfetto_cmd stops.
```
primiano: But in the case of a graceful stop, don't you want it to be auotmatically removed?
I'm not sure I understand why we need this special case. can you expand?

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -925,6 +981,22 @@ int PerfettoCmd::ConnectToServiceRunAndMaybeNotify() {
 }
 
 int PerfettoCmd::ConnectToServiceAndRun() {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
```
primiano: Why this stuff is done in ConnectToServiceAndRun?

You don't need to conenct to any service. All this code should be moved to a different function in perfetto_cmd_android.cc, and you should diverge very soon.
You are reusing perfetto_cmd for selinux reasons, but effectively this is as-if it was its own standalone main(). You should not interleave it with the real perfetto_cmd

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -925,6 +981,22 @@ int PerfettoCmd::ConnectToServiceRunAndMaybeNotify() {
...
+  if (upload_after_reboot_flag_) {
+    std::string uploader_ready_prop =
+        base::GetAndroidProp("perfetto.persistent.uploader.status");
```
primiano: I'm a bit lukewarm on doing all these pedantic checks? What I worry about is that I am not sure you can read init.svc selinux-wise

I'd just keep it easy: if you ask for --upload-after-reboot, just do it.

The uploader status shouldn't matter as WaitForService will wait for it ... forever

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1238,7 +1310,20 @@ void PerfettoCmd::FinalizeTraceAndExit() {
 #endif
   } else if (report_to_android_framework_) {
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
-    ReportTraceToAndroidFrameworkOrCrash();
+    // TODO(ktimofeev): can we be here during normal android reboot (we received
```
primiano: Do we still have trace_out_stream_ in the persist casE?
if so I think the onl think we need to add is a unlik, below line 1305.
Something like

if (trace_out_stream) { 
  ... // Thiss will do the bytes_written computation
}
if (persist_after_reboot) {
 unlink(...)
}

rest as-is

src/android_internal/tracing_service_proxy.cc
KirillTim: Yes, we need this. Otherwise the logcat will be spammed by the 
```
12-01 22:54:33.096     0     0 I servicemanager: Caller(pid=808,uid=9999,sid=u:r:perfetto:s0) Tried to start aidl service tracing.proxy as a lazy service, but was unable to. Usually this happens when a service is not installed, but if the service is intended to be used as a lazy service, then it may be configured incorrectly.
12-01 22:54:34.096     0     0 I servicemanager: Caller(pid=808,uid=9999,sid=u:r:perfetto:s0) Since 'tracing.proxy' could not be found trying to start it as a lazy AIDL service. (if it's not configured to be a lazy service, it may be stuck starting or still starting).
```
entries.

src/perfetto_cmd/perfetto_cmd.cc
KirillTim: `PerfettoCmd::ParseCmdlineAndMaybeDaemonize` uses this constant to create a trace file name (line 813):

```
persistent_trace_out_path_ = std::string(kAndroidPersistentStateDir) + "/" +
                                 trace_config_->unique_session_name() +
                                 ".pftrace";
```

not sure if we want to extract this single line to a separate function

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -768,6 +792,38 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
...
+    }
+    // PerfettoCmd assumes 'trace_out_path_' is the active output file owned by
+    // this process. When perfetto_cmd stops it automatically removes the file
+    // if '!trace_out_path_.empty()' and 'bytes_written_ == 0'.
+    // We use a separate field for the persistent trace file to prevent it from
+    // being automatically removed when perfetto_cmd stops.
+    persistent_trace_out_path_ = std::string(kAndroidPersistentStateDir) + "/" +
+                                 trace_config_->unique_session_name() +
+                                 ".pftrace";
+    PERFETTO_CHECK(trace_config_->output_path().empty());
```
primiano: THis should be an ELOG+return like the ones above

src/android_internal/tracing_service_proxy.cc
KirillTim: I've moved the wait loop to the `PerfettoCmd::ReportAllPersistentTracesToAndroidFramework`.

protos/perfetto/config/trace_config.proto
KirillTim: > No need to menion "uploaded" here as you say below about --upload.

I'd rather keep it, to make it clear that the trace file will be uploaded after the system reboot, and not post-processed, or dropped, on handled any other way

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -186,19 +189,21 @@ Light configuration flags: (only when NOT using -c/--config)
...
-                         (testing only).
-  --reset-guardrails   : Resets the state of the guardails and exits
-                         (testing only).
+  --alert-id            : ID of the alert that triggered this trace.
+  --config-id           : ID of the triggering config.
+  --config-uid          : UID of app which registered the config.
+  --subscription-id     : ID of the subscription that triggered this trace.
+  --upload              : Upload trace.
+  --upload-after-reboot : Used by Android to upload the traces saved to a 
+                          special directory after the system reboot.
```
primiano: can you say: [Android-only] Reports to the Android framework all the traces marked as persist_trace_after_reboot=true which haven't been cleanly terminated.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -412,6 +419,18 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
       return 1;
 #endif
     }
+    if (option == OPT_UPLOAD_AFTER_REBOOT) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
```
primiano, line 431: small thing. can you just set a boolean here, and move the logic below after the arg parsing?

I find a bit easier to reason about if the code here in the top does pure parsing and the logic happens below

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -738,17 +757,8 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
...
-      break;
-    case TraceConfig::STATSD_LOGGING_DISABLED:
-      statsd_logging_ = false;
-      break;
-    case TraceConfig::STATSD_LOGGING_UNSPECIFIED:
-      statsd_logging_ = upload_flag_;
-      break;
-  }
+  statsd_logging_ = ShouldLogStatsdEvents(
+      *trace_config_, /*unspecified_filed_value=*/upload_flag_);
```
primiano: why this is called unspecified_field_value? call it upload_flag (also in the arg name)

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1751,6 +1791,20 @@ void PerfettoCmd::CloneAllBugreportTraces(
   }
 }
 
+// static
+bool PerfettoCmd::ShouldLogStatsdEvents(const TraceConfig& cfg,
+                                        bool unspecified_filed_value) {
```
primiano: s/unspecified_filed_value/upload_flag/ ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -14,9 +14,11 @@
  * limitations under the License.
  */
 
+#include "perfetto/ext/base/android_utils.h"
```
primiano: nit: this should go below together with others

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
   return fd;
 }
 
+// static
+std::vector<base::ScopedFile>
+PerfettoCmd::UnlinkAndReturnPersistentTracesToUpload() {
+  base::ScopedDir dir = base::ScopedDir(opendir(kAndroidPersistentStateDir));
+  if (!dir) {
+    PERFETTO_PLOG("Failed to open a persistent traces directory '%s'",
```
primiano: Just "Failed to open %s" is enough.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+
+  std::vector<std::string> trace_paths;
+  struct dirent* entry;
+  while ((entry = readdir(*dir)) != nullptr) {
+    if (entry->d_type != DT_REG)
+      continue;
+
+    std::string path =
+        std::string(kAndroidPersistentStateDir) + "/" + entry->d_name;
+    trace_paths.push_back(path);
```
primiano: .emplace_back(std::move(path))

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  }
+
+  std::vector<base::ScopedFile> trace_fds;
+  for (std::string& path : trace_paths) {
+    base::ScopedFile fd = base::OpenFile(path, O_RDONLY);
+    if (!fd) {
+      PERFETTO_PLOG("Failed to open a persistent trace file %s", path.c_str());
```
primiano: ditto, just "Failed to open %s"

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+      continue;
+    }
+    trace_fds.emplace_back(std::move(fd));
+    if (unlink(path.c_str()) != 0) {
+      PERFETTO_PLOG("Failed to unlink a persistent trace file %s",
```
primiano: ditto: "failed to unlink %s"

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  // if this function crashes, hangs, or is never called, the property will
+  // automatically be set to "1" once "sys.boot_completed=1".
+  std::vector<base::ScopedFile> fds = UnlinkAndReturnPersistentTracesToUpload();
+
+  if (__system_property_set("perfetto.on_reboot.uploader", "1") != 0) {
+    // This should never happen, but if it does we are in trouble. In this case
+    // just crash, we don't care about traces to be reported.
+    PERFETTO_FATAL("Failed to set property 'perfetto.on_reboot.uploader'");
+  }
+
```
primiano, line 298: if (fds.empty()) 
  return true;

we can avoid all the compelx logic below if there are no traces to upload, right?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  // Now we wait until boot completes to make sure Android reporter service is
+  // available.
+  // '__system_property_wait' is only available since API level 26, so we have a
+  // simple loop here.
+  bool boot_completed = false;
```
primiano, line 307: rather than checking for an indirect effect via the boot completed, can we check if the service is actually there via checkService() (see frameworks/native/libs/binder/include/binder/IServiceManager.h)?



Let's keep it for a separate CL though to avoid too many loops on this cl.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -412,6 +419,18 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
       return 1;
 #endif
     }
+    if (option == OPT_UPLOAD_AFTER_REBOOT) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+      // The --upload-after-reboot flag is only used by the 'perfetto.rc' init
+      // script during system boot. This is a oneshot standalone operation, that
+      // doesn't consume any other flags.
+      bool all_reports_succeed = ReportAllPersistentTracesToAndroidFramework();
```
primiano, line 435: also given that we are outside of the TaskRunner, we have no watchdog protection.
so within here i'd do a 

#include "perfetto/ext/base/watchdog.h"
...

bool all_reports_succeeded = false;
RunTaskWithWatchdogGuard([&] {
  all_reports_succeeded = ReportAllPersistentTracesToAndroidFramework();
})
CHECK(return all_reports_succeed)


Rathan than returning 1, if something goes wrong we want to crash. a crash will trigger a signal that fill file bugs to us (And the logs will be in the crash)

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
   return fd;
 }
 
+// static
+std::vector<base::ScopedFile>
+PerfettoCmd::UnlinkAndReturnPersistentTracesToUpload() {
+  base::ScopedDir dir = base::ScopedDir(opendir(kAndroidPersistentStateDir));
```
primiano: can you s/dir/trace_dir/? i have the feeling "dir" is going to break on windows, because windows.h has the very bad habit of defining and exporting symbols with these common names.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+
+  std::vector<base::ScopedFile> trace_fds;
+  for (std::string& path : trace_paths) {
+    base::ScopedFile fd = base::OpenFile(path, O_RDONLY);
+    if (!fd) {
+      PERFETTO_PLOG("Failed to open a persistent trace file %s", path.c_str());
+      continue;
+    }
+    trace_fds.emplace_back(std::move(fd));
+    if (unlink(path.c_str()) != 0) {
```
primiano, line 274: s/unlink/remove/: they are the same thing, just when doing codesearch, given we used remove elsewhere, it's nice for consistency if i want to search for "where are al lthe places where we remove files?"

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  // We must do as little work as possible before setting
+  // "perfetto.on_reboot.uploader". The "traced" service will not start
+  // until this property is set to "1".
+  //
+  // A fallback mechanism exists in perfetto.rc to prevent indefinite wait:
+  // if this function crashes, hangs, or is never called, the property will
+  // automatically be set to "1" once "sys.boot_completed=1".
+  std::vector<base::ScopedFile> fds = UnlinkAndReturnPersistentTracesToUpload();
+
+  if (__system_property_set("perfetto.on_reboot.uploader", "1") != 0) {
```
primiano, line 293: let's remove/ comment out this bit for now (i'll follow up on a doc)

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+                  " seconds, persistent traces will not be "
+                  "uploaded.",
+                  kWaitForSysBootCompletedSeconds);
+    return false;
+  }
+
+  bool all_reports_succeed = true;
+  for (base::ScopedFile& fd : fds) {
+    std::optional<uint64_t> maybe_file_size = base::GetFileSize(*fd);
+    if (!maybe_file_size.has_value()) {
```
primiano, line 326: if (!maybe_file_size.has_value() || *maybe_file_size <= 0) {
  PERFETTO_PLOG("Failed to stat persistent trace");
  continue;
}
uint64_t file_size = maybe_file_size.value();

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+      continue;
+    }
+    uint64_t file_size = maybe_file_size.value();
+    if (maybe_file_size == 0) {
```
primiano: nit this should be file_size, i think the above is a cast of the bool operator actually (subtle bug i know). but if yuou follow my suggestion above this becomes more compact anyways.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    {
+      // mmap-ed file closes upon scope exit.
+      base::ScopedFile mmap_fd(dup(*fd));
+      if (!mmap_fd) {
+        continue;
+      }
+      base::ScopedMmap mmaped_file =
+          base::ScopedMmap::FromHandle(std::move(mmap_fd), file_size);
+      if (!mmaped_file.IsValid()) {
+        continue;
```
primiano, line 346: PLOG failed to mmap

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+      }
+      trace_config = ParseTraceConfigFromMmapedTrace(std::move(mmaped_file));
+    }
+
+    if (!trace_config.has_value()) {
+      continue;
```
primiano, line 353: ELOG here (it shoul really never happen)

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,6 +242,133 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  }
+  if (!boot_completed) {
+    PERFETTO_ELOG("System did not boot within %" PRId64
+                  " seconds, persistent traces will not be "
+                  "uploaded.",
+                  kWaitForSysBootCompletedSeconds);
+    return false;
+  }
+
+  bool all_reports_succeed = true;
```
primiano: i think rather than this, it's more error-proof to:
- increment the counter of suceeded reports if you reach the end
- then you return num_succeeded == fds_count (cache it at the beginning)

because here you are not accounting I think for weird cases like "the trace config could not be parsed" and so on

primiano (commented):
as per offline discussion, let's keep this CL purely on the perfetto_cmd (and proto) changes.

let's split out the test and the .rc file, as I think i have a better proposal (Writing a doc now)

src/perfetto_cmd/perfetto_cmd_android.cc
KirillTim: If we want to report an error on `fatat` failure, we need to have two separate checks:
1. `if (maybe_file_size.has_value()) { report error; continue; }`
2. `if (file_size == 0) { continue; }`

since we don't want to report an error for an empty file.
So I've fixed the `if (maybe_file_size == 0)` code and add an error reporting.

Does it look good to you? 

Or should I do
```
if (!maybe_file_size.has_value() || *maybe_file_size <= 0) {
   // don't report error, empty file is not an error
   continue;
}
```
instead?

src/perfetto_cmd/perfetto_cmd_android.cc
KirillTim: I'd rather keep it as `unlink` for the following reasons:
1. consistency with other places in `perfetto_cmd_android.cc`
2. consistency with a method name, that is called `UnlinkAndReturnPersistentTracesToUpload`
3. for me, it seems like there is a tiny semantical (in a way we name the things) difference between "remove" and "unlink": when we "delete" file we "remove" it, but here we don't "delete" it, but "hide".

what do you think?

src/perfetto_cmd/perfetto_cmd.cc
KirillTim: This group of flags is already marked as `Statsd-specific and other Android-only flags:`, so I'd rather not to add the `[Android-only] ` prefix.

I've updated the description to be `Reports to the Android framework all the traces marked as persist_trace_after_reboot=true which haven't been cleanly terminated.`

PR #3857, opened 2025-11-26
ui: Add duplicate column name validation to aggregation node

aMayzner:
Add form-level validation to prevent duplicate column names in the Explore Page query builder's aggregation node. This catches conflicts between:
- Aggregation result names and GROUP BY columns
- Aggregation result names and other aggregation results

Also adds a `validation` prop to the Form widget to support custom
validation functions that can disable the submit button.

### DIFF
--- a/ui/src/widgets/form.ts
+++ b/ui/src/widgets/form.ts
@@ -43,6 +43,11 @@ export interface FormAttrs extends HTMLAttrs {
 
   // Prevent default form action on submit. Defaults to true.
   preventDefault?: boolean;
+
+  // Custom validation function. If provided, it will be called in addition to
+  // the native HTML form validation. The submit button will be disabled if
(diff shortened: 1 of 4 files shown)


ui/src/widgets/form.ts
```diff
@@ -59,9 +66,13 @@ export class Form implements m.ClassComponent<FormAttrs> {
       resetLabel,
       onSubmit = () => {},
       preventDefault = true,
+      validation,
       ...htmlAttrs
     } = attrs;
 
+    // Store validation function for use in lifecycle hooks
+    this.validation = validation;
```
stevegolton: Please don't cache this in the private member variables. oncreate and onupdate get passed the attrs as well so you can just use attrs.validation from there and pass it to the maybeDisableSubmitButton.

PR #3879, opened 2025-11-27
Add android user type to Android Package Metadata

annathedroid:
This change will allow to understand performance of process per android user type.
Android Package Metadata already contain user id which makes it per user unique and easy to add user type.

Test performed:
```
> INCLUDE PERFETTO MODULE android.process_metadata
> SELECT upid, pid, process_name, uid, user_id, user_type  FROM android_process_metadata LIMIT 10
```

 | upid | pid |              process_name           |              uid                       |          user_id              |               user_type             |                      
 |--------|------|------------------------------------------|----------------------------------|------------------------------|-------------------------------------|
 | 0   | 0        |[NULL]                                        | [NULL]                             | [NULL]                        | [NULL]                                | 
 | 1   | 3969 |  [NULL]                                       | [NULL]                            | [NULL]                        | [NULL]                                | 
 | 2   | 3937 |  /system/bin/heapprofd         | 9999                                | 0       | android.os.usertype.system.HEADLESS  | 
 | 3   | 1058 |  /system/bin/traced_probes  | 9999                                | 0        | android.os.usertype.system.HEADLESS  | 
 | 4   | 3938 |  /system/bin/traced_perf       | 9999                                | 0        | android.os.usertype.system.HEADLESS  | 
 | 5   | 2794 |  kworker/u16:0                        | 0                                       | 0        | android.os.usertype.system.HEADLESS  | 
 | 6   | 1        |  /system/bin/init                     | 0                                        | 0        | android.os.usertype.system.HEADLESS  | 
 | 7   | 2        |  kthreadd                                 | 0                                        | 0        | android.os.usertype.system.HEADLESS  | 
 | 8   | 3        |  rcu_gp                                      | 0                                       | 0        | android.os.usertype.system.HEADLESS  | 
 | 9   | 4        |  slub_flushwq                           | 0                                       | 0        | android.os.usertype.system.HEADLESS  | 

Bug: 463959671
Flag: EXEMPT BUGFIX
Test: tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell
Change-Id: I59d24d5892b263385203609584a31e03d91ebe40

### DIFF
--- a/ui/src/components/aggregation_adapter.ts
+++ b/ui/src/components/aggregation_adapter.ts
@@ -23,7 +23,7 @@ import {
 } from '../public/selection';
 import {Trace} from '../public/trace';
 import {Track} from '../public/track';
-import {Dataset, DatasetSchema, UnionDataset} from '../trace_processor/dataset';
+import {UnionDataset, Dataset, DatasetSchema} from '../trace_processor/dataset';
 import {Engine} from '../trace_processor/engine';
 import {EmptyState} from '../widgets/empty_state';
 import {Spinner} from '../widgets/spinner';
@@ -130,7 +130,7 @@ export function selectTracksAndGetDataset<T extends DatasetSchema>(
   tracks: ReadonlyArray<Track>,
   spec: T,
   kind?: string,
-): Dataset<T> | undefined {
+) {
   const datasets = tracks
     .filter((t) => kind === undefined || t.tags?.kinds?.includes(kind))
     .map((t) => t.renderer.getDataset?.())
@@ -138,8 +138,7 @@ export function selectTracksAndGetDataset<T extends DatasetSchema>(
     .filter((d) => d.implements(spec));
 
   if (datasets.length > 0) {
-    // TODO(stevegolton): Avoid typecast in UnionDataset.
-    return (new UnionDataset(datasets) as unknown as Dataset<T>).optimize();
+    return UnionDataset.create(datasets).optimize();
   } else {
     return undefined;
   }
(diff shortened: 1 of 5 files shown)


protos/perfetto/metrics/android/process_metadata.proto
```diff
@@ -27,6 +27,10 @@ message AndroidProcessMetadata {
   // Android user-id running this process (applicable to multi-user android)
   optional int64 android_user_id = 10;
 
+  // Type of Android user running this process (applicable to multi-user
+  // android)
+  optional int64 android_user_type = 12;
```
LalitMaganti: Wouldn't it better if this was a string instead of an int? Or at least document what the ints values are possible (i.e. link to the enim)

src/trace_processor/metrics/sql/android/process_metadata.sql
```diff
@@ -29,6 +29,7 @@ SELECT NULL_IF_EMPTY(AndroidProcessMetadata(
     'name', process_name,
     'uid', uid,
     'android_user_id', user_id,
+    'android_user_type', user_type,
```
LalitMaganti, line 32: Wait what is this a string and above an int!

protos/perfetto/metrics/android/process_metadata.proto
ilkos: Hmm this is an int64 but afaict you're assigning it a string; does this work (I'm also surprised there are no test failures, but possibly there should be a test for this with the user type)?

annathedroid: Yes good catch, I will add some tests too,

annathedroid: Yep yep on it

src/trace_processor/metrics/sql/android/process_metadata.sql
annathedroid: changed to string everywhere

PR #3880, opened 2025-11-27
LoadLine plugin

khokhlov962:
The UI plugin is intended for traces from the LoadLine 2 benchmark. It shows a track with the metric overview and two subtracks with breakdowns of each metric. For all other traces, it does nothing.

### DIFF
--- a/ui/src/plugins/org.chromium.LoadLine2/index.ts
+++ b/ui/src/plugins/org.chromium.LoadLine2/index.ts
@@ -132,30 +132,31 @@ export default class implements PerfettoPlugin {
     'Show breakdown of LoadLine 2 metrics on a separate track.';
 
   async onTraceLoad(trace: Trace): Promise<void> {
-    if (await isLoadLineTrace(trace)) {
-      prepareTables(trace);
-      registerTrack(trace, METRIC_TRACK_URI, '_chrome_loadline2_plugin_metric');
-      registerTrack(
-        trace,
-        VISUAL_TRACK_URI,
-        '_chrome_loadline2_plugin_visual_breakdown',
-      );
-      registerTrack(
-        trace,
-        INTERACTIVE_TRACK_URI,
-        '_chrome_loadline2_plugin_interactive_breakdown',
-      );
-      addTracks(trace);
+    if (!(await isLoadLineTrace(trace))) {
+      return;
     }
+    prepareTables(trace);
+    registerTrack(trace, METRIC_TRACK_URI, '_chrome_loadline2_plugin_metric');
+    registerTrack(
+      trace,
+      VISUAL_TRACK_URI,
+      '_chrome_loadline2_plugin_visual_breakdown',
+    );
+    registerTrack(
+      trace,
+      INTERACTIVE_TRACK_URI,
+      '_chrome_loadline2_plugin_interactive_breakdown',
+    );
+    addTracks(trace);
   }
 }
 
 async function isLoadLineTrace(trace: Trace): Promise<boolean> {
   const queryRes = await trace.engine.query(`
-      SELECT COUNT(*) AS cnt FROM slice WHERE name GLOB 'LoadLine2/*';
+      SELECT EXISTS(SELECT * FROM slice WHERE name GLOB 'LoadLine2/*') AS res;
     `);
-  const it = queryRes.iter({cnt: NUM});
-  return it.cnt > 0;
+  const it = queryRes.iter({res: NUM});
+  return it.res > 0;
 }
 
 async function prepareTables(trace: Trace): Promise<void> {
@@ -164,11 +165,7 @@ async function prepareTables(trace: Trace): Promise<void> {
   await trace.engine.query(generateBreakdownQuery('interactive'));
 }
 
-async function registerTrack(
-  trace: Trace,
-  uri: string,
-  src: string,
-): Promise<void> {
+function registerTrack(trace: Trace, uri: string, src: string): void {
   trace.tracks.registerTrack({
     uri,
     renderer: SliceTrack.create({
@@ -186,7 +183,7 @@ async function registerTrack(
   });
 }
 
-async function addTracks(trace: Trace): Promise<void> {
+function addTracks(trace: Trace): void {
   const metricTrack = new TrackNode({
     uri: METRIC_TRACK_URI,
     name: 'LoadLine 2 metrics',


ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const METRIC_QUERY = `
+INCLUDE PERFETTO MODULE chrome.loadline_2;
+
+DROP TABLE IF EXISTS _chrome_loadline2_plugin_metric;
+CREATE PERFETTO TABLE _chrome_loadline2_plugin_metric AS
+WITH all_generated_slices AS (
+  SELECT
+    page,
+    'visual' AS name,
```
victorvianna, line 30: curious, what does this look like in the UI? Naively I thought the schema would include a single string (e.g. "amazon_visual", ts, dur)

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    story_start AS ts,
+    interactive_presentation - story_start AS dur
+  FROM chrome_loadline2_stages
+
+  UNION ALL
+  SELECT
+    page,
+    page AS name,
+    story_start AS ts,
+    story_finish - story_start AS dur
```
victorvianna, line 48: (I feel like I should have asked this before, sorry) What value does story_finish bring? I feel like this could cause confusion - some people could think the final score comes from this third slice. Especially because this is in the "MetricTrack"

I'm not even clear on the semantics of this timestamp, given it is emitted after a 100ms wait
https://source.chromium.org/chromium/chromium/src/+/main:third_party/crossbench/config/benchmark/loadline2/page_config_phone.hjson;l=45;drc=47487469f296a56dbbf83007ecf0fa8e3cb04cd4

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  FROM chrome_loadline2_stages
+)
+SELECT *
+FROM all_generated_slices
+ORDER BY page, ts, dur DESC;
```
victorvianna, line 53: add `name` here to ensure consistent order (but does order even matter here?)

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      UNION ALL
+      SELECT
+        page,
+        'network' AS name,
+        start_request AS ts,
+        end_request - start_request AS dur
+      FROM chrome_loadline2_stages
+
+      UNION ALL
```
victorvianna, line 84: this slice overlaps with the one above, does perfetto UI automatically handle that or do we have to specify the vertical position somehow?
bonus: a screenshot attached to the CL description would be neat, you can use bit.ly since it's open source

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+        renderer_ready - start_request AS dur
+      FROM chrome_loadline2_stages
+      WHERE renderer_ready > start_request
+
+      UNION ALL
+      SELECT
+        page,
+        'renderer' AS name,
+        MAX(renderer_ready, end_request) AS ts,
+        ${metric}_mark - MAX(renderer_ready, end_request) AS dur
```
victorvianna, line 98: what happens if any of these `dur`s ends up being accidentally negative? does perfetto crash?

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+const INTERACTIVE_TRACK_URI =
+  'org.chromium.LoadLine2#InteractiveBreakdownTrack';
+
+export default class implements PerfettoPlugin {
+  static readonly id = 'org.chromium.LoadLine2';
+  static readonly description =
+    'Show breakdown of LoadLine 2 metrics on a separate track.';
+
+  async onTraceLoad(trace: Trace): Promise<void> {
+    if (await isLoadLineTrace(trace)) {
```
victorvianna: nit: negate and early return

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+        '_chrome_loadline2_plugin_interactive_breakdown',
+      );
+      addTracks(trace);
+    }
+  }
+}
+
+async function isLoadLineTrace(trace: Trace): Promise<boolean> {
+  const queryRes = await trace.engine.query(`
+      SELECT COUNT(*) AS cnt FROM slice WHERE name GLOB 'LoadLine2/*';
```
victorvianna: nitty nit: SELECT EXISTS(SELECT * FROM slice WHERE name GLOB 'Loadline2/*')
(so it stops upon finding the first instance)

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+  return it.cnt > 0;
+}
+
+async function prepareTables(trace: Trace): Promise<void> {
+  await trace.engine.query(METRIC_QUERY);
+  await trace.engine.query(generateBreakdownQuery('visual'));
+  await trace.engine.query(generateBreakdownQuery('interactive'));
+}
+
+async function registerTrack(
```
victorvianna: I don't think it's async because it's not `await`ing anything. Same for addTracks() below

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+): Promise<void> {
+  trace.tracks.registerTrack({
+    uri,
+    renderer: SliceTrack.create({
+      trace,
+      uri,
+      dataset: new SourceDataset({
+        src,
+        schema: {
+          ts: LONG,
```
victorvianna, line 177: TIMESTAMP?

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
...
+          dur: LONG,
```
victorvianna, line 178: DURATION?

ui/src/plugins/org.chromium.LoadLine2/index.ts
victorvianna: since the CL is open source I mean

ui/src/plugins/org.chromium.LoadLine2/index.ts
```diff
@@ -0,0 +1,206 @@
+// Copyright (C) 2025 The Android Open Source Project
```
victorvianna, line 1: Wait, one high-level concern. Given this logic is in a plugin, will we be able to reuse it in crossbench to compute the metrics / breakdown? If not, that's bad

ui/src/plugins/org.chromium.LoadLine2/index.ts
victorvianna: Oh, I meant imgur here, sorry

stevegolton (approved):
LGTM

victorvianna:
(Sorry for the delay here, I reviewed the PR last week but didn't press "Add your review")

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: Everything required for the metrics and breakdown is in the stdlib module, `chrome.loadline_2`. The logic here only creates slices for visualisation.

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: `name` is what will be shown in the UI. `page` is there just for sorting purposes. It's probably not strictly required, but makes tables easier to read manually, in case something breaks. I'd prefer keeping the `page` column, unless there's a strong reason to remove it.

Re what it looks like in the UI: you can play around with the plugin using [this link](https://storage.googleapis.com/perfetto-ci-artifacts/gh-19739172799-ui/ui/index.html) (see the first automatic comment on the CL). Note you'll need to enable the plugin after you open the UI for the first time, it's not enabled by default.

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: [story_start, story_finish] is the period when the corresponding page is visible on the screen. It's not required for the metrics, but it helps navigating the trace, because it roughly corresponds to the period of high Renderer activity.

I think it's standard practice for custom tracks in Peffetto: high-level slice at the top, detailed breakdown underneath it (see e.g. PipelineReporter track). So it shouldn't cause any confusion.

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: My idea was to make slices line up following the critical path. So, for example, if the network is on the critical path, the breakdown looks like this
```
[browser][    network    ][renderer]
         [startup]
```
but when startup is on the critical path instead, it looks like this
```
[browser][    startup    ][renderer]
         [network]
```
and not this
```
[browser][network]        [renderer]
         [    startup    ]
```
which is not pretty IMO.

That's why I sort by duration, but not by name.

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: See above. The vertical position is determined by sorting within the table.

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: The UI will continue working, but this track will be missing. So in those cases when the mark order is variable, I added the relevant WHERE clauses (see e.g. `process_start` slice above).

In this case, `dur` being negative really means that something's severely broken, so I see no point in trying to visualise anything.

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: Done

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: Done

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: Done

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: Module '"../../trace_processor/query_result"' has no exported member 'TIMESTAMP'

ui/src/plugins/org.chromium.LoadLine2/index.ts
khokhlov962: Module '"../../trace_processor/query_result"' has no exported member 'DURATION'

PR #3892, opened 2025-11-27
docs: Add AGENTS-ui.md

primiano:


docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+      m(Button, {label: attrs.value, onclick: () => this.expanded = !this.expanded}),
+      this.expanded && m('.details', 'Expanded content'),
+    );
+  }
+}
+```
+
+**Mithril Rules:**
+
+- No need to call`m.redraw()` from within lifecycle hooks or handlers. Mithril automatically schedules redraws.
```
stevegolton: Maybe specify that redraws are triggered automatically:
- in DOM event handlers registered via mithril components
- after trace processor queries complete

Not after manually registered event handlers.

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+
+m(Gate, {open: this.isVisible}, m(ExpensiveComponent));
+```
+
+### Widget Library
+
+The `ui/src/widgets/` directory contains reusable components. Always check here before creating new UI elements:
+
+- `Button`, `ButtonBar`, `ButtonGroup` - Various button styles
+- `Menu`, `MenuItem`, `MenuDivider` - Dropdown menus
```
stevegolton: Probably makes sense to mention PopupMenu here which is usually where these widgets are used

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+- `Popup` - Floating popup containers
+- `Modal` - Modal dialogs
+- `TextInput`, `Select`, `Checkbox`, `Switch` - Form controls
+- `Tree` - Tree view component
+- `Tabs` - Tabbed interface
+- `Spinner` - Loading indicator
+- `EmptyState` - Empty state placeholder
```
stevegolton, line 159: Might be worth mentioning DataGrid here but I can add this later

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+If you really need to create a track from scratch:
+
+```typescript
+async onTraceLoad(trace: Trace): Promise<void> {
+  const uri = '/my_plugin/my_track';
+
+  // Register the track definition
+  trace.tracks.registerTrack({
+    uri,
+    renderer: new MyTrackRenderer(trace, uri),
```
stevegolton: I'm not sure this example really helps. No one should be creating their own tracks or at least it shouldn't be common enough to put into this file. Also, this example doesn't help guide how the renderer would be created, just how it should be used.

I think an example showing DatasatSliceTrack would be the most useful.

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+}
+```
+
+## CSS/SCSS Conventions
+
+Stylesheets live in `ui/src/assets/` and component-specific `.scss` files alongside components.
+
+- Use the `pf-` prefix for all CSS classes (Perfetto namespace)
+- Follow BEM-like naming: `.pf-component`, `.pf-component__element`, `.pf-component--modifier`
+- Use CSS custom properties (variables) defined in `theme.scss` for colors
```
stevegolton: The CSS colour variables live in theme_provider.scss - theme.scss contains scss utilities only (yes it's a bad name, needs renaming)

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+- Support both light and dark themes using semantic color variables
+
+## Common Pitfalls to Avoid
+
+1. **Don't create new widgets without checking existing ones** - The widget library is comprehensive.
+2. **Don't use `window`** - Use `self` for compatibility with Web Workers.
```
stevegolton: I'm not sure we actually practice this?

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+3. **Try to use the Trace object as much as possible** - Plumb the Trace object through the hierarchy and keep `AppImpl.instance.trace` as a last resort.
```
stevegolton: We should never use AppIml.instance.trace not even as a last resort!

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+**Use `ReadonlyArray<T>` for arrays that shouldn't be modified:**
+```typescript
+// Bad
+function process(items: string[]): void { ... }
+
+// Good
+function process(items: ReadonlyArray<string>): void { ... }
+```
+
+**No nested ternaries - they're never the right option:**
```
stevegolton: I'm not sure... I think they have their place!? Depends on the situation.

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+  color: var(--pf-color-foreground);
+  background: var(--pf-color-background);
+}
+```
+
+**Use CSS `:hover` pseudo-element, not JavaScript hover handlers:**
+```typescript
+// Bad
+m('div', {
+  onmouseenter: () => this.hovered = true,
```
stevegolton: I've never seen an LLM get this wrong :/. Regardless, we could probably just make it a simple bullet point to save tokens.

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+  class: this.hovered ? 'hovered' : '',
+})
+
+// Good - use CSS :hover in stylesheet
+// .pf-my-element:hover { background: var(--pf-color-hover); }
+```
+
+### Mithril-Specific Rules
+
+**Never call `m.redraw()` - it's almost never needed:**
```
stevegolton: Isn't this duplicated?

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+}
+
+// Good
+onclick: () => {
+  this.value = newValue;
+  // Mithril auto-redraws after event handlers
+}
+```
+
+**Don't use `oncreate`/lifecycle hooks for things that can be done in `view()`:**
```
stevegolton, line 332: I'm not sure I understand this rule? Doesn't it depend on what you're doing?

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+
+// Good - compute in view or use lazy initialization
+view() {
+  const computedValue = expensiveComputation();
+  return m('div', computedValue);
+}
+```
+
+**Components shouldn't cache parent data:**
+```typescript
```
stevegolton: This seems very specific to... DataGrid? Maybe?

docs/AGENTS-ui.md
```diff
@@ -0,0 +1,524 @@
...
+// Bad (if current year is 2025)
+// Copyright (C) 2024 The Android Open Source Project
+
+// Good
+// Copyright (C) 2025 The Android Open Source Project
+```
+
+### Code Organization
+
+**Use WeakMap for trace-scoped data instead of modifying Trace API:**
```
stevegolton: Not sure we should always do this?

stevegolton (approved):
In general looks good, though I do think it could be more concise perhaps.

Approved with comments - happy to land in that state and see how it goes / edit later.

docs/AGENTS-ui.md
primiano: good point. I linked at /docs/contributing/ui-plugins.md mentioning it, as that .md doc covers it well

docs/AGENTS-ui.md
primiano: oh yeah originally AI wrote this but then I manually wrote the other bit and forgot to remove this.

docs/AGENTS-ui.md
primiano: Yeah removed

docs/AGENTS-ui.md
primiano: I mean, if you need Trace-scoped data, that's the way to go no?
but is also super niche... maybe let's remove it

docs/AGENTS-ui.md
primiano: I think it picked up one of my old pet peeves. it's nicer to lazy.initialize things in the view() methods than splitting across oncreate which makes the code harder to follow IMHO

primiano: let me refine it a bit though

docs/AGENTS-ui.md
primiano: yeah fair, removed it

docs/AGENTS-ui.md
primiano: removed

PR #3896, opened 2025-11-28
Perfetto init script: get rid of `wait_for_prop`.

KirillTim:
When changing the init script to support perfetto reboot traces,
the perfetto initialization process became two phased:
1. `property:persist.traced.enable` is triggered first
2. then, on other trigger, `traced` service is started

Previously, the action was triggered immediately by step 1, but
included a `wait_for_prop` command waiting for step 2.
This blocked the boot sequence, preventing `traced` from
ever starting, resulting in a deadlock.

To fix this, we removed `wait_for_prop` and moved `sys.trace.traced_started`
to the action trigger. The action is now triggered only after
`traced` has successfully started.

The integration test for this change is add in https://github.com/google/perfetto/pull/3923


test/cts/host_side_tests/src/android/perfetto/cts/test/PerfettoBootTimeTraceHostTest.java
```diff
@@ -0,0 +1,273 @@
...
+                                mTraceStopTriggerConfigFile,
+                                BOOT_TRACE_STOP_TRIGGER_CONFIG_ON_DEVICE_PATH))
+                .isTrue();
+        CommandResult result =
+                mTestDevice.executeShellV2Command(
+                        "perfetto --txt -c " + BOOT_TRACE_STOP_TRIGGER_CONFIG_ON_DEVICE_PATH);
+        assertThat(result.getStatus()).isEqualTo(CommandStatus.SUCCESS);
+
+        // Assert boot tracing session stopped.
+        runningTracingSessionNames = assertGetRunningTracingSessionNames();
```
primiano: You should do this in a loop with 1 second delay, for at most 30 seconds

You are waiting for the exit of the client that sends the trigger, but you are not waiting that the traced instance that is sensitive to the trigger has stopped. you need to linearize with that

test/cts/host_side_tests/src/android/perfetto/cts/test/PerfettoBootTimeTraceHostTest.java
```diff
@@ -0,0 +1,273 @@
...
+            "/data/misc/perfetto-traces/boottrace.perfetto-trace";
+    private static final String BOOT_TRACE_CONFIG_ON_DEVICE_PATH =
+            "/data/misc/perfetto-configs/boottrace.pbtxt";
+    private static final String BOOT_TRACE_STOP_TRIGGER_CONFIG_ON_DEVICE_PATH =
+            "/data/misc/perfetto-configs/stopboottracetrigger.pbtxt";
+
+    private static final String BOOT_TRACE_CONFIG_SESSION_NAME = "PerfettoBootTimeTraceHostTest";
+    private static final String BOOT_TRACE_CONFIG =
+            """
+            unique_session_name: "%s"
```
primiano: add the RT boost otherwise is going to be flaky

test/cts/host_side_tests/src/android/perfetto/cts/test/PerfettoBootTimeTraceHostTest.java
```diff
@@ -0,0 +1,273 @@
...
+                    && message.contains("sys.boot_completed=1")) {
+                hasBootCompletedLine = true;
+                break;
+            }
+        }
+        assertThat(hasBootCompletedLine).isTrue();
+    }
+
+    @Test
+    public void testBootTraceStartAndStop() throws DeviceNotAvailableException, IOException {
```
primiano: can you add a comment explaining what the tests do?

test/cts/host_side_tests/src/android/perfetto/cts/test/PerfettoBootTimeTraceHostTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+        // Assert there is no trace file before reboot.
+        assertThat(mTestDevice.doesFileExist(BOOT_TRACE_RESULT_ON_DEVICE_PATH)).isFalse();
+
+        mTestDevice.reboot();
+
+        // Assert property is reset after reboot.
+        assertThat(mTestDevice.getProperty(BOOT_TRACE_ENABLE_PROPERTY)).isNotEqualTo("1");
+        // Assertf tracing session was already stopped
+        runningTracingSessionNames = assertGetRunningTracingSessionNames();
```
primiano: i'd do a loop with timeout here as well, because it's not guaranteed that perfetto will stop immediately if the system is busy

PR #3911, opened 2025-11-28
Heap flamegraph: sort roots by name to improve stability

lukaville:
Currently often the GC root could be chosen non-deterministically (based on the row number) when calculating shortest path to the GC root in java heap flamegraph table.
 
Sorting them by name could improve stability for the cases when we want to compare heap dumps. We might consider sorting all sibling nodes by name when calculating path to roots, but currently it seems like root instability causes the biggest problem. Added only sorting of the roots in this PR to ensure minimum impact on performance.

E.g. in these two traces `dalvik.system.PathClassLoader` has exactly the same set of GC roots that reference it but we chose different roots, this causes issues when trying to compare the trees (they are treated as completely separate subtrees):

image

### DIFF
--- a/test/trace_processor/diff_tests/parser/profiling/tests_heap_graph.py
+++ b/test/trace_processor/diff_tests/parser/profiling/tests_heap_graph.py
@@ -388,35 +388,6 @@ class ProfilingHeapGraph(TestSuite):
         """,
         out=Path('heap_graph_flamegraph_system-server-heap-graph.out'))
 
-  def test_heap_graph_class_tree_system_server_heap_graph(self):
-    return DiffTestBlueprint(
-        trace=DataPath('system-server-heap-graph-new.pftrace'),
-        query="""
-        INCLUDE PERFETTO MODULE android.memory.heap_graph.class_tree;
-        SELECT
-          ifnull(name, '[Unknown]') as name,
-          self_size,
-          self_count
-        FROM _heap_graph_class_tree
-        WHERE graph_sample_ts = (SELECT max(graph_sample_ts) FROM heap_graph_object)
-          AND upid = (SELECT max(upid) FROM heap_graph_object)
-        ORDER BY self_size DESC, name
-        LIMIT 10;
-        """,
-        out=Csv('''
-          "name","self_size","self_count"
-          "java.lang.String",1845640,41197
-          "java.lang.String",675400,348
-          "android.content.pm.parsing.component.ParsedActivity",577760,3611
-          "int[]",455868,1
-          "byte[]",271266,2
-          "java.lang.String",268496,3375
-          "byte[]",260021,216
-          "android.content.pm.parsing.component.ParsedActivity",212640,1329
-          "java.lang.Object[]",212348,4008
-          "android.content.pm.parsing.component.ParsedService",150400,1880
-        '''))
-
   def test_heap_graph_root_sorting(self):
     # We expect RootA to be the parent because it comes first alphabetically
     return DiffTestBlueprint(


src/trace_processor/importers/proto/heap_graph_tracker.cc
LalitMaganti, line 1133: This seems like the wrog place to do this. Sure it will work for pprof but then pprofs will not be consistent with the Perfetto UI.

Also experimental_flamegraph is intended for deletion with the pprof export code being rebased on top of the same code the UI uses: at that point again this will break.

lukaville: Ah I thought experimental_flamegraph is also used in the UI. Would it make sense to introduce similar change to both experimental_flamegraph and whatever the UI currently uses?

Seems like currenlty UI uses _heap_graph_object_min_depth_tree for this?

LalitMaganti: Yes _heap_graph_object_min_depth_tree is what you would need to modify.

lukaville: Added roots sorting for _heap_graph_object_min_depth_tree too, please have a look. Verified in the UI: now most of the roots in two different traces are the same (before this in the UI they were resolved to different roots). Thanks!

LalitMaganti (commented):
LGTM but would like Ioannis to chip in.

src/traceconv/trace_to_pprof_integrationtest.cc
```diff
@@ -200,7 +200,7 @@ TEST_F(TraceToPprofRealTraceTest, AllocationCountForClass) {
       pprof.get_samples("android.content.pm.parsing.component.ParsedActivity")
           .size(),
       5U);
-  EXPECT_EQ(pprof.get_sample_count(), 83256U);
+  EXPECT_EQ(pprof.get_sample_count(), 83028U);
```
ilkos, line 203: Hmm why did this change?

lukaville: This seems to be expected as this is the number of unique samples, so we might have different number of unique paths based on the order of roots:

E.g. let's say we have:
- RootA
- RootB
- ChildShared object of type Child (8 bytes)
- ChildB object of type Child (8 bytes)

(both roots reference ChildShared, only RootB reference ChildB)

- Traversing RootA first: visits RootA -> ChildShared, then traversing RootB: visits RootB -> ChildB
  - Total found 2 paths: RootA -> Child (8 bytes), RootB -> Child (8 bytes)
- Traversing RootB first: visits RootB -> ChildShared; RootB -> ChildB, traversing RootA: ChildShared already visited
  - Total found 1 path: RootB -> Child (16 bytes)

lukaville:
> At the same time I do wonder if this is the right traversal for your use case or whether you should be using a dominator tree comparison which is going to be much clearer, deterministic and easy to understand.

Yes, I agree that dominator traversal seems like a better fit for this, possibly it should be shown by default, we will be looking into adding this (this is in the context of Android SystemUI lab performance regression infrastructure). I have tried it and it works quite well but I think users should still have an option to choose either of the traversals.

The main downside from what I have seen is that with the dominator tree parent-child relations might not match the actual heap graph references which might be confusing. When with the shortest path traversal there is an actual reference shown to the user (even though it might be not the only one)

so imho it would be better to support both options and based on practical regressions we would be able to decide which one works better

test/trace_processor/diff_tests/parser/profiling/tests_heap_graph.py
```diff
@@ -388,6 +388,69 @@ def test_heap_graph_flamegraph_system_server_heap_graph(self):
         """,
         out=Path('heap_graph_flamegraph_system-server-heap-graph.out'))
 
+  def test_heap_graph_class_tree_system_server_heap_graph(self):
```
ilkos: What is this test verifying?

lukaville: it's more like an integration test to verify that it works and doesn't fail on a bigger real trace rather than small synthetic one
(similar to this one https://github.com/google/perfetto/pull/3911/files#diff-888bd9453422db0bb50b3efa535bf7e7c6c0b3c35b68a9b1bd26e3013056229e)

I'm happy to remove it if you think it doesn't add much value

ilkos: I'd probably remove it, these tests are not particularly debuggable, especially since it's impossible to predict what it should be doing

lukaville: Sure, I have removed the test, thanks!

PR #3916, opened 2025-11-30
Support ftrace event i2c_write and i2c_reply data recording

### LINKED ISSUE #3914: Record and show i2c data for ftrace event i2c_write and i2c_reply
For the ftrace events `i2c_write` and `i2c_reply`, it will record the raw I2C transfer data with `__data_loc __u8[] buf` field and here is the complete event format:

```
name: i2c_write
ID: 624
format:
	field:unsigned short common_type;	offset:0;	size:2;	signed:0;
	field:unsigned char common_flags;	offset:2;	size:1;	signed:0;
	field:unsigned char common_preempt_count;	offset:3;	size:1;	signed:0;
	field:int common_pid;	offset:4;	size:4;	signed:1;

	field:int adapter_nr;	offset:8;	size:4;	signed:1;
	field:__u16 msg_nr;	offset:12;	size:2;	signed:0;
	field:__u16 addr;	offset:14;	size:2;	signed:0;
	field:__u16 flags;	offset:16;	size:2;	signed:0;
	field:__u16 len;	offset:18;	size:2;	signed:0;
	field:__data_loc __u8[] buf;	offset:20;	size:4;	signed:0;

print fmt: "i2c-%d #%u a=%03x f=%04x l=%u [%*phD]", REC->adapter_nr, REC->msg_nr, REC->addr, REC->flags, REC->len, REC->len, __get_dynamic_array(buf)
```

Currently it only recorded with `uint32 buf` in the message `I2cWriteFtraceEvent` and `I2cReplyFtraceEvent`:

https://github.com/google/perfetto/blob/a8060b7bea06166ac541ce97670053882375e309/protos/perfetto/trace/ftrace/i2c.proto#L15-L22

https://github.com/google/perfetto/blob/a8060b7bea06166ac541ce97670053882375e309/protos/perfetto/trace/ftrace/i2c.proto#L28-L35

I hope Perfetto can fully capture I2C transfer data and display it in the Perfetto UI, enabling more effective hardware debugging without the need to solder wires for a logic analyzer.

The basic idea is to change the `uint32 buf` field to `bytes buf`, format it as a hexadecimal string in the `trace_processor` (maybe it is not the best way), and display it in the UI. I’ve already tried this on my [branch](https://github.com/tfx2001/perfetto/commit/5efe92ebba85061be24ce213b8df9e76db7bdd66) but am not sure what to do next. Any guidance or help would be very appreciated!

tfx2001:
# Purpose

This PR support I2C transfer data recording from ftrace event `i2c_write` and `i2c_reply`.

# Other

Resolves #3914.

### DIFF
--- a/protos/perfetto/config/perfetto_config.proto
+++ b/protos/perfetto/config/perfetto_config.proto
@@ -4933,9 +4933,19 @@ message TraceConfig {
       // Introduced in: Perfetto v54.
       optional string name = 4;
 
-      // The semantic types this rule applies to. If empty, the rule applies
-      // to all string fields regardless of their semantic type. If non-empty,
-      // the rule only applies to fields whose semantic type is in this list.
+      // The semantic types this rule applies to.
+      //
+      // SEMANTIC_TYPE_UNSPECIFIED (0) is treated as its own distinct category:
+      // - If empty, the rule applies only to fields with UNSPECIFIED type.
+      // - If non-empty, the rule applies only to fields whose semantic type
+      //   is explicitly listed. To match UNSPECIFIED fields, you must include
+      //   SEMANTIC_TYPE_UNSPECIFIED in the list.
+      //
+      // Examples:
+      //   semantic_type: []           -> matches only UNSPECIFIED fields
+      //   semantic_type: [ATRACE]     -> matches only ATRACE fields
+      //   semantic_type: [UNSPECIFIED, ATRACE] -> matches both
+      //
       // See /rfcs/0011-subset-string-filter-rules.md for design details.
       // Introduced in: Perfetto v54.
       repeated SemanticType semantic_type = 5;
--- a/protos/perfetto/config/trace_config.proto
+++ b/protos/perfetto/config/trace_config.proto
@@ -666,9 +666,19 @@ message TraceConfig {
       // Introduced in: Perfetto v54.
       optional string name = 4;
 
-      // The semantic types this rule applies to. If empty, the rule applies
-      // to all string fields regardless of their semantic type. If non-empty,
-      // the rule only applies to fields whose semantic type is in this list.
+      // The semantic types this rule applies to.
+      //
+      // SEMANTIC_TYPE_UNSPECIFIED (0) is treated as its own distinct category:
+      // - If empty, the rule applies only to fields with UNSPECIFIED type.
+      // - If non-empty, the rule applies only to fields whose semantic type
+      //   is explicitly listed. To match UNSPECIFIED fields, you must include
+      //   SEMANTIC_TYPE_UNSPECIFIED in the list.
+      //
+      // Examples:
+      //   semantic_type: []           -> matches only UNSPECIFIED fields
+      //   semantic_type: [ATRACE]     -> matches only ATRACE fields
+      //   semantic_type: [UNSPECIFIED, ATRACE] -> matches both
+      //
       // See /rfcs/0011-subset-string-filter-rules.md for design details.
       // Introduced in: Perfetto v54.
       repeated SemanticType semantic_type = 5;
--- a/protos/perfetto/trace/perfetto_trace.proto
+++ b/protos/perfetto/trace/perfetto_trace.proto
@@ -4933,9 +4933,19 @@ message TraceConfig {
       // Introduced in: Perfetto v54.
       optional string name = 4;
 
-      // The semantic types this rule applies to. If empty, the rule applies
-      // to all string fields regardless of their semantic type. If non-empty,
-      // the rule only applies to fields whose semantic type is in this list.
+      // The semantic types this rule applies to.
+      //
+      // SEMANTIC_TYPE_UNSPECIFIED (0) is treated as its own distinct category:
+      // - If empty, the rule applies only to fields with UNSPECIFIED type.
+      // - If non-empty, the rule applies only to fields whose semantic type
+      //   is explicitly listed. To match UNSPECIFIED fields, you must include
+      //   SEMANTIC_TYPE_UNSPECIFIED in the list.
+      //
+      // Examples:
+      //   semantic_type: []           -> matches only UNSPECIFIED fields
+      //   semantic_type: [ATRACE]     -> matches only ATRACE fields
+      //   semantic_type: [UNSPECIFIED, ATRACE] -> matches both
+      //
       // See /rfcs/0011-subset-string-filter-rules.md for design details.
       // Introduced in: Perfetto v54.
       repeated SemanticType semantic_type = 5;
--- a/src/protozero/filtering/string_filter.cc
+++ b/src/protozero/filtering/string_filter.cc
@@ -138,9 +138,7 @@ bool StringFilter::MaybeFilterInternal(char* ptr,
   bool atrace_find_tried = false;
   const char* atrace_payload_ptr = nullptr;
   for (const Rule& rule : rules_) {
-    // UNSPECIFIED (0) fields always match - we don't know what they contain,
-    // so apply rules conservatively. For other types, check the mask.
-    if (semantic_type != 0 && !rule.semantic_type_mask.IsSet(semantic_type)) {
+    if (!rule.semantic_type_mask.IsSet(semantic_type)) {
       continue;
     }
     switch (rule.policy) {
--- a/src/protozero/filtering/string_filter.h
+++ b/src/protozero/filtering/string_filter.h
@@ -41,9 +41,10 @@ class StringFilter {
     static constexpr size_t kBitsPerWord = sizeof(Word) * 8;
     static constexpr size_t kLimit = 128;
 
-    // Returns a SemanticTypeMask with all bits set (applies to all types).
-    static constexpr SemanticTypeMask All() {
-      return SemanticTypeMask(Word(-1), Word(-1));
+    // Returns a SemanticTypeMask with only bit 0 set (UNSPECIFIED only).
+    // UNSPECIFIED is its own category and only matches if bit 0 is set.
+    static constexpr SemanticTypeMask Unspecified() {
+      return SemanticTypeMask(Word(1), Word(0));
     }
 
     // Creates a SemanticTypeMask from raw word values (for testing).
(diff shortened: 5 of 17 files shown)


LalitMaganti:
Ping on reviewing this @primiano @rsavitski

src/traced/probes/ftrace/proto_translation_table.cc
```diff
@@ -320,13 +322,23 @@ bool InferFtraceType(const std::string& type_and_name,
...
-  if (Contains(type_and_name, "__data_loc char[] ")) {
+  if (Contains(type_and_name, "__data_loc ")) {
     if (size != 4) {
       PERFETTO_ELOG("__data_loc with incorrect size: %s (%zu)",
                     type_and_name.c_str(), size);
       return false;
     }
-    *out = kFtraceDataLoc;
+
+    if (Contains(type_and_name, "char[] ")) {
```
primiano: I think this should be still `__data_loc char[]`.
Here's the problem: i noticed that there are some events that are "__data_loca unsigned char[]".
Those should not be treated as strings, they actually should be handled by your case below, because unsigned char is == u8.

Technically this will over-parse strings, and could cause some regression.

I would keep the full `__data_loc char[]` in this branch, and in the one below go

Contains(..., "__data_loc u8[]") || Contains(..., "__data_loc __u8[]") || Contains(..., "__data_loc unsigned char[]")

src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -87,26 +119,40 @@ bool ReadDataLoc(const uint8_t* start,
...
-    PERFETTO_DFATAL("couldn't read __data_loc value");
+  const uint8_t* string_start = nullptr;
+  size_t len = 0;
+
+  if (!GetDataLocStartAndLength(start, field_start, end, field, &string_start,
+                                &len))
+    return false;
+  if (PERFETTO_UNLIKELY(len == 0))
+    return true;
+  if (PERFETTO_UNLIKELY(string_start < start || string_start + len > end)) {
```
primiano: isn't this already handled by GetDataLocStartAndLength which does the same check?

src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -87,26 +119,40 @@ bool ReadDataLoc(const uint8_t* start,
...
+  const uint8_t* data_start = nullptr;
+  size_t len = 0;
 
+  if (!GetDataLocStartAndLength(start, field_start, end, field, &data_start,
+                                &len))
+    return false;
   if (PERFETTO_UNLIKELY(len == 0))
     return true;
-  if (PERFETTO_UNLIKELY(string_start < start || string_start + len > end)) {
+  if (PERFETTO_UNLIKELY(data_start < start || data_start + len > end)) {
```
primiano: ditto. this should be handled by the common code GetDataLocStartAndLength ?

src/tools/ftrace_proto_gen/proto_gen_utils.cc
```diff
@@ -199,6 +212,8 @@ ProtoType InferProtoType(const FtraceEvent::Field& field) {
     return ProtoType::String();
   if (Contains(field.type_and_name, "char * "))
     return ProtoType::String();
+  if (Contains(field.type_and_name, "__u8[] "))
```
primiano: shouldn't this have also "u8[] " and "unsigned char[]"?

src/tools/ftrace_proto_gen/ftrace_proto_gen_unittest.cc
```diff
@@ -29,10 +29,20 @@ TEST(FtraceEventParserTest, InferProtoType) {
   EXPECT_EQ(
       InferProtoType(Field{"__data_loc char[] foo", 0, 4, false}).ToString(),
       "string");
-  EXPECT_EQ(InferProtoType(Field{"char[] foo", 0, 8, false}).ToString(),
-            "string");
```
tfx2001, line 33: Removing this because the ftrace format data does not contain any field of type `char[]`.

primiano:
> > Hey thanks a lot for the contribution and, on behalf of the team, sorry for the delay in this. This fell off our radar.
> > The contribution is really great. I have few minor suggestions, but other than that looks geat!
> 
> Thanks for taking the time to review this PR. I’ve already updated the code based on your feedback, but I still have a few questions:
> 
> 1. As part of my changes, `__data_loc u8[]` and `__data_loc __u8[]` should now be represented as `bytes` in `.proto` files. However, due to the [`GetCommon()`](https://github.com/google/perfetto/blob/c774fc2b3a46443db3adf8691a9abd9e1633cc06/src/tools/ftrace_proto_gen/proto_gen_utils.cc#L176) function, the fields remain as `uint32` in existing proto definitions. I manually updated `i2c.proto` to use the `bytes` type. Should we apply the same change to other ftrace events that use `__data_loc u8[]` ? For example:


Let me see if I get this right. THis is because UnionFields -> GetCommon() tries to maintain backwards compat with the previous definition right? In this case the problem is that the previous definition is buggy.
Just to check if you:
- Delete the uint8[] fields from the existing protos
- Then rerun ftrace_proto_gen with your patch applied
- Then those field become of the correct type (bytes) right?

If so, it's fine to do so. In this case there is no backwards compat to maintin. the past was just wrong.
And these fields, I'm pretty sure they are unused. 
 
> https://github.com/google/perfetto/blob/c774fc2b3a46443db3adf8691a9abd9e1633cc06/protos/perfetto/trace/ftrace/panel.proto#L12-L15
> 
> 2. Similarly, I’ve updated the handling of `__data_loc unsigned char[]` so that it maps to `bytes` in proto files. What should we do about existing proto files that currently represent this as `string` ? For instance:
> 
> https://github.com/google/perfetto/blob/c774fc2b3a46443db3adf8691a9abd9e1633cc06/protos/perfetto/trace/ftrace/scsi.proto#L19

Delete them from the proto (manually) and re-generate them with the correct bytes type.

Generally in .proto string -> bytes conversions are totally fine, as bytes is a superset of string and their wire encoding is identical.

> 
> 3. How should `bytes` data be displayed in the UI?

Let's figure out that later? My preference would be: 
- If most bytes (say 90% of them)  are in the printable ASCII range 32-127: show that as a string where the few non-printable chars are escaped + the hex encoding, like `hello\x00world (68 65 6C 6C 6F 00 77 6F 72 6C 64)`
- If not just show a hex string like `FF AA 01 02 04 `
- 
> 
> > P.S: did you check if it actually works? do you get i2c payloads in the trace?
> 
> Yes. I recorded a trace data and converted it into text format. You can check the gist below.
> 
> https://gist.github.com/tfx2001/c90053391a9f17a8b6713fd05fa2043c

ah nice

tfx2001:
> Let me see if I get this right. THis is because UnionFields -> GetCommon() tries to maintain backwards compat with the previous definition right? In this case the problem is that the previous definition is buggy. Just to check if you:
> 
> * Delete the uint8[] fields from the existing protos
> * Then rerun ftrace_proto_gen with your patch applied
> * Then those field become of the correct type (bytes) right?
> 
> If so, it's fine to do so. In this case there is no backwards compat to maintin. the past was just wrong. And these fields, I'm pretty sure they are unused.

I agree with you that we don't need to maintain backwards compatibility for these buggy fields. And I've updated relevant `.proto` files. 

I also found a compilation error caused by a missing `perfetto/public/protos/common/semantic_type.pzc.h` after running `tools/gen_all --proto`. Should I include the fix for this in the same PR?

src/trace_processor/perfetto_sql/stdlib/export/to_svg.sql
```diff
@@ -374,9 +374,11 @@ SELECT
       $width_pixel,
       $height_pixel,
       _state_color($state, $io_wait),
-      'Thread State: ' || $state || ' (' || _format_duration($dur) || ')' ||
-      CASE WHEN $blocked_function IS NOT NULL THEN
-      ', blocked by ' || $blocked_function ELSE '' END,
+      'Thread State: ' || $state || ' (' || _format_duration($dur) || ')' || CASE
+        WHEN $blocked_function IS NOT NULL
+        THEN ', blocked by ' || $blocked_function
+        ELSE ''
```
primiano: there seems to be only formatting changes here, but i'm curious how you ended up reformatting this file?

src/traced/probes/ftrace/cpu_reader.cc
```diff
@@ -72,6 +72,38 @@ struct EventHeader {
   uint32_t time_delta : 27;
 };
 
+// Reads a __data_loc field and extracts the data start pointer and length.
+template <typename T>
```
primiano: you don't need the template here. you have two callers and for both T=const uint8_t*, so this should be just const uint8_t**

primiano:
> I also found a compilation error caused by a missing `perfetto/public/protos/common/semantic_type.pzc.h` after running `tools/gen_all --proto`. Should I include the fix for this in the same PR?

depends on what fix and for what. if it's a more general problem, send a different PR?

src/trace_processor/perfetto_sql/stdlib/export/to_svg.sql
tfx2001: I formatted this file by running `tools/format-sources`, looks like it was introduced by another commit during rebase.

src/traced/probes/ftrace/cpu_reader.cc
tfx2001: Updated.

LalitMaganti (approved):

tfx2001:
> > I also found a compilation error caused by a missing `perfetto/public/protos/common/semantic_type.pzc.h` after running `tools/gen_all --proto`. Should I include the fix for this in the same PR?
> 
> depends on what fix and for what. if it's a more general problem, send a different PR?

Looks like it has been fixed in #4437.

tfx2001:
I think we should update the test data `test/data/trace-redaction-api-capture.pftrace` since we’ve changed the proto definition.

I found that in `FtraceParser::ParseTypedFtraceToRaw()`, it attempts to insert `DsiTxFtraceEvent::tx_buf` as a string into args because its schema type is `ProtoSchemaType::kBytes`. However, in the protobuf blob, this field is encoded with wire type `ProtoWireType::kVarInt`, which is inconsistent with the expected bytes encoding.

LalitMaganti:
We will actually need to make trace processor deal with this edge case. Perfetto needs to open traces collected for backcompat reasons and if there's a test for it, we certainly need to address it.

LalitMaganti:
Hmm so would it be possible to change the field id for the proto which would solve the problem of "we gave it the wrong type". Old trace's fields would just be dropped while new ones would work.

tfx2001:
I've updated the `.proto` files and reserved the previous field ids for back compat. Now it pass the integration tests.

Please review them again. Thanks!

PR #3917, opened 2025-11-30
tp: Fix parsing log_message trace events

altimin:
Currently the log_message trace event body is written into the
track_event.log_message field, while the source location is written
into track_event.log_message.{file_name,function_name,line_number}.

This means that these arg sets cannot be correctly serialised to JSON,
as track_event.log_message is a string primitive and cannot have child
fields.

Fix this by writing the message into track_event.log_message.message
rather than just track_event.log_message.


LalitMaganti:
This is a backwards incompatible change so please add a note to the changelog

LalitMaganti (approved):

altimin:
> Ah actually still needs the above comment fixed.

Done.

PR #3918, opened 2025-11-30
dev/altimin/reland args refactor

altimin:
Reland "ui: refactor display of args"

This reverts #3293 and relands #3032, but using the new `parseJsonWithBigInts` function
to parse JSON correctly.

Leverage the new print_args() function to avoid reconstructing the args tree in JS.

This change doesn't affect UX, just simplifies the code.

### DIFF
--- a/ui/src/plugins/dev.perfetto.Ftrace/ftrace_details_panel.ts
+++ b/ui/src/plugins/dev.perfetto.Ftrace/ftrace_details_panel.ts
@@ -15,12 +15,11 @@
 import m from 'mithril';
 import {Time} from '../../base/time';
 import {renderArguments} from '../../components/details/args';
-import {ArgsDict, getArgs} from '../../components/sql_utils/args';
-import {asArgSetId} from '../../components/sql_utils/core_types';
+import {ArgsDict, parseArgs} from '../../components/sql_utils/args';
 import {Timestamp} from '../../components/widgets/timestamp';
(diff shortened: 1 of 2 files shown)


altimin:
@stevegolton — this now fixed, PTAL when you have moment, thanks!

ui/src/plugins/dev.perfetto.Ftrace/ftrace_details_panel.ts
```diff
@@ -101,40 +95,17 @@ export class FtraceEventDetailsPanel implements TrackEventDetailsPanel {
...
-    const args: Arg[] = [];
-    for (; it.valid(); it.next()) {
-      args.push({
-        id: asArgId(it.id),
-        flatKey: it.flatKey,
-        key: it.key,
-        displayValue: it.displayValue ?? 'NULL',
-      });
+    if (it.valid() && it.arg_set_id !== null) {
+      this.args = await getArgs(this.trace.engine, asArgSetId(it.arg_set_id));
```
stevegolton: I don't love these helper functions that hide the query, meaning you have to run multiple queries - I would much prefer it if we just include the `__intrinsic_arg_set_to_json` in the original query and use the helper to convert it into an ArgsDict object.

altimin: Done.

PR #3938, opened 2025-12-02
ui: Add glob, not glob, contains and not contains filters to datagrid

stevegolton:
Add a comprehensive set of filters to DataGrid:
- Equals (using distinct columns list)
- Not equals (using distinct columns list)
- Glob
- Not glob
- Contains
- Not contains

An additional attribute has also been added to DataGrid which allows users to selectively disable certain filter operations.

Note: 'not glob' was added to the filter model as we had no way before to indicate a negative glob. Contains/not contains uses glob/not glob.

The equals and not equals menu items use the distinct values submenu:
image

The contains, not contains, glob, not glob menu items use a free form text input:
image

Related: https://b.corp.google.com/issues/405200352

### DIFF
--- a/ui/src/assets/components/data_grid.scss
+++ b/ui/src/assets/components/data_grid.scss
@@ -35,11 +35,6 @@
   &__loading {
     flex: none;
   }
-
-  // &__text-filter-form {
-  //   margin-top: -4px;
-  //   padding: 2px;
-  // }
 }
 
 .pf-aggr-cell {


aMayzner:
Is there a way to disable those in datagrid implementations?

stevegolton:
> Is there a way to disable those in datagrid implementations?

Just added it (see latest commit)

ui/src/assets/components/data_grid.scss
```diff
@@ -35,6 +35,11 @@
   &__loading {
     flex: none;
   }
+
```
aMayzner: remove

PR #3944, opened 2025-12-02
exp: Filter distinct equality on datagrid

aMayzner:
## Summary

  Add support for filtering on distinct values from the datagrid. When users select multiple distinct values to filter on, the query builder now properly handles these as multiple filters combined with OR logic.

  ## Changes

  - Extended filter system to handle arrays of filter values (FilterIn type) in addition to single values
  - Added filter normalization that expands IN/NOT IN filters into multiple equality filters connected with OR operator (since query builder doesn't have native IN support yet)
  - Refactored common graph manipulation code into reusable utilities (`insertNodeBetween`, `reconnectParentsToChildren`, `getInputNodeAtPort`, `getAllInputNodes`)
  - Centralized filter value formatting logic into `formatFilterValue` helper for consistent display across UI
  - Updated proto generation to handle array-based filter values with validation for mixed-type arrays
  - Added `setFiltersOnNode` helper to reduce code duplication when setting filters

  ## Notes

  The IN/NOT IN operators are represented in the proto as EQUAL/NOT_EQUAL with array RHS values, which get OR-ed together. This is a temporary representation until  native IN operator support is added to the query builder.

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/explore_page.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/explore_page.ts
@@ -531,13 +531,13 @@ export class ExplorePage implements m.ClassComponent<ExplorePageAttrs> {
     filterOperator?: 'AND' | 'OR',
   ) {
     // Normalize to array for uniform handling (single filter → [filter])
-    const filters = Array.isArray(filter) ? filter : [filter];
+    const filters: UIFilter[] = Array.isArray(filter) ? filter : [filter];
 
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/data_explorer.ts
stevegolton, line 310: You need to add 'in' (and maybe 'not in'?) to this to get the menu items to show up

PR #3946, opened 2025-12-02
exp: Inline editing of filters and aggregations

aMayzner:
Both aggregation and filters have modern inline editing, no more modals

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.scss
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.scss
@@ -412,7 +412,7 @@
 
   &:focus-within:not(:disabled) {
     border-color: var(--pf-color-primary);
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/widgets.scss
```diff
@@ -333,3 +333,169 @@
...
+  border-radius: 4px;
+  transition: border-color 0.15s ease-in-out;
+
+  &:hover:not(:disabled) {
+    border-color: var(--pf-color-text-secondary);
+  }
+
+  &:focus-within:not(:disabled) {
+    border-color: var(--pf-color-primary);
+    border-width: 2px;
```
stevegolton: If you add a 1px wide box shadow instead of making the border thicker, this avoids the layout shift when focused

PR #3952, opened 2025-12-02
ui: Replace SQLTable and PivotTable entirely with DataGrid

stevegolton:
This patch replaces a lot of the custom functionality in the SQL table (Open table... command), with DataGrid, removing a lot of duplicated code and functionality, and moves the functionality of loading the module definitions into the core and exposes an API on trace to open a new table explorer tab and access sql mode definitions.

Functionality:
- Opening tables defined by the SQL modules definitions.
- Adding columns from other tables based on the SQL module definitions.
- Opinionated tables are still present (from the org.chromium.OpenTableCommands plugin, which maybe disabled by default in the future).
- Sorting, column reordering, filtering, pivoting.
- Duplicate tab.
- Adding a debug track.
- Exporting data to TSV, markdown, JSON.
- Row count.

What's missing:
- Histograms and other charts (needs some careful thinking about how to integrate these nicely into the DataGrid).
- Data casting (required, will be done in an upcoming patch).
- Transformations (not sure if these are required).
- Special args links in the 'all args' columns, where clicking can add columns.
- Opening pivot tables in a new tab (but you can always duplicate the tab, so I think this is a non-issue).

### DIFF
--- a/src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.cc
+++ b/src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.cc
@@ -295,10 +295,10 @@ base::StatusOr<std::vector<std::string>> GetColumnNamesFromSelectStatement(
     if (col_name.empty()) {
       return base::ErrStatus("%s: column %u: name must not be empty", tag, i);
     }
-    if (!std::isalpha(col_name.front())) {
+    if (!std::isalpha(col_name.front()) && col_name.front() != '_') {
       return base::ErrStatus(
-          "%s: Column %u: name '%s' has to start with a letter.", tag, i,
-          col_name.c_str());
+          "%s: Column %u: name '%s' has to start with a letter or underscore.",
+          tag, i, col_name.c_str());
     }
     if (!sql_argument::IsValidName(base::StringView(col_name))) {
       return base::ErrStatus(
--- a/ui/src/components/aggregation.ts
+++ b/ui/src/components/aggregation.ts
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import m from 'mithril';
 import {ColorScheme} from '../base/color_scheme';
+import {Row, SqlValue} from '../trace_processor/query_result';
+
+export type AggregationCellRenderer = (value: SqlValue, row: Row) => m.Children;
 
 export interface ColumnDef {
   readonly columnId: string;
@@ -20,6 +24,11 @@ export interface ColumnDef {
   readonly formatHint?: string;
   readonly sum?: boolean;
   readonly sort?: 'ASC' | 'DESC';
+  readonly cellRenderer?: AggregationCellRenderer;
+  // Additional fields this column depends on for rendering.
+  // These fields will be included in queries and made available in the row
+  // parameter passed to cellRenderer.
+  readonly dependsOn?: readonly string[];
 }
 
 export interface BarChartData {
--- a/ui/src/components/aggregation_panel.ts
+++ b/ui/src/components/aggregation_panel.ts
@@ -58,8 +58,10 @@ export class AggregationPanel
           title: c.title,
           titleString: c.title,
           columnType: filterTypeForColumnDef(c.formatHint),
-          cellRenderer: getCellRenderer(c.formatHint, c.columnId),
+          cellRenderer:
+            c.cellRenderer ?? getCellRenderer(c.formatHint, c.columnId),
           cellFormatter: getValueFormatter(c.formatHint),
+          dependsOn: c.dependsOn,
         };
       }
       const schema: SchemaRegistry = {data: columnSchema};
@@ -83,8 +85,10 @@ export class AggregationPanel
           title: c.title,
           titleString: c.title,
           columnType: filterTypeForColumnDef(c.formatHint),
-          cellRenderer: getCellRenderer(c.formatHint, c.columnId),
+          cellRenderer:
+            c.cellRenderer ?? getCellRenderer(c.formatHint, c.columnId),
           cellFormatter: getValueFormatter(c.formatHint),
+          dependsOn: c.dependsOn,
         };
       }
       const schema: SchemaRegistry = {data: columnSchema};
--- a/ui/src/components/sql_table_converter.ts
+++ b/ui/src/components/sql_table_converter.ts
@@ -36,6 +36,7 @@ import {
 import {DurationWidget} from './widgets/duration';
 import {Anchor} from '../widgets/anchor';
 import {Icons} from '../base/semantic_icons';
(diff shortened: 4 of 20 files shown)


aMayzner (approved):
Great!

ui/src/components/details/sql_table_tab.ts
```diff
@@ -319,9 +300,7 @@ class SqlTableTab implements Tab {
   }
 
   getTitle(): string {
-    const rowCount = this.tableState.getTotalRowCount();
```
altimin, line 324: Showing the total number of rows is really important — for many user journeys (e.g. selecting "slice with the same name", adding a filter, etc) the number of rows it the first thing user looks at.

altimin:
It seems that `null` are no longer formatted:

Screenshot 2025-12-02 at 21 31 01

ui/src/components/widgets/sql/table/columns.ts
```diff
@@ -538,26 +541,14 @@ class ArgColumn implements TableColumn {
...
-        this.getRawColumn('real_value'),
-        tableManager,
-      );
+      return renderStandardCell(argValue['real_value'], tableManager);
     } else {
-      return renderStandardCell(
-        argValue['string_value'],
-        this.getRawColumn('string_value'),
-        tableManager,
-      );
```
altimin, line 611: Not passing SqlColumn to the `renderStandardCell` here breaks filtering by arguments:

`argValue` is a JSON containing all columns of the arg table, while we want to add a filter on the corresponding column of the args table (int_value / real_value / etc).

You can repro this by adding an arg_set_id[foo] column and trying to filter on it.

ui/src/components/details/sql_table_tab.ts
LalitMaganti: +1 can vouch for the row count being very very important.

stevegolton: This was a mistake. I have returned the row count to the tab title and on the right of the details panel.

ui/src/public/trace.ts
```diff
@@ -78,8 +80,28 @@ export interface Trace extends App {
   // code. These are on top of traceInfo.importErrors, which is a summary of
   // what TraceProcessor reports on the stats table at import time.
   get loadingErrors(): ReadonlyArray<string>;
+
   // Trace scoped disposables. Will be destroyed when the trace is unloaded.
   readonly trash: DisposableStack;
+
+  // Returns the SqlModules instance for accessing stdlib tables and modules.
+  // May return undefined if not yet initialized.
+  getSqlModules(): SqlModules | undefined;
```
LalitMaganti, line 88: So in the longer term, this function should really be on the engine. Should we move it there?

ui/src/public/trace.ts
```diff
@@ -78,8 +80,28 @@ export interface Trace extends App {
...
+
+  // Creates a new Table Explorer tab with the given configuration.
+  openTableExplorer(config: {
```
LalitMaganti: Would it make more sense to put this on the tab manager instead of here?

ui/src/public/trace.ts
stevegolton: Definitely, when it's loaded from the engine we should move it there. Moving it now would be ugly.

ui/src/public/trace.ts
stevegolton: I'm not sure about this. The tab manager deals with the tab framework, but the table explorer is a specific instance of a tab. In fact it doesn't even necessarily need to be implemented as a tab, it could open a new page for instance.

altimin:
I like the overall direction, but there are quite a few things missing — could we keep the old functionality around until we reach parity before removing it?

Specific major things that are still missing / have regressed:
- The overall experience around the pivoting has regressed considerably — the biggest issue, IMO is that "group by this column" replaces the list view — the ability to look both at the list view _and_ different pivots is a feature, which comes very handy in the data analysis.
- Ability to go from the selection to individual slices comprising that selection (incl. looking at their args, etc).
- Pivot table doesn't seem to be able to actually collapse / hide rows.

Some minor things I spotted:
- We should not hide column names too aggressively: e.g. `track` table has both `dimension_arg_set_id` and `source_arg_set_id`, which the user should be able to work with independently (e.g. look up arguments / print / etc).
- Cmd+Enter stopped working in the "select arg" menu (which was handy for adding multiple columns).

Also, transforms are very important (e.g. if you have a string, like you do in atrace, you often want to extract a part of it as an argument and continue working with it).

stevegolton:
@altimin Thanks for your review.

> I like the overall direction, but there are quite a few things missing — could we keep the old functionality around until we reach parity before removing it?

The main reasons for this PR are:
1. Remove duplicate and subtly different table implementations (e.g. the 'Slices' and 'Pivot Table' tabs when you make an area selection). Having multiple different table implementations with different UXs and feature sets makes the UI feel half baked.
2. Reduce the development overhead of having 3 table implementations (SQLTable, PivotTable, and DataGrid).

The goal here is to aim for a consistent feel and to reuse common code across all tables, even if it means losing some features temporarily. Then, once we have a more consistent and stable platform on which to work we can implement features much more quickly and they'll apply to all tables in the UI. This patch is already enormous, and I need to draw the line somewhere, so I've chosen to omit certain features, but with the intention of fixing some of them back up in future PRs.

> * The overall experience around the pivoting has regressed considerably — the biggest issue, IMO is that "group by this column" replaces the list view — the ability to look both at the list view _and_ different pivots is a feature, which comes very handy in the data analysis.

In table explorer, you can duplicate the tab if you need this functionality. In the area selection pivot table I don't think we ever had this functionality?

> * Ability to go from the selection to individual slices comprising that selection (incl. looking at their args, etc).

This works in Table Explorer but not in the various aggregations yet. This is absolutely next on the list of things to implement. It's more nuanced than it was in the previous pivot table implementation as we need to support linking to any arbitrary tracks, not only slice tracks.

> * Pivot table doesn't seem to be able to actually collapse / hide rows.

I am also planning on adding this soon. Collapsing and expanding groups while also supporting proper virtualization of rows is difficult. The old pivot table did not scale well at all as it didn't do proper virtualization at all. I'd rather go for stability now and introduce these features properly.
 
> Some minor things I spotted:
> 
> * We should not hide column names too aggressively: e.g. `track` table has both `dimension_arg_set_id` and `source_arg_set_id`, which the user should be able to work with independently (e.g. look up arguments / print / etc).

Good catch, I'll patch this before landing.

> * Cmd+Enter stopped working in the "select arg" menu (which was handy for adding multiple columns)

The menu stays open when clicking so you can add as many args as you like, or am I missing something here?

> Also, transforms are very important (e.g. if you have a string, like you do in atrace, you often want to extract a part of it as an argument and continue working with it).

If you find this feature helpful, I would be happy to accept a PR adding this as a feature to DataGrid.

PR #3955, opened 2025-12-02
ui: Nodegraph has contextMenu icon visible on hover

aMayzner:
Context menu icon visible on hover only will make it easier to focus on the + icon below the node, which is significantly more important

### DIFF
--- a/ui/src/assets/widgets/nodegraph.scss
+++ b/ui/src/assets/widgets/nodegraph.scss
@@ -85,6 +85,15 @@
   cursor: move;
   pointer-events: auto;
   width: 100%;
+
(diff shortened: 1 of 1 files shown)


ui/src/assets/widgets/nodegraph.scss
```diff
@@ -224,6 +224,15 @@
   right: 6px;
 }
 
+// Hide elements with this class, show on node hover
+.pf-show-on-hover {
```
stevegolton: I'd put this inside the pf-node so the rule looks like this:

```
.pf-node .pf-show-on-hover {...}
```

.pf-show-on-hover is quite vague and could collide with other elements.

PR #3956, opened 2025-12-02
Reapply "Revert "Add CUJ counters to stdlib. (#1992)"".

devianb:
The original change was reverted as counter metrics reported were incorrect in certain cases, due to different cuj_ids in the sysui jank table. The change is re-applied now, but using the android_jank_cujs base table from stdlib, instead of the android_sysui_jank_latency_cujs table.

Welcome to Perfetto!

For more details please see


LalitMaganti (approved):

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
```diff
@@ -12,69 +12,8 @@
...
-SELECT IFNULL(SUM(marker_name GLOB $callback_missed), 0)
-FROM cuj_marker_missed_callback
-WHERE
-  cuj_slice_name = $cuj_slice_name
-  AND ts >= $ts_min
-  AND ($ts_max IS NULL OR ts <= $ts_max)
-ORDER BY ts ASC
-LIMIT 1;
+INCLUDE PERFETTO MODULE android.cujs.cujs_base;
+INCLUDE PERFETTO MODULE android.cujs.sysui_cuj_counters;
```
elrakadm: Is the dependency on `sysui_cuj_counters` required ?

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
```diff
@@ -96,20 +35,20 @@ WITH cujs_ordered AS (
       ELSE MAX(ts, ts_end - 4000000)
     END AS ts_earliest_allowed_counter,
     LEAD(ts_end) OVER (PARTITION BY cuj_name ORDER BY ts_end ASC) AS ts_end_next_cuj
-  FROM android_jank_cuj
+  FROM android_jank_cujs
```
elrakadm: what is the difference between android_jank_cujs table and android_jank_cuj ? there are downstream perfettosql that depends on `android_jank_cuj` , should it switch and why ?

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
devianb: Yes, because the `_android_jank_cuj_counter_value` function is present in that module.

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
devianb: It is essentially the same table, just that `android_jank_cujs` has been moved to the stdlib. We want to move away from such tables that have been historically created outside the stdlib, and are being used at different places using the `RUN_METRIC()` import.

Eventually all downstream calls can be replaced with the `android_jank_cujs` stdlib table, and the old `android_jank_cuj` table can be safely deleted.

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
elrakadm: would the counters be only available for system ui cujs ? in that case `android_jank_cuj_counter_metrics`and `_android_jank_cuj_counter_value` should have `sysui` in their name. If not, then the module name is confusing.

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
elrakadm: the `android_jank_cuj` is used heavily downstream. Would it be better to remove the android_jank_cuj table and rename `android_jank_cujs` now so there is no migration effort ?

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
devianb: The counters are not specifc for sysui. Renamed to module to be more generic.

src/trace_processor/metrics/sql/android/jank/internal/counters.sql
devianb: As this new table `android_jank_cujs` is part of the stdlib now, there is a difference in the way it is imported in other files. The old table was imported like this:

`SELECT RUN_METRIC('android/jank/cujs.sql');`

Whereas the new table will be imported like this:

`INCLUDE PERFETTO MODULE android.cujs.cujs_base;`

Just renaming the table to `android_jank_cuj` won't save on the migration effort, as the import part will need to change as well.

devianb: Rebased the changes after submitting #4064 and #4080 .

elrakadm (approved):

PR #3970, opened 2025-12-03
Refactor `PerfettoCmd::ReportTraceToAndroidFramework`

KirillTim:
To support uploading multiple traces after Android reboot, we extract the
upload logic into a static method  `ReportTraceToAndroidFramework`.
It accepts input arguments, instead of relying on the internal `PerfettoCmd`
state. Each trace may have different `statsd_logging` config, so we also
pass it as an additional argument.
This also changes the error handling to return `base::Status` rather than
terminating the process.

This is a no-op refactoring for the current uploading logic.

Bug: b/382209797

Tested: atest CtsPerfettoReporterTestCases

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/add_columns_node.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/add_columns_node.ts
@@ -459,9 +459,6 @@ export class AddColumnsNode implements QueryNode {
   nextNodes: QueryNode[];
   readonly state: AddColumnsNodeState;
 
-  // Track the previous rightNode ID to detect when it actually changes
-  private previousRightNodeId?: string;
-
   constructor(state: AddColumnsNodeState) {
     this.nodeId = nextNodeId();
     this.state = state;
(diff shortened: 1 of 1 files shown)


src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -86,44 +87,67 @@ void PerfettoCmd::SaveTraceIntoIncidentOrCrash() {
...
-  const auto& cfg = trace_config_->android_report_config();
-  PERFETTO_CHECK(!cfg.reporter_service_package().empty());
-  PERFETTO_CHECK(!cfg.skip_report());
+base::Status PerfettoCmd::ReportTraceToAndroidFramework(
+    int trace_fd,
+    uint64_t trace_size,
+    const base::Uuid& uuid,
+    const std::string& unique_session_name,
+    const protos::gen::TraceConfig_AndroidReportConfig& report_config) {
+  if (report_config.reporter_service_class().empty() ||
```
primiano, line 105: check with Lalit if this can ever be empty or not (I think this code is fine and it was just an omission)

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -86,44 +87,67 @@ void PerfettoCmd::SaveTraceIntoIncidentOrCrash() {
...
-  PERFETTO_CHECK(report_to_android_framework_);
-  PERFETTO_CHECK(trace_out_stream_);
-
-  const auto& cfg = trace_config_->android_report_config();
-  PERFETTO_CHECK(!cfg.reporter_service_package().empty());
-  PERFETTO_CHECK(!cfg.skip_report());
+base::Status PerfettoCmd::ReportTraceToAndroidFramework(
+    int trace_fd,
+    uint64_t trace_size,
+    const base::Uuid& uuid,
```
primiano, line 95: note that LogUploadEvent relies on the uuid_ var so either you overwrite uuid_ in your loop, or you make sure that nobody depends on that.

The cleanest thing is making this function a static function and here you call directly android_stats::MaybeLogUploadEvent.

BEause if you recached this point you know you want to upload

but please check in which case we set statsd_logging_

KirillTim: Seems like `TraceConfig` may explicitly disable statsd reporting, even when `--upload` flag is set, see here: https://github.com/google/perfetto/blob/681c9d0824c7d5601c34c0e45e1158b608da178e/src/perfetto_cmd/perfetto_cmd.cc#L729

So I've add a `bool statsd_logging` argument to the method, to be able to pass the config for each trace we report separately.

src/perfetto_cmd/perfetto_cmd_android.cc
KirillTim: Yes, we need to check both `package` and `class`. They are used here: https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:frameworks/base/services/core/java/com/android/server/tracing/TracingServiceProxy.java;l=178;drc=1952ddfd9b56cdcac1dd2540447ef5a68e9cfee5 and `ComponentName` expects both to be `@NonNull`.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -86,44 +88,75 @@ void PerfettoCmd::SaveTraceIntoIncidentOrCrash() {
...
+                 trace_size);
+  }
+  log_upload_event_fn(PerfettoStatsdAtom::kCmdFwReportHandoff);
+  return base::OkStatus();
+}
+
+void PerfettoCmd::ReportTraceToAndroidFrameworkOrCrash() {
+  PERFETTO_CHECK(trace_out_stream_);
+  int trace_fd = fileno(*trace_out_stream_);
+  base::Uuid uuid(uuid_);
```
primiano, line 153: why this copy? can we not just pass uuid_ as argument below?

KirillTim: The `uuid_` is a string (https://github.com/google/perfetto/blob/c38e6dd325fd1751c46c4e228894290df0b18e67/src/perfetto_cmd/perfetto_cmd.h#L192), so we create an object of type `base::Uuid` from string here, and pass that object to a function.

PR #3976, opened 2025-12-04
tp: Add experimental slices creation in SQ

aMayzner:
EXPERIMENTAL

Use structured query to create slices (intervals with ts and dur)

### DIFF
--- a/protos/perfetto/perfetto_sql/structured_query.proto
+++ b/protos/perfetto/perfetto_sql/structured_query.proto
@@ -368,10 +368,12 @@ message PerfettoSqlStructuredQuery {
     // The query containing end timestamps. Required.
     optional PerfettoSqlStructuredQuery ends_query = 2;
 
-    // The column name in starts_query containing the timestamp. Required.
+    // The column name in starts_query containing the timestamp.
+    // Optional, defaults to "ts".
     optional string starts_ts_column = 3;
 
-    // The column name in ends_query containing the timestamp. Required.
+    // The column name in ends_query containing the timestamp.
+    // Optional, defaults to "ts".
     optional string ends_ts_column = 4;
   }
 
--- a/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
+++ b/src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
@@ -933,21 +933,22 @@ base::StatusOr<std::string> GeneratorImpl::CreateSlices(
   if (!create_slices.has_ends_query()) {
     return base::ErrStatus("CreateSlices must specify an ends_query");
   }
-  if (!create_slices.has_starts_ts_column()) {
-    return base::ErrStatus("CreateSlices must specify starts_ts_column");
-  }
-  if (!create_slices.has_ends_ts_column()) {
-    return base::ErrStatus("CreateSlices must specify ends_ts_column");
-  }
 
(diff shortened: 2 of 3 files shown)


protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -330,6 +330,51 @@ message PerfettoSqlStructuredQuery {
...
+  //  - Ends with no matching start are ignored
+  //  - If multiple starts occur before an end, each is paired with that end
+  message ExperimentalCreateSlices {
+    // The query containing start timestamps. Required.
+    optional PerfettoSqlStructuredQuery starts_query = 1;
+
+    // The query containing end timestamps. Required.
+    optional PerfettoSqlStructuredQuery ends_query = 2;
+
+    // The column name in starts_query containing the timestamp. Required.
```
LalitMaganti: Why not optional? We can just have default "ts" which is very sensible.

protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -330,6 +330,51 @@ message PerfettoSqlStructuredQuery {
...
+  // Schema:
+  //  1) starts_query must have the column specified in starts_ts_column
+  //  2) ends_query must have the column specified in ends_ts_column
+  //  3) Both timestamp columns must contain integer timestamps
+  //  4) Output schema: (ts, dur) where:
+  //     - ts is the start timestamp
+  //     - dur is the duration (end - start)
+  //
+  // Matching Behavior:
+  //  - Each start is paired with the first end that comes after it
```
LalitMaganti, line 360: We also need the notion of "partitioning" here. Fine to do in a followup, just letting you know.

protos/perfetto/perfetto_sql/structured_query.proto
aMayzner: I want to remove as much guessing as possible from the generation. If we allow guessing "ts" then why not "timestamp". Why not {any string}_ts? Let's add it as this. If we have complaints we can always relax it

protos/perfetto/perfetto_sql/structured_query.proto
aMayzner: Yes, I wanted to do it in followup

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -917,6 +924,56 @@ base::StatusOr<std::string> GeneratorImpl::AddColumns(
...
+    return base::ErrStatus("ends_ts_column cannot be empty");
+  }
+
+  // Generate nested sources
+  std::string starts_table = NestedSource(create_slices.starts_query());
+  std::string ends_table = NestedSource(create_slices.ends_query());
+
+  // Build the SQL to create slices
+  // For each start, find the first end that comes after it
+  std::string sql = "(WITH starts AS (SELECT * FROM " + starts_table +
```
LalitMaganti: This can be done a lot nicer with base::StackString + multi-line strings.

protos/perfetto/perfetto_sql/structured_query.proto
LalitMaganti: Stand by the original comment. 99% of cases, it will be ts.

LalitMaganti (approved):

PR #3977, opened 2025-12-04
stdlib: Make frame_id an ID

aMayzner:
Frame ID was treated just as long, which made it difficult to use it for joining with other tables. As it is unique and sorted, we make it an ID column on android_frames, and make usages type by JOINID(android_frames.frame_id)

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
@@ -28,7 +28,7 @@ RETURNS TABLE (
   -- Frame slice.
   id ID(slice.id),
   -- Parsed frame id.
-  frame_id LONG,
+  frame_id JOINID(android_frames.frame_id),
   -- Utid.
   utid JOINID(thread.id),
   -- Upid.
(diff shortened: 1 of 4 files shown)


src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
```diff
@@ -274,7 +274,7 @@ WHERE
 -- information across different layers for a given frame_id in a given process.
 CREATE PERFETTO TABLE android_frames (
   -- Frame id.
-  frame_id LONG,
+  frame_id ID,
```
LalitMaganti, line 277: My main ask here was to add an "id" column here so it could be interval intersected. The rest of the CL is nice but this is the main thing I need.

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
```diff
@@ -274,7 +274,7 @@ WHERE
 -- information across different layers for a given frame_id in a given process.
 CREATE PERFETTO TABLE android_frames (
   -- Frame id.
```
LalitMaganti, line 276: Can we add a note here explaining what the "frame id" is: where does it come from (it's a Android level concept which ends up in trace, not something we create on our side).

src/trace_processor/perfetto_sql/stdlib/android/frames/timeline.sql
LalitMaganti, line 277: Actually frame_id is *not* an id: i was mistaken because we're grouping by id *and* upid. So it's possible to have the same frame id in multiple processes. So I don't think you can do much of this CL :(

LalitMaganti: In that case, maybe we should have a separate "uframeid" concept which uniquely identifies the frame?

PR #3987, opened 2025-12-04
Speed up the CpuPerUid plugin by 100x.

rzuklie:
Precompute the total millis per track, and the aggregates "System" and "Apps" tracks so that the plugin only needs to organize the tracks into groups. On my machine, this reduces the plugin load time from 2 seconds to 20 milliseconds.

UI wise, the only change is that the tracks now use the auto-rate feature, which will show in ms per second, rather than %.

### DIFF
--- a/src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
+++ b/src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
@@ -46,7 +46,7 @@ uint64_t MakeKey(uint32_t uid, uint32_t cluster) {
 static constexpr auto kCpuPerUidBlueprint = tracks::CounterBlueprint(
     "android_cpu_per_uid",
     tracks::StaticUnitBlueprint("ms"),
-    tracks::DimensionBlueprints(tracks::UintDimensionBlueprint("uid"),
+    tracks::DimensionBlueprints(tracks::kUidDimensionBlueprint,
                                 tracks::UintDimensionBlueprint("cluster")),
     tracks::FnNameBlueprint([](uint32_t uid, uint32_t cluster) {
       return base::StackString<128>("CPU for UID %d CL%d", uid, cluster);
@@ -55,6 +55,8 @@ static constexpr auto kCpuPerUidBlueprint = tracks::CounterBlueprint(
 static constexpr auto kCpuTotalsBlueprint = tracks::CounterBlueprint(
     "android_cpu_per_uid_totals",
     tracks::StaticUnitBlueprint("ms"),
+    // TODO(lalitm): allow FnNameBlueprint and StringIdDimensionBlueprint to
+    // work together.
     tracks::DimensionBlueprints(tracks::StringDimensionBlueprint("type"),
                                 tracks::UintDimensionBlueprint("cluster")),
     tracks::FnNameBlueprint([](base::StringView type, uint32_t cluster) {


rzuklie:
This is certainly more work than the old SQL solution. However, if we can do the same for kernel wakelocks and entity state residency, we would reduce trace loading times by 6-8 seconds (and I'm on an M4 mac, my old mac was about 5x slower).

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
```diff
@@ -38,6 +38,30 @@
...
 namespace perfetto::trace_processor {
+namespace {
+uint64_t MakeKey(uint32_t uid, uint32_t cluster) {
+  return ((uint64_t(uid)) << 32) | cluster;
+}
+
+static constexpr auto kCpuPerUidBlueprint = tracks::CounterBlueprint(
+    "android_cpu_per_uid",
+    tracks::StaticUnitBlueprint("ms"),
+    tracks::DimensionBlueprints(tracks::UintDimensionBlueprint("uid"),
```
LalitMaganti: Can you make the uid a common dimension similar to utid/upid?

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
```diff
@@ -38,6 +38,30 @@
...
+                                tracks::UintDimensionBlueprint("cluster")),
+    tracks::FnNameBlueprint([](uint32_t uid, uint32_t cluster) {
+      return base::StackString<128>("CPU for UID %d CL%d", uid, cluster);
+    }));
+
+static constexpr auto kCpuTotalsBlueprint = tracks::CounterBlueprint(
+    "android_cpu_per_uid_totals",
+    tracks::StaticUnitBlueprint("ms"),
+    tracks::DimensionBlueprints(tracks::StringDimensionBlueprint("type"),
```
LalitMaganti, line 60: Can you please make this a stringid dimension?

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
```diff
@@ -85,15 +109,22 @@ void AndroidCpuPerUidModule::ParseTracePacketData(
...
-
     cluster++;
     if (cluster >= state->cluster_count) {
       cluster = 0;
       uid_it++;
     }
   }
 
+  for (auto it = system_totals_.GetIterator(); it; ++it) {
+    UpdateTotals(ts, "System", it.key(), it.value());
```
LalitMaganti, line 124: This should be interned in the constructor and you should use the stringid instead.

LalitMaganti (approved):

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
rzuklie: Done

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
rzuklie: Spoke offline, StringId dimensions don't play well with FnNameBlueprint, so added a todo instead.

src/trace_processor/importers/proto/android_cpu_per_uid_module.cc
rzuklie: Same as above.

src/trace_processor/tables/android_tables.py
```diff
@@ -58,6 +59,18 @@
     ),
 )
 
+ANDROID_CPU_PER_UID_TRACK_TABLE = Table(
+    python_module=__file__,
+    class_name="AndroidCpuPerUidTrackTable",
+    sql_name="__intrinsic_android_cpu_per_uid_track",
+    columns=[
+        C("track_id", CppTableId(TRACK_TABLE), cpp_access=CppAccess.READ),
```
LalitMaganti, line 67: This should be marked as a sorted flag as well I think.

rzuklie: Does that flag do the sorting, or do they need to be inserted sorted? Right now they're not sorted. These are emit at the end by iterating a FlatHashMap, which IIUC is iterated in arbitrary order.

LalitMaganti: Ahhhh right you're absolutely correct. Ignore me then lgtm.

PR #3990, opened 2025-12-04
Add PerfettoSQL intrinsic for package name lookup.

rzuklie:
The intrinsic finds an appropriate package name to display for a given UID. It handles system UIDs, package name preference, and multi-user scenarios.

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/package_lookup.h
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/package_lookup.h
@@ -29,6 +29,8 @@ struct PackageLookup : public sqlite::Function<PackageLookup> {
   struct Context {
     Context(TraceStorage* s)
         : storage(s),
+          // TODO(rzuklie): the uid column is not indexed (since it is not
+          // unique), consider finding a more efficient way to scan the table.
           package_list_cursor(s->mutable_package_list_table()->CreateCursor(
               {dataframe::FilterSpec{
                   tables::PackageListTable::ColumnIndex::uid,


src/trace_processor/perfetto_sql/stdlib/android/package_lookup.sql
```diff
@@ -0,0 +1,191 @@
...
+    THEN 'ISOLATED_UID'
+    ELSE 'uid=' || $uid
+  END;
+
+-- Resolves the Android package name using the `uid` column from the source
+-- table. The output contains all input columns, plus a `package_name` column.
+CREATE PERFETTO MACRO android_package_lookup(
+    -- A table or subquery containing a `uid` column for package lookup.
+    src TableOrSubquery
+)
```
LalitMaganti: Any reason this is not a table you can join against?

src/trace_processor/perfetto_sql/stdlib/android/package_lookup.sql
```diff
@@ -0,0 +1,191 @@
...
+RETURNS TableOrSubquery AS
+(
+  SELECT
+    base.*,
+    _resolve_package_name(pkg.package_name, base.uid % 100000, base.uid) AS package_name
+  FROM $src AS base
+  LEFT JOIN _package_lookup_source AS pkg
+    ON (
```
LalitMaganti: This join is going to be really inefficient as you're not joining on a constant but instead an expression.

LalitMaganti (changes requested):

src/trace_processor/perfetto_sql/stdlib/android/package_lookup.sql
rzuklie: It's a left join with post-processing. This could be split up into a table _and_ a function for post processing, but that's more work for clients and more error prone (clients need to join on TheirTable.uid % 100000, if they forget they accidentally lose all multi-user results).

src/trace_processor/perfetto_sql/stdlib/android/package_lookup.sql
rzuklie: I'm not sure how to avoid this? Something somewhere needs to do the uid -> pkgid (mod 100k) calculation on the table the _caller passes in_. Does moving this to a non-materialized CTE change this? I'd have expected it to be functionally equivalent.

LalitMaganti: Given the above and this problem I think it might make a lot of sense to make this a C++ function instead honestly.

src/trace_processor/perfetto_sql/intrinsics/functions/package_lookup.h
```diff
@@ -0,0 +1,50 @@
...
+struct PackageLookup : public sqlite::Function<PackageLookup> {
+  static constexpr char kName[] = "package_lookup";
+  static constexpr int kArgCount = 1;
+
+  struct Context {
+    Context(TraceStorage* s)
+        : storage(s),
+          package_list_cursor(s->mutable_package_list_table()->CreateCursor(
+              {dataframe::FilterSpec{
+                  tables::PackageListTable::ColumnIndex::uid,
```
LalitMaganti, line 36: Because uid is not indexed, this *might* be a full linear scan. There might be an optimization kicking in if the uids are small but this might also break down in the presence of multi-user.

Maybe not the end of the world but might be some gotcha. Maybe leave a comment to improve this?

rzuklie: Added the todo. I'm not sure if it's a linear scan, but it's a bit faster than either the left join or subquery solutions.

src/trace_processor/perfetto_sql/stdlib/android/package_lookup.sql
rzuklie: Done

src/trace_processor/perfetto_sql/stdlib/android/package_lookup.sql
rzuklie: Moved to intrinsic function.

LalitMaganti (approved):

PR #3992, opened 2025-12-04
exp: Better SQL editor

aMayzner:
Fixed an issue where the SQL source node editor required clicking precisely on "row 1" to gain focus. The rest of the editor area was not clickable.

  **Root cause**: The editor was wrapped in a div with inline styles that prevented clicks from reaching the CodeMirror editor properly.

  **Solution**: Refactored the SQL source editor to match the query page implementation:
  - Created a `SqlEditor` wrapper component with proper lifecycle management
  - Used refs and `oncreate` hook to directly access and style the editor DOM element
  - Removed the problematic wrapper div with inline styles
  - Added `ResizeHandle` for consistent resize behavior


stevegolton (approved):
NIce

PR #3999, opened 2025-12-05
ui: nodegraph: Add labels

aMayzner:
Add support for text labels in the NodeGraph widget. Labels are movable, resizable text boxes that can be placed on the canvas to annotate it. They pan and zoom like the nodes.

  ### Changes
  - Add `Label` interface with id, position, width, and text
  - Implement label rendering with drag, resize, and edit functionality
  - Add label-specific event handlers (move, resize, text change, select, remove)
  - Add label selection state (separate from node selection)
  - Support keyboard shortcuts (Delete/Backspace to remove, Escape to deselect/exit edit mode)
  - Add SCSS styling for labels with hover states, resize handles, and delete buttons
  - Update NodeGraph demo with example labels

  ### Features
  - Click to edit, click outside or Ctrl+Enter to finish editing
  - Drag labels by clicking anywhere on the label body
  - Resize labels using the handle on the right edge
  - Delete via button or keyboard shortcuts
  - Auto-focus and select text when entering edit mode

### DIFF
--- a/ui/src/plugins/dev.perfetto.WidgetsPage/demos/nodegraph_demo.ts
+++ b/ui/src/plugins/dev.perfetto.WidgetsPage/demos/nodegraph_demo.ts
@@ -20,6 +20,7 @@ import {Checkbox} from '../../../widgets/checkbox';
 import {MenuItem, PopupMenu} from '../../../widgets/menu';
 import {
   Connection,
+  Label,
   Node,
   NodeGraph,
   NodeGraphApi,
@@ -90,6 +91,7 @@ type NodeData =
 interface NodeGraphStore {
   readonly nodes: Map<string, NodeData>;
   readonly connections: Connection[];
+  readonly labels: Label[];
   readonly invalidNodes: Set<string>; // Track which nodes are marked as invalid
 }
 
@@ -435,6 +437,30 @@ export function NodeGraphDemo(): m.Component<NodeGraphDemoAttrs> {
   let store: NodeGraphStore = {
     nodes: new Map([[initialId, createTableNode(initialId, 150, 100)]]),
     connections: [],
+    labels: [
+      {
+        id: uuidv4(),
+        x: 400,
+        y: 100,
+        width: 180,
+        content: m('.pf-simple-label-text', 'Simple text label'),
+      },
+      {
+        id: uuidv4(),
+        x: 400,
+        y: 200,
+        width: 180,
+        content: m(
+          '.pf-simple-label-button',
+          m(Button, {
+            label: 'Click me!',
+            onclick: () => {
+              console.log('Label button clicked!');
+            },
+          }),
+        ),
+      },
+    ],
     invalidNodes: new Set<string>(),
   };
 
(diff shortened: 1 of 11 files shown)


ui/src/plugins/dev.perfetto.WidgetsPage/demos/nodegraph_demo.ts
```diff
@@ -1235,6 +1269,60 @@ export function NodeGraphDemo(): m.Component<NodeGraphDemoAttrs> {
...
+            if (label) {
+              label.text = text;
+              console.log(`onLabelTextChange: ${labelId} to "${text}"`);
+            }
+          });
+        },
+        onLabelSelect: (labelId: string) => {
+          selectedLabelIds.clear();
+          selectedLabelIds.add(labelId);
+          m.redraw();
```
stevegolton: Shouldn't need all the redraws

ui/src/plugins/dev.perfetto.WidgetsPage/demos/nodegraph_demo.ts
```diff
@@ -1235,6 +1269,60 @@ export function NodeGraphDemo(): m.Component<NodeGraphDemoAttrs> {
...
+          updateStore((draft) => {
+            const label = draft.labels.find((l) => l.id === labelId);
+            if (label) {
+              label.text = text;
+              console.log(`onLabelTextChange: ${labelId} to "${text}"`);
+            }
+          });
+        },
+        onLabelSelect: (labelId: string) => {
+          selectedLabelIds.clear();
```
stevegolton: I think we should have just one selection list which incorporates nodes and labels. It feels weird that I can select a node at the same time as a label.

Clicking on a label doesn't necessarily need to 'select' it (or rather it doesn't need to deselect any currently selected node). What is selection even for for labels? It might be useful to multi-select labels + nodes for instance if we wanted to move a bunch of them around.

ui/src/plugins/dev.perfetto.WidgetsPage/demos/nodegraph_demo.ts
```diff
@@ -1235,6 +1269,60 @@ export function NodeGraphDemo(): m.Component<NodeGraphDemoAttrs> {
             }
           });
         },
+        labels: store.labels,
+        selectedLabelIds: selectedLabelIds,
+        onLabelMove: (labelId: string, x: number, y: number) => {
```
stevegolton, line 1279: Looks like onLabelMove and onLabelResize are being called on every mouse move event - this will cause issues with undo/redo and is generally undesirable. Maybe just call these callbacks on mouseup

stevegolton (approved):

stevegolton (changes requested):
See comments

ui/src/plugins/dev.perfetto.WidgetsPage/demos/nodegraph_demo.ts
```diff
@@ -435,6 +437,30 @@ export function NodeGraphDemo(): m.Component<NodeGraphDemoAttrs> {
...
+      },
+      {
+        id: uuidv4(),
+        x: 400,
+        y: 200,
+        width: 180,
+        content: m(
+          '.pf-simple-label-button',
+          m(Button, {
+            label: 'Click me!',
```
stevegolton, line 456: Looks like undoing in the demo code doesn't work well for labels presumably because this data model is controlled by immer which doesn't support vnodes. But don't worry too much about this as it's only the demo

PR #4005, opened 2025-12-05
perfetto_cmd: don't delete not empty `write_into_file` trace on error.

KirillTim:
This bug happens because the `bytes_written_` field was not updated
when the `perfetto_cmd` exits with an error.
To prevent such bugs in the future, we extract the logic to a separate
method that should be called to get the the count of bytes written.

This function is called only at the end, when all the data is written so there
is no need to do something smarter than `fseek`+`ftell`.


src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1187,10 +1187,11 @@ void PerfettoCmd::ReadbackTraceDataAndQuit(const std::string& error) {
     // be marked as "E" in the event log. Hence why LOG and not ELOG here.
     PERFETTO_LOG("Service error: %s", error.c_str());
 
+    uint64_t bytes_written = GetBytesWritten();
     // In case of errors don't leave a partial file around. This happens
     // frequently in the case of --save-for-bugreport if there is no eligible
     // trace. See also b/279753347 .
-    if (bytes_written_ == 0 && !trace_out_path_.empty() &&
+    if (bytes_written == 0 && !trace_out_path_.empty() &&
```
LalitMaganti, line 1194: I'm missing something: doesn't the implementation of `GetBytesWritten` still mean zero is returned for write_into_file traces and this will be deleted?

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1282,6 +1277,20 @@ bool PerfettoCmd::OpenOutputFile() {
   return true;
 }
 
+uint64_t PerfettoCmd::GetBytesWritten() {
+  if (!trace_out_stream_)
```
LalitMaganti, line 1281: Isn't this the codepath which is taken in the case of traced written write_into_file traces? If so, will we still not try to deelte it?

src/perfetto_cmd/perfetto_cmd.cc
KirillTim: The trick is that if the `write_into_file` session is started with the `output_path` in `trace_config`, the `trace_out_path_` is empty and we ignore the `if` branch.

src/perfetto_cmd/perfetto_cmd.cc
KirillTim: we will not, because if the traced opened the file, the `trace_out_path_` is empty and we are fine.
We'd rather unify the behaviour somehow, though.

src/perfetto_cmd/perfetto_cmd.cc
KirillTim: The state is:
if `trace_out_path_` is not empty, then:
1) `trace_out_stream_` is set
2) `GetBytesWritten()` returns the real count of bytes written

Note, that on Android `trace_out_stream_` can point to the unlinked file, that implies `trace_out_path_` is empty.

PR #4006, opened 2025-12-05
python: Update deprecated arg

sashwinbalaji:
[numpy] Replace `interpolation=` with `method=` on numpy's `percentile`, `quantile`, `nanpercentile` and `nanquantile` APIs.

The `interpolation` keyword argument is removed in NumPy 2.4.


LalitMaganti:
Did you make sure we've bumped our deps to match this?

sashwinbalaji:
we currently don't have specific one in requirements.txt and this feature is available since early 2022 v1.22 https://numpy.org/doc/stable/reference/generated/numpy.percentile.html

image

PR #4023, opened 2025-12-06
Add support for print_args(arg_set_id) column to the SQLTable

altimin:
Add support for a new column type, which prints the entire arg set at the same time
and allows the user to select individual columns for further analysis. This is convenient
to be able to see the entire arg set at a glance rather than selecting individual columns manually.

Refactor TableManager interface, splitting it into more specific RenderCellContext
and ListColumnsContext and integrating the necessary bits directly into SelectColumnMenu.

Screenshot 2025-12-06 at 22 26 56

### DIFF
--- a/ui/src/components/widgets/sql/pivot_table/pivot_table.ts
+++ b/ui/src/components/widgets/sql/pivot_table/pivot_table.ts
@@ -274,16 +274,14 @@ export class PivotTable implements m.ClassComponent<PivotTableAttrs> {
             key: tableColumnId(column),
             column,
           })),
-          manager: {
-            filters: state.filters,
-            trace: state.trace,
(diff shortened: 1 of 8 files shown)


ui/src/components/widgets/sql/pivot_table/pivot_table.ts
```diff
@@ -274,16 +274,14 @@ export class PivotTable implements m.ClassComponent<PivotTableAttrs> {
...
-            filters: state.filters,
-            trace: state.trace,
-            getSqlQuery: (columns: {[key: string]: SqlColumn}) =>
-              buildSqlQuery({
-                table: state.table.name,
-                columns,
-                filters: state.filters.get(),
-              }),
-          },
+          filters: state.filters,
```
stevegolton, line 277: Nice!

PR #4025, opened 2025-12-07
sql table: refactor column creation

altimin:
This patch refactors column creation in the SqlTable: instead of exposing
the implementation of different types of columns, the public interface now
is createTableColumn(Trace, SqlColumn, PerfettoSqlType) and the specific
implementations are just an implementation details. To facilitate this,
createTableColumn is moved into columns.ts and most of the existing
implementations in columns.ts stop being exported.

Also refactor various *IdColumns to have a single IdColumn implementation
and make various id columns (slice, sched, thread_state, process, thread)
just instantiate the correct type.

These two changes are a step towards getting rid of hardcoded table definitions
and start relying on the exported stdlib definitions instead.

### DIFF
--- a/ui/src/components/widgets/sql/table/columns.ts
+++ b/ui/src/components/widgets/sql/table/columns.ts
@@ -619,9 +619,7 @@ export class ArgSetIdColumn implements TableColumn {
   }
 
   initialColumns() {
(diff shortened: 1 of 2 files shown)


stevegolton:
Looks like the PR description needs updating now?

PR #4039, opened 2025-12-08
Add plugin for Wear long battery traces

kvaleriano:
This plugin adds tracks for Wear CUJs and raw battery gauge based on statsd atoms. Also setting the plugin as default for ease of use in investigations.

On local runs the plugin startup takes about ~1ms on traces without the data sources and ~250ms on traces with them.

### DIFF
--- a/ui/src/plugins/com.android.WearLongBatteryTracing/index.ts
+++ b/ui/src/plugins/com.android.WearLongBatteryTracing/index.ts
@@ -1,4 +1,4 @@
-// Copyright (C) 2023 The Android Open Source Project
+// Copyright (C) 2025 The Android Open Source Project
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -215,16 +215,11 @@ export default class WearLongBatteryTracingPlugin implements PerfettoPlugin {
     const result = await ctx.engine.query(
       `SELECT int_value FROM metadata WHERE name = 'statsd_triggering_subscription_id'`,
     );
-    const it = result.iter({int_value: LONG});
-    let validTrace = false;
-    for (; it.valid(); it.next()) {
-      // Verify the trace is a Wear trace
-      if (!VALID_SUBSCRIPTION_IDS.includes(it.int_value)) {
-        return;
-      }
-      validTrace = true;
+    const row = result.maybeFirstRow({int_value: LONG});
+    if (!row) {
+      return;
     }
-    if (!validTrace) {
+    if (!VALID_SUBSCRIPTION_IDS.includes(row.int_value)) {
       return;
     }
 


ui/src/plugins/com.android.WearLongBatteryTracing/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * This hook is called as the trace is loading. At this point the trace is
+   * loaded into trace processor and it's ready to process queries. This hook
+   * should be used for adding tracks and commands that depend on the trace.
+   *
+   * It should not be used for finding tracks from other plugins as there is no
+   * guarantee those tracks will have been added yet.
+   */
+  async onTraceLoad(ctx: Trace): Promise<void> {
+    const atoms = await this.findAtoms(ctx.engine);
```
LalitMaganti, line 226: Right now I don't see any restriction on which traces this runs on. Can we please restrict this to only Wear traces?

LalitMaganti: You can add a restriction on unique_session_name or device name or form factor (not sure if that exists) or similar.

kvaleriano: Added filter on statsd_triggering_subscription_id. Tried looking for form factor initially but couldn't find it in the trace's metadata.

LalitMaganti:
Steve should review this not me

ui/src/plugins/com.android.WearLongBatteryTracing/index.ts
```diff
@@ -0,0 +1,236 @@
+// Copyright (C) 2023 The Android Open Source Project
```
stevegolton: Nit: 2025

ui/src/plugins/com.android.WearLongBatteryTracing/index.ts
```diff
@@ -0,0 +1,236 @@
...
+   * should be used for adding tracks and commands that depend on the trace.
+   *
+   * It should not be used for finding tracks from other plugins as there is no
+   * guarantee those tracks will have been added yet.
+   */
+  async onTraceLoad(ctx: Trace): Promise<void> {
+    const result = await ctx.engine.query(
+      `SELECT int_value FROM metadata WHERE name = 'statsd_triggering_subscription_id'`,
+    );
+    const it = result.iter({int_value: LONG});
```
stevegolton: If I'm right in thinking that you're only ever expecting on instance of `statsd_triggering_subscription_id`, you can probably clean up this code by using:

```ts
const row = result.maybeFirstRow({int_value: LONG});
if (!row) {
  return;
}
if (!VALID_SUBSCRIPTION_IDS.includes(it.int_value)) {
  return;
}
```

... which will return undefined if there are 0 rows and 1 row if there are 1 or more.

ui/src/plugins/com.android.WearLongBatteryTracing/index.ts
kvaleriano: Updated

ui/src/plugins/com.android.WearLongBatteryTracing/index.ts
kvaleriano: Yes, you're right, nice suggestion! Updated

PR #4042, opened 2025-12-08
Associate symbolization candidates with process and move to stdlib

ilkos:
Bug: 466405072

### DIFF
--- a/src/trace_processor/metrics/sql/android/unsymbolized_frames.sql
+++ b/src/trace_processor/metrics/sql/android/unsymbolized_frames.sql
@@ -36,7 +36,7 @@ SELECT UnsymbolizedFrames_Frame(
 ) AS frame_proto
 FROM (
   SELECT DISTINCT module, build_id, rel_pc, COALESCE(breakpad_module_id, build_id) AS google_lookup_id
-  FROM stacks_symbolization_candidates
+  FROM _stacks_symbolization_candidates
 );
 
 DROP VIEW IF EXISTS unsymbolized_frames_output;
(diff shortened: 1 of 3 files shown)


src/trace_processor/metrics/sql/android/unsymbolized_frames.sql
```diff
@@ -14,58 +14,30 @@
...
--- symbols indexed by "breadkpad module ID". Breakpad module ID can be derived
--- with the following formula:
---      base::StrCat({module_id->substr(6, 2), module_id->substr(4, 2),
---                    module_id->substr(2, 2), module_id->substr(0, 2),
---                    module_id->substr(10, 2), module_id->substr(8, 2),
---                    module_id->substr(14, 2), module_id->substr(12, 2),
---                    module_id->substr(16, 16), "0"});
--- See also https://source.chromium.org/chromium/chromium/src/+/main:services/tracing/public/cpp/stack_sampling/tracing_sampler_profiler.cc;l=603;drc=cba00174ca338153b9c4f0c31ddbabaac7dd38c7
--- Note that in SQL SUBSTR() indexes are 1-based, not 0 based.
+INCLUDE PERFETTO MODULE stacks.symbolization_candidates;
```
LalitMaganti, line 17: This is a problem: we're going to have circular dependencies if we do things like this b

ilkos: This is between legacy metric and new stdlib module, so I'd expect it's more or less safe. Making it private for now as discussed

PR #4043, opened 2025-12-08
perfetto_cmd: add --txt-lossy that skips unknown fields

rsavitski:
Protobufs are a good choice for configuration due to having well-defined semantics for unrecognised fields. However it's easier for humans to deal with the textproto represenation. We currently offer the --txt flag in perfetto_cmd so that it performs the serialisation (instead of having the user figure out how to serialise the protobuf), however that defaults to rejecting any unrecognised field.

This patch adds --txt-lossy as an alternative, which serialises the textproto while skipping unrecognised fields and enum values. This helps run new configs on older builds of perfetto, since most of our config changes are already shaped to support backwards compatibility.

Pre-serialised configs continue to be preferred for programmatic/production use.

Notes:
* It'd be nice to still print which fields got skipped, but that required quite a bit of plumbing back to the original caller. But I can be convinced that the value is worth the boilerplate.
* I wonder if we should start accepting numerical indices for enums (instead of only LABEL_NAME), I believe that's accepted by proto3 serialisers.


src/protozero/text_to_proto/text_to_proto.cc
```diff
@@ -438,12 +453,23 @@ class ParserDelegate {
...
         FindFieldByName(key, value,
                         {
                             FieldDescriptorProto::TYPE_MESSAGE,
                         });
     if (!field) {
+      // If unknown fields are allowed, push a skip context and continue
+      if (allow_unknown_fields_) {
+        ctx_.push(ParserDelegateContext{nullptr, nullptr, {}, true});
+        return true;
+      }
```
LalitMaganti, line 492: This could do with some tests

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -425,6 +432,12 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
       continue;
     }
 
+    if (option == OPT_PBTXT_CONFIG_LOSSY) {
```
LalitMaganti, line 457: Maybe add a comment on what to do when both are set? We probably don't want to document it as part of API but still

LalitMaganti (approved):

LalitMaganti (approved):

PR #4062, opened 2025-12-09
sdk: add privacy-safe string interning support

zezeozue:
Introduces support for interning event arguments into specific InternedData messages. This enables privacy-safe handling of app-controlled strings, starting with AndroidJobName.

Changes include:
- Proto: Added `AndroidJobName` to `InternedData` (ID 44).
- Shlib: Updated C ABI and macros to support `interned_type_id`.
- Java SDK: Added `addField` overload to support interning via JNI.

Bug: 465114453


LalitMaganti (commented):
Lgtm but would ask Primiano to review this.

include/perfetto/public/abi/track_event_hl_abi.h
```diff
@@ -60,6 +60,11 @@ struct PerfettoTeHlProtoFieldCstr {
   struct PerfettoTeHlProtoField header;
   // Null terminated string.
   const char* str;
+  // If non-zero, this is the field id of the interned data message (e.g.
+  // InternedData.event_names). `str` will be interned and the resulting iid
+  // will be written in the proto field `header.id`. If zero, `str` is written
+  // directly.
+  uint32_t interned_type_id;
```
primiano, line 75: This will break the ABI. there is a question we discussed offline about the SDK being vendor_available.
If anybody has packaged the .so and used with the expectation of being an ABI (as this name suggests) this would cause a very subtle bug.
Unfortunately you have to introdue a new PerfettoTeHlProtoFieldType (CSTR_WITH_INTERNING) and plumb it all the way.

For now I'd expose it only in the ABI (with a comment that is used only in the Java SDK) and not exposed in the macros (because I hate interning with all my heart)

include/perfetto/public/te_macros.h
```diff
@@ -102,11 +102,23 @@ struct PerfettoTeHlMacroNameAndType {
...
+      struct PerfettoTeHlProtoField*,         \
+      PERFETTO_I_TE_COMPOUND_LITERAL_ADDR(    \
+          PerfettoTeHlProtoFieldCstr,         \
+          {{PERFETTO_TE_HL_PROTO_TYPE_CSTR, ID}, VAL, 0}))
+
+// A string or bytes protobuf field (with field id `ID`) and value `VAL` (a null
+// terminated string). `VAL` will be interned. `INTERNED_TYPE_ID` is the field
+// id of one of the repeated messages in InternedData (e.g.
+// InternedData.event_names).
+#define PERFETTO_TE_PROTO_FIELD_CSTR_INTERNED(ID, VAL, INTERNED_TYPE_ID) \
```
primiano: as discussed offline, i'd avoid exposing it in the C macros until somebody needs it (and comes up with a very good ratioanle and explains they understand the risks of interning)

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrackEventBuilder.java
```diff
@@ -525,14 +524,19 @@ public PerfettoTrackEventBuilder addField(long id, double val) {
 
   /** Adds a proto field with field id {@code id} and value {@code val}. */
   public PerfettoTrackEventBuilder addField(long id, String val) {
+    return addField(id, val, 0);
+  }
+
+  /** Adds a proto field with field id {@code id} and value {@code val}. */
+  public PerfettoTrackEventBuilder addField(long id, String val, long internedTypeId) {
```
primiano: given interning has very subtle implications, can you please make it a dedicated method (addInternedField) rather than an overload which takes an extra integer.
the extra integer looks too "innocent". Unfortunately interning is not that innocent and has dramatic consequences on debugging

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrackEventExtra.java
```diff
@@ -545,6 +545,10 @@ public void setValue(long id, String val) {
       native_set_value(mPtr, id, val);
     }
 
+    public void setValue(long id, String val, long internedTypeId) {
```
primiano: ditto

src/shared_lib/track_event/hl.cc
```diff
@@ -42,13 +42,31 @@ protos::pbzero::TrackEvent::Type EventType(int32_t type) {
 }
 
 // Appends the fields described by `fields` to `msg`.
-void AppendHlProtoFields(protozero::Message* msg,
+void AppendHlProtoFields(TrackEventIncrementalState* incr,
+                         protozero::Message* msg,
                          PerfettoTeHlProtoField* const* fields) {
   for (PerfettoTeHlProtoField* const* p = fields; *p != nullptr; p++) {
     switch ((*p)->type) {
       case PERFETTO_TE_HL_PROTO_TYPE_CSTR: {
```
primiano, line 53: same comment as before, this has to be its own enum

primiano:
> Lgtm but would ask Primiano to review this.

Zim and I had some discussion offline about this.
TL;DR I'm not a fan if iterning, but given this is going to be used in 24h pwoer traces which use DISCARD Zim convinced me it's worth it. I'd limit the spreading though.

include/perfetto/public/abi/track_event_hl_abi.h
zezeozue: Done

include/perfetto/public/te_macros.h
zezeozue: Done

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrackEventBuilder.java
zezeozue: done

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrackEventExtra.java
zezeozue: done

src/shared_lib/track_event/hl.cc
zezeozue: done

src/android_sdk/perfetto_sdk_for_jni/tracing_sdk.h
```diff
@@ -367,7 +367,7 @@ class ProtoField {
       } else if constexpr (std::is_same_v<ProtoT, double>) {
         return type_identity<PerfettoTeHlProtoFieldDouble>{};
       } else if constexpr (std::is_same_v<ProtoT, const char*>) {
-        return type_identity<PerfettoTeHlProtoFieldCstr>{};
+        return type_identity<PerfettoTeHlProtoFieldCstrInterned>{};
```
primiano: is this intended? seems accidental

src/android_sdk/perfetto_sdk_for_jni/tracing_sdk.h
```diff
@@ -398,15 +398,23 @@ class ProtoField {
   }
 
   void set_value(uint32_t id, T value) {
-    if constexpr (std::is_same_v<T, int64_t>) {
+    if constexpr (std::is_same_v<T, const char*>) {
+      set_value(id, value, 0);
```
primiano: doesn't this mean that every time you set_value a string field is going to go through the PERFETTO_TE_HL_PROTO_TYPE_CSTR_INTERNED case?
that's onl true for interned strings no?

src/shared_lib/track_event/hl.cc
```diff
@@ -51,6 +52,28 @@ void AppendHlProtoFields(protozero::Message* msg,
...
+        if (field->interned_type_id) {
+          const void* str = field->str;
+          size_t len = strlen(field->str);
+          auto res = incr->iids.FindOrAssign(
+              static_cast<int32_t>(field->interned_type_id), str, len);
+          if (res.newly_assigned) {
+            auto* ser = incr->serialized_interned_data
+                            ->BeginNestedMessage<protozero::Message>(
+                                field->interned_type_id);
+            ser->AppendVarInt(perfetto_protos_AndroidJobName_iid_field_number,
```
LalitMaganti: This should not hardcode job_name: we should hardcode 1 and 2 here and add a comment saying that all "interned message fields" should have this strucutre.

src/shared_lib/track_event/hl.cc
```diff
@@ -51,6 +55,26 @@ void AppendHlProtoFields(protozero::Message* msg,
...
+          if (res.newly_assigned) {
+            auto* ser = incr->serialized_interned_data
+                            ->BeginNestedMessage<protozero::Message>(
+                                field->interned_type_id);
+            ser->AppendVarInt(kInternedStringIidFieldNumber, res.iid);
+            ser->AppendString(kInternedStringNameFieldNumber, field->str);
+          }
+          msg->AppendVarInt(field->header.id, res.iid);
+        } else {
+          msg->AppendString(field->header.id, field->str);
```
primiano: remove this as it might hide bugs.
I'd prob just add a DCHECK(field->interned_type_id > 0) and keep the if (without the else part)

src/android_sdk/perfetto_sdk_for_jni/tracing_sdk.h
zezeozue: Resolved offline and fixed by adding an is_intern template param

src/shared_lib/track_event/hl.cc
zezeozue: Good catch. Thanks

src/shared_lib/track_event/hl.cc
zezeozue: done

dreveman:
@zezeozue I'm assuming this isn't anything that the Rust SDK needs to support at this time? Should be easy to add so just let me know if that's incorrect.

primiano:
> @zezeozue I'm assuming this isn't anything that the Rust SDK needs to support at this time? Should be easy to add so just let me know if that's incorrect.

Nah. it exposes the ability to intern a string in a dedicated field name. But that's something I don't want to expose not even to the C sdk (We just exposed at the ABI level for Java). Interning, as it is right now, is something that works only when the moon is aligned with mercury and the sea temperature is between 14.2 and 14.7 C. It causes more problems than what it solves.
But zim seemed very keen because in their speific use case they seem to make good use of it. But it's not something i'd generally advise people to jump into unless they fully understand the nuance.

PR #4064, opened 2025-12-10
Add dependency on base CUJs stdlib module in jank/cujs.sql

devianb:
Remove the old android_jank_cuj table from the file, and add a direct dependency on the cujs_base stdlib module, enabling all usages of the script to have access to the android_jank_cuj stdlib table.

Test: tools/diff_test_trace_processor.py --name-filter ".*android.*" out/linux_clang_debug/trace_processor_shell


src/trace_processor/metrics/sql/android/jank/cujs.sql
```diff
@@ -13,109 +13,5 @@
...
-FROM cujs
-WHERE
-  state != 'canceled'
-  -- Older builds don't have the state markers so we allow NULL but filter out
-  -- CUJs that are <4ms long - assuming CUJ was canceled in that case.
-  OR (state IS NULL AND cujs.dur > 4e6)
-ORDER BY ts ASC;
\ No newline at end of file
+-- Add a dependency on the base CUJs module for access to the android_jank_cuj table.
+INCLUDE PERFETTO MODULE android.cujs.cujs_base;
```
nicomazz, line 17: nit, probably not for this cl: I wonder if repeting "cuj" in "cuj_base" makes sense. From an high level POV I would have actually expected to see only `INCLUDE .. android.cuj`. Would this be possible in a follow up, or do we still need to reference "base"?

LalitMaganti (approved):

src/trace_processor/metrics/sql/android/jank/cujs.sql
devianb: `INCLUDE .. android.cuj;` does not work, as a module is required. A wildcard `cuj.*` can work though, but not recommended as we do not need to include the rest of the files in the directory.
Can rename the module to base.sql (in a follow up CL).

Issue #992, opened 2025-01-22
update supported protobuf version


evanstade:
perfetto is not compatible with newer protobuf, for at least one reason: its overrides of `google::protobuf::compiler::MultiFileErrorCollector` are no longer functional; see [this change](https://github.com/protocolbuffers/protobuf/commit/d538808032d66366364e410200246c236e275e0f) and [this change](https://github.com/protocolbuffers/protobuf/commit/543fbcdbd9496de9d93a6eb4645640cb47a5514b).

In Chromium this forces us to revert the second change in our copy of protobuf to support perfetto.

LalitMaganti:
From what I understand Chromium is on the same version of protobuf as us. Is that uprev happening on the Chromium side?

evanstade:
Yes, Chromium is [trying to uprev to pb 29.3](https://crbug.com/328417294). This is the only breakage related to that which I have noticed thus far, which is fairly lucky :) [This](https://chromium-review.googlesource.com/c/chromium/src/+/6173918/22/third_party/protobuf/patches/0047-perfetto-backcompat.patch) is the workaround. After Perfetto is updated and rolled in Chromium we can simply remove that patch.

evanstade:
I have also noticed that `trace_processor_shell` should probably have a `dep` on `../../gn:protoc_lib` because it uses `Parser` which is part of the pb compiler. For some reason, Chromium currently pulls `Parser` into `protobuf_full`, however I suspect that Chromium should not do so. Making this change in perfetto and will allow Chromium to fix this oddity.

dtapuska:
I wanted to log the optimization_guide protos in a perfetto trace but it appears I can't do this because Chromium currently supports v31.0, the optimization_guide uses edition 2023. Chromium contains support for that edition but perfetto doesn't.

markafoltz:
Can Chromium's import of perfetto upgrade to proto 2023?  Or is this more complicated than just rolling proto as a dependency?

LalitMaganti:
The proto in question also needs to roll into the Perfetto repo.

LalitMaganti:
@evanstade FYI the patch in Chromium is no longer needed because of https://github.com/google/perfetto/pull/3082

evanstade:
@LalitMaganti thanks, I filed [this issue](https://issues.chromium.org/issues/449362603) to track cleanup in Chromium.

PR #4082, opened 2025-12-10
perfetto: upgrade protobuf to v26.1

LalitMaganti:
We need to add a dependency on absl indirectly because protobuf now
depends on this but it's restricted to standalone build only

Note: we temporarily have to uprev Bazel to v29 because of mismatching
versions. We'll align everything in a followup when we uprev protobuf
further.

Fixes: https://github.com/google/perfetto/issues/992

### DIFF
--- a/src/perfetto_cmd/BUILD.gn
+++ b/src/perfetto_cmd/BUILD.gn
@@ -105,9 +105,10 @@ source_set("trigger_perfetto_cmd") {
     "../tracing/ipc/producer",
   ]
   sources = [ "trigger_perfetto.cc" ]
+
   # Abseil is only used in Perfetto via the gRPC/libprotobuf.
   if (perfetto_build_standalone || is_perfetto_build_generator) {
-    assert_no_deps = [ "buildtools:abseil_cpp" ]
+    assert_no_deps = [ "../../buildtools:abseil_cpp" ]
   }
 }
 
--- a/src/profiling/memory/BUILD.gn
+++ b/src/profiling/memory/BUILD.gn
@@ -41,9 +41,10 @@ source_set("heapprofd_main") {
     "heapprofd.cc",
     "heapprofd.h",
   ]
+
   # Abseil is only used in Perfetto via the gRPC/libprotobuf.
   if (perfetto_build_standalone || is_perfetto_build_generator) {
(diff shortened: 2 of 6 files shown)


tools/install-build-deps
```diff
@@ -183,11 +183,19 @@ BUILD_DEPS_HOST = [
         'https://android.googlesource.com/platform/external/googletest.git',
         '609281088cfefc76f9d0ce82e1ff6c30cc3591e5', 'all', 'all'),
 
-    # Keep in sync with Chromium's //third_party/protobuf.
+    # Keep in sync with gRPC's third_party/protobuf.
```
primiano, line 186: don't we have to keep following chromimum?

tools/install-build-deps
```diff
@@ -183,11 +183,19 @@ BUILD_DEPS_HOST = [
...
     Dependency(
         'buildtools/protobuf',
         'https://chromium.googlesource.com/external/github.com/protocolbuffers/protobuf.git',
-        'f0dc78d7e6e331b8c6bb2d5283e06aa26883ca7c',  # refs/tags/v21.12
+        '2434ef2adf0c74149b9d547ac5fb545a1ff8b6b5',  # refs/tags/v26.1
+        'all',
+        'all'),
+
+    # Required by protobuf v22+ and gRPC.
```
primiano: Add a note saying it's not allowed in our repo (% protobuf compat)?

buildtools/BUILD.gn
```diff
@@ -128,6 +128,240 @@ source_set("gmock") {
...
+  }
+  if (is_gcc) {
+    cflags += [
+      "-Wno-stringop-overflow",
+      "-Wno-ignored-attributes",
+    ]
+  }
+}
+
+source_set("abseil_cpp") {
```
primiano, line 152: can you add an assert_no_deps to make sure that nothing ends up depending on this (other than the protolite targets)
This should not end up in our binaries or in the amalgamated SDK

tools/install-build-deps
LalitMaganti: Unfortunately, gRPC is now the furthest behind so we have to stick to it. We have to keep to whatever the oldest version is: thankfully the public API for protobuf doesn't change too often so it's OK.

PR #4092, opened 2025-12-11
ui: Tweak home page styling

stevegolton:
This patch just aims to modernize the home page styling a little, rather than a complete overhaul (which it needs, but this is just a bandaid while we work out how to improve it).

- Modernize the homepage styling.
- When on the autopush channel, show autopush in the selector (as in screenshot below). When in canary/stable channel, autopush is hidden, as before.
- Add reload button when reload is required to change the release channel.
- Use consistent anchors for external links.
- Channel colors in selector and the sidebar:
  - Stable - green (nothing in sidebar)
  - Canary - yellow
  - Autopush - red
 
image

### DIFF
--- a/ui/src/assets/home_page.scss
+++ b/ui/src/assets/home_page.scss
@@ -51,7 +51,6 @@
     .pf-home-page__hints {
       display: flex;
       flex-direction: column;
-      align-items: center;
       gap: 24px;
       margin-bottom: 32px;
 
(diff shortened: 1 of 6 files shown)


LalitMaganti:
Given were doing this, can we also add "open new trace" and "record trace" as buttons here?

primiano:
Can you show some red text below when selecting autopush, soemthing like "WARNING: autopush is updated on every new code change and is extremely unstable. Use at your own risk"

stevegolton:
> Can you show some red text below when selecting autopush, soemthing like "WARNING: autopush is updated on every new code change and is extremely unstable. Use at your own risk"

Ah so the thing is the autopush button only appears when you're actually on autopush, so it's hidden by default.

So why add it at all? If you're already on autopush, the homepage currently tells you you're on stable, which is confusing.

This is what it looks like when you're on stable/canary.

image

primiano:
> > Can you show some red text below when selecting autopush, soemthing like "WARNING: autopush is updated on every new code change and is extremely unstable. Use at your own risk"
> 
> Ah so the thing is the autopush button only appears when you're actually on autopush, so it's hidden by default.
> 
> So why add it at all? If you're already on autopush, the homepage currently tells you you're on stable, which is confusing.
> 
> This is what it looks like when you're on stable/canary.
> 
> image

Aaaah I see. makes perfect sense as-is then

LalitMaganti:
Are we still planning on landing this?

stevegolton:
> Are we still planning on landing this?

Yep, though I got hung up on adding the recent traces. Alright to land as-is then land the recents later?

LalitMaganti:
Yes please

PR #4096, opened 2025-12-11
[SDK] Add TrackEventSessionObserver::OnDidStop

etiennep-chromium:
It's not trivial as-is to implement a common pattern used in chrome: start a timer when a category is enabled, and stop the timer when the category is disabled.
This is because TRACE_EVENT_CATEGORY_ENABLED isn't reflecting the session stopping in TrackEventSessionObserver:OnStop().
This PR adds TrackEventSessionObserver:OnDidStop: as a result, TRACE_EVENT_CATEGORY_ENABLED will reflect the change, which makes the pattern described easily achievable.

HandleStopAsynchronously() becomes officially unsupported; although it was only partially supported before as multiple observer calling HandleStopAsynchronously() would not work, and never used. This allows us to reuse the same StopArgs.


include/perfetto/tracing/internal/track_event_internal.h
```diff
@@ -74,10 +74,13 @@ class PERFETTO_EXPORT_COMPONENT TrackEventSessionObserver {
...
-  // emit track events from this callback.
+  // emit track events from this callback. TRACE_EVENT_CATEGORY_ENABLED() will
+  // include the session being started.
   virtual void OnStart(const DataSourceBase::StartArgs&);
-  // Called when a track event tracing session is stopped. It is still possible
-  // to emit track events from this callback.
+  // Called when a track event tracing session is stopped.
+  // TRACE_EVENT_CATEGORY_ENABLED() will not include the session being stopped.
+  // Unlike DataSource::OnStop, it is no longer possible to emit track events
+  // from this callback.
```
LalitMaganti: Instead of making a breaking change like this, could we instead add a OnPostStop? as a separate callback? Would do what you needed and no need to break anyone today.

etiennep-chromium: I could do that, but re. StopArgs, either:
- I make a new TrackEventSessionObserver:PostStopArgs here (to plumb internal_instance_index)
- I reuse DataSourceBase::StopArgs but disallow HandleStopAsynchronously()
Thoughts?
Also, HandleStopAsynchronously() is broken as-is (and I don't think it's worth fixing)

LalitMaganti: If "HandleStopAsynchronously()" is broken, can we just make it check like you've done?

And I would just use the same args for both honestly.

primiano:
Hmm can you explain a bit more? I'm not sure I understand what you are trying to do. Why can't you stop the timer in OnStop?

Regardless of your use case, what makes me quite nervous here is that you are swapping the order of DisableTracing and OnStop... and who knows at this point what other users are doing. this seems too risky, for a reason I cannot undersand

etiennep-chromium:
I updated the title: this PR no longer changes OnStop() order, but rather adds onDidStop().
Re usage:
observer::OnStart checks for TRACE_EVENT_CATEGORY_ENABLED("my_category"), observer::OnStop can't do that.
To correctly test "my_category" is enabled/disable, we'd have to track individual sessions, but there's no API to test a category on an individual track event session (it's internal to perfetto). I found 5 observers in chrome and v8 that try to do that; they all do it wrong, and it seems like there's no way to do it right.
See for example
https://source.chromium.org/chromium/chromium/src/+/main:v8/src/profiler/tracing-cpu-profiler.cc;l=67;drc=397107e399d4b0f4a6e801e7a888d7d71778529c

primiano:
> TRACE_EVENT_CATEGORY_ENABLED

observer::OnStart checks for TRACE_EVENT_CATEGORY_ENABLED("my_category"),

Yeah but why do you care about that? I mean conceptually I don't understand a rationale that justifies the double callback.
If somebody tells you "i'm about to stop the trace" then that's information is enough. it's even giving you time to still write into the buffer.

I don't understand why you want to test that my_category is disabled, if you see an OnStop, you know it will?

I am sorry the problem here is that you are proposing to add a new callback, and you are not explaining why that is desirable.
Can you please explain your rationale?

etiennep-chromium:
> I don't understand why you want to test that my_category is disabled, if you see an OnStop, you know it will?

In a multiple sessions scenario, the observer will see multiple OnStart / OnStop. The observer is interested in turning on/off something (e.g. a repeating timer) when my_category is enabled. Something that can happen:
a: session 1 OnStart enables my_category
b: session 2 OnStart doesn't enable my_category
c: session 1 OnStop: how does that caller know whether my_category is still enabled or not?

Importantly, there's no easy way to check at (b) that this specific data source didn't enable my_category. We could check Registry->GetCategoryState(category_index) at the data_source_instance_index bit offset, but that's not documented and I'm not sure we'd want to rely on that.

primiano:
Okay now I start understanding. I think the core problem is that you are reasoning on side-effects of Start/Stop on the category system, which seems the root of most problems. You are taking something low level and trying to use to reason on something higher level. It's a bit as if you were doing layer2 ethernet inspection to figure out when it's the right time when to check when a TCP socket is closed

A proposal, if you have this need, why don't you add an OnCategoryChange observer or something like that?
it seems a better semantic than introducing an OnDidStop which goes "this is like OnStop, but has this extra semantic".
Because that's how you end up over time with OnStop, OnDidStop, OnDidStopDone, OnDidStopCompleted

If there is enough use cases for observing categories being turned on and off (Which is a use case I can be convinced easily it's a legit use case) then let's create a first class citizen for it?

etiennep-chromium:
> Because that's how you end up over time with OnStop, OnDidStop, OnDidStopDone, OnDidStopCompleted

Indeed, I'm not a fan of this either

> A proposal, if you have this need, why don't you add an OnCategoryChange observer or something like that?

One thing that concerns me is how inflexible that would be. What if we want to check 2 categories, or also check something else in the config (disabled debug annotations).
Perhaps what's missing is an easy way to test a category given a data_source_instance_index (which is already forwarded in Start/StopArgs)?

primiano:
Then maybe you have an OnCategoryChanged callback.

I honestly don't have a strong opinion right now on what's the shape of the callback (if you have time try coming up with a proposal). The only strong opinion I have right now is that for the use cases you are trying to tackle we should much rather have a category-centric callback (or set of callbacks, or observer TBD) rather than a side-effect based callback.

etiennep-chromium:
Ok how about something like this: https://github.com/google/perfetto/pull/4106

primiano:
> Ok how about something like this: #4106

I'm not opposed (I commented there).... but I don't understand how that alone helps. I thought you needed a callback at the right time? When are you going to call those functions? DOn't you have the same problem that if you call them on OnStop it's too early?

etiennep-chromium:
> When are you going to call those functions? DOn't you have the same problem that if you call them on OnStop it's too early?

We can run things in OnStart/OnStop() and it's not necessarily too early. What's missing is a way to know that a given category is being (or going to) disabled. Since we can already track individual sessions/configs with the existing TrackEventSessionObserver, adding IsCategoryEnabledByConfig() fills the missing gap.

primiano:
Ah ok. I guess we can close this PR then give #4106 just landed?

PR #4100, opened 2025-12-12
Split _android_frames_in_cuj table with layer_id consideration.

devianb:
The current _android_frames_in_cuj table disregards frame distinction based on layer_id. Distinction based on layer_id is necessary for downstream metrics, hence splitting the current table into two, for both distinction use-cases.

Test: tools/diff_test_trace_processor.py --name-filter ".*android.*" out/linux_clang_debug/trace_processor_shell

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
@@ -132,12 +132,16 @@ RETURNS STRING AS
 SELECT
   substr($cuj_slice_name, 3, length($cuj_slice_name) - 3);
 
+-- Information about all frames in a process that overlap with a CUJ from the same process.
+-- This can include multiple frames for the same frame_id (for eg. frames with different layers).
 CREATE PERFETTO VIEW _all_frames_in_cuj AS
 SELECT
   _extract_cuj_name_from_slice(cuj.cuj_slice_name) AS cuj_name,
   cuj.upid,
   cuj.process_name,
+  cie.layer_id AS cuj_layer_id,
   frame.layer_id,
+  frame.layer_name,
   frame.frame_id,
   frame.do_frame_id,
   frame.expected_frame_timeline_id,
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
```diff
@@ -132,52 +132,51 @@ RETURNS STRING AS
 SELECT
   substr($cuj_slice_name, 3, length($cuj_slice_name) - 3);
 
+CREATE PERFETTO VIEW _all_frames_in_cuj AS
```
nicomazz, line 137: let's add more comments to this, even if it's "private". Ideally someone without context should understand why we have an "all_frames" and a "distinct frames" table

src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
```diff
@@ -132,52 +132,51 @@ RETURNS STRING AS
...
+    OR (
+      (
+        frame.ts + frame.dur
+      ) >= cuj.ts AND (
+        frame.ts + frame.dur
+      ) <= cuj.ts_end
+    )
+  );
+
 -- Track all distinct frames that overlap with the CUJ slice.
```
nicomazz: let's also specify here what we mean by "distinct"

src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
```diff
@@ -201,11 +200,42 @@ SELECT
...
+  process_name,
+  frame_id,
+  do_frame_id,
+  expected_frame_timeline_id,
+  cuj_id,
+  ui_thread_utid,
+  -- In case of multiple frames for a frame_id, consider the min start timestamp.
+  min(frame_ts) AS frame_ts,
+  -- In case of multiple frames for a frame_id, consider the max end timestamp.
+  max(ts_end) AS ts_end,
```
nicomazz: In practise can we have more than one frame per layer id? If so, let's write it explicitly. If not, do we need to use min/max here?

src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
```diff
@@ -201,11 +200,42 @@ SELECT
...
 GROUP BY
   frame_id,
   cuj_id;
 
+-- Track all distinct frames with layer_id consideration that overlap with the CUJ slice.
+CREATE PERFETTO VIEW _android_distinct_frames_layers_cuj AS
+-- Captures all frames in the CUJ boundary. In cases where there are multiple actual frames, there
+-- can be multiple rows with the same frame_id.
+SELECT
+  -- Column values with no aggregation function will stay identical across rows. For eg.
```
nicomazz: I would suggest rewording this: you want to mean that after doing the "group by" specified below the remaining columns are expected to always have the same value, right?

src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
devianb: Done

src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
devianb: Removed aggregation, as having more than one frame per layer_id is not possible.

src/trace_processor/perfetto_sql/stdlib/android/cujs/base.sql
devianb: Done

nicomazz (approved):

PR #4106, opened 2025-12-12
[SDK] Expose TrackEvent::IsCategoryEnabledBySession/Config

etiennep-chromium:
This is an alternative solution to https://github.com/google/perfetto/pull/4096#issuecomment-3646496843
To enables a common pattern chrome where an observer is interested in a specific category to control e.g. a timer.

Instead of depending on side-effect of OnStart/OnStop, this PR exposes IsCategoryEnabledByConfig that can be queried directly on the TrackEvent session being started/stopped.

### DIFF
--- a/include/perfetto/tracing/internal/track_event_data_source.h
+++ b/include/perfetto/tracing/internal/track_event_data_source.h
@@ -412,7 +412,7 @@ class TrackEvent {
   static bool IsCategoryEnabledBySession(size_t internal_instance_index,
                                          size_t category_index) {
     std::atomic<uint8_t>* state = Registry->GetCategoryState(category_index);
-    return state->load(std::memory_order_release) &
+    return state->load(std::memory_order_relaxed) &
            static_cast<uint8_t>(1u << internal_instance_index);
   }
 


include/perfetto/tracing/internal/track_event_data_source.h
```diff
@@ -391,6 +391,29 @@ class TrackEvent {
     return enabled;
   }
 
+  static bool IsCategoryEnabledByConfig(
+      const protos::gen::TrackEventConfig& config,
+      const char* category_name) {
```
primiano: for consistency with the other methods this should be size_t category_index as the other methods use the index. How is the other used? is it via TRACE_EVENT_CATEGORY_ENABLED right?

include/perfetto/tracing/internal/track_event_data_source.h
```diff
@@ -391,6 +391,29 @@ class TrackEvent {
...
+
+  static bool IsCategoryEnabledByConfig(
+      const protos::gen::TrackEventConfig& config,
+      const DynamicCategory& category) {
+    return TrackEventInternal::IsCategoryEnabled(
+        *Registry, config, Category::FromDynamicCategory(category));
+  }
+
+  static bool IsCategoryEnabledBySession(size_t internal_instance_index,
+                                         const char* category_name) {
```
primiano: ditto here about using category_index like every other method

include/perfetto/tracing/internal/track_event_data_source.h
etiennep-chromium: > How is the other used? is it via TRACE_EVENT_CATEGORY_ENABLED right?

Yes I believe so

In this case, if we take in index this could be used as:
TrackEvent::IsCategoryEnabledByConfig(config, PERFETTO_GET_CATEGORY_INDEX("my_category"))

It's a bit ugly but there's only going to be a couple of callsites.

primiano: > It's a bit ugly but there's only going to be a couple of callsites.

Yeah but it's consistent with the rest of the functions. And you can add some wrapper in chrome.
Sorry unfortunately after Daniele left I have to rediscover all these things but don't always have time to dig all the archaeology. I assume there is some reason why people passed the index. and given it's not a big deal i'd stay consistent with it unless there is some big reason not to

include/perfetto/tracing/internal/track_event_data_source.h
```diff
@@ -391,6 +391,31 @@ class TrackEvent {
...
+    return TrackEventInternal::IsCategoryEnabled(
+        *Registry, config, Category::FromDynamicCategory(category));
+  }
+
+  // Determine if tracing for the given static category is enabled in a session
+  // identified by `internal_instance_index`.
+  static bool IsCategoryEnabledBySession(size_t internal_instance_index,
+                                         size_t category_index) {
+    std::atomic<uint8_t>* state = Registry->GetCategoryState(category_index);
+    return state->load(std::memory_order_release) &
```
primiano: load with std::memory_order_release is very very weird (I'm not sure most platform even support that in practice).
Is there a reason why this is not an acquire?

LalitMaganti: https://en.cppreference.com/w/cpp/atomic/atomic/load.html: it's actually undefined behaviour

etiennep-chromium: Oops I meant std::memory_order_relaxed.
The write has _release with a comment that refers to a load
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/tracing/track_event_category_registry.cc;l=60;drc=6fe04ac5cd13eaa7013d9da4b94144b4caa1fabf;bpv=1;bpt=1

But what this seemingly refers to is actually _relaxed:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/include/perfetto/tracing/internal/track_event_data_source.h;l=396?q=track_event_data_source.h&ss=chromium

I believe _relaxed is ok, because we don't synchronize with any other state.

primiano: > I believe _relaxed is ok, because we don't synchronize with any other state.

yeah worst case you can try to trace, and then the trace path will have an actual acquire at some point before writing into the buffer. relaxed here is ok.

PR #4113, opened 2025-12-12
protozero: Add v54 bytecode overlay support for trace filtering

LalitMaganti:
Add overlay support to FilterBytecodeParser and FilterBytecodeGenerator
for upgrading fields (e.g., to FilterString) without duplicating bytecode.

Parser: Two-pointer merge of base bytecode and overlay during Load().
Generator: Returns SerializeResult with bytecode and v54_overlay fields.
CLI: Add --overlay_v54_out and --overlay_v54_oct_out options.

### DIFF
--- a/src/protozero/filtering/filter_bytecode_parser.h
+++ b/src/protozero/filtering/filter_bytecode_parser.h
@@ -79,7 +79,7 @@ class FilterBytecodeParser {
   // can add new fields or upgrade existing fields (e.g., to FilterString).
   // The overlay uses the same opcode format as the base bytecode, but with an
   // explicit message index prefix:
-  //   [msg_index, (field_id << 3) | opcode, extra_args...] ... [checksum]
+  //   [msg_index, (field_id << 3) | opcode, argument] ... [checksum]
   // Entry sizes depend on the opcode (same as base bytecode).
   // Entries must be sorted by (msg_index, field_id). The base bytecode and
   // overlay are merged during loading, maintaining sorted field order.


src/protozero/filtering/filter_bytecode_generator.h
```diff
@@ -72,6 +83,10 @@ class FilterBytecodeGenerator {
   bool endmessage_called_ = false;
 
   std::vector<uint32_t> bytecode_;
+
+  // Overlay entries for v54. Each entry is [msg_index, field_word, ...].
```
primiano: `[msg_index, field_id, semantic_type]`?

src/protozero/filtering/filter_bytecode_parser.cc
```diff
@@ -16,64 +16,129 @@
...
+  bool packed_parse_err = false;
+  words->reserve(len);  // An overestimation, but avoids reallocations.
+  using BytecodeDecoder =
+      PackedRepeatedFieldIterator<proto_utils::ProtoWireType::kVarInt,
+                                  uint32_t>;
+  for (BytecodeDecoder it(data, len, &packed_parse_err); it; ++it)
+    words->emplace_back(*it);
+
+  if (packed_parse_err || words->empty())
+    return false;
```
primiano, line 69: just checking, don't you need to return true if len == 0? or do you expect a checksum? or you don't expect it at all to be called if empty (but then add a DCHECK(data && len > 0) to improve readability)

primiano, line 69: just checking, don't you need to return true if len == 0? or do you expect a checksum? or you don't expect it at all to be called if empty (but then add a DCHECK(data && len > 0) to improve readability)

src/protozero/filtering/filter_bytecode_parser.cc
```diff
@@ -16,64 +16,129 @@
...
+  uint32_t expected_csum = static_cast<uint32_t>(hasher.digest());
+  if (expected_csum != words->back()) {
+    if (!suppress_logs) {
+      PERFETTO_ELOG("Filter bytecode checksum failed. Expected: %x, actual: %x",
+                    expected_csum, words->back());
+    }
+    return false;
+  }
+
+  words->pop_back();  // Pop the checksum.
```
primiano: I would have probably written this in a single loop. Not much for perf but to avoid the fact that the output `words` will contin the checksum if you fail.

Something like:

```
uint32_t expected_csum = 0;
for (BytecodeDecoder it(data, len, &packed_parse_err); it; ) {
  uint32_t word = *it;
  ++it;
  const bool is_last_word = !it;

  if (!is_last_word) {
    words.emplace_back(word);
    hasher.update(word);
  } else {
    expected_csum = static_cast<uint32_t>(hasher.digest());
    break;
  }
}

if (expected_csum != words->back()) {
  if (!suppress_logs) {
    PERFETTO_ELOG("Filter bytecode checksum failed. Expected: %x, actual: %x",
                  expected_csum, words->back());
  }
  
  words.clear(); // Maybe also consider this?
  return false;
}

return true
```

src/protozero/filtering/filter_bytecode_parser.cc
```diff
@@ -16,64 +16,129 @@
...
-    if (!suppress_logs_for_fuzzer_) {
-      PERFETTO_ELOG("Filter bytecode checksum failed. Expected: %x, actual: %x",
-                    expected_csum, words.back());
+  // Parse the overlay (if provided).
+  std::vector<uint32_t> overlay;
+  if (overlay_data && overlay_len > 0) {
+    if (!ParseAndVerifyChecksum(overlay_data, overlay_len, &overlay,
+                                suppress_logs_for_fuzzer_)) {
+      return false;
     }
```
primiano, line 141: can you add a check to check std::is_sorted (to both) and bail out with an error?
if we ever forget this (or realistically, if somebody one day ever writes some pipeline tooling that messes with the bytecode) this will be the most hard to debug bug ever. 

TBH in order to do this, it might be beneficial to first tokenize the overlay and turn it into a vector of struct{message, field, op} and then process them. I think might make the code below easier to reason.
Technnically you can do the is_sorted check without de-structuring, but it will look ugly as hell :)

EDIT after having read below: i'm 50:50 on my suggesiton to destructure. on one hand, I feel makes the code easier to read. It adds an extra allocation for the vector (but it's a vector of pods, and you can reserve the size/3 upfront)

Up to you, i'm fine both ways. I'd probably err a bit on the side of readability TBH

src/protozero/filtering/filter_bytecode_parser.cc
```diff
@@ -90,6 +155,92 @@ bool FilterBytecodeParser::LoadInternal(const uint8_t* bytecode_data,
...
+      if (PERFETTO_UNLIKELY(overlay_idx + 1 >= overlay.size())) {
+        PERFETTO_DLOG("overlay error: truncated entry at index %zu",
+                      overlay_idx);
+        return kOverlayError;
+      }
+
+      // Parse the entry header.
+      uint32_t entry_msg = overlay[overlay_idx];
+      uint32_t entry_word = overlay[overlay_idx + 1];
+      uint32_t entry_opcode = entry_word & 0x7u;
```
primiano: can you extract 0x7u and the "3" into `kOpcode{Mask,Shift}`?
If we ever decide to change the opcode size in some upgrade, we might accidentally miss the fact that there are two 0x7u around here.

src/protozero/filtering/filter_bytecode_parser.cc
```diff
@@ -170,6 +337,7 @@ bool FilterBytecodeParser::LoadInternal(const uint8_t* bytecode_data,
       words_.insert(words_.end(), ranges.begin(), ranges.end());
       direct_indexed_fields.clear();
       ranges.clear();
+      ++current_msg_index;
```
primiano, line 367: can we add some sanitization step at the end that checks that everything is in order (e.g. ranges don't overlap)

primiano (approved):
LGTM with a few suggestions

PR #4114, opened 2025-12-12
protozero: Add semantic types for per-field string filtering

LalitMaganti:
Add support for semantic types that allow per-field string filtering rules.
This enables applying different string filter policies to different types
of data (e.g., atrace events, job names, wakelock names).

See /rfcs/0011-subset-string-filter-rules.md for design details.

### DIFF
--- a/src/protozero/filtering/string_filter_benchmark.cc
+++ b/src/protozero/filtering/string_filter_benchmark.cc
@@ -98,9 +98,9 @@ void Benchmark(
 }  // namespace
 
 static void BM_ProtozeroStringRewriterRedactMissing(benchmark::State& state) {
-  Benchmark(state,
-            {{Policy::kMatchRedactGroups,
-              R"(S\|[^|]+\|\*job\*\/.*\/.*\/(.*)\n)", "", SemanticTypeMask::All()}});
+  Benchmark(state, {{Policy::kMatchRedactGroups,
+                     R"(S\|[^|]+\|\*job\*\/.*\/.*\/(.*)\n)", "",
+                     SemanticTypeMask::All()}});
 }
 BENCHMARK(BM_ProtozeroStringRewriterRedactMissing)
     ->Unit(benchmark::kMillisecond)
@@ -137,9 +137,9 @@ BENCHMARK(BM_ProtozeroStringRewriterAtraceRedactRare)
 
 static void BM_ProtozeroStringRewriterAtraceSearchSingleRedactRare(
     benchmark::State& state) {
-  Benchmark(state,
-            {{Policy::kAtraceRepeatedSearchRedactGroups,
-              R"(VerifyClass (.*)\n)", "VerifyClass", SemanticTypeMask::All()}});
+  Benchmark(state, {{Policy::kAtraceRepeatedSearchRedactGroups,
+                     R"(VerifyClass (.*)\n)", "VerifyClass",
+                     SemanticTypeMask::All()}});
 }
 BENCHMARK(BM_ProtozeroStringRewriterAtraceSearchSingleRedactRare)
     ->Unit(benchmark::kMillisecond)
@@ -156,9 +156,10 @@ BENCHMARK(BM_ProtozeroStringRewriterRedactCommon)
 
 static void BM_ProtozeroStringRewriterAtraceRedactCommon(
     benchmark::State& state) {
-  Benchmark(state, {{Policy::kAtraceMatchRedactGroups,
-                     R"(B\|[^|]+\|Lock contention on a monitor lock (.*)\n)",
-                     "Lock contention on a monitor lock", SemanticTypeMask::All()}});
+  Benchmark(state,
+            {{Policy::kAtraceMatchRedactGroups,
+              R"(B\|[^|]+\|Lock contention on a monitor lock (.*)\n)",
+              "Lock contention on a monitor lock", SemanticTypeMask::All()}});
 }
 BENCHMARK(BM_ProtozeroStringRewriterAtraceRedactCommon)
     ->Unit(benchmark::kMillisecond)


src/protozero/filtering/string_filter.cc
```diff
@@ -94,19 +94,49 @@ void RedactMatches(const Matches& matches) {
...
 
-bool StringFilter::MaybeFilterInternal(char* ptr, size_t len) const {
+bool StringFilter::MaybeFilterInternal(char* ptr,
+                                        size_t len,
+                                        uint32_t semantic_type) const {
   std::match_results<char*> matches;
   bool atrace_find_tried = false;
   const char* atrace_payload_ptr = nullptr;
   for (const Rule& rule : rules_) {
+    // Skip rules that don't match the semantic type.
```
primiano: the only comment I can make is that for now, if you want to speculate on the fact that we'll have realistically < 64 semantic types, this could be a uint64 bitmap. It avoids moving a vector around, and replaces the loop with a bitwise and operator.
Maybe will last for a few years?
If you do that, add a CHECK/limit in the generator to stop after 64

src/protozero/filtering/filter_bytecode_generator.cc
```diff
@@ -47,12 +48,40 @@ void FilterBytecodeGenerator::AddSimpleField(uint32_t field_id) {
...
+  PERFETTO_CHECK(min_version_ >= BytecodeVersion::kV2);
+  PERFETTO_CHECK(field_id > last_field_id_);
+
+  // If min_version is at least v54, we can just use the new opcode directly.
+  if (min_version_ >= BytecodeVersion::kV54) {
+    bytecode_.push_back(field_id << 3 | kFilterOpcode_FilterStringWithType);
+    bytecode_.push_back(semantic_type);
+  } else {
+    // Otherwise we'll use plain FilterString opcode and add an overlay entry.
+    bytecode_.push_back(field_id << 3 | kFilterOpcode_FilterString);
```
primiano: discussed offline, we should deny it on baseline if it's a new filter type

src/protozero/filtering/filter_bytecode_generator.h
```diff
@@ -81,6 +104,7 @@ class FilterBytecodeGenerator {
   uint32_t last_field_id_ = 0;
   uint32_t max_msg_index_ = 0;
   bool endmessage_called_ = false;
+  BytecodeVersion min_version_;
```
primiano: add initializer?

src/tools/proto_filter/proto_filter.cc
```diff
@@ -200,11 +221,13 @@ int Main(int argc, char** argv) {
   std::string root_message_arg;
   std::set<std::string> passthrough_fields;
   std::set<std::string> filter_string_fields;
+  std::map<std::string, uint32_t> filter_string_semantic_types;
+  std::string min_bytecode_parser = "v54";  // Default to latest
```
primiano: can you add a comment in the proto saying that if we ever introduce a new version, we should update this?
Either that, or we conservatively here default to v2

src/tools/proto_filter/proto_filter.cc
```diff
@@ -284,6 +307,34 @@ int Main(int argc, char** argv) {
...
+      }
+      std::string field_name = arg.substr(0, last_colon);
+      std::string type_str = arg.substr(last_colon + 1);
+      char* endptr = nullptr;
+      uint32_t type_value =
+          static_cast<uint32_t>(strtoul(type_str.c_str(), &endptr, 10));
+      if (*endptr != '\0') {
+        fprintf(stderr, "Invalid semantic type value: %s\n", type_str.c_str());
+        exit(1);
+      }
```
primiano: This can be simplified by using just string_util's CStringToUInt32

src/tools/proto_filter/proto_filter.cc
```diff
@@ -284,6 +307,34 @@ int Main(int argc, char** argv) {
...
+      size_t last_colon = arg.rfind(':');
+      if (last_colon == std::string::npos) {
+        fprintf(stderr,
+                "Invalid semantic type syntax. Expected: "
+                "MessageName:field:type_value\n");
+        exit(1);
+      }
+      std::string field_name = arg.substr(0, last_colon);
+      std::string type_str = arg.substr(last_colon + 1);
+      char* endptr = nullptr;
```
primiano: Just use base::SplitString?

src/protozero/filtering/string_filter.h
```diff
@@ -17,17 +17,36 @@
...
 class StringFilter {
  public:
-  enum class Policy {
+  // Bitmask for semantic types. Supports up to 128 semantic types (2 * 64).
+  // Bit i in mask[j] set means semantic type i + j*64 is enabled.
+  using SemanticTypeMask = std::array<uint64_t, 2>;
+
+  // Maximum semantic type value supported.
+  static constexpr size_t kSemanticTypeLimit =
+      std::size(SemanticTypeMask()) * 64;
```
primiano: just sizeof(SemanticTypeMask) * 8 ?
1. It's more readable
2. It's less likely to go out of sync if you change tht type of the uint64

src/protozero/filtering/string_filter.h
```diff
@@ -36,28 +55,47 @@ class StringFilter {
...
   }
 
  private:
   struct Rule {
     Policy policy;
     std::regex pattern;
     std::string atrace_payload_starts_with;
+    std::string name;
+    // Bitmask of semantic types this rule applies to.
+    SemanticTypeMask semantic_type_mask;
```
primiano: default initialize to 0/{} (or  AllSemanticTypes())?

src/protozero/filtering/string_filter.h
```diff
@@ -17,17 +17,36 @@
...
 namespace protozero {
 
 // Performs filtering of strings in an "iptables" style. See the comments in
 // |TraceConfig.TraceFilter| for information on how this class works.
 class StringFilter {
  public:
-  enum class Policy {
+  // Bitmask for semantic types. Supports up to 128 semantic types (2 * 64).
+  // Bit i in mask[j] set means semantic type i + j*64 is enabled.
+  using SemanticTypeMask = std::array<uint64_t, 2>;
```
primiano: Can you do

using SemanticTypeWord = uint64_t;
using SemanticTypeMask = std::array<SemanticTypeWord, 2>;

it makes the code mor maintainable and readable

src/protozero/filtering/string_filter.h
```diff
@@ -17,17 +17,36 @@
...
+
+  // Maximum semantic type value supported.
+  static constexpr size_t kSemanticTypeLimit =
+      std::size(SemanticTypeMask()) * 64;
+
+  // Returns a SemanticTypeMask with all bits set (applies to all types).
+  static constexpr SemanticTypeMask AllSemanticTypes() {
+    return {std::numeric_limits<uint64_t>::max(),
```
primiano: return {SemanticTypeWord(-1), SemanticTypeWord(-1)}  (or static_cast if the compiler is unhappy)

src/protozero/filtering/string_filter.cc
```diff
@@ -90,23 +93,54 @@ void RedactMatches(const Matches& matches) {
...
 
+// Checks if a semantic type matches the given rule's semantic type mask.
+constexpr bool DoesRuleMatchSemanticType(
+    const StringFilter::SemanticTypeMask& mask,
+    uint32_t semantic_type) {
+  // If beyond supported range (>= 128), apply rule (safe default).
+  if (PERFETTO_UNLIKELY(semantic_type >= StringFilter::kSemanticTypeLimit)) {
+    return true;
+  }
+  uint32_t word_index = semantic_type / 64;
```
primiano: constexpr kSemanticTypeBits = sizeof(SemanticTypeWord) * 8;
uint32_t word_index = semantic_type / kSemanticTypeBits;
uint32_t bit_index = semantic_type % kSemanticTypeBits;

src/tools/proto_filter/proto_filter.cc
```diff
@@ -355,15 +406,47 @@ int Main(int argc, char** argv) {
...
+    // Helper to convert semantic types from proto to SemanticTypeMask.
+    auto convert_semantic_types =
+        [](const auto& rule) -> protozero::StringFilter::SemanticTypeMask {
+      protozero::StringFilter::SemanticTypeMask mask = {};
+      for (const auto& type : rule.semantic_type()) {
+        uint32_t semantic_type = static_cast<uint32_t>(type);
+        uint32_t word_index = semantic_type / 64;
+        uint32_t bit_index = semantic_type % 64;
+        if (word_index < mask.size()) {
+          mask[word_index] |= (1ULL << bit_index);
```
primiano: same of the other comment

constexpr kSemanticTypeBits = sizeof(SemanticTypeWord) * 8;
uint32_t word_index = semantic_type / kSemanticTypeBits;
uint32_t bit_index = semantic_type % kSemanticTypeBits;


Also c'mon this IMHO shoudl have been a funciton in the anon namespace. this feels like JS (And this code is alreday quite indented)

src/tools/proto_filter/proto_filter.cc
```diff
@@ -374,7 +457,16 @@ int Main(int argc, char** argv) {
     }
   } else if (!schema_in.empty()) {
     PERFETTO_LOG("Generating filter bytecode from %s", schema_in.c_str());
-    auto result = filter.GenerateFilterBytecode();
+
+    // Parse the bytecode version.
+    auto bytecode_version = ParseBytecodeVersion(min_bytecode_parser);
```
primiano, line 460: add a comment to the proto where the bytecode versions are defined to update this place as well if we ever introduce a new one

src/tracing/service/tracing_service_impl.cc
```diff
@@ -324,6 +354,26 @@ std::optional<protozero::StringFilter::Policy> ConvertPolicy(
   return std::nullopt;
 }
 
+using StringFilterRule =
+    protos::gen::TraceConfig::TraceFilter::StringFilterRule;
+std::optional<protozero::StringFilter::SemanticTypeMask>
```
primiano, line 360: nit add a \n before

src/tracing/service/tracing_service_impl.cc
```diff
@@ -324,6 +354,26 @@ std::optional<protozero::StringFilter::Policy> ConvertPolicy(
...
+    return protozero::StringFilter::AllSemanticTypes();
+  }
+
+  protozero::StringFilter::SemanticTypeMask mask = {};
+  for (const auto& type : rule.semantic_type()) {
+    auto semantic_type = static_cast<uint32_t>(type);
+    if (semantic_type >= protozero::StringFilter::kSemanticTypeLimit) {
+      return std::nullopt;
+    }
+    mask[semantic_type / 64] |= 1ULL << (semantic_type % 64);
```
primiano: same as above, use constexpr kSemanticTypeBits = sizeof(SemanticTypeWord) * 8;

TBH there are 3 places already. I think you could extract at this point a function like

SetSemanticType(mask, uint32) which does this, raterht than repeating the bit math in all the places

PR #4115, opened 2025-12-13
ep: Make graph state persistent

aMayzner:
Adds local storage persistence for Explore Page graph state, preserving the sidebar width and collapse state across page reloads. This improves user
  experience by maintaining their preferred layout configuration.

  ## Changes

  - Extended `ExplorePageState` to include `isExplorerCollapsed` and `sidebarWidth` fields
  - Added `saveStateToLocalStorage()` and `loadStateFromLocalStorage()` functions to persist state
  - Updated serialization/deserialization to handle new layout fields
  - Refactored `Builder` component to use state from attrs instead of local instance variables for collapse/width
  - Modified `SqlModulesPlugin.onTraceLoad()` to properly await SQL modules initialization (required for state deserialization)

  ## Notes

  State is automatically saved to local storage on every state update and restored on page load. Invalid or corrupted state is automatically cleared from storage.

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/index.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/index.ts
@@ -55,7 +55,7 @@ function loadStateFromLocalStorage(trace: Trace): ExplorePageState | undefined {
 
     return deserializeState(json, trace, sqlModules);
   } catch (error) {
-    console.warn(
+    console.debug(
       'Failed to load Explore Page state from local storage:',
       error,
     );
@@ -80,6 +80,10 @@ export default class implements PerfettoPlugin {
   // Track whether we've successfully loaded state from local storage
   private hasAttemptedStateLoad = false;
 
+  // Track whether we've auto-initialized base JSON in this session
+  // This prevents reloading base JSON when clearing all nodes
+  private hasAutoInitialized = false;
+
   onStateUpdate = (
     update:
       | ExplorePageState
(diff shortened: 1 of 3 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/index.ts
```diff
@@ -19,6 +19,54 @@ import SqlModulesPlugin from '../dev.perfetto.SqlModules';
...
+
+/**
+ * Loads the Explore Page state from local storage.
+ * Returns undefined if no state is found or if deserialization fails.
+ */
+function loadStateFromLocalStorage(trace: Trace): ExplorePageState | undefined {
+  try {
+    const json = localStorage.getItem(LOCAL_STORAGE_KEY);
+    if (!json) {
+      console.warn('No local storage state found for Explore Page.');
```
stevegolton: I don't think we should warn if this is missing - e.g. the first time anyone opens the explore page this will be missing.

ui/src/plugins/dev.perfetto.SqlModules/index.ts
```diff
@@ -40,7 +40,7 @@ export default class implements PerfettoPlugin {
   }
 
   async onTraceLoad(trace: Trace): Promise<void> {
-    docs.then(async (resolvedDocs) => {
+    await docs.then(async (resolvedDocs) => {
```
stevegolton: I'm not sure we want to block here until this has finished as it can stall the trace load.

stevegolton (approved):

ui/src/plugins/dev.perfetto.ExplorePage/index.ts
aMayzner: Maybe debug?

PR #4116, opened 2025-12-15
Add file I/O ETW events to Perfetto UI

jessemckenna:
This change adds ETW file I/O events to the Perfetto UI.

These events consist of start events (e.g., FileIo_Create, FileIo_DirEnum) and end events, which can be matched to their corresponding start events by an I/O Request Packet (IRP) identifier. If a start event or end event without a match is found, its duration is set to zero to avoid misleadingly appearing to last until the end of the trace.

Some example screenshots of the modified UI follow.

Zoomed out (the arrows are zero-duration events):
Screenshot 2025-12-16 at 2 05 13 AM

Zoomed in:
Screenshot 2025-12-16 at 2 05 34 AM


src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -367,14 +457,13 @@ void FileIoTracker::ParseFileIoDirEnum(int64_t timestamp, ConstBytes blob) {
         if (info_class) {
           // If a string version of `info_class` is known, use that as the arg.
           // Otherwise, use its numerical value.
-          const auto info_class_str =
-              kFileInfoClasses.find(static_cast<FileInfoClass>(*info_class));
+          const char* info_class_str =
```
LalitMaganti: nit (optional): these strings can all be interned in the constructor and then just referenced here. Would suggest doing that for better performance. See the approach we use elsewhere.

src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -0,0 +1,711 @@
...
+          inserter->AddArg(context_->storage->InternString(kThreadIdArg),
+                           Variadic::UnsignedInteger(*ttid));
+        }
+        if (info_class) {
+          // If a string version of `info_class` is known, use that as the arg.
+          // Otherwise, use its numerical value.
+          const char* info_class_str =
+              GetFileInfoClassString(static_cast<FileInfoClass>(*info_class));
+          inserter->AddArg(
+              context_->storage->InternString(kInfoClassArg),
```
LalitMaganti: Same with the keys for all the arguments.

src/trace_processor/importers/etw/file_io_tracker_unittest.cc
```diff
@@ -0,0 +1,340 @@
...
+  event->set_file_object(file_object);
+  event->set_file_key(file_key);
+  event->set_ttid(ttid);
+  event->set_io_size(io_size);
+  event->set_io_flags(io_flags);
+  event->set_opcode(opcode);
+  return event.SerializeAsArray();
+}
+
+std::vector<uint8_t> CreateFileIoSimpleOpEvent(uint64_t irp_ptr,
```
LalitMaganti: Please create diff tests rather than unittests here. Diff tests are a lot less brittle and much easier to maintain

src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -0,0 +1,711 @@
...
+  EndEvent(timestamp, decoder.irp_ptr(), std::move(set_args));
+}
+
+void FileIoTracker::StartEvent(uint64_t irp, FileIoEvent event) {
+  // Store started events to be added to the UI when a matching end event is
+  // parsed. Only show events with a matching end event, because some ETW events
+  // can be lost/dropped if the memory buffers are too full during tracing, and
+  // a start event whose end event was dropped would appear to last until the
+  // end of the trace (which is misleading).
+  started_events_[irp] = event;
```
LalitMaganti: This is a signifcant violation of the trace processor data model: slices *msut* be inserted as they are parsed as there are assumtpions everywhere that the slice table contains data in timestamp order.

If you want to "discard" these events, you need to mark them as instant events or define a plugin to drop them at the UI level. Storing the timestamp for later insertion is not OK.

etiennep-chromium: I guess this can be solved by using slice_tracker->Begin/End and doing essentially the same thing as TrackEventEventImporter::ParseAsyncBeginEvent, but we will get unfinished events

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: Done

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: Done

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: Done - I opted to make them instant/zero-duration events.

jessemckenna:
Many thanks for the detailed review on this! I haven't written the diff tests yet (I'm hoping to add them tomorrow), but would you mind taking a look at the latest version in the meantime? I've tried to make it more permissive to avoid skipping events as much as possible, and it now displays unmatched start/end events with a zero duration per your recommendation. Feel free to let me know if anything still seems funky. Thanks again!

src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -0,0 +1,612 @@
...
+#include "src/trace_processor/importers/common/tracks.h"
+
+namespace perfetto::trace_processor {
+
+namespace {
+
+using protozero::ConstBytes;
+
+// The value of the "Category" field for file I/O events.
+constexpr char kCategory[] = "ETW File I/O";
```
LalitMaganti: You should not set the category field at all. The category field for your events should be null: category is designed for when events get multi-plexed on a single track not what you're doing here.

src/trace_processor/importers/etw/file_io_tracker.h
```diff
@@ -0,0 +1,238 @@
...
+  // Helper function to get the readable name of the event with `opcode`, if
+  // known.
+  std::optional<StringId> GetEventName(uint32_t opcode) const;
+
+  TraceProcessorContext* context_;
+
+  // Tracks events parsed so far for which a corresponding "operation end" event
+  // has not yet been parsed. This enables events with no matching end event to
+  // be closed with a zero duration at the end of parsing.
+  std::map<Irp, StartedEvent> started_events_;
```
LalitMaganti: Any reason all of these could not be arrays or, at least, unordered_maps. They seem unnecessarily expensive as it stands right now as map is implemented with trees which is O(logn).

src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -0,0 +1,612 @@
...
+  const auto previous_event_same_irp = started_events_.find(*irp);
+  if (previous_event_same_irp != started_events_.end()) {
+    // The last event using this IRP never ended. Since the IRP is being reused,
+    // the previous event must be done and its end event must have been dropped.
+    EndUnmatchedStart(*irp, previous_event_same_irp->second.timestamp);
+  }
+
+  // `track_id` controls the row the events appear in. This must be created via
+  // `TrackCompressor` because slices may be partially overlapping, which is not
+  // supported in `SliceTracker`.
```
LalitMaganti: It's not about SliceTracker but the Perfetto datamodel.

src/trace_processor/importers/etw/file_io_tracker.cc
etiennep-chromium: I guess apart from category_, this is used in tracks::StaticNameBlueprint of kBlueprint below, which we want to keep?

src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -0,0 +1,612 @@
...
+  SliceTracker::SetArgsCallback args =
+      [this, &decoder](ArgsTracker::BoundInserter* inserter) {
+        if (decoder.has_irp_ptr()) {
+          inserter->AddArg(irp_arg_, Variadic::Pointer(decoder.irp_ptr()));
+        }
+        if (decoder.has_file_object()) {
+          inserter->AddArg(file_object_arg_,
+                           Variadic::Pointer(decoder.file_object()));
+        }
+        if (decoder.has_ttid()) {
```
etiennep-chromium, line 230: instead of setting ttid as an arg, would it be possible to make the track thread scoped?
It seems this is done by adding a tracks::DimensionBlueprints in the blueprint, an example I could find: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/importers/proto/track_event_event_importer.h;l=667;drc=7666bc1983c2a5b98e5dc6fa6c28f8f53c07d06f;bpv=1;bpt=1

src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -0,0 +1,612 @@
...
+void FileIoTracker::EndEvent(std::optional<Irp> irp,
+                             int64_t timestamp,
+                             SliceTracker::SetArgsCallback args) {
+  if (!irp.has_value()) {
+    RecordEventWithoutIrp(event_types_.at(kEndOperation), timestamp,
+                          std::move(args));
+    return;
+  }
+
+  // Get the matching start event.
```
etiennep-chromium, line 495: This seems already handled by slice_track, and returned by slice_tracker->End():
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/importers/common/slice_tracker.cc;l=238;drc=7666bc1983c2a5b98e5dc6fa6c28f8f53c07d06f;bpv=1;bpt=1

src/trace_processor/importers/etw/file_io_tracker.h
```diff
@@ -0,0 +1,238 @@
...
+  const StringId missing_start_event_;
+  const StringId missing_end_event_;
+  const StringId unknown_event_;
+  const StringId dir_enum_event_;
+  const StringId info_event_;
+  const StringId read_write_event_;
+  const StringId simple_op_event_;
+
+  // Readable descriptions for the file I/O event types.
+  std::map<EventType, StringId> event_types_;
```
etiennep-chromium: Nit: const

src/trace_processor/importers/etw/file_io_tracker.h
```diff
@@ -0,0 +1,238 @@
...
+
+  // Readable descriptions for known "File Info" argument values.
+  std::map<FileInfoClass, StringId> file_info_classes_;
```
etiennep-chromium: Nit: const

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: Done (set category to null), and inlined the name

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: Done

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: Correct me if I'm wrong, but I think I need to get the matching start event anyway to get `name`, so moving this might be a bit awkward. Feel free to let me know if I'm misunderstanding.

src/trace_processor/importers/etw/file_io_tracker.h
jessemckenna: Since changing this to an array, I'm now initializing it in the constructor body.

src/trace_processor/importers/etw/file_io_tracker.h
jessemckenna: Ditto

src/trace_processor/importers/etw/file_io_tracker.h
jessemckenna: Done - I made `started_events_` into an `unordered_map`, and made `event_types_` and `info_classes_` into arrays.

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: Thanks for the code pointer! I was able to implement this (one row per thread under the ETW File I/O subheader), but for some reason it made queries in the UI (e.g., "slices with the same name") return 0 results, so I undid it for now. If you or Lalit happen to have any ideas for how I can fix that, there's a commit from that state that I can go back to if it makes sense.

Otherwise, though, I think this may need to be a follow-up task, because I think it may necessitate tinkering with the SQL queries.

Just for posterity, here's an example query from the UI that returns 0 results when the events are split into one row per thread:
```
INCLUDE PERFETTO MODULE viz.slices;
SELECT
_viz_slices_for_ui_table_0.id AS id,
_viz_slices_for_ui_table_0.ts AS ts,
_viz_slices_for_ui_table_0.dur AS dur,
_viz_slices_for_ui_table_0.category AS category,
_viz_slices_for_ui_table_0.name AS name,
_viz_slices_for_ui_table_0.utid AS utid,
thread_1.tid AS thread_91utid_93_46tid,
thread_1.name AS thread_91utid_93_46name,
_viz_slices_for_ui_table_0.upid AS upid,
process_2.pid AS process_91upid_93_46pid,
process_2.name AS process_91upid_93_46name,
_viz_slices_for_ui_table_0.track_id AS track_id,
track_3.name AS track_91track_id_93_46name,
__intrinsic_arg_set_to_json(_viz_slices_for_ui_table_0.arg_set_id) AS print_args_40arg_set_id_41,
_viz_slices_for_ui_table_0.arg_set_id AS arg_set_id
    FROM _viz_slices_for_ui_table AS _viz_slices_for_ui_table_0
    LEFT JOIN thread AS thread_1 ON thread_1.id = _viz_slices_for_ui_table_0.utid
LEFT JOIN process AS process_2 ON process_2.id = _viz_slices_for_ui_table_0.upid
LEFT JOIN track AS track_3 ON track_3.id = _viz_slices_for_ui_table_0.track_id
    WHERE
_viz_slices_for_ui_table_0.name = 'CreateFile'
```

src/trace_processor/importers/etw/file_io_tracker_unittest.cc
jessemckenna: Done

jessemckenna:
Thank you both for the thoughtful reviews - please take a look at the latest, and feel free to let me know if anything still needs work.

src/trace_processor/importers/etw/file_io_tracker.cc
LalitMaganti: Off the top of my head, cannot think why it would not work. Needs investigation

LalitMaganti (commented):
From my perspective seems reasonable now

src/trace_processor/importers/etw/file_io_tracker.cc
jessemckenna: I'll mark this resolved for the time being and follow up with another CL to split the data into one row per thread.

etiennep-chromium (approved):

PR #4117, opened 2025-12-15
Reapply "Revert "Migrate android_jank_cuj metric to use stdlib tables and local composition."". 

devianb:
Reapplying the android_jank_cuj metric migration change after adding fixes related to downstream errors. The frame duration calculation has changed, as the considered frame duration is the difference of the ts_end of the last actual frame and the start of the expected frame.

Bug: 400854991


src/trace_processor/metrics/sql/android/android_jank_cuj.sql
```diff
@@ -85,21 +290,21 @@ SELECT
               'frame_dur_ms_p90', PERCENTILE(f.dur / 1e6, 90),
               'frame_dur_ms_p95', PERCENTILE(f.dur / 1e6, 95),
               'frame_dur_ms_p99', PERCENTILE(f.dur / 1e6, 99))
-            FROM android_jank_cuj_frame_timeline f
+            FROM android_jank_cuj_frame_trace_data f
```
elrakadm, line 293: The timeline metrics were based only on perfetto timeline, while frame metrics were based on `do-frame` slices. is it now the same ?

src/trace_processor/metrics/sql/android/android_jank_cuj.sql
```diff
@@ -85,21 +290,21 @@ SELECT
...
             WHERE f.cuj_id = cuj.cuj_id),
           'frame', (
             SELECT RepeatedField(
               AndroidJankCujMetric_Frame(
                 'frame_number', f.frame_number,
-                'vsync', f.vsync,
+                'vsync', f.frame_id,
```
elrakadm, line 299: is frame id the same as the vsync or table related id ? if it is the same could it be renamed to vsync as well

test/trace_processor/diff_tests/metrics/graphics/android_jank_cuj.out
```diff
@@ -410,13 +410,13 @@ android_jank_cuj {
       missed_app_frames: 7
       missed_sf_frames: 2
       frame_dur_max: 61000000
-      frame_dur_avg: 27192307
```
elrakadm, line 413: Should this change be no-op because it is a migration to stdlib and tests shouldn't change?

src/trace_processor/metrics/sql/android/android_jank_cuj.sql
```diff
@@ -13,14 +13,219 @@
...
+INCLUDE PERFETTO MODULE android.frames.timeline;
 
--- Creates a table that matches CUJ counters with the correct CUJs.
--- After the CUJ ends FrameTracker emits counters with the number of total
--- frames, missed frames, longest frame duration, etc.
--- The same numbers are also reported by FrameTracker to statsd.
-SELECT RUN_METRIC('android/jank/internal/counters.sql');
+-- Table captures additional data related to each frame in a CUJ. This information includes
+-- data like missed type of jank, missed app/sf frame, missed sf/hwui callbacks etc.
+DROP TABLE IF EXISTS _android_jank_cuj_frames_data;
```
elrakadm, line 25: These tables look similar to the tables created by `trace_processor/metrics/sql/android/jank/frames.sql`, should that script be moved to stdlib before removing the dependency on  `android_jank_cuj_init` ?

src/trace_processor/metrics/sql/android/android_jank_cuj.sql
devianb: In the new table we are keeping a more accurate definition of a frame, so the ts and dur change:

the `ts_end` is the end ts of the last actual frame.
And the `ts`  is `do_frame.ts`(when expected frame ts is NULL) OR `MAX(do_frame.ts_prev_do_frame_end, timeline.ts_expected)`. 

Hence this changes the test output a little as well.

test/trace_processor/diff_tests/metrics/graphics/android_jank_cuj.out
devianb: Explained in previous comment.

src/trace_processor/metrics/sql/android/android_jank_cuj.sql
devianb: This is deliberately kept in the metric file itself. For stdlib migration, we decided that we do not want to move all tables to stdlib, but only the ones that are absolutely necessary. We tried to keep the table composition to minimum possible in stdlib.
And whichever metrics need specific data, the composed tables should reside in the metric file itself.

src/trace_processor/metrics/sql/android/android_jank_cuj.sql
elrakadm: The timeline metrics currently is showing data only from the perfetto frame timeline only. there are some frames where the do frame calls is missing or do frame call exist with timeline frame missing due to data loss .. etc, and there are some quality metrics on that.
 
1. With this change, the `timeline_metrics`  output of the `android_jank_cuj` is exactly the same as `trace_metrics` output so there is no need to have two similar outputs, so I prefer to keep the `timeline_metrics` as is until it is deprecated properly

2. On changing the frame metrics duration and ts, I would prefer to split the change into two changes [1. migrating to stdlib (no-op), 2. changing the duration definition] this makes it easier to rollback specific change if there were any issues downstream and also discuss/review each change individually.

elrakadm:
Could you please change the title of the pull request ? I think the change is now different from the original cl and its revert

src/trace_processor/metrics/sql/android/android_jank_cuj.sql
elrakadm: It is fine to not migrate everything to stdlib. But `android_jank_cuj.sql` didn't create any frame table/data. The frame tables are created by `jank/frames.sql` so it feels th new tables should be added/updating the frames script not the main metric

1. Now it is hard to understand the difference between the new table `android_jank_cuj_frame_trace_data` and the table `android_jank_cuj_frame` created by the `jank/frames.sql`.
2.  it might be breaking change for metrics that depend on `android_jank_cuj.sql` metric (which creates `android_jank_cuj_frame` indirectly at the moment via the dependency on `android_jank_cuj_init`. which will raise the question for the dependents on whether they should
a. switch to depend on `android_jank_cuj_frame_trace_data` 
b. depend directly on `jank/frames.sql` metric to have `android_jank_cuj_frame` created?

elrakadm (changes requested):

PR #4118, opened 2025-12-15
ui: DataGrid: Clean up pivot table model and tweak drilldown column button styling

stevegolton:
1. Clean up the model and fix a handful of bugs:
  - Put sorting information directly on the column definition - this means that sorting on a column which doesn't exist is simply not supported by the model.
  - Disallow filtering on aggregate columns (i.e. 'having' filters) - this is no currently supported by the existing pivot table anyway so it's not a regression.
  - Make sure all the column names are formatted properly wherever they are listed: e.g. filters, pivot groups, menus, etc.

2. Put the drill down button in a virtual column between the groupby rows and the aggregate rows that appears on row hover. This serves double duty as a wide and and obvious separator between the two sides of the pivot table, and moves the drilldown button closer to the groupby columns compared to having it on the end.

See:

image

### DIFF
--- a/ui/src/assets/widgets/grid.scss
+++ b/ui/src/assets/widgets/grid.scss
@@ -255,9 +255,9 @@ $indent-guide-offset: 12px;
       }
     }
 
-    // &--border-right-thick {
-    //   border-right: $border-3;
-    // }
+    &--border-right-thick {
+      border-right: $border-3;
+    }
 
     // Drag indicator shown on all cells in the target column (header and body)
     &--drag-over-before::after,


ui/src/assets/widgets/grid.scss
```diff
@@ -242,9 +255,9 @@ $indent-guide-offset: 12px;
       }
     }
 
-    &--border-right-thick {
-      border-right: $border-3;
-    }
+    // &--border-right-thick {
```
LalitMaganti: nit: remove?

PR #4122, opened 2025-12-15
[SDK] Add optional scoped string option to Flow

etiennep-chromium:
This PR adds optional string to scope Flows' id, similar to NamedTrack.
In chrome, usage of legacy TRACE_EVENT_WITH_FLOW is common with TRACE_ID_WITH_SCOPE using a string
to scope the flow id. This offers similar option with new SDK.
This also combines Flow/TerminatingFlow implementation into a common class to avoid code duplication.


LalitMaganti:
> This also combines Flow/TerminatingFlow implementation into a common class to avoid code duplication.

Can we please do this in a separate change? Right now this is making it quite hard to review.

Also I'd like to see the TP side for this change as well at the same time.

etiennep-chromium:
I created https://github.com/google/perfetto/pull/4123 although I didn't figure out how to do chain in github

> Also I'd like to see the TP side for this change as well at the same time.

There's no change needed in TP or protos; this is really just a convenience method to hash strings that goes into flow_id.

include/perfetto/tracing/track_event_args.h
```diff
@@ -21,76 +21,86 @@
...
-    return Flow(flow_id);
+  static PERFETTO_ALWAYS_INLINE inline FlowImpl Global(uint64_t flow_id) {
+    return FlowImpl(flow_id);
+  }
+
+  // Same as above, but combines the flow id with an extra `named_scope`'s hash.
+  static PERFETTO_ALWAYS_INLINE inline FlowImpl Global(
+      uint64_t flow_id,
+      const char* named_scope) {
+    return FlowImpl(internal::Fnv1a(named_scope) ^ flow_id);
```
sashwinbalaji: minor nit: flow_id ^ internal::Fnv1a(named_scope) so every xor is aligned.

sashwinbalaji (approved):

include/perfetto/tracing/track_event_args.h
etiennep-chromium: Done
Also I changed other overloads to reuse this one.

PR #4128, opened 2025-12-16
fix build

sashwinbalaji:
Internal autorolls failing with error and this header is not needed so simple remove it

```
trace_buffer_v2.cc:27:10: error: use of private header from outside its module: '...src/base/intrusive_list.h'; [-Wprivate-header]
   27 | #include "third_party/perfetto/src/base/intrusive_list.h"
      |          ^
1 error generated.
  ```


sashwinbalaji:
updating the gn files

PR #4129, opened 2025-12-16
tracing: split out TracingServiceImpl

primiano:
This CL is a pure non-functional split of TracingServiceImpl
into several files. TSI has become too big for humans... and now
even for AI (which starts warning about the file being too large).
This CL is very trivial (I have been Cut/Paste chunks of code) but
is fully human written because I don't yet trust AI copying large
chunks of code without being creative (and the diff would be
impossible for a human to review because of the file moves)

I split it as follows:
- tracing_service_structs: holds all the auxiliary structs
  that are used across different subclasses (e.g. TriggerInfo,
  PendingClone, etc)
- tracing_service_endpoints: the boring proxy layers that
  bridge the lowest-level ext API and calls into real methods
  of TSI. I find these mostly annoying because when I codesearch
  they show up while I am looking for the real impl in TSI.
- tracing_session: it's a pure struct, but it's big enough and
  changed/read enough that warrants its own file.

The only non-copy-paste change I made has been outlining a
few functions in tracing_service_sesison.cc which used to be
inline for no good reasons.

### DIFF
--- a/src/tracing/service/tracing_service_endpoints_impl.cc
+++ b/src/tracing/service/tracing_service_endpoints_impl.cc
@@ -16,6 +16,7 @@
 
 #include "src/tracing/service/tracing_service_endpoints_impl.h"
 #include "perfetto/base/task_runner.h"
+#include "perfetto/ext/base/file_utils.h"
 #include "perfetto/ext/base/metatrace.h"
 #include "perfetto/ext/base/string_utils.h"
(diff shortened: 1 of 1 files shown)


primiano:
@LalitMaganti this is a hand-crafted cut/paste CL. I don't trust AI enough to do big moves without being creative.

PR #4130, opened 2025-12-17
ui: improve performance of visualizing arg values signifcantly

LalitMaganti:
* remove poor performing SQL which tried to compute depth: this is not
  necessary as the UI now does this for you automatically.
* materialize the dataset for arg values making working with a lot
  faster.
* use distinct where appropriate to speed up queries signifcantly


stevegolton (approved):
Nice!

PR #4131, opened 2025-12-17
tracing: Allow the use of TBV2 behind a config flag

primiano:
Add an experimental per-buffer flag that allows opting in.
Name the flag experimental as the end goal is to deprecate it
and move everything over to TBV2.

Bug: b/447426810

**Stack:**
- [#4129](https://github.com/google/perfetto/pull/4129) (svc_cleanup)
  - **[This PR] (tbv2_flag)**


LalitMaganti (approved):

primiano:
Actually let's put this on hold. I might change the boolean into an enum mode, so I can have:

- MODE_DEFAULT (v1)
- MODE_TBV2
- MODE_TBV1_WITH_TBV2_CROSSCHECK

PR #4132, opened 2025-12-17
tp: improve shell error message to point towards -Q

LalitMaganti:


sashwinbalaji (approved):
side note: I guess too late to change but -q mapping to file and -Q mapping to string seems confusing

LalitMaganti:
Far too late. This is my way of trying to patch that mistake as much as possible.

PR #4134, opened 2025-12-17
tp: add R8 retracing support and simplify deobfuscator

LalitMaganti:
R8 recently embarked on a "retracing" project which uses line numbers
to make ambiguous obfuscated method names fully deterministic when
deobfuscated. On my request, they added reference tests upstream to
verify implementations were correctly doing the deobfuscation.

This CL:
1) Ports the R8 reference tests into our codebase, split between
   deobfuscation map parsing tests (unit tests) and actual
   deobfuscation tests (diff tests)
2) Adds support for line number based deobfuscation and inline chain
   detection in deobfuscation_tracker.cc
3) Simplifies the deobfuscator code significantly:
   - MethodMapping now stores a single fully-qualified deobfuscated_name
     instead of separate class/method fields
   - Removed dual storage (deobfuscated_methods_ map) in ObfuscatedClass,
     now derives everything from method_mappings_
   - deobfuscated_methods() now handles R8 inline chains and joins
     ambiguous names with " | " format

In practice, for heap dumps we cannot use line-based deobfuscation
(since we don't have line numbers). For stack samples we could
potentially use this (since we have the PC), but that requires further
pipeline work.

### DIFF
--- a/src/trace_processor/importers/proto/deobfuscation_tracker.cc
+++ b/src/trace_processor/importers/proto/deobfuscation_tracker.cc
@@ -208,15 +208,15 @@ void DeobfuscationTracker::DeobfuscateProfiles(
   auto symbol_cursor = context_->storage->symbol_table().CreateCursor({
       dataframe::FilterSpec{
           tables::SymbolTable::ColumnIndex::symbol_set_id,
-          tables::SymbolTable::ColumnIndex::symbol_set_id,
+          0,
           dataframe::Eq{},
-          std::nullopt,
+          {},
       },
       dataframe::FilterSpec{
           tables::SymbolTable::ColumnIndex::line_number,
-          tables::SymbolTable::ColumnIndex::line_number,
+          1,
           dataframe::IsNotNull{},
-          std::nullopt,
+          {},
       },
   });
   // Deobfuscate frames using the collected mappings.


LalitMaganti:
@rsavitski might be of interest but please feel free to ignore, no action needed from you.

src/trace_processor/importers/proto/deobfuscation_tracker.cc
```diff
@@ -145,44 +159,146 @@ void DeobfuscationTracker::DeobfuscateProfiles(
...
-      for (tables::StackProfileFrameTable::Id frame_id : frames) {
-        auto* frames_tbl =
-            context_->storage->mutable_stack_profile_frame_table();
+    for (tables::StackProfileFrameTable::Id frame_id : frames) {
+      auto frame = frames_tbl->FindById(frame_id);
+      if (!frame) {
+        continue;
+      }
+
+      // Try to get line number from existing symbol entry (from symbolization).
```
ilkos: I'm a bit confused with this? How is native symbolization related to the java deobfuscation - they are entirely orthogonal - iirc the symbol table was a product of adding symbol packets.

Is this how you are modelling line numbers coming from non-libunwindstack profiling? Can this be clearly stated in the comments?

src/trace_processor/importers/proto/deobfuscation_tracker.cc
```diff
@@ -145,44 +159,146 @@ void DeobfuscationTracker::DeobfuscateProfiles(
...
+      std::optional<uint32_t> obfuscated_line;
+      if (frame->symbol_set_id().has_value()) {
+        const auto& symbol_tbl = context_->storage->symbol_table();
+        for (auto sym_it = symbol_tbl.IterateRows(); sym_it; ++sym_it) {
+          if (sym_it.symbol_set_id() == *frame->symbol_set_id()) {
```
ilkos: Is it worth creating a lookup for the symbols table given this will be quadratic across all frames?

PR #4144, opened 2025-12-17
perfetto_cmd: extract `TraceOutput` as a separate struct

KirillTim:
There are multiple ways to create a trace output file, we decouple the reference to an actual file
from the configuration passed to the `perfetto_cmd`.

This PR also unifies the cleanup behaviour for the output files created by `perfetto_cmd` and `traced`.


primiano:
Sorry I cannot tell which part of the CL is a pure refactor and which is a behavioural change, and as such makes me a little bit nervous. if I had to review this properly, in the way it's written it would take me half a day, but i'm heading off :/

in general these sort of CLs are easier to review if you split them cleanly into 
1. " a CL that has 0 intended behavioural changes, and brings the codebase in a state that is nicer fo ryour goals" 
2. A cl that introduces the behavioral change (hopefully smaller, so it's easier to reason about)
this cl is a bit of a mix, and is scary.
sorry, as-is I donj't want to land it now, because it's unflagged code. 
if there is any abug (which i am not able to prove without spending several hours on it) this will jeopardize completely tracing in android across all branches. too risky


There is something else: when we discussed offline, the impression I got is that this would have made the code somehwat easier to read/reason about. But from what I can tell, this seems to just add an indirection layer (plus some behavioural changes) so... i'm not fully sure I understand anymore what was the intend.

Can I ask you to expand a bit on "why are we doing this? what does it make better by grouping path and stream into one struct ?"

PR #4146, opened 2025-12-17
ep: Center the graph

aMayzner:
Adds automatic graph centering on load and a manual recenter button to the ExplorePage node graph. When loading a graph (either from scratch or from saved state), the view now automatically centers to show all nodes and labels. Users can also manually recenter using a new button in the graph toolbar.

  - Added automatic centering on initial graph load and after state imports
  - Added manual recenter button to graph toolbar
  - Enhanced `autofit()` function to include labels in bounding box calculations
  - Implemented callback pattern (`onRecenterReady`) to expose recenter functionality from child components to parent
  - Fixed initial centering to work both for new graphs (with auto-layout) and loaded graphs (with existing positions)

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/graph/graph.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/graph/graph.ts
@@ -603,6 +603,7 @@ export class Graph implements m.ClassComponent<GraphAttrs> {
   private nodeGraphApi: NodeGraphApi | null = null;
   private hasPerformedInitialLayout: boolean = false;
   private hasPerformedInitialRecenter: boolean = false;
+  private recenterRequired: boolean = false;
   private labels: Label[] = [];
   private labelTexts: Map<string, string> = new Map();
   private editingLabels: Set<string> = new Set();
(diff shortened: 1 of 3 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/graph/graph.ts
```diff
@@ -768,14 +780,29 @@ export class Graph implements m.ClassComponent<GraphAttrs> {
...
     ) {
       this.hasPerformedInitialLayout = true;
-      // Defer autoLayout to next tick to ensure DOM nodes are fully rendered
-      setTimeout(() => {
+      this.hasPerformedInitialRecenter = true;
+      // Call autoLayout to arrange nodes hierarchically
+      // autoLayout will call onNodeMove for each node it repositions
+      this.nodeGraphApi.autoLayout();
+      // Recenter after DOM updates with new positions
+      requestAnimationFrame(() => {
```
stevegolton: Using requestAnimationFrame here feels a bit off to me - its basically a timing hack hoping the DOM is ready by the next frame but theres no guarantee of that.

Could we instead change NodeGraph to call onReady after every render cycle - basically just call it in onupdate() as well as oncreate? We could rename it to onRenderComplete() or something if that feels better. Then you can check if the graph needs a recenter in this callback and just execute it directly on the api object passed in the callback.

```ts
m(NodeGraph, {
  onReady: (api) => {
    if (this.recenterRequired) {
      api.recenter();
      this.recenterRequired = false;
    }
  },
});
```

PR #4147, opened 2025-12-17
Disable failing TraceBufferV2Test tests

etiennep-chromium:
Link to UBSAN failure in chrome: https://ci.chromium.org/ui/p/chromium/builders/ci/Linux%20UBSan%20Tests/12411/test-results

b/469775021

PR #4153, opened 2025-12-18
tracing: fix and re-enable TBV2 tests

primiano:
Revert #4147 which disabled some failing UBSan tests.
Also fix ubsan (however our ubsan build did NOT catch
this anyways, because we don't use an annotated sysroot
like chromium does)

Test: I was able to repro adding a CHECK(src) in trace_buffer_v2.cc This makes it go away.


primiano:
@etiennep-chromium I couldn't repro (because our UBSan build doens't use the same sysroot of chromium, which seems to have extra annotation around memcpy) but I am 90% confident this was the issue (i could repro adding a CHECK)

etiennep-chromium:
Ack thanks!

PR #4160, opened 2025-12-19
sdk: Add ClearIncrementalState method to data source traits

LalitMaganti:
This change adds a new optional ClearIncrementalState method to data
source traits that allows clearing incremental state without destroying
and recreating the state object. This is useful for data sources like
track event that use large data structures (e.g., hash maps) which are
expensive to reallocate.

We also plumb this feature through to the C and Rust SDKs in the
appropriate ways for each of them.

We also implement this for the TrackEvent Data sources: this should
significantly reduce the amount of memory allocations when aggressive
intern clearing is enabled.


include/perfetto/public/data_source.h
```diff
@@ -83,19 +84,12 @@ struct PerfettoDsParams {
...
-                                 PERFETTO_NULL,
-                                 PERFETTO_DS_BUFFER_EXHAUSTED_POLICY_DROP,
-                                 false,
-                                 true};
+  struct PerfettoDsParams ret = {
+      PERFETTO_NULL, PERFETTO_NULL, PERFETTO_NULL,
+      PERFETTO_NULL, PERFETTO_NULL, PERFETTO_NULL,
+      PERFETTO_NULL, PERFETTO_NULL, PERFETTO_NULL,
+      PERFETTO_NULL, PERFETTO_NULL, PERFETTO_DS_BUFFER_EXHAUSTED_POLICY_DROP,
+      false,         true};
```
sashwinbalaji: can you also please add Commented Parameters for each, would be really helpful.

include/perfetto/tracing/data_source.h
```diff
@@ -549,6 +565,44 @@ class DataSource : public DataSourceBase {
     return nullptr;
   }
 
+  // Detection idiom for checking if DataSourceTraits::ClearIncrementalState(T*)
+  // exists. Supports traits that don't inherit from DefaultDataSourceTraits.
+  template <typename Traits, typename T, typename = void>
+  struct HasClearIncrementalState : std::false_type {};
```
sashwinbalaji, line 571: nit: variable template ?

```
  template <typename Traits, typename T, typename = void>
  static constexpr bool HasClearIncrementalState = false;
```

contrib/rust-sdk/perfetto/src/data_source.rs
```diff
@@ -292,8 +295,30 @@ impl DataSourceArgsBuilder {
...
+    ///
+    /// # Example
+    /// ```ignore
+    /// .on_clear_incr(|state: &mut MyIncrementalState| {
+    ///     state.clear();
+    ///     true
+    /// })
+    /// ```
+    #[must_use = "Builder methods return an updated builder; use the returned value or keep chaining."]
+    pub fn on_clear_incr<F>(mut self, cb: F) -> Self
```
dreveman: I prefer using a trait to expose this functionality instead of a callback in the Rust SDK. That would match how we handle initialization of incremental state and I think also make it easier to use.

ie. add something like this above:
```
pub trait Clear: Default {
    fn clear(&mut self) {
        // Replaces the current instance with a new default instance.
        *self = Self::default();
    }
}

impl Clear for IncrementalState {}
```

and then change `IncrT: Default` to `IncrT: Default + Clear` in a few places and unconditionally use that Clear trait in `on_clear_incr_trampoline`.

This means Rust SDK always supports clear but it's not much of a burden on the user as if they don't want their custom incremental state to have a custom version of this trait they can just add `impl Clear for MyCustomIncrementalState {}` and it will use the Default trait which is already required.

This is technically a breaking change to the Rust SDK so we'll need to make sure to update the `perfetto-sdk` crate version in `Cargo.toml`.  I think that's fine as the usage of the Rust SDK is limited today and the usage of custom incremental state even more so.

LalitMaganti:
Thanks both! All comments should be addressed.

contrib/rust-sdk/perfetto/src/data_source.rs
```diff
@@ -194,22 +194,23 @@ struct DsCallbacks {
...
 #[derive(Default)]
 #[must_use = "This is a builder; remember to call `.build()` (or keep chaining)."]
-pub struct DataSourceArgsBuilder {
-    args: DataSourceArgs,
+pub struct DataSourceArgsBuilder<IncrT: Default + Clear = IncrementalState> {
+    args: DataSourceArgs<IncrT>,
 }
 
-impl DataSourceArgsBuilder {
+impl<IncrT: Default + Clear> DataSourceArgsBuilder<IncrT> {
```
dreveman: is this needed? Seems like it can be removed now that we have the Trait

PR #4162, opened 2025-12-19
ui: DataGrid: Add support for compound columns

stevegolton:
Compound columns are columns that depend on multiple fields in the underlying datasource.

When a column has a dependsOn property, those fields are requested from the datasource as well as well as the referenced field. This means that all these fields will be present in the 'row' parameter passed to the cellRenderer() so that all these fields can be used to format the cell.

Compelling example: build a process column which combines upid, process.name, and process.pid into one column.

E.g. `[<upid>] <process.name> (<process.pid>)`

### DIFF
--- a/ui/src/components/widgets/datagrid/datagrid.ts
+++ b/ui/src/components/widgets/datagrid/datagrid.ts
@@ -393,12 +393,11 @@ export class DataGrid implements m.ClassComponent<DataGridAttrs> {
 
       // Check aggregate columns (those with a field) for dependencies
       for (const agg of this.pivot.aggregates ?? []) {
-        if ('field' in agg) {
-          const colInfo = getColumnInfo(schema, rootSchema, agg.field);
-          if (colInfo?.dependsOn) {
-            for (const dep of colInfo.dependsOn) {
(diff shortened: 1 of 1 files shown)


ui/src/components/widgets/datagrid/datagrid.ts
```diff
@@ -364,10 +365,57 @@ export class DataGrid implements m.ClassComponent<DataGridAttrs> {
...
+        if (colInfo?.dependsOn) {
+          for (const dep of colInfo.dependsOn) {
+            dependencyFields.add(dep);
+          }
+        }
+      }
+
+      // Check aggregate columns (those with a field) for dependencies
+      for (const agg of this.pivot.aggregates ?? []) {
+        if ('field' in agg) {
```
aMayzner: Can you refactor it to fieldName? Also, maybe it can be fieldName?: string?

stevegolton: Refactored to use the function as a tag for this union type rather than the presence of 'field' which is strictly better. Renaming to fieldname would be a larger refactor so I'll tackle in an upcoming CL.

PR #4163, opened 2025-12-19
ui: Add slice links to source slices in aggregate drilldown views

stevegolton:
For now this is only supported in the following aggregation panels:
- slice
- sched
- thread_state


ui/src/plugins/dev.perfetto.AggregationBenchmark/benchmark_tab.ts
```diff
@@ -0,0 +1,205 @@
+// Copyright (C) 2025 The Android Open Source Project
```
aMayzner, line 1: Can we just have a CoreBenchmarks, not a separete AggregationBenchmark and other benchmarks?

stevegolton: Renamed to dev.perfetto.Benchmarks and added a description.

Edit: Well, I was going to but then it automerged. Next time!

PR #4174, opened 2025-12-19
tp: interval_merge_overlapping_partitioned SQL macro

aMayzner:
Adds a new stdlib macro that merges overlapping intervals within each partition group to generate minimal covering sets. This is useful for consolidating overlapping time ranges (e.g., scheduling events, resource allocations) grouped by partition keys.

The implementation reuses the existing intervals_overlap_count_by_group infrastructure, tracking when the overlap counter transitions to/from zero to determine merged interval boundaries.


src/trace_processor/perfetto_sql/stdlib/intervals/overlap.sql
```diff
@@ -520,3 +520,71 @@ RETURNS TableOrSubquery AS
   WHERE
     value = 0
 );
+
+-- Merge overlapping intervals within each partition group to generate a minimum
```
LalitMaganti, line 524: Please make this private for now.

PR #4179, opened 2025-12-19
ui: Fix process summary mipmap in multi-VM traces

jahdiel-alvarez:
The ProcessSchedulingTrack uses the cpu id to calculate the offset of the cpu track in the process summary track mipmap. In multi-VM traces this result in incorrectly setting the verticall offset because the cpu ids could start at an arbitrarily high number (e.g. 4096) for the secondary machines.

This is fixed by retreiving the mininum cpu id for each machine and subtracting that min id when calculating the vertical offset. Doing this re-establishes the indexing at 0 which results in the correct calculation.

Bug: 470195101


LalitMaganti:
Superseded by https://github.com/google/perfetto/pull/4180 so closing

PR #4184, opened 2025-12-22
ui: Add download button to CodeSnippet

camillobruni:
Add a download to the CodeSnippet widget so we can directly download perfetto trace config files.
External tools like crossbench can use the trace config file directly and thus being able to directly save the file contents improves ergonomics.

- Implement DownloadToFileButton
- Move common code from CopyButtonHelper to new ActionButtonHelper parent class
- Add downloadFileName to CodeSnippet to conditionally enable the download button 
- Use downloadFileName in InstructionsPage and TraceInfoPage


ui/src/widgets/copy_to_clipboard_button.ts
```diff
@@ -14,56 +14,20 @@
...
-  constructor(timeout = 2000, loadingDelay = 100) {
-    this.timeout = timeout;
-    this.loadingDelay = loadingDelay;
-  }
-
-  get state(): CopyState {
-    return this._copyState;
-  }
-
+export class CopyButtonHelper extends ActionButtonHelper {
```
stevegolton: I'm not sure this entity is really required any more. Could you switch to using the ActionButtonHelper directly instead in the two places CopyButtonHelper is used?

camillobruni: DONE

PR #4185, opened 2025-12-22
ui: Improve RecordingCtl menu 

camillobruni:
The current record menu is a bit small compared to all the other menu items and it introduced another button size to the side panel.

- Use same button size for RecordingCtl buttons as for the menu items
- Use the icon of the currently selected source
- Allow two lines of device names before cropping it with elipses

New: long device description
Screenshot 2025-12-22 at 17 34 31

New:  single-line device description
Screenshot 2025-12-22 at 17 35 59

### DIFF
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/styles.scss
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/styles.scss
@@ -61,10 +61,10 @@
       visibility: hidden;
     }
     .record-target {
-      font-size: 14px;
+      font-size: var(--pf-font-size-m);
       overflow: hidden;
       text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.RecordTraceV2/styles.scss
```diff
@@ -52,27 +52,38 @@
...
-    line-height: 24px;
+    height: 50px;
+    line-height: 50px;
     .hidden {
       visibility: hidden;
     }
     .record-target {
-      font-size: 13px;
-      color: var(--pf-color-text-muted);
+      font-size: 14px;
```
stevegolton: Should we use your new font size variables here?

PR #4187, opened 2025-12-23
Add relevant threads jank CUJ module to stdlib.

devianb:
As part of migrating the android_jank_cuj metric to stdlib, move the dependent modules to stdlib first.

Bug: 400854991

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/cujs/threads.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/cujs/threads.sql
@@ -1,0 +1,65 @@
+--
+-- Copyright 2025 The Android Open Source Project
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     https://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/threads.sql
```diff
@@ -0,0 +1,64 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+INCLUDE PERFETTO MODULE android.cujs.base;
+
+-- Function returns thread information for CUJs for a given thread name.
```
nicomazz: I'm a bit confused on why we would need this function: this seems to just add the track id, but it's fine as this cl is just moving it for the sake of a refactor. 

Can you improve this doc a bit though? probably something like 

`-- Returns a table with all CUJs and an additional column for the track id of thread_name passed as parameter, if present in the same process of the cuj`

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/cujs/threads.sql
devianb: Done.

PR #4192, opened 2025-12-27
ui: implement support for server extensions

LalitMaganti:
This is an initial implementaton of RFC-0005
(https://github.com/google/perfetto/discussions/3227).


ui/src/core/app_impl.ts
```diff
@@ -106,6 +106,13 @@ export class AppImpl implements App {
   // Promise which is resolved when extra loading is completed.
   extrasLoadingDeferred = defer<undefined>();
 
+  // Deferred objects for extension server resources. These are resolved when
+  // extension loading completes. Consumers should await these promises to ensure
+  // the corresponding extra* arrays are populated before using them.
+  extensionMacrosDeferred = defer<void>();
+  extensionSqlModulesDeferred = defer<void>();
+  extensionProtoDescriptorsDeferred = defer<void>();
+
```
primiano: if you do it this way this will 100% lead to bugs, as this, as-is, is a pure gentlement agreement for "you are supposed to await the xxxDeferred before looking at the actual data thing"

Imho it's way more robust if the deferred thing carries the data you need, (so it's like extensionMacrosDeferred = defer<ExtensionMacros>()) so you can't possibly miss the await

ui/src/core/load_trace.ts
```diff
@@ -144,6 +144,10 @@ async function createEngine(
   if (app.httpRpc.newEngineMode === 'USE_HTTP_RPC_IF_AVAILABLE') {
     useRpc = (await HttpRpcEngine.checkConnection()).connected;
   }
+
+  // Wait for extension proto descriptors to load before using them
+  await app.extensionProtoDescriptorsDeferred;
```
primiano: I have a bit of a question on whether we should do these fine-grained deferred things.
On one side, they avoid blocking for a big load before you can use the UI.
OTOH they create large opportunities for "the UI silently stops working in the middle of somewhere" if a network fetch fails.

If you want to break down and do fine-grained deferred, at very least they should be wrapper in something that has a "orTimeout" so we don't hang forever, and give a soft-error in the omnibox (and in future in analytics)

ui/src/core_plugins/dev.perfetto.CoreCommands/index.ts
```diff
@@ -241,9 +241,10 @@ export default class CoreCommands implements PerfettoPlugin {
     // Register the macros from extras at onTraceReady (the latest time
     // possible).
     ctx.onTraceReady.addListener(async (_) => {
-      // Await the promise: we've tried to be async as long as possible but
+      // Await the promises: we've tried to be async as long as possible but
       // now we need the extras to be loaded.
       await app.extraLoadingPromise;
+      await app.extensionMacrosDeferred;
```
primiano: (reinforcing my previous comment) this programming model is unsustainable. there are magic awaits. Try to open this code in 3 months and go "why there is this await here? it seems to work if I remove it"

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
+// Copyright (C) 2024 The Android Open Source Project
```
primiano: here and elsewhere: 2026

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+  CommandInvocation,
+  commandInvocationArraySchema,
+} from '../../core/command_manager';
+
+// =============================================================================
+// Zod Schemas (source of truth)
+// =============================================================================
+
+/**
+ * A single proto descriptor entry.
```
primiano: Can you either keep the schemas in order (from top-level to nested) or add a comment with some sort of tree that explains how everything fits with each other.
it's not obvious just by reading this file.
Also this one itself deserves a comment to epplain what this is, what this is for and how it fits in the overall tree.
it's not obvious to 90% of people what a proto descriptor does and is for.

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+ */
+export const ProtoDescriptorSchema = z.object({
+  name: z.string(),
+  description: z.string().optional(),
+  descriptor: z.string(),
```
primiano: Is this base64 encoded?
If yes, say in the comment.
If not... are you sure it can work reliably in JS?

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+  name: z.string(),
+  modules: z.array(z.string()),
+  csp_allow: z.array(z.string()).optional(),
+});
+
+/**
+ * Macros format from {base_url}/modules/{module}/macros
+ * Maps macro names to command sequences.
+ */
+export const MacrosSchema = z.record(commandInvocationArraySchema);
```
primiano: No, this needs to be its own type defined in here (copy-pasting from the existing core.ts definition).
If you do what yo udid here, the day that Steve or someboy else changes something in that commandInvocationArraySchema, it won't realize that he is implicitly changing the schema of the server extension, which is supposed to be stable.
This is very subtle.

Instead if you make a copy here, the moment somebody touches commandInvocationArraySchema, compilation will fail on the line that passes the argument from this to runCommand

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * SQL Modules format from {base_url}/modules/{module}/sql_modules
+ * Maps module paths (e.g., "android.startup") to SQL content.
+ */
+export const SqlModulesSchema = z.record(z.string());
```
primiano: isn't this supposed to be:
z.object(modules: z.array(z.record(z.string(), z.string())));

and so on with the others, to reflect the outer dict you have?

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+ */
+export const ExtensionServersSchema = z.array(ExtensionServerSchema);
+
+/**
+ * Manifest file format from {base_url}/manifest.json
+ * Provides server metadata and available modules.
+ */
+export const ManifestSchema = z.object({
+  name: z.string(),
+  modules: z.array(z.string()),
```
primiano, line 59: say that this specifies a set of modules, and for each of them the client is supposed to fetch:

{base_url}/modules/${module}/macros -> using MacrosSchema
{base_url}/modules/${module}/sql_modules -> using SqlModuleSchema
{base_url}/modules/${module}/descriptors -> using ...

ui/src/core_plugins/dev.perfetto.ExtensionServers/url_utils.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  // HTTPS URLs pass through unchanged
+  if (trimmed.startsWith('https://')) {
+    return trimmed;
+  }
+
+  // HTTP URLs are upgraded to HTTPS
+  if (trimmed.startsWith('http://')) {
+    return trimmed.replace('http://', 'https://');
```
primiano: i'd much rather throw here.
you have to use HTTPS, so you might as well just be explicit about it.
http will never work even if we wanted because browsers don't allow mixed-content fetch anymore.
Silently upgrading http to https can lead to some hard-to-debug bugs if you have a seerver that responds with two different roots for http and https.
So rather than helping people this will end up hurting them.

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+export type ProtoDescriptors = z.infer<typeof ProtoDescriptorsSchema>;
+
+/**
+ * Runtime state for an extension server.
+ * Combines persisted config with derived and ephemeral fields.
+ * Not persisted - computed fresh on each load.
+ */
+export interface ExtensionServerState {
+  // Source configuration
+  url: string;
```
primiano: specify if this is canonicalized (https://) or not (can it be github:// here?)

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+  selectedModules: string[];
```
primiano: enabledModules? or do you mean something else here?

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+  serverKey: string;
+
+  // Fetched fields (from manifest)
+  displayName: string;
+  availableModules: string[];
+  cspAllowUrls: string[]; // From manifest.csp_allow
+
+  // Ephemeral runtime state (not persisted)
+  lastFetchStatus?: 'success' | 'error';
+  lastFetchError?: string;
```
primiano: isn't error enough? (i.e. the absence of it is a success).
TO me either you  only haave error, or you have to do something like

& ({lastFetchStatus: 'success'} | {lastFetchStatus: 'error'; lastError: string})

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * Aggregated extensions loaded from all enabled servers/modules.
+ * Used internally by the extension system.
+ */
+export interface AggregatedExtensions {
+  macros: Map<string, CommandInvocation[]>; // Key: [server_key module] macro_name
+  sqlModules: Map<string, string>; // Key: module_path
```
primiano: can you add some examples? it's unclear to me why macros has the server_key in the key, but modules and descriptors don't?

ui/src/core_plugins/dev.perfetto.ExtensionServers/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      name: 'Extension servers',
+      description:
+        'Configure external extension servers that provide additional macros, SQL modules, and proto descriptors.',
+      defaultValue: DEFAULT_EXTENSION_SERVERS,
+      schema: ExtensionServersSchema,
+      render: renderExtensionServersSettings,
+    });
+
+    // Initialize extension servers asynchronously
+    initializeExtensions(setting.get()).then((result) => {
```
primiano: can you add the type for result? this code is a bit hard to read and there are so many types around

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,210 @@
...
+// =============================================================================
+// Fetching
+// =============================================================================
+
+async function fetchJson<T>(
+  url: string,
+  schema: z.ZodSchema,
+): Promise<T | undefined> {
+  try {
+    const response = await fetch(url);
```
primiano: maybe use fetchWithTimeout so we don't hang forever? I think chrome's default timeout is like 5 mins, a bit too long

also a bit rusty on this, but I think this should be

```
fetchJson<T>(url:string, schema: T): Promise<z.infer<typeof T>|undefined>
```

to link the type of the schema and the returned type

also why do you need the | undefined part? shouldn't you bubble up fetch errors so that if somebody writes the wrong manifest with invalid stuff (or the server returns some HTTP 500) it becomes obvious?
Either that (i.e. a crash, which is better imho), or wrap in to some exception that we handled in a managed way so we don't propmpt users to file a bug.
but swallowing silently is going to hide server-side bugs.

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,210 @@
...
+  ProtoDescriptorsSchema,
+  SqlModulesSchema,
+} from './types';
+import {normalizeServerKey, resolveServerUrl} from './url_utils';
+
+// =============================================================================
+// Runtime State Management
+// =============================================================================
+
+async function buildServerState(
```
primiano: "build" -> "load" or "initialize" or something similar that makes it more obvious this shoudl be called only once?
otherwise feels like some mithril-style function that can be called frame after frame or something liek that

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,210 @@
...
+  serverUrl: string,
+): Promise<Manifest | undefined> {
+  return fetchJson(`${serverUrl}/manifest.json`, ManifestSchema);
+}
+
+// =============================================================================
+// Loading
+// =============================================================================
+
+async function loadExtension<T>(
```
primiano: loadModuleResource?

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,210 @@
...
+  state: ExtensionServerState,
+  module: string,
+  resourceType: 'macros' | 'sql_modules' | 'proto_descriptors',
+  schema: z.ZodSchema,
+  transformKey?: (
```
primiano: honestly this function is so generic that it's hard to tell what's going on at this point.
if you reach the state where you have a common function where you have to pass a "transformator" maybe just create 3 functions, like loadModuleMacros, loadModuleSqlModules etc?

ui/src/frontend/index.ts
```diff
@@ -173,8 +173,7 @@ function setupContentSecurityPolicy() {
     'connect-src': [
       `'self'`,
       'ws://127.0.0.1:8037', // For the adb websocket server.
-      'https://*.google-analytics.com',
-      'https://*.googleapis.com', // For Google Cloud Storage fetches.
+      'https:', // Allow all HTTPS for extension servers and analytics.
```
primiano: imho this shoudl work differently:

1. From the settings you know which servers you have enabled, so you know the server URLs1, which you can add here.
2. furthermore the server extension (Which you'll have to cache in localstorage) will define some futher extensions to the CSP you'll need to merge here

primiano (commented):
I left some comments. overall LG with few minor comments here and there.
to me the biggest blocker part that needs to be figured out is the interaction between core and this.
my theory is that this cannot be a pure extension (sorry for the pun). the core needs to know "something" about dynamically loadable extensions.
What I don't like is the fragile dance, where this extension loads, and then injects stuff into core, and then some other code needs to carefully await that. it's very very brittle.

I don't have a fully fledged proposal on top off my head, but let's discuss more tomorrow (book some time, steve should be also around).

In my mind the core needs to know, at least in part, of server extensions.
At least part of that types.json imho should be in core.
I don't think core needs to know about how to fetch stuff (Which can be delegated to a plugin like you did).
But in my mind core needs to know at least that "some plugins can provide server extensions, and when you need stuff like sql modules or descriptors, the core needs to query them"

This is the problem I see:
- Some of the stuff that a server offers, in theory is push-only, e.g. macros which could be injected at any time.
-I say in "theory" because in reality we are better off ensuring that they are registered very early on. because somebody might end up triggering a macro early on (can they do progammatically) and being confused about the fact it's not there (unless we say macros for now can only be invoked via user action)
- some other stuff instead is "pull": for instance sql modules. there is a precise point in time, i believe, when core initializes TP, where all the modules MUST have been gathered. same for proto descriptors (maybe?)

this is what needs to be defined.
If there is something that is a "pull at a specific point in time (e.g. while initializing TP)" core must know about that concept, and the dependency needs to be handled with something like "your plugin when initializes sets something into core like registerSqlModuleProvider(...)"

the only question I have is whether over time we'll end up promoting 80% of the extensions bits to the core, at which point we are better off promoting the whole extension thing right now?

this is the biggest open question i have: some part of this MUST be promoted to core: how much ?

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
LalitMaganti: commandInvocationArraySchema is a stable interface as well. It cannot be changed without backcompat either as that's what all settings etc will be depending on.

ui/src/core_plugins/dev.perfetto.ExtensionServers/types.ts
LalitMaganti: This is something we should discuss in more detail. There's some interesting design questions here.

LalitMaganti:
PTAL. There is a very interesting design question left about how we correctly namespace these things remaining but the code itself should now be in a much much better state.

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+import {Tooltip} from '../../widgets/tooltip';
+import {Icon} from '../../widgets/icon';
+import {MultiSelect, MultiSelectDiff} from '../../widgets/multiselect';
+import {ExtensionServer} from './types';
+import {fetchManifest} from './extension_server';
+import {resolveServerUrl} from './url_utils';
+
+type ServerType = 'github' | 'http';
+
+interface AddServerModalState {
```
primiano: slightly neater

type AddServerModalState = 
({  serverType: 'github';
   githubRepo: string; // Format: owner/repo
   githubRef ?: string; // Default to main
} | {
 serverType: 'https':
 httpUrl: string;
}) & {
  loadingManifest: boolean;
  manifestError?: string;
  availableModules: string[];
  enabledModules: Set<string>;
}

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+  httpUrl: string;
+  loadingManifest: boolean;
+  manifestError?: string;
+  availableModules: string[];
+  enabledModules: Set<string>;
+}
+
+class AddExtensionServerModal {
+  state: AddServerModalState;
+  private editingServer: ExtensionServer | null;
```
primiano: please don't use null. we tried very hard to stay consistent and always use undefined (with the only exception of queryresult when we have sql NULL.. and evern there I feel we shoul have used undefined but too late)

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  constructor(server?: ExtensionServer) {
+    this.editingServer = server ?? null;
```
primiano: drop null here

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  private createInitialState(server?: ExtensionServer): AddServerModalState {
+    const baseState = {
+      loadingManifest: false,
+      availableModules: [],
+      enabledModules: new Set(server?.enabledModules ?? []),
+    };
+
+    if (!server) {
```
primiano: server === undefined

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+      return `github://${githubRepo}/${githubRef}`;
+    } else {
+      return this.state.httpUrl;
+    }
+  }
+
+  private isUrlValid(): boolean {
+    if (this.state.serverType === 'github') {
+      const parts = this.state.githubRepo.trim().split('/');
+      return !!(
```
primiano: rather than && just add a !empty to trim?
I thought eslint would bark if you treat an empty string as boolean

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+        parts[0] &&
+        parts[1] &&
+        this.state.githubRef.trim()
+      );
+    } else {
+      return !!this.state.httpUrl.trim();
+    }
+  }
+
+  private async loadManifest(): Promise<void> {
```
primiano: I think this should be wrapped in AsyncLimiter, especially if you follow the suggestion of implicitly fetching every time there is an onchange

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+    if (!this.isUrlValid()) return;
+
+    this.state.loadingManifest = true;
+    this.state.manifestError = undefined;
+    m.redraw();
+
+    try {
```
primiano: why the try block is so wide? shouldn't it only start at fetchManifest?

ui/src/core_plugins/dev.perfetto.ExtensionServers/add_extension_server_modal.ts
```diff
@@ -0,0 +1,299 @@
...
+
+type ServerType = 'github' | 'http';
+
+interface AddServerModalState {
+  serverType: ServerType;
+  // GitHub fields
+  githubRepo: string; // Format: owner/repo
+  githubRef: string;
+  // HTTP field
+  httpUrl: string;
```
primiano: I think you don't need this, and this can be replaced by AsyncLimiter.isRunning. see my other suggestion below about using AsyncLimiter

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,231 @@
...
+      MacrosSchema,
+    );
+    return Object.entries(wrapper.macros)
+      .sort()
+      .map(([name, commands]) => [
+        `[${state.serverKey} ${module}] ${name}`,
+        commands,
+      ]);
+  } catch (e) {
+    console.warn(
```
primiano: do you wnat to swallow these silently?
for now I'd just showModal

ui/src/core_plugins/dev.perfetto.ExtensionServers/index.ts
```diff
@@ -0,0 +1,69 @@
...
+      name: 'Extension servers',
+      description:
+        'Configure external extension servers that provide additional macros, SQL modules, and proto descriptors.',
+      defaultValue: DEFAULT_EXTENSION_SERVERS,
+      schema: ExtensionServersSchema,
+      render: renderExtensionServersSettings,
+    });
+
+    // Initialize extension servers asynchronously
+    initializeExtensions(setting.get()).then(
```
primiano: Hmm no if you do this you defeat the all point of having created those addMethods that take the promise.
You are still adding the promise too late, because of the then.

This should be swapped.
You should create Deferred<> objects here, which you inject immediately.
THe in the initializeExtension.then() you do deferred.resolve()

ui/src/core_plugins/dev.perfetto.ExtensionServers/url_utils.ts
```diff
@@ -0,0 +1,136 @@
...
+  if (trimmed.startsWith('github://')) {
+    const path = trimmed.substring('github://'.length);
+    if (!path) {
+      throw new Error('Invalid GitHub URL: missing owner/repo/ref');
+    }
+    // Path format: owner/repo/ref[/optional/path]
+    return `https://raw.githubusercontent.com/${path}`;
+  }
+
+  // GCS alias: gs://bucket/path
```
primiano: just noting that you have GCS and S3 here, but they are not avialable in the settings UI. did you forget to add them there?

LalitMaganti:
PTAL. I rewrote the whole modal dialog to be a lot easier to follow and it looks nicer too.

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,232 @@
...
+  server: ExtensionServer,
+): Promise<ExtensionServerState> {
+  const canonicalUrl = resolveServerUrl(server.url);
+  const serverKey = normalizeServerKey(canonicalUrl);
+
+  const fetchManifestOrUndefined = async (canonicalUrl: string) => {
+    try {
+      return await fetchManifest(canonicalUrl);
+    } catch (e) {
+      return undefined;
```
primiano: maybe fetchManifest should return a Promise<Result<Manifest>> (Result == StatusOr in c++) converting exceptions into error) rather swallowing exceptions silently here?

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,232 @@
...
+  module: string,
+): Promise<ReadonlyArray<string>> {
+  try {
+    const wrapper = await fetchJson(
+      `${state.canonicalUrl}/modules/${module}/proto_descriptors`,
+      ProtoDescriptorsSchema,
+    );
+    return wrapper.descriptors.sort();
+  } catch (e) {
+    await showModal({
```
primiano: i probably would have bubbled up exceptions here, and try/catch in the for loop below where you call initializeExtension? (and okay if there is an error, you end up skipping all extension for that server... rarther than only skipping sql modules for instance, but seems rare)

ui/src/core_plugins/dev.perfetto.ExtensionServers/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    ctx.addMacros(
+      extPromise.then(async ({macrosPromise}) =>
+        Object.assign({}, ...(await macrosPromise)),
+      ),
+    );
+    ctx.addSqlPackages(
+      extPromise.then(async ({sqlModulesPromise}) => [
+        {
+          // TODO(lalitm): DNS. This needs to be discussed before submitting.
+          name: 'extension_servers',
```
primiano: uh? shouldn't the name here be the server key?
Also overall my comment is: why are you passing all that stuff around back and forth?
IMHO you can make initializeExtensions a non-async function, which just does itself the ctx.addXXX, rather than keep passing objects around. You are creating (imho unnecwessary) layers, within something that is already well self-contained

ui/src/widgets/section.ts
```diff
@@ -15,9 +15,9 @@
 import m from 'mithril';
 import {HTMLAttrs} from './common';
 
-export interface SectionAttrs extends HTMLAttrs {
-  // The name of the section, displayed in the title bar
-  title: string;
+export interface SectionAttrs extends Omit<HTMLAttrs, 'title'> {
+  // The content of the section header (string or custom content)
+  title: string | m.Children;
```
primiano, line 20: What do you need the changes in this file for? couldn't find a place that uses them. i'd keep it simple as it was

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,232 @@
...
+      MacrosSchema,
+    );
+    return Object.entries(wrapper.macros)
+      .sort()
+      .map(([name, commands]) => [
+        `[${state.serverKey} ${module}] ${name}`,
+        commands,
+      ]);
+  } catch (e) {
+    await showModal({
```
primiano: why don't you just bubble up the exception here and let the caller (in the for loop below) catch that and skip the server?
(and okay if you have an exeption only in the proto of submodule, it will skip the whole submodule, rather than skipping the protos... but it's enough isolation imho)

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,232 @@
...
+  const result = schema.safeParse(json);
+  if (!result.success) {
+    throw new Error(`Invalid response from ${url}: ${result.error.message}`);
+  }
+  return result.data;
+}
+
+// Fetches manifest, returning undefined on error (for graceful degradation
+// when configuring servers).
+export async function fetchManifest(serverUrl: string): Promise<Manifest> {
```
primiano: why don't you just bubble up the exception here and let the caller (in the for loop below) catch that and skip the server? (and okay if you have an exeption only in the proto of submodule, it will skip the whole submodule, rather than skipping the protos... but it's enough isolation imho)

ui/src/widgets/section.ts
LalitMaganti: Needed to put refresh in the header for the section https://github.com/google/perfetto/pull/4192/changes#diff-ac73b6e9275676b1ba182f01ee4fb46c55a622de4d64c73e0776a3dfe711c78fR138

ui/src/core_plugins/dev.perfetto.ExtensionServers/extension_server.ts
```diff
@@ -0,0 +1,216 @@
...
+import {resolveServerUrl} from './url_utils';
+import {showModal} from '../../widgets/modal';
+import {errResult, okResult, Result} from '../../base/result';
+import {AppImpl} from '../../core/app_impl';
+
+// =============================================================================
+// Helpers
+// =============================================================================
+
+const FETCH_TIMEOUT_MS = 5000; // 5 seconds
```
primiano: uh this is a bit aggressive, especially if you are tethering or such. 10 sec at least?

ui/src/widgets/section.ts
primiano: The link you pasted doensn't seem to work and I couldn't find a non-trivial case. Anyways

1. Can you try not needing it? (I can't still tell what this is for)

3. If you need something, you cannot just pass m.Children, as you'll end up caching this for more than one render cycle, which leads to UB to mithril (it does actually weird things i fyou have handlers). If you really need it, you need to pass a function that returns m.Children, so that can be invoked on every pass.The problem here is that showModal is called oneOff, but then the modal code renders stuff on every animation frame. If you pass m.children you end up caching stuff.

primiano:
Sorry as discusse offline ignore my last comment on section.ts. I tought you were doing that in Modal.ts which is special as it isn't caled on every render function.
what you were doing (passing just children) was enough, as widgets are called on every render pass

PR #4194, opened 2025-12-27
tp: significantly speed up compile of trace processor

LalitMaganti:
Many many improvements:
1) tables are not directly instantiated as dataframes
2) make all tables forward declared in tracestorage, explicitly include where needed
3) pull out args code into a separate args-utils file
4) simplify table generation


src/trace_processor/storage/trace_storage.cc
```diff
@@ -16,49 +16,91 @@
...
+template <class Variant, class T>
+struct variant_index_impl;
 
-const std::vector<NullTermStringView>& GetRefTypeStringMap() {
-  static const base::NoDestructor<std::vector<NullTermStringView>> map(
-      CreateRefTypeStringMap());
-  return map.ref();
-}
+template <class T, class... Ts>
+struct variant_index_impl<std::variant<Ts...>, T> {
```
sashwinbalaji: can we name it something else like table_init_params_traits.

src/trace_processor/util/args_utils.h
```diff
@@ -90,6 +100,71 @@ class ArgSet {
...
+            {dataframe::FilterSpec{tables::ArgTable::ColumnIndex::arg_set_id, 0,
+                                   dataframe::Eq{}, std::nullopt},
+             dataframe::FilterSpec{tables::ArgTable::ColumnIndex::key, 1,
+                                   dataframe::Eq{}, std::nullopt}})) {}
+
+  // Looks up an arg row by arg_set_id and key.
+  // Returns std::numeric_limits<uint32_t>::max() if not found.
+  uint32_t Get(uint32_t arg_set_id, const char* key) {
+    cursor_.SetFilterValueUnchecked(0, arg_set_id);
+    cursor_.SetFilterValueUnchecked(1, key);
```
sashwinbalaji, line 118: nit: just like above in constructor we can use tables::ArgTable::ColumnIndex::arg_set_id instead of 0 and similar

LalitMaganti: We cannot actually. The indexes here reference the cursor offsets *not* the column indexes.

PR #4195, opened 2025-12-29
ui: improve flamegraph filter bar signifcantly

LalitMaganti:
1) Switch from text box with tooltip to proper "wizard" box allowing
   guided adding of filters
2) Retain "quick add" feature to allow quick typing for people who are
   used to it
3) Add support for adding multiple filters at one time
4) Add a clear all button for clearing all filters in oneshot
5) Add copy button for copying all filters. Also ties with copy spport
6) Add middle elipssis for long filters
7) Add proper multi-line behaviour for many filters


ui/src/widgets/flamegraph.ts
```diff
@@ -174,6 +170,220 @@ export interface FlamegraphAttrs {
...
+        },
+      }),
+      hasPivot &&
+        this.type === 'PIVOT' &&
+        m('.pf-filter-builder__warn', 'Replaces current pivot'),
+    );
+  }
+
+  // Split by comma, respecting quoted strings and backslash escapes
+  // e.g. 'SS: foo, HF: "bar,baz", SS: "qux\"quux"'
```
aMayzner, line 346: I don't understand it, does it respect the quotes or not? In the example you removed the double quotes

ui/src/widgets/flamegraph.ts
```diff
@@ -580,52 +817,71 @@ export class Flamegraph implements m.ClassComponent<FlamegraphAttrs> {
...
-                const filters = Array.from(self.attrs.state.filters);
-                filters.splice(index, 1);
-                self.attrs.onStateChange({
-                  ...self.attrs.state,
-                  filters,
-                });
-              }
+          trigger: m(Button, {
+            label: hasFilters ? undefined : 'Add',
+            icon: 'add',
```
aMayzner: Very much a NIT, but maybe you can use the Icons icon? Here and below. For codebase consistency

ui/src/widgets/flamegraph.ts
LalitMaganti: It removes the quotes but preserves them with \" is present.  It's the same as C/C++

PR #4198, opened 2025-12-29
gn: automatically translate from x64 SIMD intrinsics to WASM

LalitMaganti:
We weren't setting -mavx2 so weren't getting the advantage of
hand-rolled SIMD we had.


sashwinbalaji (approved):

primiano:
Oooh this sounds very nice on paper. TIL they support directly X86 simd intrinsics. 
Did you check if it makes anything faster out of curiosity?

LalitMaganti:
It shouldn't because right now we have a gcc/clang intrinsic fallback for every use of intrinsic and we don't have any hand rolled SIMD. 

That is changing with some of my upcoming patches.

PR #4205, opened 2025-12-31
base: add a new FlatHashMap based on Swiss Table concepts

LalitMaganti:
Note: this CL does *not* change any production code to actually use
this. All it's doing is adding the implementation, benchmarks and tests.
Future CL will actually do the switch.

Our current FlatHashMap implementation has seved us well.
It was originally designed just to replace std::unordered_map inside of
TraceProcessor's StringPool but due to its nice perf characteristics,
it's come to be used throughout the Perfetto codebase.

But over time I've realised that it has some flows which make it not
suitable to be a truly general purpose hash table:
* very sensitive to tombstones. we've made some modifications to
  rehash if there are tombstones but in heavy insert/erase workloads
  it's not enough
* very sensitive to collisions: if there is a collision, the default is
  quadratic probing, which can cause some quite sub-optimal memory
  access patterns. And using linear probing doesn't work for high load
  factor hash maps.
* suboptimal memory layout: putting the ctrl/keys/values in different
  arrays means that we're jumping around memory instead of being in a
  single place, especially important for small maps.

All in all, these are all things that absl::flat_has_map and Swiss
tables have solved. There's no reason for us not to take advantage of at
least their memory layout and orgniaation ideas even if we cannot use
them directly. If they're good enough for much of the C++ world (and
definitely all of Google) they should be good enough for us too!

So this CL adds a new FlatHashMapV2 which is based on Swiss Table
concepths. Specifically;
* The idea of storing the control bytes, keys and values all in the same
  array
* The logic to probe using triangular numbers and do group based
  probing.
* Using SIMD to access 8/16 elements at a time and do comparisions of
  hashes.
* The idea to store 7 bytes of a hash in the control byte to do fast
  checks and reduce the probability of a false match to 1/128 before
  even checking keys
* Using SWAR for non-SIMD platforms (including ARM64 though we might
  optimize this with NEON in the future depending on what we see)

The benchmarks are quite good for general cases.

For TraceStrings:

```
BM_HashMap_InsertTraceStrings_AppendOnly                        554254588 ns    553651249 ns            1 409.798M/s   226.885M   1.17868M
BM_HashMap_InsertTraceStrings<Ours_PreHashed>                   698876584 ns    698092350 ns            1 325.008M/s   226.885M   1.17868M
BM_HashMap_InsertTraceStrings<OursV2_PreHashed>                 662225759 ns    661307563 ns            1 343.086M/s   226.885M   1.17868M
BM_HashMap_InsertTraceStrings<StdUnorderedMap_PreHashed>       2376921965 ns   2374903076 ns            1 95.5346M/s   226.885M   1.17868M
BM_HashMap_InsertTraceStrings<AbslFlatHashMap_PreHashed>        631832647 ns    631347455 ns            1 359.367M/s   226.885M   1.17868M
```

This is the only case where the old implementation wins (which makes
sense because it was specifically designed for it!). But I don't think a
loss here justifies the major wins elsewhere (especially also because in
the "normal" case, cache misses would drown out the small differences
here anyway given how widespread interning is in practice. It also only
wins with AppendOnly which is rarely used.

For TraceTids:

```
BM_HashMap_TraceTids<Ours_Tid>                                  741954852 ns    741389092 ns            1 183.063M/s    16.099k
BM_HashMap_TraceTids<OursV2_Tid>                                504387549 ns    503475605 ns            1 269.568M/s    16.099k
BM_HashMap_TraceTids<StdUnorderedMap_Tid>                      1401707453 ns   1400083835 ns            1 96.9378M/s    16.099k
BM_HashMap_TraceTids<AbslFlatHashMap_Tid>                       510576964 ns    510105976 ns            1 266.064M/s    16.099k
```

We're signifcantly better than V1 here: tids is exactly the sort of
"general behaviour" which it's very common to see in Perfetto and where
V1 behaves less well than I'd like.

For LookupRandInts:

```
BM_HashMap_LookupRandInts<Ours_Default>                         574181027 ns    573859405 ns            1 17.4259M/s   10.7378P
BM_HashMap_LookupRandInts<OursV2_Default>                       350956969 ns    350711465 ns            2 28.5135M/s   10.7378P
BM_HashMap_LookupRandInts<StdUnorderedMap_Default>              571851786 ns    571499561 ns            1 17.4978M/s   10.7378P
BM_HashMap_LookupRandInts<AbslFlatHashMap_Default>              366032708 ns    365819315 ns            2 27.3359M/s   10.7378P
BM_HashMap_LookupRandInts<StdUnorderedMap_Murmur>               611111313 ns    610638342 ns            1 16.3763M/s   10.7378P
BM_HashMap_LookupRandInts<AbslFlatHashMap_Murmur>               363292801 ns    362867428 ns            2 27.5583M/s   10.7378P
BM_HashMap_LookupRandInts<Ours_PreHashed>                       713060262 ns    712048102 ns            1  14.044M/s   10.7378P
BM_HashMap_LookupRandInts<OursV2_PreHashed>                     329546786 ns    329117625 ns            2 30.3843M/s   10.7378P
BM_HashMap_LookupRandInts<StdUnorderedMap_PreHashed>            580294840 ns    579547604 ns            1 17.2548M/s   10.7378P
BM_HashMap_LookupRandInts<AbslFlatHashMap_PreHashed>            340158501 ns    339918114 ns            2 29.4188M/s   10.7378P
```

Another handy win which appears regardless of whether the hash function
is good or not. Notice how in the V1 case, prehashed causes poor
degeneration of the function but V2 is actually *faster*.

For InsertCollidingInts:

```
BM_HashMap_InsertCollidingInts<Ours_Default>                    701989492 ns    701578722 ns            1 14.2536M/s
BM_HashMap_InsertCollidingInts<OursV2_Default>                  427725299 ns    427302016 ns            2 23.4027M/s
BM_HashMap_InsertCollidingInts<StdUnorderedMap_Default>         896047311 ns    895384570 ns            1 11.1684M/s
BM_HashMap_InsertCollidingInts<AbslFlatHashMap_Default>         420245635 ns    420006692 ns            2 23.8091M/s
```

Another big win for V2. Another similar example to the sort we'd see in
practice in Perfetto.

For InsertDupeInts:

```
BM_HashMap_InsertDupeInts<Ours_Default>                          63926278 ns     63874785 ns           11 156.556M/s
BM_HashMap_InsertDupeInts<OursV2_Default>                        40291231 ns     40268380 ns           17 248.334M/s
BM_HashMap_InsertDupeInts<StdUnorderedMap_Default>               37554643 ns     37527602 ns           19 266.471M/s
BM_HashMap_InsertDupeInts<AbslFlatHashMap_Default>               29026714 ns     29008928 ns           24 344.721M/s
BM_HashMap_InsertDupeInts<AbslFlatHashMap_Murmur>                38883999 ns     38857011 ns           18 257.354M/s
```

Same as above, another common case. Surprisingly, std does very well
here. And we seem to be bottlenecked on the hash function given how much
better absl does with its own hash vs murmur (its hash is slightly lower
quality but also faster). This is the only benchmark where this happens
thoough, others don't show this behaviour.

For *String:

```
BM_HashMap_HeterogeneousLookup_String<Ours_String>              575812136 ns    575383946 ns            1 17.3797M/s
BM_HashMap_HeterogeneousLookup_String<OursV2_String>            420531571 ns    420120838 ns            2 23.8027M/s
BM_HashMap_RegularLookup_String<Ours_String>                    820167931 ns    819494830 ns            1 12.2026M/s
BM_HashMap_RegularLookup_String<OursV2_String>                  405975726 ns    405482422 ns            2  24.662M/s
BM_HashMap_RegularLookup_String<StdUnorderedMap_String>        1577833963 ns   1576585547 ns            1 6.34282M/s
BM_HashMap_RegularLookup_String<AbslFlatHashMap_String>         491132758 ns    490802459 ns            2 20.3748M/s
BM_HashMap_RegularLookup_String<AbslFlatHashMap_String_Murmur>  571520036 ns    571134272 ns            1  17.509M/s
```

All clear wins for us. We're doing actually much better than even absl
here!

InsertVaryingSizes:

```
BM_HashMap_InsertVaryingSize<Ours_Default>/100                        754 ns          754 ns       932167 132.677M/s
BM_HashMap_InsertVaryingSize<Ours_Default>/10000                   200216 ns       200096 ns         3531 49.9759M/s
BM_HashMap_InsertVaryingSize<Ours_Default>/1000000               54744741 ns     54674135 ns           13 18.2902M/s
BM_HashMap_InsertVaryingSize<OursV2_Default>/100                     1381 ns         1380 ns       505094 72.4606M/s
BM_HashMap_InsertVaryingSize<OursV2_Default>/10000                 164971 ns       164851 ns         4250 60.6608M/s
BM_HashMap_InsertVaryingSize<OursV2_Default>/1000000             38351821 ns     38310617 ns           18 26.1024M/s
BM_HashMap_InsertVaryingSize<AbslFlatHashMap_Default>/100            1268 ns         1267 ns       551576 78.9371M/s
BM_HashMap_InsertVaryingSize<AbslFlatHashMap_Default>/10000        137933 ns       137845 ns         5065 72.5454M/s
BM_HashMap_InsertVaryingSize<AbslFlatHashMap_Default>/1000000    39980643 ns     39943874 ns           18 25.0351M/s
```

Note how v1 does better for very small maps but drops off as maps get
bigger. I think this is probably the effect of being able to have
everything in cache even with the different key/value/ctrl maps and SIMD
not being as useful with such small maps.

LookupWithMisses:

```
BM_HashMap_LookupWithMisses<Ours_Default>/0                     359571712 ns    359090314 ns            2 27.8481M/s
BM_HashMap_LookupWithMisses<Ours_Default>/50                    426399559 ns    425828348 ns            2 23.4836M/s
BM_HashMap_LookupWithMisses<Ours_Default>/100                   308777908 ns    308628319 ns            2 32.4014M/s
BM_HashMap_LookupWithMisses<OursV2_Default>/0                   195633544 ns    195481786 ns            4 51.1557M/s
BM_HashMap_LookupWithMisses<OursV2_Default>/50                  221380263 ns    221180357 ns            3  45.212M/s
BM_HashMap_LookupWithMisses<OursV2_Default>/100                 139350634 ns    139236152 ns            5 71.8204M/s
BM_HashMap_LookupWithMisses<AbslFlatHashMap_Default>/0          201567396 ns    201412429 ns            4 49.6494M/s
BM_HashMap_LookupWithMisses<AbslFlatHashMap_Default>/50         206378283 ns    206259753 ns            3 48.4826M/s
BM_HashMap_LookupWithMisses<AbslFlatHashMap_Default>/100        123735273 ns    123601106 ns            6 80.9054M/s
```

Notice how much much better we do when we have a low/high miss rate.
This is indicative of the branch predicted doing a much better job of
actually dealing wit our usecases. But OTOH even 50% we are far
suprerior.

InsertSequential:

```
BM_HashMap_InsertSequentialInts<Ours_Default>                   781732694 ns    781402467 ns            1 12.7975M/s
BM_HashMap_InsertSequentialInts<OursV2_Default>                 423644733 ns    423098276 ns            2 23.6352M/s
BM_HashMap_InsertSequentialInts<StdUnorderedMap_Default>        473308044 ns    472810618 ns            2 21.1501M/s
BM_HashMap_InsertSequentialInts<AbslFlatHashMap_Default>        420184095 ns    419919524 ns            2 23.8141M/s
```

Huge win for what is quite a common case in trace processor. Even std is
doing better than us here which is quite surprising!

LookupSequential:

```
BM_HashMap_LookupSequentialInts<Ours_Default>                   586390209 ns    585999772 ns            1 17.0649M/s
BM_HashMap_LookupSequentialInts<OursV2_Default>                 344488692 ns    344312148 ns            2 29.0434M/s
BM_HashMap_LookupSequentialInts<StdUnorderedMap_Default>         49338334 ns     49299886 ns           15  202.84M/s
BM_HashMap_LookupSequentialInts<AbslFlatHashMap_Default>        230591313 ns    230421841 ns            3 43.3987M/s
```

Std destroys everyone which I find highly surprising and don't have a
good explanation for. Maybe it's because it has a perfect hash function
basically (since by default std::hash just returns the int?). Regardless
we're doing better than V1.

```
BM_HashMap_EraseRandInts<Ours_Erasable>                     231067139 ns    230741900 ns            3 43.3385M/s
BM_HashMap_EraseRandInts<OursV2_Erasable>                   234739797 ns    234546351 ns            3 42.6355M/s
BM_HashMap_EraseRandInts<StdUnorderedMap_Default>          2731290342 ns   2728010967 ns            1 3.66567M/s
BM_HashMap_EraseRandInts<AbslFlatHashMap_Default>           346642034 ns    346395066 ns            2 28.8688M/s

BM_HashMap_InsertEraseInterleaved<Ours_Erasable>            700174237 ns    698939463 ns            1 14.3074M/s
BM_HashMap_InsertEraseInterleaved<OursV2_Erasable>          427958820 ns    427531859 ns            2 23.3901M/s
BM_HashMap_InsertEraseInterleaved<StdUnorderedMap_Default> 3887662277 ns   3884060811 ns            1 2.57462M/s
BM_HashMap_InsertEraseInterleaved<AbslFlatHashMap_Default>  376082048 ns    375696540 ns            2 26.6172M/s
BM_HashMap_InsertEraseInterleaved<AbslFlatHashMap_Murmur>   379993339 ns    379412632 ns            2 26.3565M/s

BM_HashMap_EraseTombstoneStress<Ours_Erasable>               23412930 ns     23398168 ns           30         3M
BM_HashMap_EraseTombstoneStress<OursV2_Erasable>             14905630 ns     14893152 ns           47       4.7M
BM_HashMap_EraseTombstoneStress<StdUnorderedMap_Default>     53218188 ns     53175602 ns           13       1.3M
BM_HashMap_EraseTombstoneStress<AbslFlatHashMap_Default>     18289471 ns     18265641 ns           38       3.8M
```

While erase is not the most important for most usecases as primarily we care about Find and Insert, it's important to see
it's not *terrible* either. And infact it's actually very good in general! In most cases being a step above v1 and very competitive (or even better than absl).

### DIFF
--- a/include/perfetto/ext/base/flat_hash_map.h
+++ b/include/perfetto/ext/base/flat_hash_map.h
@@ -116,20 +116,15 @@ static constexpr bool IsLookupKeyAllowed() {
   }
 }
 
-class FlatHashMapV2Base {
- public:
-  // Swiss Table control byte encoding:
-  // - Empty:   0x80 (10000000) - MSB set, easy to detect with sign bit
-  // - Deleted: 0xFE (11111110) - MSB set
-  // - Full:    0x00-0x7F - MSB clear, stores 7-bit H2 hash
-  enum ReservedTags : uint8_t {
-    kFreeSlot = 0x80,  // Empty slot
-    kTombstone = 0xFE  // Deleted slot
-  };
+// Swiss Table control byte encoding:
+// - Empty:   0x80 (10000000) - MSB set, easy to detect with sign bit
+// - Deleted: 0xFE (11111110) - MSB set
+// - Full:    0x00-0x7F - MSB clear, stores 7-bit H2 hash
+static constexpr uint8_t kFreeSlot = 0x80;   // Empty slot
+static constexpr uint8_t kTombstone = 0xFE;  // Deleted slot
 
-  // The default load limit percent before growing the table.
-  static constexpr int kDefaultLoadLimitPct = 75;
-};
+// The default load limit percent before growing the table.
+static constexpr int kDefaultLoadLimitPct = 75;
 
 }  // namespace flat_hash_map_v2_internal
 
@@ -137,8 +132,14 @@ template <typename Key,
           typename Value,
           typename Hasher = base::MurmurHash<Key>,
           typename Eq = flat_hash_map_v2_internal::HashEq<Key>>
-class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
+class FlatHashMapV2 {
  private:
+  // Import constants from internal namespace.
+  static constexpr uint8_t kFreeSlot = flat_hash_map_v2_internal::kFreeSlot;
+  static constexpr uint8_t kTombstone = flat_hash_map_v2_internal::kTombstone;
+  static constexpr int kDefaultLoadLimitPct =
+      flat_hash_map_v2_internal::kDefaultLoadLimitPct;
+
   // Slot structure holds both key and value
   struct Slot {
     Key key;
@@ -148,8 +149,8 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
  public:
   class Iterator {
    public:
-    explicit Iterator(const uint8_t* ctrl, const uint8_t* ctrl_end, Slot* slots)
-        : ctrl_(ctrl), ctrl_end_(ctrl_end), slots_(slots) {
+    explicit Iterator(const uint8_t* ctrl, const uint8_t* ctrl_end, Slot* slot)
+        : ctrl_(ctrl), ctrl_end_(ctrl_end), slot_(slot) {
       FindNextNonFree();
     }
     ~Iterator() = default;
@@ -158,30 +159,31 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
     Iterator(Iterator&&) noexcept = default;
     Iterator& operator=(Iterator&&) noexcept = default;
 
-    Key& key() { return slots_->key; }
-    Value& value() { return slots_->value; }
-    const Key& key() const { return slots_->key; }
-    const Value& value() const { return slots_->value; }
+    const Key& key() { return slot_->key; }
+    Value& value() { return slot_->value; }
+    const Key& key() const { return slot_->key; }
+    const Value& value() const { return slot_->value; }
 
     explicit operator bool() const { return ctrl_ != ctrl_end_; }
     Iterator& operator++() {
       PERFETTO_DCHECK(ctrl_ != ctrl_end_);
       ++ctrl_;
-      ++slots_;
+      ++slot_;
       FindNextNonFree();
       return *this;
     }
 
    private:
     void FindNextNonFree() {
-      for (; ctrl_ != ctrl_end_; ++ctrl_, ++slots_) {
-        if (*ctrl_ != kFreeSlot && *ctrl_ != kTombstone)
+      for (; ctrl_ != ctrl_end_; ++ctrl_, ++slot_) {
+        const uint8_t cur_ctrl = *ctrl_;
+        if (cur_ctrl != kFreeSlot && cur_ctrl != kTombstone)
           return;
       }
     }
     const uint8_t* ctrl_ = nullptr;
     const uint8_t* ctrl_end_ = nullptr;
-    Slot* slots_ = nullptr;
+    Slot* slot_ = nullptr;
   };  // Iterator
 
   explicit FlatHashMapV2(size_t initial_capacity = 0,
@@ -220,7 +222,7 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
   PERFETTO_ALWAYS_INLINE Value* Find(const K& key) const {
     size_t key_hash = Hasher{}(key);
     uint8_t h2 = H2(key_hash);
-    FindResult res = FindInternal<false>(key, key_hash, h2);
+    FindResult res = FindSlotIgnoringTombstones<false>(key, key_hash, h2);
     if (PERFETTO_UNLIKELY(res.needs_insert)) {
       return nullptr;
     }
@@ -231,7 +233,7 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
   bool Erase(const K& key) {
     size_t key_hash = Hasher{}(key);
     uint8_t h2 = H2(key_hash);
-    FindResult res = FindInternal<false>(key, key_hash, h2);
+    FindResult res = FindSlotIgnoringTombstones<false>(key, key_hash, h2);
     if (PERFETTO_UNLIKELY(res.needs_insert)) {
       return false;
     }
@@ -247,31 +249,31 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
   PERFETTO_ALWAYS_INLINE std::pair<Value*, bool> Insert(Key key, Value value) {
     size_t key_hash = Hasher{}(key);
     uint8_t h2 = H2(key_hash);
-    FindResult res = FindInternal<true>(key, key_hash, h2);
-    if (PERFETTO_UNLIKELY(res.needs_insert)) {
-      if (PERFETTO_UNLIKELY(growth_info_.growth_left == 0)) {
-        GrowAndRehash();
-        // After rehash, table has no tombstones. Find an empty slot directly
-        // instead of recursing (which would prevent inlining).
-        res.idx = FindFirstNonFull(key_hash);
-      }
-      PERFETTO_DCHECK(res.idx != kNotFound);
-      size_t insert_idx = res.idx;
-      bool is_freeslot = true;
-      if (PERFETTO_UNLIKELY(growth_info_.has_tombstones)) {
-        insert_idx = FindFirstNonFull(key_hash);
-        is_freeslot = ctrl_[insert_idx] != kTombstone;
-      }
-      new (&slots_[insert_idx].key) Key(std::move(key));
-      new (&slots_[insert_idx].value) Value(std::move(value));
-      SetCtrl(insert_idx, h2);
-      size_++;
-      if (is_freeslot) {
-        growth_info_.growth_left--;
-      }
-      return {&slots_[insert_idx].value, true};
+    FindResult res = FindSlotIgnoringTombstones<true>(key, key_hash, h2);
+    if (PERFETTO_UNLIKELY(!res.needs_insert)) {
+      return {&slots_[res.idx].value, false};
     }
-    return {&slots_[res.idx].value, false};
+    if (PERFETTO_UNLIKELY(growth_info_.growth_left == 0)) {
+      GrowAndRehash();
+      // After rehash, table has no tombstones. Find an empty slot directly
+      // instead of recursing (which would prevent inlining).
+      res.idx = FindFirstEmptyOrTombstone(key_hash);
+    }
+    PERFETTO_DCHECK(res.idx != kNotFound);
+    size_t insert_idx = res.idx;
+    bool is_freeslot = true;
+    if (PERFETTO_UNLIKELY(growth_info_.has_tombstones)) {
+      insert_idx = FindFirstEmptyOrTombstone(key_hash);
+      is_freeslot = ctrl_[insert_idx] != kTombstone;
+    }
+    new (&slots_[insert_idx].key) Key(std::move(key));
+    new (&slots_[insert_idx].value) Value(std::move(value));
+    SetCtrl(insert_idx, h2);
+    size_++;
+    if (is_freeslot) {
+      growth_info_.growth_left--;
+    }
+    return {&slots_[insert_idx].value, true};
   }
 
   Value& operator[](Key key) {
@@ -280,6 +282,7 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
   }
 
   void Clear() {
+    // Avoid trivial heap operations on zero-capacity std::move()-d objects.
     if (PERFETTO_UNLIKELY(capacity_ == 0)) {
       return;
     }
@@ -316,20 +319,28 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
     uint64_t has_tombstones : 1;
   };
 
-  // Not found sentinel for FindInternal (must fit in 63-bit FindResult.idx)
+  // Not found sentinel (must fit in 63-bit FindResult.idx)
   static constexpr size_t kNotFound = std::numeric_limits<size_t>::max() >> 1;
 
-  // Abstraction over group of control bytes.
-  // Provides SIMD-accelerated matching functions or a SWAR fallback.
+  // Abstraction over a group of control bytes that enables batch operations.
+  // On x64, uses SSE to match 16 control bytes in parallel.
+  // On other platforms, uses SWAR (SIMD Within A Register) for 8 bytes.
+  //
+  // Match() returns an iterator over slots whose control byte matches h2.
+  // MatchEmpty() returns an iterator over empty slots (kFreeSlot).
+  // MatchEmptyOrDeleted() returns an iterator over empty or deleted slots.
 #if PERFETTO_BUILDFLAG(PERFETTO_X64_CPU_OPT)
   struct Group {
    public:
     // Group size 16 for x64 SSE
     static constexpr size_t kSize = 16;
 
+    // Iterates over set bits in a match mask. Each set bit indicates a slot
+    // in the group that matched the search criteria. Call Next() to get the
+    // index of each matching slot.
     struct Iterator {
      public:
-      PERFETTO_ALWAYS_INLINE Iterator(uint16_t mask) : mask_(mask) {}
+      PERFETTO_ALWAYS_INLINE explicit Iterator(uint16_t mask) : mask_(mask) {}
       PERFETTO_ALWAYS_INLINE explicit operator bool() const { return mask_; }
 
       PERFETTO_ALWAYS_INLINE size_t Next() {
@@ -348,32 +359,34 @@ class FlatHashMapV2 : public flat_hash_map_v2_internal::FlatHashMapV2Base {
 
     PERFETTO_ALWAYS_INLINE Iterator Match(uint8_t h2) const {
       auto match = _mm_cmpeq_epi8(ctrl_, _mm_set1_epi8(static_cast<char>(h2)));
-      return {static_cast<uint16_t>(_mm_movemask_epi8(match))};
+      return Iterator(static_cast<uint16_t>(_mm_movemask_epi8(match)));
     }
 
     PERFETTO_ALWAYS_INLINE Iterator MatchEmpty() const {
-      return {static_cast<uint16_t>(
-          _mm_movemask_epi8(_mm_sign_epi8(ctrl_, ctrl_)))};
+      return Iterator(static_cast<uint16_t>(
+          _mm_movemask_epi8(_mm_sign_epi8(ctrl_, ctrl_))));
     }
 
     PERFETTO_ALWAYS_INLINE Iterator MatchEmptyOrDeleted() const {
-      return {static_cast<uint16_t>(_mm_movemask_epi8(ctrl_))};
+      return Iterator(static_cast<uint16_t>(_mm_movemask_epi8(ctrl_)));
     }
 
    private:
     __m128i ctrl_;
   };
 #else
-  // Group size 8: uses single 64-bit word (SWAR)
+  // SWAR fallback: processes 8 control bytes at a time using 64-bit arithmetic.
   struct Group {
    public:
     // Group size 8 for ARM and other platforms
     static constexpr size_t kSize = 8;
 
-    // Iterator for sparse 64-bit mask
+    // Iterates over set bits in a sparse 64-bit mask. Each set MSB indicates
+    // a matching byte in the group. Call Next() to get the index of each
+    // matching slot.
     struct Iterator {
      public:
-      PERFETTO_ALWAYS_INLINE Iterator(uint64_t mask) : mask_(mask) {}
+      PERFETTO_ALWAYS_INLINE explicit Iterator(uint64_t mask) : mask_(mask) {}
       PERFETTO_ALWAYS_INLINE explicit operator bool() const { return mask_; }
       PERFETTO_ALWAYS_INLINE size_t Next() {
         // Count zeros and divide by 8 (shift 3)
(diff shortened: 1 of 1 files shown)


include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
-    for (size_t i = 0; i < capacity_; ++i) {
-      const uint8_t tag = tags_[i];
-      if (tag != kFreeSlot && (AppendOnly || tag != kTombstone)) {
-        keys_[i].~Key();
-        values_[i].~Value();
+  PERFETTO_ALWAYS_INLINE std::pair<Value*, bool> Insert(Key key, Value value) {
+    size_t key_hash = Hasher{}(key);
+    uint8_t h2 = H2(key_hash);
+    FindResult res = FindInternal<true>(key, key_hash, h2);
+    if (PERFETTO_UNLIKELY(res.needs_insert)) {
```
primiano: I'd drop the UNLIKELY here. if anything, this is the likely case. but i'd just leave it without any hint.

Also a bit more readable if you first get out of the way the unlikely case, which is the opposite:

if (PERFETTO_UNLIKELY(!res.needs_insert)) {
    return {&slots_[res.idx].value, false};
}

// here you have the actual insertion logic

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
 
   Value& operator[](Key key) {
     auto it_and_inserted = Insert(std::move(key), Value{});
     return *it_and_inserted.first;
   }
 
-  Iterator GetIterator() { return Iterator(this); }
-  const Iterator GetIterator() const { return Iterator(this); }
+  void Clear() {
+    if (PERFETTO_UNLIKELY(capacity_ == 0)) {
```
primiano, line 286: I'd retain the original comment we had in v1 as it's quite useful:

// Avoid trivial heap operations on zero-capacity std::move()-d objects.

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+    return true;
+  } else {
+    return false;
   }
-};
+}
 
-// Non-templated base class to hold helpers for FlatHashMap.
-struct FlatHashMapBase {
+class FlatHashMapV2Base {
```
primiano: what's the benefit of this class? there is no .cc file, so no type erasion needed. it feels like this holds only some constants that could be just put in the internal namespace

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
 
    private:
-    static constexpr size_t kEnd = std::numeric_limits<size_t>::max();
-
     void FindNextNonFree() {
-      const auto& tags = map_->tags_;
-      for (; idx_ < map_->capacity_; idx_++) {
-        if (tags[idx_] != kFreeSlot && (AppendOnly || tags[idx_] != kTombstone))
+      for (; ctrl_ != ctrl_end_; ++ctrl_, ++slots_) {
+        if (*ctrl_ != kFreeSlot && *ctrl_ != kTombstone)
```
primiano: i'm never sure how much the compiler can optimize on pointer aliasing, especially when using uint8_t (Which has special rules).
Imho a bit better if you do:

const uint8_t cur_ctl = *ctrl_;
if (cur_ctl != kFreeSlot && cur_ctl != kTombstone)
  return

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
           return;
       }
-      idx_ = kEnd;
     }
-
-    const FlatHashMap* map_ = nullptr;
-    size_t idx_ = 0;
+    const uint8_t* ctrl_ = nullptr;
+    const uint8_t* ctrl_end_ = nullptr;
+    Slot* slots_ = nullptr;
```
primiano: s/slots_/slot_/ not sure why should be plural, I found  a bit confusing as I was expecting (As it is) just a pointer to the current slot

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
     Iterator(const Iterator&) = default;
     Iterator& operator=(const Iterator&) = default;
     Iterator(Iterator&&) noexcept = default;
     Iterator& operator=(Iterator&&) noexcept = default;
 
-    Key& key() { return map_->keys_[idx_]; }
-    Value& value() { return map_->values_[idx_]; }
-    const Key& key() const { return map_->keys_[idx_]; }
-    const Value& value() const { return map_->values_[idx_]; }
+    Key& key() { return slots_->key; }
```
primiano: i just noticed this issue we have even with the V1 impl: I think the key should be const Key& always.
Imagine if you iterate, take the key and modify it: it will no longer match the original hash, and terrible things will happen, as you have a key in the wrong slot with the wrong hash.

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+    if (PERFETTO_UNLIKELY(res.needs_insert)) {
       return false;
-    EraseInternal(idx);
+    }
+    PERFETTO_DCHECK(size_ > 0);
+    SetCtrl(res.idx, kTombstone);
+    slots_[res.idx].key.~Key();
+    slots_[res.idx].value.~Value();
+    size_--;
+    growth_info_.has_tombstones = 1;
```
primiano, line 245: one overall comment: I think this design is subjected to the same tombstone bitrotting problem we had with the V1, because the design principle is effectively identical. Maybe we should backport the tombstones_ count and "is_many_tombstones" logic here?
Unless I'm missing something, this will reintroduce the same problem we had with "degenerating with deletions"

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+    size_t old_capacity = capacity_;
     size_t old_size = size_;
+    uint8_t* old_ctrl = ctrl_;
+    Slot* old_slots = slots_;
+    std::unique_ptr<uint8_t[]> old_storage(std::move(storage_));
 
     // This must be a CHECK (i.e. not just a DCHECK) to prevent UAF attacks on
     // 32-bit archs that try to double the size of the table until wrapping.
+    size_t new_capacity = old_capacity * 2;
     PERFETTO_CHECK(new_capacity >= old_capacity);
```
primiano, line 547: can you plz

constexpr size_t kGrowFactor = 2;
static_assert((kGrowFactor & (kGrowFactor - 1)) == 0, "must be a pow2 because of the modulo airth in FindInternal");


I can see somebody tempted to make the grow factor a bit less to save memory, and not realizing that the the bitwise arith in FindInternal expects capacity to be strictly a power of 2.

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+
+  // Slots remaining + has_deleted flag
+  GrowthInfo growth_info_{0, 0};
+
+  // Load factor limit in % of |capacity_|.
+  int load_limit_percent_ = kDefaultLoadLimitPct;
+
+  // Cached pointers for fast access (like absl::flat_hash_map)
+  // These are updated whenever storage is allocated/reallocated.
+  uint8_t* ctrl_ = nullptr;  // Points to control bytes
```
primiano, line 638: don't you need some alignas(16) here to be sure it's aligned for SIMD operations?

EDIT: no because you access it at arbitrary offsets so I assume those SIMD op can deal with unaligned mem

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+        if constexpr (ForInsert) {
+          size_t empty_idx = (offset + it.Next()) & cap_mask;
+          return {empty_idx, true};  // Not found - empty slot for insertion
+        } else {
+          return {kNotFound, true};  // Not found - no need to compute slot
+        }
+      }
+
+      // Triangular probing (like Absl): 0, 16, 48, 96, ...
+      probe_index += Group::kSize;
```
primiano: s/probe_index/probe_size/? I got a bit confused in the beginning.  index is imho misleading

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+    static constexpr uint64_t kMsbs = 0x8080808080808080ULL;
+
+    uint64_t ctrl_;
+  };
+#endif
+
+  // The number of cloned control bytes after the main control byte array.
+  static constexpr size_t kNumClones = Group::kSize - 1;
+
+  // Returns FindResult with idx and whether the key needs to be inserted.
```
primiano: This comment should be really expanded because what FindInternal does is really subtle.

First of all I'd consider renaming to something like FindInternal_IgnoreTombstones
then you should be more explicit about the fact that if ForInsert=True, the returned index is the inex of the first empty slot, but that's not necessarily the best index to insert, as there might be some prior tombstone, and caller is suposed to make another pass to find the tombstone

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+    // Use locality hint 3 (high temporal locality) for better L1 cache usage.
+    __builtin_prefetch(ctrl_ + offset, 0, 3);
+
+    while (true) {
+      // Prefetch slots at current probe offset (like Absl).
+      __builtin_prefetch(slots_ + offset, 0, 3);
+
+      Group group(ctrl + offset);
+
+      // Match H2 tags. Use uint16_t to match Absl's BitMask<uint16_t> type.
```
primiano: what is the uint16_t comment referring to? i dont' see any uint16. h2 is a uint8

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+        // group straddles the table boundary (using cloned control bytes).
+        size_t idx = (offset + it.Next()) & cap_mask;
+        if (PERFETTO_LIKELY(Eq{}(slots_[idx].key, key))) {
+          return {idx, false};  // Found
+        }
       }
-    }  // for (idx)
-    return kNotFound;
+
+      // Check for empty slot. If we find one, the key is not present.
```
primiano: make it explicit that this does NOT look for tombstones

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
-    PERFETTO_DCHECK(size_ > 0);
-    tags_[idx] = kTombstone;
-    keys_[idx].~Key();
-    values_[idx].~Value();
-    size_--;
-    tombstones_++;
-    PERFETTO_DCHECK(size_ + tombstones_ <= capacity_);
+  // Find first empty OR deleted slot for insertion.
+  // Called only when has_deleted is set (slow path).
+  size_t FindFirstNonFull(size_t key_hash) const {
```
primiano: imho a bit easier to reason about if you call FindFirstEmptyOrTombstone
non-full is a bit ambiguous as there is precedent (FindInternal) for code that looks only at one of the two

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+  // Using bitfields like absl's GrowthInfo to avoid manual bit manipulation.
+  struct GrowthInfo {
+    uint64_t growth_left : 63;
+    uint64_t has_tombstones : 1;
+  };
+
+  // Not found sentinel for FindInternal (must fit in 63-bit FindResult.idx)
+  static constexpr size_t kNotFound = std::numeric_limits<size_t>::max() >> 1;
+
+  // Abstraction over group of control bytes.
```
primiano: plz add comments on what these functions do. I know after our offline discussion, but in 3 months if anybody has to debug this, they will hate their life.
Add coments to these on what are they expected to do

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+  // Provides SIMD-accelerated matching functions or a SWAR fallback.
+#if PERFETTO_BUILDFLAG(PERFETTO_X64_CPU_OPT)
+  struct Group {
+   public:
+    // Group size 16 for x64 SSE
+    static constexpr size_t kSize = 16;
+
+    struct Iterator {
+     public:
+      PERFETTO_ALWAYS_INLINE Iterator(uint16_t mask) : mask_(mask) {}
```
primiano: +explicit

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+  // Group size 8: uses single 64-bit word (SWAR)
+  struct Group {
+   public:
+    // Group size 8 for ARM and other platforms
+    static constexpr size_t kSize = 8;
+
+    // Iterator for sparse 64-bit mask
+    struct Iterator {
+     public:
+      PERFETTO_ALWAYS_INLINE Iterator(uint64_t mask) : mask_(mask) {}
```
primiano: +explicit

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+    growth_info_.growth_left =
+        (capacity_ * static_cast<size_t>(load_limit_percent_)) / 100;
+    growth_info_.has_tombstones = 0;
 
     if (reallocate) {
-      tags_.reset(new uint8_t[n]);
-      keys_ = AlignedAllocTyped<Key[]>(n);  // Deliberately not 0-initialized.
-      values_ =
-          AlignedAllocTyped<Value[]>(n);  // Deliberately not 0-initialized.
+      size_t slots_offset =
```
primiano, line 578: add some comment here to explain the memory layout. Initially i thought there was a bug becuse storage_ itself is NOT aligned, but there isn't.
so the deal here is that the beignning of storage_ doesn't need to be aligned because the control bytes can be accessed at arbitary indexes.
but then the slots must be aligned. so the only alignment that matters is in the middle where the slots start.
but without this explanation, this 100% looks like an alignment bug masked by the fact that the allocator always gives you 16 bytes alignment even if you dont' ask for it

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+  // buckets. The seed XOR prevents clustering when hash values have patterns
+  // (e.g., sequential keys)
+  static constexpr size_t H1(size_t hash) { return (hash >> 7); }
+  static constexpr uint8_t H2(size_t hash) { return hash & 0x7F; }
+
+  // Set control byte and update clone if needed
+  PERFETTO_ALWAYS_INLINE void SetCtrl(size_t i, uint8_t h) {
+    ctrl_[i] = h;
+    // Update clone if this is one of the first kNumClones entries
+    if (i < kNumClones) {
```
primiano: PERFETTO_UNLIKELY

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
-      kDefaultLoadLimitPct;  // Load factor limit in % of |capacity_|.
-
-  // These arrays have always the |capacity_| elements.
-  // Note: AlignedUniquePtr just allocates memory, doesn't invoke any ctor/dtor.
-  std::unique_ptr<uint8_t[]> tags_;
-  AlignedUniquePtr<Key[]> keys_;
-  AlignedUniquePtr<Value[]> values_;
+
+  // Slots remaining + has_deleted flag
+  GrowthInfo growth_info_{0, 0};
```
primiano, line 631: out of curiosity is putting the "has tombstones" in a bit in the same word something that ABSL does?
feels a bit odd, espexially considering that it requires a bitmasking on every access.
(but if they did i guess they had good reasons for it)

include/perfetto/ext/base/flat_hash_map.h
```diff
@@ -17,452 +17,586 @@
...
+  // Set control byte and update clone if needed
+  PERFETTO_ALWAYS_INLINE void SetCtrl(size_t i, uint8_t h) {
+    ctrl_[i] = h;
+    // Update clone if this is one of the first kNumClones entries
+    if (i < kNumClones) {
+      ctrl_[capacity_ + i] = h;
+    }
   }
 
+  // Owns the actual memory: [ctrl[capacity]][clones[15]][slots[capacity]]
```
primiano: Imho better if you say:

// Memory layout:
//
// [Control bytes]
// |capacity| bytes for control bytes.
// 15 bytes for control bytes clones (*)  (add a comment in appendix)
// No alignment required (Accessed at random byte offsets)
//
// [padding for Slots alignment]
//
// [Slots]
// Capacity * sizeof(Slot): contains key-value pairs
// Must be aligned at alignof(Slot)

include/perfetto/ext/base/flat_hash_map.h
LalitMaganti: Yes it is.

include/perfetto/ext/base/flat_hash_map.h
LalitMaganti: So I benchmarked this and it works faster than v1 even without any of that complexity.

absl doesn't have this either so I'm inclined to stick with what we have today.

PR #4208, opened 2026-01-02
Add additional depth validation to slice_is_ancestor function

### LINKED ISSUE #4207: slice_is_ancestor(a, b) == slice_is_ancestor(b, a) == True
While working with `slice_is_ancestor()` I've noticed some incredibly odd behavior:

`slice_is_ancestor(31744, 31749) == True`
Image

`slice_is_ancestor(31749, 31744) == True`
Image

Here are the two rows:
| id | ts | dur | track_id | category | name | depth | parent_id | arg_set_id | thread_ts | thread_dur | thread_instruction_count | thread_instruction_delta | cat | slice_id |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 31744 | 13546845003 | 1458449779 | 18 | sampleEvent | name1 | 55 | 31743 | null | null | null | null | null | sampleEvent | 31744 |
| 31749 | 13546845003 | 1458449779 | 18 | sampleEvent | name2 | 60 | 31748 | null | null | null | null | null | sampleEvent | 31749 |

I believe the cause of this bug is the fact that these two slices have the exact same `ts` and `dur`, and `slice_is_ancestor()` doesn't handle this edge case very well because it's missing an additional `depth` validation ([link](https://github.com/google/perfetto/blob/a8194d59e0766406c7e233bb84c2c1c87f8dfdd1/src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/slices.sql))

```sql
-- Given two slice ids, returns whether the first is an ancestor of the second.
CREATE PERFETTO FUNCTION slice_is_ancestor(
    -- Id of the potential ancestor slice.
    ancestor_id LONG,
    -- Id of the potential descendant slice.
    descendant_id LONG
)
-- Whether `ancestor_id` slice is an ancestor of `descendant_id`.
RETURNS BOOL AS
SELECT
  ancestor.track_id = descendant.track_id
  AND ancestor.ts <= descendant.ts
  AND (
    ancestor.dur = -1 OR ancestor.ts + ancestor.dur >= descendant.ts + descendant.dur
  )
FROM slice AS ancestor, slice AS descendant
WHERE
  ancestor.id = $ancestor_id AND descendant.id = $descendant_id;
```

dfamonteiro:
Add additional `depth` validation to `slice_is_ancestor` function.

Fixes #4207

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/slices.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/slices.sql
@@ -26,6 +26,7 @@ CREATE PERFETTO FUNCTION slice_is_ancestor(
 RETURNS BOOL AS
 SELECT
   ancestor.track_id = descendant.track_id
+  AND ancestor.depth < descendant.depth
   AND ancestor.ts <= descendant.ts
   AND (
     ancestor.dur = -1 OR ancestor.ts + ancestor.dur >= descendant.ts + descendant.dur


dfamonteiro:
There is an open topic to be discussed: what should the result of `slice_is_ancestor(a, a)` be? 

At the moment it returns `true` and this PR breaks this undocumented behavior. What's your call: do you want to preserve backwards compatibility, or do you consider this to be a bug?

LalitMaganti:
Nice spot thanks for the fix!

> At the moment it returns true and this PR breaks this undocumented behavior. What's your call: do you want to preserve backwards compatibility, or do you consider this to be a bug?

It should return false - ancestor throughout trace processor has never included the slice itself. Same with descendant.

PR #4213, opened 2026-01-05
Revert "Remove `stackCpuSampling` events. (#2745)"

keyou:
This reverts commit 6dffc26f2442ba6b369ba8bacd6e37d400cd5b97.

Reason for revert: broken chrome cpu_profiler

This is the first commit to fix crbug.com/473398637, and will be followed by another commit.

### DIFF
--- a/src/trace_processor/export_json.cc
+++ b/src/trace_processor/export_json.cc
@@ -157,6 +157,7 @@ class JsonExporter {
     RETURN_IF_ERROR(ExportSlices());
     RETURN_IF_ERROR(ExportFlows());
     RETURN_IF_ERROR(ExportRawEvents());
+    RETURN_IF_ERROR(ExportCpuProfileSamples());
     RETURN_IF_ERROR(ExportMetadata());
     RETURN_IF_ERROR(ExportStats());
     RETURN_IF_ERROR(ExportMemorySnapshots());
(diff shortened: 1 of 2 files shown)


LalitMaganti:
As the original author, @jackfranklin needs to review this.

jackfranklin:
The reason that I removed this code is because it causes the JSON traces that DevTools processes to inflate in size massively (I can find the actual numbers, but on an example React app these events take the file from roughly 300mb to 1.1gb).

If there is some consumer that needs them, then we need to figure out a way for a client to disable them when it doesn't use them, so that on the DevTools side we can do that and not have to pay the huge JSON file cost for a vast amount of events that we don't need.

I will defer to Perfetto folks on the exact path forward here but unfortunately bringing these events back exactly as they were would have a huge negative impact on folks using DevTools - particularly with frameworks like React that tend to generate a huge number of events.

keyou:
@jackfranklin  I tested locally and found that deleting the `LegacyV8CpuProfileTracker::Parse` method can also resolve the issue you mentioned, while ensuring the correctness of the cpu_profiler functionality in chrome://tracing. 

I'm not sure if the code related to `LegacyV8CpuProfileTracker::Parse` meets the criteria for deprecation and can be safely removed. It may require someone more familiar with this part of the code to evaluate.

LalitMaganti:
LegacyV8CpuProfileTracker::Parse certainly cannot be removed. It parsed V8 cpu profiling information which is widely used by everyone including Devtools themselves.

PR #4219, opened 2026-01-05
ui: Make full track header clickable

camillobruni:
Support clicking on the track-canvas header as well.
This makes it more natural to open a track when hovering on the the right side with the mouse cursor.


stevegolton:
Hey, thanks for the PR.

Unfortunately this breaks track event selection on collapsible tracks (e.g. those with children). By that I mean that if you try to select an event on one of those tracks, it'll just toggle the collapsed state instead.

I think we should just keep the collapse/uncollapse toggle hitbox in the shell only - I don't want to overload track canvas area.

Happy to discuss though.

camillobruni:
oh ok that's no good :)
Ideally I'd want this to work only in the track header, not the body itself (not sure there's a nice way to do this).
Alternative I could enable this only for opening – which would solve the main use-case here. WDYT?

stevegolton:
> oh ok that's no good :) Ideally I'd want this to work only in the track header, not the body itself (not sure there's a nice way to do this).

No no I think you're correctly adding the event handler to the header (rather than the body... unless I'm missing something). Either way, the problem is if that track header has its own events in it then clicking them will conflict with collapse/expand.

> Alternative I could enable this only for opening – which would solve the main use-case here. WDYT?

I don't think this will really help given that the headers can also contain events. It might also be confusing as it's introducing an inconsistency.

camillobruni:
as discussed offline – this causes too many inconsistency with tracks containing slices in the headers.

PR #4223, opened 2026-01-05
tp: Convert Lemon-generated C parsers to C++

LalitMaganti:
This commit simplifies the Lemon parser integration by generating C++ code
directly instead of having complex extern "C" bridging between C and C++.


LalitMaganti:
We're going to go with a different, more comprehensive approach so closing this out.

PR #4226, opened 2026-01-05
ui: add org.chromium.V8 plugin with Runtime Call Stats support

camillobruni:
Add native V8 runtime call stats support to perfetto.
This well remove the last requirement on about://tracing for V8 engineers.

- Conditionally enable the plugin if there are any RCS args available
- Display a pivot table with the extracted RCS data from the current area selection
- Calculate proportional  sums of RCS data based on the overlap ratio with the current selection


camillobruni:
Screenshot 2026-01-07 at 13 30 05

ui/src/plugins/org.chromium.V8/v8_runtime_call_stats_tab.ts
```diff
@@ -0,0 +1,308 @@
...
+    if (shouldLoad && this.previousSelection === selection) {
+      await this.updateSqlView(selection, trackIds);
+      if (this.previousSelection === selection) {
+        this.state = this.createState();
+      }
+    }
+
+    if (this.previousSelection === selection) {
+      this.loading = false;
+      this.trace.raf.scheduleFullRedraw();
```
stevegolton: I don't think you should need this redraw - redraws are scheduled automatically after queries complete

ui/src/plugins/org.chromium.V8/v8_runtime_call_stats_tab.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    if (this.loading) {
+      return m('div.pf-v8-loading-container', m(Spinner));
+    }
+
+    if (!this.state) {
+      return this.renderEmptyState();
+    }
+
+    return m(PivotTable, {
```
stevegolton: This pivot table implementation is very soon about to be deprecated and removed (I'll mark it as such in another PR as this isn't obvious). Please use DataGrid instead. If it's missing any functionality you need, I'll either add it shortly or happy to accept patches.

stevegolton (approved):
Looks good to me!

camillobruni:
Had to update the DataGrid aggregates since we not need an explicit `id` it seems.

PR #4232, opened 2026-01-06
ui: Display cumulative sizes in heap graph 'objects' view

zezeozue:
Adds cumulative Java/native size and instance counts to the disaggregated heap profile details view.


src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/class_tree.sql
```diff
@@ -46,6 +48,41 @@ FROM graph_reachable_bfs!(
 ORDER BY
   id;
 
+CREATE PERFETTO TABLE _heap_graph_object_tree_aggregation AS
```
LalitMaganti: This is a very expensive tranversal to add here and is also used on the critical path of flamegraph, slowing that down for everyone. If you need this, please separate it into a different file.

LalitMaganti (approved):

PR #4233, opened 2026-01-06
ui: unify CPU/Perf/Instruments profiling track implementations

LalitMaganti:
Consolidate three separate profiling track implementations into a single
unified createProfilingTrack() function in the CpuProfile plugin. The new
implementation uses SourceDataset as an abstraction layer and accepts a
ProfilingTrackConfig to handle differences between track types.


LalitMaganti:
@stevegolton this is purely an FYI, Ryan said he would review. ccing you because there's a todo of yours :)

rsavitski (approved):

PR #4234, opened 2026-01-06
Fix cpu_profiler broken in chrome://tracing

keyou:
The commit 5d85654 moves the registration of the `ProfileModule` object from `RegisterDefaultModules` to `RegisterAdditionalModules`, which results in the `ProtoTraceReader::ProtoTraceReader` being unable to register `ProfileModule`, on which chrome cpu_profiler depends.

This commit only reverts part of the code from 5d85654.

See crbug.com/473398637


LalitMaganti:
I would like @betasheet to comment here about whether or not supporting CPU events on chrome://tracing is something we want to continue doing given that very few (if any) Chromium developers in Google are actually using chrome://tracing at this point.

PR #4237, opened 2026-01-06
ui: use more consistent font-sizes

camillobruni:
Currently there are many places which use inconsistent font-sizes.
Changing the default font-size to 14px already solves a few issues (menus using 16px vs toolbar buttons using 14px)

- Add `--pf-font-size-XXX` variables
- Use 14px default font-size
- Reduce styles with relative font sizes
- Use h1 as modal title


camillobruni:
TBD: This does shrink the font-size in some places like Flags / Settings and Keyboard shortcuts but IMO using 14px is probably ok?

Especially the node-explorer and record-page were rather liberal in using in-between sizes, I've tried to match intentions there by adjusting it a bit.

stevegolton (approved):
Looks great, thanks!

aMayzner (approved):

PR #4238, opened 2026-01-06
ui: add chrome presets for recording

camillobruni:
- Add chrome default and  v8 preset
- Use chrome and v8 preset on android and chrome-os as well


LalitMaganti:
Looking at https://storage.googleapis.com/perfetto-ci-artifacts/gh-20782178548-1-ui/ui/index.html#!/record

These presets are showing up even in the Android menu which feels wrong. Can you please fix?

camillobruni:
Yes this is in purpose – we do investigate v8 and chrome on android.

PR #4249, opened 2026-01-06
tp: ep:  Add experimental FilterToIntervals to structured query

aMayzner:
This change adds a new `ExperimentalFilterToIntervals` operation to the structured query API, designed to filter intervals from a base query to only those that overlap with intervals from a filter set.

Unlike the more general-purpose `IntervalIntersect`, this operation is specifically optimized for the common use case of filtering intervals while preserving the original schema of the base table. It automatically handles overlapping interval merging to avoid duplicate results and supports optional clipping of timestamps to interval boundaries.

  Key features:
  - Preserves base table schema (no column suffixes)
  - Automatic merging of overlapping filter intervals to prevent duplicates
  - Optional timestamp clipping with `clip_to_intervals` flag
  - Flexible column selection via `select_columns`
  - Support for partitioned filtering via `partition_columns`
  - When clipping is enabled and ts/dur are explicitly selected, original values are preserved as `original_ts` and `original_dur`


protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -201,6 +201,95 @@ message PerfettoSqlStructuredQuery {
...
+  //
+  // Schema:
+  //  1) The query in `base` must have `id`, `ts`, and `dur` columns.
+  //  2) The query in `intervals` must have `ts` and `dur` columns.
+  //
+  //
+  // Output Schema:
+  // The output preserves the exact same columns as `base`, but:
+  //  - Rows are filtered to only those that overlap with `intervals`
+  //  - If `clip_to_intervals` is true (default), `ts` and `dur` represent
```
LalitMaganti, line 226: I like this phrasing a lot. We should use it in other places too.

protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -201,6 +201,95 @@ message PerfettoSqlStructuredQuery {
...
+  //
+  // Example in ASCII art (with clip_to_intervals = true):
+  //   base:       [-----------]     [--------]
+  //   intervals:           [-----------]
+  //   output:              [--]     [--]
+  //
+  // Example in ASCII art (with clip_to_intervals = false):
+  //   base:       [-----------]     [--------]
+  //   intervals:           [-----------]
+  //   output:     [-----------]     [--------]
```
LalitMaganti, line 258: The combination of these two is a bit confusing honestly. I'd either expect to work consistently one way or another.

aMayzner: I really like the fact that we have two options and it makes sense in the UI in filter during node. This is very thin layer on intervals intersect

LalitMaganti: I'm not talking aboutb clip_to_intervals which is intuitive. I'm talking about the cases where we merge/don't merge. The lack of consistency is confusing.

aMayzner: But it is consistent. We always try to merge if something is mergable.

LalitMaganti: I'm nto sure I agree but not worth discussing more.

PR #4272, opened 2026-01-07
sdk: Allow only prvalue DynamicStrings

sashwinbalaji:
Instead of trying to update the uses of dynamicstring to avoid passing temporary string (which is a valid scenario), update the trace_event/trace_counter to only accept prvalue dynamic strings.

### DIFF
--- a/include/perfetto/tracing/internal/track_event_data_source.h
+++ b/include/perfetto/tracing/internal/track_event_data_source.h
@@ -126,28 +126,31 @@ static constexpr bool IsValidTimestamp() {
 template <typename...>
 using void_t = void;
 
+// Helper to check if T is a pure rvalue (prvalue).
+template <typename T>
+constexpr bool IsPrvalue = !std::is_reference_v<T>;
+
 // Returns true iff `T` is a valid event name type.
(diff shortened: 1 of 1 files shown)


include/perfetto/tracing/internal/track_event_data_source.h
```diff
@@ -126,10 +126,18 @@ static constexpr bool IsValidTimestamp() {
...
+// - Temporary `perfetto::DynamicString` objects. `DynamicString` uses a raw
+//   pointer, so to prevent dangling pointers, the underlying string must
+//   outlive the event. Requiring `DynamicString` to be a temporary ensures this
+//   for the duration of the TRACE_EVENT call.
 template <typename T, typename = void>
 struct IsValidEventNameType
-    : std::is_same<perfetto::DynamicString, typename std::decay<T>::type> {};
+    : std::bool_constant<std::is_same_v<perfetto::DynamicString,
+                                        typename std::decay<T>::type> &&
+                         !std::is_reference_v<T>> {};
```
LalitMaganti: This needs a comment. And also an alias.

PR #4274, opened 2026-01-07
ui: relax CSP and move to a serviceworker based firewall

LalitMaganti:
This CL changes the CSP such that it allows all https: fetches.
This has a very nice property of allowing url= to now work for any
publicly accessible https trace which can be fetched via a GET call.

Also update the documentation to match.

### DIFF
--- a/contrib/rust-sdk/Cargo.lock
+++ b/contrib/rust-sdk/Cargo.lock
@@ -180,7 +180,7 @@ dependencies = [
 
 [[package]]
 name = "perfetto-sdk-sys"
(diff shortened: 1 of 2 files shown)


LalitMaganti:
It's so incredibly cool that now https://storage.googleapis.com/perfetto-ci-artifacts/gh-20794707740-1-ui/ui/index.html?url=https%3A%2F%2Fohadravid.github.io%2F2025-05-rav1d-faster%2Frav1d_profile_baseline.json.gz just works!

PR #4278, opened 2026-01-08
ui: simplify how macros/sql package/descriptor extensions work

LalitMaganti:
This refactors the legacy "is_internal_user" script to be an
implementation of the first type of "server extension" (i.e. data which
can be loaded from a server separate to the UI). This allows for a
smoother implementation of RFC-0005 when we do proper server extensions.

### DIFF
--- a/ui/src/core/app_impl.ts
+++ b/ui/src/core/app_impl.ts
@@ -68,8 +68,8 @@ export interface AppInitArgs {
  * and should use AppImpl instead.
  */
 export class AppImpl implements App {
+  readonly commands = new CommandManagerImpl();
   readonly omnibox = new OmniboxManagerImpl();
-  readonly commands = new CommandManagerImpl(this.omnibox);
   readonly pages = new PageManagerImpl();
   readonly sidebar: SidebarManagerImpl;
   readonly plugins = new PluginManagerImpl();


LalitMaganti:
@primiano purely as an FYI, Steve should be able to review.

ui/src/core/app_impl.ts
```diff
@@ -282,19 +282,47 @@ export class AppImpl implements App {
...
     return result;
   }
 
   navigate(newHash: string): void {
     Router.navigate(newHash);
   }
 
-  notifyOnExtrasLoadingCompleted() {
-    this.extrasLoadingDeferred.resolve();
+  addExtensionSqlPackagesPromise(
```
primiano: i'd drop "extension" from the names as, as the core is concerned, every method in here is an "extension"

ui/src/core/app_impl.ts
```diff
@@ -282,19 +282,47 @@ export class AppImpl implements App {
...
+    this._extensionSqlPackagesPromises.push(sqlPackagesResult);
+  }
+
+  async extensionSqlPackages(): Promise<ReadonlyArray<SqlPackage>> {
+    return Promise.all(this._extensionSqlPackagesPromises).then((pkgs) =>
+      pkgs.flatMap((p) => p),
+    );
+  }
+
+  addExtensionParsingDescriptorsPromise(
```
primiano: a bit more easier to understand if you call it addProtoDescriptors so it's clear it's about protos.
also make it clear what's the expected format of the string (Base64?)

ui/src/core_plugins/dev.perfetto.CoreCommands/index.ts
```diff
@@ -241,13 +241,9 @@ export default class CoreCommands implements PerfettoPlugin {
...
     ctx.onTraceReady.addListener(async (_) => {
-      // Await the promise: we've tried to be async as long as possible but
+      // Await the promises: we've tried to be async as long as possible but
       // now we need the extras to be loaded.
-      await app.extraLoadingPromise;
-      registerMacros(
-        ctx,
-        app.extraMacros.reduce((acc, macro) => ({...acc, ...macro}), {}),
-      );
+      registerMacros(ctx, await app.extensionMacros());
```
primiano: out of curiosity why do we:

1. register the settings macros immediately
2. defer the registration of the extension macros to onTraceReady?

why this is not:

app.extensionMacros.then((e) => registerMacros(ctx, e)) ?

ui/src/core/app_impl.ts
```diff
@@ -93,18 +93,20 @@ export class AppImpl implements App {
 
   // This is normally empty and is injected with extra google-internal packages
   // via is_internal_user.js
-  extraSqlPackages: SqlPackage[] = [];
+  private _extensionSqlPackagesPromises: Promise<ReadonlyArray<SqlPackage>>[] =
```
primiano: drop _extension from the names (here and below) as everythign here is technically an "extension (of core)".

Also imho 

private _extensionSqlPackagesPromises = new Array<Promise<ReadOnlyArray<SqlPackage>>()

makes it easier to read.
I originally missed the trailing [], and when I saw that i went "wait why we do one array with [] and the other with ReadOnlyArray?"


same below

ui/src/core/app_impl.ts
```diff
@@ -93,18 +93,20 @@ export class AppImpl implements App {
...
+  >[] = [];
 
   // This is normally empty and is injected with extra google-internal macros
   // via is_internal_user.js
-  extraMacros: Record<string, CommandInvocation[]>[] = [];
-
-  // Promise which is resolved when extra loading is completed.
-  extrasLoadingDeferred = defer<undefined>();
+  private _extensionMacrosPromises: Promise<
+    Record<string, ReadonlyArray<CommandInvocation>>
```
primiano: out of cusiority why Record and not Map<string, Array> ?
also clarify what the key is in a comment, as it's not obvious

ui/src/core/app_impl.ts
```diff
@@ -282,19 +282,47 @@ export class AppImpl implements App {
...
   }
 
   navigate(newHash: string): void {
     Router.navigate(newHash);
   }
 
-  notifyOnExtrasLoadingCompleted() {
-    this.extrasLoadingDeferred.resolve();
+  addExtensionSqlPackagesPromise(
+    sqlPackagesResult: Promise<ReadonlyArray<SqlPackage>>,
```
primiano: you can drop the "Promise" and take both a promise or directly the array if you do something liek

addSqlPackages(args: ReadonlyArray<SqlPackage> | Promise<ReadonlyArray<SqlPackage>>) {
  this.sqlPackages.push(Promise.resolve(args));
}

same below

ui/src/frontend/is_internal_user_script_loader.ts
```diff
@@ -25,7 +24,7 @@ const SCRIPT_URL =
 
 // This interface describes the required interface that the script expect to
 // find on window.globals.
-interface Globals {
+interface InteralUserScriptParams {
```
primiano: Neither the old nor the new. the proper way to do this is:

declare global {
  interface Window {
    isInternalUser ....

see other examples

ui/src/core_plugins/dev.perfetto.CoreCommands/index.ts
LalitMaganti: Well the settings macros are not async. These ones are. We wait until this point because startup commands could depend on these macros so we need to register them now.

LalitMaganti:
PTAL: addressed all comments.

ui/src/core/app_impl.ts
```diff
@@ -282,19 +282,47 @@ export class AppImpl implements App {
...
+
+  async protoDescriptors(): Promise<ReadonlyArray<string>> {
+    return Promise.all(this._protoDescriptorsPromises).then((desc) =>
+      desc.flatMap((d) => d),
+    );
+  }
+
+  addMacros(
+    args:
+      | Map<string, ReadonlyArray<CommandInvocation>>
```
primiano, line 318: stray leading | (which i believe is causing the weird formatting)

ui/src/frontend/is_internal_user_script_loader.ts
```diff
@@ -113,4 +100,31 @@ export async function tryLoadIsInternalUserScript(app: AppImpl): Promise<void> {
...
+  // by the script if possible.
+  return globals;
+}
+
+/**
+ * Loads a script that detects if the user is internal, allowing access to
+ * non-public features and SQL packages. Registers the loaded data directly
+ * on the app instance.
+ */
+export async function tryLoadIsInternalUserScript(app: AppImpl): Promise<void> {
```
primiano: technically this doesn't need to be async, as there is no await. it returns a promise but it's a one-shot function

ui/src/frontend/is_internal_user_script_loader.ts
```diff
@@ -113,4 +100,31 @@ export async function tryLoadIsInternalUserScript(app: AppImpl): Promise<void> {
...
+  const promise = tryLoadIsInternalUserScriptInternal();
```
primiano: s/promise/scriptLoaded/

ui/src/core/app_impl.ts
LalitMaganti: It's the other way round: Prettier is *introducing* these, I didn't add these manually.

ui/src/core_plugins/dev.perfetto.CoreCommands/index.ts
primiano: yeah but why do you wait further for onTraceReady rather than registering them just as soon as they are available?

PR #4286, opened 2026-01-08
ui: clear SQLDataSource when pivot changes to avoid stale data

camillobruni:
This fixes an issue where the DataGrid temporarily displays stale data when switching to/from pivot views.

### DIFF
--- a/ui/src/components/widgets/datagrid/sql_data_source.ts
+++ b/ui/src/components/widgets/datagrid/sql_data_source.ts
@@ -237,7 +237,9 @@ export class SQLDataSource implements DataSource {
   private shouldClearCache(newModel: DataSourceModel): boolean {
     if (!this.lastModel) return false;
     // Covering the most basic use-case.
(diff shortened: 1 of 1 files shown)


camillobruni:
This is very much a poor-man's approach to the underlying problem: whenever the number of columns / filtering change we should probably clear the table.

camillobruni:
Example failure with https://github.com/google/perfetto/pull/4226:
https://github.com/user-attachments/assets/d6766ea1-1a71-44a0-a4ff-b84c3d3249b5

camillobruni:
Forgot to assign this over to you too – I have a hunch this is just a hack and there is a better approach to solve this.

stevegolton:
I believe this https://github.com/google/perfetto/pull/4384 should fix the problem - it's a complete rewrite of the sql data source. I'm made the mistake of putting two changes in one which is making it more difficult to land unfortunately, so I might split it out

PR #4297, opened 2026-01-08
Add fwtp_perfetto_slice ftrace event

estaats-google:
This ftrace event reports slice events in Pixel firmware.

This ftrace event is only on Pixel kernels and is not upstream.

Bug: 469451931
Test: Verified FWTP slice tracepoints may be collected with Perfetto
 and displayed with traceconv.

Welcome to Perfetto!

For more details please see


primiano:
Out of curiosity did you look at https://perfetto.dev/docs/getting-started/ftrace#part-c-simple-slice-counter-visualisations-without-modifying-perfetto-code-kernel-track-events- ?

these days it should be possible to trace slices and counters from the kernel without modifying perfetto at all

@rsavitski (who created that guide and the supprot for those generic kernel events)

LalitMaganti:
> these days it should be possible to trace slices and counters from the kernel without modifying perfetto at all

To the best of my knowledge this is going to be because the timestamps come from the firmware *not* the ftrace timestamps.

primiano:
> https://perfetto.dev/docs/getting-started/ftrace#part-c-simple-slice-counter-visualisations-without-modifying-perfetto-code-kernel-track-events- ?

aaah right. nvm then.
i'll file a bug to improve that: b/474354217

estaats-google:
> > these days it should be possible to trace slices and counters from the kernel without modifying perfetto at all
> 
> To the best of my knowledge this is going to be because the timestamps come from the firmware _not_ the ftrace timestamps.

That is correct. For reference, see https://github.com/google/perfetto/pull/3653#issuecomment-3528874521 .

primiano:
btw I don't know how much this is a sailed ship or there are margins for tweaking. my only comment is that rather than muxing everything into one event called "fwtp_perfetto_slice" which has a category field, I would have created one event per category, like "fwtp_foo", "fwtp_bar" (Assuming it's tractable)

the nice thing about the latter is that you can later selectively only enable some categories.
The current one is all or nothing

having said this, if it's too late (deadline for changes are past etc) ignore this comment

estaats-google:
> btw I don't know how much this is a sailed ship or there are margins for tweaking. my only comment is that rather than muxing everything into one event called "fwtp_perfetto_slice" which has a category field, I would have created one event per category, like "fwtp_foo", "fwtp_bar" (Assuming it's tractable)
> 
> the nice thing about the latter is that you can later selectively only enable some categories. The current one is all or nothing
> 
> having said this, if it's too late (deadline for changes are past etc) ignore this comment

The category field was already added in https://github.com/google/perfetto/pull/3653 for counter traces.

PR #4299, opened 2026-01-08
android: For trigger drops also log the uuid_lsb

sashwinbalaji:
Similar to how PerfettoStatsdAtom logs the uuid_lsb extend StatsdLogTriggerEvent to also log it for better debuggability.

The uuid_lsb is logged when available for that event type else 0.

Also remove the unused function: LogTriggerEvents

### DIFF
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -4233,7 +4233,7 @@ void TracingServiceImpl::MaybeLogUploadEvent(const TraceConfig& cfg,
   if (!ShouldLogEvent(cfg))
     return;
 
-  PERFETTO_DCHECK(uuid);  // The UUID must be set at this point.
+  PERFETTO_CHECK(uuid);  // The UUID must be set at this point.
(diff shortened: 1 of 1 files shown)


src/tracing/service/tracing_service_impl.cc
```diff
@@ -4239,11 +4239,13 @@ void TracingServiceImpl::MaybeLogUploadEvent(const TraceConfig& cfg,
 }
 
 void TracingServiceImpl::MaybeLogTriggerEvent(const TraceConfig& cfg,
+                                              const base::Uuid& uuid,
                                               PerfettoTriggerAtom atom,
                                               const std::string& trigger_name) {
   if (!ShouldLogEvent(cfg))
     return;
-  android_stats::MaybeLogTriggerEvent(atom, trigger_name);
+  PERFETTO_DCHECK(uuid);  // The UUID must be set at this point.
```
LalitMaganti: PERFETTO_CHECK

sashwinbalaji: updated for MaybeLogUploadEvent as well.

PR #4304, opened 2026-01-09
ui: Add optional tooltip for buttons

stevegolton:
Useful for providing more context for icon-only buttons.

image


aMayzner:
Thank you so much!

aMayzner (approved):

PR #4306, opened 2026-01-09
ui: add esbuild support

camillobruni:
Enable esbuild with `ui/run-dev-server --esbuild` which improves incremental rebuild time by 4x (23s => 6s).


primiano:
Very curious about this. do you get a sense of what actually is taking less time?
I don't know what esbuild does, but I don't think it can avoid running TSC. so i suppose it's bundling time? does it use something else other than rollup?

camillobruni:
From what I saw this was purely the bundler time. I don't know exactly why rollup is so slow – but [esbuild](https://esbuild.github.io/) does claim to be much faster. I'm experimenting locally with supporting the [tsgo](https://devblogs.microsoft.com/typescript/progress-on-typescript-7-december-2025/) to see how much that improves the initial build.

LalitMaganti:
The only thing I'll say with tsgo is that it's still very experimental and
missing a bunch of features in the type system. I would really like to
avoid having to debug compiler issues in the UI :)

stevegolton:
If we're just using esbuild as a faster drop-in replacement for rollup and it works, I'd be happy to simply replace it - i.e. don't even give the option.

Edit: As long as the bundle sizes are roughly equivalent.

BTW what system are you using that takes 23 seconds to run an _incremental_ build?

camillobruni:
tsgo: agreed, this would be experimental and only an optional thing aimed at local iterations for the time being
build time: yes, changing some js code and then save-to-UI-reload takes 23s on my linux workstation.
bundle size: haven't looked at that yet, but let's maybe land first with the optional flag first and then flip the defaults in a follow PR?

camillobruni:
Looks like `tsgo` is mostly affecting the initial compile time – but has almost no effect on the incremental build (I guess tsc already handled that well enough).

camillobruni:
Looks like the bundle sizes are close enough?

| file name | rollup dev | esbuild dev | rollup min | esbuild min|
|---|---|---|---|---|
| engine_bundle.js | 212K | 388K | 152K | 156K |
| frontend_bundle.js | 12M | 12M | 4.9M | 5.0M |
| traceconv_bundle.js | 212K | 212K | 84K | 88K |

primiano:
> tsgo: agreed, this would be experimental and only an optional thing aimed at local iterations for the time being build time: yes, changing some js code and then save-to-UI-reload takes 23s on my linux workstation. bundle size: haven't looked at that yet, but let's maybe land first with the optional flag first and then flip the defaults in a follow PR?

Wait can you clarify what an "incremental build" is? because for me that's "run run-dev-server, touch a file after it has built the first time... time until you get a new frontend_bundle.js". That usually takes few seconds not 10-20s. There might be something off with your workstation regardless.

> bundle size: haven't looked at that yet, but let's maybe land first with the optional flag first and then flip the defaults in a follow PR?

I'd much rather figure out what it takes to replace rollup and replace it, rather than ending up in this state of having to deal with two bundlers. Either we take the plunge and go with a better one if esbuild proves to be better... or I would not maintain two.


For TSGO: no, sorry, I would not diverge from the official TS compiler. that means putting ourselves on a very narrow boat and be deemed to hit corner cases all the time.

camillobruni:
I'm experiencing roughly the same 20s incremental rebuild (`ui/run-dev-server`, edit file, worker update popup) time on my virtual workstation (and ~11s on my m1 laptop) with a fresh checkout when modifying [ui typescript code](https://github.com/google/perfetto/blob/main/ui/src/frontend/home_page.ts). A single rollup process is active at 100% CPU during that time – so I doubt there is anything special in my setup. Not sure that fits your definition of a couple of seconds?

I just noticed that [esbuild](https://esbuild.github.io/faq/#production-readiness) is marked as late-stage beta which might be an issue :/.

primiano:
> I'm experiencing roughly the same 20s incremental rebuild (ui/run-dev-server, edit file, worker update popup)

Hmm no that's 2x more than expected at least. On my Lenovo P14 (Which is a fart, my macbook air M3 is faster):

- I add a '.' to a console.log warning in `/ui/src/plugins/dev.perfetto.ExplorePage/index.ts`
- After 10-11s I see the makeManifest log in the console where run-dev-server runs (Which means the notification for frontend bundle is sent before that point, as the manifest comes last)


which laptop and OS are you using? you seem off by 2x

stevegolton:
Looks good in theory, but I'm going to push back on this a little:
1) I don't want to have two bundlers to maintain - we should choose one and deprecate the other one.
2) We have some custom configuration in rollup (at the very least, the embedded sourcemaps), which we would need to port if we want to achieve 1).

Unfortunately I don't particularly have time or that much of a compelling reason to switch bundlers + go through the process of making sure we port everything successfully, at least right now. If you want to take that on then please do.

camillobruni:
Maybe I was mistaken – seems like I don't have to wait for the service worker update alert message for the `frontend_bundle.js` ? If so then I'm well below a 10s refresh iteration which is fine.

primiano:
I have not tested the e2e when you get the alert, but you should get an alert as soon as the file is touched.
maybe you are hitting some experiment where we throttle the server side notifications we use?

primiano:
> I have not tested the e2e when you get the alert, but you should get an alert as soon as the file is touched. maybe you are hitting some experiment where we throttle the server side notifications we use?

Okay I just tried and the time for the notification goes up to 16 seconds. there might be something wrong in that logic.

I think I touched that not too long ago to prevent double notifications when you save files, maybe I did something stupid

primiano:
Did some inotify-based profilig: the actual incremental build part takes 13s on my Lenovo P14 Intel crappy laptop

```
$ inotifywait  --timefmt '%Y-%m-%d %H:%M:%S' --format '%T %w%f %e' -e modify -e create -mr out/ui/ui/dist ui/src/plugins/dev.perfetto.ExplorePage/
Setting up watches.  Beware: since -r was given, this may take a while!
Watches established.
2026-01-12 14:25:29 ui/src/plugins/dev.perfetto.ExplorePage/index.ts MODIFY
2026-01-12 14:25:29 ui/src/plugins/dev.perfetto.ExplorePage/index.ts MODIFY
...
...
2026-01-12 14:25:42 out/ui/ui/dist/v53.0-c774fc2b3/frontend_bundle.js MODIFY
2026-01-12 14:25:42 out/ui/ui/dist/v53.0-c774fc2b3/frontend_bundle.js.map MODIFY
...
2026-01-12 14:25:44 out/ui/ui/dist/v53.0-c774fc2b3/manifest.json MODIFY
```

camillobruni:
I've hacked `isDistComplete` to print the missing file and it was usually `frontend_bundle.js`, not sure if that helps.

primiano:
Also IIRC last time hjd@ profiled it most time was in writing sourcemaps.
If you don't need it (i rarely do) we could add some cmdline to skip them or write them async so it doesn't block the iteration cycle.
I haven't checked if it's still the case though, so needs to be confirmed before

camillobruni:
Ok I think I can close this PR then and see whether I can improve this separately by adding another option to ui/build.js to skip the sourcemaps.

PR #4310, opened 2026-01-09
ui: Add collapsible groups support to the pivot table

stevegolton:
image


stevegolton:
Dropped in favour of https://github.com/google/perfetto/pull/4355

PR #4316, opened 2026-01-09
[chrome] Plumb trackEventDescriptor through extension and Record UI

etiennep-chromium:
This PR updates chrome extension and perfetto record UI to query and use TrackEventDescriptor instead of categories strings.
This carries richer information (tags and description). The UI is functionally equivalent, but this will allow a follow up to improve how categories are displayed in the future (see. chrome://traces/recorder for example)
To maintain forward/backward compatibility:
- New extension version supports both GetCategories and GetTrackEventDescriptor
- New extension version has a fallback for older chrome versions
- Record UI has a fallback for older extension version


ui/src/plugins/dev.perfetto.RecordTraceV2/pages/chrome.ts
```diff
@@ -167,12 +172,13 @@ export class ChromeCategoriesWidget implements ProbeSetting {
...
-  private initializeCategories(cats: string[]) {
-    this.options = cats
+  private initializeCategories(descriptor: protos.TrackEventDescriptor) {
+    this.options = descriptor.availableCategories
+      .filter((cat) => cat != null)
       .map((cat) => ({
-        id: cat,
-        name: cat.replace(DISAB_PREFIX, ''),
-        checked: this.options.find((o) => o.id === cat)?.checked ?? false,
+        id: cat.name!,
```
LalitMaganti: It's weird you have to use !. Either assserDefined or filter these out.

etiennep-chromium: I meant for the filter above to check cat.name.
Fixing it removes the need for these.

LalitMaganti (approved):

PR #4319, opened 2026-01-09
Implement `perfetto --no-clobber` CLI option.

KirillTim:


src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -166,6 +166,7 @@ Usage: %s
                              If using CLONE_SNAPSHOT triggers, each snapshot
                              will be saved in a new file with a counter suffix
                              (e.g., file.0, file.1, file.2).
+  --no-clobber     -n      : Do not overwrite an existing output file.
```
primiano: you have a leftover "-n" here

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -522,6 +525,11 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
 #endif
     }
 
+    if (option == OPT_NO_CLOBBER) {
+      no_clobber_ = true;
```
primiano: rather than creating a member, can we:

- make no_clobber a local variable, like next to parse_as_pbtxt. 
- Add a no_clobber arg to OpenOutputFile

This has the benefit that:
- OpenOutputFile becomes a bit easier to reason about, because you don't have to think about "who can set this no_clobber_ member"
- We don't add state to perfetto cmd which is no necessary, which removes the amount of thinking.

test/cmdline_integrationtest.cc
```diff
@@ -1494,4 +1494,42 @@ TEST_F(PerfettoCmdlineTest, SaveAllForBugreport_LargeTrace) {
...
+  auto perfetto_proc_no_clobber = Exec(
+      "perfetto", {"--time", "1ms", "--no-clobber", "--out", output_file_path});
+
+  auto perfetto_proc =
+      Exec("perfetto", {"--time", "1ms", "--out", output_file_path});
+
+  StartServiceIfRequiredNoNewExecsAfterThis();
+
+  ASSERT_EQ(1, perfetto_proc_no_clobber.Run(&stderr_)) << stderr_;
+  EXPECT_THAT(stderr_, HasSubstr("(errno: 17, File exists)"));
```
primiano: drop errno:17, just "file exists". errno can change on platforms.
Also put the substring you emit as part of the error, not the one that depends on errno (Which can change with sysroot, be capitalized differently etc)

test/cmdline_integrationtest.cc
```diff
@@ -1494,4 +1495,45 @@ TEST_F(PerfettoCmdlineTest, SaveAllForBugreport_LargeTrace) {
   ExpectTraceContainsTestMessagesWithSize(trace, kMsgSize);
 }
 
+TEST_F(PerfettoCmdlineTest, NoClobber) {
+  const std::string original_file_content = "Existing file";
+  const std::string output_file_path = RandomTraceFileName();
+  const std::string expected_error_message =
```
primiano, line 1501: looking "--no-clobber" is sufficient, you don't want to tie it too much to the wording as it adds friction to change the implementation later.

PR #4322, opened 2026-01-09
ui: define new macro schema and setup an auto-migration

LalitMaganti:
We want a way to define macros which is more flexible and allows
us to add new fields. We are still early enough in the lifecycle of
macros that it's OK to do this without annoying too many people. It's
also the last point for us to do this before server extensions
and then macro schema *really* gets difficult to change.

Instead of:
```
{
  "x": [
    ... invocations
  ],
  ...
}
```
we do:
```
[
  {"name: "x", commands: [...]},
  ...
]
```

Which allows us to easily add fields to macros.


docs/visualization/ui-automation.md
```diff
@@ -214,30 +224,34 @@ Add this to **Settings > Macros** for analyzing memory when needed. This macro
...
-        "High Memory Processes (>50MB)"
-      ]
-    }
-  ]
-}
+[
+  {
+    "id": "user.example.MemoryAnalysis",
+    "name": "Memory Analysis",
+    "commands": [
```
stevegolton: Nit: While we're changing things, I'd question the naming of this a little. Maybe this should be called 'run' or 'runCommands' or something. Right now the naming might suggest the macro implements these commands.

PR #4327, opened 2026-01-10
tp: support multi-level package names with prefix ownership

LalitMaganti:
Package names can now be multi-level (e.g., "android.camera" not just
"android"). Packages own their prefix exclusively:
- Package "foo.bar" prevents registration of "foo" (prefix clash)
- Package "foo" prevents registration of "foo.bar" (prefix clash)
- Prefix clashes with different packages always fail
- allow_override only allows replacing the exact same package name

Changes:
- Add IsPackagePrefixOf helper to sql_modules.h
- Add prefix clash checking in RegisterSqlPackage
- Update module name validation to check prefix (not just first component)
- Add FindPackageForModule to look up packages by prefix match
- Update ExecuteInclude to use FindPackageForModule

### DIFF
--- a/src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.cc
+++ b/src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.cc
@@ -971,11 +971,13 @@ base::Status PerfettoSqlEngine::ExecuteInclude(
     return base::OkStatus();
   }
 
-  std::string package_name = sql_modules::GetPackageName(key);
-
-  auto* package = FindPackage(package_name);
+  // Find the package that owns this module by looking for a package whose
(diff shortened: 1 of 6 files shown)


LalitMaganti:
Actually since I've been working on top of this CL I realized the reason *why* we even did it this way: we want atomic replace semantics for packages.

What we really need to change this CL to is:
1) Add support for foo.bar.baz as the package name instead of just foo.
2) Change the logic checking for clashes to check for prefixes as well.

Will do that but changing to draft until I do.

aMayzner:
Is it fine to have package android.jank and module android.jank?

LalitMaganti:
No that wouldn't be allowed and should be prevented by this cl already.

PR #4329, opened 2026-01-10
ui: Rename sched slice aggregation column names for clarity

stevegolton:
- Wall Duration -> CPU Time
- Wall Duration as % of Total -> CPU Time %
- Wall Duration as % of Selection -> CPU Time / Wall Time

While in the area - also remove 'Partition' and 'Group' columns which are for plumbing only and should never have been visible columns.


primiano (commented):
Yup LGTM. thanks

PR #4332, opened 2026-01-10
ui: move to new API for registering SQL modules with TP

LalitMaganti:


LalitMaganti:
See comment on https://github.com/google/perfetto/pull/4327. This is actually not needed anymore.

PR #4343, opened 2026-01-12
actions: add github actions from canary -> stable and main -> canary

LalitMaganti:


.github/workflows/merge-main-to-canary.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Fetch origin/main
+        run: git fetch origin main
+
+      - name: Merge main into canary (prefer main on conflicts)
+        run: |
+          git merge origin/main -X theirs -m "Merge main into canary (automated)"
```
stevegolton, line 44: Is this necessarily enough, as in will this get canary into the exact same state in on main? We usually apply a reverse diff patch after the merge.

PR #4344, opened 2026-01-12
ui: add build watch.lock file to prevent accidental stale dev-servers

camillobruni:
Write the current build PID to ui/watch.lock and prevent running multiple build instances with --watch in the same out dir to prevent live-reload issues.


camillobruni:
This might mitigate one of the causes that I saw slightly weird live-reload behavior. launching via vscode and killing the ssh connection over night sometimes kept a devserver running somehow and I tend to only kill the actual server but not the rest.

stevegolton (approved):
Nice idea!

@primiano WDYT?

ui/build.js
```diff
@@ -922,4 +927,40 @@ function mklink(src, dst) {
...
+    if (fs.existsSync(cfg.outWatchLockFile)) {
+      const pid = fs.readFileSync(cfg.outWatchLockFile, 'utf8').trim();
+      if (pid === process.pid.toString()) {
+        fs.unlinkSync(cfg.outWatchLockFile);
+      } else {
+        console.warn(`Ignoring stale lock file PID ${pid}`)
+      }
+    }
+  } catch (e) {
+  }
```
primiano: why the empty catch? if anybody makes a typo while refactoring this will silently swallow any errors. can it be narrowed (or at least turned into a console.warn?)

camillobruni: yeah probably not needed. if this fails we have other issues. removed the try-catch altogether.

Issue #4110, opened 2025-12-12
Convert no_flush_before_write_into_file into three option enum


sashwinbalaji:
Just putting rough discussion here for tracking to ensure this blocker is fixed before next release.

Instead of silently switching to flushing on every write into file, convert it to:

enum FlushModeBeforeWriteIntoFile {
    MODE_UNKNOWN = 0;
    MODE_AUTO = 1;
    MODE_DISABLED = 2;
    MODE_ENABLED = 3;
}

MODE_AUTO will be the default mode where if:
- file_write_period_ms <= 5sec (= kDefaultWriteIntoFilePeriodMs) we will enable flush on every write
- file_write_period_ms > 5sec we will disable flush on every write

LalitMaganti:
> MODE_AUTO will be the default mode where if:

It's actually the other way round: on larger periods we will flush on every write.

> file_write_period_ms <= 5sec (= kDefaultWriteIntoFilePeriodMs) we will enable flush on every write

Now question: I think we should *still* flush in the first write after 5s have passed since the previous file write flush.

sashwinbalaji:
> It's actually the other way round: on larger periods we will flush on every 

yup, yup sorry wrote wrong updated the description

PR #4351, opened 2026-01-12
tracing: Replace no_flush_before_write_into_file with WriteFlushMode enum

sashwinbalaji:
Replace the no_flush_before_write_into_file boolean with a WriteFlushMode enum to provide granular control over flushing behavior during write_into_file sessions 

WRITE_FLUSH_AUTO (Default): Automatically determines flush frequency. For high-frequency writes (≤5s kDefaultWriteIntoFilePeriodMs), it issues a periodic flush every 5 seconds instead of flushing on every write.
Also give explicit control via ENABLED and DISABLED modes for advanced users to force or suppress flushes before writes.

Fixes: https://github.com/google/perfetto/issues/4110

### DIFF
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -1188,16 +1188,17 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
 
       if (cfg.flush_period_ms() > 0) {
         PERFETTO_LOG(
-            "Warning: flush_period_ms is not recommended and will be ignored. "
-            "Set write_flush_mode to WRITE_FLUSH_DISABLED to use "
-            "flush_period_ms.");
+            "Warning: flush_period_ms is ignored because write_flush_mode is "
+            "in AUTO mode. Set write_flush_mode to WRITE_FLUSH_DISABLED to "
+            "use flush_period_ms.");
       }
     } else if (flush_mode == TraceConfig::WRITE_FLUSH_ENABLED) {
       tracing_session->flush_strategy = TracingSession::FlushStrategy::kOnWrite;
       if (cfg.flush_period_ms() > 0) {
         PERFETTO_LOG(
-            "Warning: flush_period_ms is ignored when using "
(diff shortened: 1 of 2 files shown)


src/tracing/service/tracing_service_impl.cc
```diff
@@ -1164,6 +1171,35 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
...
+        tracing_session->flush_strategy =
+            TracingSession::FlushStrategy::kPeriodic;
+        tracing_session->periodic_flush_ms =
+            static_cast<uint32_t>(kDefaultWriteIntoFilePeriodMs);
+      } else {
+        tracing_session->flush_strategy =
+            TracingSession::FlushStrategy::kOnWrite;
+      }
+
+      if (cfg.flush_period_ms() > 0) {
```
LalitMaganti, line 1189: This is a breaking change right? We should document it as such in the changelog.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -1164,6 +1171,35 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
...
+        PERFETTO_LOG(
+            "Warning: flush_period_ms is not recommended and will be ignored. "
```
LalitMaganti: I wouldn't say just "not recommended" here. I would say it's ignored because it's in auto mode.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2419,7 +2458,12 @@ void TracingServiceImpl::PeriodicFlushTask(TracingSessionID tsid,
   if (!tracing_session || tracing_session->state != TracingSession::STARTED)
     return;
 
-  uint32_t flush_period_ms = tracing_session->config.flush_period_ms();
+  if (tracing_session->flush_strategy !=
```
LalitMaganti: Shouldn't this be a dcheck or check?

sashwinbalaji: agreed made it a check

PR #4355, opened 2026-01-13
ui: DataGrid: Collapsible groups + rollups in pivot mode

stevegolton:
Features:
- Rollups at every group level
- Expandable/collapsible groups
- Expand all / collapse all buttons on the toolbar
- Flat mode toggle switch on the toolbar

Notes:
- The grouping/rollups/ordering is still done in SQL so query virtualization still works perfectly.
- The query right now is somewhat inefficient and doesn't scale well. This will be improved in a future update.
- The toolbar looks a little sparse when in group mode and wastes vertical real estate. This will be fixed in a future update.

image


LalitMaganti:
Tried it out, looks fantastic!

Small feedback, would not block on this CL % first ask:
1) Can we make the non-grouped mode the default to preserve backcompat and for performance. The grouped approach feels sluggish even just on example Android trace
2) The toolbar approach right now feels very ad-hoc. I guess we have plans to improve how this looks?
3) I feel the group/non group selector should be more like a swtich or segmented rather than a button. It should be clear it's two state which are quite different, not a one-shot action. Sokmething closer to top-down/botom up selector in flamegraph

stevegolton:
Thanks for the feedback!

> 1. Can we make the non-grouped mode the default to preserve backcompat and for performance. The grouped approach feels sluggish even just on example Android trace

Yes, makes perfect sense! Will change.

> 2. The toolbar approach right now feels very ad-hoc. I guess we have plans to improve how this looks?

Agreed. It's the same toolbar where filters are displayed when you have any, but without much else it admittedly looks pretty sparse. I attempted to improve the look and feel but it ended up being a rabbit hole. Will improve in future.

> 3. I feel the group/non group selector should be more like a swtich or segmented rather than a button. It should be clear it's two state which are quite different, not a one-shot action. Sokmething closer to top-down/botom up selector in flamegraph

Makes sense, will change.

stevegolton:
@LalitMaganti Fixed 1 and 3, PTAL

PR #4356, opened 2026-01-13
ui: QueryPage: Database explorer + editor tabs

stevegolton:
Add editor tabs + new database explorer in the side panel

Features:
- Editor tabs with separate results list per tab 
- Database explorer in side panel
  - Accordion view
  - Fuzzy search by table name
  - List of columns and types (+ type icons)
  - Run query button which runs that query in a new editor tab

image


ui/src/assets/widgets/tab_strip.scss
```diff
@@ -0,0 +1,69 @@
+// Copyright (C) 2025 The Android Open Source Project
```
aMayzner: 2026 :(

PR #4357, opened 2026-01-13
de: Query node depends on other nodes

aMayzner:
Connect other nodes as $input_X to the Query node

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/sources/sql_source.ts
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/sources/sql_source.ts
@@ -18,12 +18,16 @@ import {
   QueryNodeState,
   NodeType,
   nextNodeId,
+  SecondaryInputSpec,
 } from '../../../query_node';
 import {notifyNextNodes} from '../../graph_utils';
 import {columnInfoFromName, newColumnInfoList} from '../../column_info';
 import protos from '../../../../../protos';
 import {Editor} from '../../../../../widgets/editor';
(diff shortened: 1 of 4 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/sources/sql_source_unittest.ts
```diff
@@ -0,0 +1,374 @@
...
+  describe('onPrevNodesUpdated', () => {
+    it('should not throw when called', () => {
+      const node = new SqlSourceNode({
+        sql: '',
+        trace: mockTrace,
+      });
+
+      expect(() => node.onPrevNodesUpdated()).not.toThrow();
+    });
+  });
```
stevegolton: How valuable is this test, really?

ui/src/plugins/dev.perfetto.ExplorePage/query_builder/nodes/sources/sql_source.ts
```diff
@@ -139,6 +152,20 @@ export class SqlSourceNode implements QueryNode {
     notifyNextNodes(this);
   }
 
+  onPrevNodesUpdated() {
+    // Called when input connections change - trigger re-render
+    m.redraw();
```
stevegolton: I'm nto super convinced this redraw is needed - when is this called?

aMayzner:
> Don't we use the dollar sign `$foo` inside macros? Could this collide with those?

The SQL nodes shouldn't create new artifacts like tables/views/functions/macros. Also, this is already something that is implemented on the C++ level

PR #4359, opened 2026-01-13
TP: prompt to use venv when running diff tests

primiano:


python/requirements.txt
```diff
@@ -1,2 +1,4 @@
-yapf==0.40.2
+pandas>=2.3.3
+protobuf>=6.33.4
 sqlglot==26.9.0
+yapf==0.40.2
```
LalitMaganti: We need to add grpcio here.

PR #4363, opened 2026-01-13
tp: stdlib: Add perf_sample table to stdlib

aMayzner:


LalitMaganti:
I'm fine to LGTM. But I would really like if you also separately migrated *all* the tables over.

PR #4368, opened 2026-01-13
tp: plumb memory_size_mb to metadata tables

sashwinbalaji:
Added it in below tables:
- machine
- metadata

### DIFF
--- a/src/trace_processor/importers/common/machine_tracker.cc
+++ b/src/trace_processor/importers/common/machine_tracker.cc
@@ -61,8 +61,8 @@ void MachineTracker::SetAndroidSdkVersion(int64_t sdk_version) {
   getRow()->set_android_sdk_version(sdk_version);
 }
 
-void MachineTracker::SetMemorySizeMb(uint32_t memory_size_mb) {
-  getRow()->set_memory_size_mb(memory_size_mb);
+void MachineTracker::SetMemorySizeBytes(int64_t memory_size_bytes) {
+  getRow()->set_memory_size_bytes(memory_size_bytes);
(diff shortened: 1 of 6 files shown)


protos/perfetto/common/system_info.proto
```diff
@@ -86,7 +86,8 @@ message SystemInfo {
   // The size of the memory in MB.
   //
   // On Linux/Android: (sysinfo.totalram * sysinfo.mem_unit) / (1024 * 1024).
-  // On Mac: (sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGESIZE)) / (1024 * 1024).
+  // Other POSIX (e.g. Mac): (sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGESIZE)) /
```
LalitMaganti: Please change this to be in bytes not mb. We always prefer to use the base units.

primiano:
sorry arrived late to this. before we start adopting it, can we call it system_ram_mb:
1. memory size is too confusing (can feel some heap profiler stuff)
2. MB is nicer to read and write in queries than doing long multiplications to get bytes
thanks!

primiano:
I'll send a patch!

LalitMaganti:
> MB is nicer to read and write in queries than doing long multiplications to get bytes

I disagree. We always put things in base units in Perfetto to avoid loss of precision and I think this is no different. Otherwise there's an argument of "why are we not putting time in ms because that's easier to read" or same with heap profiling.

You can display it as MB in the UI without making it that way in the tables.

PR #4380, opened 2026-01-14
ui: preserve selected area selection tab across share links

LalitMaganti:


ui/src/core/state_serialization_schema.ts
```diff
@@ -41,6 +41,7 @@ const SELECTION_SCHEMA = z.discriminatedUnion('kind', [
     start: zTime,
     end: zTime,
     trackUris: z.array(z.string()).readonly(),
+    areaSelectionTabId: z.string().optional(),
```
stevegolton: Maybe call this currentSelectionSubTab or something as we might reuse it for more than just the area selection?

stevegolton (approved):
LGTM

PR #4384, opened 2026-01-14
ui: Refactor DataGrid SQL datasource

stevegolton:
- Migrate to QuerySlots.
- Break up SQL datasource implementation into multiple files.
- Materialize perfetto table to make tree queries more efficient - fixes slow group collapsing/expanding and sorting when in pivot+tree mode.
- Fix bug where a flash of stale data is shown while the new data loads when switching between datagrid modes like flat->pivot or pivot->drilldown.


ui/src/components/widgets/datagrid/sql_data_source.ts
```diff
@@ -0,0 +1,250 @@
...
+  constructor(config: DatagridEngineSQLConfig) {
+    this.engine = config.engine;
+    this.sqlSchema = config.sqlSchema;
+    this.rootSchemaName = config.rootSchemaName;
+    this.preamble = config.preamble;
+    this.queue = config.queue ?? new SerialTaskQueue();
+    this.preambleSlot = new QuerySlot<void>(this.queue);
+    this.distinctValuesSlot = new QuerySlot<readonly SqlValue[]>(this.queue);
+    this.parameterKeysSlot = new QuerySlot<readonly string[]>(this.queue);
+    const uuid = shortUuid();
```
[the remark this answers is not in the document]
stevegolton: This is not a secure context in the slightest. It's just used to generate a random uuid for creating unique temporary table names.

ui/src/components/widgets/datagrid/sql_data_source.ts
```diff
@@ -12,1385 +12,235 @@
...
-    }
-
-    const result = await runQueryForQueryTable(
-      this.wrapQueryWithPrelude(query),
+    this.preamble = config.preamble;
+    this.queue = config.queue ?? new SerialTaskQueue();
+    this.preambleSlot = new QuerySlot<void>(this.queue);
+    this.distinctValuesSlot = new QuerySlot<readonly SqlValue[]>(this.queue);
+    this.parameterKeysSlot = new QuerySlot<readonly string[]>(this.queue);
+    const uuid = shortUuid();
```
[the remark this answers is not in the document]
stevegolton: This is not a security context.

ui/src/components/widgets/datagrid/sql_data_source/flat.ts
```diff
@@ -0,0 +1,249 @@
...
+        return result.firstRow({}) as Row;
+      },
+    });
+  }
+
+  getRows(model: FlatModel): DataSourceRows {
+    const {columns, filters = [], pagination, sort} = model;
+
+    // Load the row count first
+    const rowCountResult = this.rowCountSlot.use({
```
LalitMaganti, line 97: Depending on what the query is, 2xing like this can be expensive. Do we really need to do this?

ui/src/components/widgets/datagrid/sql_data_source/flat.ts
```diff
@@ -0,0 +1,249 @@
...
+        filters: serializeFilters(filters),
+      },
+      queryFn: async () => {
+        const query = buildQuery(this.sqlSchema, this.rootSchemaName, {
+          mode: 'flat',
+          columns,
+          filters,
+        });
+        const result = await this.engine.query(
+          `SELECT COUNT(*) as count FROM (${query})`,
```
aMayzner, line 110: Can we maybe get this from the query result (trace processor) rather than executing another query?

ui/src/components/widgets/datagrid/sql_data_source/flat.ts
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Filters can contain values that are not JSON friendly - e.g. Uint8Arrays.
+ * This function returns a serialized version of the filters suitable for use as
+ * a cache key.
+ */
+function serializeFilters(filters: FlatModel['filters']) {
+  return filters?.map((filter) => filterToSql(filter, filter.field));
+}
+
+function buildQuery(
```
aMayzner, line 170: Maybe in the future it would be possible to use structured query here? It materializes the table for you, keeps track of the count and time it took to run it, and you don't have to worry about order of operations or how to translate the operations to SQL

aMayzner (approved):

ui/src/components/widgets/datagrid/sql_data_source/flat.ts
stevegolton: Good point. We need to know the row count in order to know how large to make the virtual table.

1. This hasn't changed  from before - the refactor just moves things around.
2. The other query is paginated so we must run a separate query to get the full row count.
3. The base rollup table is materialized now so this query is a lot cheaper than it used to be.
4. Because of the key, it's only re-run when the columns or the filters change (actually we could probably omit the columns as only the filters really change the number of rows).

ui/src/components/widgets/datagrid/sql_data_source/flat.ts
stevegolton: As above:

We need to know the row count in order to know how large to make the virtual table.

This hasn't changed from before - the refactor just moves things around.
The other query is paginated so we must run a separate query to get the full row count.
The base rollup table is materialized now so this query is a lot cheaper than it used to be.
Because of the key, it's only re-run when the columns or the filters change (actually we could probably omit the columns as only the filters really change the number of rows).

ui/src/components/widgets/datagrid/sql_data_source/flat.ts
stevegolton: Maybe, thought the SQL in question is quite complex, so I fear would essentially have to re-invent all this in SQ or use the raw sql escape hatch at which point there's not much point in using SQ at all.

PR #4385, opened 2026-01-14
ui: reduce binary size impact for Chromium

LalitMaganti:


sashwinbalaji (approved):
LGTM to unblock but sad to see to increase in compile time :(

PR #4398, opened 2026-01-15
protozero: refactor filtering tests to be based on descriptors

LalitMaganti:
Pure refactoring change, no behavior change.

Will work a lot better for upcoming changes to migrate to an annotation
based proto system

### DIFF
--- a/src/trace_processor/importers/proto/winscope/surfaceflinger_layers_rect_computation.cc
+++ b/src/trace_processor/importers/proto/winscope/surfaceflinger_layers_rect_computation.cc
@@ -367,6 +367,15 @@ TraceRectTableId RectComputation::InsertLayerTraceRectRow(
     row.opacity = opacity;
   }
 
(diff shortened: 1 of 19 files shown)


src/protozero/filtering/filter_util.cc
```diff
@@ -150,6 +148,53 @@ bool FilterUtil::LoadMessageDefinition(
...
+  google::protobuf::DescriptorPool pool;
+  for (const auto& file : fds.file()) {
+    if (!pool.BuildFile(file)) {
+      PERFETTO_ELOG("Failed to build file descriptor for %s",
+                    file.name().c_str());
+      return false;
+    }
+  }
+
+  const google::protobuf::Descriptor* root_msg =
```
sashwinbalaji, line 181: nit: introduce using declaration?

LalitMaganti: This will all change anyway. Leaving it for now.

PR #4400, opened 2026-01-15
protozero: switch filter generation to be based on proto annotations

LalitMaganti:
Until now we were using command line arguments to tell the proto filter
generation code "this is a string field", "this is a passthrough", "this
is a semantic type" etc. But it's clear that this is just not scaling
and is making it extremely difficult to reason about which fields are
allowed and how they are processed.

The fix for this is to use the proto itself as the way to define this so
how the field is processed is 1:1 with the field itself. To do this, we
switch to an annotation based approach for defining how fields should be
handled and remove all the complex command line options.


src/protozero/filtering/BUILD.gn
```diff
@@ -89,7 +89,7 @@ source_set("filter_util") {
     "../../../gn:default_deps",
     "../../../gn:protobuf_full",
     "../../../protos/perfetto/common:semantic_type_zero",
-    "../../../protos/perfetto/config:zero",
+    "../../../protos/perfetto/proto_filtering:lite",
```
primiano, line 92: how come do you need lite? shouldn't be enough to have the protoc lib (you have "full" above) and the descriptors?
or is it for the option

primiano (approved):

src/protozero/filtering/BUILD.gn
LalitMaganti: this is for the option parsing yes.

PR #4409, opened 2026-01-15
pbzero: add support for specifying semantic types to be allowed on v2

LalitMaganti:
Allows for backwards compatibility wherw we treat the atrace filter
string as allowed in v2 bytecode, unconditionally running all filters.

### DIFF
--- a/src/protozero/filtering/filter_bytecode_generator.cc
+++ b/src/protozero/filtering/filter_bytecode_generator.cc
@@ -41,7 +41,7 @@ void FilterBytecodeGenerator::EndMessage() {
 // Allows a simple field (varint, fixed32/64, string or bytes).
 void FilterBytecodeGenerator::AddSimpleField(uint32_t field_id) {
   PERFETTO_CHECK(field_id > last_field_id_);
-  bytecode_.push_back(field_id << 3 | kFilterOpcode_SimpleField);
+  bytecode_.push_back(field_id << kOpcodeShift | kFilterOpcode_SimpleField);
   last_field_id_ = field_id;
   endmessage_called_ = false;
 }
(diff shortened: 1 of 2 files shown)


src/protozero/filtering/filter_bytecode_generator.cc
```diff
@@ -70,10 +71,15 @@ void FilterBytecodeGenerator::AddFilterStringFieldWithType(
     bytecode_.push_back(field_id << 3 | kFilterOpcode_FilterStringWithType);
     bytecode_.push_back(semantic_type);
   } else {
-    // On bytecode v2, the field will just be totally denied. Just don't add
-    // anything.
+    // On v2 bytecode, if add_to_v2 is true, emit the field as a simple
+    // filter string (without semantic type). On v54 it will be enhanced via
+    // overlay.
+    if (add_to_v2) {
+      bytecode_.push_back(field_id << 3 | kFilterOpcode_FilterString);
```
sashwinbalaji: can we define some inline function or a macro for << 3 for readabilitiy

src/protozero/filtering/filter_bytecode_generator_unittest.cc
```diff
@@ -143,7 +143,7 @@ TEST(FilterBytecodeGeneratorTest, Nested) {
 TEST(FilterBytecodeGeneratorTest, SemanticTypeOverlayV2) {
   // Test that generating for v2 with semantic types creates an overlay
   FilterBytecodeGenerator gen(FilterBytecodeGenerator::BytecodeVersion::kV2);
-  gen.AddFilterStringFieldWithType(1u, 42u);  // field_id=1, semantic_type=42
+  gen.AddFilterStringFieldWithType(1u, 42u, false);  // field_id=1, type=42
```
sashwinbalaji: nit: , add_to_v2=false

PR #4411, opened 2026-01-15
infra: mark ensure-ci-is-green as cancelled when jobs are cancelled

LalitMaganti:
When upstream jobs are cancelled (e.g., due to concurrency rules), the
ensure-ci-is-green job now cancels itself instead of failing. This
prevents spurious CI failures when a newer commit supersedes an older
one.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -143,19 +143,35 @@ jobs:
     steps:
       - name: Ensure all non-skipped jobs have suceeded
         shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
+          failed=0
+          cancelled=0
+
+          # Possible values: success, failure, cancelled, skipped
+          # Treat unknown values as failures to be defensive.
           check() {
             job="$1"
             result="$2"
-            if [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
+            if [ "$result" = "success" ] || [ "$result" = "skipped" ]; then
+              :
+            elif [ "$result" = "cancelled" ]; then
+              echo "::notice::$job was cancelled"
+              cancelled=1
+            else
               echo "::error::$job failed with result: $result"
-              exit 1
+              failed=1
             fi
           }
           warn() {
             job="$1"
             result="$2"
-            if [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/analyze.yml
```diff
@@ -143,20 +143,30 @@ jobs:
...
+            elif [ "$result" = "cancelled" ]; then
+              echo "::notice::$job was cancelled"
+              cancelled=1
             fi
           }
           warn() {
             job="$1"
             result="$2"
-            if [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
+            if [ "$result" = "failure" ]; then
```
sashwinbalaji: Are these the only possible values of result ?

sashwinbalaji: Acc to this there are more which we might miss ?
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks

Conclusion | Description
-- | --
action_required | The check run provided required actions upon its completion. For more information, see Using the REST API to interact with checks.
cancelled | The check run was cancelled before it completed.
failure | The check run failed.
neutral | The check run completed with a neutral result. This is treated as a success for dependent checks in GitHub Actions.
skipped | The check run was skipped. This is treated as a success for dependent checks in GitHub Actions.
stale | The check run was marked stale by GitHub because it took too long.
success | The check run completed successfully.
timed_out | The check run timed out.

LalitMaganti: I've changed it so that anything which we don't explicitly deal with is a failure.

.github/workflows/analyze.yml
```diff
@@ -143,19 +143,35 @@ jobs:
...
           }
           warn() {
             job="$1"
             result="$2"
-            if [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
+            if [ "$result" = "success" ] || [ "$result" = "skipped" ]; then
+              :
+            elif [ "$result" = "cancelled" ]; then
+              cancelled=1
+            else
```
sashwinbalaji, line 174: a little confusing that for warn optional runs if failed we don't block but if cancelled we cancel everything

LalitMaganti: Fixed.

sashwinbalaji (approved):

PR #4412, opened 2026-01-15
install-build-deps: give actionable instructions for airlock

primiano:


tools/install-build-deps
```diff
@@ -787,6 +787,18 @@ def CheckPythonVenv():
 
 
 def InstallPythonVenv():
+  # Give actionable instructions to Googlers on how to go/corp-airlock.
+  if shutil.which("gpkg"):
+    DEVNULL = subprocess.DEVNULL
+    res = subprocess.run(["gpkg", "auth"], stderr=DEVNULL, stdout=DEVNULL)
+    if res.returncode != 0:
+      print('Python packages require an update. Please run in a new terminal:')
```
LalitMaganti, line 795: I don't think it's a different terminal this runs in. It sets some environment variable so I think you have to run tools/install-build-deps in that terminal *after* you run gpkg setup.

primiano: nono. i thought the same when i saw the print. it's not necessary. 
it actually stores the cookies in a file, i traced with strace (Details in chat)

PR #4414, opened 2026-01-15
traced,tp: memory_size -> system_ram

primiano:
Follow up to #4362 and #4368:
- Rename memory_size -> system_ram to disambiguate
- Add GB variant to make queries easier. I speculate the
  ambiguity of "shoudl I divide by 1024 or 1000" will cause
  a lot of bugs down the line from users.

Bug: b/475245044

### DIFF
--- a/src/trace_processor/importers/common/machine_tracker.cc
+++ b/src/trace_processor/importers/common/machine_tracker.cc
@@ -16,8 +16,6 @@
 
 #include "src/trace_processor/importers/common/machine_tracker.h"
 
-#include <cmath>
-
 #include "src/trace_processor/storage/trace_storage.h"
 
 namespace perfetto::trace_processor {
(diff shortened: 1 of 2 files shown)


src/trace_processor/importers/common/machine_tracker.cc
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "src/trace_processor/importers/common/machine_tracker.h"
+
```
sashwinbalaji, line 18: nit:unwanted line

sashwinbalaji (approved):

src/trace_processor/importers/common/machine_tracker.cc
primiano: tbh was wrong before :)

as per https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
the order is

#include "header_matching_the_cc_file.h"
blank line
other headers

this is to keep the matching header at the top even when you clang-format (clang-format treats \n as a sorting boundary)

sashwinbalaji: my bad missed it was this `header_matching_the_cc_file`

PR #4419, opened 2026-01-15
trace_processor: Support V8IcEvent

camillobruni:
- Add new V8ICEvent trace packet and basic 
- Add basic v8_tracker_unittest.cc
- Add JitCache::FindCode to look jit code entries based on a single address


camillobruni:
Taking a first stab at this – not sure how reasonable my code is :)

PR #4425, opened 2026-01-15
ui: Support tags for selecting chrome trace categories

camillobruni:
Add TabStrip to the chrome record settings with new Tags and the existing Presets.
- Display "tags" if the extension returns them with the TrackEventDescriptor
- Use unwrapResult to properly handle chromeCategoryGetter results
- Improve serialisation logic and support categories, presets and privacy settings 
- Add "Details" header to better distinguish the "manual" settings from the presets/tags
- Add Icons to to various chrome category settings
- Move missing-chrome-extension warning to the top in the chrome probe settings page


camillobruni:
If the latest extension is installed we list the clickable tags, otherwise we keep the original Presets.
Screenshot 2026-01-21 at 19 01 09

stevegolton:
I'm getting a crash when navigating to the record/chrome page on this build. Maybe I don't have the right version installed but I think it should probably handle this more elegantly.

camillobruni:
I see that now too – that's definitely new... let me fix this.

camillobruni:
I've added defensive to code to guard against a missing tracing extension.

camillobruni:
This now depends on PR #4606 to land first.

camillobruni:
Without extension:
Screenshot 2026-02-02 at 14 41 28

With the old extension:
Screenshot 2026-02-02 at 14 40 19

With the latest extension:
Screenshot 2026-02-02 at 14 41 46

PR #4428, opened 2026-01-15
Make partitioned interval merge faster and variadic

rzuklie:
The new algorithm is at least 2x faster in all cases tested, primarily driven by the fact that the input set is never doubled in size (count_by_group creates a row for start and end, doubling the size).

This change also uses __intrinsic_token_apply and ColumnNameList to be able to partition across 1 or more columns.


rzuklie:
Anna, can I remove the underscore on this as well to make it public? It is a very convenient macro I'd like to use elsewhere.

src/trace_processor/perfetto_sql/stdlib/intervals/overlap.sql
```diff
@@ -521,6 +521,13 @@ RETURNS TableOrSubquery AS
     value = 0
 );
 
+-- Helper to unparenthesize a column list wiht __intrinsic_token_apply.
```
aMayzner: typo

aMayzner:
> Anna, can I remove the underscore on this as well to make it public? It is a very convenient macro I'd like to use elsewhere.

Yes, I believe so. The API looks sensible

aMayzner (approved):

src/trace_processor/perfetto_sql/stdlib/intervals/overlap.sql
rzuklie: Fixed

PR #4435, opened 2026-01-16
tp: add new fields to view capture args

BogdanFloris:
This PR adds two new fields to the `viewcapture` args proto definition. It also adds makes sure that the fields can be parsed by the trace processor. See: b/436963256

### DIFF
--- a/src/trace_processor/importers/proto/winscope/viewcapture_args_parser.cc
+++ b/src/trace_processor/importers/proto/winscope/viewcapture_args_parser.cc
@@ -106,6 +106,13 @@ std::optional<ConstChars> ViewCaptureArgsParser::TryDeinternString(
     return DeinternString<InternedData::kViewcaptureWindowNameFieldNumber>(
         iid, snapshot_row_, &ViewCaptureRow::set_window_name);
   }
+  if (base::EndsWith(key.key, "content_description_iid")) {
+    return DeinternString<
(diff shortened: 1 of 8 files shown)


src/trace_processor/importers/proto/winscope/viewcapture_args_parser.cc
```diff
@@ -106,23 +106,35 @@ std::optional<ConstChars> ViewCaptureArgsParser::TryDeinternString(
...
-  if (row) {
-    base::StringView sv(chars);
+  return ConstChars{reinterpret_cast<const char*>(blob.data), blob.size};
+}
+
+template <uint32_t FieldNumber, typename RowRef>
+std::optional<ConstChars> ViewCaptureArgsParser::DeinternString(
+    uint64_t iid,
+    RowRef* row,
+    void (RowRef::*setter)(StringPool::Id)) {
```
PriyankaSPX, line 123: nit: define before DeinternString(uint64_t) ?

PriyankaSPX (approved):
lgtm - please test via the diff tests in tests_viewcapture.py

src/trace_processor/importers/proto/winscope/viewcapture_args_parser.cc
BogdanFloris: Added before.

PR #4437, opened 2026-01-16
20260116 add c protos

erikarn:
Add some more generated c protobuf definitions that are useful for building external providers

* sys_stats and sys_info - to generate linux format system stats / info (from freebsd, I'm working on getting that in useful shape for commit)
* add dependencies upon them
*


primiano:
Hmm i am a bit conflicted on this.
Could you solve your problem by running the C generator and generating the proto you need in your project?

So fare we only generate C bindings for the packets that are strictly necessary to support the core functionalities of the SDK (Track event, and the core trace packet properties).
If we say yes to this, then every time somebody wants to emit a packet for whatsoever (legit) reason, we'd have to keep adding bindings here, which doesn't scale (and creates an API boundary on the name of each event)

I think this would be better solved by you running the generator for the protos you need in your project, rather than requiring that upstream has all the bindings.

sashwinbalaji:
Some discussions on this topic: https://discord.com/channels/629013440635207702/1461769936446820446

PR #4442, opened 2026-01-17
ui: improve perf by caching track height instead of fetching every dom redraw

LalitMaganti:
This CL caches the value of the minimum track height in the track tree
view class instead of doing a localstorage read and a full zod
deserialization once per track every dom redraw. We also make the
setting require reload because we're now caching it.

This cuts DOM redraw times on traces with lots of expanded tracks
by 2x.

Before:

image

After:

image


stevegolton:
> instead of doing a localstorage read and a full zod deserialization once per track every dom redraw

We don't actually do a localstorage fetch every time - just the zod.safeParse(), which is admittedly slower than I would expect.

> We also make the setting require reload because we're now caching it.

I'd really rather not do this if at all possible. Please see the following PR (https://github.com/google/perfetto/pull/4466) which makes setting.get() fast across the board by caching inside SettingsManager instead. It performs the same as this PR, which is expected, but doesn't require the 'requiresReload' flag to be set.

LalitMaganti:
Great thanks! Looks good: your PR is probably even better than this one because on DOM redraws, in TimeAxisPanel [1], there are two reads of the timestampFormat setting (once in that function, once in getTimeAxisOrigin)

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/frontend/timeline_page/time_axis_panel.ts;l=65?q=%22getTimeAxisOrigin%22%20f:perfetto

stevegolton:
> in TimeAxisPanel [1], there are two reads of the timestampFormat setting (once in that function, once in getTimeAxisOrigin)

Right, it's a footgun. One would expect setting.get() to be trivial so ideally it should be. Thanks for finding the this other example!

PR #4443, opened 2026-01-17
ui: Add ColorCache to avoid repeated color computations in render loops

LalitMaganti:
Adds a generic ColorCache<K, V> class that caches color computations
and automatically invalidates when the theme or USE_CONSISTENT_COLORS
flag changes.

Updated CpuSliceTrack and GroupSummaryTrack to use ColorCache instead
of calling colorForThread/colorForTid directly in their render loops.
This avoids repeated hash computations and Color object allocations
for the same utid/pid values on every frame.

### DIFF
--- a/ui/src/components/colorizer.ts
+++ b/ui/src/components/colorizer.ts
@@ -43,6 +43,35 @@ const USE_CONSISTENT_COLORS = featureFlags.register({
   defaultValue: false,
 });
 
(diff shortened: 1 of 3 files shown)


LalitMaganti:
Holding of on this. We might want to do this in a slightly different way (i.e. with a local cache instead).

PR #4449, opened 2026-01-18
rust-sdk: generate new proto bindings

dreveman:
- python: include contrib/rust-sdk/tools in formatted python code
- rust-sdk: generate new proto bindings


LalitMaganti:
Unfortunately the power rails protos are *not* stable for other users outside of Perfetto to depend on. They are a heavy reflection of the way Android power rails work and will change the day those change (which is very possible with recent discussions we've been having).

I would suggest we define some new protos, more suited to you if you need these. Maybe tying in with the "GenericKernel" set of protos.

dreveman:
> Unfortunately the power rails protos are _not_ stable for other users outside of Perfetto to depend on. They are a heavy reflection of the way Android power rails work and will change the day those change (which is very possible with recent discussions we've been having).
> 
> I would suggest we define some new protos, more suited to you if you need these. Maybe tying in with the "GenericKernel" set of protos.

This isn't critical. I mostly needed it for an Android app example but will avoid adding a crate for this if not considered stable.

I dropped the power rails protos from this PR so it's now just fixing some formatting and updating the existing generated proto bindings for the rust sdk. ptal.

LalitMaganti (approved):

PR #4462, opened 2026-01-19
tools: Do not use venv python3 for setting up fresh venv

camillobruni:
Use 'python3' instead of the hardcoded sys.executable for the venv setup.
sys.executable might point to the current venv's python which is removed in InstallPythonVenv.


camillobruni:
I'm not sure this is fully safe but I seem to run into the "python3 does not exist" warning regularly when trying to update / re-install deps.

primiano:
Hmm given that the when we run install-build-deps we just need a python3 interpreter (it's the only case where we don't need to use the venv because we are installing it) it's the one place where we could skip the sys,interpreter

I think the only reason why that code does `sys.executable` rather than `python3` it's purely to deal with `python3` vs `python3.exe` (on windows)

So I think instead of doing all that change, we can simply do 
```
'python3' + ('.exe' if sys.platform.startswith('win') else '')
```

camillobruni:
Ah true, we'll default to another even if the venv is PATH since it's already deleted at that point.

tools/install-build-deps
```diff
@@ -799,28 +788,15 @@ def CheckPythonVenv():
 
 def InstallPythonVenv():
   venv_pip = os.path.join(PYTHON_VENV_BIN_DIR, 'pip3')
-  cur_python_interpreter = sys.executable
+  python_interpreter = 'python3' if sys.platform != 'win32' else 'python3.exe'
```
primiano, line 804: add a comment saying "We cannot use sys.interpreter because it will get deleted when reinstalling venv"

otherwise in a few months I can see myself going "this is silly, we should use sys.executable here" and undo this :D

test/data/collapsed_stack_simple.txt.sha256
primiano: accidental change?

test/data/collapsed_stack_with_comments.txt.sha256
primiano: ditto

PR #4464, opened 2026-01-19
ui: add org.chromium.V8 V8SourceView

camillobruni:
Implement a V8 sources tab that displays all v8_js_script and it's inner functions from the v8.jit module.

Drive-by-fix:
- Use DataGrid's fillHeight attr in more places

### DIFF
--- a/ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
+++ b/ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
@@ -260,6 +260,7 @@ export class V8SourcesTab implements Tab {
       data: this.functionsDataSource,
       schema: v8JsFunctionUiSchema,
       rootSchema: V8_JS_FUNCTION_SCHEMA_NAME,
+      fillHeight: true,
       initialFilters: this.functionsFilters,
       onFiltersChanged: (filters: readonly Filter[]) => {
         this.functionsFilters = filters;
@@ -346,6 +347,7 @@ export class V8SourcesTab implements Tab {
           data: this.dataSource,
           schema: v8JsScriptUiSchema,
           rootSchema: V8_JS_SCRIPT_SCHEMA_NAME,
+          fillHeight: true,
           filters: this.filters,
           onFiltersChanged: (filters: readonly Filter[]) => {
             this.filters = filters;
--- a/ui/src/plugins/org.chromium.V8/v8_runtime_call_stats_tab.ts
+++ b/ui/src/plugins/org.chromium.V8/v8_runtime_call_stats_tab.ts
@@ -129,6 +129,7 @@ export class V8RuntimeCallStatsTab implements Tab {
       toolbarItemsLeft: this.renderGroupFilter(),
       toolbarItemsRight: this.renderExportButton(),
       data: this.dataSource,
+      fillHeight: true,
       initialPivot: {
         groupBy: [{field: 'v8_rcs_group', id: 'v8_rcs_group'}],
         aggregates: [


camillobruni:
Screenshot 2026-01-19 at 18 09 23

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -0,0 +1,391 @@
...
+              EmptyState,
+              {
+                fillHeight: true,
+                icon: 'no_sim',
+                title: 'No script selected',
+              },
+              'Select a script from the list to view details.',
+            )
+          : [
+              m(TabStrip, {
```
stevegolton: Could you possible switch to the `Tabs` component here instead of `TabStrip`. It handles the tab content as well so that's less you have to deal with switching pages manually.

The only major difference is the tab styling, but I'm trying to normalize this across the UI anyway. It would be nice if all tabs used the same component so we could change it all in one go if we wanted to.

camillobruni: ok, if we update the styling it's fine. Currently the `Tabs` component looks quite bad on the the settings page since it has a different background color behing the tabs on either side. Let me see what I can do 👍

camillobruni: oops this comment was for the other PR... nvm. changed to `Tabs`.

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -0,0 +1,390 @@
...
+        CASE
+          WHEN name GLOB 'chrome://*' THEN 'chrome'
+          WHEN name GLOB 'v8/*' THEN 'v8'
+          WHEN name GLOB 'extensions::*' THEN 'extensions'
+          ELSE STR_SPLIT(name, '/', 2)
+        END AS domain
+      FROM v8_js_script;
+    `);
+    this.isReady = true;
+    m.redraw();
```
stevegolton: Nit: You shouldn't need this. Redraws are automatically scheduled after queries complete.

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -0,0 +1,390 @@
...
+      };
+      this.functionsFilters = [
+        {
+          field: 'v8_js_script_id',
+          op: '=',
+          value: id,
+        },
+      ];
+    }
+    m.redraw();
```
stevegolton: As above

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -0,0 +1,390 @@
...
+      v8JsScript: {
+        v8_js_script_id: {
+          title: 'ID',
+          cellRenderer: (value: unknown, row: Row) => {
+            return m(
+              Anchor,
+              {
+                onclick: (e: Event) => {
+                  e.preventDefault();
+                  this.selectScript(row.v8_js_script_id as number);
```
stevegolton: If the user clicks faster than the query can run, this could lead to unbounded head of line blocking.

If possible could you switch to using a QuerySlot for this? I have literally just added this to the codebase but it's perfect for stuff like this.

It would work something like this:

```ts
// At the top of the class ...
private readonly slot = new QuerySlot<ScriptResult>(new SerialTaskQueue());

// In your view function ...
const selectedId = this.selectedId;
const {data, isPending} = this.slot.use({
  key: {id: selectedId}, // Declare that these results depend on this key alone
  retainOn: ['id'], // Optional - keep returning stale data if the id has changed until the new data has loaded - prevents ugly flashes of empty data while queries are running in the background.
  queryFn: () => this.selectScript(selectedId), // Describe how to load the data
});

if (!data) return 'Loading...';
// Render `data` here if defined

// In your Anchor element, just set `this.selectedId` and mithril will automatically redraw and start loading the data if things have changed.
onclick: () => {
  this.selectedId = row.v8_js_script_id as number;
}
```

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
camillobruni: done.

camillobruni:
Tried to us the `QuerySlot` and it seems to look ok :)

PR #4468, opened 2026-01-19
ui: improve performance by reducing frequency of dom redraws

LalitMaganti:
The state at HEAD: on every mouse move, we would trigger a full DOM
redraw because it's possible the tooltip state may have been updated.
This is unnecessary as most of the time when mouse move happens, we're
either staying on the same hovered object *OR* (more common), we're not
hovering over anything at all.

Fix this by keeping track of the previous hovered object and only triggering
a redraw if it really changes.

Before (notice how the dom redraw count goes up as I move my mouse):

[cast.webm](https://github.com/user-attachments/assets/b9b5d7ca-f312-4dfc-af72-d5697701c988)

After (notice how the count doesn't change):

[After screencast.webm](https://github.com/user-attachments/assets/53b0aab8-d2a3-43d2-90d9-154a9a802d80)


LalitMaganti:
Ack happy to park this!

LalitMaganti:
No worries, thanks!

PR #4469, opened 2026-01-19
tp: Time queries in TP rather than in the UI

stevegolton:
Move query timing from the UI into TraceProcessor's RPC layer, returning elapsed time in each query result batch message.

## Problem

Queries are currently timed from when they enter the queue until the response is received. This includes queue wait time, which misleads when debugging. For example, imagine we simultaneously place 3 queries in the queue, the query log might look like this:

- fast query (3ms) <- fine
- slow query (1021ms) <- fine (ish)
- fast query (1024ms)  <- completely wrong: actually too 3ms but spent 1021ms in queue

## Solution

Time queries in TraceProcessor's RPC layer, after they've been dequeued. This provides accurate measurements of actual query execution time.

We cannot time queries in the wasm bridge since it's merely a byte-oriented proxy with no knowledge of the message contents. The RPC layer in trace processor is the most appropriate place as it also works for httpd mode.

This approach doesn't capture time in transit or proto deserialization overhead in the UI, but this is the best we can reasonably do. Query execution and result iteration time are usually the main concerns when debugging performance anyway.

For streaming results, the elapsed time is included in each batch message so the UI can track incremental progress.

### DIFF
--- a/ui/src/trace_processor/query_result.ts
+++ b/ui/src/trace_processor/query_result.ts
@@ -388,7 +388,10 @@ export interface QueryResult {
   // Returns the last SQL statement.
   lastStatementSql(): string;
 
-  // Returns the wall time taken to run the query in milliseconds.
+  // Returns the time spend processing the query so far in milliseconds. This is
+  // timed by trace processor to avoid any time spend in the queue. It's updated
+  // every time we get a new batch to convey the time taken do far, so it should
+  // be monotonically increasing and stable when isComplete() is true.
   elapsedTimeMs(): number;
 }
 
--- a/include/perfetto/ext/trace_processor/rpc/query_result_serializer.h
+++ b/include/perfetto/ext/trace_processor/rpc/query_result_serializer.h
@@ -25,6 +25,8 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include "perfetto/base/time.h"
+
 namespace perfetto {
 
 namespace protos::pbzero {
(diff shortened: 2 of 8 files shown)


protos/perfetto/trace_processor/trace_processor.proto
```diff
@@ -249,6 +249,9 @@ message QueryResult {
 
   // The last statement in the provided SQL.
   optional string last_statement_sql = 6;
+
+  // Time taken in milliseconds to execute the query.
```
LalitMaganti: As it stands, it seems like this will be set on *each* QueryResult proto sent. So this field is not quite accurate as on any non-final fields, it's a provisional, "time taken until this message was sent".

Can we either:
1) clarify this in the proto
2) change the rpc.cc logic to only send this on the final message and add a note of this in the proto

ui/src/trace_processor/query_result.ts
```diff
@@ -589,6 +596,10 @@ class QueryResultImpl implements QueryResult, WritableQueryResult {
           this._lastStatementSql = reader.string();
           break;
 
+        case 7:
+          this._elapsedTimeMs = reader.double();
```
LalitMaganti, line 603: Do we expect to be keep overwritten on every QueryResult proto so the last one is taken?

protos/perfetto/trace_processor/trace_processor.proto
stevegolton: The idea is that you get the incremental time taken so far. Or if you don't care, just check the last message to just get the overall time taken. I'll add that to the comment.

ui/src/trace_processor/query_result.ts
stevegolton: Yes, it gives you the time taken so far, or just the final time if you wait for all messages to come through.

PR #4472, opened 2026-01-19
ui: Indent Toggle description to align with label

camillobruni:
- Change pf-toggle indent to 30px + 0.33em to vertically align with the label
- Use the inherited font size for the toggle description (by removing the font-size css style)

### DIFF
--- a/ui/src/assets/widgets/checkbox.scss
+++ b/ui/src/assets/widgets/checkbox.scss
@@ -31,6 +31,7 @@
 // When transitioning from checked to unchecked, the animation plays in reverse,
 // and the border width is set to 0 right at the end in order to make the tick
 // completely invisible again.
+
 $box-label-padding: 0.33em;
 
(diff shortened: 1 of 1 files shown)


camillobruni:
Screenshot 2026-01-19 at 19 13 44

stevegolton (changes requested):
The padding should be proportional to the font size - why are you doing this please?

camillobruni:
Took now another stab at this by using the scss $box-label-padding variable (this is probably a bit hacky). In turn we need to have the same font size for the description as the label (which is probably a good thing IMO).


Motivation: The left-aligned description makes it a bit harder to scan what belongs to the toggle vs the next paragraph IMO.

Before:
Screenshot 2026-01-19 at 20 03 19

After:
Screenshot 2026-01-19 at 19 59 01

Happy to iterate on this.

PR #4479, opened 2026-01-20
ui: Improve assert messages

stevegolton:
Add a Babel plugin that automatically injects expression text into assertion calls, so `assertExists(foo.bar)` becomes `assertExists(foo.bar, "foo.bar")` at build time.

This adds a lot more context to the assert messages which makes them more useful than what we had before which was fairly opaque messages like 'Value is undefined' or 'Value doesn't exist'.

| Function | Error Message |
|----------|---------------|
| `assertExists(x)` | `` `x` doesn't exist`` |
| `assertDefined(x)` | `` `x` is undefined`` |
| `assertTrue(x)` | `` `x` is falsy`` |
| `assertFalse(x)` | `` `x` is truthy`` |
| `assertIsInstanceOf(x, Y)` | `` `x` is not an instance of Y`` |
| `assertUnreachable(x)` | ``Unreachable code reached when `x` = <value>`` |

Testing: Run the UI and use the command palette to trigger any "Chaos: assert*" command.

### DIFF
--- a/ui/config/babel-plugin-assert-messages.js
+++ b/ui/config/babel-plugin-assert-messages.js
@@ -1,0 +1,92 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


primiano (commented):
Oh this looks great. my only question: did you check if this slows down compilation? that's my only worry. if not ship it!

primiano (approved):

stevegolton:
Ah, good catch. It slows down a clean build on my machine from 27sec -> 34sec.

I'll see if I can speed it up.

stevegolton:
Closing due to limited value this feature adds given that we have the source-maps which can provide infinitely more information and context around an assertion failure.

PR #4481, opened 2026-01-20
ui: Add export button to V8RuntimeCallStatsTab

camillobruni:
We still use an old tool that reads the serialized RCS data to compare between multiple browser versions and pages.
Add an export button that serializes all RCS data from all processes to a json file.

- Add dedicated RcsJsonExporter class
- Use the chrome.process_label[0] as page name if possible, otherwise the unique id

### DIFF
--- a/ui/src/plugins/org.chromium.V8/v8_runtime_call_stats_tab.ts
+++ b/ui/src/plugins/org.chromium.V8/v8_runtime_call_stats_tab.ts
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import m from 'mithril';
-import {download} from '../../base/download_utils';
 import {Icons} from '../../base/semantic_icons';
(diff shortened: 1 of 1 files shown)


ui/src/plugins/org.chromium.V8/v8_runtime_call_stats_tab.ts
```diff
@@ -246,6 +255,117 @@ export class V8RuntimeCallStatsTab implements Tab {
     });
   }
 
+  private renderExportButton() {
+    return m(Button, {
```
stevegolton: Could this be a DownloadToFileButton in order to add some feedback?

camillobruni: oops yes... I forgot to upload the latest local commit – fixed it there already.

PR #4483, opened 2026-01-20
uI: Enable virtual scrolling flag by default

stevegolton:


LalitMaganti:
Just to confirm, this renders the track title off screem right? Ctrl-f should still work?

stevegolton:
> Just to confirm, this renders the track title off screem right? Ctrl-f should still work?

Correct, offscreen titles are still rendered. Ctrl+F still works.

LalitMaganti:
Lgtm then thanks!

PR #4488, opened 2026-01-20
tp: Diff tests can execute structured query

aMayzner:
This change adds support for executing structured queries in diff tests, enabling end-to-end testing of the 
structured query generator.                                                                                 
                                                                                                              
  **Changes:**                                                                                                
  - Added `STRUCTURED_QUERY` test type to the diff test framework                                             
  - Implemented `StructuredQueryTestExecutor` that serializes textproto queries to binary proto and invokes   
  trace processor with `--structured-query` flag                                                              
  - Added `--structured-query` flag to trace processor shell that reads binary proto, generates SQL via       
  `StructuredQueryGenerator`, and executes it                                                                 
  - Added helper `BuildSqlWithModulesAndPreambles()` to construct complete SQL including module imports and   
  preambles

### DIFF
--- a/src/trace_processor/trace_processor_shell.cc
+++ b/src/trace_processor/trace_processor_shell.cc
@@ -902,20 +902,6 @@ Advanced:
  --extra-checks                       Enables additional checks which can catch
                                       more SQL errors, but which incur
                                       additional runtime overhead.
- --structured-query-spec SPEC_PATH    Parses the spec at the specified path and
-                                      makes queries available for execution.
-                                      Spec files must be instances of the
-                                      perfetto.protos.TraceSummarySpec proto.
-                                      If the file extension is `.textproto` then
-                                      the spec file will be parsed as a
-                                      textproto. If the file extension is `.pb`
-                                      then it will be parsed as a binary
-                                      protobuf. Otherwise, heuristics will be
-                                      used to determine the format.
- --structured-query-id ID             Specifies that the structured query with
-                                      the given ID should be executed. The spec
-                                      for the query must exist in one of the
-                                      files passed to --structured-query-spec.
  -e, --export FILE                    Export the contents of trace processor
                                       into an SQLite database after running any
                                       metrics or queries specified.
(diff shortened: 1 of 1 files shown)


src/trace_processor/trace_processor_shell.cc
```diff
@@ -806,6 +853,10 @@ General purpose:
                                       If used with --run-metrics, the query is
                                       executed after the selected metrics and
                                       the metrics output is suppressed.
+ --structured-query FILE              Execute a structured query from FILE.
+                                      The file should contain a binary
```
LalitMaganti: Please make this a separate seciton in advanced for now. And also we should allow taking txt inpiut as well not just binary proto./

aMayzner: It's completely not finished, please don't review it yet!

src/trace_processor/trace_processor_shell.cc
```diff
@@ -806,6 +906,20 @@ General purpose:
                                       If used with --run-metrics, the query is
                                       executed after the selected metrics and
                                       the metrics output is suppressed.
+ --structured-query-spec SPEC_PATH    Parses the spec at the specified path and
```
LalitMaganti: As I requested, please put this in its own section independent of PerfettoSQL. This is not PerfettoSQL, it should be in its own structured query (advanced) section

src/trace_processor/trace_processor_shell.cc
```diff
@@ -894,6 +902,20 @@ Trace summarization:
  --extra-checks                       Enables additional checks which can catch
                                       more SQL errors, but which incur
                                       additional runtime overhead.
+ --structured-query-spec SPEC_PATH    Parses the spec at the specified path and
```
LalitMaganti: As I requested last time, please give this its own section.

aMayzner: Now it's done. Between advanced and metrics v1

PR #4490, opened 2026-01-20
Add an option to traceconv text to skip unknown fields instead of failing

edgararriagag:
An example on how to use the new option:
traceconv text trace trace.pbtxt --skip-unknown

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/input.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/input.sql
@@ -15,6 +15,8 @@
 
 INCLUDE PERFETTO MODULE android.frames.timeline;
 
+INCLUDE PERFETTO MODULE intervals.intersect;
+
 INCLUDE PERFETTO MODULE slices.with_context;
 
 CREATE PERFETTO TABLE _input_message_sent AS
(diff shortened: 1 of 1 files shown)


src/trace_processor/util/protozero_to_text.cc
```diff
@@ -410,13 +414,16 @@ void ProtozeroToTextInternal(const std::string& type,
     const auto* opt_field_descriptor =
         opt_proto_descriptor ? opt_proto_descriptor->FindFieldByTag(field.id())
                              : nullptr;
+    if (skip_unknown_fields && opt_field_descriptor == nullptr) {
+      continue;
```
edgararriagag, line 418: I can also add a log here either always or in some verbose option if we deem it appropriate, for now I just kept it silent.

src/traceconv/main.cc
```diff
@@ -74,7 +74,8 @@ CONVERSION MODES AND THEIR SUPPORTED OPTIONS:
    --full-sort                        Forces full trace sorting
 
  text                                 Converts to human-readable text format
-   (no additional options)
+   --skip-unknown                     Skip unknown fields when converting to text
+                                      (default: fail on unknown fields)
```
LalitMaganti: We don't fail right? We just print it as field number instead.

edgararriagag: You're right, my bad, got mixed up with the parsing side. We fail on the parsing side, not on the serializing, updating...

PR #4491, opened 2026-01-21
ui: Autofocus search input on settings pages

camillobruni:
When selecting a search page (flags, plugins or settings) autofocus directly on the search input to required fewer mouse inputs.

### DIFF
--- a/ui/src/core/page_manager.ts
+++ b/ui/src/core/page_manager.ts
@@ -46,8 +46,8 @@ export class PageManagerImpl implements PageManager {
     // scrolling position is retained between page flips, which can be handy
     // when quickly switching between pages that have long scrolling content
     // such as the timeline page.
-    return Array.from(this.previousPages.entries()).map(
-      ([key, {page, subpage}]) => {
+    return Array.from(this.previousPages.entries())
+      .map(([key, {page, subpage}]) => {
         const open = key === route.page;
         const maybeRenderedPage = this.renderPageForRoute(page, subpage, open);
         // If either the route doesn't exist or requires a trace but the trace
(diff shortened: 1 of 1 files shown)


stevegolton:
Unfortunately as pages are retained in memory even after switching away (in order to avoid losing things like scroll position), this only works the first time the pages are navigated to, which feels a little janky.

I'm not aware of an elegant solution to this problem, but open to suggestions.

ui/src/public/page.ts
```diff
@@ -51,6 +52,40 @@ export interface PageHandler {
...
+export type FocusPageAttrs = HTMLInputAttrs & {
+  readonly open?: boolean;
+};
+
+/**
+ * Abstract class that tracks whether the current Page is open or not.
+ * Whenever the Page transitions from closed to open, the focus() methods is
+ * called.
+ */
+export abstract class FocusPage<T extends FocusPageAttrs>
```
camillobruni, line 69: Likely not the right place to add this – but this was the easiest for now.

camillobruni:
Took another stab at this by adding a FocusPage that internally tracks the `open` state and update `PageManagerImpl` to pass along the `open` state to the render method too.

stevegolton (commented):
I'm not particularly keen on this approach because:
- Manual drilling of the 'open' state through potentially many layers of components.
- FocusPage needs oncreate and onupdate but it's all to easy to override by mistake in subclases.
- Only works with pages (not subpages, or tabs, which also mostly use Gate to control visibility).

I've been playing around with my take on this problem here: https://github.com/google/perfetto/pull/4603 - I'd be interested to hear your thoughts.

PR #4492, opened 2026-01-21
ui: Allow downloading .gz trace files with SafeFilePicker

camillobruni:
window.showSaveFilePicker complains when trying to download a trace with '.pftrace.gz' file ending.
Allowing '.gz' file ending in filePickerAcceptTypes solves this problem

### DIFF
--- a/ui/src/frontend/sidebar.ts
+++ b/ui/src/frontend/sidebar.ts
@@ -99,7 +99,7 @@ function downloadTrace(trace: TraceImpl) {
   const filePickerAcceptTypes = [
     {
       description: 'Perfetto trace',
-      accept: {'*/*': ['.pftrace', '..gz']},
+      accept: {'*/*': ['.pftrace', '.gz']},
     },
   ];
   if (src.type === 'URL') {


camillobruni:
Not sure whether this is a chrome-canary issue only, just FYI this is the warning popup I get otherwise:
Screenshot 2026-01-21 at 10 45 13

camillobruni:
Now I wonder whether we should explicitly set 'application/gzip' in the `ARRAY_BUFFER` and add a filePickerAcceptTypes setting just for that?

primiano:
> Now I wonder whether we should explicitly set 'application/gzip' in the `ARRAY_BUFFER` and add a filePickerAcceptTypes setting just for that?

A bunch of thoughts

The problem is that not all traces are gzipped. so you can do that only if you know that the trace was gzipped in the first place. Which is a bit trickier when traces are pushed via postmessage or url.

Even worse, a trace can be "internally gziped" (if you record it with `COMPRESSION_TYPE_DEFLATE`, it will compress its contents, but not the traceconfig and other metadata, so it's still a valid proto) or "externally gzipped" (if you take a normal trace and gzip it).

But I think in your case you only care about what was the original file name (which can be pushed as well, but it's optional), so maybe add that only if the original file was .gz. Alternatively you could inspect the firs two bytes of the blob, and if they are `0x1f8b` it's a gzip.

Note: in reality the extension is irrelevant for us. you can even call files .txt. it changes nothing in the way we processes them. what really matters is the heuristic in TraceProcessor that guesses the file type by inspecting its content.

But then there is a good value for having the file being called .gz, it prevents people to go "oh i'll gzip it to make it smaller" ending up with traces that are doubly gzipped, which we don't support

camillobruni:
ACK, so then I guess it's fine if we just allow the `.gz` extension and let the user decide if they want to mess with it or not :).

PR #4493, opened 2026-01-21
ui: Full track virtualization + custom track search

stevegolton:
- "Full" virtualization for tracks (no DOM nodes created at all for offscreen tracks). Significantly reduces jank when displaying long track lists.
  - E.g. 30ms -> 2ms for DOM redraws with 10k tracks open
- A new track search feature mapped to Mod+F, that works just like the browser search but works on virtualized tracks + offers more features.
  - Regex option
  - Search within collapsed track groups option
  - A help text with links to releveant settings:
    - Enable/disable full virtualization
    - Change hotkey to Shift+F instead of overriding Mod+F

### Screenshot

image

### Demo
[c7daa3ec-ddca-41e1-ae2c-91cd9c6e0fd4.webm](https://github.com/user-attachments/assets/66d88f7b-2486-48b7-a8e3-36da10e15c85)

### Notes
1. Why is this not in the omnibox?
  a. **Simpler implementation:** Track search is only relevant on the timeline page, and the hotkey should only be active when on this page. Putting the dialog inside the timeline page means we can keep everything including all the search logic on the timeline page and avoid having to have hooks to make the onmibox and allow hotkeys to page-aware.
  b. **Familiar experience to browser search and other applications:** Browser search appears as a little box in the top right, vscode has different search boxes per panel that appear in the top right, codemirror also has a unique find/replace box that appears when pressing Mod+F, which appears at the bottom, but within the bounds of the editor window.
  c. **Separate dialogs makes it possible to run commands while searching, or run event searches and simultaneously search for tracks:** The problem with the omnibox is it can really only support one long-lived operation at a time, and right now that's event search. It's possible to quickly run a one-shot operation (command, query) while in the middle of an event search as the UI just hides the search while we're doing our operation, but if we wanted to do a track search *and* search for events, it's not clear which one takes precedence. The omnibox is a great entry point - a place to kick off one shot tasks, but it shouldn't really retain any long lived state itself. This is the approach that vscode takes for instance. In short, the omnibox needs a redesign and I don't want to block this PR on it.
2. Why is this not combined with or part of track filtering?
  a. **Track filtering is typically more long lived compared to track search:** Separating the two means it's possible to create a long lived track filter, and then run an ephemeral search within those results.

### DIFF
--- a/ui/src/core_plugins/dev.perfetto.Timeline/index.ts
+++ b/ui/src/core_plugins/dev.perfetto.Timeline/index.ts
@@ -130,7 +130,7 @@ class TimelinePage implements m.ClassComponent<TimelinePageAttrs> {
   private trackSearchModel: TrackSearchModel = {
     searchTerm: '',
     useRegex: false,
-    searchWithinCollapsedGroups: false,
+    searchWithinCollapsedGroups: true,
   };
   private trackSearchBarVisible = false;
   private trackSearchBarApi?: TrackSearchBarApi;
--- a/ui/src/core_plugins/dev.perfetto.Timeline/track_search_bar.ts
+++ b/ui/src/core_plugins/dev.perfetto.Timeline/track_search_bar.ts
@@ -205,7 +205,7 @@ export class TrackSearchBar implements m.ClassComponent<TrackSearchBarAttrs> {
             onclick: onStepBackwards,
           }),
         },
-        'Previous match (Shift+Enter)',
+        m('', 'Previous match ', m(HotkeyGlyphs, {hotkey: 'Shift+Enter'})),
       ),
       m(
         Tooltip,
@@ -217,7 +217,7 @@ export class TrackSearchBar implements m.ClassComponent<TrackSearchBarAttrs> {
             onclick: onStepForward,
           }),
         },
-        'Next match (Enter)',
+        m('', 'Next match ', m(HotkeyGlyphs, {hotkey: 'Enter'})),
       ),
       m(
         Tooltip,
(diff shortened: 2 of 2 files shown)


LalitMaganti:
Overall this looks reasonable to me but I'd like Primiano to comment if he's happy with this direction before I do a proper review.

primiano:
I played around with it a bit. here are my comments:

- okay overall makes sense. I think we past the point where enough important users deal with 10k tracks (sigh) so it's harder to ignore that
- UX wise I like it, great job. One idle though: maybe the filter bar could also have a filter button [here](https://screenshot.googleplex.com/83GmhBRJPg8SUUc) which turns the currnt search into a filter (As if you went through the other route?) could improve discoverability of track filtering. (happy to discuss separately / different PR) to avoid scope creep
- functionality wise: it's broken XD
  - I searched for traced
  - it found some results
  - then i scrolled down
  - it switched to "no matches"
  
 Somehow it feels like it's only searching in the visible viewport, which goes against the problem we are trying to solve

primiano:
Okay there is one thing to discuss tomorrow: how should it work when you ctrl-f on a details tab below

there is a legitimate use-case of wanting to ctrl-f for some details of a slice, or in the result table. 
my theory:

- We should hook up ctrl-f only if the focus is on the timeline, not anything else (sidebar or tabs or another page)
- We should probably have an escape hatch so that if you "double ctrl-f" we let it buble up to the browser. like... measure the time from the last ctrl-f, if it's <500ms, let it show the default thing. this recovers edge cases like "i'm angry, why did you take away my ctrl-f?"

PR #4494, opened 2026-01-21
perfetto: fix SEGV in ComputePacketHash

sashwinbalaji:
Update MurmurHash Combine from taking pointers.

Bug: b/477506828

Pass std::string_view to MurmurHashCombiner instead of (ptr, size). The previous code was calling hasher.Combine(ptr, size), which was interpreted as two separate calls to CombineOne(). The first call CombineOne(const char*) was invoking the std::string_view(const char*) constructor, which calls strlen(). Since trace buffer slices are not null-terminated, this resulted in a SEGV.

### DIFF
--- a/include/perfetto/ext/base/murmur_hash.h
+++ b/include/perfetto/ext/base/murmur_hash.h
@@ -235,6 +235,12 @@ constexpr bool AllowsHeterogeneousLookup() {
          (IsStringLike<T>() && IsStringLike<U>());
 }
 
+// Helper to detect pointers in Combine(...).
+// Hashing pointers directly is prohibited as it often indicates a misuse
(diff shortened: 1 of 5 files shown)


LalitMaganti (approved):

LalitMaganti:
This is a footgun. Can you also add some check in the combine function to prevent two argument cases with a pointer and a size like this?

primiano:
Uuuh thanks for spotting and fixing this. but also Combine looks very easy to misuse (like i did).
I even quesiton if we need the Combine(T...). can't we have something more explicit like Combine(std::tuple ?

sashwinbalaji:
will fix tuple change in https://github.com/google/perfetto/pull/4507

PR #4495, opened 2026-01-21
ui: WIP support tab.selectionURI to override currentSelectionTab

camillobruni:
First hacky approach to showcase how V8RuntimeCallStatsTab could hijack the current selection and stay focused.
Previously we'd always switch to the current_selection tab in `TraceImpl.onSelectionChange` which breaks usability for custom tabs that render content based on the current selection.

### DIFF
--- a/ui/src/core/trace_impl.ts
+++ b/ui/src/core/trace_impl.ts
@@ -190,7 +190,7 @@ export class TraceImpl implements Trace, Disposable {
       const currentTabUri = this.tabs.currentTabUri;
       const tab = this.tabs.resolveTab(currentTabUri);
       if (tab && tab.content instanceof TraceTab) {
-        this.tabs.showTab(tab.content.getSelectionURL(selection));
+        this.tabs.showTab(tab.content.getSelectionURI(selection));
       } else {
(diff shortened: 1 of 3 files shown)


ui/src/core/trace_impl.ts
```diff
@@ -179,12 +180,20 @@ export class TraceImpl implements Trace, Disposable {
...
+    const {clearSearch = true, switchToCurrentSelectionTab = false} = opts;
     if (clearSearch) {
       this.search.reset();
     }
     if (switchToCurrentSelectionTab && selection.kind !== 'empty') {
       this.tabs.showCurrentSelectionTab();
+    } else {
+      const currentTabUri = this.tabs.currentTabUri;
+      const tab = this.tabs.resolveTab(currentTabUri);
+      if (tab && tab.content instanceof TraceTab) {
```
camillobruni, line 192: this needs some better cleaner approach for sure :) (double dispatch or a custom TraceTabManagger or so)

camillobruni:
Probably that would just work fine for our use-cases and it would definitely remove the ambiguity between toplevel tabs vs selection-based subtabs.

PR #4499, opened 2026-01-21
tp: Don't have to specify the id, ts and dur columns

aMayzner:
Add another SQL function that allows you to specify the id, ts, dur

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/intervals/intersect.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/intervals/intersect.sql
@@ -156,12 +156,12 @@ RETURNS TableOrSubquery AS
 -- Each table can have different column names.
 --
 -- Example:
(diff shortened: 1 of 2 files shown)


src/trace_processor/perfetto_sql/stdlib/intervals/intersect.sql
```diff
@@ -114,6 +151,47 @@ RETURNS TableOrSubquery AS
...
+--
+-- Each table can have different column names.
+--
+-- Example:
+--   SELECT * FROM interval_intersect_with_col_names!(
+--     table1, id1, ts1, dur1,
+--     table2, id2, ts2, dur2,
+--     (partition_col)
+--   )
+CREATE PERFETTO MACRO interval_intersect_with_col_names(
```
LalitMaganti: Please don't make this public.

PR #4501, opened 2026-01-21
tracing: emit TraceProvenance packet

rukkal:
Initial (minimal) implementation of TraceProvenance packet specified in RFC-002: https://github.com/google/perfetto/discussions/4093

This implementation covers the bare minimum needed for the ProtoVM integration

### DIFF
--- a/protos/perfetto/trace/perfetto/trace_provenance.proto
+++ b/protos/perfetto/trace/perfetto/trace_provenance.proto
@@ -1,0 +1,36 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+message TraceProvenance {
+  message Sequence {
+    // The trusted_packet_sequence_id reported in each TracePacket.
+    optional uint32 id = 1;
+
+    // ID of the producer, as per Producer.id.
+    optional int32 producer_id = 2;
+  }
+
+  // Describes a trace buffer and all the sequences writing to it.
+  message Buffer {
+    repeated Sequence sequences = 1;
+  }
+
+  repeated Buffer buffers = 2;
+}
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -122,6 +122,7 @@
 #include "protos/perfetto/common/trace_stats.pbzero.h"  // IWYU pragma: keep
 #include "protos/perfetto/config/trace_config.pbzero.h"
 #include "protos/perfetto/trace/clock_snapshot.pbzero.h"
+#include "protos/perfetto/trace/perfetto/trace_provenance.pbzero.h"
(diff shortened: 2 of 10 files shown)


rukkal:
@primiano I need your review here too. The ProtoVM integration PR is based on this one.

protos/perfetto/trace/perfetto/trace_provenance.proto
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+message TraceProvenance {
+  message Sequence {
+    // The trusted_sequence_id reported in each TracePacket.
```
primiano: typo: trusted_packet_sequence_id (just to make codesearch easy)

protos/perfetto/trace/perfetto/trace_provenance.proto
```diff
@@ -0,0 +1,36 @@
...
+    optional uint32 id = 1;
+
+    // ID of the producer, as per Producer.id.
```
primiano, line 26: just to double-check, this is going to be opaque to whoever reads the trace / trace processor.
You cannot join this with anything else? I guess because this is enough for you today? (just checking, i'm fine if for now we only add partial provenance info)

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2745,6 +2746,7 @@ std::vector<TracePacket> TracingServiceImpl::ReadBuffers(
     EmitUuid(tracing_session, &packets);
     if (!tracing_session->config.builtin_data_sources().disable_system_info()) {
       EmitSystemInfo(&packets);
+      EmitTraceProvenance(tracing_session, &packets);
```
primiano: can you move this one line up outside of the "disable_system_info"? as it's a bit unexpected that disable_system_info affects this.
If anything (I think you did it because some tests annoyingly count the number of expected packets) move it in the group above of "if (!disable_trace_config)" as this sort of data matches more with traceconfig etc than systeminfo

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4103,6 +4105,39 @@ void TracingServiceImpl::EmitSystemInfo(std::vector<TracePacket>* packets) {
...
+    TracingSession* tracing_session,
+    std::vector<TracePacket>* packets) {
+  protozero::HeapBuffered<protos::pbzero::TracePacket> packet;
+  packet->set_timestamp(static_cast<uint64_t>(clock_->GetBootTimeNs().count()));
+  auto* provenance = packet->set_trace_provenance();
+  for (const BufferID buf_id : tracing_session->buffers_index) {
+    TraceBuffer* buf = GetBufferByID(buf_id);
+    if (!buf)
+      continue;
+    auto* buffer_proto = provenance->add_buffers();
```
primiano: can you move this line before the if (!buf)

TBH I cannot think of any reason why the buffer should not exist. but if it does, it would be nicer to emit an "empty" buffer section (wihc is what add_buffers() would do) so that the buffer numbering doesn't go out of sync wiht the buffers declared in the trace config

protos/perfetto/trace/perfetto/trace_provenance.proto
rukkal: done

src/tracing/service/tracing_service_impl.cc
rukkal: Done. Probably a combination of sloppy AI change and sloppy review on my part

src/tracing/service/tracing_service_impl.cc
rukkal: Done

rukkal: > I cannot think of any reason why the buffer should not exist

yeah I guess the `if (!buf)` could even be a PERFETTO_CHECK

protos/perfetto/trace/perfetto/trace_provenance.proto
rukkal: Correct. As discussed offline, this is just the bare minimum needed for the ProtoVM integration.

PR #4507, opened 2026-01-21
Use tuple rather than variadic args

sashwinbalaji:


sashwinbalaji:
Closing the PR as this was done in order to have 1:1 mapping with abseil flat_hash_map so in future if we port there won't be much compatability issues.

PR #4511, opened 2026-01-21
Migrate battery plugins to package_lookup built-in

rzuklie:


src/trace_processor/perfetto_sql/stdlib/android/cpu/cpu_per_uid.sql
```diff
@@ -29,22 +29,15 @@ CREATE PERFETTO TABLE android_cpu_per_uid_track (
...
   total_cpu_millis,
-  (
-    SELECT
-      package_name
-    FROM package_list
-    WHERE
-      uid = track.uid % 100000 AND uid >= 10000
-    LIMIT 1
-  ) AS package_name
+  package_lookup(track.uid) AS package_name
```
LalitMaganti, line 40: I just realised I missed this before: can I ask that this C++ funciton is renamed from package_lookup -> __intrinsic_android_package_lookup and if you want to expose it, please do so through a stdlib module in the android package.

rzuklie: Is there some magic to re-exporting the __intrinsic? Wrapping it in a function like this makes it quite a bit slower on larger tables:

```
CREATE PERFETTO FUNCTION android_package_lookup(uid LONG)
RETURNS STRING AS SELECT __intrinsic_android_package_lookup($uid);
```

LalitMaganti: Yes see https://source.chromium.org/chromium/_/chromium/external/github.com/google/perfetto/+/86e4c1946d31cfe60c06fe9d7b1c207ff365b7b5:src/trace_processor/perfetto_sql/parser/perfetto_sql_parser_unittest.cc;l=221

rzuklie: Oh, cool! I can't seem to get it to work though, it's not finding any functions:

Target function '__intrinsic_android_package_lookup' not found in registry. Make sure it has been registered as an available function for delegation.

Looking at the code, I only see RegisterFunctionAndAddToRegistry called to register the function doing the delegating, but I don't see any connection between RegisterFunction<PackageLookup> and the intrinsic table...

LalitMaganti: Thanks for letting me know. https://github.com/google/perfetto/pull/4651 should fix.

PR #4514, opened 2026-01-22
ui: Optimize BreakdownTracks performance and update Binder Viz plugin

gignat-dev:
This CL introduces significant performance optimizations to the BreakdownTracks
controller and applies them to the Android Binder Visualization plugin to
resolve loading bottlenecks.

Key changes:
1. Iterative Track Creation: Modified `createTracks` to BreakdownTracks from recursive to iterative approach.
   Unlike the original recursive method which issued separate queries for each
   node , this approach fetches the entire hierarchy (Aggregation
   -> Slice -> Pivot) in a single SQL query and builds the tree in memory.
   This drastically reduces sql engine round-trip overhead.

2. Optional Sorting: Introduced a `sortTracks` boolean prop. Previously,
   track sorting required eager computation of max values for every track,
   which accounted for ~90% of the plugin's load time.
   This can now be disabled for performance-critical tracks.

3. Binder Plugin Updates:
   - made `sortTracks` false.
   - removed out Pivot tracks and removed the dependency on the
     `android.binder_breakdown` module, because on some traces the import alone took +40s to complete.

### DIFF
--- a/ui/src/components/tracks/breakdown_tracks.ts
+++ b/ui/src/components/tracks/breakdown_tracks.ts
@@ -238,6 +238,159 @@ export class BreakdownTracks {
       .join('\n');
   }
 
+  async createTracks() {
+    if (this.modulesClause !== '') {
+      await this.props.trace.engine.query(this.modulesClause);
+    }
+
+    if (this.props.aggregationType !== BreakdownTrackAggType.COUNT) {
+      await this.props.trace.engine.query(`
+        CREATE OR REPLACE PERFETTO FUNCTION _ui_dev_perfetto_breakdown_tracks_is_spans_overlapping(
+          ts1 LONG,
+          ts_end1 LONG,
+          ts2 LONG,
+          ts_end2 LONG)
+        RETURNS BOOL
+        AS
+        SELECT (IIF($ts1 < $ts2, $ts2, $ts1) < IIF($ts_end1 < $ts_end2, $ts_end1, $ts_end2));
+
+        ${this.getIntervals()}
+      `);
+    }
+
+    const rootTrackNode = await this.createCounterTrackNode(
+      `${this.props.trackTitle}`,
+      [],
+    );
+
+    const aggColNames = this.props.aggregation.columns;
+    const sliceColNames = this.props.slice?.columns ?? [];
+    const pivotColNames = this.props.pivots?.columns ?? [];
+
+    const allColNames = [];
+    allColNames.push(...aggColNames);
+    allColNames.push(...sliceColNames);
+    allColNames.push(...pivotColNames);
+
+    const query = `
+      SELECT ${allColNames.join(', ')}
+      FROM ${this.props.aggregation.tableName}
+      ${this.sliceJoinClause ?? ''}
+      ${this.pivotJoinClause ?? ''}
+      GROUP BY ${allColNames.join(', ')}
+    `;
+    const res = await this.props.trace.engine.query(query);
+
+    await this.createBreakdownHierarchy(
+      rootTrackNode,
+      res,
+      aggColNames,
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.android.AndroidBinderViz/index.ts
```diff
@@ -55,22 +56,24 @@ export default class implements PerfettoPlugin {
...
-        tsCol: 'ts',
-        durCol: 'dur',
-        joins: [
-          {
-            joinTableName: 'android_binder_client_server_breakdown',
-            joinColumns: ['binder_txn_id'],
-          },
-        ],
-      },
+      // pivots: {
```
LalitMaganti: Please don't comment out code like this - go/comments#comment_out

ui/src/plugins/com.android.AndroidBinderViz/index.ts
```diff
@@ -35,7 +35,8 @@ export default class implements PerfettoPlugin {
     const binderCounterBreakdowns = new BreakdownTracks({
       trace: ctx,
       trackTitle: `Binder ${perspective} Transaction Counts`,
-      modules: ['android.binder', 'android.binder_breakdown'],
+      // modules: ['android.binder', 'android.binder_breakdown'],
```
LalitMaganti: As below.

ui/src/plugins/com.android.AndroidBinderViz/index.ts
```diff
@@ -55,22 +56,24 @@ export default class implements PerfettoPlugin {
...
+      //       joinColumns: ['binder_txn_id'],
+      //     },
+      //   ],
+      // },
+      sortTracks: false,
     });
 
     ctx.defaultWorkspace.addChildInOrder(
-      await binderCounterBreakdowns.createTracks(),
+      // await binderCounterBreakdowns.createTracks(),
```
LalitMaganti: As above

ui/src/components/tracks/breakdown_tracks.ts
LalitMaganti: This is unused now right? If so please remove it.

ui/src/plugins/com.android.AndroidBinderViz/index.ts
gignat-dev: Removed, thank you for the reference

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -461,14 +332,170 @@ export class BreakdownTracks {
...
+    }
 
     return new TrackNode({
       name,
       uri,
       sortOrder: sortOrder !== undefined ? -sortOrder : undefined,
     });
   }
+
+  async createTracks() {
```
LalitMaganti, line 241: It would be nice if this stayed where it was above both because:
1) it makes the diffs cleaner
2) generally public functions should appear before private ones anyway.

ui/src/plugins/com.android.AndroidBinderViz/index.ts
gignat-dev: Done

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Removed

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Agreed, moved it

ui/src/plugins/com.android.AndroidBinderViz/index.ts
gignat-dev: Done

PR #4522, opened 2026-01-22
ui: Metrics page supports trace summary

aMayzner:
Extends the Metrics V2 page to support full trace summary specifications in addition to individual metric specs. Users can now switch between "Metric Spec" mode (for single metrics) and "Full Summary" mode (for complete TraceSummarySpec with multiple metrics, queries, and templates).


LalitMaganti:
Can we please make v2 and full summary the default tabs? Also there should be an option to indicate *which* metric you want to run.

It would also be nice if you could parse out the rows from the metric and show it as a table instead of just showing the proto I'll be honest.

PR #4524, opened 2026-01-22
ui: Android ANR plugin

gignat-dev:
Introduces a new plugin to visualize ANR events from the `android_anrs` stdlib module.

Key changes:
- Adds a "Android ANRs" slice track to the workspace.
- Implements route argument parsing (`packageName`,`processName`, `errorId`) to support deep linking.
- Adds `selectAnrMainThread` logic to automatically pin the ANR track, `scroll` to the app's main thread, and `select` the relevant time interval.

URL params:
  	- `com.android.AndroidAnr` (flag)
  	- `com.android.AndroidAnr.packageName`
  	- `com.android.AndroidAnr.processName`
  	- `com.android.AndroidAnr.errorId`
- If specific filters (packageName, processName, errorId) are provided, the last
ANR matching these filters is selected.
- If only `com.android.AndroidAnr` flag is present, the last ANR
in the trace is selected. This would be the *implicit behaviour*.

Tested manually:
- go to http://localhost:10000/#!?com.android.AndroidAnr, open a
trace and it will automatically select and navigate to the main thread
of the last ANR.
- go to
http://localhost:10000/#!?com.android.AndroidAnr.packageName=com.android.example,
open a trace and it will automatically select and navigate to the main
thread of the last ANR that has the package_name =
`com.android.example'
- similar for 
http://localhost:10000/#!?com.android.AndroidStartup.processName=com.android.example.process and http://localhost:10000/#!?com.android.AndroidStartup.errorId=<uuid\>

### DIFF
--- a/ui/src/plugins/com.android.AndroidAnr/index.ts
+++ b/ui/src/plugins/com.android.AndroidAnr/index.ts
@@ -188,12 +188,26 @@ export default class AndroidAnr implements PerfettoPlugin {
       trackNode.pin();
     }
 
-    // Construct the track URI. Typically, thread tracks use the format /slice_{track_id}.
-    const mainThreadTrackUri = `/slice_${anrInfo.mainThreadTrackId}`;
     const startTime = Time.fromRaw(BigInt(anrInfo.ts));
     const endTime = Time.fromRaw(BigInt(anrInfo.ts + anrInfo.dur));
 
     ctx.onTraceReady.addListener(async () => {
+      // Find the main thread track by its track ID via the track tags.
+      const mainThreadTrackNode = ctx.currentWorkspace.flatTracks.find(
+        (track) => {
+          if (!track.uri) {
+            return false;
+          }
+          const trackDesc = ctx.tracks.getTrack(track.uri);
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.android.AndroidAnr/index.ts
```diff
@@ -0,0 +1,229 @@
...
+    };
+
+    // 1. Pin the Android ANRs track first.
+    const trackNode = ctx.currentWorkspace.getTrackByUri(ANR_TRACK_URI);
+    if (trackNode) {
+      trackNode.pin();
+    }
+
+    // Construct the track URI. Typically, thread tracks use the format /slice_{track_id}.
+    const mainThreadTrackUri = `/slice_${anrInfo.mainThreadTrackId}`;
```
stevegolton: This is brittle and will stop working if we change the URI format - you can filter tracks using their tags and these tracks usually put a tag called 'trackIds' which is an array of containing track ids IIRC

gignat-dev: Done, it was indeed tags.trackIds

PR #4526, opened 2026-01-22
tp: outline functions from bytecode interpreter

LalitMaganti:
This outlines a handful of functions from the interpreter which were
quite expensive anyway so the function call overhead was negligible.
Instead put them in the CC file to reduce number of instances and
compile time.

It also sets us up for the infra to do this sort of thing in the future
too.


LalitMaganti:
@primiano FYI

PR #4528, opened 2026-01-22
ui: Add relation visualisations to related events plugin

ivan-chong:
This adds a library consisting of two files: arrow_visualiser and relation_visualiser. 

arrow_visualiser provides a generic arrow drawing API based off flows, which when given a trackUri and a depth of two events (ArrowConnection), draws a line between the two.

relation_visualiser then handles the building of the ArrowConnections for all the related events. It currently assumes that events are joined sequentially in timestamp order.


ivan-chong:
Superseded by #4767

PR #4530, opened 2026-01-22
ui: Add related events plugin

ivan-chong:
This plugin provides an API for future plugins (uploaded in separate PRs) to have the concept of related events. 

These can be defined by any set of criteria defined by a set of rules which retrieve a dataset and the id of a track event, after matching on some related field - e.g. an id in an ATrace name, or an arg from the args table. 

The related events can then be presented to the user through a tab. The tab shows a list of all related events with the ability to navigate to them through clicking on them. The current implementation is a base class, but can be extended to support more sophisticated usecases as demonstrated in future PRs. 

The idea is to also support visualisation of flow-like arrows between them which will also come in a future PR.


ivan-chong:
Superseded by #4767

PR #4531, opened 2026-01-23
tp: tidy up how interpreter works

LalitMaganti:
The previous change to outline was a bit of a papering over a deeper
design flaw in how the interpreter is organized: there's little reaosn
for it to be a templated impl class now that state was factored out:
instead it can just be a bunch of free functions. As a bonus, it
completely removes the templating layer around all the bytecode allowing
easier outlining going forward.

There's going to be further follow up changes to rationlise the
namespacing which has gotten out of hand and needs to be flattened but
I'll do that in a followup.

### DIFF
--- a/src/trace_processor/core/dataframe/impl/bytecode_interpreter_impl.cc
+++ b/src/trace_processor/core/dataframe/impl/bytecode_interpreter_impl.cc
@@ -232,4 +232,4 @@ uint32_t* StringFilterRegexImpl(const StringPool* string_pool,
                 RegexComparator{string_pool});
 }
 
-}  // namespace perfetto::trace_processor::dataframe::impl::bytecode
+}  // namespace perfetto::trace_processor::dataframe::impl::bytecode::ops
--- a/src/trace_processor/core/dataframe/impl/bytecode_interpreter_impl.h
+++ b/src/trace_processor/core/dataframe/impl/bytecode_interpreter_impl.h
@@ -236,6 +236,42 @@ template <typename Comparator, typename ValueType>
   return o_write;
 }
 
+template <typename T>
+inline PERFETTO_ALWAYS_INLINE auto GetComparableRowLayoutReprInteger(T x) {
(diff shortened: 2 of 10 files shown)


src/trace_processor/core/dataframe/impl/bytecode_interpreter_impl.cc
```diff
@@ -228,8 +228,8 @@ uint32_t* StringFilterRegexImpl(const StringPool* string_pool,
   if (!regex.ok()) {
     return output;
   }
-  return Filter(data, begin, end, output, regex.value(),
+  return ops::Filter(data, begin, end, output, regex.value(),
                 RegexComparator{string_pool});
 }
 
-}  // namespace perfetto::trace_processor::dataframe::impl::bytecode::outlined
+}  // namespace perfetto::trace_processor::dataframe::impl::bytecode
```
sashwinbalaji: ::ops

src/trace_processor/core/dataframe/impl/bytecode_interpreter_impl.h
```diff
@@ -208,1381 +236,1384 @@ template <typename Comparator, typename ValueType>
...
+    const InterpreterState& state,
+    const StringPool::Id* data,
+    const uint32_t* begin,
+    const uint32_t* end,
+    uint32_t* output,
+    const char* val) {
+  std::optional<StringPool::Id> id =
+      state.string_pool->GetId(base::StringView(val));
+  if (!id) {
+    memcpy(output, begin, size_t(end - begin));
```
sashwinbalaji, line 1010: this is not new code but just for my understanding doesn't memcpy take the number of bytes and end - begin will give the number of elements only ?

LalitMaganti: Very good catch! I will fix this in a followup with tests, want to keep this CL pure refactor.

PR #4546, opened 2026-01-23
ui: Add Google analytics to opening the Query and Data Explorer pages

aMayzner:
Only for the first open on during the session

### DIFF
--- a/ui/src/core/page_manager.ts
+++ b/ui/src/core/page_manager.ts
@@ -16,6 +16,7 @@ import m from 'mithril';
 import {assertExists, assertTrue} from '../base/logging';
 import {Registry} from '../base/registry';
 import {PageHandler, PageManager} from '../public/page';
+import {Analytics} from '../public/analytics';
 import {Router} from './router';
 import {Gate} from '../base/mithril_utils';
 
@@ -25,6 +26,13 @@ export class PageManagerImpl implements PageManager {
     string,
     {page: string; subpage: string}
   >();
+  // Track the current page for analytics - log when it changes
+  private currentPage: string | undefined;
+  private readonly analytics: Analytics;
+
+  constructor(analytics: Analytics) {
+    this.analytics = analytics;
+  }
 
   registerPage(pageHandler: PageHandler): Disposable {
     assertTrue(/^\/\w*$/.exec(pageHandler.route) !== null);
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/index.ts
```diff
@@ -131,6 +134,14 @@ export default class implements PerfettoPlugin {
     trace.pages.registerPage({
       route: '/explore',
       render: () => {
+        if (!this.hasLoggedPageOpen) {
```
primiano: I don't think we should do this in this way. A few comments

1. Doesn't GA already log the fragment by itself? I don't remember this
2. I'd much rather add something in UImain when we do renderPageForRoute, so we always emit a GA even every time you switch to a new page

What I don't particularly like here is that you are taking some decisions (only log the first event) that should be taken by the server instead (That's what happens with the rest). Send all the page changes, and then we can infer the rest, rather than folding everything here.

Also since we are here would be nice to know which pages are used most, not just about this one page

ui/src/core/page_manager.ts
```diff
@@ -25,6 +26,14 @@ export class PageManagerImpl implements PageManager {
     string,
     {page: string; subpage: string}
   >();
+  // Track the current page for analytics - log when it changes
+  private currentPage: string | undefined;
+  // Getter for analytics to avoid circular dependency with AppImpl
+  private readonly getAnalytics: () => Analytics;
+
+  constructor(getAnalytics: () => Analytics) {
```
primiano: why this is a getter function rather than just passing Analytics?
I think the Analytics itself object doens't change, just at some point we dynaically call .initialize() on it. but it's there all the time since App is created

ui/src/core/page_manager.ts
```diff
@@ -36,6 +45,13 @@ export class PageManagerImpl implements PageManager {
   // Called by index.ts upon the main frame redraw callback.
   renderPageForCurrentRoute(): m.Children {
     const route = Router.parseFragment(location.hash);
+
+    // Log page changes to analytics
+    if (this.currentPage !== route.page) {
+      this.currentPage = route.page;
+      this.getAnalytics().logEvent('User Actions', route.page);
```
primiano: can you please wrap this in a setTimeout(() => this.analytics.log ...., 0) so this doesn't slow down rendering with a fetch request?

PR #4555, opened 2026-01-24
tp: fully decouple interpreter from dataframe

LalitMaganti:
This CL is a pure refactoring.

This CL finally achieves to goal over the last few CLs of making the
interpreter decoupled from the dataframe. This is necessary to allow for
graphs/trees support to be added to the interpreter.

### DIFF
--- a/src/trace_processor/core/dataframe/cursor_impl.h
+++ b/src/trace_processor/core/dataframe/cursor_impl.h
@@ -20,7 +20,6 @@
 #include <cstddef>
 #include <cstdint>
 
-#include "perfetto/base/logging.h"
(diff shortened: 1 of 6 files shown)


src/trace_processor/core/dataframe/cursor_impl.h
```diff
@@ -20,6 +20,7 @@
 #include <cstddef>
 #include <cstdint>
 
+#include "perfetto/base/logging.h"
```
sashwinbalaji: why is this needed ?

LalitMaganti: emoved

PR #4558, opened 2026-01-25
tp: introduce new GetCell method and switch current uses to it

LalitMaganti:
There are two current patterns which are not great:
1) Calling PlanQuery/Execute with no filters or sorts etc. This ships
   multiple full copy of the interpreter for zero reson
2) Calling GetCellLegacy

In both cases, we can migrate to a new GetCell method which allows for
type-safe callback style method for extracting data from a dataframe.
We're also going to use this in upcoming code.

### DIFF
--- a/src/trace_processor/core/dataframe/adhoc_dataframe_builder.cc
+++ b/src/trace_processor/core/dataframe/adhoc_dataframe_builder.cc
@@ -116,8 +116,7 @@ base::StatusOr<Dataframe> AdhocDataframeBuilder::Build() && {
         }));
         break;
       case base::variant_index<DataVariant, core::FlexVector<int64_t>>(): {
-        auto& data =
-            base::unchecked_get<core::FlexVector<int64_t>>(state.data);
+        auto& data = base::unchecked_get<core::FlexVector<int64_t>>(state.data);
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/intrinsics/table_functions/winscope_proto_to_args_with_defaults.cc
```diff
@@ -55,6 +57,28 @@ namespace perfetto::trace_processor {
...
+  void OnCell(int32_t) {}
+};
+
+// Helper to extract int64 values via Cell() callback.
+struct Int64CellExtractor : dataframe::CellCallback {
+  int64_t result = 0;
+  void OnCell(int64_t v) { result = v; }
+  void OnCell(double) {}
+  void OnCell(NullTermStringView) {}
+  void OnCell(std::nullptr_t) {}
```
sashwinbalaji, line 77: why not nullopt ?

LalitMaganti: It's that way when dataframe was written and in the central code nullptr makes more sense than nullopt_t.

PR #4569, opened 2026-01-26
Add the weighted jank metrics

pmuetschard:
Adds the new weighted jank per animation second metric values to the jank metrics based on the frame timeline data.

Bug: b/430220320


src/trace_processor/metrics/sql/android/android_jank_cuj.sql
```diff
@@ -55,6 +55,8 @@ SELECT
               'missed_frames', SUM(app_missed OR sf_missed),
               'missed_app_frames', SUM(app_missed),
               'missed_sf_frames', SUM(sf_missed),
+              'weighted_missed_app_frames', SUM(app_missed * jank_score),
```
nicomazz: don't you need to change some diff test as well?

pmuetschard: Done

PR #4570, opened 2026-01-26
tp: Make metadata per machine-trace

sashwinbalaji:
CHROME_STDLIB_MANUAL_ROLL=Large scale trace processor change


src/trace_processor/importers/common/metadata_tracker.cc
```diff
@@ -32,43 +34,129 @@ namespace perfetto::trace_processor {
...
+// hierarchy. |nullopt| is considered the ultimate ancestor of everything.
+bool IsAncestor(TraceStorage* storage,
+                std::optional<uint32_t> possible_parent,
+                uint32_t child) {
+  if (!possible_parent.has_value()) {
+    return true;
+  }
+  const auto& table = storage->trace_file_table();
+  std::optional<uint32_t> current = child;
+  while (current) {
```
LalitMaganti: Please use for loops.

src/trace_processor/importers/common/metadata_tracker.cc
```diff
@@ -32,43 +34,129 @@ namespace perfetto::trace_processor {
...
+// Returns true if |possible_parent| is an ancestor of |child| in the trace file
+// hierarchy. |nullopt| is considered the ultimate ancestor of everything.
+bool IsAncestor(TraceStorage* storage,
+                std::optional<uint32_t> possible_parent,
+                uint32_t child) {
+  if (!possible_parent.has_value()) {
+    return true;
+  }
+  const auto& table = storage->trace_file_table();
+  std::optional<uint32_t> current = child;
```
LalitMaganti: This should be traceFileTable::Id

src/trace_processor/importers/common/metadata_tracker.cc
```diff
@@ -32,43 +34,129 @@ namespace perfetto::trace_processor {
...
+  while (current) {
+    auto row = table.FindById(tables::TraceFileTable::Id(*current));
```
LalitMaganti: You should CHECK on the result of this.

src/trace_processor/importers/common/metadata_tracker.h
```diff
@@ -28,22 +28,32 @@
...
  public:
-  explicit MetadataTracker(TraceStorage* storage);
+  explicit MetadataTracker(TraceProcessorContext* context);
 
   // Example usage:
   // SetMetadata(metadata::benchmark_name,
   //             Variadic::String(storage->InternString("foo"));
   // Returns the id of the new entry.
-  MetadataId SetMetadata(metadata::KeyId key, Variadic value);
+  MetadataId SetMetadata(
```
LalitMaganti: Move to GlobalMetadataTracker

src/trace_processor/importers/common/global_metadata_tracker.cc
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+
+  // Special case for trace_uuid: it's possible that trace_uuid was set
+  // globally (with trace_id=null) before the actual trace_id was known (e.g. by
+  // TraceProcessorStorageImpl). In this case, we "upgrade" the existing global
+  // entry by associating it with the current trace context instead of
+  // inserting a new row.
+  if (key == metadata::trace_uuid) {
+    for (auto it = metadata_table.IterateRows(); it; ++it) {
```
LalitMaganti: The fact that we need to iterate like this is non ideal and has the potential to become a performance problem. Can we keep a hash map to the machine if instead?

src/trace_processor/importers/common/trace_file_tracker.cc
```diff
@@ -74,10 +86,12 @@ void TraceFileTracker::DoneParsing(tables::TraceFileTable::Id id, size_t size) {
   auto row = *context_->storage->mutable_trace_file_table()->FindById(id);
   row.set_size(static_cast<int64_t>(size));
 
-  // First file (root)
-  if (id.value == 0) {
-    context_->metadata_tracker->SetMetadata(metadata::trace_size_bytes,
-                                            Variadic::Integer(row.size()));
+  // Log trace_size_bytes only for non-container traces.
```
LalitMaganti, line 89: I think you could do this for containers as well honestly

src/trace_processor/perfetto_sql/stdlib/viz/summary/trace.sql
```diff
@@ -17,7 +17,7 @@
 CREATE PERFETTO FUNCTION _is_linux_trace()
```
LalitMaganti: Ohhhh I don't know about this. We might need to check where this is used and figure out what we should do here. At the very least we need to rename this to better reflect the semantics of what we're doing.

src/trace_processor/tables/metadata_tables.py
LalitMaganti: While you're here, can you please move this to the prelude and make this __intrinsic?

src/trace_processor/importers/common/trace_file_tracker.cc
sashwinbalaji: my worry is it will break a lot of people who are used to getting trace_type. But since we are going to make the table __intrinsic, how about we log everyting here inclyuding container and then in perfetto table metadata we drop them ?

src/trace_processor/perfetto_sql/stdlib/viz/summary/trace.sql
sashwinbalaji: nice catch!

src/trace_processor/importers/common/trace_file_tracker.cc
LalitMaganti: Ah I see the point. It's fine just leave it.

sashwinbalaji:
Leaving comment as can't merge until ui change also approved

PR #4576, opened 2026-01-26
ui: Add Android Input Lifecycle Plugin

ivan-chong:
This plugin allows the lifecycle of input events (from twoshay, through system_server and to the frame being drawn) to be computed on the fly for any given input event / frame. 

This facilitates the correlation of janky frames to the input events responsible in a trivial way through
1.  Flow-like arrows between the events, as well as 
2. A tab which lists lifecycles per channel. 

These events can be navigated through by clicking on the names of events through the tab, as well as with left and right arrow keys.


ivan-chong:
Superseded by #4716

Issue #4527, opened 2026-01-22
Feature: weight perf flame graphs by event count instead of sample count.


cphlipot1:
Every perf sample has an event count associated with it (eg. number of cpu cycles, instructions, cache misses, etc.)

Currently other tools like https://github.com/brendangregg/FlameGraph and linux perf's "perf report" weight their stack frames in  calltrees/flamegraphs not by number of samples attributed to that stack frame, but rather the number of events (eg. cpu cycles) attributed to that stack frame.

Depending how sampling is configured weighting by sample counts instead of event counts can present a misleading picture to the user.

LalitMaganti:
Yup makes total sense. I would even switch to this mode as a default and leave sample count as another metric.

Do you want to send a PR? If not, I will probably look into this in a few weeks, I'm currently doing some plumbing work to signifcantly improve flamegraph perf.

cphlipot1:
I can look at putting up a PR for this.

Question though. it looks like the counter table contains the event counts, where as the perf_sample table contains the samples and call stacks, but there is no clear way to join them. Should we add event count to the perf_sample table, or should we use existing schema and join with counter table on ts since no join id is available?

LalitMaganti:
Yes in fact we discussed this very thing in a design weekly a couple of months ago (cc @rsavitski). Let me dig that up and summarize that as an rfc for you.

LalitMaganti:
Please see https://github.com/google/perfetto/discussions/4545. Hope it's clear :)

LalitMaganti:
I believe this should now be fixed?

PR #4579, opened 2026-01-26
add perf_sample_with_counters to stdlib

cphlipot1:
This implements RFC-0015 (#4545) to make it easy to correlate sample callstacks with the associated counter values. The intention is to use this to help weight perf event flamegraphs by the event values, rather than sample values
Bug: #4527

### DIFF
--- a/src/trace_processor/tables/profiler_tables.py
+++ b/src/trace_processor/tables/profiler_tables.py
@@ -532,6 +532,7 @@ PERF_SAMPLE_TABLE = Table(
             CppOptional(CppUint32()),
             cpp_access=CppAccess.READ_AND_HIGH_PERF_WRITE,
             cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
+            sql_access=SqlAccess.HIGH_PERF,
         ),
     ],
     tabledoc=TableDoc(


src/trace_processor/importers/proto/profile_module.cc
```diff
@@ -253,20 +255,35 @@ void ProfileModule::ParsePerfSample(
...
   // Populate the |perf_sample| table with everything except the recorded
   // counter values, which go to |counter|.
-  context_->event_tracker->PushCounter(
+  // Collect counter IDs for counter set association
+  std::vector<CounterId> counter_ids;
+
+  auto timebase_counter_id = context_->event_tracker->PushCounter(
       ts, static_cast<double>(sample.timebase_count()),
       sampling_stream.timebase_track_id);
+  if (timebase_counter_id) {
```
LalitMaganti, line 281: We should be consistent and also update the perf.data format parser which will be used by simpleperf.

src/trace_processor/importers/proto/perf_counter_set_tracker.cc
```diff
@@ -0,0 +1,47 @@
...
+#include "src/trace_processor/tables/profiler_tables_py.h"
+
+namespace perfetto::trace_processor {
+
+PerfCounterSetTracker::PerfCounterSetTracker(TraceProcessorContext* context)
+    : context_(context) {}
+
+uint32_t PerfCounterSetTracker::AddCounterSet(
+    const std::vector<CounterId>& counter_ids) {
+  auto* table = context_->storage->mutable_perf_counter_set_table();
```
LalitMaganti: I don't think this is worth for just this function with no state. I'd suggest just doing this as a function inline honestly.

src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.cc
```diff
@@ -0,0 +1,110 @@
...
+  const char* counter_name = sqlite::value::Text(argv[1]);
+
+  auto* user_data = GetUserData(ctx);
+  auto* storage = user_data->storage;
+
+  // Look up the sample to get counter_set_id.
+  const auto& perf_sample_table = storage->perf_sample_table();
+  if (sample_id >= perf_sample_table.row_count()) {
+    // Invalid sample ID.
+    return;
```
LalitMaganti: This should set error.

src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.cc
```diff
@@ -0,0 +1,110 @@
...
+    // No counter set for this sample.
+    return;
+  }
+
+  // Iterate through the counter set to find the counter with matching name.
+  const auto& perf_counter_set_table = storage->perf_counter_set_table();
+  const auto& counter_table = storage->counter_table();
+  const auto& track_table = storage->track_table();
+
+  for (auto it = perf_counter_set_table.IterateRows(); it; ++it) {
```
LalitMaganti: This is a O(N) scan. What we should really do use memoize a cursor and then use that. See what I do for EXTRACT_ARG.

src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.cc
```diff
@@ -0,0 +1,110 @@
...
+    // Get the track for this counter.
+    TrackId track_id = counter_table[counter_id.value].track_id();
+    if (track_id.value >= track_table.row_count()) {
+      continue;
+    }
+
+    // Check if the track name matches.
+    auto track_name_str =
+        storage->GetString(track_table[track_id.value].name());
+    if (track_name_str == counter_name) {
```
LalitMaganti: Instead of this, would suggest going the other way: lookup the key in the string pool and then do a id equality (or really the cursor should do it for you).

src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.cc
LalitMaganti: You can do this in amortized O(1) time if you do that.

src/trace_processor/perfetto_sql/stdlib/linux/perf/counters.sql
```diff
@@ -0,0 +1,72 @@
...
+-- Returns the counter value for a perf sample given the sample ID
+-- and counter name.
+CREATE PERFETTO FUNCTION linux_perf_counter_for_sample(
+    -- The ID of the perf sample.
+    sample_id JOINID(perf_sample.id),
+    -- The name of the counter (e.g., 'cpu-clock', 'instructions').
+    counter_name STRING
+)
+-- The counter value, or NULL if not found.
+RETURNS DOUBLE AS
```
LalitMaganti: Please use delegates to syntax to avoid the overhead here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/perfetto_sql/parser/perfetto_sql_parser_unittest.cc;l=237?q=%22delegates%20to%22%20f:perfetto&ss=chromium

src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.cc
cphlipot1: are you suggesting we do this for every return case here where id is not found, or just sample_id?

src/trace_processor/perfetto_sql/stdlib/linux/perf/counters.sql
cphlipot1: I tried using "DELEGATE TO" syntax in sql, but at least locally ./tools/format-sql-sources seems to be unable to recognize that syntax:
```
Failed to format SQL: file src/trace_processor/perfetto_sql/stdlib/linux/perf/counters.sql, Expected 'AS'. Line 14, Col: 10.
  tructions').
    counter_name STRING
)
-- The counter value, or NULL if not found.
RETURNS DOUBLE
  DELEGATES TO __intrinsic_perf_counter_for_sample;
  ```
  
Is this what you were suggesting? if yes, do you have a suggestion to handle this error?

**Comment by cphlipot1** [comment]
Thanks. i tried to address feedback but had a few questions.

File: src/trace_processor/importers/proto/perf_counter_set_tracker.cc (hunk 2 above)
Comment (cphlipot1) [reply src/trace_processor/importers/proto/perf_counter_set_tracker.cc]: I got rid of this file.

File: src/trace_processor/importers/proto/profile_module.cc (hunk 1 above)
Comment (cphlipot1) [reply src/trace_processor/importers/proto/profile_module.cc:281]: I added updates to  importers/perf/record_parser.cc . Am i correct in assuming you are talking about that one?

Comment (LalitMaganti) [reply src/trace_processor/importers/proto/profile_module.cc:281]: Yup!

File: src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.cc (hunk 3 above)
Comment (LalitMaganti) [reply src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.cc]: Just when the id is beyond the row count.

File: src/trace_processor/perfetto_sql/stdlib/linux/perf/counters.sql (hunk 6 above)
Comment (LalitMaganti) [reply src/trace_processor/perfetto_sql/stdlib/linux/perf/counters.sql]: Please turn off formatting for the file: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/perfetto_sql/stdlib/intervals/intersect.sql;l=16?q=third_party%2Fperfetto%2Fsrc%2Ftrace_processor%2Fperfetto_sql%2Fstdlib%2Fintervals%2Fintersect.sql&ss=chromium

File: src/trace_processor/perfetto_sql/intrinsics/functions/perf_counter.h [hunk 7]
```diff
@@ -0,0 +1,74 @@
...
+#include "src/trace_processor/storage/trace_storage.h"
+#include "src/trace_processor/tables/profiler_tables_py.h"
+
+namespace perfetto::trace_processor {
+
+// A reusable cursor for looking up perf counters by counter_set_id.
+// This avoids creating a new cursor for each lookup.
+class PerfCounterExtractor {
+ public:
+  explicit PerfCounterExtractor(
```
LalitMaganti: All the functions here can live in the CC file instead of .h

LalitMaganti: Will reduce the headers leaking transitively.

src/trace_processor/importers/perf/perf_counter.h
```diff
@@ -19,11 +19,14 @@
 
 #include <cstdint>
 
+#include "src/trace_processor/storage/trace_storage.h"
 #include "src/trace_processor/tables/counter_tables_py.h"
 #include "src/trace_processor/tables/track_tables_py.h"
 
 namespace perfetto::trace_processor::perf_importer {
 
+using trace_processor::CounterId;
```
LalitMaganti: nit: shouldn't be necessary.

src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h
```diff
@@ -440,8 +440,8 @@ base::Status PerfettoSqlEngine::RegisterFunction(
   function_count_++;
   const char* name = args.name ? args.name : Function::kName;
   int argc = args.argc.has_value() ? args.argc.value() : Function::kArgCount;
-  return engine_->RegisterFunction(name, argc, Function::Step, ctx, nullptr,
-                                   args.deterministic);
+  return RegisterFunctionAndAddToRegistry(name, argc, Function::Step, ctx,
```
LalitMaganti, line 443: Is this leftover from my change? I think this might need an origin/main merge.

src/trace_processor/tables/profiler_tables.py
```diff
@@ -492,6 +525,12 @@
         ),
         C('unwind_error', CppOptional(CppString())),
         C('perf_session_id', CppTableId(PERF_SESSION_TABLE)),
+        C(
+            'counter_set_id',
+            CppOptional(CppUint32()),
+            cpp_access=CppAccess.READ_AND_LOW_PERF_WRITE,
```
LalitMaganti: I would make this HIGH_PERF_WRITE + sql_access HIGH_PERF as well.

LalitMaganti:
Largely LGTM! Just a few last nits.

src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h
cphlipot1: not sure what happened. here, but merge with origin/main made this delta go away.

src/trace_processor/tables/profiler_tables.py
LalitMaganti: This is still not sql_access high perf?

cphlipot1: Sorry about. should be fixed now.

PR #4582, opened 2026-01-26
ui: Add custom details panel for AndroidBinderViz slices

gignat-dev:
This change introduces a custom details panel `BinderSliceDetailsPanel`
for the slices on the binder transactions breakdown tracks,
with the following information:
- transaction type: 'Client binder transaction'
or 'Server binder reply'
- interfaceName
- methodName
- a SliceRef link to the actual thread slice for easy navigation
- Standard slice details via the existing `renderDetails()` method.


ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
```diff
@@ -0,0 +1,132 @@
...
+import {asSliceSqlId, SliceSqlId} from '../../components/sql_utils/core_types';
+import {SliceRef} from '../../components/widgets/slice';
+import {getSlice, SliceDetails} from '../../components/sql_utils/slice';
+import {renderDetails} from '../../components/details/slice_details';
+import {Engine} from '../../trace_processor/engine';
+import {STR_NULL} from '../../trace_processor/query_result';
+import {Tree, TreeNode} from '../../widgets/tree';
+import {Section} from '../../widgets/section';
+
+// Define an interface for the binder transaction details
```
zezeozue: Comment a bit redundant

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
```diff
@@ -0,0 +1,132 @@
...
+interface BinderTxnDetails {
+  side: string;
```
zezeozue: maybe use the word perspective to be consistent with the other parts of the code

gignat-dev: I don't think it's the same meaning ?
From what I understand, in the `Binder server Transaction Counts` where `perspective=server` we consider the `client binder transaction` slices.

zezeozue: ack. maybe add a comment on this? I find 'side' too vague, maybe role?

zezeozue: ack. maybe add a comment? explaining this

gignat-dev: Changed to `txnRole` and added comment `// Whether this is the client or server side of the binder transaction.`

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
gignat-dev: Removed

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
```diff
@@ -0,0 +1,132 @@
...
+  private sliceDetails: SliceDetails | undefined;
+  private binderTxnDetails: BinderTxnDetails | undefined;
+  private isLoading = true;
+
+  constructor(
+    private readonly trace: Trace,
+    private readonly id: bigint,
+  ) {
+    const sliceId = asSliceSqlId(Number(this.id));
+    Promise.all([
```
stevegolton: Please do this loading inside `TrackEventDetailsPanel`'s `async load()` hook.

If the user quickly changes selection faster than these loads are able to complete, we can end up clogging up the trace processor query queue and blocking other queries from being run for a potentially unbounded amount of time.

If you use the load hook, it'll automatically cancel selections that are no longer relevant because they've been skipped over quickly.

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
```diff
@@ -0,0 +1,132 @@
...
+      getSlice(this.trace.engine, sliceId),
+      getBinderTxnDetails(this.trace.engine, sliceId),
+    ])
+      .then(([slice, binderTxn]) => {
+        this.sliceDetails = slice;
+        this.binderTxnDetails = binderTxn;
+      })
+      .catch((error) => {
```
stevegolton: Consider why this can fail and make it visible. Console messages are typically ignored.

- If it's because of a developer error (bad query, missing table) then let it crash so the user has the stacktrace and can raise a bug (we get analytics for this as well).
- If it's because you sometimes expect it to fail for legitimate reasons (e.g. no relevant data in this trace) then show this to the user as a message in the details panel: E.g. 'Unable to load data because you don't have XYZ'.

I suspect it's the former, in which case just let it crash.

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
```diff
@@ -0,0 +1,132 @@
...
+      })
+      .finally(() => {
+        this.isLoading = false;
+        m.redraw();
+      });
+  }
+
+  render() {
+    if (this.isLoading) {
+      return m('.details-panel', m('h2', 'Loading...'));
```
stevegolton: What is this class for? If it's for styling, where are the styles defined?

For constancy with other panels, please use the DetailsShell widget.

stevegolton (changes requested):

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
gignat-dev: Done, used DetailsShell

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
gignat-dev: Done, removed the catch; if it crashes is developer error

ui/src/plugins/com.android.AndroidBinderViz/details_panel.ts
gignat-dev: Done, moved the getSlice() and getBinderTxnDetails() inside `async load()` of TrackEventDetailsPanel

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -26,8 +26,10 @@ import {
   STR,
   LONG_NULL,
   QueryResult,
+  NUM,
 } from '../../trace_processor/query_result';
 import {TrackRenderer} from '../../public/track';
+import {TrackEventDetailsPanel} from 'src/public/details_panel';
```
stevegolton: This import should be relative.

If using vscode you can configure this by adding the following to your settings:

"typescript.preferences.importModuleSpecifier": "relative",

https://perfetto.dev/docs/contributing/build-instructions#ide-setup

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -145,6 +151,10 @@ export interface BreakdownTrackProps {
    * Whether to sort the tracks based on their maximum value.
    */
   sortTracks?: boolean;
+  /**
+   * Optional custom details panel for the slice tracks.
+   */
+  detailsPanel?: new (trace: Trace, id: bigint) => TrackEventDetailsPanel;
```
stevegolton: Having to pass a class here is a little unusual, I'd just make it a callback function and then the caller can return what they like in the callback (including a new class instance)

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -429,6 +441,15 @@ export class BreakdownTracks {
               ${filtersClause}
             `,
           }),
+          detailsPanel: this.props.detailsPanel
+            ? (row: {
+                id: number;
+                ts: bigint;
+                dur: bigint | null;
+                name: string;
+              }) =>
```
stevegolton: You go to some effort to pull out the id from the results here but then you don't even end up using it in BinderSliceDetailsPanel? Maybe just remove it?

gignat-dev: Done, I didn't realize that I already had the slice id when selecting it in the UI

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Done, it's clearer now with the function

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Done

PR #4543, opened 2026-01-23
tp: AnalyzeStructuredQuery takes TraceSummary and analyzes one of the queries inside

aMayzner:
Refactors the structured query analysis API to support query dependencies via `inner_query_id` references. The API now
   takes a `TraceSummarySpec` containing all queries and a single `query_id` to analyze, enabling proper resolution of  
  cross-query references.                                                                                               
                                                                                                                                         
  - Updated StructuredQueryBuilder to use reference-based composition (`innerQueryId`) for single-input operations            
  - Generator's `AddQuery` now returns the query ID for efficient lookup

PR #4593, opened 2026-01-27
tp: Reapply #4588 - Revert most of the structured_query_generator change

aMayzner:
With the latest structured_query_generator change we went from the table creation for each shared query, to the CTE approach, which wasn't actually needed for the purpose of the PR. That introduced a couple of subtle bugs if the structured_query was reused. This mostly reverts to the previous logic, while making sure that the PR is still doing what it was supposed to do in #4543


aMayzner:
> Can we please have a 1:1 test of what failed internally here?

Added and tested manually (using our new Metrics page!)

LalitMaganti (approved):

PR #4594, opened 2026-01-27
tp: Refactor AdhocDataframeBuilder to use Storage instead of DataVariant

LalitMaganti:
Pure refactoring change, no behavior change.

Change ColumnState to use std::optional<Storage> instead of the
custom DataVariant type. This simplifies unnecessarily complex
code required when you use std::variant directly.

Key changes:
- Replace DataVariant with std::optional<Storage> in ColumnState
- Add per-column nullability_type to ColumnState
- Update all methods to work with new Storage-based structure
- Remove global nullability_type_ member variable

### DIFF
--- a/src/trace_processor/core/dataframe/adhoc_dataframe_builder.h
+++ b/src/trace_processor/core/dataframe/adhoc_dataframe_builder.h
@@ -205,7 +205,7 @@ class AdhocDataframeBuilder {
       EnsureNullOverlayExists(state);
     }
     state.null_overlay->push_back_multiple(false, count);
-    if (nullability_type_ == NullabilityType::kDenseNull) {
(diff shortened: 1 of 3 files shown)


src/trace_processor/core/dataframe/adhoc_dataframe_builder.h
```diff
@@ -269,6 +265,16 @@ class AdhocDataframeBuilder {
     return res >= -kMaxDoubleRepresentible && res <= kMaxDoubleRepresentible;
   }
 
+  // Maps value type T to the corresponding core storage type tag.
+  template <typename T>
+  using StorageTypeFor = std::conditional_t<
```
sashwinbalaji: nit how about using template specializiong to make this easier to track

PR #4599, opened 2026-01-28
Split file I/O events by thread

jessemckenna:
This change is a follow up to [Add file I/O ETW events to Perfetto UI](https://github.com/google/perfetto/pull/4116). This change splits the file I/O events into one row per thread, per a reviewer request on the original PR ([discussion](https://github.com/google/perfetto/pull/4116#discussion_r2624358608)).

Before:
Screenshot 2026-01-27 at 4 21 49 PM

After:
Screenshot 2026-01-28 at 10 16 58 AM

### DIFF
--- a/test/trace_processor/diff_tests/parser/etw/tests.py
+++ b/test/trace_processor/diff_tests/parser/etw/tests.py
@@ -56,7 +56,6 @@ class Etw(TestSuite):
         """),
         query="""
         SELECT
-          track.name AS track,
           slice.ts,
           slice.dur,
           slice.name,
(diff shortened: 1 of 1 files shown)


src/trace_processor/importers/etw/file_io_tracker.cc
```diff
@@ -30,12 +30,12 @@ namespace {
...
-// Display file I/O events in a single row titled "ETW File I/O" under the "IO"
-// headeer (per the schema for type "etw_fileio" in `slice_tracks.ts`).
+// Display file I/O events under the "IO" heading. Split data into one row per
+// thread, each labeled "File I/O (<thread ID>)".
 const auto kBlueprint = TrackCompressor::SliceBlueprint(
     "etw_fileio",
-    tracks::DimensionBlueprints(),
-    tracks::StaticNameBlueprint("ETW File I/O"));
+    tracks::DimensionBlueprints(tracks::kThreadDimensionBlueprint),
+    tracks::StaticNameBlueprint("File I/O"));
```
LalitMaganti: I'm confused: how are you getting the thread id appended to the name in the screenshots while keeping this StaticName? That doesn't make sense to me...

jessemckenna: This seems to happen on the UI side. I think it's the logic in [ui/src/public/utils.ts::getTrackName](https://source.chromium.org/chromium/chromium/src/+/ce898154e8c3339b74b4a106869f4d13708b2604:third_party/perfetto/ui/src/public/utils.ts;l=72).

In the latest I removed the name altogether and got automatic thread names e.g., "Thread 123", and put them under a common header, which I think is a bit cleaner (I've updated the screenshot).

LalitMaganti: Are you sure you want this and not properly associated by thread/process? Dunno I feel from Linux syscall analysis, the other feels generally more useful. But up to you.

jessemckenna: I don't have strong feelings on this, but when I set the track name to e.g., "Thread 123", the UI appended "(123)", so it resulted in "Thread 123 (123)" which felt a bit confusing. Since there's already a thread ID field in the data itself (`ttid`), I think it should still be queryable without the track name containing the thread ID. I'm happy to change it either way, though.

etiennep@: do you have any preference?

etiennep-chromium: > Are you sure you want this and not properly associated by thread/process

Does this mean using the track created by track_event_tracker (right under the usual thread / process)? I'd lean that way but I'm not sure what this entail in terms of implementation, do you have an example?

jessemckenna: Resolving this so I can merge the PR (got approval to merge etiennep@ over chat) but I'm happy to do another follow-up if we decide to.

PR #4600, opened 2026-01-28
tp: Save dependent queries as CTEs

aMayzner:
This simplifies the execution significantly - just add shared queries as CTEs each time they are needed, rather than materializing them upfront. This also makes them more independent, as structured query will never materialize anything now. 

This change is required for the current implementation of the Explore Page, in which each node depends on the other nodes in the CTE. As this is a clear simplification, we might go back to the previous implementation.

Add a todo to go back to a tables based lifecycle management when we integrate properly with the Explore Page.


LalitMaganti:
Isn't this going to regress anyone who is using metric computation? This is not just a codepaths used by explore page but also summarization in Google.

aMayzner:
> Isn't this going to regress anyone who is using metric computation? This is not just a codepaths used by explore page but also summarization in Google.

It will regress anyone who is executing multiple metrics using the same shared query. I am working on a different fix, to avoid this regression

PR #4602, opened 2026-01-28
ui: Display warning in chrome probe if extension is not available

camillobruni:
- Display warning about missing perfetto extension in the chrome probe settings
- Use default line-height for Callout to make it more readable


camillobruni:
Especially with the #4425 PR we have to make it a bit more visible why only limited information is visible when no extension is active.

stevegolton (approved):
LGTM

PR #4603, opened 2026-01-28
ui: Autofocus search box when navigating to settings, plugins or flags page

stevegolton:
- Add `GateDetector` component that detects when an ancestor `Gate` opens/closes
- Use it to autofocus and select the search box when navigating to Settings, Plugins, and Flags pages

## Background: The Gate Component

The `Gate` component is used throughout the UI to wrap pages, tabs, and other content that needs to persist its DOM state when hidden. It works by:

1. **Caching children** - When closed, Gate returns the same vnode reference from the previous render, causing Mithril to skip re-rendering that subtree
2. **CSS hiding** - Uses `display: none` to visually hide content while keeping DOM intact
3. **State preservation** - Scroll positions, form inputs, and component state persist across page/tab switches

Gate is used in:
- `PageManager` - wraps each page so switching pages preserves state
- `Tabs` widget - wraps tab content for the same reason
- `DrawerPanel` - wraps bottom panel tabs
- `CurrentSelectionTab` - wraps selection detail tabs

## The Problem

Components inside a Gate can't tell when they become visible vs hidden. The standard Mithril lifecycle hooks (`oncreate`/`onremove`) only fire once when the DOM is actually created/destroyed, not when the Gate opens/closes. This makes it impossible to do things like "focus the search box when the user navigates to this page".

## The Solution: GateDetector

`GateDetector` is a wrapper component that:
1. Finds the closest ancestor Gate via `[data-gate-open]` attribute
2. Uses MutationObserver to watch for attribute changes
3. Fires `onVisibilityChanged(visible, dom)` when visibility changes (e.g. when the gate opens/closes
4. Also fires on initial mount if already visible

Usage:
```typescript
view() {
  return m(GateDetector, {
    onVisibilityChanged: (visible, dom) => {
      if (visible) {
        (dom.querySelector('input') as HTMLInputElement)?.focus();
      }
    }
  }, m('.my-page', ...));
}
```

## Test plan
- Navigate to Settings page → search box should be focused and selected
- Navigate away and back → search box should be focused again
- Same behavior for Plugins and Flags pages
- Typing in search box should work normally

### DIFF
--- a/ui/src/base/mithril_utils.ts
+++ b/ui/src/base/mithril_utils.ts
@@ -117,12 +117,11 @@ export class GateDetector implements m.ClassComponent<GateDetectorAttrs> {
     this.dom = vnode.dom;
 
     // Find closest Gate wrapper using data attribute
-    this.gateElement = (vnode.dom as HTMLElement).closest(
-      '[data-gate-open]',
-    ) as HTMLElement | undefined;
+    const gateElem = this.dom.closest('[data-gate-open]') ?? undefined;
+    this.gateElement = gateElem as HTMLElement | undefined;
     if (!this.gateElement) return;
 
-    this.observer = new MutationObserver(() => this.checkVisibility());
+    this.observer = new MutationObserver(this.checkVisibility.bind(this));
     this.observer.observe(this.gateElement, {
       attributes: true,
       attributeFilter: ['data-gate-open'],


ui/src/base/mithril_utils.ts
```diff
@@ -83,6 +84,72 @@ export class Gate implements m.ClassComponent<GateAttrs> {
...
+  view({children}: m.Vnode<GateDetectorAttrs>): m.Children {
+    return children;
+  }
+
+  oncreate(vnode: m.VnodeDOM<GateDetectorAttrs>) {
+    this.callback = vnode.attrs.onVisibilityChanged;
+    this.dom = vnode.dom;
+
+    // Find closest Gate wrapper using data attribute
+    this.gateElement = (vnode.dom as HTMLElement).closest(
```
camillobruni: nit: `this.dom.closest(...)`

ui/src/base/mithril_utils.ts
```diff
@@ -83,6 +84,72 @@ export class Gate implements m.ClassComponent<GateAttrs> {
...
+      '[data-gate-open]',
+    ) as HTMLElement | undefined;
+    if (!this.gateElement) return;
+
+    this.observer = new MutationObserver(() => this.checkVisibility());
```
camillobruni: nit: I prefer .bind(), slightly more accidental-leak-proof, but no strong opinion:
`    this.observer = new MutationObserver(this.checkVisibility.bind(this));`

ui/src/core_plugins/dev.perfetto.SettingsPage/settings_page.ts
```diff
@@ -130,6 +135,25 @@ export class SettingsPage implements m.ClassComponent<SettingsPageAttrs> {
             }),
       ),
     );
+
+    return m(
+      GateDetector,
+      {
+        onVisibilityChanged: (visible: boolean, dom: Element) => {
+          if (visible) {
+            const input = findRef(
```
camillobruni, line 144: nit: would this be worth a common helper on GateDetector or so?

stevegolton: Not sure I follow you here..?

camillobruni: I mean something like `GateDetector.focusRef(...)` so we don't have to repeat the boilerplate code.

stevegolton: Ack re the boilerplate code. But focusing text inputs by ref is unrelated to GateDetector.

I could get behind a separate utility like `focusInputByRef(...)` perhaps?

camillobruni: SGTM, definitely just nice-to have :)

PR #4606, opened 2026-01-28
ui: Prevent duplicate chrome trace categories

camillobruni:
Chrome's new tracing extension can return duplicate entries for V8 categories
which in turn breaks the MultiSelect view in the chrome probe.

Add a temporary fix until the extension API on the chrome-side is fixed.

### DIFF
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/chrome.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/chrome.ts
@@ -198,19 +198,20 @@ export class ChromeCategoriesWidget implements ProbeSetting {
   }
 
   private initializeCategories(descriptor: protos.TrackEventDescriptor) {
-    const options: MultiSelectOption[] = [];
-    for (const cat of descriptor.availableCategories) {
-      if (!cat.name) continue;
-      options.push({
+    this.options = descriptor.availableCategories
+      .filter(
+        (
+          cat,
+        ): cat is protos.ITrackEventCategory & {
+          name: string;
+        } => cat.name != null,
+      )
+      .map((cat) => ({
         id: cat.name,
         name: cat.name.replace(DISABLED_PREFIX, ''),
         checked: this.options.find((o) => o.id === cat.name)?.checked ?? false,
-      });
-    }
-    options.sort((a, b) =>
-      a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
-    );
-    this.options = options;
+      }))
+      .sort((a, b) => a.name.toLowerCase().localeCompare(b.name.toLowerCase()));
   }
 
   getEnabledCategories(): string[] {


etiennep-chromium:
> Add a temporary fix until the extension API on the chrome-side is fixed.

Nit: I'm not sure this should be fixed in the extension. Chrome returns duplicates because the category exists in both chrome and v8; we can ideally fix that in chrome, but it makes sense for the recorder UI to de-duplicate anyways.

camillobruni:
> > Add a temporary fix until the extension API on the chrome-side is fixed.
> 
> Nit: I'm not sure this should be fixed in the extension. Chrome returns duplicates because the category exists in both chrome and v8; we can ideally fix that in chrome, but it makes sense for the recorder UI to de-duplicate anyways.

fair, maybe we can just fix in the chrome-extension instead of the API?

camillobruni:
I've now updated the PR to filter the duplicate categories in the chrome extension target and made initializeCategories a bit easier to read.

etiennep-chromium:
> to filter the duplicate categories in the chrome extension target

I was confused by this because "chrome extension target" is not the perfetto extension code, but a file in perfetto ui recorder code.

ui/src/plugins/dev.perfetto.RecordTraceV2/pages/chrome.ts
```diff
@@ -182,20 +182,19 @@ export class ChromeCategoriesWidget implements ProbeSetting {
...
-    this.options = descriptor.availableCategories
-      .filter(
-        (
-          cat,
-        ): cat is protos.ITrackEventCategory & {
-          name: string;
-        } => cat.name != null,
-      )
-      .map((cat) => ({
+    const options: MultiSelectOption[] = [];
```
stevegolton: Is this just a different form of the same logic or am I missing something?

stevegolton (approved):
Makes sense, though if we have two identical categories is it just the UI that falls over? As in, can we actually use these two categories in any meaningful way?

My point is, could we generate a different key in the multiselect based on the name + other fields in the category struct to fix the UI or does it break elsewhere?

etiennep-chromium:
The categories are identified by name in the DataSourceConfig as well, so we can't distinguish them in a meaningful way.

stevegolton:
> The categories are identified by name in the DataSourceConfig as well, so we can't distinguish them in a meaningful way.

ACK, thanks.

ui/src/plugins/dev.perfetto.RecordTraceV2/pages/chrome.ts
camillobruni: Yes, just with one intermediate array less and easier to read IMO since we don't need the additional type annotations for the arrow functions.

camillobruni: removed this for now since I don't see typescript complaining anymore.

PR #4609, opened 2026-01-28
Add Megascale Perfetto plugin

bbadawi1:
This adds important statistics upon trace load. This plugin is disabled by default for now.

### DIFF
--- a/ui/src/plugins/org.openxla.Megascale/index.ts
+++ b/ui/src/plugins/org.openxla.Megascale/index.ts
@@ -1,0 +1,127 @@
+// Copyright (C) 2024 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {PerfettoPlugin} from '../../public/plugin';
+import {Trace} from '../../public/trace';
+import {addQueryResultsTab} from '../../components/query_table/query_result_tab';
+
+export default class MegascalePlugin implements PerfettoPlugin {
+  static readonly id = 'org.openxla.Megascale';
+
+  async onTraceLoad(trace: Trace): Promise<void> {
+    trace.onTraceReady.addListener(() => {
+      addQueryResultsTab(trace, {
+        query: `
+          SELECT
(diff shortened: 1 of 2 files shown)


ui/src/plugins/org.openxla.Megascale/index.ts
```diff
@@ -0,0 +1,127 @@
+// Copyright (C) 2024 The Android Open Source Project
```
stevegolton, line 1: Nit 2026

ui/src/plugins/org.openxla.Megascale/index.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export default class MegascalePlugin implements PerfettoPlugin {
+  static readonly id = 'org.openxla.Megascale';
+
+  async onTraceLoad(trace: Trace): Promise<void> {
+    trace.onTraceReady.addListener(() => {
+      addQueryResultsTab(trace, {
+        query: `
+          SELECT
+            a.int_value AS network_latency_us,
```
LalitMaganti, line 27: Drive by: this and below can be simplified tremendously with some careful use of EXTRACT_ARG: https://perfetto.dev/docs/analysis/perfetto-sql-getting-started#helper-functions

ui/src/plugins/org.openxla.Megascale/index.ts
```diff
@@ -0,0 +1,127 @@
...
+import {PerfettoPlugin} from '../../public/plugin';
+import {Trace} from '../../public/trace';
+import {addQueryResultsTab} from '../../components/query_table/query_result_tab';
+
+export default class MegascalePlugin implements PerfettoPlugin {
+  static readonly id = 'org.openxla.Megascale';
+
+  async onTraceLoad(trace: Trace): Promise<void> {
+    trace.onTraceReady.addListener(() => {
+      addQueryResultsTab(trace, {
```
LalitMaganti, line 24: Is there a reason why btw this could not be done with a macro + a startup command - https://perfetto.dev/docs/visualization/ui-automation?

You can even land this stuff in Google3 instead of needing to have a plugin in our repo.

bbadawi1: This could work. I see that I can share macros internally in Google. Is it possible to also share them externally? I only see "Configure them in Settings > Macros and run them via the command palette" in the public documentation.

LalitMaganti: We're actively working on it - see https://github.com/google/perfetto/discussions/3227 and https://github.com/LalitMaganti/perfetto-test-extensions

bbadawi1: This would work. One current limitation is that I cannot name the tabs when using commands in a macro. I sent you https://github.com/google/perfetto/pull/4619 for that.

bbadawi1:
We have the macros working (internally only for now) so I will abandon this PR. If/when you need early testers for new extensions servers feature, we'd love to be involved.

PR #4613, opened 2026-01-29
ui: Add AZERTY keyboard support for navigation

### LINKED ISSUE #3849: Perfetto usability is low with an AZERTY keyboard
Hello, thanks for this tool. 

There's actually a few issues with AZERTY (french) keyboard, and I spotted some [related issues](https://github.com/google/perfetto/issues/463) / PR so I thought I could make the fix myself but the files changed and I don't know enough TS fuckery to be of any use. :(

Issues I spotted: 
1/ In https://ui.perfetto.dev/, the "tooltip" which says "Use "`W` `A` `S` `D`" to navigate the trace." is incorrect (it should be `Z` `Q` `S` `D`). 

Image

2/ Much more annoying, using `Q` (me - azerty) / `A` (you - qwerty) to go to the left opens the "Current selection" tab, wasting 50% of the screenspace. 

https://github.com/user-attachments/assets/a11fd058-0900-48ed-9c0e-63c88d36d62f

BryanBradfo:
Fixes #3849

This PR resolves the conflict between ZQSD navigation on AZERTY keyboards and the "Toggle Drawer" shortcut.

**The Issue:**
On AZERTY keyboards, the physical key for moving left is 'Q' (which maps to `KeyA`). However, Perfetto also bound the 'Q' character key to "Toggle Drawer". This meant that navigating left would unintentionally open/close the sidebar.

**The Fix:**
Following the suggestion from @stevegolton:
1. Added an **"AZERTY layout"** toggle in the Settings (similar to the existing Dvorak mode).
2. When enabled, the "Toggle Drawer" shortcut is remapped from **'Q'** to **'A'**.
   - 'Q' is now free for navigation (Pan Left).
   - 'A' becomes the dedicated key for the drawer (located top-left on AZERTY).

**Test Plan:**
- [x] Enable "AZERTY layout" in Settings.
- [x] Press 'Q' (middle-left key): Confirmed it pans the timeline left without toggling the drawer.
- [x] Press 'A' (top-left key): Confirmed it now toggles the drawer.
- [x] Disable "AZERTY layout": Confirmed behavior reverts to standard QWERTY ('Q' toggles drawer).

Thanks @primiano and @stevegolton for the hints on the implementation!
cc: @aviv-nino (original reporter)

### DIFF
--- a/ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
+++ b/ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
@@ -17,6 +17,7 @@ import {OmniboxMode} from '../../core/omnibox_manager';
 import {Trace} from '../../public/trace';
 import {PerfettoPlugin} from '../../public/plugin';
 import {AppImpl} from '../../core/app_impl';
+import {TraceImpl} from '../../core/trace_impl';
 import {getTimeSpanOfSelectionOrVisibleWindow} from '../../public/utils';
 import {exists, RequiredField} from '../../base/utils';
 import {LONG, NUM, NUM_NULL} from '../../trace_processor/query_result';
@@ -52,16 +53,13 @@ export default class TrackUtilsPlugin implements PerfettoPlugin {
     });
   }
 
-  async onTraceLoad(ctx: Trace): Promise<void> {
+  async onTraceLoad(ctx: TraceImpl): Promise<void> {
     ctx.commands.registerCommand({
       id: 'dev.perfetto.ToggleDrawer',
       name: 'Toggle drawer',
       defaultHotkey: TrackUtilsPlugin.azertySetting.get() ? 'A' : 'Q',
       callback: () => {
-        const tabsManager = ctx.tabs as unknown as {
-          toggleTabPanelVisibility: () => void;
-        };
-        tabsManager.toggleTabPanelVisibility();
+        ctx.tabs.toggleTabPanelVisibility();
       },
     });
 


ui/src/core_plugins/dev.perfetto.CoreCommands/index.ts
```diff
@@ -746,10 +746,13 @@ export default class CoreCommands implements PerfettoPlugin {
       defaultHotkey: 'R',
     });
 
+    const azertySetting = ctx.settings.get<boolean>('azertyMode');
```
stevegolton: This introduces a hidden dependency on the TrackUtils plugin. If it's working right now it's only working by chance.

You could make it an explicit dependency, or ...

The separation between these two plugins is somewhat arbitrary, they both provide useful 'core' commands. But you could think of the TrackUtilsPlugin as more of a 'timeline' plugin, in which case it makes sense to move the `dev.perfetto.ToggleDrawer` command there instead (seeing as the drawer is part of the timeline). Then you can locate everything in the one plugin.

BryanBradfo: Thanks for the review!
Indeed, the dependency between plugins was fragile.
I have refactored the code as suggested!

ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
```diff
@@ -40,9 +41,30 @@ export default class TrackUtilsPlugin implements PerfettoPlugin {
...
+      id: 'azertyMode',
+      defaultValue: false,
+      name: 'AZERTY layout',
+      description: 'Adjusts shortcuts for AZERTY keyboards (ZQSD navigation).',
+      schema: z.boolean(),
+      requiresReload: true,
+    });
   }
 
   async onTraceLoad(ctx: Trace): Promise<void> {
```
stevegolton: Turn this `Trace` into a `TraceImpl` and you should be able to avoid the type assertions below.

BryanBradfo: Done! Changed onTraceLoad to accept TraceImpl instead of Trace, which gives direct access to TabManagerImpl.toggleTabPanelVisibility() without the type assertions. Thanks for the suggestion!

aMayzner (approved):

BryanBradfo:
> Looks great! Thanks for the contributation.

Thanks for the review, Steve! I'm really excited to have my first contribution to Perfetto (and to a Google repo!) merged. I appreciate the guidance on the implementation.

PR #4617, opened 2026-01-29
Increase ui.perfetto.dev build timeout to 60 min

primiano:
Also update the comment. Since we moved to GH there is
no need anymore to copy the yaml into the trigger setting.

Bug: b/479830079


primiano:
@LalitMaganti @stevegolton there is something else I noticed as part of this

I think we messed up something when we migrated to GH, as the perfetto-ui-trigger-release is NOT referring to cloudbuild_release.yaml, but all the triggers refer to cloudbuild.yaml

So right now cloudbuild_release.yaml seems unused

This seems an accidental mistake... however things seem to work ???

we need to understand what's going on before touching too much.
For now i'm only increasing timeouts

PR #4630, opened 2026-01-30
ui: Add WebGL renderer

stevegolton:
Add WebGL renderer to accelerate rendering of large rects / chevrons efficiently.

## Design
- An additional canvas is positioned behind the usual canvas and a new `Renderer` abstraction has been created which tracks can use to render rects or markers (e.g. chevrons) using WebGL in the same imperative and immediate fashion.
- WebGL rendering is currently behind a flag, as it's still experimental and has some limitations (see below).
- The track implementation has changed very little, mainly swapping `ctx.fillRect()` for `renderer.drawRect()`.
- When WebGL is not available (e.g. on VMs) we fall back to a canvas2d based renderer which performs similarly to before.
- Chevrons are implemented as SDF textures, for resolution independent rendering.
- Rects and chevrons are drawn using instanced rendering to minimize buffer sizes.
- Crosshatching and fadeout patterns implemented in fragment shaders for RT and incomplete slices.

## Limitations
- Unable to customize marker styles (all markers appear as chevrons).
- Slices appear behind the gridlines as opposed to on top (seeing as all webgl stuff is rendered to a different canvas element which is positioned behind the usual canvas.
- Text is not supported yet so this is still done using canvas2d calls.
- Incomplete slices not fully supported yet by the renderer so they are still done using canvas2d calls.
- In its current implementation, the webgl renderer involves streaming buffers every frame. This is not a particularly efficient way of using WebGL, in theory each track could load buffers into WebGL only when they actually change (e.g. when new data is pulled from TP) and leave them in GPU memory, meaning draws would be a uniform update and a draw call. However this would involve creating a potentially unbound number of buffers which is a problem (several per track). Uploading the buffers every frame is still extremely performant compared to canvas calls that it doesn't really matter, and we can make further improvements in the future if needed. The main overhead comes from the work done in JS to fill the buffers out in the first place, which definitely can be further improved.

## Testing
Average canvas render time:
Old Canvas2D: **21ms**
New Canvas2D: **21ms** (identical)
WebGL: **5ms**

## Test setup
- Macbook Pro 16 M1
- Android example trace
- Expand ftrace group, collpase cpu freq group, dismiss details panel - looks like this:
image

### DIFF
--- a/ui/src/base/renderer.ts
+++ b/ui/src/base/renderer.ts
@@ -72,6 +72,11 @@ export interface Renderer {
     pattern?: number,
   ): void;
 
-  // Flush all pending draw calls to the GPU.
+  // Flush all pending draw/marker calls to the underlying backend and
+  // invalidate caches. Users should ensure that they call flush before
+  // accessing the canvas2d context directly in order to synchronize draws and
+  // avoid visual glitches. However, excessive flushing can degrade performance,
+  // so it should be used judiciously. If possible, try to batch as many draw
+  // calls together as possible inbetween flushes.
   flush(): void;
 }


ui/src/components/tracks/base_slice_track.ts
```diff
@@ -919,29 +916,8 @@ export abstract class BaseSliceTrack<
...
-    end = Time.add(end, this.slicesKey.bucketSize);
-
-    let slices = filterVisibleSlices<CastInternal<SliceT>>(
-      this.slices,
-      start,
-      end,
-    );
-    slices = slices.concat(this.incomplete);
+  private getVisibleSlicesInternal(): Array<CastInternal<SliceT>> {
+    const slices = this.slices.concat(this.incomplete);
```
LalitMaganti, line 920: One thing I found in my profiling of this was that this array causes object churn and GC stalls due to creating a anew array every render and it's very easy to avoid by just structuring it as callbacks operating on each of these.

Not sure you want to do in this CL, maybe in a followup?

ui/src/components/tracks/base_slice_track.ts
```diff
@@ -547,23 +535,28 @@ export abstract class BaseSliceTrack<
...
+          () => this.drawChevron(ctx, slice.x, y, sliceHeight),
+        );
       } else if (slice.flags & SLICE_FLAGS_INCOMPLETE) {
+        // Draw incomplete slice
         const w = CROP_INCOMPLETE_SLICE_FLAG.get()
           ? slice.w
           : Math.max(slice.w - 2, 2);
+        // Flush renderer before accessing the canvas2d context directly to
+        // synchronize any reordering and invalidate caches.
+        renderer.flush();
```
LalitMaganti, line 559: Is it really OK to flush on every incomplete slice like this?

ui/src/plugins/dev.perfetto.Sched/cpu_slice_track.ts
```diff
@@ -384,9 +381,12 @@ export class CpuSliceTrack implements TrackRenderer {
         color = colorScheme.base;
         textColor = colorScheme.textBase;
       }
-      ctx.fillStyle = color.cssString;
 
-      if (data.flags[i] & CPU_SLICE_FLAGS_INCOMPLETE) {
+      if (flags & CPU_SLICE_FLAGS_INCOMPLETE) {
+        // Flush renderer before accessing the canvas2d context directly to
+        // synchronize any reordering and invalidate caches.
+        renderer.flush();
```
LalitMaganti, line 388: As above: in general it would be nice if there was some commentary of how expensive "flush" is meant to be. Because it seems like from your usage, it should be cheap while the naming makes me think it might be very expensive. And I can imagine that might trick us into accidentally adding future expensive things to flush for that reason.

stevegolton: To this comment and above:

- This API is supposed to be a temporary shim in order to avoid having to change the track impls too much, it'll go away soon in favor of a batched API, but that requires a larger refactor in the tracks.
- In canvas mode, flush() is effectively a no-op and just trashes the cached fillstyle. If you called it every slice it would just make things a little slower as we would have to set the fillStyle every rect. Note: we need to do this anyway if we set a custom fillstyle while rendering the incomplete slice.
- In webgl mode it pushes the buffers to webgl and issues a draw call. Doing this on every slice tanks perf, but:
  - Incomplete slices are infrequent anyway.
  - Webgl is experimental and behind a default disabled flag.

I'll add a note to flush() to indicate the above.

ui/src/components/tracks/base_slice_track.ts
stevegolton: As below

ui/src/components/tracks/base_slice_track.ts
stevegolton: Good catch, I'll take a look in a followup. Though I suspect this will all change anyway in some upcoming PRs.

LalitMaganti (approved):

PR #4633, opened 2026-01-30
ui: Reduce jank while tracks are processing query results

stevegolton:
Schedule track query results processing on a task with (for now) 'user-visible' priority, yielding every now and again to avoid blocking the main thread for too long, which reduces frame jank.

Note: 'background' priority would be a good option as it'll never block rendering, but for now our rafs are too busy which starves background tasks leading to long loading times.

This makes use of the scheduling API, with a fallback on setTimeout for browsers that don't support it (Safari, mainly). See: https://developer.chrome.com/blog/use-scheduler-yield

Testing: Scroll around on a busy timeline, observe less jank.

### DIFF
--- a/ui/src/components/tracks/base_slice_track.ts
+++ b/ui/src/components/tracks/base_slice_track.ts
@@ -790,7 +790,9 @@ export abstract class BaseSliceTrack<
     this.onUpdatedSlices(slices);
     this.slices = slices;
 
-    raf.scheduleCanvasRedraw();
+    // We need to schedule a full redraw here as the track height could have
+    // changed, which requires a full DOM redraw.
+    raf.scheduleFullRedraw();
   }
 
   private rowToSliceInternal(row: RowT): CastInternal<SliceT> {


ui/src/base/chunked_task.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Polyfill for scheduler.postTask()
+export function postTask(
+  callback: (args: void) => void,
+  options?: {priority?: Priority},
+): void {
+  if (globalThis.scheduler?.postTask) {
+    globalThis.scheduler.postTask(callback, options);
+  } else {
```
LalitMaganti, line 37: One concern I have: on Safari I believe this will be posted to the *back* of the queue causing significant loading bars. Have we tested this and do we think it's a problem?

ui/src/components/tracks/base_counter_track.ts
```diff
@@ -1130,6 +1131,8 @@ export abstract class BaseCounterTrack implements TrackRenderer {
       );
     `);
 
+    const task = await deferChunkedTask({priority: 'background'});
```
LalitMaganti: When I tested this, on Chrome, setting background would cause significant "Loading" bars to appear. Maybe it's less of a problem after other changes but I'd take a largeish trace and test zooming/panning around etc and see the effect.

stevegolton: I was on the fence about whether to use 'background' or 'user-visible' for this. 'background' led to longer loading bars but with less frame jank, whereas 'user-visible' seemed to delay RAFs by up to a frame or two in order to get the work done introducing slightly more jank, but the loading times were noticeably shorter.

'user-visible' semantically seems like it should be the correct option, and overall it is probably a good balance - I'll change it

ui/src/base/chunked_task.ts
stevegolton: On chrome it behaves like 'user-visible'. The overall loading time is no different.

Yield continuation prioritization should mean that tracks that started loading earlier will run to completion before starting to process new tracks, whereas with setTimeout they will tend to interleave. In theory this means that we should start seeing the first tracks completing sooner with yield but the overall loading time should be the same - but in practice I don't see much difference TBH.

I haven't tested natively on Safari, yet...

ui/src/components/tracks/base_counter_track.ts
stevegolton: Looking into this more, I think background is probably technically what we want if aiming for 0 jank.

Rendering does not take priority over enqueued 'user-visible' tasks. This means if we have N track's worth of continuations queued up, the RAF would be blocked for up to N*4ms.  Assuming a worst case of ~30 busy tracks on screen that's up to 120ms of jank.

For example, you can see all the task continuations blocking the raf which was clearly scheduled and triggered way earlier:
image

Rendering does however take priority over 'background' tasks, so the most the RAF is blocked by is 4ms. The problem is we're doing a lot in our RAFs so when they're constantly being triggered, the background tasks are getting starved. Something has to give, we either do some loading and jank, or do run at 60fps and never load.

The alternative is we keep track of how long it's been since the last RAF and limit the amount of work we can do before yielding to stay under 16ms... but that's a bit hacky.

General grumble: requestIdleCallback *should* have solved this, but in practice it can delay arbitrarily for a number seconds :(

To summarize: 'background' prio perfectly solves our problem IMO, assuming our problem is jank. But even with 'user-visible' we still get the advantage of yielding every 4ms which is much better than not yielding at all.

LalitMaganti: In my prototype I think I was round-robining between background and user-visible to try and workaround this and prevent this "storm" of every track taking priority over the RAF. Do we want to go in that direction or do you think it's too complex and it's fine as is?

ui/src/base/chunked_task.ts
LalitMaganti: Ah OK. In that case LGTM (I thought it was more similar to requestIdleCallback but maybe I was mistaken).

I would just check once on Safari but otherwise all good.

stevegolton: I don't see much difference in safari between autopush and this TBH. I think we are a bit smoother thanks to the chunking but total loading times seem very similar.

LalitMaganti: Ack thanks for checking!

ui/src/components/tracks/base_counter_track.ts
stevegolton: > In my prototype I think I was round-robining between background and user-visible to try and workaround this and prevent this "storm" of every track taking priority over the RAF.

Ahh I see what you were doing, switching priorities every other continuation. Though if we get blocked by continuous rafs when in 'background' mode then we're still blocking potentially for a quite a long time before we manage to get a 'free' 'user-visible' chunk in..?

I think I'll change the priority to a fixed 'user-visible' priority for now in this PR, same priority as before but at least we still get the cunking. We're still too busy in rafs for 'background' priority to be practical. Once we do less (offscreen rendering / webgl) then I think ultimately 'background' is the correct priority if we want zero jank. In practice I think we should play with a few techniques including your round robin approach, because it probably makes sense to drop a frame or two to see slices sooner.

As you mentioned before we could also probably reduce the amount of fields per slice we load from TP in most tracks which would speed things up anyway.

LalitMaganti: > Though if we get blocked by continuous rafs when in 'background' mode then we're still blocking potentially for a quite a long time before we manage to get a 'free' 'user-visible' chunk in..?

Right exactly it's by no means a silver bullet, just maybe tries to balance a bit more. 

> As you mentioned before we could also probably reduce the amount of fields per slice we load from TP in most tracks which would speed things up anyway.

Sounds good. Let's discuss Monday. 

For what it's worth I am also doing some changes on the TP side to the mipmap operator to reduce the time taken on busiest tracks from 10+ ms to <1ms.

stevegolton: > For what it's worth I am also doing some changes on the TP side to the mipmap operator to reduce the time taken on busiest tracks from 10+ ms to <1ms.

Very exciting! Though in this case the query time isn't the issue - it's all the deserialization on the JS end :(. But faster queries are always better than slower queries :)

LalitMaganti: Yes on the example Android trace, the query is already very fast but on v large traces (O(GB}} query time is also a problem. 

I think both would help improving performance: JS side for all traces, TP side for bigger traces.

PR #4638, opened 2026-01-31
ui: weight flamegraphs by perf event count

cphlipot1:
give user option to weight flamegraphs by perf event count, and weight them this way by default.

tested with `./tools/open_trace_in_ui test/data/perf_sample_multicallstacks.pftrace --dev-server`

Bug: #4527

image

### DIFF
--- a/ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
+++ b/ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
@@ -428,7 +428,6 @@ export default class LinuxPerfPlugin implements PerfettoPlugin {
   private createAreaSelectionTab(trace: Trace) {
     let previousSelection: AreaSelection | undefined;
     let flamegraphWithMetrics: QueryFlamegraphWithMetrics | undefined;
-    let isLoading = false;
 
     return {
       id: 'perf_sample_flamegraph',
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -501,6 +561,39 @@ export default class LinuxPerfPlugin implements PerfettoPlugin {
     });
     return {flamegraph: new QueryFlamegraph(trace), metrics};
   }
+
+  private async getCounterTypesForSessions(
```
LalitMaganti: Any chance we can do this in onTraceLoad and cache it here? Would be a nicer user experience if we didn't have to do this just-in-time.

cphlipot1:
@LalitMaganti I am seeing failures in UI related screenshot tests, presumably because i updated the contents of the flamegraph selector. I am unable to see screenshots from the CI as far as i can tell, and am unclear how to reproduce locally.

when i run `./ui/run-all-tests` locally i see all 77 tests pass where CI shows 75 passing and 2 failing.

I have merged this PR with main after https://github.com/google/perfetto/pull/4747 has closed, so i don't think i'm impacted by the issues that PR supposedly fixed.

LalitMaganti:
https://storage.googleapis.com/perfetto-ci-artifacts/gh-21851583282-1-ui/ui-test-artifacts/index.html is the screenshot tests

Seems like everything changed is WAI. You need access to the GCS bucket to update screenshots tho so let me download this patch and do it for you.

ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -431,16 +475,27 @@ export default class LinuxPerfPlugin implements PerfettoPlugin {
     };
   }
 
-  private computePerfSampleFlamegraph(
+  private async computePerfSampleFlamegraph(
     trace: Trace,
     currentSelection: AreaSelection,
-  ): QueryFlamegraphWithMetrics | undefined {
+  ): Promise<QueryFlamegraphWithMetrics | undefined> {
```
LalitMaganti: Why async at this point? We can remove that I think. That's the whole reason I proposed prefiltering :)

cphlipot1: agree. after last fix i don't think this is needed anymore.

LalitMaganti (approved):

PR #4641, opened 2026-02-02
ui: Add MultiSelect options validation

camillobruni:
Providing duplicate ids in the MultiSelectAttrs options can lead to very unexpected side effects and late UI corruption.
This adds a minimal validation step with an explicit warning to speed up debugging.

### DIFF
--- a/ui/src/widgets/multiselect.ts
+++ b/ui/src/widgets/multiselect.ts
@@ -255,7 +255,7 @@ export class PopupMultiSelect
     const {options, showNumSelected, label} = attrs;
 
     if (showNumSelected) {
-      const numSelected = options.filter(({checked}) => checked).length;;
+      const numSelected = options.filter(({checked}) => checked).length;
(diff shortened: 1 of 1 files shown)


ui/src/widgets/multiselect.ts
```diff
@@ -244,7 +255,10 @@ export class PopupMultiSelect
     const {options, showNumSelected, label} = attrs;
 
     if (showNumSelected) {
-      const numSelected = options.filter(({checked}) => checked).length;
+      const numSelected = options.reduce(
```
stevegolton: This change seems unrelated

stevegolton (changes requested):
Looks good to me, thanks. Just a nit about the unrelated change

ui/src/widgets/multiselect.ts
camillobruni: removed.

PR #4653, opened 2026-02-03
Add OOM Adjuster Score visualization plugin

gignat-dev:
This plugin introduces a new Perfetto UI plugin to visualize Android's OOM Adjuster scores. It provides:
- A command "OOM Adjuster Score: Visualize (over selection)" to generate tracks.
- Concurrency tracks showing the overlap count of OOM intervals per bucket.
- Individual OOM Score tracks for each process within a selected time window and bucket.

The plugin utilizes SQL queries against `android_oom_adj_intervals` to surface this data in the Perfetto UI, helping in the analysis of OOM related behaviors.

### DIFF
--- a/ui/src/plugins/com.android.OomAdjScoreViz/index.ts
+++ b/ui/src/plugins/com.android.OomAdjScoreViz/index.ts
@@ -22,12 +22,20 @@ import {
 import {uuidv4} from '../../base/uuid';
 import {getTimeSpanOfSelectionOrVisibleWindow} from '../../public/utils';
 import {time, TimeSpan} from '../../base/time';
+import {z} from 'zod';
 
 export default class OomAdjScoreViz implements PerfettoPlugin {
   static readonly id = 'com.android.OomAdjScoreViz';
 
   private static readonly TBL_INTERVALS = `_${OomAdjScoreViz.id.replace(/\./g, '_')}_oom_intervals`;
   private static readonly TBL_COUNT = `_${OomAdjScoreViz.id.replace(/\./g, '_')}_oom_count`;
+
+  private static readonly TimeSpanSchema = z
+    .object({
+      startTime: z.string().optional(),
+      endTime: z.string().optional(),
+    })
+    .default({});
 
   async onTraceLoad(ctx: Trace): Promise<void> {
     await ctx.engine.query(`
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.android.OomAdjScoreViz/index.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  private static readonly TBL_INTERVALS = `_${OomAdjScoreViz.id.replace(/\./g, '_')}_oom_intervals`;
+  private static readonly TBL_COUNT = `_${OomAdjScoreViz.id.replace(/\./g, '_')}_oom_count`;
+
+  async onTraceLoad(ctx: Trace): Promise<void> {
+    await ctx.engine.query(`
+      INCLUDE PERFETTO MODULE android.oom_adjuster;
+      INCLUDE PERFETTO MODULE intervals.overlap;
+
+      CREATE OR REPLACE PERFETTO TABLE ${OomAdjScoreViz.TBL_INTERVALS} AS
```
zezeozue, line 45: Do we need a separate table or just cte in the one below?

Also, can we chop at process death with something like:

create or replace perfetto table oom as
select 
    row_number() over () as id, 
    oom.ts,
    -- Clip the duration: use the smaller of (current end) or (process end)
    min(oom.ts + oom.dur, ifnull(p.end_ts, oom.ts + oom.dur)) - oom.ts as dur
...
from android_oom_adj_intervals oom
join process p using (upid)

zezeozue (approved):

ui/src/plugins/com.android.OomAdjScoreViz/index.ts
gignat-dev: > Do we need a separate table or just cte in the one below?

Yes, I think we need the separate table because we use it in two other queries

> Also, can we chop at process death with something like:

Right, I clipped the oom intervals by the process start_ts and end_ts

ui/src/plugins/com.android.OomAdjScoreViz/index.ts
```diff
@@ -0,0 +1,214 @@
...
+        group_name AS bucket,
+        value AS concurrency
+      FROM intervals_overlap_count_by_group!(${OomAdjScoreViz.TBL_INTERVALS}, ts, dur, bucket);
+    `);
+
+    ctx.commands.registerCommand({
+      id: `${OomAdjScoreViz.id}.visualize`,
+      name: 'OOM Adjuster Score: Visualize (over selection)',
+      callback: async (...args: unknown[]) => {
+        const params = args[0] as {[key: string]: unknown} | undefined;
```
stevegolton: There's a lot of type assertions going on here - I'd suggest zod might be able to clean this up but I'll leave that up to you

ui/src/plugins/com.android.OomAdjScoreViz/index.ts
```diff
@@ -0,0 +1,214 @@
...
+      SELECT
+        ts,
+        lead(ts) OVER(PARTITION BY group_name ORDER BY ts) - ts AS dur,
+        group_name AS bucket,
+        value AS concurrency
+      FROM intervals_overlap_count_by_group!(${OomAdjScoreViz.TBL_INTERVALS}, ts, dur, bucket);
+    `);
+
+    ctx.commands.registerCommand({
+      id: `${OomAdjScoreViz.id}.visualize`,
```
stevegolton: We typically avoid putting the plugin id directly in the command name so that they can be moved around. Call it something like `com.android.visualizeOomAdjScore` or similar.

gignat-dev: Done, renamed to com.android.visualizeOomAdjScore

ui/src/plugins/com.android.OomAdjScoreViz/index.ts
gignat-dev: Done, declared a zod schema for the command params

PR #4656, opened 2026-02-03
Change type of LayerProto#metadata to map<int32, bytes>

chominskib-goog:
Parsing the previous, map<int32, string>, version in Java was troublesome because Java proto parser expects strings to be in UTF-8 and the values of this metadata map are usually not correct UTF-8 strings.

### DIFF
--- a/protos/perfetto/trace/perfetto_trace.proto
+++ b/protos/perfetto/trace/perfetto_trace.proto
@@ -6755,7 +6755,7 @@ message LayerProto {
   // Layer's corner radius
   optional float corner_radius = 41 [deprecated = true];
   // Metadata map. May be empty.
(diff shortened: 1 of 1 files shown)


aMayzner (approved):

LalitMaganti:
@PabiGamito @rukkal if this is going to be also put in our logs proto, we're going to have to discuss the use of bytes. The general policy is that we don't allkw bytes fields (with rare exceptions) in logs protos.

chominskib-goog:
I amended the PR because the presubmits were failing due to `protos/perfetto/trace/perfetto_trace.proto` being out of sync with the original `proto` file I've changed. Could you please PTAL and start a new presubmit run?

PR #4661, opened 2026-02-03
tp: fix handling of perf.data records without tses in merged traces

LalitMaganti:
In merged traces, we would previously mark all MMAP/MMAP2/COMM records as having
timestamp 0 to force them to be processed before all samples. However, this is
incorrect as really the timestamp should be the timestamp of the start of the
file as that's really when these records were (appoximately) captured.

This was causing problems in the presence of merged traces which have
task_newtask events which forcefully cause a new process to be created: in that
case, the mmaps would be associated to one process and the actual slices/ftrace
events would be associated to another even though it was really the same process

Rationalise this by giving the records the timestamp of the *next* even which had
a timestamp which is as good as we are going to get. It's still not totally
accurate but is better than the current behaviour of 0 and fixes the issue
well enough in the vast majority of cases.

Bug: b/481290800

### DIFF
--- a/ui/src/plugins/org.chromium.ChromeScrollJank/index.ts
+++ b/ui/src/plugins/org.chromium.ChromeScrollJank/index.ts
@@ -33,6 +33,8 @@ import {SourceDataset} from '../../trace_processor/dataset';
 import {SliceTrack} from '../../components/tracks/slice_track';
 import {escapeQuery} from '../../trace_processor/query_utils';
 import {ThreadSliceDetailsPanel} from '../../components/details/thread_slice_details_tab';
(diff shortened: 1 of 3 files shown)


src/trace_processor/importers/perf/perf_data_tokenizer.cc
sashwinbalaji, line 580: should we flush pending_records_without_timestamp_ here also if min_timestamp_ else log error ?

LalitMaganti: Unfortunately we cannot do it in notifeof because mmapp records are needed *before* the samples: if they happen to be at the same timestamp, we stably sort so it will be the wrong way round.

PR #4663, opened 2026-02-03
traceconv: add support for Android traces to bundle command

LalitMaganti:
The bundle command until now only supported Linux traces. Let's also add
support for Android traces. This also needs us to support deobufscation
as well, not just symbolization.

### DIFF
--- a/include/perfetto/base/compiler.h
+++ b/include/perfetto/base/compiler.h
@@ -99,13 +99,9 @@ extern "C" void __asan_unpoison_memory_region(void const volatile*, size_t);
 #define PERFETTO_ASAN_UNPOISON(addr, size)
 #endif  // __clang__
 
-#if defined(__clang__)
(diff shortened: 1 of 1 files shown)


LalitMaganti:
Holding off on this. I want to change the approach.

LalitMaganti:
OK now should be a lot cleaner. Though followups will improve things further.

include/perfetto/base/compiler.h
```diff
@@ -99,6 +99,17 @@ extern "C" void __asan_unpoison_memory_region(void const volatile*, size_t);
 #define PERFETTO_ASAN_UNPOISON(addr, size)
 #endif  // __clang__
 
+#if defined(__clang__)
+#if __has_feature(memory_sanitizer)
```
sashwinbalaji: nit: combine the  two ifs ?

PR #4665, opened 2026-02-03
tp: Wattson Metric v2 implementation

runpinglai-creator:
Implemented Wattson metric v2. Refactored and moved essential metric calculation logics to stdlib, exposing metric-specific logics in trace summary. Wattson metric v1 protos are heavily nested. v2 flattened the metric proto structure and put most essential data into flat views for metricv2 consumption.

windows.sql defines the analysis windows (e.g., Marker, Startup, CUJ), which are passed as parameters to the
views in rails_metrics.sql and threads_metrics.sql; these views then invoke centralized aggregation macros in aggregation.sql to calculate the final power attributions flat table for textproto to consume.

bug: 477599147


src/trace_processor/perfetto_sql/stdlib/wattson/textprotos/rails.textproto
```diff
@@ -0,0 +1,343 @@
+#
```
LalitMaganti: The whole point of metrics v2 is that these textprotos defining metrics no longer live in our repo but in whichever location makes most sense for the metrics system you're integrating with. Please remove them from here.

src/trace_processor/perfetto_sql/stdlib/wattson/aggregation.sql
```diff
@@ -0,0 +1,433 @@
+--
```
LalitMaganti, line 1: In general, putting all of these in a "metrics" folder doesn't really make sense IMO. Fundamentally metrics are simply a subset of data in the trace. There's nothing special about them which warrants their own folder. Especially in your case as you're simply aggregating data across the trace.

src/trace_processor/perfetto_sql/stdlib/wattson/textprotos/rails.textproto
```diff
@@ -0,0 +1,343 @@
...
+# limitations under the License.
+
+metric_spec {
+  id: "wattson_metadata"
+  dimensions: "metric_version"
+  dimensions: "power_model_version"
+  value: "is_crude_estimate"
+  query: {
+    table: {
+      table_name: "_wattson_metric_metadata"
```
LalitMaganti: Since these metric files will no longer live in our repo, all tables which are referenced by these should be made public and should be properly documented and become part of the public API of the wattson package.

src/trace_processor/perfetto_sql/stdlib/wattson/aggregation.sql
runpinglai-creator: Agree. Merged the contents from "metrics" folder to other reusable places.

src/trace_processor/perfetto_sql/stdlib/wattson/textprotos/rails.textproto
runpinglai-creator: Agree. The textprotos should be kept at metric owners' place and metric consumers' place.

src/trace_processor/perfetto_sql/stdlib/wattson/textprotos/rails.textproto
runpinglai-creator: done

PR #4666, opened 2026-02-03
tp: split NotifyEndOfFile function into two stages

LalitMaganti:
This CL splits the NotifyEndOfFile function which is plumbed everywhere
in trace processor into two stages:
1) Stage 1 is responsbile for pushing data to the sorter
2) Stage 2 is responsible for making any final adjustments to tables

This largely corresponds to the two types of data processing which
happens in NotifyEndOfFile.

The main motivation for this change is that it's currently *impossible*
to have a merged trace file which waits until end of all processing to
push data because any data pushed prior will be extracted on
TraceProcessorStorage::Flush() leading to sorter_out_of_order stats and
data just being dropped.

After this change, we instead correctly push all data to the queues, run
a one-shot extraction and only then add any remaining data.

Bug: https://github.com/google/perfetto/issues/1000

### DIFF
--- a/src/trace_processor/importers/common/chunked_trace_reader.h
+++ b/src/trace_processor/importers/common/chunked_trace_reader.h
@@ -41,6 +41,8 @@ class ChunkedTraceReader {
   // For other trace formats, this may write directly to storage tables.
   [[nodiscard]] virtual base::Status OnPushDataToSorter() = 0;
 
+  // Phase 2 (implicit): TraceSorter extracts and sorts all pushed events.
+
   // Phase 3: Called after events are extracted from sorter.
   // Parsers do post-extraction processing and cleanup here.
   virtual void OnEventsFullyExtracted() = 0;
--- a/src/trace_processor/importers/perf/perf_data_tokenizer.cc
+++ b/src/trace_processor/importers/perf/perf_data_tokenizer.cc
@@ -572,11 +572,7 @@ base::Status PerfDataTokenizer::OnPushDataToSorter() {
 
 void PerfDataTokenizer::OnEventsFullyExtracted() {
   // Phase 3: Finalize tracker
-  auto status = perf_tracker_.OnEventsFullyExtracted();
-  if (!status.ok()) {
-    PERFETTO_ELOG("Failed to finalize perf tracker: %s",
-                  status.message().c_str());
-  }
+  perf_tracker_.OnEventsFullyExtracted();
 }
 
 }  // namespace perfetto::trace_processor::perf_importer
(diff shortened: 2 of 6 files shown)


src/trace_processor/importers/pprof/pprof_trace_reader.cc
```diff
@@ -73,12 +73,19 @@ base::Status PprofTraceReader::Parse(TraceBlobView blob) {
   return base::OkStatus();
 }
 
-base::Status PprofTraceReader::NotifyEndOfFile() {
+base::Status PprofTraceReader::OnPushDataToSorter() {
+  // Idempotency: if buffer is empty, we've already pushed
   if (buffer_.empty()) {
-    return base::ErrStatus("Empty pprof data");
+    return base::OkStatus();
```
sashwinbalaji, line 84: shall we maintain any other variabl like "first clear" to check if it's actuall empty and throw error like before.

src/trace_processor/importers/common/chunked_trace_reader.h
```diff
@@ -36,8 +36,14 @@ class ChunkedTraceReader {
   // The buffer size is guaranteed to be > 0.
   virtual base::Status Parse(TraceBlobView) = 0;
 
-  // Called after the last Parse() call.
-  [[nodiscard]] virtual base::Status NotifyEndOfFile() = 0;
+  // Phase 1: Push accumulated data to sorter or storage.
+  // For proto traces, this pushes deferred packets to the sorter.
+  // For other trace formats, this may write directly to storage tables.
+  [[nodiscard]] virtual base::Status OnPushDataToSorter() = 0;
+
```
sashwinbalaji, line 43: can we please mention here that phase 2 is the sorting done by trace sorter

src/trace_processor/importers/perf/perf_data_tokenizer.cc
```diff
@@ -562,12 +562,21 @@ base::Status PerfDataTokenizer::ProcessItraceStartRecord(Record record) {
...
   }
-  RETURN_IF_ERROR(perf_tracker_.NotifyEndOfFile());
   return base::OkStatus();
 }
 
+void PerfDataTokenizer::OnEventsFullyExtracted() {
+  // Phase 3: Finalize tracker
+  auto status = perf_tracker_.OnEventsFullyExtracted();
+  if (!status.ok()) {
+    PERFETTO_ELOG("Failed to finalize perf tracker: %s",
```
sashwinbalaji: should we have stats for them ?

sashwinbalaji (approved):

PR #4672, opened 2026-02-04
tp: add support for breakpad symbolization in traceconv bundle

LalitMaganti:
Also unify the codepaths for symbolization to make everyone use the new function and simplify the enrichment logic.


src/trace_processor/util/symbolizer/symbolize_database.cc
```diff
@@ -171,21 +162,44 @@ std::string SymbolizeDatabaseWithSymbolizer(trace_processor::TraceProcessor* tp,
...
+      result.symbols += SymbolizeDatabaseWithSymbolizer(tp, symbolizer.get());
+    }
+  }
+
+  // Run "find" mode symbolizer if paths are provided.
+  if (has_find) {
+    auto symbolizer = CreateFindSymbolizer(config);
+    if (symbolizer) {
+      result.symbols += SymbolizeDatabaseWithSymbolizer(tp, symbolizer.get());
+    }
```
sashwinbalaji: nit: here and above, use if statement with initializer ?

PR #4673, opened 2026-02-04
Add disp_frame_start/done_timeout ftrace

creek0810:
Bug: 456638739
Test: can use gfx to record, show event on ui correctly


src/trace_processor/importers/ftrace/pixel_display_tracker.cc
```diff
@@ -0,0 +1,103 @@
...
+
+namespace perfetto::trace_processor {
+
+PixelDisplayTracker::PixelDisplayTracker(TraceProcessorContext* context)
+    : context_(context),
+      frame_start_timeout_name_(
+          context->storage->InternString("frame_start_timeout")) {}
+
+void PixelDisplayTracker::ParseDpuDispFrameStartTimeout(
+    int64_t timestamp,
```
LalitMaganti, line 46: If these are just being stored as instants, what do we need any parsing at all? Why is it not just using ftrace events tab? Also should we not make slices?

src/trace_processor/importers/ftrace/pixel_display_tracker.cc
```diff
@@ -0,0 +1,103 @@
...
+
+  TrackId track_id = context_->track_tracker->InternTrack(
+      kBluePrint, tracks::Dimensions(ex.display_id()));
+  StringId slice_name_id = frame_start_timeout_name_;
+
+  context_->slice_tracker->Scoped(
+      timestamp, track_id, kNullStringId, slice_name_id, 0,
+      [&](ArgsTracker::BoundInserter* inserter) {
+        inserter->AddArg(
+            context_->storage->InternString(base::StringView("output_id")),
```
LalitMaganti: All constant strings should be interned in the constructor.

src/trace_processor/importers/ftrace/pixel_display_tracker.cc
creek0810: Yes. They are just being stored as instants. 
 
>  Why is it not just using ftrace events tab

There are two reasons here. 
1. There are too many events in ftrace tab. (I assume you are talking about the "Ftrace Events" track in perfetto ui) We would like to have a place to find our display events only and easily. That's why I put the tracks under Hardware > Display.
2. We plan to add those two events in AOT. From my understanding, I need to upstream the proto defintion here and parse it.

Please let me know if you have any better idea to achieve it.

LalitMaganti:
@sashwinbalaji @aMayzner can you take care of this? Going to be out unexpectedly next couple of weeks

src/trace_processor/importers/ftrace/pixel_display_tracker.cc
sashwinbalaji: Apologies for the delay.  

> We plan to add those two events in AOT. From my understanding, I need to upstream the proto defintion here and parse it.

You only need to upstream the protos, parsing them is not necessarily needed.

But considering you specifically need to display it in the Hardware > Display, I think the change is fine.

PR #4674, opened 2026-02-04
tp: keep track of mutation count and update trace bounds when needed

LalitMaganti:
Reduces the need for callers of Flush to perform gymnastics to avoid
calling Flush + NotifyEndOfFile in sequence.


src/trace_processor/trace_processor_impl.cc
sashwinbalaji: nit: align all the comments ?

src/trace_processor/trace_processor_impl.cc
```diff
@@ -423,53 +424,67 @@ sql_modules::NameToPackage GetStdlibPackages() {
...
+  return std::apply([](const auto&... t) { return (t.mutations() + ...); },
+                    tables);
+}
+
+template <typename T>
+std::pair<int64_t, int64_t> GetTraceTimestampBoundsNs(const T& tables) {
   int64_t start_ns = std::numeric_limits<int64_t>::max();
   int64_t end_ns = 0;
-  for (auto it = storage.ftrace_event_table().IterateRows(); it; ++it) {
+  for (auto it = std::get<0>(tables).IterateRows(); it; ++it) {
```
sashwinbalaji: not a big fan of using indexes for the tables seems really brittle.

How about:

1) struct with const& members or
2) make it generic using templates

```
#include <type_traits>

// 1. Detection helpers using variable templates (C++17 style)
template <typename T, typename = void>
inline constexpr bool has_dur_v = false;

template <typename T>
inline constexpr bool has_dur_v<T, std::void_t<decltype(std::declval<T>().dur())>> = true;

template <typename T, typename = void>
inline constexpr bool has_graph_ts_v = false;

template <typename T>
inline constexpr bool has_graph_ts_v<T, std::void_t<decltype(std::declval<T>().graph_sample_ts())>> = true;

// 2. Generic helper for a single table
template <typename T>
std::pair<int64_t, int64_t> GetTableBounds(const T& table) {
  int64_t start = std::numeric_limits<int64_t>::max();
  int64_t end = 0;

  for (auto it = table.IterateRows(); it; ++it) {
    int64_t ts;
    // Variable template detection makes this very readable
    if constexpr (has_graph_ts_v<decltype(it)>) {
      ts = it.graph_sample_ts();
    } else {
      ts = it.ts();
    }

    int64_t dur = 0;
    if constexpr (has_dur_v<decltype(it)>) {
      dur = it.dur();
    }

    start = std::min(ts, start);
    end = std::max(ts + dur, end);
  }
  return {start, end};
}

// 3. Non-brittle main function using std::apply and fold expressions
template <typename T>
std::pair<int64_t, int64_t> GetTraceTimestampBoundsNs(const T& tables) {
  int64_t global_start = std::numeric_limits<int64_t>::max();
  int64_t global_end = 0;

  std::apply([&](const auto&... t) {
    auto update = [&](std::pair<int64_t, int64_t> b) {
      global_start = std::min(global_start, b.first);
      global_end = std::max(global_end, b.second);
    };
    // C++17 fold expression iterates over the tuple members
    (update(GetTableBounds(t)), ...); 
  }, tables);

  if (global_start == std::numeric_limits<int64_t>::max()) {
    return {0, 0};
  }
  return {global_start, std::max(global_end, global_start + 1)};
}
```

LalitMaganti: This is too much complexity. I'd rather just be very explicit. Will change to it.

src/trace_processor/trace_processor_impl.cc
LalitMaganti: This is autoformatted.

sashwinbalaji (approved):

PR #4675, opened 2026-02-04
tp: log in the case of empty build ids

LalitMaganti:


src/trace_processor/util/symbolizer/symbolize_database.cc
```diff
@@ -97,6 +111,28 @@ std::map<UnsymbolizedMapping, std::vector<uint64_t>> GetUnsymbolizedFrames(
...
+};
+
+MappingsWithoutBuildId GetMappingsWithoutBuildId(
+    trace_processor::TraceProcessor* tp) {
+  MappingsWithoutBuildId result;
+  Iterator it = tp->ExecuteQuery(kQueryMappingsWithoutBuildId);
+  while (it.Next()) {
+    result.names.emplace_back(it.Get(0).AsString());
+    int64_t count = it.Get(1).AsLong();
+    PERFETTO_CHECK(count >= 0);
```
sashwinbalaji, line 121: why is this check needed ?

src/trace_processor/util/symbolizer/symbolize_database.cc
```diff
@@ -97,6 +111,28 @@ std::map<UnsymbolizedMapping, std::vector<uint64_t>> GetUnsymbolizedFrames(
   return res;
 }
 
+struct MappingsWithoutBuildId {
+  std::vector<std::string> names;
+  uint32_t frame_count = 0;
+};
```
sashwinbalaji: should we make this vector of {name, count} so when printing we can know for which one how many were missing rather than total ?

src/trace_processor/util/symbolizer/symbolize_database.cc
LalitMaganti: It's just for safety. We by the structure of the query we should never get a count which is <=0 because the count should not have returned it.

PR #4682, opened 2026-02-04
tp: de: Serializer responsiblity moved to Trace Processor

aMayzner:
This change moves query materialization responsibility from the UI to Trace Processor (TP), establishing TP as the single source of truth for all materialization state.

## Changes

**Public API (include/perfetto/trace_processor/):**
- Added new experimental `Summarizer` interface for managing lazy materialization of structured queries
- Added `CreateSummarizer()` method to `TraceProcessor` (returns ownership via out parameter)
- Added `SummarizerUpdateSpecResult` with nested `QuerySyncInfo` for sync status
- Added `SummarizerQueryResult` for query results
- All Summarizer APIs marked as EXPERIMENTAL

**Trace Processor (C++):**
- Added `SummarizerImpl` class with smart optimizations:
  - Lazy materialization: tables created only when `Query()` is called
  - Proto-based change detection via hashing
  - Dependency propagation: if query A changes, all dependents (B→C→D) re-materialize
  - Table-source substitution: unchanged queries reference materialized tables directly
- RPC layer manages Summarizer IDs (C++ TraceProcessor layer is ID-agnostic)
- Summarizers automatically cleared when trace processor is reset

**UI (TypeScript):**
- Simplified `QueryExecutionService` by removing ~500 lines of complex state management
- Removed UI-side query hash caching, materialization lifecycle, staleness detection
- TP is now queried on-demand for table names and state
- Updated `Engine` interface with new `createSummarizer()`, `updateSummarizerSpec()`, `querySummarizer()`, and `destroySummarizer()` methods


protos/perfetto/trace_processor/trace_processor.proto
```diff
@@ -142,10 +144,12 @@ message TraceProcessorRpc {
     ResetTraceProcessorArgs reset_trace_processor_args = 107;
     // For TPM_REGISTER_SQL_PACKAGE.
     RegisterSqlPackageArgs register_sql_package_args = 108;
-    // For TPM_ANALYZE_STRUCTURED_QUERY.
-    AnalyzeStructuredQueryArgs analyze_structured_query_args = 109;
```
LalitMaganti, line 146: Here and below: please reserve removed fields.

protos/perfetto/trace_processor/trace_processor.proto
```diff
@@ -424,3 +411,39 @@ message TraceSummaryResult {
   }
   optional string error = 3;
 }
+
+// Input for TPM_SYNC_DATA_EXPLORER.
```
LalitMaganti: Uhhhh there should be no refernece to "data explorer" in trace processor. Please remove all references.

include/perfetto/trace_processor/trace_processor.h
```diff
@@ -242,15 +242,17 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessor : public TraceProcessorStorage {
...
-  // metadata. The `spec` contains a TraceSummarySpec with all queries (in its
-  // `query` field). Each query must have an `id` field set. The `query_id`
-  // specifies which query from the spec should be analyzed (must match one of
-  // the queries' id fields).
-  virtual base::Status AnalyzeStructuredQuery(
-      const TraceSummarySpecBytes& spec,
+  // Syncs Data Explorer query state. Compares proto hashes to detect changes,
+  // auto-drops removed queries, marks changed queries for re-materialization.
+  // Materialization is lazy (deferred to FetchNodeForDataExplorer).
+  virtual base::Status SyncDataExplorer(const TraceSummarySpecBytes& spec,
```
LalitMaganti: Please remove references to DataExplorer. As we discussed this is just "create summarizer" which should really return a class which you can call "Query" on.

aMayzner:
> I think the SQL data used to produce hermetic SQL that we could just run without depending on any intermediate tables but this seems to no longer be the case

Fixed

include/perfetto/trace_processor/trace_processor.h
```diff
@@ -237,20 +237,6 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessor : public TraceProcessorStorage {
   // loaded by trace processor shell at runtime. The message is encoded as
   // DescriptorSet, defined in perfetto/trace_processor/trace_processor.proto.
   virtual std::vector<uint8_t> GetMetricDescriptors() = 0;
-
-  // =================================================================
```
LalitMaganti, line 241: I expect to see a CreateSummaizer method here + a Summarizer interface? It seems to be missing.

The Rpc layer should not have dependencies on anything except the public API of trace processor, that's a layering violation.

src/trace_processor/rpc/BUILD.gn
```diff
@@ -40,6 +40,9 @@ source_set("rpc") {
     "../../base:version",
     "../../protozero",
     "../../protozero:proto_ring_buffer",
+    "../trace_summary",
```
LalitMaganti: This is incorrect. The Rpc layer should only depend on the public API of trace processor not internal details like this.

include/perfetto/trace_processor/summarizer.h
```diff
@@ -0,0 +1,87 @@
+/*
```
LalitMaganti, line 1: Please mark all contents of this file as experimental until we nail down the shape of this interface

include/perfetto/trace_processor/trace_processor.h
```diff
@@ -239,18 +240,26 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessor : public TraceProcessorStorage {
...
-  // Analyzes a structured query and returns the generated SQL along with
-  // metadata. The `spec` contains a TraceSummarySpec with all queries (in its
-  // `query` field). Each query must have an `id` field set. The `query_id`
-  // specifies which query from the spec should be analyzed (must match one of
-  // the queries' id fields).
-  virtual base::Status AnalyzeStructuredQuery(
-      const TraceSummarySpecBytes& spec,
-      const std::string& query_id,
-      AnalyzedStructuredQuery* output) = 0;
+  // Creates a new Summarizer instance for managing lazy materialization of
```
LalitMaganti: This should all still be marked as experimental.

include/perfetto/trace_processor/trace_processor.h
```diff
@@ -239,18 +240,26 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessor : public TraceProcessorStorage {
...
+  // structured queries. The returned Summarizer is owned by this TraceProcessor
+  // instance and will be cleaned up when destroyed or when DestroySummarizer()
+  // is called.
+  //
+  // |summarizer_id| is a caller-provided unique identifier for this summarizer.
+  // Returns an error if a summarizer with the same ID already exists.
+  virtual base::Status CreateSummarizer(const std::string& summarizer_id) = 0;
```
LalitMaganti: This is incrorect; we should return base::StatusOr<std::unique_ptr<Summarizer>> which it's up to the caller to decide how to handle.

src/trace_processor/rpc/rpc.h
```diff
@@ -26,6 +26,7 @@
 #include <vector>
 
 #include "perfetto/base/status.h"
+#include "perfetto/ext/base/flat_hash_map.h"
```
LalitMaganti, line 29: The RPC layer will still need to keep a hashmap of summarizers. It's responsbile of ids. The c++ layer should not know anything about ids, that's RPC's responsibility.

include/perfetto/trace_processor/summarizer.h
```diff
@@ -0,0 +1,87 @@
...
+#include <string>
+#include <vector>
+
+#include "perfetto/base/export.h"
+#include "perfetto/base/status.h"
+
+namespace perfetto::trace_processor {
+
+// Per-query sync info returned by Summarizer::UpdateSpec().
+struct PERFETTO_EXPORT_COMPONENT SummarizerQuerySyncInfo {
```
LalitMaganti: Please nest this struct inside of SummarizerUpdateSpecResult

include/perfetto/trace_processor/trace_processor.h
LalitMaganti: Ah actually we don't have base::StatusOr avaialble here. We should isntead return base::Status and take a pointer to an std::unique_ptr as an out parameter instead which we fill in.

PR #4692, opened 2026-02-05
Fix: Isolate clock tracking for multi-trace TAR loading

maunvz:
When multiple trace files are loaded from a TAR container, they were sharing the same clock tracker (machine_id=0), causing timestamp synchronization issues. Events from subsequent traces would appear offset by the duration of previous traces.

Fix: Use unique machine IDs (based on file_id) for each trace file to ensure each gets its own isolated clock tracker context.

Before:
image

After:
image

### DIFF
--- a/src/trace_processor/forwarding_trace_parser.cc
+++ b/src/trace_processor/forwarding_trace_parser.cc
@@ -128,7 +128,11 @@ base::Status ForwardingTraceParser::Init(const TraceBlobView& blob) {
   // TODO(b/334978369) Make sure kProtoTraceType and kSystraceTraceType are
   // parsed first so that we do not get issues with
   // SetPidZeroIsUpidZeroIdleProcess()
-  trace_context_ = input_context_->ForkContextForTrace(file_id_, 0);
+  // Use file_id as the machine ID to ensure each trace file in a multi-trace
+  // session gets its own clock tracker. This prevents timestamp offsets when
+  // loading multiple traces from a TAR container.
+  trace_context_ =
+      input_context_->ForkContextForTrace(file_id_, file_id_.value);
   if (trace_type_ == kProtoTraceType || trace_type_ == kSystraceTraceType) {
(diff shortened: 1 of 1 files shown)


LalitMaganti:
Hey thanks for the PR! However I don't think this change is correct. Clock trackers *should* be shared across traces on the same machine. 

I believe actually the problem is that you are using incremental clocks and we don't correctly multiplex those per tracker.

If you want to update this PR to go that direction, please go ahead! Alternatively I can also take a look if you prefer.

maunvz:
> Hey thanks for the PR! However I don't think this change is correct. Clock trackers _should_ be shared across traces on the same machine.
> 
> I believe actually the problem is that you are using incremental clocks and we don't correctly multiplex those per tracker.
> 
> If you want to update this PR to go that direction, please go ahead! Alternatively I can also take a look if you prefer.

Thanks for the quick reply! Yeah I think I misunderstood, if you have time to look into that fix I can abandon this PR

LalitMaganti:
Would help if you had a trace I could use to debug this :)

Even just a synthetic one, as long as it's representative, it's enough.

maunvz:
Here are traces from a server / client pair, run on the same machine, loaded with the "load multiple traces" mode - on the current production build you'll see one of the threads is offset:
 
[traces.zip](https://github.com/user-attachments/files/25109740/traces.zip)

LalitMaganti:
I believe https://github.com/google/perfetto/pull/4708 should address this same issue. Once the CI runs, I will test your trace on it.

LalitMaganti:
Confirmed, looks good to me now!

I will need to check performance of that change with a big Chrome trace to make sure we're not regressing load time. Will do that tomorrow :)

Screenshot 2026-02-06 at 00 01 57

LalitMaganti:
Closing in favor of https://github.com/google/perfetto/pull/4708. I verified there is no perf impact as a result of my change

PR #4698, opened 2026-02-05
tp: Add support for variadic arguments in delegate functions

LalitMaganti:
Adds support for variadic function arguments using the ANY... syntax
in PerfettoSQL delegate functions. This allows intrinsic functions to
accept a variable number of arguments at runtime.

Changes:
- Add kAny type to sql_argument::Type enum
- Add is_variadic field to ArgumentDefinition
- Update grammar to parse TYPE... syntax for variadic args
- Variadic arguments are validated to be last in the parameter list
- Variadic args only allowed in delegate functions (not SQL functions)
- Skip type checking for ANY type in created_function.cc
- Use argc=-1 (SQLite's variadic indicator) for variadic functions

### DIFF
--- a/test/trace_processor/diff_tests/syntax/function_tests.py
+++ b/test/trace_processor/diff_tests/syntax/function_tests.py
@@ -294,3 +294,20 @@ class PerfettoFunction(TestSuite):
         14,1
         15,1
       """))
+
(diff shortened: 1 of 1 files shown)


aMayzner:
Please add a diff test for this

PR #4702, opened 2026-02-05
tp: Add tree filtering feature to TreeTransformer

LalitMaganti:
Adds the ability to filter tree structures using SQL WHERE clause
semantics. The FilterTree method takes SQL filter expressions and
uses bytecode execution to efficiently prune tree nodes while
maintaining tree structure integrity.

Key additions:
- New tree-specific bytecode instructions in bytecode_instructions.h
- TreeValueFetcher for evaluating SQL expressions on tree nodes
- FilterTree method in TreeTransformer for applying filters
- SQL intrinsics for tree filtering (tree_filter function)
- Diff tests for tree filtering functionality

### DIFF
--- a/src/trace_processor/core/interpreter/bytecode_instructions.h
+++ b/src/trace_processor/core/interpreter/bytecode_instructions.h
@@ -635,7 +635,7 @@ struct MakeChildToParentTreeStructure : Bytecode {
 //
 // Example: For a tree with parent array [NULL, 0, 0, 1] representing:
 //     0 (root)
-//    / \
+//    / |
 //   1   2
 //   |
 //   3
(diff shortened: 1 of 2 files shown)


src/trace_processor/core/interpreter/bytecode_instructions.h
```diff
@@ -610,6 +610,117 @@ struct Reverse : Bytecode {
...
+//   2. For each node, track closest surviving ancestor
+//   3. Build compacted parent array with reparenting
+//   4. Build compacted original_rows array
+//
+// The node count is derived from parent_span_register.size().
+// After filtering, both span.e pointers are updated to reflect the new count.
+//
+// Scratch registers:
+//   - scratch1: size = n*2, used for surviving_ancestor (first n) and queue
+//   - scratch2: size = n, used for old_to_new mapping
```
sashwinbalaji: can we use this to name the variables rather than 1 & 2.

PR #4708, opened 2026-02-05
tp: make sequence scoped clock ids multi-trace aware

LalitMaganti:
This solves the problem of merged trace files where multiple traces
use the same sequence clock id and same machine id. With incremental
clocks, this can end up with events stacking on top of each other

Instead, change so that clocks are correctly namespaced across machines
and builds.

Inspired by https://github.com/google/perfetto/pull/4692

### DIFF
--- a/src/trace_processor/importers/proto/proto_trace_reader_unittest.cc
+++ b/src/trace_processor/importers/proto/proto_trace_reader_unittest.cc
@@ -187,13 +187,13 @@ TEST_F(ProtoTraceReaderTest, CalculateClockOffset) {
   // Estimated offsets: (10000 + 20000)/2 - 120000 = -105000,
   //                    20000 - (120000 + 140000) / 2 = -110000.
   // Average = -107500.
-  ASSERT_EQ(-107500, clock_offsets[BOOTTIME]);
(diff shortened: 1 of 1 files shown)


src/trace_processor/util/clock_synchronizer.h
```diff
@@ -126,7 +127,45 @@ class TraceProcessorContext;
...
+  // the sequence ID (for sequence-scoped clocks), and the trace file ID
+  // (to isolate sequence-scoped state across different trace files in a TAR).
+  // Non-sequence clocks have seq_id=0, trace_file_id=0.
+  struct ClockId {
+    uint32_t clock_id = 0;
+    uint32_t seq_id = 0;
+    uint32_t trace_file_id = 0;
+
+    constexpr ClockId() = default;
+    constexpr explicit ClockId(int64_t cid)
```
sashwinbalaji: this downcasting seems a bit dangerous. add a dcheck ?

LalitMaganti:
Actually after discussing with Primiano, I think I want to hold off on this until I think a bit of how we want to deal with clocks tracker in merged traces.

Issue #4696, opened 2026-02-05
JSON traces with `"ts": null` fail to load after commit 4329542


ZRHann:
After commit [4329542](https://github.com/google/perfetto/commit/4329542a322e586ef3615c6ac49d69f988a22f56), JSON traces containing `"ts": null` cause errors.

**Questions**

1. Is this strict validation intentional?

2. Should the parser handle null (e.g., default to 0 or skip the event) for backward compatibility?

I am willing to submit a PR to fix this if it's not the intended behavior.

LalitMaganti:
The move to in general tighten up our parsing is intentional. However, if we get a semantic error like this, we generally skip any problematic events and continue parsing. 

Afaik reading that commit, we do just skip the event and not error out the whole trace. Are you seeing some different behavior?

Happy to accept a PR if you've found a case where we're not doing so.

ZRHann:
@LalitMaganti  Thank you for the clarification.

I checked my legacy JSON files and found that `"ts": null` was included specifically in metadata events. According to the Chrome Trace Event Format documentation, metadata events are not supposed to contain a ts field.

Currently, it seems the parser attempts to validate the ts field even for these metadata events. It causes these metadata events to be skipped entirely. Should we skip `ts` parsing entirely for metadata events? Or is the current strict enforcement intended?

LalitMaganti:
Happy to take a PR where we don't try to look at the TS for metadata events *or* just ignore it if it fails.

ZRHann:
Hi @LalitMaganti I have submitted a PR.

LalitMaganti:
This should be fixed now. Thanks for the PR!

PR #4712, opened 2026-02-06
fix: relax ts validation for metadata events

ZRHann:
Fixes a regression from https://github.com/google/perfetto/commit/4329542a322e586ef3615c6ac49d69f988a22f56 where strict timestamp checks caused metadata events () to be rejected.

issue: https://github.com/google/perfetto/issues/4696

### DIFF
--- a/src/trace_processor/importers/json/json_trace_tokenizer.cc
+++ b/src/trace_processor/importers/json/json_trace_tokenizer.cc
@@ -598,7 +598,6 @@ bool JsonTraceTokenizer::ParseTraceEventContents() {
   JsonEvent event;
   base::Status status;
   int64_t ts = std::numeric_limits<int64_t>::max();
-  bool ts_parse_failed = false;
   std::optional<IdResult> id2_local;
   std::optional<IdResult> id2_global;
   for (;;) {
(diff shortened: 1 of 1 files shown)


src/trace_processor/importers/json/json_trace_tokenizer.cc
```diff
@@ -617,9 +618,9 @@ bool JsonTraceTokenizer::ParseTraceEventContents() {
       event.phase = ph.size() >= 1 ? ph[0] : '\0';
     } else if (it_.key() == "ts") {
       if (!CoerceToTs(it_.value(), ts, status)) {
-        PERFETTO_DLOG("%s", status.c_message());
-        context_->storage->IncrementStats(stats::json_tokenizer_failure);
-        return false;
+        ts_parse_failed = true;
```
LalitMaganti: I feel like this can be simplified - why not just leave ts == std::numeric_limits<int64_t>::max() and use that as the failure condition? When we log, we don't distinguish between the two anyway?

src/trace_processor/importers/json/json_trace_tokenizer.cc
```diff
@@ -754,9 +755,12 @@ bool JsonTraceTokenizer::ParseTraceEventContents() {
     context_->storage->IncrementStats(stats::json_tokenizer_failure);
     return true;
   }
-  // Metadata events may omit ts. In all other cases error.
-  if (ts == std::numeric_limits<int64_t>::max()) {
+  // Don't check ts for metadata events. In all other cases error.
+  if (ts == std::numeric_limits<int64_t>::max() || ts_parse_failed) {
     if (event.phase != 'M') {
+      if (ts_parse_failed) {
+        PERFETTO_DLOG("%s", status.c_message());
```
LalitMaganti: I would just dlog this unconditionally.

src/trace_processor/importers/json/json_trace_tokenizer.cc
ZRHann: Done.

Issue #4713, opened 2026-02-06
memory access out of bounds @ trace_processor_memory64.wasm


Ahren-Li:
`UI: https://ui.perfetto.dev/v53.0-a3fe107e3

RuntimeError: memory access out of bounds @ trace_processor_memory64.wasm.perfetto::trace_processor::SliceMipmapOperator::Create(sqlite3*, void*, int, char const* const*, sqlite3_vtab**, char**)
 - trace_processor_memory64.wasm.perfetto::trace_processor::SliceMipmapOperator::Create(sqlite3*, void*, int, char const* const*, sqlite3_vtab**, char**) (wasm://wasm/trace_processor_memory64.wasm-02ee9b3e:wasm-function[4542]:0x4b3979)
 - trace_processor_memory64.wasm.vtabCallConstructor (wasm://wasm/trace_processor_memory64.wasm-02ee9b3e:wasm-function[488]:0x8f04f)
 - trace_processor_memory64.wasm.sqlite3_step (wasm://wasm/trace_processor_memory64.wasm-02ee9b3e:wasm-function[361]:0x7087a)
 - trace_processor_memory64.wasm.perfetto::trace_processor::SqliteEngine::PreparedStatement::Step() (wasm://wasm/trace_processor_memory64.wasm-02ee9b3e:wasm-function[3812]:0x3ee8d8)
 - trace_processor_memory64.wasm.perfetto::trace_processor::PerfettoSqlEngine::ExecuteUntilLastStatement(perfetto::trace_processor::SqlSource) (wasm://wasm/trace_processor_memory64.wasm-02ee9b3e:wasm-function[3670]:0x3d8d3d)
 - trace_processor_memory64.wasm.perfetto::trace_processor::TraceProcessorImpl::ExecuteQuery(std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char>> const&) (wasm://wasm/trace_processor_memory64.wasm-02ee9b3e:wasm-function[127]:0x33676)
 - trace_processor_memory64.wasm.trace_processor_on_rpc_request (wasm://wasm/trace_processor_memory64.wasm-02ee9b3e:wasm-function[306]:0x5da0b)
 -  (engine_bundle.js:1209:14)
 - Object.ccall (engine_bundle.js:5282:15)
 - WasmBridge.onMessage (engine_bundle.js:10937:34)
 -  (stderr: )
 -  (warning: unsupported syscall: __syscall_mprotect)
[...]

PR #4714, opened 2026-02-06
ui: fix SliceMipmapOperator::Create memory access out of bounds

Ahren-Li:
* if sqlite3_column_int64 return -1，static_cast uint32 to 0xffffffff
*0xffffffff + 1 = 0, causing by_depth to be resized to 0.

test: The local run-dev-server does not crash when loading the same perfetto trace.
fix: https://github.com/google/perfetto/issues/4713


stevegolton:
Could you share the trace you're using for testing? I'm quite surprised the UI ever creates a mipmap table on any raw data that has negative depths.

Besides, I don't think we should just continue, we should probably error early as this will presumably cause tracks to break silently.

Ahren-Li:
> Could you share the trace you're using for testing? I'm quite surprised the UI every creates a mipmap table on any raw data that has negative depths.
> 
> Besides, I don't think we should just continue, we should probably error early as this will presumably cause tracks to break silently.

Due to confidentiality issues, I'm not available to provide it right now. I'll try providing a trace without sensitive information.

PR #4721, opened 2026-02-09
ui: Expand chart widgets

aMayzner:
## Summary

Expand the chart widget library with three new SVG-based chart components BarChart, LineChart, PieChart) and their corresponding SQL loaders, alongside shared utilities and a reusable brush controller extracted from the existing Histogram.

## Changes

- Add BarChart component with vertical/horizontal orientation, log scale, brush selection, and an `aggregateBarChartData()` helper
- Add LineChart component with multi-series support, legend, data point, hover, brush selection, and a downsampling utility
- Add PieChart component with donut mode, legend, slice click, and percentage labels
- Add SQL loaders (SQLBarChartLoader, SQLLineChartLoader, SQLPieChartLoader) using QuerySlot for caching and deduplication
- Extract shared chart utilities into chart_utils.ts (tick generation, number formatting, aggregation types, SQL helpers, color palette)
- Extract SvgBrush into its own reusable class, refactoring Histogram to use it
- Add integer dimension support to Histogram (integer tick snapping, integer bucket boundaries)
- Replace old histogram_demo.ts with unified charts_demo.ts on the widgets page
- Add SCSS styles for all three new chart types

### DIFF
--- a/ui/src/components/widgets/charts/bar_chart.ts
+++ b/ui/src/components/widgets/charts/bar_chart.ts
@@ -590,9 +590,9 @@ function aggregate(values: number[], agg: AggregationType): number {
     case 'AVG':
       return values.reduce((a, b) => a + b, 0) / values.length;
     case 'MIN':
-      return Math.min(...values);
+      return values.reduce((a, b) => Math.min(a, b), Infinity);
     case 'MAX':
-      return Math.max(...values);
+      return values.reduce((a, b) => Math.max(a, b), -Infinity);
     case 'COUNT':
       return values.length;
     case 'COUNT_DISTINCT':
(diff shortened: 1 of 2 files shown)


ui/src/components/widgets/charts/pie_chart_loader.ts
stevegolton, line 1: Do we really need separate loaders for pie/bar charts? I feel like some of these charts are just different representations of the same data (e.g. the same aggregates)?

stevegolton (approved):
LGTM!

ui/src/components/widgets/charts/pie_chart_loader.ts
aMayzner: Discussed offline: yes, for now we need. There is a lot of similarities and subtle differences between the underlying data.

PR #4722, opened 2026-02-09
ui: Massive 'slice track' clean up/refactor + improve WebGL performance

stevegolton:
1. Merge `BaseSliceTrack` into `SliceTrack`.
2. Migrate `CpuSliceTrack` over to `SliceTrack`.
3. Improve WebGL rect and marker rendering performance in `SliceTrack` using batching (remove incremental API including `flush()`).
4. Render incomplete slices up to the start of the next slice at that depth, rather than to the end of the window. (Note this only affects incomplete slices where the depth is controlled by the UI as incomplete TP will not reassign a depth if it contains an incomplete slice).
5. Remove `cropIncompleteSlice` and `fadeThinSlices` flags.
   a. `cropIncompleteSlice` is largely superseded by 4.
   b. `fadeThinSlices` logic is broken anyway.
6. Fix issue where instants close to non-instant slices can disappear when zoomed out enough.


stevegolton:
> For incomplete slices on slice tracks, this will be until end of trace anyway as the trace processor will not reassign a depth if it's incomplete.

Ah good point - this only really applies to tracks where the depth is controlled by the UI then.

PR #4724, opened 2026-02-09
TraceBufferV2: improve stats

primiano:
The original PR ended up losing too much information
and made the stats hard to reason about.
This PR is less smart at the cost of more memory

Bug: b/447426810

### DIFF
--- a/src/tracing/service/trace_buffer_v1_with_v2_shadow.cc
+++ b/src/tracing/service/trace_buffer_v1_with_v2_shadow.cc
@@ -48,9 +48,8 @@ uint64_t ComputePacketHash(
   return hasher.digest();
 }
 
-template <typename T>
-void IncrementWithSaturation(T* v) {
-  *v += *v < std::numeric_limits<T>::max() ? *v + 1 : 0;
+void IncrementWithSaturation(uint16_t* v) {
+  *v += *v < UINT16_MAX ? 1 : 0;
 }
 
 }  // namespace


src/tracing/service/trace_buffer_v1_with_v2_shadow.cc
```diff
@@ -46,6 +48,11 @@ uint64_t ComputePacketHash(
   return hasher.digest();
 }
 
+template <typename T>
+void IncrementWithSaturation(T* v) {
+  *v += *v < std::numeric_limits<T>::max() ? *v + 1 : 0;
```
sashwinbalaji: should it be just:
*v +=  *v < std::numeric_limits<T>::max() ?  1 : 0;

src/tracing/service/trace_buffer_v1_with_v2_shadow.cc
```diff
@@ -46,6 +48,11 @@ uint64_t ComputePacketHash(
   return hasher.digest();
 }
 
+template <typename T>
+void IncrementWithSaturation(T* v) {
```
sashwinbalaji: why template ?

src/tracing/service/trace_buffer_v1_with_v2_shadow.cc
primiano: omg very well spotted. i tested it locally but I guess I didn't hit cases of >1 entries in the same bucket.

src/tracing/service/trace_buffer_v1_with_v2_shadow.cc
primiano: Yeah i was very conflicted myself. My only reason was "if I change my mind on the size of the field then the function here doesn't silently go out of sync".
BUt frankly this all code is not going to live much long, so let's be pragmatic.

PR #4726, opened 2026-02-09
Update pinning logic for Android Input Lifecycle

ivan-chong:
- Use the global workspace as the source of truth when pinning, ensuring that when new selections are focused, existing pinned tracks remain pinned.


ivan-chong:
Superseded by #4767

Issue #4658, opened 2026-02-03
Add support for scroll jank v4 metric to org.chromium.ChromeScrollJank plugin


petrcermak:
See https://docs.google.com/document/d/1AaBvTIf8i-c-WTKkjaL4vyhQMkSdynxo3XEiwpofdeA and `scroll_jank_v4*.{h,cc}` source files in https://source.chromium.org/chromium/chromium/src/+/main:cc/metrics/ for more details about the v4 metric.

PR #4729, opened 2026-02-09
Generalize `SliceRef` and `SqlRef`

petrcermak:
1. Generalize `SliceRef` to `TrackEventRef` so that it could be used with plugin slices and non-slice track events (e.g. counters). This pull request allows the caller to explicitly set the table containing the events. It also relaxes the requirement that the ID has to be a `SliceSqlId`.
2. Generalize `SqlRef` to support tables whose ID column name isn't 'id'. This pull request adds a new optional `idColumnName` string argument to `SqlRefAttrs` which allows the caller to specify an ID column name other than 'id'. If the new argument is set and different from 'id', the reference is rendered as `${table}[${idColumnName}=${id}]`.

Issue: #4658

### DIFF
--- a/ui/src/widgets/sql_ref.ts
+++ b/ui/src/widgets/sql_ref.ts
@@ -26,18 +26,25 @@ export interface SqlRefAttrs {
   // The ID of our row.
   // If not provided, `table[Unknown]` is shown with no popup menu.
   id?: number | bigint;
+  // The name of the ID column of the table.
+  // If not provided, defaults to 'id'.
+  idColumnName?: string;
   // Optional additional popup menu items.
   additionalMenuItems?: m.Children;
 }
 
 export class SqlRef implements m.ClassComponent<SqlRefAttrs> {
   view({attrs}: m.CVnode<SqlRefAttrs>) {
-    const {table, id, additionalMenuItems} = attrs;
+    const {table, id, idColumnName = 'id', additionalMenuItems} = attrs;
     if (id !== undefined) {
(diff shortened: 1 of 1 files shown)


ui/src/components/widgets/slice.ts
```diff
@@ -25,8 +25,9 @@ import {
 
 interface SliceRefAttrs {
   readonly trace: Trace;
-  readonly id: SliceSqlId;
+  readonly id: SliceSqlId | number;
```
stevegolton: There's not much point in having the branded type SliceSqlId here if you're going to allow numbers anyway, might as well just omit SliceSqlId entirely.

Also, maybe accept bigint as well.

`readonly id: number | bigint;`

ui/src/components/widgets/slice.ts
```diff
@@ -25,8 +25,9 @@ import {
 
 interface SliceRefAttrs {
```
stevegolton: This is no longer a 'slice' specific widget, it's more of a SqlEventRef widget so could probably be renamed?

petrcermak: I renamed this to `TrackEventRef`, as discussed offline.

ui/src/components/widgets/slice.ts
petrcermak: I removed `SliceSqlId`. I didn't add `bigint` because I thought track events had to fit into 32 bits.

PR #4732, opened 2026-02-10
ui: Clean up WebGL rendering issues

stevegolton:
- Fix layout of slice titles and subtitles on cpu slice tracks and color of subtitles
- Make diagonal lines on RT slices more obvious - Use CSS px rather than physical px to define stripe size and spacing in order to scale correctly on high DPI displays.
- Render gridlines using WebGL instead of canvas so they appear behind the slices instead of in front.
- Clamp incomplete slices to the end of the window to make the fadeout end at the edge of the screen.


primiano (approved):
Thanks, I played with it and it looks great!
(some tests might need rebaselining)

Issue #3495, opened 2025-11-03
Feature Request: Metadata/notes data source


auselen:
(Sorry if this feature already exists in a way that I'm failing to see.)

I would like to embed some notes / my own metadata to traces. Something like
```
data_sources {
  config {
    name: "trace.notes"
    trace_note {
      key: "pmu_set"
      value: "1"
} } }
```

I would like to be able to inject these from command line as well
```
perfetto ... --note "screensize" "$(wm size)" --note "run_id" "$RUN_ID"
```
(command line should be able to override notes embedded in config file)

Would this make sense?

primiano:
so just to double check, you just want to dump some data end to end in the trace.
Which use would you make of this? Just looking at that by querying trace processor?

In theory you could do it today in the way you suggest, something like

```
data_sources {
  config {
    name: "trace.notes"
    legacy_config: "pmu_set=1"
} }
```

And then you could query it in the config (info & stats page of the ui, and whatever TP table it queries to obtain that)

however it's a bit of a hack as you'd be abusing and speculate on the absence of thata datasource

Maybe we could have a properly supported way, which would be a dedicate field in DataSources.
but we'd have to agree on what's the shape it takes in TraceProcessor. maybe we say we only support key-value pairs where the value is a string, and the data ends up in the metadata table as "notes-name1", "notes-name2" columns?

LalitMaganti:
Could we maybe tie it to the concept of built-in data sources we already have instead of doing anything too custom: I could see this makes sense as a "consumer controlled" field of perfetto/traced: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/config/trace_config.proto;drc=0fde4213e1af1de023b881b886d273eeea4493b4;l=105

auselen:
> Which use would you make of this?

Like you describe, it would have its own table, and as you suggest it could be highlighted in info & stats page.

> Maybe we could have a properly supported way, which would be a dedicate field in DataSources.

Sorry for being terse, that's what I was asking for in the title.

> "consumer controlled" field of perfetto/traced

Sure, as long as it is key/value pair and easy to query. Though I would require an example.

LalitMaganti:
> Sure, as long as it is key/value pair and easy to query. Though I would require an example.

Example:
```
builtin_data_sources: {
  notes: {
    key: "foo"
    value: "bar"
  }
}
```

[...]

PR #4735, opened 2026-02-10
perfetto_cmd: allow adding notes to traces (#3495)

auselen:
Add TraceConfig.notes so users can attach arbitrary key/value metadata to a trace, and plumb it through perfetto_cmd via --add-note.

--add-note accepts:
  - key=value (split on the first '=' so values can contain '=')
  - key        (shorthand for key=, i.e. empty value)

Example
  perfetto -o trace.pftrace --time 10s \
      --add-note freq=fixed --add-note conf=X=1 \
      --add-note env1=$ENV1


LalitMaganti (approved):
LGTM thanks!

LalitMaganti:
@primiano friendly ping on this.

PR #4737, opened 2026-02-10
ui: Add pretty print support to V8SourcesTab

camillobruni:
Adding a pretty printer helps debug and analyse minified sources better.

The PrettyPrinter will cache up to 20MB of formatted source code.

### DIFF
--- a/ui/src/plugins/org.chromium.V8/pretty_print_utils.ts
+++ b/ui/src/plugins/org.chromium.V8/pretty_print_utils.ts
@@ -23,7 +23,7 @@ export class PrettyPrintedSource {
   ) {}
 
   get sourceMap(): Int32Array {
-    throw new Error("Not Implemented yet");
+    throw new Error('Not Implemented yet');
   }
 
   // Returns a rough estimate of the entry size, might be off for certain


ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -204,12 +218,81 @@ export class V8SourcesTab implements Tab {
     m.redraw();
   }
 
+  private async togglePrettyPrint() {
+    this.showPrettyPrinted = !this.showPrettyPrinted;
+    this._maybeFormatSources();
```
stevegolton, line 223: Nit: wouldn't bother with the leading underscore

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -204,12 +218,81 @@ export class V8SourcesTab implements Tab {
...
-      language: 'javascript',
-      readonly: true,
-    });
+    if (this.showPrettyPrinted && this.prettyPrintedSource !== undefined) {
+      source = this.prettyPrintedSource.formatted;
+    }
+    return m(
+      '.pf-v8-source-container',
+      {
+        style: {
```
stevegolton, line 264: Static styles should go in the stylesheet

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -204,12 +218,81 @@ export class V8SourcesTab implements Tab {
...
+      m(Editor, {
+        text: source,
+        language: 'javascript',
+        readonly: true,
+        fillHeight: true,
+      }),
+      m(
+        '.pf-v8-floating-toolbar',
+        {
+          style: {
```
stevegolton, line 278: Static styles should go in the stylesheet

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -176,6 +186,10 @@ export class V8SourcesTab implements Tab {
         value: id,
       },
     ];
+    this.selectedScriptSource = it.source;
+    this.prettyPrintedSource = undefined;
+    // Redo source formatting in case it was enabled before.
+    this._maybeFormatSources();
```
stevegolton, line 192: Try and keep this function pure if possible - return everything back through the query slot.

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
stevegolton, line 364: Since the formatting is async - I think this can be elegantly solved using two query slots: one to do the loading and one to do the formatting.

- Make both slots use the same queue to keep things synched.
- The format slot should use the 'shouldFormat' state + selectedId as its key.
- Both slot functions should be pure functions and return the query results back throught the queryslot framework.

I think this current approach is racy if the user selects a different source while formatting for vice versa.

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -204,12 +218,81 @@ export class V8SourcesTab implements Tab {
...
+  }
+
+  async _maybeFormatSources() {
+    if (!this.showPrettyPrinted) return;
+    if (this.prettyPrintedSource) return;
+
+    this.isPrettyPrinting = true;
+    this.prettyPrintedSource = undefined;
+
+    if (!this.prettyPrinter.has(this.selectedScriptSource)) {
```
stevegolton, line 233: This is quite awful, are you sure we need this?

I'm hoping the use of queryslots can make it redundant, i'm not too familiar with how prettier works - looks like it's async but blocks the main thread anyway?

Edit:

Ahh OK I've realized what's going on. Even with queryslot, the browser is blocked from rendering the loading spinner because the continuation runs immediately in the microtasks, blocking the main thread.

Maybe use `await deferChunkedTask();` here instead - that'll make the busy work run in a scheduled task, so the loading spinner can be applied. Or a simple setTimeout() would do TBH.

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -204,12 +218,81 @@ export class V8SourcesTab implements Tab {
...
+      await updateDelay;
+    }
+
+    try {
+      this.prettyPrintedSource = await this.prettyPrinter.format(
+        this.selectedScriptSource,
+      );
+    } finally {
+      this.isPrettyPrinting = false;
+      m.redraw();
```
stevegolton, line 253: I'm hoping all these m.redraw()s can go after migration to queryslot too

stevegolton (changes requested):

ui/src/plugins/org.chromium.V8/v8_sources_tab.ts
```diff
@@ -204,12 +218,81 @@ export class V8SourcesTab implements Tab {
...
+        '.pf-v8-floating-toolbar',
+        {
+          style: {
+            position: 'absolute',
+            bottom: '20px',
+            right: '20px',
+            zIndex: 10,
+          },
+        },
+        m(Button, {
```
stevegolton, line 285: This button is really hard to see in light mode - suggest moving it to a toolbar or something.

PR #4739, opened 2026-02-10
ui: Fix query results tab not updating properly

stevegolton:
This patch fixes a bug where running a new query on the query page then switching back to the timeline would not actually update the columns in the query results tab on the timeline. The first set of columns this tab sees is what it sticks with forever, despite query results changing underneath it.

Testing:
1. Run a query in the query page.
2. Switch to the timeline page.
3. Switch back to the query page and modify the query a different query with a different set of columns.
4. Switch back to the timeline page, confirm the query tab has the right set of columns.

Because this patch involves basically disabling mutable columns, this blocks the user from modifying columns at all in the query tab view - this includes adding, removing, or reordering columns but also sorting. The datagrid component currently doesn't hide these controls when despite the fact that they do nothing, so this patch also fixes this so the controls are no longer visible.

While in the area:
- Remove fixed filters from query results tab, there's no reason not to allow filters
- In DataGrid if filters are immutable (controlled but no onFiltersX callback passed, like above), don't show any of the filter controls.

### DIFF
--- a/ui/src/components/query_table/query_table.ts
+++ b/ui/src/components/query_table/query_table.ts
@@ -293,12 +293,13 @@ export class QueryResultsTable
     return m(DataGrid, {
       schema,
       rootSchema: 'data',
-      initialColumns: resp.columns.map((col) => ({id: col, field: col})),
-      // If filters are defined by no onFilterChanged handler, the grid operates
-      // in filter read only mode.
-      enablePivotControls: false, // In-memory datasource does not support pivoting
+      // Fixed columns - ensures that any changes to the query response are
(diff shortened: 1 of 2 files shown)


sashwinbalaji:
I believe this is also the same issue I hit in http://b/482389144 ?

stevegolton:
> I believe this is also the same issue I hit in http://b/482389144 ?

Yep, there was no reason for filters to be immutable on this grid. Not sure why I set them as such.

I think it would be ideal if the filters synched between the querypage and the timeline tab but that's for another PR.

Issue #4738, opened 2026-02-10
ui: process name does not change in UI despite reading new track_descriptor for the same uuid/pid


adamgreloch:
Hi, I am generating a trace where a process track descriptor will later need to be renamed in the stream. First, I emit a track descriptor with a (gibberish) name:
```
track_descriptor {
  uuid: 79
  process {
    pid: 15
    process_name: "\'\'"
  }
}
```
And after some time I emit a track descriptor for the same UUID/PID but with a different `process_name`:
```
track_descriptor {
  uuid: 79
  process {
    pid: 15
    process_name: "\'sleep 1\'"
  }
}
```
However, the UI keeps the old name:

<img width="248" height="176" alt="Image" src="https://github.com/user-attachments/assets/2d8f6ad4-567c-4bde-b999-9864bf0c46b2" />

I have managed to update the UI metadata of a generic track_descriptor by emitting a new one for the same UUID, but it seems it's not the same for process-like descriptors. Is this behavior intended/am I missing something?

Tested on v53.0-936041655

LalitMaganti:
Can you please provide a minimal self contained trace demonstrating this issue? I believe process renaming should work but maybe there's something subtle I'm missing.

adamgreloch:
Sure, chopped it to 10 packets: [process_renaming.pftrace.zip](https://github.com/user-attachments/files/25213977/process_renaming.pftrace.zip)

<img width="716" height="135" alt="Image" src="https://github.com/user-attachments/assets/74678959-af41-4c65-879a-29569ab5e5bd" />

protoc dump:
<details>

```
$ protoc --decode=perfetto.protos.Trace perfetto_trace.proto < process_renaming.pftrace
packet {
  track_descriptor {
    uuid: 42
    name: "CPUs"
  }
}
packet {
  track_descriptor {
    uuid: 43
    name: "KERNEL"
  }
}
packet {
  track_descriptor {
    uuid: 79
    process {
      pid: 15
      process_name: "\'\'"
    }
  }
}
packet {
  track_descriptor {
    uuid: 75
    thread {
      pid: 15
      tid: 33
    }
  }
}
packet {
  track_descriptor {
    uuid: 76
    name: "sched"
    parent_uuid: 75
  }
}
packet {
  track_descriptor {
    uuid: 77
    name: "events"
    parent_uuid: 75
  }
}
packet {
  track_descriptor {
    uuid: 78
    name: "prio"
    parent_uuid: 75
    counter {
[...]

PR #4742, opened 2026-02-10
tp: fix proc descriptor names not overriding each other

LalitMaganti:
Before this CL, we were accidentally not allowing renaming of process
names originating from processdescriptor tracks. This was because we
were trying to prevent overriding system names which we want to take
priority. But as a consequence we ended up also meaning with two
descriptors, the second would not override the first when it should.

Introduce a priority system for process names, very similar to the
existing one for thread names to deal with both problems at once. This,
as a bonus, also makes process name and thread name handling symmetric
which is great!

Fixes: https://github.com/google/perfetto/issues/4738

### DIFF
--- a/src/trace_processor/importers/common/process_tracker.h
+++ b/src/trace_processor/importers/common/process_tracker.h
@@ -56,9 +56,10 @@ enum class ThreadNamePriority : uint8_t {
 // different sources.
 enum class ProcessNamePriority : uint8_t {
   kOther = 0,
-  kTrackDescriptor = 1,
-  kChromeProcessLabel = 2,
-  kSystem = 3,
+  kChromeProcessLabel = 1,
+  kTrackDescriptor = 2,
+  kChromeProcessLabelRenderer = 3,
+  kSystem = 4,
 };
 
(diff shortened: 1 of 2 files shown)


src/trace_processor/importers/common/process_tracker.cc
```diff
@@ -430,11 +430,19 @@ void ProcessTracker::SetProcessUid(UniquePid upid, uint32_t uid) {
...
   auto& pt = *context_->storage->mutable_process_table();
-  if (auto rr = pt[upid]; !rr.name().has_value()) {
-    rr.set_name(process_name_id);
+  if (PERFETTO_UNLIKELY(process_name_priorities_.size() <= upid)) {
+    process_name_priorities_.resize(upid + 1);
+  }
+  if (priority >= process_name_priorities_[upid]) {
+    pt[upid].set_name(process_name_id);
+    process_name_priorities_[upid] = priority;
   }
```
sashwinbalaji, line 446: nit: add a dlog when we override ?

sashwinbalaji (approved):

PR #4743, opened 2026-02-10
Android tests: switch to libc++

primiano:
Now that libc++ is available as a shlib, switch to
that in our test targets.
libc++_static seemst to create ODR violations if
somebody else (e.g. libandroid.so) pull a shared version of it.

Bug: b/482864728
Bug: b/481873842
Bug: b/483005117

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -492,7 +492,7 @@ cc_library_shared {
     header_libs: [
         "bionic_libc_platform_headers",
     ],
-    stl: "libc++",
+    stl: "libc++_static",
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Should we land this?

LalitMaganti (approved):

primiano:
Keep this on hold. The author of the original breakage says that ag/38385143 is supposed to fix it.
If it does, i'd wait until bots cycle green, and only if they do i'd try to land this (But I want to see them green first)

PR #4744, opened 2026-02-10
perf: Allow raw perf events to specify dev by name

auselen:
Add perf_device option to let raw_perf events derive type from /sys/bus/event_source/devices/<perf_device>/type

### DIFF
--- a/protos/perfetto/common/perf_events.proto
+++ b/protos/perfetto/common/perf_events.proto
@@ -160,10 +160,14 @@ message PerfEvents {
   // struct. Primarily for local use-cases, since the event availability and
   // encoding is hardware-specific.
   message RawEvent {
+    // Either |type| or |perf_device| can be set.
     optional uint32 type = 1;
     optional uint64 config = 2;
     optional uint64 config1 = 3;
     optional uint64 config2 = 4;
+    // The name of a PMU under /sys/bus/event_source/devices
+    // (e.g. "armv8_pmuv3").
+    optional string perf_device = 5;
   }
 
   // Subset of clocks that is supported by perf timestamping.
--- a/protos/perfetto/common/perf_events.proto
+++ b/protos/perfetto/common/perf_events.proto
@@ -160,14 +160,17 @@ message PerfEvents {
   // struct. Primarily for local use-cases, since the event availability and
   // encoding is hardware-specific.
   message RawEvent {
-    // Either |type| or |perf_device| can be set.
+    // Either |type| or |pmu_name| can be set.
     optional uint32 type = 1;
     optional uint64 config = 2;
     optional uint64 config1 = 3;
     optional uint64 config2 = 4;
-    // The name of a PMU under /sys/bus/event_source/devices
+    // The name of a dynamic PMU under /sys/bus/event_source/devices
     // (e.g. "armv8_pmuv3").
-    optional string perf_device = 5;
+    // If set, leave |type| unset.
+    // Android: the relevant sysfs paths are not allowlisted by default, so
+    // this option will require a rooted device with selinux disabled.
+    optional string pmu_name = 5;
   }
 
   // Subset of clocks that is supported by perf timestamping.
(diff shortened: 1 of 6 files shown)


auselen:
@rsavitski ?

src/profiling/perf/event_config.cc
```diff
@@ -371,12 +382,49 @@ PerfCounter PerfCounter::RawEvent(std::string name,
...
+          return devs;
+        }());
+    const auto& devices = perf_devices.ref();
+    auto it = devices.find(perf_device);
+    if (it != devices.end()) {
+      ret.attr_type = it->second;
+    } else {
+      PERFETTO_ELOG("RawEvent config perf_device couldn't be found");
+    }
+  }
```
rsavitski: Can we instead:
* pull this logic into a helper function (see ParseTracepointAndResolveId as an example), otherwise there's a lot of clutter inlined for a "read number from file" step.
* instead of iterating over everything and building a static map, look up exactly that one /sys/.../${name_from_config}/type file, without any caching. Note that you'll need to sanitize the path, the following should be sufficient imo:
```
  if (base::Contains(raw_cfg_name, '/') ||
      base::StartsWith(raw_cfg_name, "..")) {
    return std::nullopt;
  }
```

So the function will effectively do: "sanitize name -> read sysfs path -> parse string into number".

src/profiling/perf/event_config.cc
```diff
@@ -301,8 +306,14 @@ std::optional<PerfCounter> MakePerfCounter(
                                      tracepoint_pb.filter(), *maybe_id);
     } else if (event_desc.has_raw_event()) {
       const auto& raw = event_desc.raw_event();
-      return PerfCounter::RawEvent(name, raw.type(), raw.config(),
-                                   raw.config1(), raw.config2());
+      if (raw.has_perf_device() && !raw.perf_device().empty() &&
+          raw.has_type()) {
+        PERFETTO_ELOG("raw_event cannot specify both type and perf_device.");
+        return std::nullopt;
+      }
```
rsavitski: Similar to the "ParseTracepointAndResolveId()" callsite above, I would call the [new] helper function to get a numeric value for "type", instead of deferring that to PerfCounter::RawEvent constructor.

protos/perfetto/common/perf_events.proto
```diff
@@ -160,10 +160,14 @@ message PerfEvents {
...
   message RawEvent {
+    // Either |type| or |perf_device| can be set.
     optional uint32 type = 1;
     optional uint64 config = 2;
     optional uint64 config1 = 3;
     optional uint64 config2 = 4;
+    // The name of a PMU under /sys/bus/event_source/devices
+    // (e.g. "armv8_pmuv3").
+    optional string perf_device = 5;
   }
```
rsavitski, line 174: I would prefer we call this field "pmu_name". My suggestion (including a clarifier for users on android):
```
  message RawEvent {
    // Alternatively, set |pmu_name|.
    optional uint32 type = 1;
    optional uint64 config = 2;
    optional uint64 config1 = 3;
    optional uint64 config2 = 4;
    // The name of a dynamic PMU under /sys/bus/event_source/devices
    // (e.g. "armv8_pmuv3").
    // If set, leave |type| unset.
    // Android: the relevant sysfs paths are not allowlisted by default, so
    // this option will require a rooted device with selinux disabled.
    optional string pmu_name = 5;
  }
```

rsavitski (changes requested):
Very sorry, I missed this. Will respond way quicker going forward.

auselen:
@rsavitski Just managed to get back to this. Can you have another look please?

auselen:
ping @rsavitski.

rsavitski:
Apologies once again, I suck. To save on extra roundtrips, I'm resending your rebased change as https://github.com/google/perfetto/pull/5475 with a couple tiny fixups. I believe your authorship will be retained in the git log after it's merged.

LalitMaganti:
Closing as this landed in https://github.com/google/perfetto/pull/5475

auselen:
Appreciated.

PR #4745, opened 2026-02-10
tp: Change frame logic in android_input_events 

ivan-chong:
Currently frames are correlated under the assumption that the next event following the input event is the correct frame. This logic is brittle and not necessarily true. This change instead matches input to frame by correlating the frame's deliverInputEvent to the input_event_id.

Bug: 482994760

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/input.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/input.sql
@@ -145,36 +145,47 @@ SELECT
 FROM _interval_intersect!(
   (_clean_android_frames, _clean_deliver_events),
   (utid)
-) ii;
+) AS ii;
 
 CREATE PERFETTO TABLE _input_events_pending_frame_match AS
-SELECT *
+SELECT
+  *
 FROM _clean_deliver_events
-WHERE id NOT IN (SELECT event_id_key FROM _input_event_frame_intersections);
+WHERE
+  NOT id IN (
+    SELECT
+      event_id_key
+    FROM _input_event_frame_intersections
+  );
 
 -- Speculative Match: Find the immediate next frame for non-vsync-aligned events
 CREATE PERFETTO TABLE _input_event_frame_speculative_matches AS
-WITH _ordered_future_frames AS (
-  SELECT
-    e.id AS event_id_key,
-    f.id AS frame_id_key,
-    ROW_NUMBER() OVER (PARTITION BY e.id ORDER BY f.ts ASC) AS rn
-  FROM _input_events_pending_frame_match e
-  JOIN _clean_android_frames f
-    ON e.utid = f.utid
-    AND f.ts >= e.ts
-)
+WITH
+  _ordered_future_frames AS (
+    SELECT
+      e.id AS event_id_key,
+      f.id AS frame_id_key,
+      row_number() OVER (PARTITION BY e.id ORDER BY f.ts ASC) AS rn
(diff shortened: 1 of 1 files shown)


vishniakou (approved):

vishniakou (approved):

vishniakou (approved):

src/trace_processor/perfetto_sql/stdlib/android/input.sql
```diff
@@ -77,66 +79,87 @@ WHERE
...
+      enqeue_slice.name AS enqueue_name,
+      thread_slice.utid,
+      thread_slice.thread_name,
+      str_split(str_split(send_message_slice.name, '=', 1), ',', 0) AS event_channel
+    FROM slice AS send_message_slice
+    JOIN slice AS publish_slice
+      ON send_message_slice.parent_id = publish_slice.id
+    JOIN slice AS start_dispatch_slice
+      ON publish_slice.parent_id = start_dispatch_slice.id
+    JOIN slice AS enqeue_slice
```
vishniakou: enqueue is misspelled

src/trace_processor/perfetto_sql/stdlib/android/input.sql
```diff
@@ -77,66 +79,87 @@ WHERE
...
+      ON start_dispatch_slice.parent_id = enqeue_slice.id
+    JOIN thread_slice
+      ON send_message_slice.id = thread_slice.id
+    WHERE
+      send_message_slice.name GLOB 'sendMessage(*'
+      AND thread_slice.thread_name = 'InputDispatcher'
+  )
+SELECT
+  str_split(str_split(name, '=', 2), ',', 0) AS event_seq,
```
vishniakou, line 104: It would be nice to move away from all this string processing by changing the way this data is reported from the framework. But, this should be dealt with in a separate change.

ivan-chong: Acknowledged

gignat-dev:
I'm not really familiar with this table, but the change seems sound.
@ivan-chong can you also run diff tests? `tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell`

src/trace_processor/perfetto_sql/stdlib/android/input.sql
```diff
@@ -77,66 +79,87 @@ WHERE
...
-JOIN slice AS start_dispatch_slice
-  ON publish_slice.parent_id = start_dispatch_slice.id
-JOIN slice AS enqeue_slice
-  ON start_dispatch_slice.parent_id = enqeue_slice.id
-JOIN thread_slice
-  ON send_message_slice.id = thread_slice.id
+  do_frame_id AS id,
+  ts,
+  dur,
+  CAST(ui_thread_utid AS INTEGER) AS utid,
```
gignat-dev: nit: cast_int!() instead of CAST

ivan-chong:
> I'm not really familiar with this table, but the change seems sound. @ivan-chong can you also run diff tests? `tools/diff_test_trace_processor.py out/linux_clang_release/trace_processor_shell`

@gignat-dev, I saw the diff test failed before and after discussion with @vishniakou, the two options were 1) to ignore these cases where frame does not overlap with event, or 2) speculatively mark the next frame as the correlated frame as a fallback.

I've implemented 2) with a boolean to mark whether the frame was speculatively correlated. This now makes the test pass as before, PTAL thanks!

PR #4746, opened 2026-02-10
ui: Enable webgl rendering by default

stevegolton:
- Configure the CI and playwright containers to install llvmpipe, a software rasterizer which can be used to test webgl on the GPU-less CI.
- Wrap playwright in xvfb-run which creates a virtual frame buffer and x server to allow chrome to see llvmpipe as an opengl driver and use it for webgl.
- Flip the WebGL flag on by default.
- When WebGL is not supported, we fallback to the normal canvas renderer instead.


LalitMaganti:
Really great job on this! Excited for this to be a headline feature for the next release :)

stevegolton:
### CI Testing Notes

- Chrome on the CI seems to want to run using Swiftshader as a WebGL fallback.
- Swiftshader is a software rasterizer built into Chrome that's designed to be used for compositing and as a WebGL fallback when no hardware GPU is present.
```bash
# Testing Swiftshader on cloudtop
/opt/google/chrome/chrome --enable-unsafe-swiftshader --use-gl=angle --use-angle=swiftshader-webgl
```
- It seems Swiftshader does not support the fragment shader we use for rendering chevrons properly, as we end up with artifacts around the edges that are supposed to be discarded or transparent. It also seems to be very **very** slow when testing on my cloudtop at least, and really is not a practical solution. We really don't want users to end up falling back to this. [CI results](https://github.com/google/perfetto/actions/runs/21940271805/job/63363945680).

#### Rendering artifacts around chevrons
image

#### Half finished lines
image

  - The artifacts seem to be a Swiftshader issue rather than an issue with our fragment shader, as the latter works just fine with AMD, Intel, and Apple drivers.
- Thankfully, swiftshader has now been deprecated as a WebGL fallback due to security issues and poor performance. This started being rolled out in Chrome 139 released August 5th, 2025 [source](https://groups.google.com/a/chromium.org/g/blink-dev/c/yhFguWS_3pM?pli=1) [source](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/gpu/swiftshader.md). Thus, I don't think we need to do anything special here.
- We were using **Chrome 138** version in the CI up until yesterday when I happened to update it to **145** for unrelated reasons [source](https://github.com/google/perfetto/pull/4748/changes#diff-25db25f7c22b5c3caa8492dd43c3d54f3ba1bd5758cf9922b74ab402633a381b). This is why there was a discrepancy between cloudtop and the CI. Now, the CI runs with our canvas2d fallback by default.

### Options for testing WebGL path in the CI

1. Run swiftshader, make our fragment shaders compatible with it.
  a. It might actually be too slow to even run on the CI - we still end up with half rendered frames when the screenshot is taken.
  b. Optimizing for a rasterizer that no one is going to be running.
2. Run a VM with a real hardware GPU.
  a. Expensive, impractical.
3. Run an external software implementation (e.g. llvmpipe).
  a. Tests WebGL render path, not too slow, representative of real hardware.

### LLVMPipe

LLVMPipe is a software rasterizer, that actually performs better than our canvas2d fallback, and more importantly, seems to be much more compliant and works with our fragment shaders without artifacts. We can use this in headless mode on our CI using EGL, but we need to install a few more packages in the container. Chrome typically blocklists LLVMPipe as a GPU driver, so we must pass `--ignore-gpu-blocklist`.

```bash
# Testing LLVMPipe on cloudtop: 
/opt/google/chrome/chrome --ignore-gpu-blocklist
```

LalitMaganti:
So what's the plan in that case? Do we want to enable llvmpipe on the CI and then land this?

stevegolton:
> So what's the plan in that case? Do we want to enable llvmpipe on the CI and then land this?

I think that makes the most sense. I'm attempting to update the CI container now.

PR #4750, opened 2026-02-10
ui: Improve track shell/button layout

stevegolton:
- Since we the track shell is now resizable, the user can simply resize the shell to see long track names:
  - Remove middle ellipsis track titles - just use normal CSS 'end' ellipsis
  - Remove full track title tooltips on hover
- Make track shell buttons display: none when not hovered to avoid taking up any space at all, leaving more room for the track titles. The end ellipsis (as opposed to the middle ellipsis) works very nicely with this approach and avoids unpleasant jumping around.
  - This is how various panels in vscode behave - it's a common UX.
- Note: The permanent (i.e. not appear-on-hover) buttons (e.g. pin, checkbox) have been moved to the end so that they don't shift around when hovered which would be infuriating!

[a50f6458-1cf9-4c98-88ef-1a9d4dad5bd2.webm](https://github.com/user-attachments/assets/37c0e437-6000-4179-a6b7-c7c5c9a2a5da)


stevegolton:
Supersedes https://github.com/google/perfetto/pull/2944

stevegolton:
@rsavitski @primiano @LalitMaganti curious to see your thoughts on this - is it a nicer UX or are we throwing the baby out with the bathwater by removing the middle ellipsis & the track popup?

LalitMaganti:
I personally like it a lot. I always felt middle ellipsis was at best a workaround for non having resizable title section.

stevegolton:
> I personally like it a lot. I always felt middle ellipsis was at best a workaround for non having resizable title section.

Agreed! Perhaps we could get this landed then?

LalitMaganti (approved):

PR #4756, opened 2026-02-11
de: tp: Add metrics v2 node

aMayzner:
DE: Adds metrics v2 node, with support for one metric column

### DIFF
--- a/ui/src/assets/explore_page/node_info/metrics.md
+++ b/ui/src/assets/explore_page/node_info/metrics.md
@@ -1,0 +1,38 @@
+# Metrics
+
+**Purpose:** Define trace-based metrics from your query results. This node packages your data into a `TraceMetricV2TemplateSpec` proto (metric bundle) that can be exported and used in trace analysis pipelines. The selected value column becomes the metric value, and all other columns become dimensions.
+
+**How to use:**
+
+1. **Connect an input:** This node requires a source of data (e.g., from a Table Source or after filtering/aggregating data)
+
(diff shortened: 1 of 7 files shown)


include/perfetto/trace_processor/trace_processor.h
```diff
@@ -250,6 +250,16 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessor : public TraceProcessorStorage {
   // materialization of structured queries. On success, |out| is populated with
   // the new instance and ownership is transferred to the caller.
   virtual base::Status CreateSummarizer(std::unique_ptr<Summarizer>* out) = 0;
+
+  // EXPERIMENTAL: Converts a proto from binary format to textproto format.
```
LalitMaganti: Uhhhhh why do you need this? This seems totally unnecessary.

LalitMaganti: No sorry please don't add a function like this, this is totally not the right thing to do. You should be able to convert to text proto in JavaScript without going to trace processor. If you need to do it on c++ see how it's done for trace config and generalize. I've allowed ad-hoc changes on TP and not looked too closely but this is where I draw the line. I don't want to add random things in TP just because it's convenient for explore page.

aMayzner: I will add functionality to go from protojs to text proto without TP, but I don't get the strong reaction. So far, this is what we have been always doing. To get trace config, in recording page, in Data explorer. The idea of continuing with it was not so random.

LalitMaganti: No the reaction here is because you are basically just dumping a shortcut to do proto to prototext of any proto trace processor understands. That's not the job of trace processor and I think that should be reasonably clear

LalitMaganti:
No TP code left so removing myself.

PR #4759, opened 2026-02-11
Trace Processor: Add regexp_replace function to Perfetto SQL

simonmacm:
Bug: http://b/483616653

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/utils.h
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/utils.h
@@ -398,8 +398,8 @@ struct RegexpExtract : public sqlite::Function<RegexpExtract> {
   }
 };
 
-struct RegexpReplace : public sqlite::Function<RegexpReplace> {
-  static constexpr char kName[] = "regexp_replace";
+struct RegexpReplaceSimple : public sqlite::Function<RegexpReplaceSimple> {
+  static constexpr char kName[] = "regexp_replace_simple";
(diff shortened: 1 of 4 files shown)


test/trace_processor/diff_tests/stdlib/prelude/regexp_replace.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class RegexpReplace(TestSuite):
+
+  def test_regexp_replace(self):
+    return DiffTestBlueprint(
+        trace=TextProto(''),
+        query="""
+        SELECT
+          REGEXP_REPLACE('abcde', 'c', 'C') AS c,
```
LalitMaganti: These strings don't seem consistent with the behaviour of http://go/f1-1#regexp-replace. You don't seem to support groups etc properly.

Either can we implement something which is ~1:1 compatible with F1 *or* call this something diffferent (maybe regexp_replace_simple) or similar to avoid confusing people and leave open the possibility to implement the fully featured operator in the future.

simonmacm: Let's go for regexp_replace_simple.

Unfortunately I managed to get git into a very confused state on this PR. I think it needs to be abandoned and we can pick up at https://github.com/google/perfetto/pull/4763.

simonmacm:
This got trashed by mistake, continue at https://github.com/google/perfetto/pull/4763

PR #4761, opened 2026-02-11
de: Migrate charts to echarts

aMayzner:
- Add dependency on echarts
- Migrate existing charts to echarts
- Add scatterplot and treemap charts
- Add grouping: multiple charts changing together.

### DIFF
--- a/ui/src/components/widgets/charts/bar_chart.ts
+++ b/ui/src/components/widgets/charts/bar_chart.ts
@@ -15,16 +15,14 @@
 import m from 'mithril';
 import type {EChartsCoreOption} from 'echarts/core';
 import {AggregationType, extractBrushRange, formatNumber} from './chart_utils';
-import {
-  EChartView,
-  EChartEventHandler,
-  getPerfettoThemeColors,
-} from './echart_view';
+import {EChartView, EChartEventHandler} from './echart_view';
 import {
   buildAxisOption,
   buildGridOption,
   buildBrushOption,
+  buildTooltipOption,
 } from './chart_option_builder';
+import {getChartThemeColors} from './chart_theme';
 
 /**
  * A single bar in the bar chart.
@@ -155,12 +153,7 @@ function buildBarOption(
   } = attrs;
   const fmtMeasure = formatMeasure ?? formatNumber;
 
-  // Only get theme for custom color overrides
-  const theme =
-    barColor === undefined || barHoverColor === undefined
-      ? getPerfettoThemeColors()
-      : undefined;
-
+  const theme = getChartThemeColors();
   const horizontal = orientation === 'horizontal';
   const labels = data.items.map((item) => String(item.label));
 
@@ -192,17 +185,18 @@ function buildBarOption(
 
   const option: Record<string, unknown> = {
     animation: false,
+    color: [...theme.chartColors],
     grid: buildGridOption({
       bottom: dimensionLabel && !horizontal ? 45 : 25,
     }),
-    tooltip: {
+    tooltip: buildTooltipOption({
       trigger: 'axis' as const,
       axisPointer: {type: 'shadow' as const},
       formatter: (params: Array<{name?: string; value?: number}>) => {
         const p = Array.isArray(params) ? params[0] : params;
         return `${p.name ?? ''}<br>${measureLabel}: ${fmtMeasure(p.value ?? 0)}`;
       },
-    },
+    }),
     xAxis: horizontal ? valueAxis : categoryAxis,
     yAxis: horizontal ? categoryAxis : valueAxis,
     series: [
@@ -210,12 +204,11 @@ function buildBarOption(
         type: 'bar',
         data: data.items.map((item) => item.value),
         itemStyle: barColor !== undefined ? {color: barColor} : undefined,
-        emphasis:
-          barHoverColor !== undefined
-            ? {itemStyle: {color: barHoverColor}}
-            : theme !== undefined
-              ? {itemStyle: {color: theme.accentColor}}
-              : undefined,
+        emphasis: {
+          itemStyle: {
+            color: barHoverColor ?? theme.accentColor,
+          },
+        },
       },
     ],
   };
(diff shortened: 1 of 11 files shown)


ui/src/components/widgets/charts/echart_view.ts
```diff
@@ -0,0 +1,343 @@
...
+    this.activateBrush(attrs.activeBrushType);
+  }
+
+  private scheduleResize(): void {
+    // Batch all resize requests via requestAnimationFrame to avoid
+    // triggering ResizeObserver loop warnings. Multiple resize requests
+    // within the same frame are coalesced into a single resize.
+    if (this.resizePending || this.chart === undefined) return;
+    this.resizePending = true;
+    requestAnimationFrame(() => {
```
stevegolton: Why do we need requestAnimationFrame? And below? This is usually a hack.

ui/src/components/widgets/charts/echart_view.ts
```diff
@@ -0,0 +1,481 @@
...
+let themeObserver: MutationObserver | undefined;
+
+/**
+ * Starts observing theme provider class changes to detect theme switches.
+ * Only creates the observer when the first chart mounts.
+ */
+function startThemeObserver(): void {
+  if (themeObserver !== undefined) return;
+
+  const themeProvider = document.querySelector('.pf-theme-provider');
```
stevegolton: The theme is not necessarily provided by a theme provider, so this class might not exist and we shouldn't depend on it to listen for changes.

Instead I think we should just read the relevant colours out every frame and update the chart if they've changed with `getPropertyValue(--pf-color-....)`

This may cause a reflow, but I think that's fine. We cause a reflow in a few other places inside onupdate anyway.

ui/src/components/widgets/charts/echart_view.ts
```diff
@@ -0,0 +1,481 @@
...
+      // last option, so ECharts' cached dimensions could be stale.
+      this.chart.resize();
+      this.activateBrush(attrs.activeBrushType);
+    }
+    this.syncHandlers(attrs.eventHandlers ?? []);
+  }
+
+  private initChart(attrs: EChartViewAttrs): void {
+    if (this.container === undefined || attrs.option === undefined) return;
+    this.chart = echarts.init(this.container, this.currentTheme);
```
stevegolton: I think it might be more elegant here to control theme colors using setOption() if possible and avoid themes entirely as we can modify the colors without having to reinitialize the entire chart.

ui/src/components/widgets/charts/scatter_chart_loader.ts
```diff
@@ -0,0 +1,249 @@
...
+  dispose(): void {
+    this.querySlot.dispose();
+  }
+}
+
+/**
+ * Deterministically sample points to reduce to maxPoints.
+ * Uses stride-based sampling for uniform distribution across the dataset.
+ */
+function samplePoints(
```
stevegolton: Can we offload this to SQL?

ui/src/components/widgets/charts/scatterplot.ts
stevegolton, line 1: This is usually called a scatterplot not a scatter_chart IIUC

ui/src/components/widgets/charts/treemap.ts
stevegolton, line 1: Maybe just call this treemap, not treemap_chart?

ui/src/components/widgets/charts/chart_theme.ts
```diff
@@ -0,0 +1,79 @@
...
+  readonly backgroundColor: string;
+  readonly accentColor: string;
+  readonly chartColors: readonly string[];
+}
+
+/**
+ * Returns true if dark theme is currently active.
+ */
+export function isDarkTheme(): boolean {
+  const themeProvider = document.querySelector('.pf-theme-provider');
```
stevegolton: We really shouldn't be depending on theme provider as it embedded builds could omit it entirely.

Just listen for changes to the css variables by querying the property value from the DOM every frame and comparing it against the previous colors.

ui/src/components/widgets/charts/bar_chart.ts
```diff
@@ -115,437 +117,157 @@ export interface BarChartAttrs {
...
-  chartHeight: number;
-  slotSize: number;
-  data: BarChartData;
-  measureTicks: number[];
-  valueToMeasure: (v: number) => number;
-  formatMeasure: (v: number) => string;
-  formatLabel: (l: string | number) => string;
-  dimensionLabel: string | undefined;
-  measureLabel: string | undefined;
+  return option as EChartsCoreOption;
```
stevegolton, line 226: Do we need this type assertion?

ui/src/components/widgets/charts/echart_view.ts
```diff
@@ -0,0 +1,481 @@
...
+  private resizeObs?: Disposable;
+  private prevHandlers: ReadonlyArray<EChartEventHandler> = [];
+  private prevOptionJson?: string;
+  private currentTheme: 'perfetto-light' | 'perfetto-dark' = 'perfetto-light';
+
+  oncreate({dom, attrs}: m.CVnodeDOM<EChartViewAttrs>) {
+    ensureEChartsSetup();
+    this.currentTheme = getCurrentThemeName();
+
+    const container = dom.querySelector(
```
stevegolton: assertExists() is convention for developer errors like this

ui/src/components/widgets/charts/echart_view.ts
```diff
@@ -0,0 +1,481 @@
...
+      {
+        className: classNames(
+          attrs.fillParent && 'pf-echart-view--fill-parent',
+          attrs.className,
+        ),
+        // When fillParent is set, let the CSS class control height (100%)
+        // instead of setting an explicit pixel height via inline style.
+        style: attrs.fillParent ? undefined : {height: `${height}px`},
+      },
+      [
```
stevegolton: Nit: Unnecessary array

PR #4762, opened 2026-02-11
tp: SQ should depend on referenced queries

aMayzner:
Introduces `referenced_query` as a replacement for `inner_query_id` in the structured query proto, along with a new `referenced_queries` repeated field for explicitly declaring cross-query dependencies. This makes dependency graphs explicit rather than inferred, enabling validation that all references are accounted for and improving correctness of dependency propagation.


docs/analysis/trace-summary.md
```diff
@@ -787,10 +787,19 @@ A query's source can be one of the following:
 - **`sql`**: An arbitrary SQL `SELECT` statement.
 - **`simple_slices`**: A convenience for querying the `slice` table.
 - **`inner_query`**: A nested structured query.
-- **`inner_query_id`**: A reference to a shared structured query.
+- **`referenced_query`**: A reference to a shared structured query by its `id`.
```
LalitMaganti, line 790: Sorry but we cannot break backcompat. people are almost certainly using this.

protos/perfetto/perfetto_sql/structured_query.proto
```diff
@@ -589,10 +589,20 @@ message PerfettoSqlStructuredQuery {
   // Optional.
   repeated string referenced_modules = 11;
 
-  // Represents the "source" of the query which will be translared to an SQL
+  // The IDs of other structured queries referenced via `referenced_query`
```
LalitMaganti, line 592: No sorry I don't want this thing to be added. I don't understand why we're needing this - we should be able to figure this out in trace processor.

docs/analysis/trace-summary.md
aMayzner: I'm not, I just stopped mentioning this in the doics. Maybe I should, just mark it as legacy.

protos/perfetto/perfetto_sql/structured_query.proto
aMayzner: The problem I'm trying to fix - Summarizer is trying to track dependencies, but as we don't know what fields the dependencies are hidden in, we need to hard code the dependencies per each source. This way we would have one nice set of dependencies, which would make dependencies tracking cleaner, for Summarizer and generally to understand how your query works

PR #4767, opened 2026-02-11
ui: Add a generic Related Events API

ivan-chong:
This abstracts the logic from AndroidInputLifecycle into components which can be more generically consumed by other plugin creators who wish to represent related events with common utilities to create a tab for organisation and navigation of details, as well as drawing arrows.

The ui/src/components/related_events API exposes 
- A generic arrow_visualiser which draws arrows between two points on the track using trackUri, ts and depths
- An overlay which allows rendering of these arrows by a plugin, built from a set of connections consisting of start and end points and an optional arrow color
- Various utilities e.g. getting a trackid by trackUri, calculating the depth of a slice on a track and a manager which handles pinning of sets of tracks.

### DIFF
--- a/ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
+++ b/ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
@@ -79,7 +79,7 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
     });
 
     trace.commands.registerCommand({
-      id: 'openAndroidInputLifecycleTab',
+      id: 'com.android.openAndroidInputLifecycleTab',
       name: 'Android: View Input Lifecycle',
       callback: () => {
         trace.tabs.showTab('com.android.AndroidInputLifecycleTab');


ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
```diff
@@ -0,0 +1,318 @@
...
+    const relations: Relation[] = [];
+    const overlayEvents: RelatedEvent[] = [];
+    const overlayRelations: Relation[] = [];
+
+    const it = result.iter({
+      input_id: STR,
+      channel: STR,
+      total_latency: NUM_NULL,
+
+      ts_reader: NUM_NULL,
```
stevegolton: These ts and durs should really be of type LONG_NULL. That way they'll be materialized as a bigint, eliminating precision errors.

ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
```diff
@@ -0,0 +1,318 @@
...
+      it.id_reader !== null &&
+      it.ts_reader !== null &&
+      it.track_reader !== null &&
+      it.dur_reader !== null
+    ) {
+      const trackUri = getTrackUriForTrackId(this.trace, it.track_reader);
+      if (trackUri) {
+        readerEvent = {
+          id: it.id_reader,
+          ts: BigInt(it.ts_reader) as time,
```
stevegolton: If ts_reader were of type LONG_NULL instead of NUM_NULL, the BigInt() wouldn't be required.

Also, instead of type asserting here, we have a safer Time.fromRaw() function which ensures the input is at least a bigint.

ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
```diff
@@ -0,0 +1,318 @@
...
+      it.id_receive !== null &&
+      it.ts_receive !== null &&
+      it.track_receive !== null &&
+      it.dur_receive !== null
+    ) {
+      const trackUri = getTrackUriForTrackId(this.trace, it.track_receive);
+      if (trackUri) {
+        receiveEvent = {
+          id: it.id_receive,
+          ts: BigInt(it.ts_receive) as time,
```
stevegolton: Same as above (and all below)

ui/src/plugins/dev.perfetto.RelatedEvents/index.ts
stevegolton, line 1: Though I appreciate the separation of the plugin, I think since this plugin is a pure library, this stuff could be put directly into ui/src/components, which is the right place for stuff like this e.g. track implementations, general purpose panels, etc.

ui/src/plugins/dev.perfetto.RelatedEvents/index.ts
```diff
@@ -0,0 +1,80 @@
...
+// - Interfaces from './interface' (e.g., RelatedEventData, EventSource)
+//   to structure your event relationship data.
+// - 'drawRelatedEvents' from './arrow_visualiser' to render arrows on the timeline.
+// - 'GenericRelatedEventsOverlay' from './generic_overlay' for a quick overlay implementation.
+// - Utility functions from './utils' for common tasks like track URI lookups.
+//
+// Example usage:
+//
+// // 1. Implement EventSource
+// class SimpleEventSource implements EventSource {
```
stevegolton: I don't really see why EventSource as an interface needs to exist - it doesn't seem like any one uses it.

ui/src/plugins/com.android.AndroidInputLifecycle/tab.ts
```diff
@@ -121,242 +118,71 @@ export class AndroidInputTab implements Tab {
...
-          it.id_receive,
-          it.id_consume,
-          it.id_frame,
-        ].includes(clickedEventId);
-
-        if (matchesClickedEvent && rowToHighlight === undefined) {
-          rowToHighlight = uniqueId;
+      for (const event of data.events) {
+        if (event.type === 'InputLifecycle') {
+          const parsedArgs = InputLifecycleArgsSchema.safeParse(
```
stevegolton: We know we're using an AndroidInputEventSource here, why do we need to route custom args through a wide interface then parse them back to their specific types? We could just return the correct types in the first place.

ivan-chong:
@stevegolton Friendly ping, PTAL thanks!

ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
```diff
@@ -0,0 +1,217 @@
...
+          receiver: this.createStageArgs(receiveEvent, dispatchEvent),
+          consumer: this.createStageArgs(consumeEvent, receiveEvent),
+          frame: this.createStageArgs(frameEvent, consumeEvent),
+          allTrackIds: [
+            it.track_reader,
+            it.track_dispatch,
+            it.track_receive,
+            it.track_consume,
+            it.track_frame,
+          ].filter((t) => t !== null) as number[],
```
stevegolton: I'd avoid the type assertion here, TS is smart enough to narrow the output type of the filter, and besides `customArgs` are `unknown` anyway right?

ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
```diff
@@ -0,0 +1,217 @@
...
+        consumeEvent,
+        frameEvent,
+      ] = stages;
+
+      const tabEvent: RelatedEvent = {
+        id: eventId,
+        ts: (readerEvent?.ts ??
+          dispatchEvent?.ts ??
+          receiveEvent?.ts ??
+          0n) as time,
```
stevegolton: I'd try to avoid the type assertion here as well by using `Time.fromRaw(0n)` or `Time.ZERO`.

Also, does it really make sense to use 0 here if none of the others are defined? Might be better to show nothing, or drop this related event.

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -12,40 +12,188 @@
...
+        this.visibleRowIds.clear();
+        this.lastAppliedEventId = undefined;
       },
     });
 
     trace.commands.registerCommand({
-      id: 'com.android.AndroidInputLifecycle#ViewFlow',
-      name: 'Android: View Input Flow',
-      callback: () => trace.tabs.showTab(tabUri),
+      id: 'openAndroidInputLifecycleTab',
```
stevegolton: One last thing:

this should have a com.android prefix

something like: `com.android.OpenAndroidInputLifecycleTab`

PR #4768, opened 2026-02-11
tp: integrate ProtoVM

rukkal:
Integrate ProtoVM into TP as specified in go/perfetto-proto-vm.

### DIFF
--- a/src/trace_processor/importers/proto/proto_trace_reader.cc
+++ b/src/trace_processor/importers/proto/proto_trace_reader.cc
@@ -213,17 +213,6 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   // Any compressed packets should have been handled by the tokenizer.
   PERFETTO_CHECK(!decoder.has_compressed_packets());
 
-  if (decoder.has_trace_provenance()) {
-    protovm_.ProcessTraceProvenancePacket(decoder.trace_provenance());
-  } else if (decoder.has_protovms()) {
-    protovm_.ProcessProtoVmsPacket(decoder.protovms(), packet);
-  } else if (auto new_packet = protovm_.TryProcessPatch(decoder, packet);
-             new_packet) {
-    packet = std::move(*new_packet);
-    decoder =
-        protos::pbzero::TracePacket::Decoder(packet.data(), packet.length());
-  }
-
   // When the trace packet is emitted from a remote machine: parse the packet
   // using a different ProtoTraceReader instance. The packet will be parsed
   // in the context of the remote machine.
@@ -247,6 +236,19 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   // machine_id is set only for remote machines.
   PERFETTO_DCHECK(decoder.has_machine_id() ==
                   (context_->machine_id() != MachineId(kDefaultMachineId)));
+
+  // Execute ProtoVM logic right after the "machine ID fork" as detailed in
+  // go/perfetto-proto-vm.
+  if (decoder.has_trace_provenance()) {
+    protovm_.ProcessTraceProvenancePacket(decoder.trace_provenance());
+  } else if (decoder.has_protovms()) {
+    protovm_.ProcessProtoVmsPacket(decoder.protovms(), packet);
+  } else if (auto new_packet = protovm_.TryProcessPatch(decoder, packet);
+             new_packet) {
+    packet = std::move(*new_packet);
+    decoder =
+        protos::pbzero::TracePacket::Decoder(packet.data(), packet.length());
+  }
 
   uint32_t seq_id = decoder.trusted_packet_sequence_id();
   auto [scoped_state, inserted] = sequence_state_.Insert(seq_id, {});
--- a/src/trace_processor/importers/proto/protovm_incremental_tracing.cc
+++ b/src/trace_processor/importers/proto/protovm_incremental_tracing.cc
@@ -108,6 +108,9 @@ TraceBlobView ProtoVmIncrementalTracing::SerializeIncrementalState(
         proto->set_trusted_pid(patch.trusted_pid());
         proto->set_trusted_packet_sequence_id(
             patch.trusted_packet_sequence_id());
+        if (patch.has_machine_id()) {
+          proto->set_machine_id(patch.machine_id());
+        }
       });
 }
 


src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -332,6 +332,23 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
...
+
+  if (decoder.has_protovms()) {
+    protovm_incremental_tracing_.InstantiateProtoVms(decoder.protovms());
+  }
+
+  auto inflated_state = protovm_incremental_tracing_.TryProcessPatch(packet);
+  if (inflated_state.IsOk()) {
+    packet = TraceBlobView(std::move(*inflated_state));
+  } else if (inflated_state.IsAbort()) {
+    context_->storage->IncrementStats(stats::protovm_abort);
```
LalitMaganti: Please use the new import logging infra and properly log the reason for this.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -332,6 +332,23 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
     ParseTraceConfig(decoder.trace_config());
   }
 
+  if (decoder.has_trace_provenance()) {
```
LalitMaganti: All of this code should also be in its own module. That way, you can just have the "ProtoVM module" holding all this code and below. You can have all the logic below happen at tokenization time.

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,108 @@
...
+#include "protos/perfetto/trace/trace_packet.pbzero.h"
+
+namespace perfetto {
+namespace trace_processor {
+
+namespace {
+TraceBlob MakeIncrementalStatePacket(
+    const protovm::Vm& vm,
+    const protos::pbzero::TracePacket::Decoder& patch) {
+  std::string incremental_state_without_trusted_fields =
```
LalitMaganti: This method needs to be rewritten to take as input a protozero::Message* to avoid a copy. Right now RwProto::SerializeAsString() is itself making two copies.

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,108 @@
...
+      vm.SerializeIncrementalState();
+
+  protozero::HeapBuffered<protos::pbzero::TracePacket> incremental_state;
+  incremental_state->AppendRawProtoBytes(
```
LalitMaganti: This is copy number 3

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,108 @@
...
+      incremental_state_without_trusted_fields.data(),
+      incremental_state_without_trusted_fields.size());
+  incremental_state->set_trusted_uid(patch.trusted_uid());
+  incremental_state->set_trusted_pid(patch.trusted_pid());
+  incremental_state->set_trusted_packet_sequence_id(
+      patch.trusted_packet_sequence_id());
+
+  auto serialized = incremental_state.SerializeAsString();
```
LalitMaganti: Copy number 4

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,108 @@
...
+  return TraceBlob::CopyFrom(serialized.data(), serialized.size());
```
LalitMaganti: Copy number 5!

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+protovm::StatusOr<TraceBlob> ProtoVmIncrementalTracing::TryProcessPatch(
+    const TraceBlobView& blob) {
+  protos::pbzero::TracePacket::Decoder patch(blob.data(), blob.size());
+  if (!patch.has_trusted_packet_sequence_id()) {
+    return protovm::StatusOr<TraceBlob>::Error();
+  }
+  auto producer_id =
```
LalitMaganti: This double-lookup is really ugly. I'd really want to see a very fast sequence id -> vector<ProtoVm> here so we only have a O(1) flatHashMap lookup. Instead I see here two O(logn) lookups including one on multimap which is extra slow.

At the very least the sequence_id_to_producer_id_ happens on every packet because TraceProvenance is emitted uncondiontally in my experience.

Also when the above is refactored into a protoModule, all ofthis needds to be refactored such that it lives in the same file to allow for inlning.

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,108 @@
...
+        instance.program(), 1024 * instance.memory_limit_kb(), state));
+    protovm::Vm* vm = vms_.back().get();
+    for (auto producer_id = instance.producer_id(); producer_id;
+         ++producer_id) {
+      producer_id_to_vm_.emplace(*producer_id, vm);
+    }
+  }
+}
+
+protovm::StatusOr<TraceBlob> ProtoVmIncrementalTracing::TryProcessPatch(
```
LalitMaganti: It's unclear to me why you have protovm::StatusOr and how this is different from base::StatusOr?

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
rukkal: Apologies, I totally forgot I still needed to make this right. Happy to change the ProtoVM interface to fix it.

src/trace_processor/importers/proto/proto_trace_reader.cc
rukkal: A ProtoVM module sounds great.

Sanity check: you are suggesting we move all the work into a ProtoVmModule::TokenizePacket(), called by ProtoTraceReader::TimestampTokenizeAndPushToSorter():
```
base::Status ProtoTraceReader::TimestampTokenizeAndPushToSorter(
    TraceBlobView packet) {
  protos::pbzero::TracePacket::Decoder decoder(packet.data(), packet.length());
  ...

  auto& modules = module_context_.modules_by_field;
  for (uint32_t field_id = 1; field_id < modules.size(); ++field_id) {
    if (!modules[field_id].empty() && decoder.Get(field_id).valid()) {
      for (ProtoImporterModule* module : modules[field_id]) {
        ModuleResult res = module->TokenizePacket(
            decoder, &packet, timestamp, state->current_generation(), field_id);
        if (!res.ignored())
          return res.ToStatus();
      }
    }
  }
  ...
```
correct?

Then we'll have
```
ModuleResult ProtoVmModule::TokenizePacket(const protos::pbzero::TracePacket::Decoder& decoder, TraceBlobView* packet, ...);
```
which might perform a transform from patch to full incremental state, so when the transform happens we'll have to update the `decoder` and `packet` parameters. Correct? Any preference on how we update `decode`?

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
rukkal: We did consider re-using base::StatusOr, but we went for protovm::StatusOr as we really need to represent three states (ok, error, abort) instead of just two.

src/trace_processor/importers/proto/proto_trace_reader.cc
rukkal: > when the transform happens we'll have to update the decoder

Just realized that once the ProtoVmModule does the tokenization the decoder is not used anymore.

I think we'll still need to update the `packet` parameter though:
```
base::Status ProtoTraceReader::TimestampTokenizeAndPushToSorter(TraceBlobView packet)
```

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
LalitMaganti: Ok but then I expect to see a converter between protovm and bases status or?

src/trace_processor/importers/proto/proto_trace_reader.cc
LalitMaganti: I'm lost as to why you need to touch ProtoTraceReader::TimestampTokenizeAndPushToSorter. If you look at other modules, you see they directly push to sorter. Whu can we not do the same?

rukkal: ah found it! Man I didn't know that `module_context_->trace_packet_stream->Push()` was a thing in a few modules

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
rukkal: Done the refactoring on the ProtoVM side. Also available in the PR if you want to merge it separate:

https://github.com/google/perfetto/pull/4816

src/trace_processor/importers/proto/proto_trace_reader.cc
rukkal: Done

src/trace_processor/importers/proto/proto_trace_reader.cc
rukkal: Changed to `trace_context_->import_logs_tracker->RecordTokenizationError(stats::protovm_abort, packet->offset())`

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
rukkal: Now the protovm::StatusOr remains isolated within ProtoVmModule

LalitMaganti:
Let's wait for https://github.com/google/perfetto/pull/4816 to be merged and rebase

rukkal:
> Let's wait for #4816 to be merged and rebase

Done

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
rukkal: Replaced with a single `base::FlatHashMap<uint32_t, std::vector<protovm::Vm*>> sequence_id_to_vms_`

src/trace_processor/importers/proto/default_modules.cc
```diff
@@ -49,6 +50,8 @@ void RegisterDefaultModules(ProtoImporterModuleContext* module_context,
       new ChromeSystemProbesModule(module_context, context));
   module_context->modules.emplace_back(
       new MetadataMinimalModule(module_context, context));
+  module_context->modules.emplace_back(
```
LalitMaganti: This should be in additional_modules not here to prevent bloating Chromium binary size.

src/trace_processor/importers/proto/protovm_module.cc
```diff
@@ -0,0 +1,147 @@
...
+
+namespace perfetto {
+namespace trace_processor {
+
+ProtoVmModule::ProtoVmModule(ProtoImporterModuleContext* context,
+                             TraceProcessorContext* trace_context)
+    : ProtoImporterModule(context), trace_context_(trace_context) {
+  RegisterForField(protos::pbzero::TracePacket::kTraceProvenanceFieldNumber);
+  RegisterForField(protos::pbzero::TracePacket::kProtovmsFieldNumber);
+  RegisterForField(
```
LalitMaganti: Why this? This feels like a signifcant layer violation.

If we need to work on this, the correct way to layer this is to have a protovm_tracker which is called by both protovm module *and* the winscope module.

rukkal: > This feels like a signifcant layer violation.

What do you mean? Are you expecting all the packets to be passed down to ProtoVmModule? Here my assumption was that you wanted to minimizet the "passing down"

> a protovm_tracker which is called by both protovm module and the winscope module.

No idea what that protovm_tracker would be. Well, I guess it's something that gemini can decode for me :)

LalitMaganti: > What do you mean? Are you expecting all the packets to be passed down to ProtoVmModule? Here my assumption was that you wanted to minimizet the "passing down"

I don't understand why the protovm module needs to know at all about surfaceflinger layer protos. That feels quite confusing no? What ties layers with protovm? Yes you happen to use ProtoVM for surfaceflinger layers but that's just how you *happen* to use it - nothing about protovm restricts it to happen for those packets.

The only code which should use or know about surfaceflinger protos is the winscope module.

> No idea what that protovm_tracker would be. Well, I guess it's something that gemini can decode for me :)

Well you need some code which both the protovm module and the winscope module can talk to such that there's a "shared understanding" of how the protos should be modified. The general pattern for this in trace processor is by introducing a new shared "tracker" which both modules can talk to. Basically take all the stateful logic in this module and extract it into a separate class (as modules themselves cannot be referenced by other modules, that's a layering problem if they do).

rukkal: That's correct, ProtoVmModule doesn't need to know anything about surfaceflinger or other winscope protos.

We could replace the
```
RegisterForField(protos::pbzero::TracePacket::kSurfaceflingerTransactionsFieldNumber);
```
with something lke
```
// just process any packet, including those that we already know the ProtoVMs don't care about
RegisterForField(protos::pbzero::TracePacket::kTrustedPacketSequenceIdFieldNumber);
```
and stuff would work.

Shall we go with that "shared tracker" approach instead?

LalitMaganti: > We could replace the

Yes we could but then *every* packet will end up in this code and this code becomes a lot more perf sensitive. Potentially at that state, we might need to add short-circuiting remove the module entirely and do this in ProtoTraceReader as before to avoid the virtual function call

However, right now I don't want to think in those terms because I'm very sure Chrome will start complaining about binary size if we start including code in core trace processor we don't need for JSON conversion. So I would prefer we go with the tracker approach. That keeps things flexible and localizes the perf impact to Winscope only.

src/trace_processor/types/trace_processor_context.h
```diff
@@ -148,6 +148,7 @@ class TraceProcessorContext {
   GlobalPtr<StackProfileTracker> stack_profile_tracker;
   GlobalPtr<Destructible> deobfuscation_tracker;  // DeobfuscationTracker
   GlobalPtr<BlobPacketWriter> blob_packet_writer;
+  GlobalPtr<Destructible> protovm_tracker;  // ProtoVmTracker
```
LalitMaganti: This should be PerTraceAndMachine not global.

src/trace_processor/importers/proto/BUILD.gn
```diff
@@ -56,6 +56,10 @@ source_set("minimal") {
     "proto_trace_reader.h",
     "proto_trace_tokenizer.cc",
     "proto_trace_tokenizer.h",
+    "protovm_module.cc",
```
LalitMaganti: This should be in the full target not minimal.

src/trace_processor/importers/proto/protovm_tracker.cc
```diff
@@ -0,0 +1,122 @@
...
+    RefPtr<PacketSequenceStateGeneration> sequence_state) const {
+  protozero::HeapBuffered<protos::pbzero::TracePacket> proto;
+  vm.SerializeIncrementalState(proto.get());
+  proto->set_trusted_uid(patch.trusted_uid());
+  proto->set_trusted_pid(patch.trusted_pid());
+  proto->set_trusted_packet_sequence_id(patch.trusted_packet_sequence_id());
+  auto [data, size] = proto.SerializeAsUniquePtr();
+  auto blob = TraceBlob::TakeOwnership(std::move(data), size);
+
+  uint64_t timestamp;
```
rukkal: @LalitMaganti  I suspect you'll have questions about this. FYI here we read the timestamp from the serialized incremental state, so that we give a ProtoVM's program also the power to determine timestamps. Primiano wanted this feature so you might want to chat directly with him. Then we\ can always make the implementation below more efficient avoiding to construct a full TracePacket::Decoder.

src/trace_processor/importers/proto/winscope/winscope_module.cc
```diff
@@ -68,11 +70,22 @@ WinscopeModule::WinscopeModule(ProtoImporterModuleContext* module_context,
...
-    RefPtr<PacketSequenceStateGeneration> /*state*/,
+    RefPtr<PacketSequenceStateGeneration> sequence_state,
     uint32_t field_id) {
   switch (field_id) {
+    case TracePacket::kSurfaceflingerTransactionsFieldNumber: {
+      std::optional<ProtoVmTracker::SerializedIncrementalState> state =
+          protovm_tracker_->TryProcessPatch(decoder, *packet,
+                                            std::move(sequence_state));
+      if (state) {
+        module_context_->trace_packet_stream->Push(state->timestamp,
```
rukkal: @LalitMaganti any chance we can move the `trace_packet_stream->Push()` into ProtoVmTracker? A trace_packet_stream doesn't seem to be available from the tracker and not sure if exposing it breaks architectural boundaries

src/trace_processor/importers/proto/protovm_tracker.cc
```diff
@@ -0,0 +1,122 @@
...
+  return std::nullopt;
+}
+
+ProtoVmTracker::SerializedIncrementalState
+ProtoVmTracker::SerializeIncrementalState(
+    const protovm::Vm& vm,
+    const protos::pbzero::TracePacket::Decoder& patch,
+    const TraceBlobView& packet,
+    RefPtr<PacketSequenceStateGeneration> sequence_state) const {
+  protozero::HeapBuffered<protos::pbzero::TracePacket> proto;
```
LalitMaganti: Please use the new class I introduced for you - https://github.com/google/perfetto/pull/4799

src/trace_processor/importers/proto/protovm_tracker.cc
```diff
@@ -0,0 +1,122 @@
...
+}
+
+void ProtoVmTracker::AddProtoVm(std::unique_ptr<protovm::Vm> vm,
+                                const std::vector<int32_t>& producer_ids) {
+  protovm::Vm* p = vm.get();
+  vms_.push_back(std::move(vm));
+  for (int32_t producer_id : producer_ids) {
+    auto* sequence_ids = producer_id_to_sequence_ids_.Find(producer_id);
+    if (!sequence_ids) {
+      context_->storage->IncrementStats(stats::protovm_registration_error);
```
LalitMaganti: please use import log tracker rather than stats. Here and throughout.

src/trace_processor/importers/proto/protovm_tracker.cc
LalitMaganti: The problem is that by doing this, your logic is now inconsistent with the protoTraceReader. I wasn't aware that you were planning on rewriting the full trace packet otherwise I wouldn't have told you to make a module at all.

In this case I actually think you have no choice but to do this very very proactively, basically at the very start of ProtoTraceReader, before *any* of the logic of that function. Otherwise, this code will drift between the (very complicted) timestamp logic which exists in [1]

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/importers/proto/proto_trace_reader.cc

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -236,6 +237,18 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
...
+    protovm_.ProcessTraceProvenancePacket(decoder.trace_provenance());
+  } else if (decoder.has_protovms()) {
+    protovm_.ProcessProtoVmsPacket(decoder.protovms());
+  } else if (auto new_packet = protovm_.TryProcessPatch(packet); new_packet) {
+    packet = TraceBlobView(std::move(*new_packet));
+    // TODO(keanmariotti): fix this hack
+    decoder.protos::pbzero::TracePacket::Decoder::~Decoder();
+    new (&decoder)
+        protos::pbzero::TracePacket::Decoder(packet.data(), packet.length());
```
rukkal, line 250: @LalitMaganti what about extending decoder's interface to fix this hack? Either an assignment operator or a Reset() member function

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -236,6 +237,18 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   PERFETTO_DCHECK(decoder.has_machine_id() ==
                   (context_->machine_id() != MachineId(kDefaultMachineId)));
 
+  if (decoder.has_trace_provenance()) {
```
LalitMaganti, line 242: This needs to all happen even before machine id right? In theoery this should always be handled on the root level eben before the above.

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,109 @@
...
+      for (auto sequence_id : *sequence_ids) {
+        sequence_id_to_vms_[sequence_id].push_back(vm);
+      }
+    }
+  }
+}
+
+std::optional<TraceBlob> ProtoVmIncrementalTracing::TryProcessPatch(
+    const TraceBlobView& blob) {
+  protos::pbzero::TracePacket::Decoder patch(blob.data(), blob.size());
```
LalitMaganti: Why are we creating *another* decoder here? Why not just pass it in?

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,109 @@
...
+                                      ? instance.state()
+                                      : protozero::ConstBytes{nullptr, 0};
+    vms_.push_back(std::make_unique<protovm::Vm>(
+        instance.program(), 1024 * instance.memory_limit_kb(), state));
+    protovm::Vm* vm = vms_.back().get();
+    for (auto producer_id = instance.producer_id(); producer_id;
+         ++producer_id) {
+      auto* sequence_ids = producer_id_to_sequence_ids_.Find(*producer_id);
+      if (!sequence_ids) {
+        context_->storage->IncrementStats(stats::protovm_registration_error);
```
LalitMaganti: Please use import logging.

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
```diff
@@ -0,0 +1,109 @@
...
+      context_->storage->IncrementStats(stats::protovm_abort);
+    }
+  }
+  return std::nullopt;
+}
+
+TraceBlob ProtoVmIncrementalTracing::SerializeIncrementalState(
+    const protovm::Vm& vm,
+    const protos::pbzero::TracePacket::Decoder& patch) const {
+  protozero::HeapBuffered<protos::pbzero::TracePacket> proto;
```
LalitMaganti: Again please use the BlobPacketWriter to avoid copies.

src/trace_processor/importers/proto/proto_trace_reader.cc
LalitMaganti: Let's add the move assignment operator.

src/trace_processor/importers/proto/proto_trace_reader.cc
rukkal: uhm I am a bit lost on how multiple-machine traces are produced. If with multiple machines we have only one TraceProvenance packet and one ProtoVms packet (because there is only one traced instance that emits them), then yes, I should move the protovm stuff before machine id.

LalitMaganti: Well it's possible for there to be multiple traced instances which are being cominbed into one fianl trace file. This is how VM tracing works. It really depends on whether you can affect machine_id via a VM trace. You also seem to have a dependency on trusted_packet_sequence_id not being modifuable.

I suggest defining in the protovm documentation which fields are eligible to be targetted via protovm and which are not. Will make resolving things like this a lot easier.

src/trace_processor/importers/proto/protovm_tracker.cc
rukkal: Nice. I was going to ask if we wanted to get rid of that allocation as well. Done

src/trace_processor/importers/proto/proto_trace_reader.cc
rukkal: Done

src/trace_processor/importers/proto/protovm_incremental_tracing.cc
rukkal: is it still a tokenization error at this stage?

rukkal: Done with `context_->import_logs_tracker->RecordTokenizationError(...)` for now

include/perfetto/protozero/proto_decoder.h
```diff
@@ -532,6 +532,19 @@ class TypedProtoDecoder : public TypedProtoDecoderBase {
 
   TypedProtoDecoder(TypedProtoDecoder&& other) noexcept
       : TypedProtoDecoderBase(std::move(other)) {
+    MaybeCopyOnStackStorage(other);
+  }
+
+  TypedProtoDecoder& operator=(TypedProtoDecoder&& other) noexcept {
+    if (this != &other) {
+      TypedProtoDecoderBase::operator=(std::move(other));
```
LalitMaganti, line 540: Uhhhh isn't this infinite recursion?

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -212,6 +213,17 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   // Any compressed packets should have been handled by the tokenizer.
   PERFETTO_CHECK(!decoder.has_compressed_packets());
 
+  if (decoder.has_trace_provenance()) {
```
LalitMaganti: Do we have a resolution on whether this should be before or after machine id? Is there some doc etc we can point to in a comment about why the order is the way it is?

rukkal: I suspect we'll have to move it again after machine id, but not sure. Have a look at go/perfetto-proto-vm (TP integration tab) and feel free to edit. I just drafted something there, including the list of fields controllable by ProtoVM.

include/perfetto/protozero/proto_decoder.h
rukkal: Why? It is calling the assignment operator in the base class

src/trace_processor/importers/proto/proto_trace_reader.cc
LalitMaganti: Yeah let's move it after. I had a look, I think machine_id should be treated as "trusted"

include/perfetto/protozero/proto_decoder.h
LalitMaganti: Ahhh I misread sorry.

src/trace_processor/importers/proto/proto_trace_reader.cc
rukkal: Done. And the full-state packets now inherits machine_id from the patch.

PR #4769, opened 2026-02-11
ui: Add android monitor lock contention plugin

ivan-chong:
# Android Lock Contention Plugin

This PR introduces the `com.android.AndroidLockContention` plugin, which allows navigation between events contending for a lock, and presents data about the events in a tabular format.

### Features

-   **Analysis Tab:** Displays detailed information about selected lock contention events:
    -   **Blocked Thread (Victim):** Thread name, TID, main thread status, method, and source location.
    -   **Contention Details:** Lock name, duration, monotonic duration, and a list of other waiters.
    -   **Blocking Thread (Culprit):** Thread name, TID, main thread status, method, and source location.
    -   **Blocking Thread States:** Table summarizing CPU states.
    -   **Kernel Functions:** Table summarizing time spent in kernel functions (if applicable).
    -   **Binder IPC Support:** Visualizes inbound/outbound Binder transactions related to the contention.
    -   **Nested Contention Support:** Displays a warning if the thread holding the lock is also blocked by another lock, with a shortcut to navigate to the root cause.
-   **Navigation History:** Supports navigating backward and forward through inspected contention events with shortcuts.

### Usage

1.  Run the command **"Android Lock Contention: Go Forward"** (or use shortcut `Ctrl+]`) to start navigating.
2.  Navigate backward using shortcut `Ctrl+[`.
3.  Inspect details in the **"Lock Contention Analysis"** tab.

Bug: 289343169

### DIFF
--- a/ui/src/plugins/com.android.AndroidLockContention/android_lock_contention_event_source.ts
+++ b/ui/src/plugins/com.android.AndroidLockContention/android_lock_contention_event_source.ts
@@ -1,0 +1,280 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {QueryResult, QuerySlot, SerialTaskQueue} from '../../base/query_slot';
+import {duration, Duration, time, Time} from '../../base/time';
+import {getTrackUriForTrackId} from '../../components/related_events/utils';
+import {Trace} from '../../public/trace';
+import {
+  LONG,
+  LONG_NULL,
+  NUM,
+  NUM_NULL,
+  STR,
+  STR_NULL,
+} from '../../trace_processor/query_result';
+
+export interface ContentionState {
+  state: string;
+  dur: duration;
+  count: number;
+}
+
+export interface ContentionBlockedFunction {
+  func: string;
+  dur: duration;
+  count: number;
+}
+
+export interface ContentionWaiter {
+  threadName: string;
+  tid: number | null;
+  eventId: number;
+}
+
+export interface LockContentionDetails {
+  id: number;
+  ts: time;
+  dur: duration | null;
+  monotonicDur: duration | null;
+  waiterCount: number;
+
+  lockName: string;
+
+  blockedThreadName: string;
+  blockedThreadTid: number | null;
+  isBlockedThreadMain: boolean;
+  blockedMethod: string;
+  blockedSrc: string;
+
+  blockingThreadName: string;
+  blockingThreadTid: number | null;
+  isBlockingThreadMain: boolean;
+  blockingMethod: string;
+  blockingSrc: string;
+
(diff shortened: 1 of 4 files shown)


ui/src/plugins/com.android.AndroidLockContention/android_lock_contention_event_source.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    const events: RelatedEvent[] = [];
+
+    if (!it.valid()) {
+      return {events: [], relations: []};
+    }
+
+    const blockedEventId = it.contention_id!;
+    const blockedTs = BigInt(it.contention_ts!) as time;
```
stevegolton: Use LONG_NULL above and you can avoid the bigint conversion. Also the ! at the end is dangerous and can hide bugs. if you're sure it shouldn't be null the nuse LONG rather than LONG_NULL. Also use Time.fromRaw() instead of the type assertion.

ui/src/plugins/com.android.AndroidLockContention/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    const tab = new AndroidLockContentionTab({trace, source});
+
+    trace.tabs.registerTab({
+      uri: 'com.android.AndroidLockContentionTab',
+      isEphemeral: false,
+      content: tab,
+    });
+
+    trace.commands.registerCommand({
+      id: 'toggleContentionNavigation',
```
stevegolton: com.android.ToggleContentionNavigation

ui/src/plugins/com.android.AndroidLockContention/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      name: 'Android Lock Contention: Toggle Blocked/Blocking Slice',
+      defaultHotkey: ']',
```
stevegolton: I think ] already maps to 'move to next flow event'. Not sure if this collides.

ui/src/plugins/com.android.AndroidLockContention/tab.ts
```diff
@@ -0,0 +1,207 @@
...
+  source: AndroidLockContentionEventSource;
+}
+
+export class AndroidLockContentionTab implements Tab {
+  private event: RelatedEvent | null = null;
+  private isLoading = false;
+
+  constructor(private config: LockContentionTabConfig) {}
+
+  syncSelection() {
```
stevegolton: Who's calling this?

ui/src/plugins/com.android.AndroidLockContention/tab.ts
```diff
@@ -0,0 +1,207 @@
...
+    const selection = this.config.trace.selection.selection;
+    if (selection.kind === 'track_event') {
+      this.loadData(selection.eventId);
+    } else {
+      this.event = null;
+      m.redraw();
```
stevegolton: You typically don't need m.redraw()s unless you're doing something out of the ordinary like setTimeout, fetch, yield. Mithril will trigger a redraw automatically after DOM events including command callbacks.

stevegolton (changes requested):

ui/src/plugins/com.android.AndroidLockContention/index.ts
ivan-chong: Replaced with 'Ctrl+]', wdyt?

ui/src/plugins/com.android.AndroidLockContention/tab.ts
ivan-chong: This was previously used by a base class for the tab, but I have replaced the inheritance with composition so this is no longer necessary

ui/src/plugins/com.android.AndroidLockContention/index.ts
stevegolton: Sounds good to me

PR #4770, opened 2026-02-11
Add Proto for ProcState events

parthsane:
This change introduces a new proto message, AndroidProcState, for perfetto tracing of process state.

This will help understand memory usage.

### DIFF
--- a/protos/perfetto/trace/android/android_track_event.proto
+++ b/protos/perfetto/trace/android/android_track_event.proto
@@ -33,6 +33,14 @@ message AndroidMessageQueue {
   optional uint64 message_delay_ms = 4;
 }
 
+// Information about ProcState of process in the system
+message ProcState {
+  // Name of the Process State.
+  optional string proc_state = 1;
+  // Count of processes in the proc_state
+  optional int32 count_of_processes = 2;
+}
+
 message AndroidJobSchedulerJob {
   // Job id.
   optional int32 job_id = 1;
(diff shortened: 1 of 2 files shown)


protos/perfetto/trace/android/android_track_event.proto
```diff
@@ -33,6 +33,14 @@ message AndroidMessageQueue {
   optional uint64 message_delay_ms = 4;
 }
 
+// Information about ProcState of process in the system
+message ProcState {
```
LalitMaganti, line 37: This is an *ultra* generic name. Please make this more specific. At the very least adding an android prefix but I would consider making this even more specific (because it's some sort of aggregate counter *not* actually an event on every proc state).

Also why are we even having this proto at all? IMO this should be a set of CounterTrack track events which is going to be both far more efficient *and* will remove the need to do de-interning in the trace processor.

parthsane: I wasn't sure if this was the correct way to do this,  or if this was necessary to do.

I have implemented this as a counter track in the code. If this is a counter track, then does it not need this proto change? I can close the PR then.

LalitMaganti: If you can do it with Counter tracks, you should. @zezeozue had exactly this convo in https://github.com/google/perfetto/pull/4795 but the opposite was reached because it wasn't possible to emit in one shot.

PR #4772, opened 2026-02-12
ui: Add auth support and refactor extension servers

LalitMaganti:
- Add GitHub PAT auth for private repos
- Refactor server config from flat url-based schema to discriminated
  union (github vs https) with per-type auth
- Switch GitHub fetches to Contents API unconditionally
- Centralize URL utilities into url_utils.ts
- Use Form/FormSection widgets in add server modal
- Add tests for buildFetchRequest and buildAuthConfig


LalitMaganti:
@primiano FYI only.

PR #4774, opened 2026-02-12
ui: add link sharing for server extensions

LalitMaganti:
This CL adds link sharing support for server extensions making it very
easy people to share these around and add these to their UI in just one
click.


stevegolton (approved):
LGTM! Sharing settings via the url args sounds entirely reasonable to me!

Issue #4547, opened 2026-01-23
Request: Allow any origin for websocket_bridge


helo-xperi:
Hi.

I noticed that websocket_bridge in Perfetto uses `AddAllowedOrigin()` to restrict access to specific origins:

```cpp
srv.AddAllowedOrigin("http://localhost:10000");
srv.AddAllowedOrigin("http://127.0.0.1:10000");
srv.AddAllowedOrigin("https://ui.perfetto.dev");
```

However, the origin of a HTTP request can be spoofed in multiple ways:

1. **Non-browser clients:** Tools like curl or custom scripts can send arbitrary Origin.
2. **Browser-based spoofing:** By manipulating DNS resolution in the client (e.g., pointing `localhost` or `ui.perfetto.dev` to another server) and for HTTPS origins disable certificate validation or use a self-signed certificate where the CA is trusted inside the browser.

What I think this means is that the origin check in `websocket_bridge` provides limited security guarantees for the system running `websocket_bridge`, even against browser-based attacks, since a malicious client can bypass it entirely by crafting the request as it sees fit. Unlike CORS headers that are intended to protect non-malicious clients by informing the client of any allowed origins - this seems to want to protect the host.

I have two questions:

1. What is the threat model that this access control avoids?
2. Request: Is it possible to change the behaviour of `websocket_bridge` to allow any HTTP origin by default?

The reason for this request is to allow a Perfetto UI instances hosted inside a corporate network to connect to `websocket_bridge` without modifying the source code.

primiano:
The two spoof scenarios you listed are not the concern.

The real concern is that you visit a random malicious website, and that website tries connects to the websocket bridge and takes over of the device connected.

If you assume localhost access (e.g. a non-browser) there is no need to use websocket bridge, just talk directly to ADB on localhost:5037. websocket_bridge will not give you any gain if you can run malicious/untrusted code on localhost.

So the deal is "we trust ui.perfetto.dev (and localhost for development) but we cannot trust the whole internet"

> Request: Is it possible to change the behaviour of websocket_bridge to allow any HTTP origin by default?

No absolutely not. that's a big security risk. a random website on the internet could run arbitrary commands on your android device and install malware on it.

If you want you can send a CL that takes an optional argument --http-additional-cors-origins similar to what traceprocessor does.

helo-xperi:
> there is no need to use websocket bridge, just talk directly to ADB 

[...]

PR #4776, opened 2026-02-12
 Changes to websocket_bridge+trace_processor_shell to support using non-default WebUI origin

helo-xperi:
This set of patches implements changes suggested in #4547 to better support using websocket_bridge and trace_processor with non-default origins, that is origins that are neither of https://ui.perfetto.dev/ or http://localhost:1000/.

It does so by making websocket_bridge support --http-additional-cors-origins argument from trace_processor_shell and also adds a GN arg perfetto_http_additional_cors_origins that can be used to compile both binaries with a by-default supported origin already included.


src/tools/http_additional_cors_origins/BUILD.gn
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+declare_args() {
+  # Comma-separated list of additional CORS origins to allow in trace_processor
+  # and websocket_bridge.
+  perfetto_http_additional_cors_origins = ""
```
LalitMaganti: This needs to be in https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/gn/BUILD.gn as part of gen_buildflags not adhoc like this.

helo-xperi: Hi. 

Thanks for your review. I'm sorry for providing a late response, it seems the notification got lost in my email flow and I just noted it now when I was looking to "ping" the ticket of this PR. 

How do we want to add this to `gen_buildflags`? From what I can tell the other buildflags in that block are of true/false nature and not a comma-separated list of strings?

LalitMaganti: Well gen_buildflags allows you to put anything you want there. So there's nothing stopping you generating a comma separated list of strings if you wanted!

helo-xperi: Thanks. I added it to BUILDFLAGS. Not familiar with how those work, PTAL!

src/tools/http_additional_cors_origins/http_additional_cors_origins.cc
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#include "src/tools/http_additional_cors_origins/http_additional_cors_origins.h"
+
+#include "perfetto/base/build_config.h"
+#include "perfetto/ext/base/string_utils.h"
+
+namespace perfetto {
+
+std::vector<std::string> GetHttpAdditionalCorsOrigins() {
```
LalitMaganti: Two problems:
1) given how small this function is I don't think it'd worth having two files and a build rule for this. Please inline 
2) even if we *did" have an abstraction tools/ is definitely the wrong place for this. Core libraries should not have a decency on tool. Likely this would go in base but realistically it should just be inclined.

helo-xperi: Updated according to suggestion, PTAL!

PR #4777, opened 2026-02-12
protovm: fix chromium build

rukkal:


include/perfetto/base/BUILD.gn
```diff
@@ -20,6 +20,8 @@ source_set("base") {
     "compiler.h",
     "export.h",
     "flat_set.h",
+    "intrusive_list.h",
```
LalitMaganti: Please put this in ext/base not base. This will become part of the public API for Perfetto if it's in base which we definitely don't want.

rukkal: Got it. Moved to ext/base

PR #4779, opened 2026-02-12
sdk: Add Android process lifecycle and broadcast track events

zezeozue:
AndroidTrackEvents to capture:
1. process starts
2. process deaths
3. process state changes
4. broadcast deliveries
5. freezer transitions

We already have jobs. And we'll add service bindings in a follow up.

We capture process names at start and death to allow TraceProcessor detect PID reuse without the overhead of ftrace events and complexity of interning. Other events will be keyed by pid. This is a pragmatic first provides the equivalent of task_new_task and sched_process_free for managed Java processes.

Also enabled javastream generation for these protos so the Android framework can use generated constants instead of hard-coded IDs.

Test: Manual with ag/38360511


zezeozue:
Trace file example: https://ui.perfetto.dev/#!/?s=2e4834ab27c0f4bfc93b1f48b225bc811f1a66c9 (just 3mins)
Proto profile: https://pprof.corp.google.com/?id=6715fb48d583b13260f74520bf0aa9be&tab=flame&source=cli

PR #4781, opened 2026-02-12
tp: Rename trace_config_utils to proto_utils and extend

aMayzner:
Renames trace_config_utils to proto_utils to better reflect its
expanded purpose as a general proto-to-textproto conversion utility.

Extends the utility to support TraceSummarySpec and TraceMetricV2Spec
in addition to the existing TraceConfig support. Adds new C++ functions
TraceSummarySpecPbToTxt() and TraceMetricV2SpecPbToTxt() following the
same pattern as TraceConfigPbToTxt().

The TypeScript/WASM wrapper is moved from the RecordTraceV2 plugin to
ui/src/base/ to make it accessible across the entire UI codebase.

All changes maintain backwards compatibility for existing TraceConfig
functionality.


LalitMaganti:
Would suggest calling this proto_utils to allow for other proto functions to be put here (e.g. conversion to JSON)

LalitMaganti:
Would suggest calling this proto_utils to allow for other proto functions to be put here (e.g. conversion to JSON)

LalitMaganti:
Would suggest calling this proto_utils to allow for other proto functions to be put here (e.g. conversion to JSON)

aMayzner:
> Would suggest calling this proto_utils to allow for other proto functions to be put here (e.g. conversion to JSON)

Renamed

src/proto_utils/pb_to_txt.cc
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "src/proto_utils/pb_to_txt.h"
+
+#include <string>
+
+#include "src/proto_utils/config.descriptor.h"
+#include "src/proto_utils/trace_summary.descriptor.h"
+#include "src/trace_processor/util/descriptors.h"
```
LalitMaganti, line 22: Sorry but no you should not have that this module dpeends on trace processor. The dependency should be reversed: trace procesosr should depend on this just like how it works 

Please always check that your architecture make sense and that a "more specific" module doesn't depend on a "more generic" module.

src/proto_utils/wasm.cc
```diff
@@ -77,6 +78,17 @@ uint32_t trace_config_txt_to_pb(uint32_t size) {
   memcpy(&wasm_buf[1], res->data(), resp_size);
   return static_cast<uint32_t>(resp_size);
 }
+
+uint32_t EMSCRIPTEN_KEEPALIVE trace_summary_spec_to_text(uint32_t size);
+uint32_t trace_summary_spec_to_text(uint32_t size) {
+  return pb_to_txt(perfetto::TraceSummarySpecPbToTxt, size);
+}
+
+uint32_t EMSCRIPTEN_KEEPALIVE trace_metric_v2_spec_to_text(uint32_t size);
```
LalitMaganti: Why do you need both this and above? Doesn't make sense to me.

src/proto_utils/pb_to_txt.cc
aMayzner: trace_config_utils already depended on trace processor. This is a significant refactor. I will do it in a separate PR.

aMayzner: As discussed offline - it requires more thoughtful untangling, I will submit it as is.

src/proto_utils/wasm.cc
aMayzner: I want to be able to only decode metric and trace summary. They share the descriptor (because they are nested), but I would like to have separate decoders.

aMayzner: Ok, I removed the metrics. Considering that some metrics require other queries to run, it is a better idea to always talk about trace summary

PR #4795, opened 2026-02-13
sdk: Add the process state changed sequence id proto

zezeozue:


protos/perfetto/trace/android/android_track_event.proto
```diff
@@ -630,6 +630,9 @@ message AndroidProcessStateChangedEvent {
 
   // Reason for the process state change.
   optional OomChangeReasonEnum reason = 9;
+
+  // Sequence id to identify all changes in one event.
+  optional int64 seq_id = 10;
```
LalitMaganti, line 637: I don't understand the semantic of this. Why do we need this?

zezeozue: This is an id from oomadjuster: https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:frameworks/base/services/core/java/com/android/server/am/psc/OomAdjuster.java;l=284?q=oomadjuster.java&sq=repo:googleplex-android%2Fplatform%2Fsuperproject%2Fmain%20b:main%20case:no

It's a bit easier to write queries when we know all the process states that changed due to one oom adjust event.

LalitMaganti: Would it be possible to just emit it as one repeated field in that case? Or does thi shappen in different stages?

At the very least would suggest linking to the docs etc (note that ideally the link would be on cs.android.com).

zezeozue: We emit each process's state change independently. We could consider having one proto for all the process state changes but that will require a bit more plumbing on the platform side. Added the doc link

PR #4800, opened 2026-02-13
Fix Windows compilation and enable libperfetto_c on Windows

jgaillard85:
- Update gen_android_bp to enable Windows build of libperfetto_c in Android.
- Adjust pid_t and ssize_t for Windows in sys_types.h to match expected types (64-bit on Win64).
- Fix SetConsoleCtrlHandler compilation error by using a WINAPI trampoline function instead of a lambda.

### DIFF
--- a/src/base/ctrl_c_handler.cc
+++ b/src/base/ctrl_c_handler.cc
@@ -51,7 +51,7 @@ void InstallCtrlCHandler(CtrlCHandlerFunction handler) {
   g_handler = handler;
 
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
-  ::SetConsoleCtrlHandler(Trampoline, TRUE);
+  ::SetConsoleCtrlHandler(Trampoline, true);
 #elif PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX) || \
     PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID) || \
     PERFETTO_BUILDFLAG(PERFETTO_OS_APPLE)
--- a/Android.bp
+++ b/Android.bp
@@ -1033,6 +1033,16 @@ cc_library {
                 "perfetto_flags_c_lib",
             ],
         },
+        windows: {
+            enabled: true,
+            cflags: [
+                "-Wno-shift-count-overflow",
+                "-Wno-unknown-pragmas",
+            ],
+            host_ldlibs: [
+                "-lws2_32",
+            ],
+        },
     },
 }
 


tools/gen_android_bp
```diff
@@ -333,6 +333,14 @@ additional_args = {
         ('min_sdk_version', '30'),
         ('export_include_dirs', {'include'}),
         ('cflags', {'-DPERFETTO_SHLIB_SDK_IMPLEMENTATION'}),
+        ('windows', {
```
LalitMaganti, line 336: I expect to see a change to Andorid.bp here which apepars to be missing?

src/base/ctrl_c_handler.cc
```diff
@@ -34,21 +34,24 @@ namespace base {
...
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
-  auto trampoline = [](DWORD type) -> int {
-    if (type == CTRL_C_EVENT) {
-      g_handler();
-      return true;
-    }
-    return false;
-  };
-  ::SetConsoleCtrlHandler(trampoline, true);
+  ::SetConsoleCtrlHandler(Trampoline, TRUE);
```
LalitMaganti: Do we really need to change true -> TRUE?

include/perfetto/ext/base/sys_types.h
```diff
@@ -31,13 +31,19 @@
...
 using pid_t = int;
+#endif  // _WIN64
+
 #endif  // !GCC
 
 #if defined(_WIN64)
 using ssize_t = int64_t;
 #else
-using ssize_t = long;
+using ssize_t = int;
```
primiano, line 46: wait why? this change is not obvious. can you explain?

primiano:
A more general comment: can I ask what this is for? using the C sdk on windows puts you in very uncharted territory: it's a combination nobody ever tested (and probably nobody else ever will). Expect dragons ;)

include/perfetto/ext/base/sys_types.h
jgaillard85: This is to make it consistent with the definitions from Mingw (that are present in the Android codebase)

jgaillard85:
> A more general comment: can I ask what this is for? using the C sdk on windows puts you in very uncharted territory: it's a combination nobody ever tested (and probably nobody else ever will). Expect dragons ;)

This is part of an effort to build parts of Android to run on desktop (including Windows). The main application is rendering Android graphics in Android dev tools such as Android Studio or enable emulator-free screenshot testing of Android apps. Perfetto is a dependency of some graphics libraries, and enabling the Windows build would reduce the maintenance burden of a separate fork for Windows.
Though we only ship a 64-bit version of the libraries, the Android build system still supports the 32-bit targets, hence my need to make the type definitions consistent in both 32 and 64 bit.

tools/gen_android_bp
jgaillard85: I thought it was done automatically when imported in AOSP. My bad. This has been updated

src/base/ctrl_c_handler.cc
jgaillard85: It's not strictly required, I reverted it

PR #4803, opened 2026-02-13
sdk: Accept a callstack in the track event sdk instead of dumping

zezeozue:
This allows callers to pass cached stacks. In this case, drop the 'expensive' from the name since the perfetto API is no longer responsible for dumping.

### DIFF
--- a/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
+++ b/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
@@ -304,7 +304,7 @@ public final class PerfettoTrace {
    * A stack can be captured with Thread.getStackTrace() but note that it is expensive
    * and should only be used for local debugging or low-frequency diagnostic events.
    */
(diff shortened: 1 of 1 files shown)


src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -300,25 +300,22 @@ public static void registerWithDebugChecks(boolean isBackendInProcess) {
   }
 
   /**
-   * Captures and emits the current thread's call stack to the Perfetto trace.
-   * * WARNING: This is an expensive operation. Thread.getStackTrace() requires
-   * a full stack walk and symbol resolution. Use this ONLY for local debugging
-   * or low-frequency diagnostic events. Do not use in production hot paths.
```
LalitMaganti, line 306: I think we should still call this function expensive. Sorry but we should not encourage use of this API. I am very concerned people will misuse this.

zezeozue: Sure

PR #4811, opened 2026-02-15
traced_relay: refactor traced_relay to extract library target

dreveman:
Extract a thin wrapper interface for traced_relay which allows
it to be included in tracebox and similar binaries layered on
top of perfetto without duplicating large amounts of code.


LalitMaganti:
Thanks for the contribution! Would it be possible to split the pure refactoring change from the change to add new functionality? This code is now starting to be used inside core Android so we need to be a bit more careful with  changes here than we were in the past.

dreveman:
> Thanks for the contribution! Would it be possible to split the pure refactoring change from the change to add new functionality? This code is now starting to be used inside core Android so we need to be a bit more careful with changes here than we were in the past.

Done. This is now limited to the refactor. I'll upload the other changes as a different PR.

PR #4814, opened 2026-02-16
ui: MemoryViz add new mem tracks

gignat-dev:
This change adds new memory counter tracks to the MemoryViz plugin: mem.heap ('Heap size (KB)'), mem.rss.shmem, mem.dmabuf_rss, mem.locked and 'GPU Memory'.

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/utils.h
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/utils.h
@@ -332,10 +332,8 @@ struct Regexp : public sqlite::Function<Regexp> {
       }
       return sqlite::result::Long(ctx, aux->Search(text));
     } else {
-      // Always-true branch to avoid spurious no-return warnings.
-      if (ctx) {
-        PERFETTO_FATAL("Regex not supported");
-      }
+      return sqlite::utils::SetError(
+          ctx, "regexp: regex is not supported on this platform");
     }
   }
 };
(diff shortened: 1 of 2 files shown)


ui/src/plugins/com.android.MemoryViz/index.ts
```diff
@@ -34,16 +34,31 @@ export default class MemoryViz implements PerfettoPlugin {
...
+          ...tracks.map(([trackName, displayName]) =>
+            this.createBreakdownTrack(ctx, window, trackName, displayName),
+          ),
         );
-        ctx.defaultWorkspace.pinnedTracksNode.addChildLast(rssFileTrack);
+        const createdTracks = await Promise.all(trackPromises);
+
+        for (const track of createdTracks) {
+          if (track) {
+            ctx.defaultWorkspace.pinnedTracksNode.addChildLast(track);
```
stevegolton, line 59: Adding tracks directly to the pinned track node is slightly unusual, I should probably not have exposed that! You can just call TrackNode.pin().

gignat-dev: For a command it seems natural to add the created tracks directly to the pinned tracks. 
IMO it seems a bit strange to create it on the normal workspace and then pin it, at least in this use case of a command (considering you may call multiple times).

Please let me know what you think.
If we have any concerns I'll follow up in the future.

PR #4815, opened 2026-02-16
ui: Fix crash in slice track tooltips when no dur present in dataset

stevegolton:
On ToT, there is a type mismatch between what TS things the type of each raw row is and the actual type. Typescript thinks each row should have `ts, dur, depth, layer, id` fields but in reality each row only has what's specified in the dataset's schema used when the track was created. This mismatch was missed by TS due to a dodgy type assertion.

Some tracks that only contain instant events (such as stack trace tracks) simply don't define `dur` in their dataset. The problem is that tooltipRenderer assumes dur is of type `bigint | null`, and so doesn't handle the `undefined` case. Typescript didn't complain due to the aforementioned type mismatch above.

This fix is just to use the track's dataset schema as the type, and handle the situation where dur is undefined in the tooltip renderer. Thus, everywhere that assumes the full type has now been reverted to simply use the dataset type (T), and the tooltipRenderer has been changed to handle the case where dur === undefined.

While in the area, minimize the type assertion and extract into a separate little function to contain the unsafe code.

Fixes: https://b.corp.google.com/issues/483051607

### DIFF
--- a/ui/src/components/tracks/slice_track.ts
+++ b/ui/src/components/tracks/slice_track.ts
@@ -855,13 +855,6 @@ export class SliceTrack<T extends RowSchema> implements TrackRenderer {
         if (task.shouldYield()) await task.yield();
       }
 
-      // Clone raw data out of the iterator
-      const row: Record<string, SqlValue> = {};
-      // eslint-disable-next-line guard-for-in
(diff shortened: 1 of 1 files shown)


stevegolton:
@rsavitski FYI

PR #4817, opened 2026-02-16
sdk: Add binder died event

zezeozue:
AppExitInfoTracker can have a 15s delay for some kills. See discussion in ag/38360511. To mitigate this, we introduce a binder died event which will be logged quite early without death reasons.

This will simplify downstream analysis, especially with pid tracking.

### DIFF
--- a/protos/perfetto/trace/android/android_track_event.proto
+++ b/protos/perfetto/trace/android/android_track_event.proto
@@ -586,7 +586,7 @@ message AndroidProcessDiedEvent {
   // Pid of the dying process.
   optional int32 pid = 2;
 
-  // The process name.
+  // The process name of the dying process.
   optional string process_name = 3;
 
   // The reason code of why the process dies.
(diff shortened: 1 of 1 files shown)


protos/perfetto/trace/android/android_track_event.proto
```diff
@@ -603,6 +605,19 @@ message AndroidProcessDiedEvent {
...
+// This is expected to come earlier than AndroidProcessDiedEvent.
+// For death reasons, see AndroidProcessDiedEvent.
+message AndroidBinderDiedEvent {
+  // Uid of the package of the dying process.
+  optional int32 uid = 1;
+
+  // Pid of the dying process.
+  optional int32 pid = 2;
+
+  // The process name.
```
LalitMaganti: Of the dying process?

protos/perfetto/trace/android/android_track_event.proto
```diff
@@ -603,6 +605,19 @@ message AndroidProcessDiedEvent {
   optional int32 has_foreground_services = 8;
 }
 
+// This is expected to come earlier than AndroidProcessDiedEvent.
+// For death reasons, see AndroidProcessDiedEvent.
+message AndroidBinderDiedEvent {
```
ilkos, line 610: Possibly name it more concretely (binder is too broad)? AndroidAppThreadDiedEvent ? And document when it is emitted in the comment?

PR #4820, opened 2026-02-17
ui: Input events overlaps visualization command 

gignat-dev:
This change adds a new command "Input Events: Visualize event overlaps (over selection)"
to the InputEvents plugin. This allows users to visualize concurrent input processing density over a selected time range.

The input event intervals are measured
from the moment of dispatch until the ACK is received (i.e. dispatch_ts, total_latency_dur).

- Registers a new visualizeOverlaps command that calculates event concurrency using the intervals_overlap_count logic.
- Adds a parent "Input Events" counter track showing total concurrency across all channels.
- Creates sub-tracks for each individual process that received input events.
- For each process, creates sub-tracks for each individual input event channel.
- Orders tracks by their peak value (descending).
- Clips intervals to the selected time window (selection or visible window).

Screenshots:
image

### DIFF
--- a/ui/src/plugins/com.android.InputEvents/index.ts
+++ b/ui/src/plugins/com.android.InputEvents/index.ts
@@ -83,38 +83,55 @@ export default class AndroidInputEvents implements PerfettoPlugin {
 
   async visualizeOverlaps(ctx: Trace): Promise<void> {
     const window = await getTimeSpanOfSelectionOrVisibleWindow(ctx);
+    const rootNode = await this.createRootTrack(ctx, window);
+    ctx.defaultWorkspace.pinnedTracksNode.addChildLast(rootNode);
 
-    const parentUri = `com.android.InputEvents.event_overlaps_parent.${uuidv4()}`;
-    const parentNode = await this.createTrack(
-        ctx,
-        parentUri,
-        `
-        SELECT *
-        FROM intervals_overlap_count!(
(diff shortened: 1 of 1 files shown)


zezeozue (approved):

ui/src/plugins/com.android.InputEvents/index.ts
```diff
@@ -66,4 +80,206 @@ export default class implements PerfettoPlugin {
       .getOrCreateStandardGroup(ctx.defaultWorkspace, 'USER_INTERACTION');
     group.addChildInOrder(node);
   }
+
+  async visualizeOverlaps(ctx: Trace): Promise<void> {
+    const window = await getTimeSpanOfSelectionOrVisibleWindow(ctx);
+    const rootNode = await this.createRootTrack(ctx, window);
+    ctx.defaultWorkspace.pinnedTracksNode.addChildLast(rootNode);
```
stevegolton, line 87: Adding a track directly to the pinned track node is unusual. You usually want to put the track in the normal workspace and the pin it, which means that when the user unpins it then it is still available in the normal track tree. Having said that there's nothing technically wrong with this I don't think, it's just unexpected. What's your intent here?

gignat-dev: This is a command and I don't think creating the tracks in the workspace is a good thing to do, considering we can call this command multiple times (over different selections)

PR #4822, opened 2026-02-17
ui: introduce new "embedder" interface to place deployment vars

LalitMaganti:
This CL introduces a new "embedder" interface to allow for people
deploying their own instances of the Perfetto UI to be able to easily
configure a bunch of changes without needing to make changes to the
source code smeared across the codebase.

This interface will eventually host all the deployment specific things
but for now I've just moved one thign over which is the GA id. Followup
changes will move the "default extension server" here also.

### DIFF
--- a/ui/src/core/analytics_impl.ts
+++ b/ui/src/core/analytics_impl.ts
@@ -75,12 +75,7 @@ export function initAnalytics(
   // local storage.
   // Skip analytics if the user has disabled analytics.
   // Skip analytics if the embedder does not provide an analytics ID.
-  if (
-    analyticsId !== undefined &&
-    !testingMode &&
-    !embeddedMode &&
-    enable
-  ) {
+  if (analyticsId !== undefined && !testingMode && !embeddedMode && enable) {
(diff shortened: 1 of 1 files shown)


LalitMaganti:
@primiano FYI

ui/src/core/embedder/embedder.ts
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Interface for embedder-specific behavior. Different implementations allow
+ * the UI to adapt to the environment it's running in (e.g. ui.perfetto.dev
+ * vs a third-party embedding).
+ */
+export interface Embedder {
```
LalitMaganti, line 20: Do we like Embedder as a name or do we prefer "deployment" or something else?

stevegolton: I think we've all know it best as embedder now so let's leave it as is.

PR #4829, opened 2026-02-17
proto: factor out MultifileErrorCollector

primiano:
This class is about to pass the "three strikes and you refactor"
test, as I'm about to add a third instance of MFE.


LalitMaganti:
BTW I only see two instances? Is there another one hiding?

primiano:
> BTW I only see two instances? Is there another one hiding?

Ah sorry my count was off by one. I counted in the other editor where I was writing the 3rd instance.

PR #4840, opened 2026-02-18
tp: Expand interval_intersect to 10 tables

wusamuel6:
Expand interval_intersect() to 10 tables, and leverage this expansion in the Wattson SQL to save some memory and performance.

Bug: 484354364

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/wattson/cpu/pivot.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/wattson/cpu/pivot.sql
@@ -105,7 +105,7 @@ SELECT
   *
 FROM _cpu_stats_subquery!(7, cpu7_curve, cpu7_static, freq_7, idle_7);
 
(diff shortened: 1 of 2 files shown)


src/trace_processor/perfetto_sql/engine/table_pointer_module.cc
```diff
@@ -221,7 +221,7 @@ int TablePointerModule::FindFunction(sqlite3_vtab*,
                                      void**) {
   if (base::CaseInsensitiveEqual(name, "__intrinsic_table_ptr_bind")) {
     *fn = [](sqlite3_context* ctx, int, sqlite3_value**) {
-      sqlite::result::Error(ctx, "Should not be called.");
+      sqlite::result::Long(ctx, 1);
```
wusamuel6: I'm not quite sure when this callflow is invoked, but I needed to stub it here in order for interval_intersect() to work.

PR #4844, opened 2026-02-18
ui: Simplify charts loader

aMayzner:
Introduces `ChartSource` — a unified SQL query builder — and `SQLChartLoader<TConfig, TData>` — an abstract base class that centralises caching (`QuerySlot`), lifecycle management (`SerialTaskQueue`), and the load/parse loop. All six chart loaders (bar, pie, line, scatter, histogram, treemap) are refactored to extend `SQLChartLoader`, replacing ~880 lines of duplicated SQL-construction logic with ~50-line subclass implementations.

## Changes

- **New `chart_source.ts`**: `ChartSource` supports three query shapes via a discriminated-union config — aggregated (with optional top-N + "Other" row and hierarchical grouping), point-based (with optional breakdown and stride-sampling), and histogram. Includes `inFilter()` and `rangeFilters()` helpers that build on the datagrid's `Filter` / `filterToSql` instead of hand-rolled SQL utilities.
- **New `chart_source_unittest.ts`**: 505 lines of tests covering all query modes — aggregated (simple, top-N with Other, hierarchical), points (basic, breakdown, stride-sampled), histogram, and filter/WHERE-clause generation.
- **Refactored loaders**: `bar_chart_loader`, `pie_chart_loader`, `line_chart_loader`, `scatterplot_loader`, `histogram_loader`, and `treemap_loader` now extend `SQLChartLoader` and only implement `buildQueryConfig()` + `parseResult()`.
- **Removed `sqlInClause` / `sqlRangeClause`** from `chart_utils.ts` — replaced by the new filter helpers.

### DIFF
--- a/ui/src/components/widgets/charts/chart_sql_source.ts
+++ b/ui/src/components/widgets/charts/chart_sql_source.ts
@@ -13,7 +13,11 @@
 // limitations under the License.
 
 import {assertUnreachable} from '../../../base/logging';
-import {QuerySlot, SerialTaskQueue} from '../../../base/query_slot';
+import {
+  QuerySlot,
+  SerialTaskQueue,
+  type QueryResult,
+} from '../../../base/query_slot';
 import type {Engine} from '../../../trace_processor/engine';
 import type {QueryResult as TPQueryResult} from '../../../trace_processor/query_result';
 import {AggregateFunction, Filter} from '../datagrid/model';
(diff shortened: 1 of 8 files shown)


ui/src/components/widgets/charts/chart_sql_source.ts
```diff
@@ -0,0 +1,661 @@
...
+ *
+ * const sql = source.buildQuery({
+ *   type: 'aggregated',
+ *   dimensions: [{column: 'process_name'}],
+ *   measures: [{column: 'dur', aggregation: 'SUM'}],
+ *   limit: 10,
+ * });
+ * ```
+ */
+export class ChartSource {
```
stevegolton: I'm not sure the point of this class - it seems like it only has one function buildQuery() which could just be a standalone function...

ui/src/components/widgets/charts/chart_sql_source.ts
```diff
@@ -0,0 +1,661 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// SQLChartLoader base class
+// ---------------------------------------------------------------------------
+
+/**
+ * Result returned by SQL chart loaders.
+ */
+export interface ChartLoaderResult<TData> {
```
stevegolton: Isn't this just QueryResult? from query_slot.ts? You can just return it directly.

ui/src/components/widgets/charts/chart_sql_source.ts
```diff
@@ -0,0 +1,661 @@
...
+    });
+    return {data: result.data, isPending: result.isPending};
+  }
+
+  dispose(): void {
+    this.querySlot.dispose();
+  }
+
+  /** Build the QueryConfig from the per-use config. */
+  protected abstract buildQueryConfig(config: TConfig): QueryConfig;
```
stevegolton: I think we worked out a long time ago that composition is better than inheritance. Do we need the inheritance tree?

stevegolton (approved):
LGTM % nits

PR #4847, opened 2026-02-18
WIP: Simple lezer-based SQL formatter

stevegolton:
Add a SQL formatting feature to the query page and explore page.

Use `Opt/Alt+Shift+F` to auto format when any editor is highlighted.


stevegolton:
Dropped in favor of Lalit's WASM based formatter (upcoming)

PR #4857, opened 2026-02-18
heapprofd: add PERFETTO_HEAPPROFD_BLOCKING_EXIT env variable

### LINKED ISSUE #4801: heap profiling (53.0) not working with absl time lib (Linux)
Hi, I have tried the old and latest absl libs and the problem remains - here is the minimum code to reproduce the problem in our env:

```
#include <iostream>
#include "absl/time/time.h"

int main(int argc, char** argv) {
    std::cout << "====" << absl::UTCTimeZone() << std::endl; // <==============THIS LINE
    auto leaked = new int [100000];
    for (auto i = 0; i < 100000; ++i) {
        leaked[i] = i % 37;
    }
    std::cout << "************leaked output=" << leaked[0] + leaked [100000 - 1]  << std::endl;
    return 0;
}
```

Just comment out the marked out line, the generated trace show the leaks calling stack correctly - otherwise it shows:

```
root
ERROR MEMORY_INVALID (the leaked size seems to be correct but no symbols)
unknown
```

LalitMaganti:
Short-lived processes can exit before heapprofd finishes reading and
unwinding all pending stacks from the shared ring buffer, causing
ERROR MEMORY_INVALID in the resulting trace.

Add a new PERFETTO_HEAPPROFD_BLOCKING_EXIT=1 environment variable
that registers an atexit handler to wait (up to 10s) for heapprofd
to drain the ring buffer before the process exits. This is independent
of PERFETTO_HEAPPROFD_BLOCKING_INIT which controls blocking on first
malloc for initialization.

Fixes: https://github.com/google/perfetto/issues/4801

### DIFF
--- a/src/profiling/memory/client_api.cc
+++ b/src/profiling/memory/client_api.cc
@@ -32,6 +32,7 @@
 
 #include "perfetto/base/build_config.h"
 #include "perfetto/base/logging.h"
+#include "perfetto/base/time.h"
 #include "perfetto/ext/base/string_utils.h"
 #include "perfetto/ext/base/unix_socket.h"
 #include "perfetto/ext/base/utils.h"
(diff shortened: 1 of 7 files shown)


LalitMaganti:
@rsavitski should we find the time to discuss this at some point?

src/profiling/memory/client_api.cc
```diff
@@ -586,3 +611,12 @@ __attribute__((visibility("default"))) bool AHeapProfile_initSession(
 
   return true;
 }
+
+bool AHeapProfile_initSessionWithBlockingExit(void* (*malloc_fn)(size_t),
```
primiano, line 615: this smells a bit bad (as in: are we going to create a new initSession variant for each new behaviour?)
Why this is not just a call made right after AHeapProfile_initSession() ?
there doens't seem to be any reason I can see why you want to call this before the actual init

What I'm suggesting is: why doesn't the code in factory_standalone just do:

```
AHeapProfile_initSession(malloc, free)
if (g_blocking_exit)
  AHeapProfile_installExitHandler();
```

or if the malloc interception creates problems, do them in opposite order?

src/profiling/memory/client_api.cc
```diff
@@ -280,6 +281,30 @@ void AtForkChild() {
...
+      return;
+    client = *GetClientLocked();
+  }
+  if (!client)
+    return;
+
+  constexpr uint32_t kTimeoutUs = 10000 * 1000;
+  constexpr uint32_t kSleepUs = 10 * 1000;
+  uint32_t elapsed_us = 0;
+  while (client->read_avail() > 0 && elapsed_us < kTimeoutUs) {
```
primiano, line 302: hmm this might cause some never ending loop where you might never converge if you exit() a multithreaded program. 
the way this should work is that you need to wirte some sort of fence in the shmem an know when you reached it.
**I think** (but didn't look to deep) at the end of the day the semantic you want here is: you need to observe the current wirite_pos and keep sleeping until you see that the shmem's read_pos has reached that write_pos

PR #4876, opened 2026-02-20
[Recorder] Show category description and tags in chrome tab

etiennep-chromium:
This PR adds UI for showing category descriptions and the list of tags.
Tags: directly following the category name
Description: Shown on hover
This also merges disabled-by-default and "normal" categories to the same list, because "disabled-by-default" categories aren't treated any differently from "normal" categories with the slow tag.

image

### DIFF
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/chrome.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/chrome.ts
@@ -14,6 +14,7 @@
 
 import m from 'mithril';
 import protos from '../../../protos';
+// import {Tooltip} from '../../../widgets/tooltip';
 import {
   RecordSubpage,
   RecordProbe,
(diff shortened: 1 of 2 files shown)


camillobruni:
- I like merging the two lists 👍 
- 
- I think we should hide the tags and only show the in the tooltip (or use an actual formatted popup)? I find it quite hard to scan the list for names if both `(disabled-by-default)` and tags are present.
- It's a bit confusing that we can't click on the info icon (but that's probably also fine)

Screenshot 2026-02-23 at 13 45 36

etiennep-chromium:
> I think we should hide the tags and only show the in the tooltip 

I tried using Tooltip widget to make a structured tooltip but I ended up fighting with checkbox mouseover.
I ended up with the simpler solution of using "title" attribute, but displaying tags a Chip (see below)

image

camillobruni:
I like the chips 👍 .
I wonder, do we still need the disabled-by-default information in the case we have tags?
it's somehow confusing in the UI in general, since we only really care if it's enabled or disabled, and with the nice tags we have even better `slow` annotations?

etiennep-chromium:
> I wonder, do we still need the disabled-by-default information in the case we have tags?

I think it's (unfortunately) necessary to distinguish between e.g. "base" and "disabled-by-default-base", because the disabled-by-default prefix is part of the name identifying the category (otherwise the 2 categories would be indistinguishable), as opposed to tags, and "slow" tag isn't guaranteed to be present.
The old "disabled-by-default" mechanism is all gone though, so now it's only a part of the category name, so ideally we eventually replace all disabled-by-default prefix by something more precise / informative.

PR #4877, opened 2026-02-20
ui: Add more charts

aMayzner:
Extends the chart library with three new chart types
- CDF (Cumulative Distribution Function)
- Boxplot
- Heatmap 
each with a corresponding SQL-backed loader. Also improves `createChartLoader` to retain stale data
while new queries are in-flight, preventing loading spinner flashes during brush/filter changes.

### DIFF
--- a/ui/src/components/widgets/charts/boxplot.ts
+++ b/ui/src/components/widgets/charts/boxplot.ts
@@ -1,0 +1,197 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


ui/src/components/widgets/charts/boxplot.ts
```diff
@@ -0,0 +1,197 @@
...
+    }),
+    series: [
+      {
+        type: 'boxplot',
+        data: boxData,
+      },
+    ],
+  };
+
+  return option as EChartsCoreOption;
```
stevegolton, line 196: Assert required?

ui/src/components/widgets/charts/heatmap.ts
```diff
@@ -0,0 +1,190 @@
...
+          itemStyle: {
+            shadowBlur: 10,
+            shadowColor: 'rgba(0, 0, 0, 0.5)',
+          },
+        },
+      },
+    ],
+  };
+
+  return option as EChartsCoreOption;
```
stevegolton, line 189: Type assertion required?

stevegolton (approved):

ui/src/components/widgets/charts/heatmap.ts
aMayzner: Yup, as the option is of the type of Record<string, unknown>, to be able to freely modify the options from ExhartsCoreOption. We do the same in bar_chart.ts.

Issue #860, opened 2024-08-18
Implementation to add data to info and stats page (Generic way not tied to Chrome)


brandonlo11:
I am trying to display data in tabs that is not part of the data rendered in tracks. The only ways I've found to do that is to render an event in a track to be able to display the data in tabs or to use ChromeMetaData packets. (This issue was brought up to lalitm on discord)

LalitMaganti:
I've scoped out how to do this and I have a plan: will send a CL soon.

LalitMaganti:
@primiano and I chatted: we've agreed that we need to not just solve this issue but also add a bunch of protos to add slices, counters, stack traces and trace metadata in a much simpler way than track_event and much more efficient way than JSON. We're going to be looking into designing this in the coming months.

LalitMaganti:
So the above PR actually adds the proto but we don't have documentation for this @tikurahul: do you mind adding a section in our "converting to Perfetto" docs page?

PR #4879, opened 2026-02-20
Add the ability to add generic metadata pertaining to a trace or the …

tikurahul:
Add the ability to add generic metadata about a trace or the environment for a tracing session. 

* Traces can have multiple `TraceMetadata` packets.
* Keys **must** be namespaced to avoid collisions. 
* The last `value` for a given `key` wins when storing metadata. So we have overwrite semantics.
* Currently the `metadata` table supports `string`, `int64` and `json` types.
  I am therefore restricting the value types to `string` and `int64`.

Addresses: FR https://github.com/google/perfetto/issues/860
Test: Added unit tests


protos/perfetto/common/trace_metadata.proto
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Helps annotate a trace with top level metadata (key value pairs) that can be
```
LalitMaganti: The way I would phrase it would be more "Arbitrary key value pairs which can attached to a trace to give context about any facet of the trace. This can include, what data it contains, properties of the machine the trace it was collected on, pre-baked analysis etc".

Then I would give some examples of what type of things you think would fit here

LalitMaganti: Also some documentation about where this will be displayed in the UI *and* where it will appear in the trace processor tables would be helpful.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -499,6 +504,23 @@ void ProtoTraceReader::HandleFirstPacketOnSequence(
...
+    protos::pbzero::TraceMetadata_Metadata::Decoder decoded_metadata(*it_buf);
+    auto key = decoded_metadata.key();
+    auto key_id = context_->storage->InternString(key);
+    if (decoded_metadata.has_long_value()) {
+      auto value = Variadic::Integer(decoded_metadata.long_value());
+      context_->metadata_tracker->SetDynamicMetadata(key_id, value);
+    } else if (decoded_metadata.has_string_value()) {
+      auto value = Variadic::String(
+          context_->storage->InternString(decoded_metadata.string_value()));
+      context_->metadata_tracker->SetDynamicMetadata(key_id, value);
```
LalitMaganti, line 542: IMO this should be prefixed somehow. Maybe with trace_metadata.<key> or similar. Something which indicates the source and namespaces otherwise you can just accidentally overwrite critical keys to TP.

LalitMaganti:
We also should think about where/how this will be displayed in the UI as well. Right now it's only going to show up in the "advanced" page. is that the right place? Or do we want something better?

src/trace_processor/importers/proto/proto_trace_reader.cc
tikurahul: Added the `trace_metadata_` prefix. Let me know if there are better ways of doing this.

protos/perfetto/common/trace_metadata.proto
tikurahul: Added more documentation about key prefixes (`trace_metadata_`), and where this metadata is visible when using Perfetto UI.

tikurahul:
> We also should think about where/how this will be displayed in the UI as well. Right now it's only going to show up in the "advanced" page. is that the right place? Or do we want something better?

I think that is a good start.

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -5521,6 +5521,41 @@ message SystemInfo {
 
 // End of protos/perfetto/common/system_info.proto
 
+// Begin of protos/perfetto/common/trace_metadata.proto
+
+// Arbitrary key value pairs which can attached to a trace to give context about
+// any facet of the trace. This can include, what data it contains, properties
+// of the host / machine the trace was collected on, and other interesting
+// information about a trace. These keys should be namespaced to avoid
+// collisions. All keys get prefixed with `trace_metadata_`.
```
LalitMaganti: No the namespacing was about the trace processor side, they don't need to be double namespaced here.

LalitMaganti: ... in the UI and in the trace processor.

protos/perfetto/common/trace_metadata.proto
```diff
@@ -0,0 +1,49 @@
...
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Arbitrary key value pairs which can attached to a trace to give context about
+// any facet of the trace. This can include, what data it contains, properties
+// of the host / machine the trace was collected on, and other interesting
+// information about a trace. These keys should be namespaced to avoid
+// collisions. All keys get prefixed with `trace_metadata_`.
+
```
LalitMaganti: All of these should be one continuous set. Empty lines should just have ///

protos/perfetto/common/trace_metadata.proto
```diff
@@ -0,0 +1,49 @@
...
+// Examples include:
+// gradle_version = "9.0.10-alpha01"
+// java_major_version = 24
+
+// These key value pairs show up in the `Overview -> Info and Stats (advanced)`
```
LalitMaganti: Just in the overview page. The subpage I think can and should change (likely we need to create a new page) and I don't want to overcommit to thatl.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -520,6 +525,25 @@ void ProtoTraceReader::HandleFirstPacketOnSequence(
   }
 }
 
+void ProtoTraceReader::HandleTraceMetadata(protozero::ConstBytes blob) {
+  protos::pbzero::TraceMetadata::Decoder decoder(blob);
+  for (auto it_buf = decoder.metadata(); it_buf; ++it_buf) {
+    protos::pbzero::TraceMetadata_Metadata::Decoder decoded_metadata(*it_buf);
+    auto key = decoded_metadata.key();
+    // Prefix all custom metadata keys with `trace_metadata_`.
```
LalitMaganti: The prefix should be `trace_metadata.` (. is used as a universally separator in args like thing between fields).

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -5521,6 +5521,40 @@ message SystemInfo {
...
+// gradle_version = "9.0.10-alpha01"
+// java_major_version = 24
+///
+// All keys get prefixed with `trace_metadata.` in the UI and in
+// `traceprocessor`. These key value pairs show up in the `Overview` section in
+// Perfetto UI. Additionally, these key, value pairs can also be queried using
+// Pefetto SQL by using the following query:
+///
+// `SELECT * FROM metadata WHERE name GLOB 'trace_metadata.*';`
+
```
LalitMaganti, line 5543: nit: space

nit: we also need a comment saying these are trace prefixed *and* if there are multiple in a trace for the same key, the latest will win.

primiano:
out of curiosity how frequently do you add new keys? Could have been nicer if they were actual keys in the protos, rathre than generic key-value pairs, as now this creates a new indirection layer.
TP could do automatlically the symbolziation (so you don't have to add new C++ code every time you add a new key) but would have made codesearch and understanding what's going on much easier.

Now if you see a trace that has metadata "foobar" you need to codesearch for which project injects a metadata with key=foo, rathern thatn setting that specific field

It's all a matter of how frequently you need more keys TBH. if highly flexible, then you need this. If they change rather slowly, imho would have been nicer to add theme explicitly

LalitMaganti:
I think the idea here was this was the "blessed" way for people to dump *any* metadata into the trace. So yes we don't actually want to need to modify protos every time someone wants to add a new thing, this is equivalent to "track event args"

PR #4890, opened 2026-02-23
ui: Add rectangle brush to scatterplot and optional gridlines to all charts

aMayzner:


ui/src/plugins/dev.perfetto.WidgetsPage/demos/charts_demo.ts
```diff
@@ -192,6 +199,7 @@ export function renderCharts(app: App): m.Children {
           enableBrush: opts.enableBrush,
           horizontal: opts.horizontal,
           aggregation: opts.aggregation as AggregateFunction,
```
stevegolton: Driveby - I don't think these type assertions are necessary

aMayzner:
> Looks good in general
> 
> The scatter plot box selection doesn't work as expected. Basically it seems the axes start at 0,0 no matter where the selection is made but all the points outside the selection are hidden, which is confusing.
> 

Thank you!

PR #4894, opened 2026-02-23
tp: add TraceProcessorShellMain

dreveman:
Small wrapper around TraceProcessorShell class that provides a library interface that matches the traced service interfaces.

This makes bundling into a single binary like tracebox easier and more consistent.

### DIFF
--- a/include/perfetto/ext/trace_processor/trace_processor_shell.h
+++ b/include/perfetto/ext/trace_processor/trace_processor_shell.h
@@ -89,6 +89,8 @@ class TraceProcessorShell_PlatformInterface {
       std::function<void(size_t)> progress_callback) = 0;
 };
 
+int PERFETTO_EXPORT_COMPONENT TraceProcessorShellMain(int argc, char** argv);
+
 }  // namespace perfetto::trace_processor
 
 #endif  // INCLUDE_PERFETTO_EXT_TRACE_PROCESSOR_TRACE_PROCESSOR_SHELL_H_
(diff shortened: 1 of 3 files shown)


LalitMaganti:
Can I ask what the goal is here? I'm not opposed to this change but I am worried when I hear "tracebox" and "shell" at the same time :)

@primiano and I have had some disucssions and we've agreed that trace_processor_shell needs to evolve a bit and become the "tracebox of offline processing". This means bundling the functionality of `traceconv` inside it (and removing traceconv as a separate binary) and also making it capable of having a http server for the sake of having locally provided extension servers (https://github.com/google/perfetto/discussions/3227)

dreveman:
> Can I ask what the goal is here? I'm not opposed to this change but I am worried when I hear "tracebox" and "shell" at the same time :)

It's specific to some experiments I'm doing for my specific use-case where I'd like to bundle everything (tracebox + trace_processor_shell) into a single statically linked binary. It might turn into nothing useful and certainly not something I expect upstream to support or turn tracebox into. This change is not critical for this, works ok without, just a bit cleaner with this change in upstream and it seemed like a general improvement to use this existing pattern upstream.

> 
> @primiano and I have had some disucssions and we've agreed that trace_processor_shell needs to evolve a bit and become the "tracebox of offline processing". This means bundling the functionality of `traceconv` inside it (and removing traceconv as a separate binary) and also making it capable of having a http server for the sake of having locally provided extension servers (#3227)

That sounds great. If what I'm experimenting with turns out useful then I'll benefit from this.

PR #4895, opened 2026-02-23
Changed severity to error

srapiteanu:
Bug: b/363786060
This is a small change in winscope_protolog_view_config_collision severity from info to error to match the task description "trace process error".


LalitMaganti:
This was done intentionally in http://github.com/google/perfetto/pull/1866. Have we fixed the underlying causes? Because if not it will just cause APC pipeline problems.

srapiteanu:
Thank you @LalitMaganti for your input! You are right, I will cancel this PR.

PR #4900, opened 2026-02-23
ui: add embedder API for sidebar branding badges

dreveman:
Add a BrandingBadge interface to the Embedder API that allows embedders to
display custom branding (text, icon, or inline image) in the sidebar header
next to the Perfetto logo.

The badge is rendered above the existing channel indicator (CANARY/AUTOPUSH),
which shifts down automatically via CSS :has() only when a badge is present.

The API supports three visual modes:
- Text only (e.g. {text: 'INTERNAL', color: '#0668E1'})
- Material icon + text (e.g. {icon: 'star', text: 'BRAND'})
- Inline data URI image + text (e.g. {image: 'data:image/svg+xml,...'})

Both DefaultEmbedder and PerfettoUiEmbedder set brandingBadge to undefined
(no branding). A fork wanting custom branding would either modify
default_embedder.ts directly or create a custom embedder class.

### DIFF
--- a/ui/src/frontend/sidebar.ts
+++ b/ui/src/frontend/sidebar.ts
@@ -386,6 +386,21 @@ export class Sidebar implements m.ClassComponent {
       m(
         `header.pf-sidebar__channel--${getCurrentChannel()}`,
         m(`img[src=${assetSrc('assets/brand.png')}].pf-sidebar__brand`),
+        app.embedder.brandingBadge &&
+          m(
+            'span.pf-sidebar__branding-badge',
+            {style: {color: app.embedder.brandingBadge.color}},
+            app.embedder.brandingBadge.image
+              ? m('img.pf-sidebar__branding-badge-img', {
+                  src: app.embedder.brandingBadge.image,
+                })
+              : app.embedder.brandingBadge.icon &&
+                  m(Icon, {
+                    icon: app.embedder.brandingBadge.icon,
+                    className: 'pf-sidebar__branding-badge-icon',
+                  }),
+            app.embedder.brandingBadge.text,
+          ),
         m(Button, {
           icon: 'menu',
           className: 'pf-sidebar-button',
--- a/ui/src/assets/sidebar.scss
+++ b/ui/src/assets/sidebar.scss
@@ -52,6 +52,25 @@
       height: 36px;
       margin-top: 4px;
     }
+    .pf-sidebar__branding-badge {
+      display: flex;
+      align-items: center;
+      position: absolute;
+      font-size: var(--pf-font-size-xs);
+      line-height: 1;
+      font-family: var(--pf-font);
+      top: 7px;
+      right: 48px;
+      z-index: 10;
+      i.pf-icon {
+        font-size: var(--pf-font-size-xs);
+        margin-right: 2px;
+      }
+      .pf-sidebar__branding-badge-img {
+        height: 10px;
+        margin-right: 2px;
+      }
+    }
     &::before {
       z-index: 10;
     }
@@ -64,6 +83,12 @@
       font-family: var(--pf-font);
       top: 7px;
       right: 48px;
+    }
+    &:has(.pf-sidebar__branding-badge) {
+      &.pf-sidebar__channel--canary::before,
+      &.pf-sidebar__channel--autopush::before {
+        top: 18px;
+      }
     }
     &.pf-sidebar__channel--canary::before {
       content: "CANARY";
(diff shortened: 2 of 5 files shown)


dreveman:
Given that a stable plugin API is not in the plans we're likely going to keep having alternative instances of the UI with extra plugins. The purpose of this change is to make it easier for the user to tell if they are using the upstream UI or some alternative instance. This is useful to the user but maybe even more so to developers triaging bugs as the inclusion of a screenshot in the report makes it clear if the upstream UI is used or not. Wdyt?

stevegolton:
Sounds like a solid plan to solve the screenshot identity crisis, though talking with the team, we think this would be best living in the new embedders.ts file that @LalitMaganti just added rather than the plugin API.

embedders.ts file is really a deployment config file for the UI, providing a single file that can be modified in a fork making it simpler to merge from upstream. I guess this could almost be a config file in the long run, but for now it's a compiled typescript file to allow more powerful configuration such as functions to be defined rather than just pure data.

For this PR I think the best thing to do would be to move the branding API into the embedders file API, then define or customize this file for your own deployment - that is assuming you have a distinct branding per installation of the UI - otherwise we might have to rethink.

dreveman:
> Sounds like a solid plan to solve the screenshot identity crisis, though talking with the team, we think this would be best living in the new embedders.ts file that @LalitMaganti just added rather than the plugin API.
> 
> embedders.ts file is really a deployment config file for the UI, providing a single file that can be modified in a fork making it simpler to merge from upstream. I guess this could almost be a config file in the long run, but for now it's a compiled typescript file to allow more powerful configuration such as functions to be defined rather than just pure data.
> 
> For this PR I think the best thing to do would be to move the branding API into the embedders file API, then define or customize this file for your own deployment - that is assuming you have a distinct branding per installation of the UI - otherwise we might have to rethink.

@stevegolton Thanks for the review. This makes sense. I've updated this PR as you suggested. Ptal.

primiano:
One side comment: (maybe in a separate PR) we should add the branding to `const GLOBAL_MASKS` in perfetto_ui_test_helper.ts. I don't know if @dreveman runs our UI test (realistically I don't they they worry about them), but if somebody does, it's nicer to make the test branding-invariant.
but this requires rebaselining all the tests expectations, and this is by far not a blocker, so I suggest doing that in a separate PR after the release

dreveman:
> One side comment: (maybe in a separate PR) we should add the branding to `const GLOBAL_MASKS` in perfetto_ui_test_helper.ts. I don't know if @dreveman runs our UI test (realistically I don't they they worry about them), but if somebody does, it's nicer to make the test branding-invariant. but this requires rebaselining all the tests expectations, and this is by far not a blocker, so I suggest doing that in a separate PR after the release

Not sure exactly what this entails but happy to work on this as a follow.

stevegolton (commented):
Looking good!

One thing I'm wondering is whether this should just be a function returning mithril vnodes:

`branding: () => m('img', {src: ...})`

It would be a simpler narrower interface, much more flexible and more resistant to feature creep.

@lalitm @primiano WDYT?

Apologies for the pedantry, but it want to get this right as it'll be hard to change later.

LalitMaganti:
> One thing I'm wondering is whether this should just be a function returning mithril vnodes:

Makes sense. 

> Apologies for the pedantry, but it want to get this right as it'll be hard to change later.

Curious why this is? Like plugin api I think we should only have very soft backwards compatibility guarantees for embedder.ts

dreveman:
@stevegolton I've updated the code to use a function returning mithril vnodes. ptal

stevegolton:
> Curious why this is? Like plugin api I think we should only have very soft backwards compatibility guarantees for embedder.ts

Well we're advertising the embedders API as a way to customise non-Google UI installations, which is necessarily out-of-tree, so we should offer a more stable API no?

@dreveman thanks for applying the update. I'll take a look this afternoon.

ui/src/frontend/sidebar.ts
```diff
@@ -386,6 +396,12 @@ export class Sidebar implements m.ClassComponent {
       m(
         `header.pf-sidebar__channel--${getCurrentChannel()}`,
         m(`img[src=${assetSrc('assets/brand.png')}].pf-sidebar__brand`),
+        app.embedder.branding &&
+          m(
+            '.pf-sidebar__branding',
+            app.embedder.branding(),
```
stevegolton: Apologies for the churn, after seeing the impl and discussing with the team, I think what you had previously was ideal.

I'll approve this now, but please revert the last commit back to what it was and we can get this landed.

dreveman: No problem. Switched back to the BrandingBadge interface in latest version of PR. I'll merge this as soon as it passed CI checks.

PR #4902, opened 2026-02-24
add trace config option to preserve logcat buffers

cphlipot1:
Add and option to preserve logcat buffers, analogous to the existing
[preserve_ftrace](https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto//protos/perfetto/config/ftrace/ftrace_config.proto;l=255) option.

keeping logcat from before the trace starts is useful in similar situations
where you would want to do this for ftrace, such as tracing the boot sequence.

## testing:
test_preserve_log_buffer.pbtx:
```
buffers: {
  size_kb: 32768
  fill_policy: DISCARD
}

data_sources: {
  config {
    name: "android.log"
    android_log_config {
      # Include all common log buffers
      log_ids: LID_DEFAULT
      log_ids: LID_SYSTEM
      log_ids: LID_EVENTS
      log_ids: LID_CRASH
      log_ids: LID_KERNEL

      preserve_log_buffer: true
    }
  }
}

data_sources: {
  config {
    name: "linux.ftrace"
    ftrace_config {
      ftrace_events: "sched/sched_switch"
      ftrace_events: "sched/sched_waking"
      ftrace_events: "sched/sched_wakeup"
    }
  }
}

duration_ms: 5000
```

```
./tools/record_android_trace --sideload-path ./out/android_release_arm64/tracebox -c test_preserve_log_buffer.pbtx
```

image

### DIFF
--- a/protos/perfetto/config/android/android_log_config.proto
+++ b/protos/perfetto/config/android/android_log_config.proto
@@ -31,4 +31,11 @@ message AndroidLogConfig {
   // If non-empty ignores all log messages whose tag doesn't match one of the
   // specified values.
   repeated string filter_tags = 4;
+
+  // If true, includes the contents of the logcat buffers that were already
+  // present before the data source started. By default (false), only new
+  // log entries produced after the data source starts are recorded.
+  // Can be useful for boot traces or situations where logs from before
+  // traced started are important.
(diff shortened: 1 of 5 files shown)


primiano:
makes sense, thanks for the PR

PR #4905, opened 2026-02-24
ui: Tidy up chart theming

stevegolton:
- Centralize theming in EchartView to avoid duplication.
- Avoid depending on `pf-theme-provider` and remove fallback to documentElement as this won't have the CSS variables set.
- Fixes reactivity in scroll position chart in chrome scroll plugin
- Use themes rather than options to specify colors, which avoids having to fiddle around with injecting theme colors into options.

### DIFF
--- a/ui/src/components/widgets/charts/treemap.ts
+++ b/ui/src/components/widgets/charts/treemap.ts
@@ -18,18 +18,6 @@ import {formatNumber} from './chart_utils';
 import {EChartView, EChartEventHandler, EChartClickParams} from './echart_view';
 import {buildTooltipOption} from './chart_option_builder';
 
-// Default chart colors - these will be overridden by EChartView's theme application
-const DEFAULT_CHART_COLORS = [
-  '#5470c6',
-  '#91cc75',
-  '#fac858',
-  '#ee6666',
-  '#73c0de',
-  '#3ba272',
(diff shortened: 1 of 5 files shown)


ui/src/components/widgets/charts/treemap.ts
```diff
@@ -17,7 +17,18 @@ import type {EChartsCoreOption} from 'echarts/core';
...
+const DEFAULT_CHART_COLORS = [
+  '#5470c6',
+  '#91cc75',
+  '#fac858',
+  '#ee6666',
+  '#73c0de',
+  '#3ba272',
+  '#fc8452',
+  '#9a60b4',
+];
```
aMayzner: AI did the same for me, but do we want to have them hardcoded this way in .ts?

stevegolton: I think the problem is that TreeMap is assigning colors manually to each node instead of using the theme.

I just updated a PR to make the TreeMap use the theme colors instead, and remove the hard coded colors.

This also meant that the 'category' field which could be assigned per node of the input data was no longer used anywhere, so I removed it - the TreeMap chart automatically assigns a similar color to each of the top level groups using the theme colors which I think actually looks better than what we had before.

Old: 
image

New:
image

PR #4908, opened 2026-02-24
ui: Consolidate fonts by using Roboto Flex for condensed text

stevegolton:
We are already using Roboto Flex for the normal Roboto fonts but a separate RobotoCondensed font file for compact fonts. With the modern Roboto Flex font we no longer need two font files, and we can just use the single file with a different width.

This saves space, and also looks a lot nicer especially on standard DPI displays.

Before
Font|Size|Notes
---|---|---
RobotoCondensed.woff2|50K|
Roboto.woff2|30K|No width axis

After
Font|Size|Notes
---|---|---
Roboto|56K|Variable width axis


stevegolton:
It's not possible to make RobotoFlex look exactly like RobotoCondensed, giving up.

stevegolton:
> I'd go with Flex TBH! Thanks for this

Thanks for testing!

PR #4909, opened 2026-02-24
Support critical path analysis for Fuchsia traces

anp:
As of a few months ago, Fuchsia traces now include enough information to reconstruct the graph of wakes/unblocks that led to a particular slice of execution on a CPU.

The included change extends the importer to put that information in the analysis database and updates the critical path queries to account for quirks of Fuchsia's scheduler traces. It includes unit test coverage for the importer and a diff test for the query updates.

This is my first time contributing to this project, so please let me know if I'm holding this wrong. In particular I wasn't sure if y'all prefer to have these two commits as separate branches/PRs like in a Gerrit-style flow.


src/trace_processor/perfetto_sql/stdlib/sched/thread_executing_span.sql
```diff
@@ -55,6 +55,17 @@ FROM thread_state
 WHERE
   thread_state.dur != -1 AND thread_state.waker_id IS NOT NULL;
 
+-- Defines if the trace is a fuchsia trace.
+CREATE PERFETTO TABLE _is_fuchsia AS
```
zezeozue, line 59: Can use a PERFETTO FUNCTION

zezeozue: More specifically. I'm thinking something like:

-- Returns TRUE if the state is 'R' (standard Linux) 
-- or 'W' (specific to Fuchsia traces).
CREATE PERFETTO FUNCTION _is_runnable_state(state STRING)
RETURNS BOOL AS
SELECT 
  $state = 'R' OR (
    $state = 'W' AND EXISTS(
      SELECT 1 FROM metadata 
      WHERE name = 'trace_type' AND str_value = 'fuchsia'
    )
  );

LalitMaganti: Honestly I'm a bit worried about both using a function *and* how this patch is structured. Isn't this very perfomance sensitive query wise?

zezeozue: The usages shoudn't be. It's just prepopulating some initial state (and final) state. Worth double checking though

zezeozue: Ok. You are right. The second usage in wakeup_map with the window function is bad with the perfetto function but this is fine:

-- Defines if the trace is a fuchsia trace.
CREATE PERFETTO TABLE _is_fuchsia AS
SELECT
  EXISTS(
    SELECT
      1
    FROM metadata
    WHERE
      name = 'trace_type' AND str_value = 'fuchsia'
  ) AS is_fuchsia;

CREATE OR REPLACE PERFETTO FUNCTION _is_runnable_state(state STRING)
RETURNS BOOL AS
SELECT
$state = 'R' OR (
$state = 'W' AND (SELECT is_fuchsia FROM _is_fuchsia)
);

So still defining the _is_fuchsia table and using it from the function

anp: I'll give this a shot, thanks!

I can get some numbers, is there an existing/documented flow that y'all use?

anp: I used [this script](https://gist.github.com/anp/52d100776f3a4d60c4ec132287ff91b3) to measure the wakeup graph reporting overhead before/after my changes.

On `main`:

```
Iterations: 10
Min: 1733 ms
Max: 1825 ms
Mean: 1782.80 ms
Median: 1787.0 ms
StdDev: 25.53 ms
```

Using a function for `_is_runnable_state`:

```
Iterations: 10
Min: 2502 ms
Max: 2643 ms
Mean: 2568.90 ms
Median: 2579.5 ms
StdDev: 48.31 ms
```

That's a pretty big regression! I tried converting the function to a macro and now I get:

```
Iterations: 10
Min: 1730 ms
Max: 1836 ms
Mean: 1771.90 ms
Median: 1770.5 ms
StdDev: 32.39 ms
```

Seems pretty comparable to the baseline numbers. Pushing that change up now.

aMayzner (approved):

anp:
Thanks for the reviews! I was OOO for a couple of weeks but I'm back now and I'm wondering what the next steps to land this are. I updated the branch to see if that would get the "2 owners' approvals" check to pass.

anp:
Thanks!

PR #4910, opened 2026-02-25
tp: Fix index corruption in StringFilterNe for missing strings

sashwinbalaji:
When a string being compared  was not present in the string pool, the interpreter used memcpy to overwrite the output buffer with the source buffer.

This corrupted the row mapping for all subsequent operations in the query, causing lookups in other columns to return data from incorrect rows.

Since '!=' is always true if the string is missing, we simply keep the existing table indices in the output buffer. 

Bug: b/487405013

### DIFF
--- a/src/trace_processor/core/interpreter/bytecode_interpreter_unittest.cc
+++ b/src/trace_processor/core/interpreter/bytecode_interpreter_unittest.cc
@@ -925,7 +925,69 @@ TEST_F(BytecodeInterpreterTest, StringFilterNeStringNotInPool) {
       GetSpan(update_buffer), GetStoragePtr<String>(0));
 
   // All 4 indices should be returned since "nonexistent" != any string.
-  EXPECT_THAT(GetRegister<Span<uint32_t>>(2), ElementsAre(100, 101, 102, 103));
+  const auto& res = GetRegister<Span<uint32_t>>(2);
+  EXPECT_THAT(res, ElementsAre(100, 101, 102, 103));
+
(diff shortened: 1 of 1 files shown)


src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
```diff
@@ -1002,7 +1002,6 @@ inline PERFETTO_ALWAYS_INLINE uint32_t* StringFilterNe(
   std::optional<StringPool::Id> id =
       state.string_pool->GetId(base::StringView(val));
   if (!id) {
-    memcpy(output, begin, size_t(end - begin) * sizeof(*begin));
```
LalitMaganti, line 1005: This is only correct if source = destination. I think the problem actually is that if there are overlapping buffers, memcpy is undefined. I think we need to change this to memmove which should fix the issue.

LalitMaganti: Misunderstood this fix is correct.

src/trace_processor/core/interpreter/bytecode_interpreter_unittest.cc
```diff
@@ -924,8 +924,8 @@ TEST_F(BytecodeInterpreterTest, StringFilterNeStringNotInPool) {
       CastFilterValueResult::Valid("nonexistent"), GetSpan(source_indices),
       GetSpan(update_buffer), GetStoragePtr<String>(0));
```
LalitMaganti, line 926: Please let's improve the tests.

PR #4920, opened 2026-02-25
[android sdk] add ravenwood policy for native functions

parthsane:
Having a ravenwood policy file here will allow directly using
the perfetto sdk from android framework instead of having
a shim layer in frameworks/base which takes care of the ravenwood policy via
annotations.

### DIFF
--- a/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
+++ b/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
@@ -19,6 +19,7 @@ package dev.perfetto.sdk;
 import dalvik.annotation.optimization.CriticalNative;
 import dalvik.annotation.optimization.FastNative;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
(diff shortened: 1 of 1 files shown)


src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -187,6 +187,11 @@ public byte[] close() {
 
   private static native byte[] native_stop_session(long ptr);
 
+  // A function to support ravenwood infrastructure.
```
zezeozue, line 193: Why does this need the ravenwood call and other don't?

zezeozue (approved):

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
parthsane: The return value is the issue.

for functions returning primitives, you can just use the ignore annotation/directive to return a dummy value.

Since this returns an array, we need to have a dummy function and use the replace directive

zezeozue: Got it. Please add this comment in the code

zezeozue (approved):

PR #4922, opened 2026-02-25
sdk: Move system registration check from Android framework

parthsane:
Adds a static method `PerfettoTrace.getAttemptedSystemRegistration()` to the Java SDK. This returns true if `register(isBackendInProcess = false)` has been called in the current process.

Motivation:
This allows components to dynamically check if the Perfetto SDK has been initialized for system-wide tracing in the current process.

This optimization provides lower overhead and avoids potential SELinux issues when triggering from privileged processes like Android's system_server, which cannot easily fork-exec the trigger binary.

The heuristic of checking for *attempted* registration is sufficient for this optimization.


primiano:
can I ask how you plan to use this e2e and wha's the problem it's solving? it feels a bit of a weird API tbh. but I cannot offer better proposals if I don't understand the whole problem.

Also it's unclear how this is related with fork+exec mentioned in the PR description

PR #4933, opened 2026-02-26
ui: add support for collecting long traces via UI

LalitMaganti:
Not for landing yet, still needs a lot of testing. See commentary on b/423048711

Fixes: b/423048711


primiano:
Okay maybe this is not as bad as I thought and I should pick this up.
I am not a fan of this PR to be honest, mainly because I don't buy we need to duplicate the TracingSession class. I think code can be kept way more maintainable if we add that feature to the existing tracing session.
I'll pick it up, but I also need to sync up with @stevegolton as he might be touchign this code as part of his real time memory  inspector.

PR #4934, opened 2026-02-26
Add extra debug fields to WM Shell Transition proto

PabiGamito:


protos/perfetto/trace/android/shell_transition.proto
```diff
@@ -72,8 +74,8 @@ message ShellTransition {
   // The transition type of this transition (e.g. TO_FRONT, OPEN, CLOSE).
   optional int32 type = 13;
 
-  // The list of targets that are part of this transition.
-  repeated Target targets = 14;
+  // The list of changes that are part of this transition.
+  repeated Change changes = 14;
```
PriyankaSPX, line 78: this won't be backwards compatible so older perfetto traces will no longer be supported (and you'll need to make changes in how we convert legacy traces in Winscope)

can we not just add a new `Change` field/message instead of replaying `Target`?

PriyankaSPX: or alternatively just extend `Target` - we can rename in the UI if that's unclear to users

rukkal: This change is not breaking wire format compatibility and this new version of TP will be able to process older traces.

However, we are obviously breaking the API compatibility. However, will this actually break the winscope legacy-to-perfetto conversion @PriyankaSPX ? It shouldn't break, unless the conversion logic is accessing the ShellTransition's fields

PriyankaSPX: yeah we do access the ShellTransition's fields - it should only be a minor change to fix - there will be a few other places related to parsers/file readers/operations that you'll need to address the field name change

PabiGamito: Made the respective changes on the Winscope side. Waiting for this to merge into internal before merging there.

PR #4946, opened 2026-02-27
tp: best-fit scratch registers + DataframeRegisterCache

LalitMaganti:
Replace the slot_id-based scratch register API with a best-fit
allocation strategy: when requesting scratch of a given size, the
smallest existing free slot that fits is reused. This removes the
need for callers to manually assign slot numbers.

Extract the scope-based register caching from BytecodeBuilder into
a new DataframeRegisterCache helper keyed by (reg_type, void*),
using column/index pointer identity as the key instead of abstract
numeric indices. This makes the caching more type-safe and removes
the CreateCacheScope/ClearCacheScope API.

In TreeTransformer, heap-allocate both the BytecodeBuilder and
DataframeRegisterCache via unique_ptr so the cache's internal
reference to the builder survives moves (TreeTransformer is moved
into MovePointer by tree_conversion/tree_filter).

### DIFF
--- a/src/trace_processor/core/interpreter/bytecode_builder.cc
+++ b/src/trace_processor/core/interpreter/bytecode_builder.cc
@@ -17,6 +17,8 @@
 #include "src/trace_processor/core/interpreter/bytecode_builder.h"
 
 #include <cstdint>
+#include <limits>
+#include <optional>
 
 #include "perfetto/base/logging.h"
 #include "src/trace_processor/core/interpreter/bytecode_core.h"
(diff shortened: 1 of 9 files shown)


src/trace_processor/core/interpreter/bytecode_builder.cc
```diff
@@ -26,63 +28,78 @@
...
 namespace perfetto::trace_processor::core::interpreter {
 
-uint32_t BytecodeBuilder::CreateCacheScope() {
-  auto scope_id = static_cast<uint32_t>(scope_caches_.size());
-  scope_caches_.emplace_back();
-  return scope_id;
+std::optional<uint32_t> BytecodeBuilder::FindBestFitSlot(uint32_t size) const {
+  std::optional<uint32_t> best;
+  uint32_t best_size = std::numeric_limits<uint32_t>::max();
+  for (uint32_t i = 0; i < scratch_slots_.size(); ++i) {
```
sashwinbalaji: how about having a ordered set of unused slots by size so can quickly search the best fit?

PR #4961, opened 2026-03-01
perfetto: Speed up tools with parallelism and caching

LalitMaganti:
Key improvements:
- Parallelize GN build description creation in gen_bazel (3 gn gen
  calls now run concurrently instead of sequentially).
- Reuse the GN output directory for buildflags generation in
  gen_android_bp and gen_bazel, avoiding a redundant gn gen + ninja
  invocation.
- Run all presubmit checks concurrently via ThreadPoolExecutor.
- Use tempfile.mkdtemp for temp directories in gn_utils to avoid
  collisions when running in parallel.

Measured on a Mac (M-series), wall-clock time:
- run_presubmit: 5.6s -> 3.0s (~1.9x faster)
- gen_bazel --check-only: 3.2s -> 1.9s (~1.7x faster)

### DIFF
--- a/tools/gn_utils.py
+++ b/tools/gn_utils.py
@@ -18,6 +18,7 @@
 from __future__ import print_function
 import collections
 from compat import iteritems
+from concurrent.futures import ThreadPoolExecutor
 import errno
 import filecmp
 import json
@@ -165,7 +166,6 @@ def create_build_descriptions(gn_args_list, root=repo_root()):
     entered). The caller must call .close() or use cleanup_build_descriptions()
     when done.
     """
-  from concurrent.futures import ThreadPoolExecutor
 
   def _create(gn_args):
     bd = BuildDescription(gn_args, root)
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Friendly ping!

tools/gn_utils.py
```diff
@@ -110,13 +111,78 @@ def create_build_description(gn_args, root=repo_root()):
...
+    shutil.copy(src, os.path.join(repo_root(), target_file))
+
+
+def create_build_descriptions(gn_args_list, root=repo_root()):
+  """Like create_build_description but runs multiple GN descriptions in
+    parallel. Returns a list of BuildDescription context managers (already
+    entered). The caller must call .close() or use cleanup_build_descriptions()
+    when done.
+    """
+  from concurrent.futures import ThreadPoolExecutor
```
sashwinbalaji: nit: import at top.

tools/run_presubmit
```diff
@@ -15,6 +15,7 @@
 
 from __future__ import print_function
 import argparse
+import concurrent.futures
```
sashwinbalaji: nit: follow same as above
``` from concurrent.futures import ThreadPoolExecutor```

PR #4964, opened 2026-03-02
stdlib: add lock name to android_monitor_contention

gignat-dev:
This change adds lock name to android_ monitor_contention table: if a slice `monitor contention*` is surrounded by `${name}_lock_acquire` on left and `${name}_lock_held` on right, then we have contention on `${name}` lock (at the moment just for big locks: ams, proc, wms, pms).

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/monitor_contention.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/monitor_contention.sql
@@ -137,7 +137,8 @@ WHERE
 GROUP BY
   slice.id;
 
--- Contains parsed monitor contention slices.
+-- Contains parsed Java monitor contention slices, detailing the blocking and blocked
+-- threads, methods involved, and the duration of the contention.
 CREATE PERFETTO TABLE android_monitor_contention (
   -- Name of the method holding the lock.
   blocking_method STRING,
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/stdlib/android/monitor_contention.sql
```diff
@@ -192,8 +193,51 @@ CREATE PERFETTO TABLE android_monitor_contention (
...
+  all_slices_on_contention_tracks AS (
+    SELECT
+      s.id,
+      s.track_id,
+      s.ts,
+      s.name,
+      lag(s.name) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS prev_name,
+      lead(s.name) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS next_name
+    FROM slice AS s
+    WHERE
```
zezeozue, line 210: should we add a condition to also only do this for s.name glob 'monitor contention*'?

ditto below

zezeozue (approved):

src/trace_processor/perfetto_sql/stdlib/android/monitor_contention.sql
gignat-dev: I may be wrong, but I think the filters are applied before LEAD/LAG are computed, and this is not what we want in this case.

zezeozue: right. nvm then

PR #4968, opened 2026-03-02
ui: Improve minimap

stevegolton:


LalitMaganti:
Small drive-by since I saw this CL. The existing minimap on non-Android traces is really ugly. Espeically if it's a very simple trace and there's only a single track. In that case the whole minimap would just be a black bar.

If it's easy then can we please consider doing something nicer (especially if we can minimize the vertical space used) or even hide the minimap entirely is perfectly OK - for simple traces like that, you don't need much help navigating it.

LalitMaganti:
Please don't consider the above blocking, it was only something I was thinking about discussing with you at some point anyway but since you're here, thought it would be good to share. If it's too much work or it becomes a rabbit hole, I'd much rather some improvement lands!

PR #4978, opened 2026-03-02
Consider doFrames when extending frame timeline

IQ01660:
For blocking calls per frame duration metrics
we extend the frame timeline to cover cases
where blocking calls starting within a certain frame extend beyond it.

- Problem:

Until now frames were extended till the next available frame in the actual frametine. This however can be problematic when UI thread gets vsync updates and makes blocking calls for a period of time where no draws actually happen.
This means the next actual frame might be much later. This leads to skewed metrics that collect blocking calls way into the future after a specific frame had already ended.

- Solution:

A better approach could be to consider both the next actual frame and the next Choreographer#doFrame slice, picking the earliest option. This updated definition of extended frame should make the metric more meaningful and stable.

Test: tools/diff_test_trace_processor.py --name-filter ".\*android.\*" out/android/trace_processor_shell

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/frame_blocking_calls/blocking_calls_aggregation.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/frame_blocking_calls/blocking_calls_aggregation.sql
@@ -65,7 +65,12 @@ SELECT
   cuj_name,
   -- Earliest of (Next Frame Start, Next doFrame Start).
   -- Fallback to original ts_end only if both are NULL.
-  coalesce(min(next_do_slice_after_frame, next_frame_start), original_ts_end) AS ts_end
+  coalesce(
+    min(next_do_slice_after_frame, next_frame_start),
+    next_do_slice_after_frame,
+    next_frame_start,
+    original_ts_end
+  ) AS ts_end
 FROM _frames_with_extension_options
 ORDER BY
   frame_id;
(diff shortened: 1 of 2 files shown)


LalitMaganti:
I have a general comment here: it seems like we are shoving more and more sysui specific things into what was meant to be a generic "blocking call" stdlib module which applied no matter the app. I am feeling quite uncomfortable with this apporach. Is it possible to refactor the sysui things out into their own modules?

nicomazz:
> I have a general comment here: it seems like we are shoving more and more sysui specific things into what was meant to be a generic "blocking call" stdlib module which applied no matter the app. I am feeling quite uncomfortable with this apporach. Is it possible to refactor the sysui things out into their own modules?

I wouldn't consider this a "sysui" specific change: the scenario above (doFrame that doesn't result in a draw) could happen in any app, and I think it makes sense to change the boundaries to consider until the next doFrame instead of the next "doFrame that results in a draw" (that would result in also an actual timeline slice entry).

From discussions in the comments, a possible good point is why we don't always and only consider the doFrame for boundaries: I would say that considering both doFrame and actualTimeline slices makes this moer generic, as it would also work for processes that generate the actual timeline despite not being based on "doFrame" based frame rendering.

LalitMaganti:
Ack thanks for the commentary. Lgtm.

PR #4982, opened 2026-03-03
stdlib: memory_breakdown expose schema

gignat-dev:
This change refactors memory_breakdown.sql:
- moves the computation of zygote_adjusted_value before the left span join, and added it together with the raw counter value
- marks tracks with spikes bigger than 100MiB (instead of removing)
- removed the trace limits filter (will filter only for metrics)
- added table schemas
- renamed tables

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
@@ -232,7 +232,7 @@ CREATE VIRTUAL TABLE _memory_breakdown_mem_oom_span_join USING SPAN_LEFT_JOIN (
 -- NOTE: Some tracks may have a spike greater than 100MiB. This can be legitimate or
 -- an accounting issue: see b/418231246 for more details.
 CREATE PERFETTO TABLE android_process_memory_intervals_by_oom_bucket (
-  -- Unique identifier of this memory interval.
+  -- Id.
   id LONG,
   -- The start timestamp of the interval.
   ts TIMESTAMP,


src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,22 +18,30 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
 
 INCLUDE PERFETTO MODULE counters.intervals;
 
--- Create a table containing intervals of memory counters values, adjusted to process lifetime.
-CREATE PERFETTO TABLE _memory_breakdown_mem_intervals_raw AS
+-- A table of process memory counter intervals, clipped to the process lifetime.
+-- Provides a comprehensive view of memory usage, including raw values,
+-- zygote-adjusted values, and a flag for tracks with a spike bigger than 100 MiB.
+CREATE PERFETTO TABLE android_process_memory_intervals (
+  -- Timestamp of the memory counter sample.
```
aMayzner: Would it be possible to also return counter id?

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,22 +18,30 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
...
+  -- Duration of the sample.
+  dur DURATION,
+  -- The unique process id.
+  upid JOINID(process.id),
+  -- The name of the memory counter track.
+  track_name STRING,
+  -- The id of the memory counter track.
+  track_id JOINID(track.id),
+  -- Whether the track has a spike greater than 100MiB.
+  has_spike_gt_100mib BOOL,
```
aMayzner: Do we have to mark the whole track or maybe we can just filter out those spikes?

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,22 +18,30 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
...
+-- zygote-adjusted values, and a flag for tracks with a spike bigger than 100 MiB.
+CREATE PERFETTO TABLE android_process_memory_intervals (
+  -- Timestamp of the memory counter sample.
+  ts TIMESTAMP,
+  -- Duration of the sample.
+  dur DURATION,
+  -- The unique process id.
+  upid JOINID(process.id),
+  -- The name of the memory counter track.
+  track_name STRING,
```
aMayzner: So RSS and do on?

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: We mark the whole track as having "large spikes" (useful for metrics)

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Yeah, they track_name are 'mem.rss.anon', 'mem.rss.file', 'mem.swap', etc.

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Good idea

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
aMayzner: Put this column at the end and write in a comment that this is related to the bug, and this column might be removed in the future

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -34,6 +34,8 @@ CREATE PERFETTO TABLE android_process_memory_intervals (
   track_id JOINID(track.id),
   -- Whether the track has a spike greater than 100MiB.
   has_spike_gt_100mib BOOL,
+  -- The counter id
+  counter_id JOINID(counter.id),
```
aMayzner: This column should be first I think, to maintain the standard id, ts, dur. It's not unique, is it?

gignat-dev: It is unique in `android_process_memory_intervals`, but might not be unique in `android_process_memory_intervals_by_oom_bucket` because of the breakdown by oom_bucket

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
aMayzner: memory_track_name please, so that people know that's the important string.

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -223,6 +229,8 @@ CREATE PERFETTO TABLE android_process_memory_intervals_by_oom_bucket (
   bucket STRING,
   -- Whether the track has a spike greater than 100MiB.
   has_spike_gt_100mib BOOL,
+  -- The counter id
+  counter_id JOINID(counter.id),
```
aMayzner, line 255: As above

gignat-dev: Here, the counter_id might not unique because we break the memory intervals by the oom_bucket. That is why we use `row_number()` as id

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,24 +18,35 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
...
+  -- Duration of the sample.
+  dur DURATION,
+  -- The unique process id.
+  upid JOINID(process.id),
+  -- The name of the memory counter track.
+  memory_track_name STRING,
+  -- The id of the memory counter track.
+  track_id JOINID(track.id),
+  -- The value of the memory counter.
+  value LONG,
```
aMayzner, line 50: In bytes? Please write it.

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done, renamed counter_id to id and moved it to the top

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: As previously mentioned, counter_id in this table `android_process_memory_intervals_by_oom_bucket` is not unique

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,24 +18,35 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
 
 INCLUDE PERFETTO MODULE counters.intervals;
 
--- Create a table containing intervals of memory counters values, adjusted to process lifetime.
-CREATE PERFETTO TABLE _memory_breakdown_mem_intervals_raw AS
+-- A table of process memory counter intervals, clipped to the process lifetime.
+-- Provides a comprehensive view of memory usage, including raw values,
+-- zygote-adjusted values, and a flag for tracks with a spike bigger than 100 MiB.
```
aMayzner: The spikes I would add as a -- NOTE:, as this is kinda a detail and. Maybe the same with zygote, another note in which you explain what do you mean by it

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,24 +18,35 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
...
+CREATE PERFETTO TABLE android_process_memory_intervals (
+  -- The counter id
```
aMayzner: `The id of the memory counter value`

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,24 +18,35 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
...
+  id JOINID(counter.id),
+  -- Timestamp of the memory counter sample.
+  ts TIMESTAMP,
+  -- Duration of the sample.
```
aMayzner: `How long this memory track had this value`.

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,24 +18,35 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
...
+  dur DURATION,
+  -- The unique process id.
```
aMayzner: The process whose memory is being measured - or sth like that. Can we maybe get a process name too?

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done, added 2 notes in the table description: about zygote adjustment and the spikes.

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done, added name and improved comment

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -18,24 +18,46 @@ INCLUDE PERFETTO MODULE android.oom_adjuster;
...
+  pid JOINID(process.pid),
+  -- The name of the memory counter track (e.g. 'mem.rss.anon').
+  memory_track_name STRING,
+  -- The id of the memory counter track.
+  track_id JOINID(track.id),
+  -- The value of the memory counter in bytes.
+  value LONG,
+  -- The value of the memory counter in bytes, adjusted with the zygote's memory usage.
+  zygote_adjusted_value LONG,
+  -- Whether the track has a spike greater than 100MiB.
```
aMayzner, line 53: Please mention that this column might disappear in the future, when the bug is fixed!

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
```diff
@@ -142,82 +215,65 @@ WHERE
...
+-- NOTE: For 'mem.rss.anon', 'mem.swap', 'mem.rss.file', and 'mem.heap' tracks, we
+-- subtract the Zygote's average memory usage. This provides a better estimate
+-- of the child process's unique memory usage by accounting for the
+-- baseline memory inherited from the Zygote.
+--
+-- NOTE: Some tracks may have a spike greater than 100MiB. This can be legitimate or
+-- an accounting issue: see b/418231246 for more details.
+CREATE PERFETTO TABLE android_process_memory_intervals_by_oom_bucket (
+  -- Unique identifier of this memory interval.
+  id LONG,
```
aMayzner, line 236: Just call it in id. You can create your own ids.

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: I think we might still want to keep it just for devices with the old version of trace processor that contains the bug?
@zezeozue wdyt?

zezeozue: yes. + there are several other unfixed bugs it appears. like b/487412953

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Done

src/trace_processor/perfetto_sql/stdlib/android/memory/memory_breakdown.sql
gignat-dev: Ack

PR #4985, opened 2026-03-03
ui: Configurable table links in query results

stevegolton:
Add a dropdown to the query results page results that allows users to select what the id columns of the query results are interpreted as in order to customize where the links in the 'id' column navigate to.

The first item in the list is 'Auto-Detect' which uses the same heuristic approach based on the presence of various columns in the result set. Nothing changes there. This just allows users to select a different table if they so please.

Note: This is a temporary measure - once we can parse queries and extract lineage information from them, we will no longer need to have this dropdown (probably), or at the very least, the 'Auto-Detect' will be much more accurate.

The change has been made in the query page and the results tab on the timeline page.

image


aMayzner:
I think we can go further and maybe set it as a per column properly? We might as well have multiple columns that can be a table link. Even slice.id and slice.parent_id, both of them should be links.

stevegolton:
> I think we can go further and maybe set it as a per column properly? We might as well have multiple columns that can be a table link. Even slice.id and slice.parent_id, both of them should be links.

Yes, but I want it to be discoverable. I don't want to bury it in some column menu. This is a good first step and something that a lot of people have been asking for. We can add more functionality later.

aMayzner:
> Yes, but I want it to be discoverable. I don't want to bury it in some column menu. This is a good first step and something that a lot of people have been asking for. We can add more functionality later.

I think you should say what column is being interpreted at least. If I have android_frames table, with 5 slice.id columns, I want to see the string make sense. "Interpret as" on it's own, is kinda hard to understand the context

stevegolton:
Added an indication that it's the 'id' column that's being interpreted

PR #4988, opened 2026-03-03
tp: Add intervals_create C++ Perfetto SQL function

aMayzner:
Replace the O(n*m) SQL-based `create_slices` implementation with an O(n+m) C++ two-pointer algorithm. The structured query generator now calls `_intervals_create!` macro (backed by `__intrinsic_interval_create`) instead
of inlining a (inefficient) subquery.

## Changes

- Add `SortedTimestamps` type and `__intrinsic_timestamp_set_agg` aggregate
  to collect and sort timestamps from SQL into a C++ vector
- Add `__intrinsic_interval_create` scalar function that takes two
  `SortedTimestamps` pointers and produces a dataframe with `ts`/`dur` columns
- Add `_intervals_create!` PerfettoSQL macro in `intervals/create_slices.sql`
  that wires the aggregate + scalar function together
- Update `StructuredQueryGenerator::CreateSlices` to emit the macro call
  and register the module dependency
- Add 8 diff tests covering basic, interleaved, partial match, empty,
  equal-timestamp, unsorted, and no-match scenarios

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/create_intervals.cc
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/create_intervals.cc
@@ -47,8 +47,8 @@ using ColType = dataframe::AdhocDataframeBuilder::ColumnType;
 // For each start timestamp, finds the minimum end timestamp that is strictly
 // greater than the start. Multiple starts may be paired with the same end.
 // Starts with no matching end are excluded from the output.
-struct CreateIntervals : public sqlite::Function<CreateIntervals> {
-  static constexpr char kName[] = "__intrinsic_create_intervals";
+struct IntervalCreate : public sqlite::Function<IntervalCreate> {
+  static constexpr char kName[] = "__intrinsic_interval_create";
   static constexpr int kArgCount = 2;
 
   struct UserData {
@@ -83,7 +83,7 @@ struct CreateIntervals : public sqlite::Function<CreateIntervals> {
     const auto& end_ts = ends->timestamps;
 
     // Two-pointer matching: O(n + m).
-    // Both arrays are already sorted (guaranteed by TimestampSetAgg).
+    // Both arrays are already sorted (guaranteed by ORDER BY in the SQL macro).
     // For each start, we find the first end strictly greater than it.
     // Since starts are sorted, the end pointer only advances forward.
     size_t end_idx = 0;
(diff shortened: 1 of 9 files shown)


aMayzner:
> Having something which is called "slices" in a module called "intervals" is confusing and a layering violation:
> 
> * intervals = ts + dur
> * slices = ts + dur + name
> 
> Please fix the layering here. Either put it in slices module if you're creaitng slicdes or call it _create_intervals if you're creating intervals.

Done. I renamed everything except data explorer/sq to create intervals.

src/trace_processor/perfetto_sql/stdlib/intervals/create_intervals.sql
```diff
@@ -0,0 +1,67 @@
...
+-- Example:
+-- ```
+-- SELECT * FROM _create_intervals!(
+--   (SELECT ts FROM starts_table),
+--   (SELECT ts FROM ends_table),
+--   ts,
+--   ts
+-- )
+-- ```
+CREATE PERFETTO MACRO _create_intervals(
```
LalitMaganti: _interval_create to match the pattern by _interval_intersect.

src/trace_processor/perfetto_sql/stdlib/intervals/create_intervals.sql
```diff
@@ -0,0 +1,67 @@
...
+    -- Table or subquery containing start timestamps.
+    starts_table TableOrSubquery,
+    -- Table or subquery containing end timestamps.
+    ends_table TableOrSubquery,
+    -- Name of the timestamp column in the starts table.
```
LalitMaganti: Can we please rename the input instead and hardcode ts as the column. This approach doesn't scale.

src/trace_processor/perfetto_sql/intrinsics/functions/type_builders.cc
```diff
@@ -435,6 +436,37 @@ struct IntervalTreeIntervalsAgg
...
+    data.timestamps.push_back(sqlite::value::Int64(argv[0]));
+  }
+
+  static void Final(sqlite3_context* ctx) {
+    auto raw_agg_ctx = AggCtx::GetContextOrNullForFinal(ctx);
+    if (!raw_agg_ctx) {
+      return sqlite::result::Null(ctx);
+    }
+    auto& ts = raw_agg_ctx.get()->data.timestamps;
+    std::sort(ts.begin(), ts.end());
```
LalitMaganti: Why are we sorting here instead of asking SQL to sort, eliding the sort if the data is already sorted?

src/trace_processor/perfetto_sql/intrinsics/functions/create_intervals.cc
```diff
@@ -0,0 +1,117 @@
...
+    auto* ends = sqlite::value::Pointer<SortedTimestamps>(
+        argv[1], SortedTimestamps::kName);
+
+    std::vector<std::string> col_names{"ts", "dur"};
+    std::vector<ColType> col_types{ColType::kInt64, ColType::kInt64};
+
+    dataframe::AdhocDataframeBuilder builder(
+        col_names, GetUserData(ctx)->pool,
+        dataframe::AdhocDataframeBuilder::Options{
+            col_types, dataframe::NullabilityType::kSparseNullWithPopcount});
```
LalitMaganti, line 72: Generally always prefer DenseNull for functions returning dataframes.

LalitMaganti (approved):

src/trace_processor/perfetto_sql/intrinsics/functions/create_intervals.cc
aMayzner: Done

PR #4991, opened 2026-03-03
de: Cleanup the navigation sidepanel

aMayzner:
Aesthetic cleanup. 
From:
image

To: 
image

### DIFF
--- a/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/builder.scss
+++ b/ui/src/plugins/dev.perfetto.ExplorePage/query_builder/builder.scss
@@ -206,7 +206,7 @@
   gap: 12px;
 
   span {
-    font-size: 11px;
+    font-size: var(--pf-font-size-xs);
     font-weight: 600;
     text-transform: uppercase;
     letter-spacing: 0.08em;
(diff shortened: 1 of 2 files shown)


ui/src/plugins/dev.perfetto.ExplorePage/query_builder/builder.scss
```diff
@@ -189,51 +190,191 @@
...
+
+  .pf-nav-list-item__arrow {
+    flex-shrink: 0;
+    opacity: 0;
+    transition:
+      opacity 0.15s ease,
+      transform 0.15s ease;
+
+    .pf-icon {
+      font-size: 18px;
```
stevegolton: Can we use the font-size css variable from common.scss

aMayzner:
> Looks better overall - like the labels and the general layout a lot.
> 
> A few points:
> 
> 1. Each category of card appears to have different hover colors / right icons / dropshadows, despite the fact that all of them do the same type of thing (open a graph).
> 2. The different accent color on the 'smart graph' is a little inconsistent (accent vs primary) - looks like someone guessed the blue value and got it wrong. Maybe just make them all primary or accent.

Fixed

PR #4992, opened 2026-03-03
tp: Remove descendant_slices() false positives

aMayzner:
Fix false positives in `descendant_slice()` where an instant child slice at
the exact boundary between its parent and a subsequent sibling ("uncle") was
incorrectly returned as a descendant of the uncle. The timestamp-based filter
alone cannot distinguish these cases, so a parent_id validation pass is added.

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/table_functions/descendant.cc
+++ b/src/trace_processor/perfetto_sql/intrinsics/table_functions/descendant.cc
@@ -37,6 +37,25 @@
 namespace perfetto::trace_processor {
 namespace {
 
+// Walks the parent chain of |candidate| to check whether |ancestor_id| is an
+// ancestor. Stops early when the depth drops to |ancestor_depth| or below.
+bool IsAncestor(const tables::SliceTable& slices,
+                tables::SliceTable::ConstRowReference candidate,
+                SliceId ancestor_id,
+                uint32_t ancestor_depth) {
+  for (auto id = candidate.parent_id(); id;) {
+    if (*id == ancestor_id) {
+      return true;
+    }
+    auto ref = slices.FindById(*id);
+    if (!ref || ref->depth() <= ancestor_depth) {
+      return false;
+    }
+    id = ref->parent_id();
+  }
+  return false;
+}
+
 bool GetDescendantsInternal(
     const tables::SliceTable& slices,
     tables::SliceTable::ConstCursor& cursor,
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/intrinsics/table_functions/descendant.cc
```diff
@@ -66,8 +66,20 @@ bool GetDescendantsInternal(
     ts_upper_bound = std::numeric_limits<int64_t>::max();
   }
   cursor.SetFilterValueUnchecked(3, ts_upper_bound);
+
+  // The timestamp filter can produce false positives at boundaries (e.g. a
```
LalitMaganti: Creating an unordered set here in an extreme hot path is not a good idea. Can we please do a stiatic check based on timestamps similar to what the above + SliceTracker do.

aMayzner: A pure timestamp check can't fully eliminate false positives because at boundaries, timestamps are ambiguous (a child of a sibling at the exact same ts is indistinguishable from a child of the target slice by timestamps alone). I have another idea though, one that uses the fact that we know that for each depth the slices are non overlapping. I will see if it works. I doesn't require a set initialisation.

aMayzner: SliceTracker's timestamp rules determine whether a slice could be a child (i.e., whether the parent is still on the stack), but when multiple slices share a boundary timestamp, there are multiple candidates on the stack, and only parent_id tells you which one was the actual parent.

The new per-depth vector approach keeps the parent_id check but avoids the unordered_set allocation.

LalitMaganti: The problem was the allocation, not the fact you were using unordered_set... You still have an allocation.

This is a critical path for many chrome queries. Just try a chrome trace and you'll see the regression.

IMO it should be possible to determine this statically without needing any side-memory regardless of the datastructure.

aMayzner: No allocations now. It works and should be fast because of the outside check for ts equality.

src/trace_processor/perfetto_sql/intrinsics/table_functions/descendant.cc
```diff
@@ -66,8 +65,34 @@ bool GetDescendantsInternal(
...
+
+  // The timestamp filter can produce false positives at the start boundary
+  // (candidate.ts == start.ts) where a child of a slice ending at start.ts
+  // shares the same timestamp. For such candidates, walk the parent chain to
+  // verify ancestry. For candidates strictly inside the interval (ts >
+  // start.ts), same-depth non-overlapping guarantees they are true descendants.
+  int64_t start_ts = start_ref->ts();
   for (cursor.Execute(); !cursor.Eof(); cursor.Next()) {
-    row_numbers_accumulator.emplace_back(cursor.ToRowNumber());
+    auto row_num = cursor.ToRowNumber();
```
LalitMaganti, line 95: OK nice this is a lot closer to what I was expecting. But I think it would be a lot easier to read if this was factored out into a helper function :)

PR #4994, opened 2026-03-03
de: Add tabs to Data Explorer

aMayzner:
Add multi-tab support to the Data Explorer, allowing users to work on multiple independent graph explorations simultaneously. Each tab maintains its own state (nodes, layout, history, materialization) and tabs persist across page reloads via localStorage and permalinks.

## Changes

- Refactored `ExplorePage` component to render multiple tabs using the existing
  `Tabs` widget, with per-tab services (query execution, cleanup, history)
- Added `explore_tabs_storage.ts` for persisting tab layout to localStorage
  using zod schema validation
- Updated plugin (`index.ts`) to manage an array of tabs instead of a single
  state, with debounced saves and backward compatibility for v1 single-graph
  permalinks and recentGraphsStorage
- Fixed SVG marker ID collisions in `nodegraph.ts` when multiple NodeGraph
  instances exist (e.g. across tabs) by using instance-unique IDs and scoped
  DOM queries
- Added `onTabDblClick` callback to the `Tabs` widget for tab renaming
- Registered a `beforeunload` handler to flush pending saves on page unload

### DIFF
--- a/ui/src/widgets/nodegraph.ts
+++ b/ui/src/widgets/nodegraph.ts
@@ -53,6 +53,8 @@ import {Icon} from './icon';
 import {PopupMenu} from './menu';
 import {classNames} from '../base/classnames';
 import {Icons} from '../base/semantic_icons';
+import {assertExists} from '../base/assert';
+import {shortUuid} from '../base/uuid';
 
 // Default height estimate for labels (used for box selection calculations)
 const DEFAULT_LABEL_MIN_HEIGHT = 30;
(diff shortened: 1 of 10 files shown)


ui/src/widgets/nodegraph.ts
```diff
@@ -1055,7 +1068,10 @@ export function NodeGraph(): m.Component<NodeGraphAttrs> {
         ? `[data-node="${nodeId}"] .pf-port[data-port="${portType}-${portIndex}"]`
         : `[data-node="${nodeId}"] [data-port="${portType}-${portIndex}"] .pf-port`;
 
-    const portElement = document.querySelector(selector);
+    // Scope to this NodeGraph instance to avoid matching elements from other
+    // instances (e.g. hidden tabs with the same node IDs).
+    const scope = canvasElement ?? document;
```
stevegolton: Maybe `assertExists(canvasElement)` - we don't ever want to fall back to document by accident

aMayzner:
> Looks good in general: A few notes:
> 
> * I get a brief flicker when opening an new tab, like the nodegraph is being resized manually in JS or something?
> * Odd choice of icon for a nodegraph - mabye 'account_tree' or 'graph_X' or one of the more boring document icons?
> * What is the default nodegraph I get displayed when I create a new tab? Seems quite opinionated?

1 and 2 will be fixed in the next commit
3. It is heavily opinionated, but it's pretty, simple, non aggressive. I will change it in the future, but for now it looks good.

PR #5000, opened 2026-03-03
Record empty power rails packets as an error

simonmacm:
If we fail to read power rails data for whatever reason we'll write an empty packet.

It would be good to be able to keep track of whether this happens, since Westworld power rails pulls are quite unreliable - do we ever see similar problems in Perfetto?


src/trace_processor/importers/proto/android_probes_module.cc
```diff
@@ -116,6 +116,10 @@ ModuleResult AndroidProbesModule::TokenizePacket(
 
     parser_.ParseRailDescriptor(evt);
 
+    if (!evt.has_energy_data()) {
+      context_->storage->IncrementStats(stats::power_rail_empty_packet);
```
LalitMaganti: Please use the import log tracking infrastructure instead of using stats. Also you should return Handled.

simonmacm: Done.

PR #5011, opened 2026-03-04
ui: Highlight current page in the sidebar

stevegolton:
If the current page has a corresponding link in the sidebar that link will be highlighted. This helps ground users as to which page they're currently on and helps navigation.


aMayzner:
image

This gap seems like a bug

stevegolton:
> image
> This gap seems like a bug

Fixed

PR #5016, opened 2026-03-04
ci: Fix playwright report CSS

stevegolton:


stevegolton:
The problem was I was seeing broken CSS when opening  the playwright HTML report from the CI - e.g. https://storage.googleapis.com/perfetto-ci-artifacts/gh-22673935466-1-ui/ui-test-artifacts/index.html

The issue was that since the CI builds of the UI are hosted from the same domain, they have access to the same local storage settings as the UI. Some local storage setting from the UI was breaking the CSS theme colors.

PR #5018, opened 2026-03-04
heapprofd: expose a reallocarray hook

rsavitski:
Allow bionic to find a `heapprofd_reallocarray`. Additionally, short-circuit
the overflow case as it is defined to cause a clean error.

Bug: 489692215


cferris1000:
This does fix the loading of the shared library. I don't know if there are other issues besides this though.

PR #5019, opened 2026-03-04
ui: Pin per thread Wattson power tracks

wusamuel6:
Pin the top threads by power within the window of interest (e.g. app_startup, jank window, etc.) when it's requested via the URL. This gives a more informative view of the most potentially problematic threads in the window of interest. Per SysUI team's request.

Test:
http://localhost:10000/#!/viewer?local_cache_key=<MY_LOCAL_KEY>#dev.perfetto.PinAndroidPerfMetrics:metrics=perfetto_wattson_markers_rails_cpu-estimated_mws-mean, http://localhost:10000/#!/viewer?local_cache_key=<MY_LOCAL_KEY>#dev.perfetto.PinAndroidPerfMetrics:metrics=perfetto_wattson_apps_rails_cuj_estimated_mws
Bug: 483752560

### DIFF
--- a/ui/src/plugins/org.kernel.Wattson/index.ts
+++ b/ui/src/plugins/org.kernel.Wattson/index.ts
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import {addWattsonThreadTrack} from './wattson_thread_utils';
+import {App} from '../../public/app';
 import {createAggregationTab} from '../../components/aggregation_adapter';
 import {
   BaseCounterTrack,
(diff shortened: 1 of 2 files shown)


ui/src/plugins/org.kernel.Wattson/index.ts
```diff
@@ -36,9 +38,29 @@ import {
...
   static readonly id = `org.kernel.Wattson`;
   static readonly dependencies = [SchedPlugin];
+  private static windowsOfInterest: string[] = [];
+
+  static onActivate(_app: App): void {
+    const cujRegex = /perfetto_wattson_(markers|trace|apps|app_startup)/g;
+    const windows = new Set<string>();
+
+    // Split by common delimiters in the hash/URL to find potential metrics
+    for (const component of location.hash.split(/[?&]/)) {
```
stevegolton: We actually already have a mechanism plumbed in for this.

If you pass urls prefixed with the plugin's name - like this:

```
dev.perfetto.PinAndroidPerfMetrics:metrics=perfetto_wattson_markers_rails_cpu-estimated_mws-mean
```

They'll be translated to an object and passed as the second argument to onActivate().

```ts
static onActivate(app: App, args: RouteArgs) {
  console.log(args.metrics); // Should be 'perfetto_wattson_markers_rails_cpu-estimated_mws-mean'
}
```

wusamuel6: Thanks for the discussion, resolved.

PR #5024, opened 2026-03-05
tp: Make Summarizers truly independent

aMayzner:
Make summarizer instances truly independent by incorporating a unique id into materialized table names. Previously, multiple summarizers sharing the same TraceProcessor would generate colliding table names (e.g. `_exp_mat_0`),
causing CREATE TABLE failures. Each summarizer now produces namespaced tables like `_exp_mat_{id}_{seq}`.

### DIFF
--- a/include/perfetto/trace_processor/trace_processor.h
+++ b/include/perfetto/trace_processor/trace_processor.h
@@ -249,8 +249,7 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessor : public TraceProcessorStorage {
   // EXPERIMENTAL: Creates a new Summarizer instance for managing lazy
   // materialization of structured queries. On success, |out| is populated with
   // the new instance and ownership is transferred to the caller.
-  virtual base::Status CreateSummarizer(const std::string& id,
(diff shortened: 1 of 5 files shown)


include/perfetto/trace_processor/trace_processor.h
```diff
@@ -249,7 +249,8 @@ class PERFETTO_EXPORT_COMPONENT TraceProcessor : public TraceProcessorStorage {
   // EXPERIMENTAL: Creates a new Summarizer instance for managing lazy
   // materialization of structured queries. On success, |out| is populated with
   // the new instance and ownership is transferred to the caller.
-  virtual base::Status CreateSummarizer(std::unique_ptr<Summarizer>* out) = 0;
```
LalitMaganti, line 252: Playing devil's advocate, why not make this a proeprt of trace processor where there's an automatic incrementing id which is passed to the Summarizer. Is there any reason it needs to become part of the API?

aMayzner: This also works, even if it pushes the responsibility to TP. If we are fine with it, the commit implemented this change

PR #5028, opened 2026-03-05
tp: Fix HPROF importer for ahat parity

zezeozue:
Match ahat's computation of self_size, native_size, reachability,
root_type, and qualified field names so that the raw table output
is identical to ahat for the same .hprof file.

 - Fixed reference_set_id to match the row index of the first reference.
 - Classified reference types (Weak, Soft, Phantom, Finalizer) to
   correctly skip "referent" fields during reachability analysis,
   matching ahat's behavior.
 - Implement root type precedence and shortest-path root distance
   calculation to align with proto heap graph expectations.


zezeozue:
Will reupload with @google email

PR #5031, opened 2026-03-05
rust-sdk: fix nested message naming and improve macro formatting

dreveman:
Fix the Rust protobuf bindings generator to prepend parent type names
to nested messages, matching the existing behavior for enums. This
prevents naming collisions when multiple parent types have nested
messages with the same name (e.g., Row.Value.Null vs Row.Dimension.Null
now generate RowValueNull vs RowDimensionNull).

Also improve pb_msg!/pb_enum! macro formatting by emitting single-line
invocations when appropriate. This makes empty messages more
compact: `pb_msg!(MessageName {});`


zytyz:
Should we add an underscore when prepending parent types since the names are already camel case, e.g. `GpuCounterEvent_GpuCounter`? Otherwise the naming still isn't unique (although it is unlikely that would happen).

dreveman:
> Should we add an underscore when prepending parent types since the names are already camel case, e.g. `GpuCounterEvent_GpuCounter`? Otherwise the naming still isn't unique (although it is unlikely that would happen).

I considered that but decided against it as struct names in Rust supposed to be UpperCamelCase and we'd have to suppress a bunch of format warnings. And it might look a bit weird in code that uses this crate.

primiano (approved):

PR #5038, opened 2026-03-06
ui: AndroidAnr expand process on missing main thread

gignat-dev:
This change treats the case when an app that ANRed doesn't have a main thread track associated with it (really common).
In this case, we will scroll to the process group of the ANRing app, expand it and select the ANR region. If the main thread track exists, this track is selected.

### DIFF
--- a/ui/src/plugins/com.android.AndroidAnr/index.ts
+++ b/ui/src/plugins/com.android.AndroidAnr/index.ts
@@ -248,8 +248,8 @@ export default class AndroidAnr implements PerfettoPlugin {
     ts: bigint,
     dur: bigint,
   ) {
-    const startTime = Time.fromRaw(ts);
-    const endTime = Time.fromRaw(ts + dur);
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.android.AndroidAnr/index.ts
```diff
@@ -172,7 +181,10 @@ export default class AndroidAnr implements PerfettoPlugin {
       dur: LONG_NULL,
       main_thread_track_id: NUM,
     });
+
     if (!it.valid()) {
+      // Main thread not found, fallback to process track
```
zezeozue: Is there any difference if we always just open the process track even with main thread? since the main thread is always the first track inside process anyways?

gignat-dev: Yeah, makes sense

gignat-dev: Done, now we scroll to process group and then select main thread track if exists

PR #5050, opened 2026-03-08
tp: Support shared resources in Gecko trace importer

LalitMaganti:
Refactor the Gecko tokenizer to handle profiles where
frames and stacks are in a top-level "shared" object
rather than duplicated per-thread. Extract InternFrame
as a reusable method and split thread processing into
separate frames/stacks and samples phases so shared
callsites can be built once and reused across threads.


LalitMaganti:
[gecko_shared_resources.json.gz](https://github.com/user-attachments/files/25824428/gecko_shared_resources.json.gz)

SHA256 needs to be added

aMayzner (approved):

PR #5059, opened 2026-03-09
tp: stdlib: Update android_input_events to parse eventTime

ivan-chong:
The track event for UnwantedInteractionBlocker was recently changed to include the eventTime. This PR updates the parsing of inputEventId whilst maintaining backward compatibility, as well as adding and propagating eventTime to android_input_events.

Bug: 454761692

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/input.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/input.sql
@@ -70,13 +70,23 @@ ORDER BY
   event_seq;
 
 CREATE PERFETTO TABLE _input_read_time AS
+WITH
+  _extracted_input_read_args AS (
+    SELECT
+      name,
+      str_split(str_split(str_split(name, 'id=', 1), ',', 0), ')', 0) AS input_event_id,
+      str_split(str_split(name, 'eventTime=', 1), ')', 0) AS event_time_str,
+      ts AS read_time
+    FROM slice
+    WHERE
+      name GLOB 'UnwantedInteractionBlocker::notifyMotion*'
+  )
 SELECT
   name,
-  str_split(str_split(name, '=', 1), ')', 0) AS input_event_id,
-  ts AS read_time
-FROM slice
-WHERE
-  name GLOB 'UnwantedInteractionBlocker::notifyMotion*';
+  input_event_id,
+  cast_int!(event_time_str) AS event_time,
+  read_time
+FROM _extracted_input_read_args;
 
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/stdlib/android/input.sql
```diff
@@ -354,7 +365,9 @@ CREATE PERFETTO TABLE android_input_events (
   -- Vsync Id associated with the input. Null if an input event has no associated frame event.
   frame_id LONG,
   -- Indicates if the frame association was speculative rather than exact based on id match.
-  is_speculative_frame BOOL
+  is_speculative_frame BOOL,
+  -- Timestamp when the input event actually occurred.
+  event_time LONG
```
aMayzner: The type should be TIMESTAMP

aMayzner: Same with read_time(). Also, I would think about having input_event_id be an actual ID type

ivan-chong: Thanks! I saw this and kept it as LONG to be consistent with read_time which was pre-existing, and planned to look into this for a future refactor of this table. Will change this to TIMESTAMP here and look into the input_event_id in a future PR, wdyt?

aMayzner: I wouldn't overcomplicate it. If this is the place where you introduce the ID (sorted, unique) than just change it to the ID and all of the places that use it should refer to it as JOINID({table_name}.{id_col_name}). If it's a significant change then sure, do it in a different PR.

ivan-chong: The `input_event_id` is a hex string was pre-existing alongside the `android_input_events` table. The naming of `input_event_id` and `android_input_events` table is also a bit misleading as `input_event_id` is not actually a unique ID within the `android_input_events` table as the same value can be present across multiple rows as the same "unique" input event can be sent through multiple `event_channel`s. The actual value itself is also not a typical incrementing ID and is created by random generation. This means it is not sorted nor is it technically unique. Do you know if it still possible to have an ID type here?

I've also now changed event_time and read_time to now be TIMESTAMP, wdyt?

aMayzner: For the future, maybe it would be useful to actually have an ID for each input event, or at least correspond it to sth from a counter/slice table to be able to fetch more context. For now though it's a great change, LGTM

PR #5065, opened 2026-03-10
ui: Add tab renaming

aMayzner:
Add an way to rename a tab from a Tabs widget with double click.

### DIFF
--- a/ui/src/widgets/tabs.ts
+++ b/ui/src/widgets/tabs.ts
@@ -43,13 +43,10 @@ export interface TabsAttrs {
   onTabChange?(key: string): void;
   // Called when a tab's close button is clicked.
   onTabClose?(key: string): void;
-  // Called when a tab is double-clicked. When onTabRename is also set,
-  // double-click triggers rename mode in addition to calling this callback.
-  onTabDblClick?(key: string): void;
   // Called when a tab's title is renamed via inline editing. When set, tabs
-  // with a string title become renamable on double-click. If the input is
-  // cleared (empty after trim), the rename is cancelled and this callback is
-  // not fired.
+  // with a string title become renamable on double-click (tabs with non-string
+  // titles are not affected). If the input is cleared (empty after trim) or
+  // Escape is pressed, the rename is cancelled and this callback is not fired.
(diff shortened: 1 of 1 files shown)


ui/src/widgets/tabs.ts
```diff
@@ -147,7 +163,32 @@ class TabHandle implements m.ClassComponent<TabHandleAttrs> {
...
+            oninput: (e: InputEvent) => {
+              const target = e.target as HTMLInputElement;
+              onRenameInput?.(target.value);
+            },
+            onkeydown: (e: KeyboardEvent) => {
+              if (e.key === 'Enter') {
+                onRenameCommit?.();
+                e.preventDefault();
+              } else if (e.key === 'Escape') {
+                onRenameCancel?.();
```
stevegolton, line 180: Pressing esc doesn't actually cancel the rename. renameInputValue is still set and onblur is still triggered leading to onRenameCommit() still being called with the pending name.

ui/src/widgets/tabs.ts
```diff
@@ -43,8 +43,14 @@ export interface TabsAttrs {
   onTabChange?(key: string): void;
   // Called when a tab's close button is clicked.
   onTabClose?(key: string): void;
-  // Called when a tab is double-clicked (e.g. for renaming).
+  // Called when a tab is double-clicked. When onTabRename is also set,
+  // double-click triggers rename mode in addition to calling this callback.
   onTabDblClick?(key: string): void;
```
stevegolton: I don't think this is used anywhere and the interaction with onTabRename is a little confusing. Perhaps remove it, YAGNI?

ui/src/widgets/tabs.ts
```diff
@@ -43,8 +43,14 @@ export interface TabsAttrs {
...
+  // Called when a tab's title is renamed via inline editing. When set, tabs
+  // with a string title become renamable on double-click. If the input is
+  // cleared (empty after trim), the rename is cancelled and this callback is
+  // not fired.
+  onTabRename?(key: string, newTitle: string): void;
```
stevegolton, line 50: Perhaps leave a note about how this only works if the tab's content is a string?

PR #5073, opened 2026-03-11
ui: Add context menu on tabs

aMayzner:
- Add optional context menu on tab in Tabs
- Add it to the demo
- Use it for the data explorer copy the graph option

### DIFF
--- a/ui/src/assets/widgets/tabs.scss
+++ b/ui/src/assets/widgets/tabs.scss
@@ -68,21 +68,16 @@
     }
   }
 
-  &__tab-menu {
-    display: flex;
-    align-items: center;
-    opacity: 0;
-    pointer-events: none;
+  &__tab-menu-btn {
+    visibility: hidden;
 
-    &--open {
-      opacity: 1;
-      pointer-events: auto;
+    &.pf-active {
+      visibility: visible;
     }
   }
 
-  &__tab:hover &__tab-menu {
-    opacity: 1;
-    pointer-events: auto;
+  &__tab:hover &__tab-menu-btn {
+    visibility: visible;
   }
 
   &__tab-title {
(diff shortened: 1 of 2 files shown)


ui/src/widgets/tabs.ts
```diff
@@ -186,6 +195,28 @@ class TabHandle implements m.ClassComponent<TabHandleAttrs> {
...
+          m(
+            PopupMenu,
+            {
+              trigger: m(Button, {
+                compact: true,
+                icon: Icons.ContextMenuAlt,
+              }),
+              position: PopupPosition.Bottom,
+              onChange: (isOpen: boolean) => {
+                this.menuOpen = isOpen;
```
stevegolton: Remove this - use pf-active in CSS (as mentioned above)

ui/src/widgets/tabs.ts
```diff
@@ -186,6 +195,28 @@ class TabHandle implements m.ClassComponent<TabHandleAttrs> {
             onclick: (e: Event) => e.stopPropagation(),
           })
         : m('.pf-tabs__tab-title', children),
+      menuItems !== undefined &&
+        m(
+          '.pf-tabs__tab-menu',
```
stevegolton: I don't think you should need this wrapper div. Just apply the class to the button directly

ui/src/assets/widgets/tabs.scss
```diff
@@ -68,6 +68,23 @@
     }
   }
 
+  &__tab-menu {
```
stevegolton: When the popup is open, it'll apply `pf-active` to the trigger (the button).

You can use the presence of this class to keep it visible when the popup is showing.

Also I'd use visibility: hidden and visibility: visible rather than opacity to hide the button.

Issue #5095, opened 2026-03-12
vkQueueSubmit and multiple command buffers


olehkuznetsov:
Currently trace processor drops all but the first command buffer

It would be great if it doesn't.

LalitMaganti:
PRs welcome to fix. All of this code was contributed externally, the Perfetto team doesn't have the knowledge to know what the right fixes here are.

olehkuznetsov:
Thank you, I'm already working on a PR.

PR #5096, opened 2026-03-12
tp: Support multiple command buffers for vkQueueSubmit

olehkuznetsov:
Adds all command buffers as Args

Fixes: [5095](https://github.com/google/perfetto/issues/5095)


src/trace_processor/importers/proto/gpu_event_parser.cc
```diff
@@ -833,9 +847,21 @@ void GpuEventParser::ParseVulkanApiEvent(int64_t ts, ConstBytes blob) {
...
+          uint32_t count = 0;
+          for (auto it = event.vk_command_buffers(); it; ++it, ++count) {
+            if (count == 0) {
+              // Record first command buffer in a separate column for legacy
+              // compatibility.
+              inserter->AddArg(command_buffer_id_,
+                               Variadic::UnsignedInteger(*it));
+            }
+            if (count < command_buffer_handle_ids_.size()) {
+              inserter->AddArg(command_buffer_handle_ids_[count],
```
LalitMaganti: Why are not using the native support for arg arrays in the argstracker instead of reinventing a subset of that functionality here?

olehkuznetsov: I've updated it to use GetNextArrayEntryIndex and GetNextArrayEntryIndex.

Is this native support you mean or there is even better one?

src/trace_processor/importers/proto/gpu_event_parser.cc
```diff
@@ -833,9 +833,22 @@ void GpuEventParser::ParseVulkanApiEvent(int64_t ts, ConstBytes blob) {
         inserter->AddArg(pid_id_, Variadic::Integer(event.pid()));
         inserter->AddArg(tid_id_, Variadic::Integer(event.tid()));
         if (event.has_vk_command_buffers()) {
-          inserter->AddArg(command_buffer_id_,
-                           Variadic::Integer(static_cast<int64_t>(
-                               *event.vk_command_buffers())));
+          uint32_t count = 0;
+          for (auto it = event.vk_command_buffers(); it; ++it, ++count) {
+            if (count == 0) {
+              // Compatibility: first buffer as a flat "command_buffer"
```
LalitMaganti, line 840: it's quite confusing to have both an array and a primitive field with the same name. It will break e.g. globs, json export of args (which the UI uses) etc.

Can we please rename the array field with a different name if we're keep this for backcompat.

olehkuznetsov: renamed array to "command_buffers"

PR #5110, opened 2026-03-13
ui: Add pf-test-volatile class which can be used to hide certain elements during screenshot tests

stevegolton:
Some elements are volatile and unpredictable - e.g. the app version number in the sidebar and the printout of how long queries took to run in results tables.

We used to mask out these values manually in playwright tests, but that was error prone.

This patch introduces a new concept, where elements that are known to be volatile can be assigned a special class `pf-test-volatile`, which in a no-op in production but hidden when testing. This flips the dependency, and while admittedly this approach leaks test knowledge into the production DOM, it is simply a lot cleaner and easier to work with.

Also hiding is better than masking, because masking covers the element with a specific color, and if that element changes width, that size of the mask will also change. This led to a practice of masking out the closest stable ancestor of the volatile element instead, which could end up masking out other important elements.


stevegolton:
> Would it be possible to have a counter, like with "any" to have some control over this class? I can imagine it leaking into some things that aren't actually that volatile and marking them as such just hides the inherent flakiness.

I see your point, though I'm not sure it's worth the extra machinery... yet anyway. If it becomes a problem we can add one.

Besides since the DOM is a tree, count doesn't really reflect how much in being obscured. One could move the class up to the document root which would render all the tests pointless, yet the count wouldn't change.

PR #5118, opened 2026-03-13
ui: Add stacking and series hovering to line charts

stevegolton:
Add two new features to line chart:
- Stacking of multi series data
- Hovering over a point highlights the entire series and dims the others

image


ui/src/components/widgets/charts/chart_option_builder.ts
```diff
@@ -26,7 +26,7 @@ import type {EChartsCoreOption} from 'echarts/core';
  * Configuration for an axis in a chart.
  */
 export interface AxisConfig {
-  readonly type: 'value' | 'category' | 'log';
+  readonly type: 'value' | 'category' | 'log' | 'time';
```
aMayzner, line 29: Time is a value, no? This is what value is - not categorical columns

stevegolton: Discussed offline. This is an internal type and it's needed for the echarts stacking hack. It's not exposed on any external API.

PR #5122, opened 2026-03-13
perfetto: add test for SMB scraping in TraceBufferV2

LalitMaganti:
In write_into_file traces with long flush periods (e.g. 1h),
SMB scraping can re-introduce chunks whose index entries were
already evicted by ring buffer wraparound. When this happens,
CopyChunkUntrusted treats the chunk as entirely new (since
index_.find() misses), resetting num_fragments_read to 0.
ReadNextTracePacket then re-reads all fragments, producing
duplicate packets that cause out-of-order errors in the trace
sorter (e.g. "trace_sorter.cc: out of order events").

The root cause is that DeleteNextChunksFor removes index entries
(and their read progress) when the ring buffer wraps, but SMB
scraping doesn't change the producer's shared memory state. So
the next scrape finds the same chunk, and without the index entry,
there's no record that fragments were already consumed.

We don't want to fix this in TraceBufferV1 because its quite risky
*and* we're anyway trying to move to v2 due to other, unfixable bugs
with v1.

Just add a regression test to V2 to ensure that this bug doesn't reappear.

### DIFF
--- a/src/tracing/service/trace_buffer_v2_unittest.cc
+++ b/src/tracing/service/trace_buffer_v2_unittest.cc
@@ -2345,4 +2345,106 @@ TEST_F(TraceBufferV2Test, Overwrite_SizeDiffLessThanChunkHeader) {
   ASSERT_THAT(ReadPacket(), ElementsAre(FakePacketFragment(32 - 16, 'c')));
 }
 
+// -------------------------------------------
+// Re-scrape after ring-buffer eviction tests
+// -------------------------------------------
+
+// V2 is immune to the rescrape-after-eviction bug because it tracks
+// last_chunk_id_consumed per writer and rejects chunks with IDs that have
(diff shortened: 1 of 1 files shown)


LalitMaganti:
After discussing with Ioannis, I think in practice it might not make much sense to land the v1 side of this patch given the other (unfixable) bug where write into file appears to cause data loss.

The best we can do might be to simply migrate everyone to V2 ASAP and land the tests in this CL.

primiano:
Yeah honestly given where we are, let's sort out the remaining mysteries on v2 (ashwin figured out some repros) and let's jsut migrated over to v2, rather than adding more risks of destabilizing by making changes to  v1

LalitMaganti:
Changed this PR to just be a test against V2 instead of fixing anything with V1.

sashwinbalaji (approved):

PR #5128, opened 2026-03-13
tp: add common_flags, query subcommand, and dispatcher

LalitMaganti:
## Summary
- Shared infrastructure: common_flags.h/.cc (GlobalOptions, BuildConfig, SetupTraceProcessor, LoadTraceFile with symbolization/deobfuscation)
- Query subcommand with -f, -c, -i, -W, -p flags
- ClassicSubcommand: moves entire classic path into a Subcommand, uses shared infra
- Thin dispatcher: finds subcommand or falls back to classic
- Integration tests for query subcommand and classic backcompat

Part 2/7 of the subcommand CLI implementation (RFC-0018).

## Stack
1. #5134 — SubcommandContext + GetLongOptions() interface
2. **#5128** ← this PR — common_flags + query + ClassicSubcommand + dispatcher
3. #5129 — repl + serve subcommands
4. #5130 — summarize + export subcommands
5. #5131 — metrics subcommand
6. #5132 — help system + file collision + classic edge case tests
7. #5133 — BUILD.gn consolidation

### DIFF
--- a/src/trace_processor/shell/subcommand.h
+++ b/src/trace_processor/shell/subcommand.h
@@ -74,6 +74,11 @@ class Subcommand {
  public:
   virtual ~Subcommand();
 
+  Subcommand(const Subcommand&) = delete;
+  Subcommand& operator=(const Subcommand&) = delete;
+  Subcommand(Subcommand&&) = delete;
+  Subcommand& operator=(Subcommand&&) = delete;
+
   // The name of the subcommand as it appears on the command line
   // (e.g. "query", "export").
   virtual const char* name() const = 0;
@@ -86,6 +91,9 @@ class Subcommand {
 
   // Runs the subcommand. |ctx| provides access to shared resources.
   virtual base::Status Run(const SubcommandContext& ctx) = 0;
+
+ protected:
+  Subcommand() = default;
 };
 
 // Result of FindSubcommandInArgs(). If |subcommand| is non-null, a subcommand


src/trace_processor/shell/common_flags.h
```diff
@@ -0,0 +1,129 @@
...
+  OPT_GLOBAL_OVERRIDE_STDLIB,
+  OPT_GLOBAL_REGISTER_FILES_DIR,
+  OPT_GLOBAL_METATRACE_BUFFER_CAPACITY,
+  OPT_GLOBAL_METATRACE_CATEGORIES,
+};
+
+// getopt_long entries for global flags. Subcommands should append these
+// to their own option arrays. The 'm' short option is for --metatrace.
+// Note: does NOT include the terminating {nullptr,0,nullptr,0} sentinel.
+#define GLOBAL_LONG_OPTIONS                                              \
```
primiano: why is this a define rather than a  `const option[]` defined in the .cc file?

src/trace_processor/shell/query_subcommand.cc
```diff
@@ -0,0 +1,193 @@
...
+    if (option == 'p') {
+      perf_file = optarg;
+      continue;
+    }
+
+    PrintUsage(argv[0]);
+    return option == 'h' ? 0 : 1;
+  }
+
+  if (query_file.empty() && query_string.empty()) {
```
primiano: WDYT about adding something along the lines of:

if (query_file == "-") -> read from stdin. it can be nice to pipe things through cat

src/trace_processor/shell/subcommand.cc
```diff
@@ -26,18 +29,26 @@ FindSubcommandResult FindSubcommandInArgs(
     int argc,
     char** argv,
     const std::vector<Subcommand*>& subcommands,
-    const std::vector<std::string>& flags_with_arg) {
+    const std::vector<Subcommand*>& all_subcommands) {
+  // Build the set of long flags that consume the next argument, derived
```
primiano: TBH a bit dubious about all this logic (I trust it, but TBH i can't fully think about all the implications).

In my mind it would have been cleaner if you had a layer that does the classic -> modern translation of the cmdline, rather than trying to handle both here.  I feel it's nicer for debugging as well.

src/trace_processor/trace_processor_shell.cc
```diff
@@ -870,339 +640,411 @@ TraceSummaryOutputSpec::Format GetSummaryOutputFormat(
...
-  // Enable metatracing as soon as possible.
-  if (!options.metatrace_path.empty()) {
-    metatrace::MetatraceConfig metatrace_config;
-    metatrace_config.override_buffer_size = options.metatrace_buffer_capacity;
-    metatrace_config.categories = options.metatrace_categories;
-    tp->EnableMetatrace(metatrace_config);
-  }
+    base::TimeNanos t_query_start = base::GetWallTimeNs();
+    if (!options.pre_metrics_v1_path.empty()) {
+      auto status =
```
primiano: how come you undid all the RETURN_IF_ERROR?

primiano:
Some other comment after looking at the other CLs. My theory is that a lot of boilerplate (and potentials future bugs) can be avoided if:

1. Each command returns a base::Status rather than int (do you actually use the int for anything that is needed for backwards compat? I think we always return 1 in case of error)

2. Rather than repeating the getopt in each subcommand, what if each subcommand exposed, as part of its interface, the list of flags it takes , with helps and specs, so there is only one place doing the getopt at the top level. This would remove the need of maintaining PrintUsage, and reduce a lot of boilerplate all over the place.

I'm thinking of something where each subcommand defines an array of cmdline flags, with a lambda for doing the parsing, plus some helpers for the trivial cases (e.g. StoreString, StoreInt) something like:

```
const Flags** SummarizeSubcommand::flags() const {
  return {
    Flag("spec",  kRequired, [this] (const char* optarg) {this->spec_ = optarg;},
     Flag("metrics-v2", kRequired, [this] (const char* optarg) {this->metrics_v2_ = optarg;},
  };
}
```

but then you can just simplify as 

 Flag("spec",  kRequired, StoreString(this, spec_)


pretty sure a good AI agent can help figuring out the details

src/trace_processor/shell/query_subcommand.cc
LalitMaganti: Done.

LalitMaganti:
Changed the implementation as you suggested. I'm holding off on doing anything with the classic mode until *after* we have all subcommands implemented as it will be a lot easier with less merge conflicts.

src/trace_processor/shell/query_subcommand.h
```diff
@@ -0,0 +1,44 @@
...
+
+class QuerySubcommand : public Subcommand {
+ public:
+  const char* name() const override;
+  const char* description() const override;
+  std::vector<FlagSpec> GetFlags() override;
+  base::Status Run(const SubcommandContext& ctx) override;
+
+ private:
+  std::string query_file_;
```
primiano, line 36: disallow copy and assign? otherwise the flag mechanism that takes pointers will break.
or maybe if you disablein the base class shoudl be enough

src/trace_processor/shell/subcommand.h
```diff
@@ -43,14 +81,11 @@ class Subcommand {
   // A short one-line description shown in help output.
   virtual const char* description() const = 0;
 
-  // Runs the subcommand. |ctx| provides access to shared resources like the
-  // platform interface. |argc| and |argv| are the original command line with
-  // the subcommand name removed (argv[0] is the program name).
-  // Returns 0 on success, non-zero on failure.
-  virtual int Run(const SubcommandContext& ctx, int argc, char** argv) = 0;
+  // Returns the flags this subcommand accepts.
```
primiano, line 89: disallow copy and move

primiano (approved):

PR #5129, opened 2026-03-13
tp: add interactive and server subcommands

LalitMaganti:
## Summary
- Interactive subcommand (interactive SQL shell with -W flag, renamed from repl per RFC-0018)
- Server subcommand (RPC server with http/stdio positional modes, renamed from serve per RFC-0018)
- Uses new declarative flag system from parent branch
- All sources folded into single :subcommand BUILD.gn target
- Integration tests for both subcommands

## Test plan
- [x] All 33 TraceProcessorShellIntegrationTest tests pass
- [x] Builds cleanly with no warnings


src/trace_processor/shell/repl_subcommand.cc
```diff
@@ -0,0 +1,117 @@
...
+#include "perfetto/ext/trace_processor/trace_processor_shell.h"
+#include "perfetto/trace_processor/trace_processor.h"
+#include "src/trace_processor/shell/common_flags.h"
+#include "src/trace_processor/shell/interactive.h"
+#include "src/trace_processor/shell/metatrace.h"
+
+namespace perfetto::trace_processor::shell {
+
+const char* ReplSubcommand::name() const {
+  return "repl";
```
primiano: how about "interactive" or just "shell"? I feel a good number of people won't know what "repl" is

primiano (approved):

src/trace_processor/shell/repl_subcommand.cc
LalitMaganti: You're reviewing an old patch. It's already "interactive"

primiano: No it was a stale comment that got stuck in vscode when I reviewed yesterday. i saw already the updated one, just the comment was queued yesterday. ignore

PR #5132, opened 2026-03-13
tp: add help system, file collision detection, classic tests

LalitMaganti:
## Summary
- Top-level help (`-h`) showing all subcommands with examples
- `help <command>` for per-subcommand flag reference
- `--help-classic` for legacy flat-flag reference
- File collision detection hint when positional arg matches both a subcommand and a file on disk
- Restored classic CLI integration tests

## Test plan
- [x] All 45 TraceProcessorShellIntegrationTest tests pass
- [x] Builds cleanly with no warnings

### DIFF
--- a/ui/src/assets/perfetto.scss
+++ b/ui/src/assets/perfetto.scss
@@ -66,6 +66,7 @@
 @import "widgets/overlay_container";
 @import "widgets/popup";
 @import "widgets/resize_handle";
+@import "widgets/scorecard";
 @import "widgets/section";
 @import "widgets/segmented_buttons";
 @import "widgets/select";
--- a/ui/src/assets/widgets/scorecard.scss
+++ b/ui/src/assets/widgets/scorecard.scss
@@ -1,0 +1,53 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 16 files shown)


src/trace_processor/trace_processor_shell.cc
```diff
@@ -162,7 +162,59 @@ struct CommandLineOptions {
   std::vector<std::string> raw_metric_v1_extensions;
 };
 
-void PrintUsage(char** argv) {
+void PrintSubcommandHelp(const char* argv0) {
```
primiano, line 165: shouldn't a lot of this be dynamically built out of the information in each Subcommand? feels more scalable

primiano (approved):
LG with a comment

src/trace_processor/trace_processor_shell.cc
LalitMaganti: This is the top level documentation. For this I want to be hihgly opinionated on what is shown as this is what 99% of AI and people will see. So this is less a reference and more "I think you should know about X".

primiano: But do we list all available subcommands somewhere?

LalitMaganti: This is that list.

PR #5152, opened 2026-03-16
tp: improve help text for subcommand CLI

LalitMaganti:
Add usage_args() and detailed_help() to Subcommand base class so each
subcommand provides its own positional arg description and detailed
help text. Rework the top-level help to use a clean command table,
common flags table, examples section, and classic interface note.
Fix PrintSubcommandUsage to use per-command usage_args instead of
hardcoded "[SQL]", and add FILE* parameter for output stream control.
Update metrics description to reference summarize --metrics-v2.

### DIFF
--- a/ui/src/plugins/dev.perfetto.DataExplorer/dashboard/dashboard.scss
+++ b/ui/src/plugins/dev.perfetto.DataExplorer/dashboard/dashboard.scss
@@ -129,6 +129,10 @@
   display: flex;
   flex-direction: column;
   user-select: none;
+  // Smooth settling when snapped away from a divider.
+  transition:
+    left 0.1s ease-out,
+    top 0.1s ease-out;
 
   // Custom width/height via CSS custom properties set inline.
   &--custom-width {
@@ -143,6 +147,7 @@
     opacity: 0.6;
     z-index: 100;
     cursor: grabbing;
+    transition: none;
   }
 
   // Position resize handles at the edges of the card.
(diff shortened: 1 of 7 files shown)


src/trace_processor/shell/common_flags.cc
```diff
@@ -50,33 +50,44 @@ namespace perfetto::trace_processor::shell {
...
 
 }  // namespace
 
-void PrintSubcommandUsage(const char* argv0, Subcommand* cmd) {
-  fprintf(stderr, "Usage: %s %s [FLAGS] <trace_file> [SQL]\n\n", argv0,
-          cmd->name());
-  fprintf(stderr, "%s\n\n", cmd->description());
+std::string FormatSubcommandUsage(const char* argv0, Subcommand* cmd) {
+  std::string out;
+  char buf[256];
```
primiano: base::StackString buf<256>("Usage , ...."...)

src/trace_processor/shell/common_flags.cc
```diff
@@ -223,7 +234,11 @@ base::Status ParseFlags(Subcommand* cmd,
       break;
 
     if (opt == '?') {
-      return base::ErrStatus("Unknown flag. Use --help for usage.");
+      // getopt_long already printed a diagnostic to stderr; mark the
+      // status so callers know not to print the message again.
+      base::Status s = base::ErrStatus("Unknown flag");
```
primiano, line 237: Honestl given that this is the only case where you do the has_printed_error dance, isn't more pragmatic to just make it ErrStatus(""), with a comment explaiing getopt already prints the error.
The other side will just print a \n, which is fine.

LalitMaganti: we have a PERFETTO_CHECK against empty error message unfortunately. And printing anything else is confusing as there's two, potentially slightly different formatted error messages.

Issue #5060, opened 2026-03-09
Is it feasible to make process/thread tracks behave like generic event tracks as an opt-in feature


anakryiko:
I'm aware of special treatment of process/thread tracks in terms of parent/child relationship, sibling merging behavior, child ordering, etc. I assume there are some very good backwards compatibility preserving reasons for that.

But if backwards compat is not an issue, would it be possible to make thread/process tracks behave just like normal track in all regards, except for keeping that (extremely useful, mind you, which is one of the reasons I still haven't switch to generic tracks just yet) thread/process information attached.

Sorry if this was extensively explained before for *why not as an opt-in?*, but I haven't found that. Thanks!

LalitMaganti:
The short answer is somewhere between: "no" and "it's extremely difficult to the point of being a no".

The long answer is that the "process container" and "thread container" in Perfetto is not a concept owned by any single data source (e.g. track event). Instead, it's shared and multiplexed amongst all 10s of different data formats, data producers etc that Perfetto supports. This means it's very difficult to support all the track event features at the thread/process level without also considering its impact on the other data sources and situations like "what happens when someone merges a JSON trace and a track event trace with these annotations" etc.

It would help if you could give concrete examples of thigns you want to do but don't work. I'm not convinced that thigns like e.g. sibling merging event make sene at the thread/process level - you can some many (most) of the problems by creating child tracks of the process/thread tracks and then applying merging there.

LalitMaganti:
At least for sorting, I think we could have a separate "sort" enum at the ThreadDescriptor/ProcessDescriptor level, which operates as a hint to the process/thread sorting logic, which we could piggy back of similar to what I did for JSON - https://github.com/google/perfetto/pull/3273

We cannot use the same sorting protos as we do for "regular" track event tracks for the reasons described previously.

Another thing I think I want to add support for: attaching key-value pairs to TrackDescriptors which become source_arg_set_id args at the proto level and we can display one/some/all of those on the track shell. I think this would be more productive for "I want to see the a number similar to the pid/tid" etc.

anakryiko:
[...]

PR #5154, opened 2026-03-17
tp: add In filter support to TypedCursor and optimize In bytecode

LalitMaganti:
Add SetFilterValueListUnchecked to TypedCursor allowing callers to
pass a pointer+size array of FilterValue for In filters without
allocation. Plumb this through the codegen'd ConstCursor/Cursor.

Optimize the In bytecode by pre-building lookup structures during
CastFilterValueList instead of rebuilding on every Execute():
- For dense Id/Uint32: BitVector (built once, not per-call)
- For large sparse integer/string lists: FlatHashMapV2 for O(1)
- For small lists (<=16): linear scan (cache-friendly)

The lookup is stored as a variant in CastFilterValueListResult,
replacing the separate value_list field.

Bug: https://github.com/google/perfetto/issues/5060

### DIFF
--- a/src/trace_processor/core/interpreter/interpreter_types.h
+++ b/src/trace_processor/core/interpreter/interpreter_types.h
@@ -175,8 +175,7 @@ struct CastFilterValueListResult {
 
   // The lookup used by In. For small lists (<=16 elements), we keep the
   // FlexVector and do a linear scan. For dense Id/Uint32, a BitVector.
-  // For large sparse lists, a HashLookup. If we build a BitVector or
-  // HashLookup, the FlexVector is not kept.
+  // For large sparse lists, a HashLookup.
   using Lookup = std::variant<ValueList, BitVector, HashLookup>;
 
   static CastFilterValueListResult Valid(Lookup l) {


src/trace_processor/core/interpreter/interpreter_types.h
```diff
@@ -170,22 +171,26 @@ struct CastFilterValueListResult {
...
                                  FlexVector<double>,
                                  FlexVector<StringPool::Id>>;
+  using HashLookup = base::FlatHashMapV2<int64_t, bool>;
 
-  static CastFilterValueListResult Valid(ValueList v) {
-    return CastFilterValueListResult{CastFilterValueResult::Validity::kValid,
-                                     std::move(v)};
+  // The lookup used by In. For small lists (<=16 elements), we keep the
+  // FlexVector and do a linear scan. For dense Id/Uint32, a BitVector.
+  // For large sparse lists, a HashLookup. If we build a BitVector or
```
sashwinbalaji: nit: this part seems unnecessary " If we build a BitVector or
  // HashLookup, the FlexVector is not kept."

PR #5158, opened 2026-03-17
tp: add indexed path to FilterIn bytecode

LalitMaganti:
## Summary

Extend the `In` bytecode (renamed to `FilterIn<T, N>`) with an optional
`index_register` that enables an indexed path using the sorted
permutation vector:

- When `index_register` is absent: linear scan with hash/bitvector/linear
  lookup (same behavior as the old `In` bytecode).
- When `index_register` is present: binary search for each IN-list value
  in O(k log N). For large IN-lists (k > 64), falls back to linear scan
  with hash lookup since the O(m log m) sort cost dominates. The
  threshold was determined empirically via constant-sweeping benchmarks.

Also:
- Correctly compare strings by content (via `string_pool->Get()`)
  rather than raw `StringPool::Id` in the indexed binary search path.
- Avoid allocating unused popcount/null_bv registers for NonNull columns.
- Add constant-sweeping benchmarks (`BENCHMARK_CONSTANT_SWEEPING=1`) to
  calibrate the binary-search vs linear-scan threshold.

## Stack
1. #5154 - tp: add In filter support to TypedCursor and optimize In bytecode
2. **#5158 - tp: add indexed path to FilterIn bytecode** (this PR)
3. #5155 - tp: migrate experimental_slice_layout to use In filter on track_id

Bug: https://github.com/google/perfetto/issues/5060

### DIFF
--- a/src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
+++ b/src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
@@ -1172,6 +1172,47 @@ IndexToStorageIndex(uint32_t index,
   }
 }
 
+// Binary-searches a sorted index for all entries whose resolved storage
+// value equals |target|. Returns the [lb, ub) range of matching index
+// entries.
+//
+// |target| should be a NullTermStringView for String columns (i.e. a
+// pre-resolved string_pool->Get result) or the raw value for other types.
+// This avoids repeated string_pool->Get calls inside the comparators.
+template <typename T, typename N, typename Resolved>
+inline PERFETTO_ALWAYS_INLINE std::pair<uint32_t*, uint32_t*> IndexEqualRange(
+    uint32_t* begin,
+    uint32_t* end,
+    const Resolved& target,
+    const typename T::cpp_type* data,
+    const BitVector* const* null_bv,
+    const Slab<uint32_t>* popcnt,
+    const StringPool* string_pool) {
+  auto* lb =
+      std::lower_bound(begin, end, target, [&](uint32_t idx, const Resolved&) {
+        uint32_t si = IndexToStorageIndex<N>(idx, null_bv, popcnt);
+        if (si == std::numeric_limits<uint32_t>::max())
+          return true;
+        if constexpr (std::is_same_v<T, String>) {
+          return string_pool->Get(data[si]) < target;
+        } else {
+          return data[si] < target;
+        }
+      });
+  auto* ub =
+      std::upper_bound(lb, end, target, [&](const Resolved&, uint32_t idx) {
+        uint32_t si = IndexToStorageIndex<N>(idx, null_bv, popcnt);
+        if (si == std::numeric_limits<uint32_t>::max())
+          return false;
+        if constexpr (std::is_same_v<T, String>) {
+          return target < string_pool->Get(data[si]);
+        } else {
+          return target < data[si];
+        }
+      });
+  return {lb, ub};
+}
+
 template <typename T, typename N>
 inline PERFETTO_ALWAYS_INLINE void IndexedFilterEq(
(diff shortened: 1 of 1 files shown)


src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
```diff
@@ -1481,6 +1221,420 @@ inline PERFETTO_ALWAYS_INLINE void IndexedFilterEq(
...
+    uint32_t* write = dest.b;
+    for (const auto& cmp_val : vl) {
+      auto* lb = std::lower_bound(
+          index->b, index->e, cmp_val,
+          [&](uint32_t idx, const ValElem& target) {
+            uint32_t si = IndexToStorageIndex<N>(idx, null_bv, popcnt);
+            if (si == std::numeric_limits<uint32_t>::max())
+              return true;
+            if constexpr (std::is_same_v<T, String>) {
+              return string_pool->Get(data[si]) < string_pool->Get(target);
```
sashwinbalaji: can we cache the value of string_pool->Get(target), to avoid calling it repeatedly.

src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
```diff
@@ -1481,6 +1221,420 @@ inline PERFETTO_ALWAYS_INLINE void IndexedFilterEq(
...
+      return;
+    }
+  }
+  const auto& vl = base::unchecked_get<VL>(cast_result.value_list);
+  if (!vl.empty() && vl.size() <= kFilterInKeyScanThreshold) {
+    dest.e = FilterInLinearScan<T>(data, source.b, source.e, dest.b, vl);
+    return;
+  }
+  const auto& hm = base::unchecked_get<HM>(cast_result.hash_map);
+  dest.e = FilterInHashMap<T>(data, source.b, source.e, dest.b, hm);
```
sashwinbalaji: Should we also sort dest here, as it's possible source is index

src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
```diff
@@ -1481,6 +1221,420 @@ inline PERFETTO_ALWAYS_INLINE void IndexedFilterEq(
...
+    }
+
+    const auto* data =
+        state.ReadStorageFromRegister<T>(bytecode.arg<B::storage_register>());
+    const Slab<uint32_t>* popcnt =
+        state.MaybeReadFromRegister(bytecode.arg<B::popcount_register>());
+    const BitVector* const* null_bv =
+        state.MaybeReadFromRegister(bytecode.arg<B::null_bv_register>());
+    const StringPool* string_pool = state.string_pool;
+
```
sashwinbalaji, line 1556: nit: use equal_range ?

src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
LalitMaganti: Fixed this.

src/trace_processor/core/interpreter/bytecode_interpreter_impl.h
```diff
@@ -1481,6 +1221,470 @@ inline PERFETTO_ALWAYS_INLINE void IndexedFilterEq(
...
+}
+
+// Attempts the indexed binary search path for FilterIn. Looks up the
+// index register; if present and the IN list is small enough, binary-
+// searches the sorted index for each value and populates dest. Returns
+// true if binary search was performed, false if the caller should use
+// a different strategy.
+//
+// TODO(lalitm): since the index is sorted, galloping search could replace
+// repeated std::lower_bound for better locality.
```
sashwinbalaji, line 1522: +1

PR #5162, opened 2026-03-17
Allow heapprofd_standalone_client to build for Chrome

Sentimentron:
This allows libheaprofd_standalone_client.so to be included in the chrome_public_apk target for tracing Chrome's custom allocators.

Main changes:
* Re-plumbs the GN build to remove/simplify dependencies for Chromium.
* Adds a "lib" prefix to `kSkipMaps`.

AI disclosure: I used Gemini CLI to get started and find the main things that needed to change. This PR is derived from that, alongside human review and cleanup.

### DIFF
--- a/gn/BUILD.gn
+++ b/gn/BUILD.gn
@@ -424,8 +424,13 @@ if (enable_perfetto_trace_processor_linenoise) {
 # Only used by src/profiling in standalone and android builds.
 if (enable_perfetto_heapprofd || enable_perfetto_traced_perf) {
   group("libunwindstack") {
-    public_configs = [ "//buildtools:libunwindstack_config" ]
-    public_deps = [ "//buildtools:libunwindstack" ]
+    if (build_with_chromium && is_android) {
+      public_configs = [ "//third_party/libunwindstack:libunwindstack_config" ]
+      public_deps = [ "//third_party/libunwindstack" ]
+    } else {
+      public_configs = [ "//buildtools:libunwindstack_config" ]
+      public_deps = [ "//buildtools:libunwindstack" ]
+    }
   }
 }
 
--- a/src/profiling/memory/BUILD.gn
+++ b/src/profiling/memory/BUILD.gn
@@ -135,12 +135,16 @@ shared_library("heapprofd_api_noop") {
 
   # This target does absolutely nothing, so we do not want to depend on
   # liblog.
-  configs -= [ "//gn/standalone:android_liblog" ]  # nogncheck
+  if (perfetto_build_standalone) {
+    configs -= [ "//gn/standalone:android_liblog" ]  # nogncheck
+  }
   sources = [ "client_api_noop.cc" ]
 }
 
 shared_library("heapprofd_client_api") {
-  configs -= [ "//gn/standalone:android_liblog" ]  # nogncheck
+  if (perfetto_build_standalone) {
+    configs -= [ "//gn/standalone:android_liblog" ]  # nogncheck
+  }
   if (perfetto_build_with_android) {
     cflags = [ "-DPERFETTO_ANDROID_ASYNC_SAFE_LOG" ]
   } else {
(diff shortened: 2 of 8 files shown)


primiano:
I'm happy to look at this. I just am a bit delayed due to traveling in CN right now. i'll try to get this later today.
I skimmed through that and seems sensible

gn/BUILD.gn
```diff
@@ -424,8 +424,13 @@ if (enable_perfetto_trace_processor_linenoise) {
 # Only used by src/profiling in standalone and android builds.
 if (enable_perfetto_heapprofd || enable_perfetto_traced_perf) {
   group("libunwindstack") {
-    public_configs = [ "//buildtools:libunwindstack_config" ]
-    public_deps = [ "//buildtools:libunwindstack" ]
+    if (build_with_chromium && is_android) {
+      public_configs = [ "//third_party/libunwindstack:libunwindstack_config" ]
+      public_deps = [ "//third_party/libunwindstack" ]
```
primiano, line 429: so this is likely to cause problems in the rolls.
libunwind stack is subjected to breaking changes every now and then. when it happens, we have to catch up with upstream (Android). But this means whenver we do, the chromium roll will fail.

It's the biggest worry I have about this PR. This is going to doom us to get stuck on the chromium rolls.
I think the best ay forward would be to create an indirection layer in the unwinder, and allow chrome to plug in its unwinder (you can use your own copy of libunwindstack, and that's better because there is no direct coupling)

This dependency is going to be a time-bomb unfortunately

src/profiling/memory/BUILD.gn
```diff
@@ -135,12 +135,16 @@ shared_library("heapprofd_api_noop") {
 
   # This target does absolutely nothing, so we do not want to depend on
   # liblog.
-  configs -= [ "//gn/standalone:android_liblog" ]  # nogncheck
+  if (perfetto_build_standalone) {
```
primiano, line 138: (here and below) well also `|| build_with_android) no? (or !build_with_chromium to be more conservative)

src/profiling/memory/wire_protocol.h
```diff
@@ -47,7 +47,7 @@ constexpr size_t kMaxRegisterDataSize =
               sizeof(uint64_t) * unwindstack::ARM64_REG_LAST,
               sizeof(uint32_t) * unwindstack::X86_REG_LAST,
               sizeof(uint64_t) * unwindstack::X86_64_REG_LAST,
-              sizeof(uint64_t) * unwindstack::RISCV64_REG_COUNT});
+              sizeof(uint64_t) * unwindstack::RISCV64_REG_MAX});
```
primiano, line 50: this is preceisely the problem I was describing. You are effetively trying to undo https://android-review.git.corp.google.com/c/platform/external/perfetto/+/2918494 likely becuase your libunwindstack version is behind upstream.

Unfortunately this creates a upstream XOR your_version situation that doesn't work for us

primiano (commented):
Unfortunately, as much as I'm sympathetic with what you are trying to do, this change as proposed is unfeasible because of version skews of libunwindstack. see comments below.

primiano:
+ @rsavitski FYI (no action needed)

gn/BUILD.gn
Sentimentron: So I've noticed that libunwindstack in Chrome is also broken (it does not successfully strip the pointer-authentication bits). Is there an underlying reason why libunwindstack is used for Perfetto instead of the canonical LLVM libunwind?

src/profiling/memory/wire_protocol.h
Sentimentron: As far as I can tell, Chrome's libunwindstack is mostly only used for Perfetto so I think it's fine to align the two versions.

primiano: > As far as I can tell, Chrome's libunwindstack is mostly only used for Perfetto 

Hold on a sec, I don't this this is true.
As far as I can see, that is used by chrome's base/profiler, which is not perfetto. I don't see any build rule that has a dependency from third_party/perfetto to third_party/libunwindstack in chromium, mainly because it would require changes like the ones you are making here to happen.

Also libunwindstack today requires `enable_perfetto_heapprofd | enable_perfetto_traced_perf` and both of them today seem false in chromium.

I think what might have happened by looking at git log is that somebody (presumably from cronet?) copied/took inspiration from some perfetto code and added some base/profiler code that uses libunwindstack in a similar way perfetto does. but that code is NOT perfetto

Which means that you would have to make sure that chrome's base/profiler and third_party/perfetto are aligned on the same version of libunwindstack... plus the below

> I think it's fine to align the two versions.

the problem is that libunwindstack is a pure android project, and as such is legitimately allowed to depend on any random part of the android tree. 
It's alredy the case today that, in order to make it work, we have to pull random files from various android_base / android_utils libraries (thankfully headers only mostly)

The risk for you (and recursively for us) is that one day they will add some dependency to a bigger android library, and doing so will require you to have to pull larger parts of the android tree into chromium.

I really don't think it's a viable solution, it creates too much risk of getting stuck (in a state where the chronium roll is blocked with no good solution to unblock it) 

Again, I think the only viable solution here is to create a swappable indirection layer in our profilers, so you can re-implement it using chrome's version of libwundstack (or anything else of your choice).
But that would avoid the deadlock situation i mendioned, because you can decide to stay on an older version and things will be fine.

gn/BUILD.gn
LalitMaganti: libunwindstsck is the "official" unwinding library on Android that's supported and that's able to deal with all the nuances of Android unwinding especially with respect to Java code in its various states (interpreted, Jitted, ahead of time compiled(

primiano: the main feature is the ability to decompress the mnified dwarf debugging sections on the flight, which is something specific to android's toolchain.
If I remember correctly now, whoever introduced libunwindstack on chrome it might have done it because they wanted to do similar tricks.
You need to check git log, maybe it has been done only for chromeos back then. for sure libunwindstack cannot work on windows or mac.

PR #5214, opened 2026-03-20
tp: Track process lifecycle from Android framework events

zezeozue:
In some traces, ftrace only captures an initial process snapshot. Subsequent starts and deaths come from the Android framework via AndroidProcessStartEvent (process_bound) and AndroidBinderDiedEvent (binder_died) track events carrying pid, uid, and process_name.

Hook into ProtoToArgsParser::AddParsingOverrideForType to intercept these events during arg parsing. For starts, reuse the existing process if ftrace already created it (filling in uid only), otherwise create one via StartNewProcessWithoutMainThread. For deaths, set uid/name on the process row before marking it ended.

Add ProcessTracker::EndProcess() to handle ending processes created both with and without a main thread. Add ProcessTracker::UpidForPid() as a public lookup (previously test-only).

### DIFF
--- a/src/trace_processor/importers/common/process_tracker.cc
+++ b/src/trace_processor/importers/common/process_tracker.cc
@@ -133,6 +133,24 @@ void ProcessTracker::EndThread(int64_t timestamp, int64_t tid) {
   pids_.Erase(tid);
 }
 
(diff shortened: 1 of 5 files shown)


zezeozue:
Will re-upload a new patch to address offline comments

PR #5223, opened 2026-03-21
ui: Add full path in pinned tracks

stevegolton:
When you pin a track, it doesn't show any context about where it came from. For process-scoped tracks such as thread state or counter tracks, their process context is lost.

The fix is to make pinned tracks show the full path, but first we need to change how pinned tracks work. Currently, pinned tracks are a separate mini workspace, and each TrackNode in it is a standalone TrackNode. When we pin, we take a complete copy of the node and add that copy to the pinned tracks workspace.

This patch changes how this works, so that pinned tracks are actually references to TrackNodes in the workspace.

This does also involve changing debug tracks and some plugins which added tracks to the pinned area directly, as we now must add a track or group to the workspace first before it can be pinned, but the result is the same.

Before:
image

After:
image

### DIFF
--- a/ui/src/components/tracks/debug_tracks.ts
+++ b/ui/src/components/tracks/debug_tracks.ts
@@ -252,7 +252,8 @@ async function addPivotedSliceTracks(
     });
 
     const trackNode = new TrackNode({uri, name, removable: true});
-    trace.currentWorkspace.pinnedTracksNode.addChildLast(trackNode);
+    trace.currentWorkspace.addChildLast(trackNode);
+    trackNode.pin();
   }
 }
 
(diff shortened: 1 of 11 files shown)


LalitMaganti (approved):

bsarden:
Hi @stevegolton! 

I found myself reaching for just this today :P thanks for all the hard work on the UI I really appreciate it!
Lmk if there's anything you need help with here to get this across the finish line. Happy to get my feet wet.

stevegolton:
Hey @bsarden, sorry for the delay. I parked this PR due to architectural issues I can't quite remember right now. Let me remind myself what those issues were and we can hopefully land this.

PR #5282, opened 2026-03-25
ui: support auto-height DrawerPanel

abhmen01:
Add a `notFixedHeight` option to `DrawerPanel` so callers can render drawers that size to their content instead of forcing a fixed visible height.

When this mode is enabled, the panel skips the inline height style in the visible state while preserving the existing collapsed and fullscreen behavior. Drag-resizing is also disabled for drawers that do not use a fixed height.

### DIFF
--- a/ui/src/widgets/drawer_panel.ts
+++ b/ui/src/widgets/drawer_panel.ts
@@ -107,6 +107,9 @@ export interface DrawerPanelAttrs {
   // What height should the drawer be initially?
   readonly startingHeight?: number;
 
+  // Whether the drawer height should not be fixed.
+  readonly notFixedHeight?: boolean;
+
   // Called when the drawer visibility is changed.
   onVisibilityChange?(visibility: DrawerPanelVisibility): void;
 }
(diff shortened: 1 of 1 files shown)


abhmen01:
Hi @stevegolton 
Can you please review this?  It would be great to get your feedback soon.

stevegolton (commented):
I'm trying to figure out your use case behind this PR so bear with me.

The use case as I understand it is that you want a tabbed panel in the new PMU recording page, is that right?

In that case I think the best option will be to use the `Tabs` widget, which I think is a much better fit.

abhmen01:
@stevegolton 
You’re right. For PMU I only need a tabbed content container, not the drawer semantics. When I originally wrote it sometime back, I was using SplitPanel. If I remember correct, Tabs were not available then. Later when rebased, the SplitPanel got replaced by DrawerPanel. 
I'm closing this pull request, as I could use Tabs for the PMU recorder page.

stevegolton:
Apologies for the churn!

Issue #3288, opened 2025-10-14
Linux journald logs datasource


sabae-valve:
Perfetto's android logcat capture source is of great help in tying together high-level system-wide events with low-level performance details.  When working on systemd Linux systems, the journal log stream provides a similar mechanism as the logcat.  It would be great for Linux system performance measurement to have a journald datasource that can be surfaced in the same way as the Android logcat logs are.

LalitMaganti:
Hey. Thanks for filing this! In general, this sounds good I think we would be supportive of adding a new datasource for fetching logs from journald.

However, from the core team perspective, this would be a feature that we would not necessarily use ourselves and nor would any of our internal clients so it would be a lower priority for us. If someone externally wanted to implement it, I'd be happy to write up something explaining how to do this.

If we were do this, this would probably only happen if someone in the team becomes motivated out of personal interest rather than something we would do more formally due to our work.

PR #5331, opened 2026-03-29
Add `linux.journald` data source

staticfloat:
This PR adds a `linux.journald` data source to capture journal messages on systemd systems, inspired by the Android logcat viewer.  This PR satisfies https://github.com/google/perfetto/issues/3288.

The design is fairly straightforward; `libsystemd` is added as an optional dependency, `traced_probes` is taught how to open a connection to the system journal, and messages are serialized in a new protobuf message and displayed in a new UI plugin that mimics that of the logcat messages.

This PR was co-written by AI; while I iterated on it until the look and feel felt correct, there may be context or use cases that I am unaware of.  I am open to all review comments, and will do my best to address them.


LalitMaganti:
Hi. Thanks for the contribution! Really great to see this happening. However there are several concerns I have with this patch as it stands:
1) this is a very large patch for us to review and increases the risk of reverts if one piece is wrong. It would be better if this split into at least 3 patches: on device, trace processor and UI. Even more is preferable if it's cleanly possible. 
2) On device, it seems like in the probes code, we're relying on a header installed on the system and linking against systemd directly.
    * This is something we don't do in Perfetto because it means you then cannot compile unless systemd is installed which it very well might not be, especially in docker containers or VMs where we compile code for distribution 
    * The use of systemd functions directly instead of going through dlopen is also problematic because it means, at runtime we will need the library installed or the dynamic linker will fail to start traced_probes. That code needs to be rewritten to treat systemd purely as an optional dependency at runtime not compile time. 
3) I see you've added a new systemd logs table. I'm torn on this but on balance I think we should likely do the opposite refactor: create a common "logs" table and make Android logcat and systemd logs just views over the top of it. This should reduce unnecessary duplication while still allowing custom entries through the use of args. 
4) this also applies to the UI. You seem to have duplicated a *lot* of code at the UI level when really we should have figured out how to refactor and share that code instead. Specifically, we should probably have a common "logs" plugin which hosts all the shared code with the Android logs and systemd plugins depending on it. 

This is my high level comments: I'm really excited to see this happen but I want to make sure this goes in while fitting in with the overall direction of the project!

staticfloat:
Thank you for your comments, they are quite helpful.  As you can see, I'm working on this in my (not-so-copious) spare time, I apologize for the long iteration time.

> It would be better if this split into at least 3 patches: on device, trace processor and UI. Even more is preferable if it's cleanly possible.

I have separated this into three commits; would you prefer that I submit each commit as its own pull request?  If so, seems like a good opportunity for that new stacked PRs feature GitHub touted a while back.

> That code needs to be rewritten to treat systemd purely as an optional dependency at runtime not compile time.

Good idea, I have changed this to a runtime-loaded dependency.

> create a common "logs" table and make Android logcat and systemd logs just views over the top of it. This should reduce unnecessary duplication while still allowing custom entries through the use of args.

I believe I've now done this, please let me know if what I've done looks wrong.

> This also applies to the UI. You seem to have duplicated a lot of code at the UI level when really we should have figured out how to refactor and share that code instead. Specifically, we should probably have a common "logs" plugin which hosts all the shared code with the Android logs and systemd plugins depending on it.

I have done my best to follow this guidance, however this is the area that I am least confident in, so you'll need to tell me if the organization of the code makes sense.

LalitMaganti:
> I have separated this into three commits; would you prefer that I submit each commit as its own pull request? 

Yes please, putting them on own PRs will be best. I will look at this immediately after you do so!

PR #5354, opened 2026-03-31
Update the jank track pin plugin for the JPS metrics.

pmuetschard:
Bug: b/430220320

### DIFF
--- a/ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinCujScoped.ts
+++ b/ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinCujScoped.ts
@@ -37,18 +37,17 @@ class PinCujScopedJank implements MetricHandler {
    */
   public match(metricKey: string): CujScopedMetricData | undefined {
     const matcher =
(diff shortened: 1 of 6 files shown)


ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinCujScoped.ts
```diff
@@ -83,6 +82,10 @@ class PinCujScopedJank implements MetricHandler {
       jankTypeFilter = ' AND sf_missed > 0';
       jankTypeDisplayName = 'sf';
     }
+    if (metricData.isWeighted) {
+      jankTypeFilter += ' AND jank_score > 0';
```
nicomazz, line 86: I think it would be nice to also add the jank score directly in the pinned slices (but feel free to do that in a follow up in case)

nicomazz (approved):

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinCujScoped.ts
pmuetschard: Done

PR #5392, opened 2026-04-02
ui: Improve UX of com.android.AndroidLockContention

ivan-chong:
This CL improves the UX of navigation by creating custom tracks above each blocking thread to mirror the lock contention slices of blocked threads.

These custom "Blocking Contention" tracks allow the user to navigate from a lock contention slice to its blocking thread, whilst visualising the full context of all blocked slices on that lock.

This also allows a target for arrows to be drawn between blocked and blocking slices across threads, with the "Blocking Contention" track providing detailed breakdowns of all contentions in the details panel, including blocking methods, source files and line numbers for monitor contentions.

The navigation is also replaced with a flow-like UX which uses '[' and ']' to navigate back and forth respectively.

Bug: 289343169


ui/src/plugins/com.android.AndroidLockContention/lock_contention_sql.ts
```diff
@@ -0,0 +1,101 @@
...
+CREATE OR REPLACE PERFETTO VIEW android_lock_contention AS
+WITH raw_contentions AS (
+  SELECT
+    s.id,
+    s.ts,
+    s.dur,
+    s.name,
+    s.utid,
+    cast_int!(STR_SPLIT(STR_SPLIT(s.name, '(owner tid: ', 1), ')', 0)) AS owner_tid
+  FROM thread_slice AS s
```
stevegolton, line 31: I think you need to add an include of the `slices.with_context` module at the top of this query.

stevegolton (changes requested):

stevegolton (approved):

PR #5402, opened 2026-04-03
tp/ui: Wattson GPU attribution in package level

runpinglai-creator:
## Summary
This PR implements Wattson GPU attribution at the package level and refactors the UI to support the separated CPU and GPU attribution tables. 

This work is split into 2 commits:

### 1. tp: implement Wattson GPU attribution in package level
- **Table Isolation**: Separated GPU from CPU tables because GPU only tracks package-level granularity (coarser than CPU), and having separate tables avoids wasting UI loading time for users focused on only one domain.
- **Attribution Logic**: Active power (tasks > 0) is shared among concurrent UIDs. Idle transition power (tasks = 0) in active pulses is attributed to the causing UID based on proximity.
### 2. ui: refactor Wattson package aggregator for split CPU/GPU tables
- Refactored `WattsonPackageSelectionAggregator` into dedicated CPU and GPU aggregators sharing a common base class. The GPU aggregator now directly queries the standalone GPU table.

Example of GPU Attribution:
image

Bug: 499375826
test: UI tests, Trace Processor diff tests


src/trace_processor/perfetto_sql/stdlib/wattson/tasks/attribution.sql
```diff
@@ -13,15 +13,66 @@
...
+SELECT
+  ii.ts,
+  ii.dur,
+  ta.uid,
+  ta.gpu_id,
+  -- Calculate attributed power (mW) proportionally shared
+  iif(ta.active_tasks > 0, p.gpu_mw / ta.active_tasks, 0.0) AS estimated_mw
+FROM _interval_intersect!(
+  (
+    _ii_subquery!(tasks_with_total),
```
wusamuel6: You can just directly use `tasks_with_total` (without `_ii_subquery`) if line 39 column is stored as `id` instead of `_auto_id`.

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/attribution.sql
```diff
@@ -13,15 +13,66 @@
...
+      ON tot._auto_id = id_1
+  )
+-- Then align with Wattson's power estimates
+SELECT
+  ii.ts,
+  ii.dur,
+  ta.uid,
+  ta.gpu_id,
+  -- Calculate attributed power (mW) proportionally shared
+  iif(ta.active_tasks > 0, p.gpu_mw / ta.active_tasks, 0.0) AS estimated_mw
```
wusamuel6, line 64: Can we make this based on the percentage shown in the UID window instead of distributing it evenly?

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/pulses.sql
```diff
@@ -0,0 +1,156 @@
+--
+-- Copyright 2025 The Android Open Source Project
```
wusamuel6: 2026

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/pulses.sql
```diff
@@ -0,0 +1,156 @@
...
+INCLUDE PERFETTO MODULE wattson.gpu.estimates;
+
+INCLUDE PERFETTO MODULE wattson.gpu.freq_idle;
+
+INCLUDE PERFETTO MODULE wattson.tasks.gpu_tasks;
+
+INCLUDE PERFETTO MODULE wattson.utils;
+
+-- Step 1: Find active GPU islands (contiguous freq > 0)
+CREATE PERFETTO TABLE _gpu_pulses AS
```
wusamuel6: What does pulses mean? Is this the GPU active region? I think more precise name would be better

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/gpu_tasks.sql
```diff
@@ -0,0 +1,64 @@
...
+  t.uid,
+  t.gpu_id
+FROM slice AS s
+JOIN android_gpu_work_period_track AS t
+  ON s.track_id = t.id
+WHERE
+  s.dur > 0;
+
+-- Calculate the number of active GPU tasks at any point in time.
+CREATE PERFETTO TABLE _gpu_active_task_count AS
```
wusamuel6, line 32: We should add test case for tables with these complex calculations

wusamuel6: And also other newly added GPU attribution tables

wusamuel6 (changes requested):

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/attribution.sql
runpinglai-creator: done

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/attribution.sql
runpinglai-creator: I see your point! To clarify, the percentage inside the gpu_work_period is an independent duty cycle for each individual UID (i.e., package A utilizes x% of its own period, package B utilizes y% of its own period). They aren't necessarily partitions of a single shared 100% pie. Also all percentages I've seen so far are always 100% so it doesn't useful in here

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/pulses.sql
runpinglai-creator: done

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/pulses.sql
runpinglai-creator: Yes it's GPU active regions. I've renamed them. Naming them pulses because each active freq region is like independent pulses. Sorry didn't make it clear

src/trace_processor/perfetto_sql/stdlib/wattson/tasks/gpu_tasks.sql
runpinglai-creator: done

wusamuel6 (approved):

LalitMaganti (approved):

stevegolton (approved):

PR #5409, opened 2026-04-06
sdk: Add track_event_buf_exhausted_policy to TracingInitArgs

bsarden:
Add a track_event_buffer_exhausted_policy field to TracingInitArgs,
allowing apps to set the buffer exhausted policy (drop vs stall)
for TrackEvent at initialization time.

The value is stashed in TrackEventInternal static state during
Tracing::Initialize() and read during TrackEvent::Register(),
following the same pattern as use_monotonic_clock and
disallow_merging_with_system_tracks. A protected Register(dsd,
params) overload on DataSource allows TrackEventDataSource to
inject custom DataSourceParams without intermediate storage.

TrackEvent is used in critical system services (e.g.,
system_server) where an accidental kStall policy could cause a
device soft-reboot. Plumbing through TracingInitArgs removes
ordering ambiguity (Initialize always precedes Register) and
avoids redundant state in DataSourceType. See https://github.com/google/perfetto/pull/1312 and
b/384007571.

I don't have a use-case for also allowing track_event_buf_exhaust_policy to be configurable.
For now, making a hard-choice when we initialize the process is ok.
If we need configurabilty, the extension points are natural; we can follow footsteps of this PR.

cc: @dreveman

### DIFF
--- a/include/perfetto/tracing/internal/track_event_data_source.h
+++ b/include/perfetto/tracing/internal/track_event_data_source.h
@@ -246,6 +246,10 @@ class PERFETTO_EXPORT_COMPONENT TrackEventDataSource
  public:
   static constexpr bool kRequiresCallbacksUnderLock = false;
 
+  static BufferExhaustedPolicy GetDefaultBufferExhaustedPolicy() {
+    return TrackEventInternal::GetBufferExhaustedPolicy();
+  }
+
   // DataSource implementation.
   void OnSetup(const DataSourceBase::SetupArgs& args) override {
     auto config_raw = args.config->track_event_config_raw();
@@ -362,6 +366,8 @@ class PERFETTO_EXPORT_COMPONENT TrackEventDataSource
 
   static void ResetForTesting() {
     TrackEventInternal::GetInstance().ResetRegistriesForTesting();
+    TrackEventInternal::SetBufferExhaustedPolicy(
+        GetDefaultBufferExhaustedPolicy());
   }
 
  private:
@@ -373,6 +379,7 @@ template <const TrackEventCategoryRegistry* Registry>
 class TrackEvent {
  public:
   using TraceContext = TrackEventDataSource::TraceContext;
+
   static bool Register() { return TrackEventDataSource::AddRegistry(Registry); }
 
   // Add or remove a session observer for this track event data source. The
--- a/include/perfetto/tracing/data_source.h
+++ b/include/perfetto/tracing/data_source.h
@@ -280,6 +280,12 @@ class DataSource : public DataSourceBase {
   constexpr static BufferExhaustedPolicy kBufferExhaustedPolicy =
       BufferExhaustedPolicy::kDrop;
 
+  // Returns the effective buffer exhausted policy. Derived classes can
+  // override this to provide a runtime value (e.g. from TracingInitArgs).
+  static BufferExhaustedPolicy GetDefaultBufferExhaustedPolicy() {
+    return DerivedDataSource::kBufferExhaustedPolicy;
+  }
+
   // Whether the kBufferExhaustedPolicy policy above is overridable via config.
   constexpr static bool kBufferExhaustedPolicyConfigurable = false;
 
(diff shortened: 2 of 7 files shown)


LalitMaganti:
I think the reason for this behaviour is pure oversight. Configurable buffer exhausted modes were only introduced signifcantly after track event and I think we never thought about retroactively applying that to TrackEvent.

@primiano @sashwinbalaji would be the right people to confirm the above.

primiano:
So I did some digging in #1312 and the internal bug b/384007571

I think this feature was introduced to allow SDK users that implement data sources to make it configurable by calling PerfettoDsSetBufferExhaustedPolicyConfigurable(true)

I think back then nobody thought about track event, it was focused only on the DS layer.

Now there is a legitimate question that is: should TrackEvent itself be configurable?

I think depends on whether you want:

A) TrackEvent in an app you control to be configurable: I'd be absolutely positive about this (But shoudl be done in a different way than this PR)

B) TrackEvent to be always configurable. I'd be a bit more careful here. My problem here is the fact that trak event is used in key system services (e.g system_server). Making this universally configurable (incl system_server) means making it so that if somebody copy/pastes the wrong flag into the config without knowing what they are doing, they will cause a device soft-reboot, which is unideal


So if you want to send out a PR for A, i'm more then happy to review, but needs to be something that a client of TrackEvent calls explicitly when initializing.

bsarden:
@primiano took a stab at A). Idea is that client calls `SetBufferExhaustedPolicy` before calling `Register` if they want to change the buffer exhausted policy (opt-in).
There's a matching `SetBufferExhaustedPolicyConfigurable(bool)` helper as well for completeness.

AI Disclaimer: I used claude code to help w/ planning+impl for this PR. It was quite helpful as it was my first time digging through TrackEventDataSource+friends. Looking forward to first round of feedback :)

primiano:
Hmm so I have a bit of mixed feelings here, for a few reasons:

1. This adds a SetBufferExhaustedPolicyConfigurable to TrackEvent, but in reality this method has effect only if you call it before the very first Register, otherwise it's too late
2. There is a bit of redundant state in track_event.h now, because if you look at the Register method, it already takes a internal::DataSourceParams, which then contains the buffer stall policy there in `bool buffer_exhausted_policy_configurable`

I think a better way would be plumbing this via the TracingInitArgs passed to Tracing::Initialize. That removes the ambiguity of when you can set it.

I know that in theory TrackEvent is "a data source like all the others". But in practice it's special.
What you have to figure out is how to make it so that in TrackEventInternal::Initialize, when we fill the DataSourceDescriptor, we can poke at the TracingInitArgs.
Maybe TracingINitialize can stash them in some TrackEvent static state somewhere (just remember to clear it in ResetForTesting)?

bsarden:
@primiano Took another stab at this based on your suggestion to plumb `track_event_buf_exhausted_policy` through TracingInitArgs. I dropped the "configurable" part of this MR since if we are doing this at init-time it's really only "settable". 

Will we be setting the track_event_buf_exhausted_policy for all processes in the system when we run in system mode? Or since we are doing this with `TracingInitArgs` now do we lose the per-app opt-in behavior.

Thanks for being patient with me!

include/perfetto/tracing/internal/track_event_data_source.h
```diff
@@ -362,6 +366,7 @@ class PERFETTO_EXPORT_COMPONENT TrackEventDataSource
 
   static void ResetForTesting() {
     TrackEventInternal::GetInstance().ResetRegistriesForTesting();
+    TrackEventInternal::SetBufferExhaustedPolicy(BufferExhaustedPolicy::kDrop);
```
primiano: shouldn't this be something like TrackEventInternal::GetDefaultBufferExhaustedPolicy()?

primiano (approved):
This one works for me (with one small question, but not a big deal).
Just note that now the PR is out of sync with your initial PR description (Which probably needs rewording).

Your initial intent was to make the policy runtime-configurable (on a per-session basis). 
Now you are making a hard choice while initializing the process.

TBH I am fine with the latter, just make sure you update your PR description if we go with this.

primiano (approved):

include/perfetto/tracing/internal/track_event_data_source.h
bsarden: Nice catch, updated so `GetDefaultBufferExhaustedPolicy` uses the `DerivedDataSource::BufferExhaustedPolicy`, now `ResetForTesting` can safely call:

```cpp
TrackEventInternal::SetBufferExhaustedPolicy(GetDefaultBufferExhaustedPolicy());
```

bsarden:
@primiano We need another Google org member to approve. Who would be best for that? 

@sashwinbalaji maybe?

sashwinbalaji (approved):
outdated

PR #5435, opened 2026-04-10
Fix handling of spaces in query regex

emrekultursay:
The '\\s' doesn't seem to behave as expected in regex, causing filters. For instance when the slice name is as follows:

```
location=/data/app/.../oat/arm64/split_MeasurementDynamite_installtime.vdex status=up-to-date filter=verify reason=vdex
```

...then we get `filter=verify rea` because the regex reads until it sees the letter 's', not whitespace.

Before: 
before

After:
after

### DIFF
--- a/ui/src/plugins/com.android.AndroidStartup/optimizations.ts
+++ b/ui/src/plugins/com.android.AndroidStartup/optimizations.ts
@@ -84,7 +84,7 @@ export async function optimizationsTrack(
       AS
       SELECT
         MAX(slice_ts) AS compile_ts,
-        regexp_extract(slice_name, 'filter=([^\\s]+)') as filter
+        regexp_extract(slice_name, 'filter=([^ ]+)') as filter
       FROM android_thread_slices_for_all_startups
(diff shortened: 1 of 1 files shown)


LalitMaganti:
a) feels like we should fix the underlying issue here rather than working around.
b) this might have something to do with the changes to start using RE2. @sashwinbalaji should look into that.

emrekultursay:
SGTM. I'll let you handle the root cause then.

sashwinbalaji:
I think this is an POSIX regex limitation which should be fixed once we move to re2/pcre/std::regex.

I verified it locally and our ToT shows `verify rea` but once we re-revert go/pgh/5417 we get `verify`

Posix regex doesn't support \s. as \s is a Perl/PCRE extension. So that's why it's parsed as literal characters \ and s causing match till rea

sashwinbalaji:
the regex library update has landed [go/pgh/5460](http://goto.google.com/pgh/5460). Please try now and see if the issue still happens. Thanks

PR #5436, opened 2026-04-11
ui: Add GpuCompute plugin for GPU kernel performance analysis

dreveman:
Adds the dev.perfetto.GpuCompute plugin for analyzing GPU compute kernel performance across multiple vendors and hardware architectures in the Perfetto trace viewer.

When a compute GPU slice (render stage event) is selected, the Compute tab automatically opens and loads detailed metrics for the selected kernel.

Core plugin (dev.perfetto.GpuCompute):
- Compute tab that auto-opens on compute slice selection
- Details tab with per-kernel metric tables, collapsible sections, baseline comparison, and percent bars
- Summary table with sortable columns and double-click navigation
- Toolbar with kernel selector, terminology picker, humanize toggle, and baseline comparison controls
- Declarative Section registry for metric group definitions
- Terminology registry for adapting GPU concepts (thread, warp, block, SM, etc.) to different vendors
- Well-known metric registry for vendor-agnostic toolbar and summary display (duration, cycles, frequency, throughput)
- Analysis provider interface for optional analysis backends

Built-in vendor support:
- CUDA (NVIDIA) — metrics, terminology, well-known metrics
- AMD — metrics, well-known metrics
- OpenCL — terminology

Additional vendors can register support through companion plugins that provide terminologies, metric sections, well-known metrics, and analysis providers.

Built-in sections (section/):
- Speed of Light — compute and memory throughput overview
- Launch Statistics — kernel launch configuration
- Occupancy — warp occupancy and limiting factors
- Workload Analysis — per-pipeline utilization


ui/src/plugins/dev.perfetto.GpuCompute/index.ts
```diff
@@ -0,0 +1,389 @@
...
+            console.warn('GpuCompute: failed to fetch slice metrics:', e);
+            content.setSliceId(sliceId, false, {
+              hasMetrics: false,
+              toolbar: undefined,
+            });
+          });
+      } else {
+        content.setSliceId(undefined);
+      }
+    }, 100);
```
dreveman: @stevegolton Is there a better way to do this? what I'm doing here is obviously a hack but it achieves my goal of automatically opening and navigating to a "Compute" tab when a GPU slice with compute details is selected. How do you feel about adding a `trace.selection.onChange` if there's not a way to do this already?

dreveman:
@mohammadawwad FYI

dreveman:
Here are some screenshots of example GPU traces showing what it looks like:

Summary view:
summary

Detailed view:
detailed

Detailed kernel compare view:
compare

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
stevegolton: The whole architecture is purposely lazy. We don’t want to kick off heavy data loading for a tab the user isn't even looking at, so plugins are only really woken up when mithril redraws.

Whenever the selection changes, it triggers a redraw. If your tab is open, this ends up calling your tab's `render()` function. At that point, you could manually compare the current selection state against the previous state to see if you need to reload, much like you're doing here, but that can get messy as you have to handle the change detection manually as well as avoid queueing up a mountain of queries in the case where someone spams the selection faster than your query completes.

I’d strongly recommend using `QuerySlot`. It's very similar to how [TanStack Query](https://tanstack.com/query/latest) works if you've ever used that.

The idea is that you poll it every single render cycle. You give it a key and an async function; if the key matches what it has in memory, it returns the data synchronously. If the key has changed, it handles the fetch in the background and returns a pending state so you can show a loading spinner. When the query completes, a redraw is automatically triggered, so your render code will run again, poll the query slot which will return the cached query data so you can render the tab content.

Here’s a quick example of how that looks in practice:

```ts
// One queue and one slot per async resource on the component.
// The queue serialises all queries; multiple slots can share one queue.
const queue = new SerialTaskQueue();
const rowsSlot = new QuerySlot<string[]>(queue);

// Call use() every render cycle — it is synchronous and safe to call repeatedly.
// Changing the key triggers a new fetch; the same key returns the cached result.
function render(search: string) {
  const result = rowsSlot.use({
    key: {search},
    queryFn: async () => {
      const rows = await engine.query(
        `SELECT name FROM t WHERE name LIKE '%${search}%'`,
      );
      return rows.iter({name: STR}).map((r) => r.name);
    },
  });

  if (result.data === undefined) {
    return 'Loading...';
  }

  return result.data.join(', ');
}
```

I think this is the way to go for this plugin.

dreveman: Thanks @stevegolton! I switched to using QuerySlot to automatically show the compute tab for compute slices and the plugin is now automatically enabling the compute tab for traces that contain at least one gpu slice with compute category.

I know this is a large change and I don't expect you to review it in detail but I would appreciate if you could look over it when you have time to make sure I'm not doing something that is an anti-pattern or something that will make it hard to maintain.

LalitMaganti:
Given how complex the details panel looks, would it make more sense if we create a new page for all of this instead of tryign to stuff everything in the detgails panel. This is the appraoch we took for the heap dump explorer and I think people find it works really well to have a summary in the details panel and then a separate page for the more detailed internals.

ui/src/plugins/dev.perfetto.GpuCompute/section/launch_statistics.ts
```diff
@@ -0,0 +1,109 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {registerSection} from './index';
+
+export function registerLaunchStatisticsSection(): void {
+  registerSection({
+    id: 'dev.perfetto.GpuCompute.Section.LaunchStatistics',
```
stevegolton: I wouldn't go too nuts on the namespacing here as you're only registering things with your own plugin, but that's up to you.

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    // Rendering view
+    return m('div', {key: wrapKey}, [toolbar, body]);
+  }
+}
+
+// Plugin entry point, registers the tab and menu command
+export default class GpuComputePlugin implements PerfettoPlugin {
+  static readonly id = 'dev.perfetto.GpuCompute';
```
stevegolton: Can we please call this `com.meta.GpuCompute` to give the indication that this plugin is not owned by our team?

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+        if (list.length > 0) {
+          trace.tabs.showTab(tabUri);
+        }
+      })
+      .catch((e) => {
+        console.warn('GpuCompute: failed to fetch kernel launch list:', e);
+        content.setOptions([]);
+      });
+
+    trace.trash.use({
```
stevegolton: I don't think this disposal is actually necessary.Closing the trace will automatically cancel any pending work anyway, and since all references to the queryslot object are lost it should be GCd.

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+    const content = new Compute(trace.engine, trace, tabUri);
+    trace.tabs.registerTab({
+      isEphemeral: false,
+      uri: tabUri,
+      content: content,
+    });
+
+    // Load dropdown options once. If compute slices exist, auto-open
+    // the tab so it can react to selection changes via its QuerySlot.
+    await fetchKernelLaunchList(trace.engine)
```
stevegolton: Combined `await` and `then` is unusal. You'd usually use one or the other... like:

```ts
try {
  const list = await fetchKernelLauchList(trace.engine);
  ...
} catch (e) {
  console.warn(...);
}
```

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+import {
+  PerformanceAnalysisResult,
+  AnalysisCache,
+  getAnalysisProvider,
+} from './analysis';
+import {SerialTaskQueue, QuerySlot} from '../../base/query_slot';
+
+class Compute {
+  constructor(
+    private engine: Engine,
```
stevegolton: Nit: could be readonly

ui/src/plugins/com.meta.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+          const hasMetrics = Array.isArray(data) && data.length > 0;
+          return {
+            hasMetrics,
+            toolbar: hasMetrics ? data[0].toolbar : undefined,
+          };
+        },
+      });
+
+      if (
+        result.isFresh &&
```
stevegolton, line 270: Can remove the isFresh check, it's always true - only applies when you use queryslot with retainOn keys.

ui/src/plugins/com.meta.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+    ) => {
+      this.sectionAnalysisCache.set(
+        this.getSectionCacheKey(sliceId, sectionName),
+        result,
+      );
+    },
+  };
+
+  public getTitle() {
+    return 'Compute';
```
stevegolton, line 124: 'Compute' is slightly mysterious, could you make it a little more descriptive? 'GPU Compute' perhaps?

ui/src/plugins/com.meta.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  // Updates which kernel is shown from the dropdown. Fetches toolbar
+  // metrics for the selected kernel and switches to the appropriate tab.
+  public async setSliceId(sliceId: number, suppressAutoDetails = false) {
+    const firstId = this.options[0]?.id ?? -1;
+    this.sliceId = sliceId !== -1 ? sliceId : firstId;
+    m.redraw();
```
stevegolton, line 132: I think the vast majority of these m.redraws() are unnecessary.  Mithril will automatically schedule a redraw when:
- A trace processor query completes
- A DOM event registered through mithril is fired

Try removing them, I suspect everything will work just fine.

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+    // --- Handle selection-driven updates via QuerySlot ---
+    const sel = this.trace.selection.selection;
+    const selSliceId =
+      sel.kind === 'track_event'
+        ? (sel as TrackEventSelection).eventId
+        : undefined;
+
+    const selectionChanged = selSliceId !== this.lastSelectionSliceId;
+    this.lastSelectionSliceId = selSliceId;
+    if (selectionChanged) {
```
stevegolton: Do we still need all this manual change detection if we're using the queryslot?

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+    } else if (selectionChanged) {
+      // Selection cleared — fall back to first option with summary
+      const firstId = this.options[0]?.id ?? -1;
+      this.sliceId = firstId;
+      this.setToolbarInfo(undefined);
+      setInfoTab('summary');
+    }
+    // --- End selection handling ---
+
+    // The key forces a remount when the inputs change so the data refreshes
```
stevegolton: We shoudn't need to do this. What happens if you remove it?

ui/src/plugins/com.meta.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+import {registerCUDATerminology} from './terminology/cuda';
+import {registerOpenCLTerminology} from './terminology/opencl';
+import {
+  PerformanceAnalysisResult,
+  AnalysisCache,
+  getAnalysisProvider,
+} from './analysis';
+import {SerialTaskQueue, QuerySlot} from '../../base/query_slot';
+
+class Compute {
```
stevegolton, line 52: The state management situation of this class makes me uneasy, but it's difficult to get a handle on exactly what it needs to do without trying it out. Do you have any representative traces you could share publicly - something we could upload to our test traces?

ui/src/plugins/dev.perfetto.GpuCompute/toolbar.ts
```diff
@@ -0,0 +1,334 @@
...
+      ),
+    ]),
+
+    // Toolbar divider
+    m('hr.pf-gpu-compute__toolbar-divider'),
+  ];
+}
+
+// Whether metric values are displayed in human-readable units.
+export let enableHumanizeMetrics = true;
```
stevegolton: There's a lot of global mutable state in this plugin that's making me uneasy (not just here). Can you please pass round a settings object rather than exposing all these settings as mutable globals.

In fact a context object that you pass around could eliminate a lot of the globals quite neatly.

ui/src/plugins/com.meta.GpuCompute/index.ts
```diff
@@ -0,0 +1,392 @@
...
+    return m('div', {key: wrapKey}, [toolbar, body]);
+  }
+}
+
+// Plugin entry point, registers the tab and menu command
+export default class GpuComputePlugin implements PerfettoPlugin {
+  static readonly id = 'dev.perfetto.GpuCompute';
+
+  static onActivate(): void {
+    registerCUDATerminology();
```
stevegolton, line 349: The registries that these functions contribute to: could these just be defined as readonly literal maps instead?

This would entirely avoid the registration ceremony and eliminate any global mutation or ordering dependencies.

stevegolton (changes requested):
Thanks for this. I haven't gone through this PR with a fine toothed comb but from what I've seen it does a hell of a lot, but I'd like to see the state management tidied up a bit before landing.

if you could share some representative traces for testing that would be also be great, if you have any that you can share publicly.

Also re @LalitMaganti 's comment on page vs tab, I'll leave that one up to you - make it a page if you want the real estate, make it a tab if you want to see the timeline simultaneously, or put a summary in a tab, and the full thing in a page.

mohammadawwad:
> Given how complex the details panel looks, would it make more sense if we create a new page for all of this instead of tryign to stuff everything in the detgails panel. This is the appraoch we took for the heap dump explorer and I think people find it works really well to have a summary in the details panel and then a separate page for the more detailed internals.

I personally lean towards the tab-based approach here.

One of the biggest advantages is that it lets us keep the timeline visible while inspecting compute metrics, which is really valuable when you’re trying to correlate behaviour across kernels. With a separate page, it may be harder for a user to fully understand the context.

Another key benefit is how it integrates with selection. Right now, when a user selects a GPU slice that contains metrics, we automatically open the Compute tab and jump straight into the relevant details tab instead of the usual Current Selection tab. That makes the workflow much faster and more discoverable for users compared to having to manually navigate somewhere else and select the kernel via the drop-down. This interaction model feels more natural in a tab than in a separate page, since it’s directly tied to timeline exploration.

That said, I do see the argument for needing more space as the details grow and potentially more tabs get added other than Details and Summary. A hybrid approach could work well too where we can keepa summary and details view in the tab for quick inspection, and potentially adding a deeper “full-page” view later if needed.

I would defer to @dreveman for his thoughts on what direction would make the most sense moving forward.

dreveman:
[gpu.pb](https://github.com/user-attachments/files/26861533/gpu.pb.txt)

ui/src/plugins/dev.perfetto.GpuCompute/section/launch_statistics.ts
dreveman: I kept this for now but adjusted to "com.meta". we support multiple Section.LaunchStatistics so keeping this is useful

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
dreveman: Done

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
dreveman: Done

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
dreveman: Done

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
dreveman: Done

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
dreveman: Not fully. We need some of it to limit the change to when it changed. ptal.

ui/src/plugins/com.meta.GpuCompute/index.ts
dreveman: Attached gpu.pb trace. It's fake data for testing but will allow you to explore the GpuCompute plugin UI and all the new GPU features added recently.

ui/src/plugins/com.meta.GpuCompute/index.ts
dreveman: Done this way to match how it's done by plugins that want to register a Terminology dynamically. I prefer to keep it this way but could use an alternative way for the built-in terminologies if you feel strongly about it?

dreveman:
> > Given how complex the details panel looks, would it make more sense if we create a new page for all of this instead of tryign to stuff everything in the detgails panel. This is the appraoch we took for the heap dump explorer and I think people find it works really well to have a summary in the details panel and then a separate page for the more detailed internals.
> 
> I personally lean towards the tab-based approach here.
> 
> One of the biggest advantages is that it lets us keep the timeline visible while inspecting compute metrics, which is really valuable when you’re trying to correlate behaviour across kernels. With a separate page, it may be harder for a user to fully understand the context.
> 
> Another key benefit is how it integrates with selection. Right now, when a user selects a GPU slice that contains metrics, we automatically open the Compute tab and jump straight into the relevant details tab instead of the usual Current Selection tab. That makes the workflow much faster and more discoverable for users compared to having to manually navigate somewhere else and select the kernel via the drop-down. This interaction model feels more natural in a tab than in a separate page, since it’s directly tied to timeline exploration.
> 
> That said, I do see the argument for needing more space as the details grow and potentially more tabs get added other than Details and Summary. A hybrid approach could work well too where we can keepa summary and details view in the tab for quick inspection, and potentially adding a deeper “full-page” view later if needed.
> 
> I would defer to @dreveman for his thoughts on what direction would make the most sense moving forward.

I second @mohammadawwad. To add some context, we're also trying to keep this consistent with the most commonly used GPU compute UIs out there (e.g. Nsight Compute) to make it familiar to people but we're also not afraid to improve things when we find ways that area clearly better. @mohammadawwad has done most of the UI design here and improved some things like percentage bars when comparing kernels.

We very much intend to keep iterating on this and make adjustments when convinced that alternative ways improve the experience. I'll take a more in-depth look at the heap dump explorer and other plugins to see if there are things we can use here.

ui/src/plugins/dev.perfetto.GpuCompute/toolbar.ts
dreveman: Good call. Fixed. But some globals related to things that can be extended with plugins remain

ui/src/plugins/dev.perfetto.GpuCompute/index.ts
dreveman: Removed

LalitMaganti:
I'll leave the UX up to you folks as fundamentally you are the ones that will be using it. I think directionally though, in the Perfetto team, our feeling is as folllows:
1) Anything which needs a notion of time to make sense, of course *should* go in the timeline page, makes total sense and is a no brainer
2) If you get into the game of "I need to deep-dive into X" where X is some specific part of the trace with a lot of internal strucutre (e.g. heap dumps), then I think other pages start making a lot of sense honestly.
3) Then the "Overview' page is going to be our "gateway" page where the idea is that you get a summary of *all* the data in the trace, and even potentially probes you could have enabled but are *not* in the trace. The idea of this page is to act as a central hub connecting and summarizing all the data in the trace and "deeplinking" to the timeline or custom pages as appropriate"

ui/src/plugins/com.meta.GpuCompute/index.ts
stevegolton: Are you planning on making this an extension point that other plugins can contribute to? In that case, the registry makes sense. But please don't make it global. Expose it through a public method on the plugin, similar to this: https://github.com/google/perfetto/blob/4f3a228656829bc8a3310b27cab91e42c5022a5b/ui/src/plugins/dev.perfetto.StandardGroups/index.ts#L60

Why do I feel so strongly about global mutable state? We have users who initialize multiple instances of the UI in the same page, and global mutable state tends to break this.

ui/src/plugins/com.meta.GpuCompute/index.ts
dreveman: Done

ui/src/plugins/com.meta.GpuCompute/index.ts
dreveman: Done

ui/src/plugins/com.meta.GpuCompute/index.ts
dreveman: Done. Worked fine

ui/src/plugins/com.meta.GpuCompute/index.ts
dreveman: Yes, we already have internal meta plugins that use these extension points. The globals should have been removed from the latest version of this change.

dreveman:
> I'll leave the UX up to you folks as fundamentally you are the ones that will be using it. I think directionally though, in the Perfetto team, our feeling is as folllows:
> 
> 1. Anything which needs a notion of time to make sense, of course _should_ go in the timeline page, makes total sense and is a no brainer
> 2. If you get into the game of "I need to deep-dive into X" where X is some specific part of the trace with a lot of internal strucutre (e.g. heap dumps), then I think other pages start making a lot of sense honestly.
> 3. Then the "Overview' page is going to be our "gateway" page where the idea is that you get a summary of _all_ the data in the trace, and even potentially probes you could have enabled but are _not_ in the trace. The idea of this page is to act as a central hub connecting and summarizing all the data in the trace and "deeplinking" to the timeline or custom pages as appropriate"

Thanks for the sharing this. Makes sense.

The way I typically use this plugin is by navigating the timeline of large traces with lots of kernels and switching between kernels by clicking on different render stage slices of interest and seeing an updated view of the key metrics in the details view of the GPU Compute tab. That way of using the plugin wouldn't work well if I had to switch to a different page to see the details and that's my main hesitation to changing it in this initial version of the plugin.

There might be a bit too much contents in the details tab now than what's ideal but I think that's OK for this initial version. When it comes to extending the plugin going forward I think we should move to a separate page for deep-dives and limit the tab to key metrics as that's what makes it useful in the timeline view. The fundamentals you provided makes sense and should be something we apply to this plugin going forward.

stevegolton:
@dreveman do you mind if I push a few patches to this PR? There's some stuff I need to play around with before I can know what to recommend, so if I'm going to fix it anyway, I might as well just push it. I won't (intentionally) change any functionality - just architectural stuff.

If that's alright, could we avoid modifying history from now on (no force push)?

dreveman:
> @dreveman do you mind if I push a few patches to this PR? There's some stuff I need to play around with before I can know what to recommend, so if I'm going to fix it anyway, I might as well just push it. I won't (intentionally) change any functionality - just architectural stuff.
> 
> If that's alright, could we avoid modifying history from now on (no force push)?

@stevegolton Sure, go for it! Let me know if you need me to test anything on my end.

stevegolton:
Hey @dreveman, I fixed a few bits that could cause issues in the future - namely:
- Depending on TrackShell to manage collapsing sections - switched to Accordion which is designed for specifically for this (I had to add support for multiple sections to be open at a time which wasn't supported before).
- Dependence on some track tooltip class, which could have broken things subtly if changed.
- Moved some inline styles to scss files (small nitpick/tidyup).

I'm still not too convinced by the loading strategy with the queryslots. I think we're doing a fair bit of double loading, but I think if it works for now and the loading seems light then we can kick the can down the road. I think the main points of tension are the fact that we can show either the current selected slice or the kernel selected from the dropdown, and also we the fact that we need to load some data from TP before we know whether to switch tabs. These are valid usecases that we don't support particularly well, which is something we need to fix, but I don't want to block you on that.

If you wouldn't mind giving it a quick test to confirm I haven't broken anything that'd be great, and I'm happy to approve  after you've confirmed this.

ui/src/plugins/com.meta.GpuCompute/index.ts
```diff
@@ -0,0 +1,414 @@
...
+
+    // Rendering view
+    return m('div', [toolbar, body]);
+  }
+}
+
+// Plugin entry point, registers the tab and menu command
+export default class GpuComputePlugin implements PerfettoPlugin {
+  static readonly id = 'com.meta.GpuCompute';
+  static readonly description = 'TODO: Description for GPU Compute plugin';
```
stevegolton: Would you mind filling in this plugin description - it shows up in the plugins page and helps users understand what it does.

stevegolton (changes requested):

dreveman:
> Hey @dreveman, I fixed a few bits that could cause issues in the future - namely:
> 
> * Depending on TrackShell to manage collapsing sections - switched to Accordion which is designed for specifically for this (I had to add support for multiple sections to be open at a time which wasn't supported before).
> * Dependence on some track tooltip class, which could have broken things subtly if changed.
> * Moved some inline styles to scss files (small nitpick/tidyup).

Thanks!

> 
> I'm still not too convinced by the loading strategy with the queryslots. I think we're doing a fair bit of double loading, but I think if it works for now and the loading seems light then we can kick the can down the road. I think the main points of tension are the fact that we can show either the current selected slice or the kernel selected from the dropdown, and also we the fact that we need to load some data from TP before we know whether to switch tabs. These are valid usecases that we don't support particularly well, which is something we need to fix, but I don't want to block you on that.

Yes, there's some things that are not ideal but I think this is a good start that I can iterate on.

> 
> If you wouldn't mind giving it a quick test to confirm I haven't broken anything that'd be great, and I'm happy to approve after you've confirmed this.

Just did some testing and all seems ok.

ui/src/plugins/com.meta.GpuCompute/index.ts
dreveman: Done

stevegolton (approved):

PR #5451, opened 2026-04-14
protovm: adds the surfaceflinger patch to the trace packet

BogdanFloris:
Adds the LayersPatchProto message to the winscope extensions proto message.


rukkal (approved):

LalitMaganti:
Why is this being added here instead of an extension? We are reducing how much stuff we put in the main TracePacket instead, putting more things in extensions.

BogdanFloris:
Wasn't aware we should be adding this to winscope extensions, thanks for the notice. Just updated the pr.

PR #5454, opened 2026-04-14
ci: install gh CLI in sync-protos and stop gating gen_merged_protos

LalitMaganti:
The self-hosted runner used by the daily sync-protos workflow doesn't
ship with the gh CLI, so the "Close existing proto sync PRs" step has
been failing with "gh: command not found" on every run. Install gh via
the official apt repo at the top of the job so the step can run.

Also drop CheckMergedTraceConfigProto from tools/run_presubmit so that
perfetto_config.proto / perfetto_trace.proto freshness is handled by
the sync-protos action rather than blocking CL presubmit.

### DIFF
--- a/tools/run_presubmit
+++ b/tools/run_presubmit
@@ -538,35 +538,6 @@ def CheckBinaryDescriptors(changed_files: List[str]) -> List[str]:
   return []
 
 
-def CheckMergedTraceConfigProto(changed_files: List[str]) -> List[str]:
-  if sys.platform == 'win32':
-    return []
-  tool_rel_path = 'tools/gen_merged_protos'
-  tool = os.path.join(REPO_ROOT, tool_rel_path)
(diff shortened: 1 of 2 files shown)


tools/run_presubmit
```diff
@@ -538,35 +538,6 @@ def CheckBinaryDescriptors(changed_files: List[str]) -> List[str]:
   return []
 
 
-def CheckMergedTraceConfigProto(changed_files: List[str]) -> List[str]:
```
sashwinbalaji, line 541: can we keep it as a warning/info ?

sashwinbalaji (approved):

tools/run_presubmit
LalitMaganti: There's no point in warnings like this. What is someone meant to do? By definition, our script should be taking care of this.

PR #5472, opened 2026-04-15
tp: delete the legacy PerfettoSqlPreprocessor

LalitMaganti:
**Stack:**

1. #5614 — `tp: prep PerfettoSqlParser for a second backend`
2. #5615 — `tp: add syntaqlite-driven macro expansion behind a flag`
3. #5525 — `tp: replace sqlfluff formatter with syntaqlite` (parallel; based on #5615)
4. #5616 — `tp: route PerfettoSqlParser through syntaqlite macro expansion`
5. **#5472 — `tp: delete the legacy PerfettoSqlPreprocessor`** ← you are here

---

Final PR in a 4-PR stack that splits the original "replace the handrolled preprocessor with syntaqlite" change. With trace_processor routing through the syntaqlite-driven macro path (see the previous PR in the stack), the handwritten preprocessor and the `LegacyImpl`/`constexpr`/`std::optional<>` machinery that gated its existence have no remaining production callers.

This PR:

 * Deletes `src/trace_processor/perfetto_sql/preprocessor/` entirely
   (~2.6k lines of generated grammar + handwritten preprocessor +
   tests) and drops its `BUILD.gn` / `Android.bp` / `BUILD` references
   from the parser, engine, and root `trace_processor` targets.
 * Inlines the real `Macro` struct on `PerfettoSqlParser` (was a
   `using` alias to the preprocessor type) and drops the parser/engine
   includes of the preprocessor header.
 * Collapses `PerfettoSqlParser::Impl` from the dual-impl wrapper into
   the syntaqlite implementation directly: no more `LegacyImpl`,
   `kUseSyntaqliteMacros` / `kUsesSyntaqliteMacros`, `std::optional<>`
   slots, or `if constexpr` dispatch.
 * Drops the `kBang` token alias added to `SqliteTokenizer` when the
   legacy preprocessor needed to recognise `!` as a macro indicator.
 * Drops the `kUsesSyntaqliteMacros` test gate and the
   `ExpandsTokenApplyIntrinsic` skip-on-legacy guard.
 * Refreshes the engine.cc architecture comment to drop the
   preprocessor layer from the diagram.

The earlier PRs in the stack:

 1. `tp: prep PerfettoSqlParser for a second backend` — pure refactor.
 2. `tp: add syntaqlite-driven macro expansion behind a flag` — adds
    `SyntaqliteMacroImpl` + `intrinsic_macro_expansion.{h,cc}` +
    `SqlSource::FromMacroExpansion` + the syntaqlite library bump,
    behind `kUsesSyntaqliteMacros = false`.
 3. `tp: route PerfettoSqlParser through syntaqlite macro expansion` —
    flips the flag.


LalitMaganti:
The `ExpandsMacroInsideArg` crash is partly an upstream syntaqlite issue: the header comment says `body_call_length == 0` signals an arg-internal call, but the implementation actually uses `SYNTAQLITE_MACRO_BODY_CALL_ARG_INTERNAL` (= `UINT32_MAX`).

So `BuildForUserMacro`:

```cpp
if (c.body_call_length == 0) continue;
```

never trips for the arg-internal case.

src/trace_processor/perfetto_sql/parser/intrinsic_macro_expansion.cc
```diff
@@ -0,0 +1,146 @@
...
+  return joined;
+}
+
+}  // namespace
+
+ExpandResult TryExpandIntrinsicMacro(std::string_view name,
+                                     const SyntaqliteToken* args,
+                                     uint32_t arg_count) {
+  if (name == "__intrinsic_stringify" ||
+      name == "__intrinsic_stringify_ignore_table") {
```
sashwinbalaji, line 146: hmm functionality wise are these same ? if so why keep 2 ?

src/trace_processor/perfetto_sql/parser/intrinsic_macro_expansion.h
```diff
@@ -0,0 +1,54 @@
...
+#include <cstdint>
+#include <string>
+#include <string_view>
+
+#include "src/trace_processor/perfetto_sql/syntaqlite/syntaqlite_perfetto.h"
+
+namespace perfetto::trace_processor::perfetto_sql {
+
+struct ExpandResult {
+  enum : uint8_t {
```
sashwinbalaji: nit: enum class

src/trace_processor/perfetto_sql/parser/intrinsic_macro_expansion.cc
```diff
@@ -0,0 +1,146 @@
...
+  auto a = SplitParenList(std::string_view(args[1].text, args[1].length));
+  if (!a)
+    return std::nullopt;
+  std::vector<std::string> b;
+  if (arg_count == 3) {
+    auto parsed =
+        SplitParenList(std::string_view(args[2].text, args[2].length));
+    if (!parsed)
+      return std::nullopt;
+    b = std::move(*parsed);
```
sashwinbalaji: a is std::optional<std::vector<std::string>>
b is std::vector<std::string>

can we name them better and also align both

src/trace_processor/perfetto_sql/parser/intrinsic_macro_expansion.cc
```diff
@@ -0,0 +1,146 @@
...
+  size_t start = 0;
+  for (size_t i = 0; i < s.size(); ++i) {
+    char c = s[i];
+    if (c == '(') {
+      ++depth;
+    } else if (c == ')') {
+      if (depth == 0)
+        return std::nullopt;
+      --depth;
+    } else if (c == ',' && depth == 0) {
```
sashwinbalaji: is it possible to have , within a token ?

src/trace_processor/perfetto_sql/parser/intrinsic_macro_expansion.cc
LalitMaganti: It's because we've changed how we handle macros: in the past, we used to expand macros *eagerly* (i.e. in a CREATE PERFETTO MACRO statement, we would expand any nested macros).

But now we do it lazily so we delay it (i.e. we only expand when we're *not* in a macro creation context). This flattens the difference between these two.

We should remove this in a followup not here.

sashwinbalaji (approved):

PR #5476, opened 2026-04-15
Allow partial recovery on packet drop for ProtoLog messages

PabiGamito:
Currently, we strictly drops packets with the SEQ_NEEDS_INCREMENTAL_STATE flag when the sequence state is invalid (e.g., after a packet drop). This causes some ProtoLog messages to be unnecessarily lost, even when they could be partially recovered or decoded.

### DIFF
--- a/src/trace_processor/importers/proto/proto_trace_reader.cc
+++ b/src/trace_processor/importers/proto/proto_trace_reader.cc
@@ -367,7 +367,12 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
     }
     scoped_state->needs_incremental_state_total++;
 
-    if (!state->IsIncrementalStateValid()) {
+    protos::pbzero::TracePacket::Decoder local_decoder(packet.data(),
+                                                       packet.length());
+    if (!state->IsIncrementalStateValid() &&
+        !local_decoder
+             .Get(protos::pbzero::TracePacket::kProtologMessageFieldNumber)
+             .valid()) {
       if (context_->content_analyzer) {
         // Account for the skipped packet for trace proto content analysis,
         // with a special annotation.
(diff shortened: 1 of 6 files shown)


src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -367,7 +367,12 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
     }
     scoped_state->needs_incremental_state_total++;
 
-    if (!state->IsIncrementalStateValid()) {
+    protos::pbzero::TracePacket::Decoder local_decoder(packet.data(),
+                                                       packet.length());
+    if (!state->IsIncrementalStateValid() &&
+        !local_decoder
```
LalitMaganti, line 373: Sorry but this sort of change is not OK. You cannot just add hacks in core code like this because it's convenient for you. It causes us severe problems for several reasons:
a) when we have to make refactorings or improvements and we have to workaround code like this
b) when someone comes to us with "why is this trace broken" and we have to constantly reason about "is this hack the problem or not"
c) you block things like trace processor pluginzation (e.g. https://github.com/google/perfetto/pull/5363) because now core has references to concepts which only exist for Winscope plugin reasons. And I'm very close to putting all of winscope code in a self-contaiend disablable plugin so this is hindering that very directly.

Not to mention now on every packet type, if incremental state happens to be invalid (not uncommon on AOT traces), you will spend a bunch of cyucles checking "is this protolog packet", hugely inefficient.

It's funny because I actually have a patch for this problem (https://github.com/google/perfetto/pull/5288) which solves this properly. If this problem is analyzed from first principles, you would see there's a very specific property which needs to be adhered to for recovery to be possible. I don't know if protolog adheres to that property or not but if it doesn't, it needs to do so on device and then you can set the new flag and things will just work. I just didn't land that patch because I didn't have a solid usecase, if you can confirm it works for you, I'm happy to land this.

In the future, I would really appreciate if sort of thing was discussed upfront internally rather than just sending patches like this. If you are changing something core like this (I hope this is clear why this is so core), we talk through "what is the right way to solve this". Becuase it cannot be "I'm going to add a hack".

PabiGamito: Ok I see, very fair points. Apologies for the hacky PR and skipping the internal discussion stage. I'm closing this CL and we can discuss offline what a proper solution for this would look like.

PR #5477, opened 2026-04-16
perfetto: add string_view overloads for base::TrimWhitespace

LalitMaganti:
Introduces std::string_view and const char* overloads of
base::TrimWhitespace and refactors the existing std::string
implementation on top of them.  This avoids a mandatory copy when
callers only need to observe the trimmed range and opens the door to
using TrimWhitespace in code that already works in terms of
string_view.

### DIFF
--- a/src/base/string_utils.cc
+++ b/src/base/string_utils.cc
@@ -134,8 +134,9 @@ std::string_view TrimWhitespace(std::string_view str) {
     return {};
   std::string_view front_trimmed = str.substr(front_idx);
   size_t end_idx = front_trimmed.find_last_not_of(kWhitespaces);
-  return end_idx == std::string_view::npos ? std::string_view()
-                                           : front_trimmed.substr(0, end_idx + 1);
+  return end_idx == std::string_view::npos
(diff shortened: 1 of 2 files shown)


src/base/string_utils.cc
```diff
@@ -127,16 +127,24 @@ std::vector<std::string> SplitString(const std::string& text,
...
-  std::string whitespaces = "\t\n ";
+std::string_view TrimWhitespace(std::string_view str) {
+  static constexpr char kWhitespaces[] = "\t\n ";
+  size_t front_idx = str.find_first_not_of(kWhitespaces);
+  if (front_idx == std::string_view::npos)
+    return {};
+  std::string_view front_trimmed = str.substr(front_idx);
+  size_t end_idx = front_trimmed.find_last_not_of(kWhitespaces);
+  return end_idx == std::string_view::npos
+             ? std::string_view()
```
sashwinbalaji: I don't think we need this check. end_idx will always be atleast = 0 right ?

sashwinbalaji (approved):

PR #5479, opened 2026-04-16
ui: Move legacy trace actions to a dedicated plugin (disabled by default)

stevegolton:
This patch moves the 'Open in legacy UI' and 'Switch to legacy UI' sidebar actions as well as the corresponding commands and logic to a new 'dev.perfetto.Catapult' plugin, which is disabled by default.

It's there is anyone really wants it, but it's hidden by default. We could remove this in the future very simply by deleting the plugin.


LalitMaganti (approved):

primiano (approved):
Great idea.

PR #5480, opened 2026-04-16
stdlib: Wattson: Handle multiple forms of perf events

wusamuel6:
Depending on whether the perf event comes from the Perfetto config or from running simpleperf concurrently as the Perfetto trace, the data format saved is different. Handle this.

If the perf event is from Perfetto config polling, then the counters will be monotonically increasing. Example config below:

data_sources {
  config {
    name: "linux.perf"
    perf_event_config {
      target_cpu: 0
      timebase {
        poll_period_ms: 10
        raw_event {
          type: 13
          config: 43
        }
        timestamp_clock: PERF_CLOCK_MONOTONIC
        name: "arm_dsu_0-l3d_cache"
      }
    }
  }
}

If the counters come from concurrently running simpleperf, then the table has counts per period. Example command below:

simpleperf stat -a -e arm_dsu_0/l3d_cache/ --interval 10

Bug: 503368151

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/wattson/cpu/arm_dsu.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/wattson/cpu/arm_dsu.sql
@@ -34,7 +34,13 @@ SELECT
   ts,
   lead(ts) OVER (PARTITION BY track_id ORDER BY ts) - ts AS dur,
   -- Rate of event accesses in a section (i.e. count / dur).
-  value / (
+  -- If the event name ends in '_cpu0', then the counter is "counts per period".
+  -- If the event name does not end in '_cpu0', then the counter is monotonic.
+  iif(
+    $event GLOB "*_cpu0",
+    value,
+    lead(value) OVER (PARTITION BY track_id ORDER BY ts) - value
+  ) * 1.0 / (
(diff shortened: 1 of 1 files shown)


LalitMaganti:
This seems very suspicious change. Why are you not using perf record instead of perf stat? That should be 1:1 with Perfetto. This sort of "merging" you are doing here cannot possibly scale and will cause you pain in the long term I can guarantee. 

Up to you I will lgtm if you ask but I would suggest you reconsider

wusamuel6:
> This seems very suspicious change. Why are you not using perf record instead of perf stat? That should be 1:1 with Perfetto. This sort of "merging" you are doing here cannot possibly scale and will cause you pain in the long term I can guarantee.

Fair point, we want this to scale moving forward. I will move everything over to use `perf record` or the Perfetto config version of it, since those seem to be the canonical way.

> Up to you I will lgtm if you ask but I would suggest you reconsider

There are quite a few pre-existing traces that have used `perf stat`, so I'd like to get this in to at least retain support for those traces. Fortunately, this is the only event/case of this, so moving forward, everything else we can guarantee will be on the `perf record` version.

LalitMaganti (approved):

PR #5481, opened 2026-04-16
ui: allow ui/build to act as a client of a running dev-server

primiano:
Previously, running ui/build while ui/run-dev-server was already active
would fail out: the two would clash in out/ui/ and a PID lockfile
prevented the second invocation. This made it awkward to iterate on
TypeScript sources while a watch-mode server was keeping tsc/rollup
warm, especially for AI agents that edit code and need synchronous
feedback on whether the build succeeded.

Repurpose ui/build, when a watch-mode server is detected, as a thin
client that rides the server's existing tsc-watch/rollup-watch
pipeline. Communication is purely file-based under out/ui/:

  - out/ui/watch.lock: JSON with {pid, watch}. Replaces the
    previous PID-only lockfile. prepareBuildLock() now also detects
    "server alive + watch: true" and returns the server's info so the
    caller can enter client mode; concurrent full builds still bail.

  - out/ui/ui_build.log: append-mode log of tsc/rollup/sass/etc. stdout+stderr
    plus this script's console.log, each line timestamped and
    ANSI-stripped. Truncated at server startup and after each client
    run, so it acts as a per-iteration buffer.

  - out/ui/dev_server.port: written on HTTP "listening" so other tools (test
    runners, agents) can discover the dev server's port without
    parsing logs.

  - out/ui/tsc/gen/perfetto_version.ts: the client rewrites SCM_REVISION to
    "perfetto_incremental_build:N" (monotonic). tsc-watch and
    rollup-watch cascade this through the bundle as a fence -- once
    frontend_bundle.js contains the same marker, every edit made before
    the client ran is guaranteed to be in the bundle.

The client polls the bundle for the fence while tailing ui_build.log
to its own stdout. Exit codes: 0 on fence match, 1 on tsc "Found N
errors" or rollup "[!] ..." detection or 120s timeout.

Server-side plumbing: exec() now pipes and tees subprocess output when
logStream is active; the one synchronous stdout-capturing call (the
version script) was inlined with spawnSync to keep exec() simple.


stevegolton (approved):
Looks good to me % one small issue insofar that the ninja output of the wasm build is no longer showing up in the terminal.

Edit: It also did seem to be a bit racy for me. Let's discuss offline.

primiano:
Unnecessary after moving to vite

PR #5485, opened 2026-04-17
doc: Document aflags data source

sashwinbalaji:


docs/data-sources/android-aflags.md
```diff
@@ -0,0 +1,80 @@
...
+from android_aflags
+where value_picked_from != 'default'
+```
+
+package | name | value | value_picked_from | storage_backend
+--------|------|-------|-------------------|----------------
+com.android.window.flags | enable_multi_window | enabled | server | device_config
+com.android.systemui.flags | new_notification_header | disabled | local | aconfigd
+
+If the `aflags` tool fails at runtime, a per-trace error is recorded under the stat name `android_aflags_errors` in the `_trace_import_logs` table.
```
LalitMaganti: Never document private tables like this. Very bad practice to do do. Either make the table public or remove this.

LalitMaganti (approved):

PR #5489, opened 2026-04-17
actions: use deploy-key for pushing to canary/stable

LalitMaganti:


stevegolton (approved):

primiano:
Wait insn't there some other way to allow a GH action to push to the same repo?
it feels quite weird that we have to bake a ssh key into a secret for that

PR #5492, opened 2026-04-17
ui: Drive channel deploys from canary/stable branches, race-free

LalitMaganti:
Replace channels.json + ui-canary/ui-stable with the long-lived
canary/stable branches as the source of truth (per RFC-0022). Each
Cloud Build trigger now builds exactly one channel from the branch
HEAD that fired it: main->autopush, canary->canary, stable->stable.

To stay parallel-safe without touching the UI bootstrap or AppEngine,
each deploy modifies only its own entry in the shared
data-perfetto_version map on root /index.html, using GCS
x-goog-if-generation-match as a CAS primitive. Stable additionally
swaps the HTML body and writes /service_worker.* (still preserving
the other channels' map entries). Canary/autopush never write the
body or service_worker, so canary instability cannot break stable.

- New ui/release/build_channel.py replaces build_all_channels.py.
- ui_builder_entrypoint.sh now takes the channel name as $1; no
  git checkout (Cloud Build already places us on the right SHA).
- cloudbuild.yaml passes 'autopush'; cloudbuild_release.yaml still
  passes $BRANCH_NAME (canary or stable).
- channels.json, roll_branch.py, roll_stable.sh, roll_canary.sh
  are deleted -- branch HEADs are now the source of truth.


stevegolton (approved):
Nice!

PR #5501, opened 2026-04-20
ui: Highlight same-name slices globally

stevegolton:
When hovering a slice, highlight all slices across all tracks with the same name.

Related: https://github.com/google/perfetto/pull/5494

### DIFF
--- a/ui/src/public/timeline.ts
+++ b/ui/src/public/timeline.ts
@@ -204,6 +204,7 @@ export interface Timeline {
 
   hoveredNoteTimestamp: time | undefined;
   highlightedSliceId: number | undefined;
+  highlightedSliceName: string | undefined;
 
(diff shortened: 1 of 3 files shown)


ui/src/public/timeline.ts
```diff
@@ -204,6 +204,7 @@ export interface Timeline {
 
   hoveredNoteTimestamp: time | undefined;
   highlightedSliceId: number | undefined;
+  highlightedSliceName: string | undefined;
```
primiano, line 207: maybe add a comment on highlightedSliceId saying it's deprecated and is there only for flow events with a TODO

primiano (approved):

ui/src/public/timeline.ts
stevegolton: Done

PR #5509, opened 2026-04-20
traced_probes: defer aflags Flush() callback on in-flight subprocess

sashwinbalaji:
When a trace with android.aflags enabled ends while the aflags_updatable subprocess is still running, the previous Flush() implementation acked the producer immediately. The producer then destroyed the data source, killing the subprocess mid-write and dropping the aflags packet from the trace.

```
  04-15 11:03:50.410 F DEBUG  : Abort message: 'failed printing to stdout: Broken pipe (os error 32)'
  ```

This change defers the Flush() callback while a subprocess is in flight:

- Fast path: if aflags_process_ is empty, flush the writer immediately.
- Deferred path: store the callback in pending_flushes_ keyed by flush_request_id, and arm an 800ms safety-net timeout.
- FinalizeAflagsCapture() drains pending_flushes_ at the end of its successful-completion path, so the producer is acked only after the aflags packet has been written. If the subprocess runs past the timeout, OnFlushTimeout logs it and delegates to OnFlushComplete so the producer isn't left waiting.

```
  04-20 15:05:27.694 I perfetto: Triggering final flush for 2
  04-20 15:05:27.696 I perfetto: ProbesProducer::Flush(2) begin
  04-20 15:05:27.913 I perfetto: Flush 2 acked by data source 2
  04-20 15:05:27.914 I perfetto: FlushAndDisableTracing(2) done, success=1
  04-20 15:05:27.916 I perfetto: Producer stop (id=2)
  04-20 15:05:27.964 I perfetto: Wrote 469696 bytes into /data/.../aflags.pftrace  
  04-20 15:05:27.694 I perfetto: Triggering final flush for 2
  04-20 15:05:27.696 I perfetto: ProbesProducer::Flush(2) begin
  04-20 15:05:27.913 I perfetto: Flush 2 acked by data source 2
  04-20 15:05:27.914 I perfetto: FlushAndDisableTracing(2) done, success=1
  04-20 15:05:27.916 I perfetto: Producer stop (id=2)
  04-20 15:05:27.964 I perfetto: Wrote 469696 bytes into /data/.../aflags.pftrace
  ```

Also documents ReadFileDescriptor()'s append-and-EOF-return semantics.

Bug: 502852101
Test: perfetto_unittests --gtest_filter='AndroidAflagsDataSourceTest.*' 
Test: manually captured a 100ms duration_ms trace on device with android.aflags enabled; aflags packet now lands in the trace instead of being dropped, no aflags_updatable tombstone.

### DIFF
--- a/include/perfetto/ext/base/file_utils.h
+++ b/include/perfetto/ext/base/file_utils.h
@@ -49,6 +49,8 @@ inline constexpr char kFopenReadFlag[] = "re";
 
 constexpr FileOpenMode kFileModeInvalid = static_cast<FileOpenMode>(-1);
 
+// Cross-platform variant of ReadFileDescriptor() that takes a PlatformHandle.
+// On Windows normalizes ERROR_BROKEN_PIPE to EOF so behavior matches POSIX.
 bool ReadPlatformHandle(PlatformHandle, std::string* out);
 
 // Reads from |fd|, appending what is currently available into |*out|.
@@ -59,7 +61,11 @@ bool ReadPlatformHandle(PlatformHandle, std::string* out);
 //    IsAgain(errno) and retry on the next readability notification.
 bool ReadFileDescriptor(int fd, std::string* out);
 
+// Convenience wrapper around ReadFileDescriptor() that takes a FILE*.
 bool ReadFileStream(FILE* f, std::string* out);
+
+// Opens |path| read-only and reads its contents into |*out|.
+// Returns false if the file cannot be opened.
 bool ReadFile(const std::string& path, std::string* out);
 
 // A wrapper around read(2). It deals with Linux vs Windows includes. It also
--- a/include/perfetto/ext/base/file_utils.h
+++ b/include/perfetto/ext/base/file_utils.h
@@ -50,7 +50,15 @@ inline constexpr char kFopenReadFlag[] = "re";
 constexpr FileOpenMode kFileModeInvalid = static_cast<FileOpenMode>(-1);
 
 bool ReadPlatformHandle(PlatformHandle, std::string* out);
+
+// Reads from |fd|, appending what is currently available into |*out|.
+// Returns:
+//  True: EOF reached (all writers of |fd| have closed their end).
+//  False: read error. On a non-blocking |fd| this includes EAGAIN (no data
+//    currently available but writers are still alive); callers can check
+//    IsAgain(errno) and retry on the next readability notification.
 bool ReadFileDescriptor(int fd, std::string* out);
+
 bool ReadFileStream(FILE* f, std::string* out);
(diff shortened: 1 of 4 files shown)


include/perfetto/ext/base/file_utils.h
```diff
@@ -50,7 +50,15 @@ inline constexpr char kFopenReadFlag[] = "re";
...
 
 bool ReadPlatformHandle(PlatformHandle, std::string* out);
+
+// Reads from |fd|, appending what is currently available into |*out|.
+// Returns:
+//  True: EOF reached (all writers of |fd| have closed their end).
+//  False: read error. On a non-blocking |fd| this includes EAGAIN (no data
+//    currently available but writers are still alive); callers can check
+//    IsAgain(errno) and retry on the next readability notification.
 bool ReadFileDescriptor(int fd, std::string* out);
```
LalitMaganti, line 62: Comment here seems unnecesary and inconsistent with everything else.

LalitMaganti (approved):

src/traced/probes/android_aflags/android_aflags_data_source.h
```diff
@@ -73,12 +75,23 @@ class AndroidAflagsDataSource : public ProbesDataSource {
...
+
   base::TaskRunner* const task_runner_;
   std::unique_ptr<TraceWriter> writer_;
 
   // Polling frequency in ms. 0 means one-shot capture at startup.
   uint32_t poll_period_ms_ = 0;
 
+  // Flush() callbacks deferred while an aflags subprocess is in flight.
+  // Drained when the subprocess completes (or when the flush timeout fires).
+  std::unordered_map<FlushRequestID, std::function<void()>> pending_flushes_;
```
LalitMaganti, line 93: Why does this need to be a map btw? AFAIK we cannot have multiple flushes active at any one time.

sashwinbalaji: So there is a very corner case, somebody sets a very small flush_period_ms. It's not practical but possible

Our current check is it needs to be just greater than 0.

```
  if (cfg.flush_period_ms() > 0) {
    tracing_session->flush_strategy = TracingSession::FlushStrategy::kPeriodic;
    tracing_session->periodic_flush_ms = cfg.flush_period_ms();
  } else {
    tracing_session->flush_strategy = TracingSession::FlushStrategy::kDisabled;
  }
```

include/perfetto/ext/base/file_utils.h
sashwinbalaji: It took me an dprimiano sometime to understand what the function is doing. I would like to keep the comment and instead added some more for the surrounding ones to make it consistent

PR #5510, opened 2026-04-20
ui: Accordion widget - Add 'multi' option

stevegolton:
Accordion widget currently only allows one section to be open at one time (by design - this is a feature).

It can be handy to be able to open and close multiple sections at once.

This patch adds a 'multi' option, to enable this behavior, and also allows each section to be start open with a `defaultOpen` prop.

This patch also refactors the Accordion widget to use a composite component approach:
```ts
m(Accordion,
  m(AccordionSection, {header: 'Foo'}, 'Foo content'),
  m(AccordionSection, {header: 'Bar'}, 'Bar content'),
)
```

### DIFF
--- a/ui/src/widgets/accordion.ts
+++ b/ui/src/widgets/accordion.ts
@@ -28,7 +28,10 @@ export interface AccordionItem {
 export interface AccordionAttrs {
   // The list of accordion items to display.
   readonly items: AccordionItem[];
-  // Currently expanded item id (controlled mode).
+  // Allow multiple items to be open simultaneously. Defaults all items to
+  // expanded on first render. Incompatible with controlled mode (expanded).
+  readonly multi?: boolean;
+  // Currently expanded item id (controlled single-open mode).
   // If omitted, the accordion manages its own state (uncontrolled mode).
   readonly expanded?: string;
   // Callback when an item is toggled.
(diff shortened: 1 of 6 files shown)


ui/src/widgets/accordion.ts
```diff
@@ -28,7 +28,10 @@ export interface AccordionItem {
 export interface AccordionAttrs {
   // The list of accordion items to display.
   readonly items: AccordionItem[];
-  // Currently expanded item id (controlled mode).
+  // Allow multiple items to be open simultaneously. Defaults all items to
```
aMayzner: I'd rather see things defaulting to collapsed, otherwise it's kinda like sections with collapsibility.

aMayzner (approved):

ui/src/widgets/accordion.ts
stevegolton: I've changed them to default to collapsed in the latest commit, but you can configure each section to be open by default.

ui/src/widgets/accordion.ts
```diff
@@ -13,90 +13,93 @@
...
+  readonly key?: string;
+  readonly multi?: boolean;
 }
 
 export class Accordion implements m.ClassComponent<AccordionAttrs> {
-  // Internal state for uncontrolled mode.
-  private internalExpanded: string | undefined = undefined;
-  // Track which item should be scrolled into view after render.
-  private pendingScrollId: string | undefined = undefined;
+  private readonly uuid = shortUuid();
```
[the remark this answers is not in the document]
stevegolton: This is absolutely not in a security context.

PR #5515, opened 2026-04-20
traced: Simpler config for perf tracing

wusamuel6:
This change allows users to specify hardware events by their name (e.g.,
'l3d_cache') instead of raw config IDs. The profiler resolves these
names at runtime by reading from sysfs. To ensure safety, it also reads
the PMU's format/event file to verify the event variable maps directly
to the lowest bits of the config field.

Example Perfetto config:

Old:
raw_event {
  pmu_name: "arm_dsu_0"
  config: 43
}

New:
raw_event {
  pmu_name: "arm_dsu_0"
  event_name: "l3d_cache"
}

Bug: 501821381
Test: manual test

### DIFF
--- a/src/profiling/perf/event_config.cc
+++ b/src/profiling/perf/event_config.cc
@@ -100,6 +100,38 @@ std::optional<uint32_t> ReadPmuTypeFromSysfs(const std::string& pmu_name) {
   return base::StringToUInt32(base::TrimWhitespace(buf));
 }
 
+std::optional<uint64_t> ReadEventConfigFromSysfs(
+    const std::string& pmu_name,
+    const std::string& event_name) {
+  if (pmu_name.empty() || event_name.empty())
+    return std::nullopt;
+  if (base::Contains(pmu_name, '/') || base::StartsWith(pmu_name, ".."))
+    return std::nullopt;
+  if (base::Contains(event_name, '/') || base::StartsWith(event_name, ".."))
+    return std::nullopt;
+
+  std::string format_path =
+      "/sys/bus/event_source/devices/" + pmu_name + "/format/event";
+  std::string format_buf;
+  if (!base::ReadFile(format_path, &format_buf)) {
+    PERFETTO_ELOG("Failed to read %s", format_path.c_str());
+    return std::nullopt;
+  }
+  if (!base::StartsWith(base::TrimWhitespace(format_buf), "config:0-")) {
+    PERFETTO_ELOG("Unsupported event format '%s' for PMU '%s'",
+                  base::TrimWhitespace(format_buf).c_str(), pmu_name.c_str());
+    return std::nullopt;
+  }
+
+  std::string event_path =
+      "/sys/bus/event_source/devices/" + pmu_name + "/events/" + event_name;
+  std::string buf;
+  if (!base::ReadFile(event_path, &buf))
+    return std::nullopt;
+
+  return EventConfig::ParseEventConfigValue(buf);
+}
+
 // |T| is either gen::PerfEventConfig or gen::PerfEventConfig::Scope.
 // Note: the semantics of target_cmdline and exclude_cmdline were changed since
 // their original introduction. They used to be put through a canonicalization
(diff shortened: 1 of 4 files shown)


src/profiling/perf/perf_producer.cc
```diff
@@ -101,15 +101,61 @@ bool IsCpuOnline(uint32_t cpu) {
...
-// automatically downscope events to the cpus that they're present on (matters
-// for heterogeneous cores and ppmu/uncore events).
-// This lets users use "perf record -a" without worrying about cpu scopes.
-std::vector<uint32_t> CreateCpuMask(const protos::gen::PerfEventConfig& cfg) {
-  const auto& target_cpus_raw = cfg.target_cpu();
-  std::set<uint32_t> target_cpus(target_cpus_raw.begin(),
-                                 target_cpus_raw.end());
+}  // namespace
+
+std::set<uint32_t> ParseCpuList(const std::string& input) {
```
rsavitski: if this function is still needed given comments below: There's already one case of this parser in Tracefs::GetOfflineCpus() (at src/traced/probes/ftrace/tracefs.cc).

It'd be nicer if we could move that function to src/kernel_utils/ and make it a general linux cpulist parser, called from here for both `IsCpuOnline` and the new code.

src/profiling/perf/perf_producer_unittest.cc
```diff
@@ -200,3 +200,16 @@ TEST(TargetFilterTest, AdditionalCmdlines) {
 }  // namespace
 }  // namespace profiling
 }  // namespace perfetto
+
+TEST(PerfProducerTest, ParseCpuList) {
```
rsavitski: nit: should be inside the namespaces

src/profiling/perf/perf_producer.cc
```diff
@@ -101,15 +101,61 @@ bool IsCpuOnline(uint32_t cpu) {
...
+  for (uint32_t cpu : cfg.target_cpu()) {
+    target_cpus.insert(cpu);
+  }
+
+  // If no CPUs were specified in the config, and we have PMU events,
+  // consult the PMUs' cpumasks.
+  if (target_cpus.empty()) {
+    auto add_pmu_cpus = [&](const std::string& pmu_name) {
+      if (pmu_name.empty())
+        return;
```
rsavitski: Several thoughts in no particular order:
* if we support cpumask, we should do it consistently for all configurations (even if someone manually sets the pmu `type` via raw events), this patch will surprise people since only "symbolic pmu name + empty cpu list" kicks in the cpumask handling. But then we'll need to iterate over the sysfs pmus to find the path for a raw `type`. Maybe it's just a case of documenting the caveats in the config proto, but I'd err on the side of not committing partial support for masking.
* imo the cpumask should be an additional filtering layer for the config-specified cpu list, with an error if the set of cpus is empty after said filtering. I.e. make sure that even if the user specifies explicit cpus, the masking corrects them. This is what the desktop `perf` tool does afaik.
* one alternative to /cpumask could be just trying the syscall and handling ENOENT gracefully, but I'm still unsure if there are caveats for the latter approach.

I'd probably just drop the cpumask part of the patch for now, and focus on sysfs event_name.

src/profiling/perf/event_config.cc
```diff
@@ -362,6 +395,22 @@ bool IsSupportedUnwindMode(
...
+std::optional<uint64_t> EventConfig::ParseEventConfigValue(
+    const std::string& content) {
+  for (const auto& part : base::SplitString(content, ",")) {
+    std::string trimmed = base::TrimWhitespace(part);
+    if (base::StartsWith(trimmed, "event=")) {
+      return base::StringToUInt64(trimmed.substr(6), 16);
+    }
+    if (base::StartsWith(trimmed, "config=")) {
+      return base::StringToUInt64(trimmed.substr(7), 16);
+    }
```
rsavitski: This is making too many assumptions without verifying them, and the "config=" case in particular looks very ad-hoc.

Technically the way it works is that .../pmu/format/* defines which variables map onto which bits of config/config1/config2, and .../pmu/events/* have the comma-separated key=value set, and the perf tool needs to derive the syscall values using that mapping.

I am very open to limit scope and initially support only the "event=" key (which should cover most typical PMUs), but we should still verify using pmu/format/event that it maps onto the lowest bits of `config`.

wusamuel6: ACK, thanks for the feedback. Makes sense, and dropped the cpumask stuff.

PR #5519, opened 2026-04-21
trace_processor: Parse adreno_cmdbatch_retired/submitted events

rossning92:
Add custom parsing for kgsl adreno_cmdbatch_retired and adreno_cmdbatch_submitted ftrace events to display GPU slices in Perfetto UI.

Test: built and loaded trace from Adreno device, verified GPU slices appear under correct tracks in Perfetto UI:
image

Cross-validated with Qualcomm's Render stage trace (useful for GPU debugging but has large CPU and some GPU overhead): 
image


src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -1920,6 +1932,33 @@ void FtraceParser::ParseKgslGpuFreq(int64_t timestamp, ConstBytes blob) {
...
+void FtraceParser::ParseKgslAdrenoCmdbatchRetired(int64_t timestamp,
+                                                  protozero::ConstBytes data) {
+  protos::pbzero::KgslAdrenoCmdbatchRetiredFtraceEvent::Decoder evt(data);
+
+  static constexpr auto kBlueprint = TrackCompressor::SliceBlueprint(
+      "gpu_adreno_cmdbatch",
+      tracks::DimensionBlueprints(tracks::UintDimensionBlueprint("context_id"),
+                                  tracks::UintDimensionBlueprint("prio")),
+      tracks::FnNameBlueprint([](uint32_t context_id, uint32_t prio) {
+        return base::StackString<64>("GPU (Ctx=%u, Prio=%u)", context_id, prio);
```
LalitMaganti: The name should usually indicate the actual source of these events. If someone just see "gpu" they are going to have absolutely no clue what the data inside that contains.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -136,6 +136,13 @@ using protos::pbzero::perfetto_pbzero_enum_KprobeEvent::KprobeType;
 using protozero::ConstBytes;
 using protozero::ProtoDecoder;
 
+// Adreno GPU ticks run at 19.2 MHz.
+constexpr uint64_t kGpuTicksPerUs = 19200;
```
LalitMaganti: If this is Adreno specific, why not just inline it? I don't see how this would be reused in another context?

LalitMaganti: Also are you sure that *all* GPUs run at this frequency or just the one you care about? I feel quite suspicious about this constant tbh.

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -69,6 +70,13 @@ namespace {
 
 constexpr uint32_t kSequenceScopedClockId = 64;
 
+// Adreno GPU ticks run at 19.2 MHz.
+constexpr uint64_t kGpuTicksPerUs = 19200;
+
+uint64_t GpuTicksToNs(uint64_t ticks) {
```
LalitMaganti: Again this is Adreno specific and so should be inlined

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -301,6 +309,41 @@ void FtraceTokenizer::TokenizeFtraceEvent(
...
+    auto retired_field = GetFtraceEventField(
+        protos::pbzero::FtraceEvent::kKgslAdrenoCmdbatchRetiredFieldNumber,
+        event);
+    if (retired_field.has_value()) {
+      protos::pbzero::KgslAdrenoCmdbatchRetiredFtraceEvent::Decoder evt(
+          retired_field->as_bytes());
+
+      const uint32_t ts = evt.timestamp();
+      auto* info = cmdbatch_sync_map_.Find(ts);
+      if (info && evt.start() >= info->ticks) {
```
LalitMaganti: Sorry but I think this whole flow you are doing is not making any sense to me.

I don't understand why we are not just passing through both the retired and submitted as individual ftrace events and then doing the slicification on the parser side.

Right now this feels like you are doing too much computation here and not enough in the parser. Specifically, you want to plumb these through as proper begin/end events and you can kill almost all the code here. I definitely don't think you should need any hashmap at all, that feels off.

src/trace_processor/importers/ftrace/ftrace_parser.cc
rossning92: I deeply appreciated thoughtful suggestions!  
Done - renamed to "Adreno Cmdbatch (Ctx=%u, Prio=%u)".

src/trace_processor/importers/ftrace/ftrace_parser.cc
rossning92: Yup all Adreno GPUs rely on this internal GPU clock:  
`#define KGSL_XO_CLK_FREQ 19200000` (defined in kernel/qcom/drivers/gpu/msm/kgsl_pwrctrl.h), and we've tested for all GPUs from a5xx, a6xx, all the way to the latest a8xx GPUs.
(Revised the comment to clarify that too)

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
rossning92: Done - inlined.

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
rossning92: So the reason that I've done it in the tokenizer here is that timestamp correction must happen in the tokenizer (before sorting) because the ftrace' CPU timestamps of adreno_submitted/adreno_retired events do NOT correspond to when the GPU actually started/finished work - they reflect when the CPU enqueued or was notified, which can be significantly later.

In order to get real GPU execution time, adreno_submitted event provides a CPU <-> GPU clock sync point (secs/usecs + ticks). We use this to map the GPU start tick count back to the raw clock domain: raw_timestamp = sync_time + (start_ticks - sync_ticks) / tick_rate. And it's very important that the GPU event is accurately aligned with the existing CPU event in the perfect trace.

(I added comments in the code to clarify this as well.)

But please definitely let me know if there are better ways to do this!

LalitMaganti: I'm understand why you're need code here in the tokenizer (i.e. for fixing up timestamps). I guess I'm objecting more to how you are doing it rather than the fact you are doing it.

Well a few things:
1) You could use clock tracker and register this as a new sequence scoped clock or similar and get TP to do the clock tracking for you. It's generally better to use central infra for things like this instead of ad-hoc.
2) Even if you do it manuallty, I don't understand why youa re passing them through as scoped events rather than begin/end and letting TP do the slicification. Generally scoped events are something we try to avoid as much as possible as they have lots of gotchas.

rossning92: Many thanks for the suggestion @LalitMaganti,  
1. I updated the PR to use `clock_tracker` for CPU<->GPU clock domain synchronization, following the existing code path. 
2. I fully agree that it would be cleaner if we called Begin() on the adreno_submitted event and End() for the adreno_retired event. However the problem is that both actual GPU timestamps (evt.start() and evt.retire(), as in GPU ticks) are only in the adreno_retired event. i.e., the submitted event can't call Begin() because it doesn't know when the GPU actually started, which is why I initially use `Scoped` when parsing the Adreno_retired event.
Let me know if this makes sense to you!

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -301,6 +304,60 @@ void FtraceTokenizer::TokenizeFtraceEvent(
     return;
   }
 
+  // The submitted event provides a CPU <-> GPU clock sync point; we register
+  // it as a sequence-scoped clock so the clock tracker can convert GPU tick
+  // timestamps to trace time, skipping duplicates since the GPU clock runs at
+  // a fixed frequency and a single snapshot is sufficient.
+  // Adreno GPU ticks run at 19.2 MHz (KGSL_XO_CLK_FREQ in kgsl_pwrctrl.h).
+  constexpr int64_t kAdrenoGpuTicksPerUs = 19200;
+  if (event_id ==
```
LalitMaganti: Please refactor this code out into functions which would make them much easier to read.

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -301,6 +304,60 @@ void FtraceTokenizer::TokenizeFtraceEvent(
...
+      protos::pbzero::FtraceEvent::kKgslAdrenoCmdbatchSubmittedFieldNumber) {
+    auto submitted_field = GetFtraceEventField(
+        protos::pbzero::FtraceEvent::kKgslAdrenoCmdbatchSubmittedFieldNumber,
+        event);
+    if (submitted_field.has_value()) {
+      protos::pbzero::KgslAdrenoCmdbatchSubmittedFtraceEvent::Decoder evt(
+          submitted_field->as_bytes());
+      if (!adreno_gpu_clock_registered_) {
```
LalitMaganti: This will cause only the first one to be used for all subsequent. Maybe you want that, maybe not, not sure, depends on your event.

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -301,6 +304,60 @@ void FtraceTokenizer::TokenizeFtraceEvent(
...
+              ClockTracker::ClockTimestamp(gpu_clock, gpu_ticks_ns),
+              ClockTracker::ClockTimestamp(clock_id, sync_time_ns),
+          });
+          adreno_gpu_clock_registered_ = true;
+        }
+      }
+    }
+  } else if (event_id == protos::pbzero::FtraceEvent::
+                             kKgslAdrenoCmdbatchRetiredFieldNumber) {
+    auto retired_field = GetFtraceEventField(
```
LalitMaganti: Same asa bove, please pull this out.

Also like the other Tokenize* funcitons, instead of modifying timestmaps and falling through, it would benicer if you just pushed to the sorter queues yoruself

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -301,6 +304,60 @@ void FtraceTokenizer::TokenizeFtraceEvent(
     return;
   }
 
+  // The submitted event provides a CPU <-> GPU clock sync point; we register
```
LalitMaganti: This needs some diff tests in the style of other events.

rossning92: Thanks @LalitMaganti ,

1. Addressed all comments and added a diff test.
2. Build and confirm that the Perfetto UI still works exactly as expected as before.

```
$ tools/ninja -C out/host trace_processor_shell
ninja: Entering directory `out/host'
[1/7] compile ../../src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
[2/7] compile ../../src/trace_processor/importers/ftrace/ftrace_module_impl.cc
[3/7] compile ../../src/trace_processor/importers/proto/additional_modules.cc
[4/7] compile ../../src/trace_processor/importers/ftrace/ftrace_parser.cc
[5/7] stamp obj/src/trace_processor/importers/ftrace/full.stamp
[6/7] stamp obj/src/trace_processor/importers/proto/full.stamp
[7/7] link ./trace_processor_shell

$ .venv/bin/python3 tools/diff_test_trace_processor.py out/host/trace_processor_shell --name-filter 'AdrenoCmdbatch'
[==========] Running 1 tests.
[ RUN      ] AdrenoCmdbatch:adreno_cmdbatch_retired_slice
[       OK ] AdrenoCmdbatch:adreno_cmdbatch_retired_slice (ingest: 494.77 ms query: 0.56 ms)
[==========] Name filter selected 1 tests out of 1357.
[==========] 1 tests ran out of 1 total. (1066 ms total)
[  PASSED  ] 1 tests.
```

LalitMaganti (approved):

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
```diff
@@ -493,6 +512,91 @@ FtraceTokenizer::HandleFtraceClockSnapshot(
...
+      gpu_clock, gpu_start_ns, std::nullopt, /*suppress_errors=*/true);
+  if (ts.has_value()) {
+    module_context_->PushFtraceEvent(
+        cpu, *ts, TracePacketData{std::move(event), std::move(state)});
+    return;
+  }
+
+  std::optional<int64_t> timestamp = context_->clock_tracker->ToTraceTime(
+      clock_id, static_cast<int64_t>(raw_timestamp));
+  if (!timestamp.has_value()) {
```
sashwinbalaji, line 593: optional: you can define a separate stat similar to ftrace_bundle_tokenizer_errors in this file and log missing timestamp 

https://crsrc.org/c/third_party/perfetto/src/trace_processor/storage/stats.h

sashwinbalaji (approved):

src/trace_processor/importers/ftrace/ftrace_tokenizer.cc
rossning92: Thanks for the suggestion @sashwinbalaji ! 
I looked into this further. 1) It seems that the ClockTracker already records error stats internally (via RecordConversionError in ClockSynchronizer::ConvertSlowpath) when ToTraceTime fails, so the failure is already surfaced in the stats table (clock_sync_failure_unknown_target_clock).
2) I also took a look at other ToTraceTime fallback paths in this file that follow the same pattern of silently returning. 
So I'll keep it consistent and skip the extra stat here.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/slice_tracks.ts
```diff
@@ -248,6 +248,11 @@ export const SLICE_TRACK_SCHEMAS: ReadonlyArray<SliceTrackTypeSchema> = [
     topLevelGroup: 'HARDWARE',
     group: undefined,
   },
+  {
+    type: 'adreno_gpu_cmdbatch',
+    topLevelGroup: 'HARDWARE',
```
LalitMaganti: Would it not make more sense to put this in the GPU top level group?

ui/src/plugins/dev.perfetto.TraceProcessorTrack/slice_tracks.ts
```diff
@@ -248,6 +248,11 @@ export const SLICE_TRACK_SCHEMAS: ReadonlyArray<SliceTrackTypeSchema> = [
...
+    group: 'GPU Adreno',
```
LalitMaganti: You probablhy don't need a group given it's a single track. Feels unnecessary.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/slice_tracks.ts
rossning92: Done! simply reuse the GPU top level group.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/slice_tracks.ts
rossning92: Here I still kept a sub-group ('Adreno Cmdbatch') since a real trace will produce multiple tracks (one per GPU context):
Tested with latest change:
image

LalitMaganti (approved):

rossning92:
@sashwinbalaji could I get an extra approval with the last changes on reusing `topLevelGroup: 'GPU'` as suggested by Lalit. thanks for that!

Issue #5097, opened 2026-03-12
[GPU] Improve Perfetto support for General-Purpose GPU (GPGPU) compute workloads


dreveman:
## Overview

Modern GPUs are used far beyond traditional graphics rendering — powering scientific simulation, machine learning, image processing, physical modeling, and other massively parallel workloads through General-Purpose Computing on Graphics Processing Units (GPGPU). Perfetto's GPU tracing support was designed primarily around graphics pipelines (Vulkan, OpenGL, GLES), and while it can partially represent compute workloads today, several gaps in the data model, trace config, and UI make it less than ideal to analyze GPGPU workloads effectively.

This issue tracks a collection of related improvements spanning API coverage, counter collection controls, UI capabilities, and the render stage data model — with the goal of making Perfetto a first-class tracing backend for GPU compute workloads across CUDA, HIP, SYCL, and other GPGPU runtimes.

---

## Motivation

GPGPU profiling tools like CUPTI (CUDA), ROCProfiler/ROCTracer (HIP/ROCm), and Intel's GPU profiling stack already expose rich performance data. Producers such as PyTorch Profiler (via libkineto) and vendor profiling SDKs typically have some Perfetto support today but typically limited to using track events and debug annotations to record GPU events. This limits the UI plugins and tools that can be built on top of Perfetto traces for analyzing GPU workloads. The improvements outlined below describe how to address some of these gaps.

---

### 1. Missing API Enums for CUDA / HIP

The current `GpuRenderStageEvent` and related protos lack enum values for CUDA and HIP APIs. Producers targeting these runtimes are forced to use placeholder or vendor-specific values, making cross-tool interoperability fragile.

**Proposed:** Add first-class enum entries for `CUDA` and `HIP` (and potentially `SYCL`, `Metal Compute`, `OpenCL`) in the relevant GPU proto definitions.

---

### 2. Multi-GPU Support in the UI

Perfetto's UI does not clearly differentiate or aggregate events across multiple GPU devices. For multi-GPU compute jobs running across 4, 8, or more devices, the current UI falls apart.

**Proposed (v1):**
- Surface `gpu_id` as a first-class grouping dimension in the UI track hierarchy
- Allow users to view per-GPU timelines side-by-side or in a collapsible grouped view
- Ensure SQL queries and metrics surface `gpu_id` consistently

**Proposed (v2):**
- Deprecate `gpu_id` and introduce a `gpu_iid` and a GPU description proto message (InternedGpu)
[...]

PR #5523, opened 2026-04-21
tp: Add gpu_context table for InternedGraphicsContext

dreveman:
Add a new gpu_context table that maps GPU graphics context IDs to their associated process ID and API type (OPEN_GL, VULKAN, OPEN_CL, CUDA, HIP). This resolves the TODO in gpu_event_parser.cc to create a proper table for graphics context data.

Previously, the context field on GpuRenderStageEvent was used to look up InternedGraphicsContext but only the pid was extracted -- the api field was ignored and no structured table existed for context metadata. Now the full context metadata is materialized into a queryable table.

The table is populated lazily: each unique context_id is inserted on first encounter during GpuRenderStageEvent parsing.

The view is exposed via a new gpu stdlib module at gpu.context, accessible with INCLUDE PERFETTO MODULE gpu.context.

Schema:
  gpu_context(id, context_id, pid, api)

Example query:
  INCLUDE PERFETTO MODULE gpu.context;
  SELECT gc.api, gs.name, gs.dur
  FROM gpu_slice gs
  JOIN gpu_context gc ON gs.context_id = gc.context_id
  WHERE gc.api = 'CUDA'

Related: https://github.com/google/perfetto/issues/5097

### DIFF
--- a/src/trace_processor/importers/proto/gpu_event_parser.cc
+++ b/src/trace_processor/importers/proto/gpu_event_parser.cc
@@ -257,6 +257,24 @@ const char* MeasureUnitToString(int32_t unit) {
   }
 }
 
+const char* GraphicsApiToString(int32_t api) {
+  using Api = protos::pbzero::InternedGraphicsContext::Api;
+  switch (api) {
+    case Api::OPEN_GL:
+      return "OPEN_GL";
+    case Api::VULKAN:
+      return "VULKAN";
+    case Api::OPEN_CL:
+      return "OPEN_CL";
+    case Api::CUDA:
+      return "CUDA";
+    case Api::HIP:
+      return "HIP";
+    default:
+      return "UNDEFINED";
+  }
+}
+
 }  // namespace
 
 StringId GpuEventParser::FormatCounterUnit(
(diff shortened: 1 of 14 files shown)


src/trace_processor/importers/proto/gpu_event_parser.cc
```diff
@@ -692,6 +692,42 @@ void GpuEventParser::ParseGpuRenderStageEvent(
         protos::pbzero::InternedGraphicsContext>(context_id);
     if (decoder) {
       pid = decoder->pid();
+
+      // Insert into gpu_context table if not already present.
+      if (!gpu_contexts_inserted_.Find(context_id)) {
```
LalitMaganti: It would be nice if this was factored out a bit instead of being inline like this. Likely into an Intern methoid with early return etc.

src/trace_processor/importers/proto/gpu_event_parser.cc
```diff
@@ -692,6 +692,42 @@ void GpuEventParser::ParseGpuRenderStageEvent(
...
+        gpu_contexts_inserted_[context_id] = true;
+        tables::GpuContextTable::Row ctx_row;
+        ctx_row.context_id = static_cast<uint32_t>(context_id);
+        if (decoder->has_pid()) {
+          ctx_row.pid = static_cast<uint32_t>(decoder->pid());
+        }
+        if (decoder->has_api()) {
+          const char* api_str = nullptr;
+          switch (decoder->api()) {
```
LalitMaganti: Would be much cleaner as a pure function in anonymous namespace.

src/trace_processor/importers/proto/gpu_event_parser.cc
dreveman: Done

src/trace_processor/importers/proto/gpu_event_parser.cc
dreveman: Done

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
```diff
@@ -622,6 +622,21 @@ SELECT
   *
 FROM __intrinsic_vulkan_memory_allocations;
 
+-- GPU graphics context.
+CREATE PERFETTO VIEW gpu_context (
```
LalitMaganti: This should definitely *not* be in the prelude. We don't allow adding stuff here anymore (except very rarely). We should really figure out a place in the standard library where this can be exposed.

dreveman: Moved it to stdlib/gpu. Let me know if there's a better place to put it.

tools/check_sql_modules.py
LalitMaganti, line 48: Please follow the guidance here to put in the std toplevel package.

dreveman: Done. But I had to add "std" here instead, which I assume is ok?

LalitMaganti (approved):

PR #5525, opened 2026-04-21
tp: replace sqlfluff formatter with syntaqlite

LalitMaganti:
**Stack:**

1. #5614 — `tp: prep PerfettoSqlParser for a second backend`
2. #5615 — `tp: add syntaqlite-driven macro expansion behind a flag`
3. **#5525 — `tp: replace sqlfluff formatter with syntaqlite` (parallel; based on #5615)** ← you are here
4. #5616 — `tp: route PerfettoSqlParser through syntaqlite macro expansion`
5. #5472 — `tp: delete the legacy PerfettoSqlPreprocessor`

---

- Reformat the PerfettoSQL stdlib using the new formatter.
- Narrow the `perfetto_fmt_dialect` shared library's exported symbol
  surface: instead of dropping the project-wide `visibility_hidden`
  config (which exported every internal helper), define
  `SYNTAQLITE_SHARED` so only the 11 symbols tagged `SYNTAQLITE_API` /
  `SYNTAQLITE_DIALECT_API` (the CLI `dlsym` entry point plus parser/
  tokenizer dispatch) are exported.


python/tools/code_format_sql.py
```diff
@@ -26,33 +27,89 @@
...
+  # GN's shared_library emits `libperfetto_fmt_dialect.so` on Linux/macOS and
+  # `perfetto_fmt_dialect.dll` on Windows.
+  if platform.system() == 'Windows':
+    return os.path.join(out_dir, f'{DIALECT_TARGET}.dll')
+  return os.path.join(out_dir, f'lib{DIALECT_TARGET}.so')
+
+
+def _resolve_out_dir(repo_root: str) -> str:
+  env_out = os.environ.get('OUT')
+  if env_out:
```
aMayzner, line 50: Why is it optional? Wouldn't it be better to always specify the OUT dir rather than the most recent one?

aMayzner (approved):

PR #5526, opened 2026-04-21
traced_perf - support setting period/frequency for follower events.

abhmen01:
This change adds support for configuring the sampling period or frequency for follower  events.

Previously, only the group leader event could be configured to trigger samples. With this change, follower events can also independently trigger samples by specifying the 'period' or 'frequency' field in the config.

This aligns Perfetto's behavior more closely with `perf` semantics, e.g., `perf record -e '{context-switches,cpu-clock/period=<SAMPLE_PERIOD>}'`, and allows more flexible PMU event sampling.

### DIFF
--- a/src/profiling/perf/event_reader.cc
+++ b/src/profiling/perf/event_reader.cc
@@ -84,6 +84,14 @@ bool MaybeApplyTracepointFilter(int fd, const PerfCounter& event) {
   return true;
 }
 
+std::optional<uint64_t> GetPerfEventId(int fd) {
+  uint64_t id = 0;
+  if (ioctl(fd, PERF_EVENT_IOC_ID, &id) != 0) {
+    return std::nullopt;
+  }
+  return id;
+}
+
 }  // namespace
 
 PerfRingBuffer::PerfRingBuffer(PerfRingBuffer&& other) noexcept
@@ -212,11 +220,15 @@ EventReader::EventReader(uint32_t cpu,
                          perf_event_attr event_attr,
                          base::ScopedFile perf_fd,
                          std::vector<base::ScopedFile> followers_fds,
+                         uint64_t timebase_stream_id,
+                         std::vector<uint64_t> follower_stream_ids,
                          std::optional<PerfRingBuffer> ring_buffer)
     : cpu_(cpu),
       event_attr_(event_attr),
       perf_fd_(std::move(perf_fd)),
       follower_fds_(std::move(followers_fds)),
+      timebase_stream_id_(timebase_stream_id),
+      follower_stream_ids_(std::move(follower_stream_ids)),
       ring_buffer_(std::move(ring_buffer)) {}
 
 EventReader& EventReader::operator=(EventReader&& other) noexcept {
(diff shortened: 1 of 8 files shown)


src/profiling/perf/event_reader.cc
```diff
@@ -272,6 +272,12 @@ std::optional<EventReader> EventReader::ConfigureEvents(
       return std::nullopt;
     }
   }
+
+  // redirect the follower events to the leader's buffer.
+  for (size_t i = 0; i < follower_fds.size(); ++i) {
+    ioctl(follower_fds[i].get(), PERF_EVENT_IOC_SET_OUTPUT, timebase_fd.get());
```
rsavitski, line 305: This seems very incomplete. It redirects events, but the patch doesn't have any code for adjusting how events are read when using either the sampling or the polling interface.

I'd expect a lot more code for any buffer sharing. In particular extra IDs in the samples (see PERF_SAMPLE_STREAM_ID) and demultiplexing them on the reading side.

What kind of config were you testing with?

rsavitski (changes requested):
I feel like this patch is motivated due to a misunderstanding of the current config, and lacks a big chunk of implementation of the buffer sharing part.

Currently, each "perf.linux" data source is one recording group that uses leader sampling (https://man7.org/linux/man-pages/man1/perf-list.1.html#LEADER_SAMPLING). So only the leader sampling is by design.

If you need to independently record several counters, you can specify several "linux.perf" data sources in your config. That's the equivalent of "perf record -e A,B,C".

abhmen01:
The current behavior is indeed leader sampling, and I agree that this is how perfetto works today.
The intent of this patch is to extend that behavior, not to make events independent. In particular, this is not trying to model `perf record -e A,B`, where each event is recorded separately. The goal is to keep the events in a group and read them together, but allow a follower event to trigger sampling when its own period overflows.

So today the behavior is effectively equivalent to: `perf record -e '{A,B}':S` 
where only the leader can trigger samples, I presume. 

With this patch, if a period is specified for follower `B`, the intended behavior becomes: `perf record -e '{A,B/period=100/}'`

That is, the group is still read as a unit regardless of which event triggered the sample, but sampling is no longer restricted to the leader only. So I think the key difference is:

current behavior: grouped read, leader-only sample trigger
intended behavior with this patch: grouped read, leader or follower can trigger if that follower has its own sampling period configured.

src/profiling/perf/event_reader.cc
abhmen01: For the grouped case, I did not think extra demultiplexing was needed, because we always use PERF_FORMAT_GROUP, so the whole group is read together regardless of which event triggered the overflow. In that path, PERF_SAMPLE_STREAM_ID should not be required for decoding, which is why I did not add it.

On further review, I think the non-grouped case still needs to be handled explicitly, even though PERF_FORMAT_GROUP is always enabled during configuration. There we would need to maintain a mapping from event index to stream ID and use that during decoding.

abhmen01: Modified the events reading code to use the stream id to recognize the event which produced the sample.

abhmen01:
@rsavitski 
A concrete use case is periodic IPC measurement for a process. For example:

`perf record -e '{sched:sched_switch/period=1/,cpu-cycles/period=10000/,inst_retired}'`

This keeps the given events in one group so they are read together as a consistent snapshot. `sched:sched_switch/period=1/ `can trigger a sample on every context switch, while `cpu-cycles/period=10000/` can also trigger a sample every 10,000 cycles. At each sample point, both the cycle count and retired-instruction count are captured together, which allows IPC to be computed periodically. 

The equivalent perfetto configuration will be:
```
data_sources: {
    config {
        name: "linux.perf"
        perf_event_config {
            timebase {
                period: 1
                tracepoint {
                    name: "sched/sched_switch"
                }
                name: "sched_switch"
            }
            followers: {
                period: 10000
                raw_event {
                    type: 8
                    config: 17
                }
                name: "CPU_CYCLES"
            }
            followers: {
                raw_event {
                    type: 8
                    config: 8
                }
                name: "INST_RETIRED"
            }
            ignore_open_failure: true

        }
    }
}
```

LalitMaganti:
I think this is one of the things which *could* be nice but seems like quite a bit of subtle code to reason about. Are we sure there's no way to do this with just specifying multiple linux.perf data sources in a single config?

abhmen01:
@LalitMaganti 
Multiple linux.perf data sources would create independent groups, not one grouped read. That is fine for recording counters separately, but not equivalent to what this patch is trying to do.
The intent here is to keep a single perf group, read all events together as one snapshot, and allow a follower overflow to trigger that group read. Multiple data sources do not preserve that behavior.

LalitMaganti:
Sure. I think the thing we have to weigh is whether the cost of adding something which we would not use in the forseeable *and* is so niche is worth it. Right now, the team is already under a bit of short-staffing pressure give some engineers who have moved on from the team so I'm not sure if a subtle and complex patch like this is really something we want to take on right now especially if there are teething issues for this going forward as well.

We might need to park this for now and revisit at some point in the future.

PR #5527, opened 2026-04-22
traced: Handle sys_stats pread() returning fewer bytes

wusamuel6:
Handle pread() for sys_stats returning fewer bytes than expected.

Bug: 503864690


src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -749,15 +749,21 @@ void SysStatsDataSource::Flush(FlushRequestID, std::function<void()> callback) {
 size_t SysStatsDataSource::ReadFile(base::ScopedFile* fd, const char* path) {
   if (!*fd)
     return 0;
-  ssize_t res = pread(**fd, read_buf_.Get(), kReadBufSize - 1, 0);
-  if (res <= 0) {
+  char* buf = static_cast<char*>(read_buf_.Get());
+  size_t total_read = 0;
+  ssize_t res;
+  while (total_read < kReadBufSize - 1 &&
```
LalitMaganti: Please break this down into multiplelines rarher cramming everything like this. This is quite unreadable.

src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -749,15 +749,26 @@ void SysStatsDataSource::Flush(FlushRequestID, std::function<void()> callback) {
...
-  ssize_t res = pread(**fd, read_buf_.Get(), kReadBufSize - 1, 0);
-  if (res <= 0) {
-    PERFETTO_PLOG("Failed reading %s", path);
+  char* buf = static_cast<char*>(read_buf_.Get());
+  size_t total_read = 0;
+  ssize_t res = 0;
+  while (total_read < kReadBufSize - 1) {
+    // pread() is not guaranteed to return the same number of bytes requested,
+    // so continuouslly call pread() until it returns 0, indicating EOF.
+    size_t remainingBufSize = kReadBufSize - 1 - total_read;
```
LalitMaganti: nit: we use camel_case not pascalCase

src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -749,15 +749,26 @@ void SysStatsDataSource::Flush(FlushRequestID, std::function<void()> callback) {
...
+    res = pread(**fd, buf + total_read, remainingBufSize,
```
LalitMaganti: I was suspicious of this so I asked AI and it said that this is not really valid to do this (which also I tend to agreee with myself):

> This loop changes ReadFile() from a single pread() snapshot into multiple independent pread() calls with increasing offsets. That is risky for the /proc/* inputs this helper serves (/proc/stat, /proc/vmstat, /proc/meminfo, /proc/diskstats, /proc/pressure/*): Linux seq_file documentation describes a read “session” as roughly one read() call, so stitching together multiple calls is not guaranteed to represent one coherent snapshot. In practice that can produce internally inconsistent samples, e.g. fields from /proc/stat or /proc/diskstats coming from different moments in time inside one emitted SysStats packet. The old code could truncate, but it at least never merged multiple procfs snapshots into one record.

LalitMaganti: Orthogonally, if this is OK to do, I also think we need to wrap this PERFETTO_EINTR which was missing.

wusamuel6: > Orthogonally, if this is OK to do, I also think we need to wrap this PERFETTO_EINTR which was missing.

Agree, ACK

Regarding single vs multiple `pread()`, I think there are 2 issues.

1. The Linux seq_file framework calls seq_start/seq_stop per page, so for procfs files larger than a page, one definitely needs multiple `pread()` calls to capture the entire file. Actually this is not obvious but would lead to truncated output like [this](https://lore.kernel.org/all/20251204000348.1413593-1-tjmercier@google.com/). 
2. Multiple `pread()`s guarantee atomicity across pages, but not atomicity for the entire file. So feedback above was inclined to only have a single `pread()` call.

I think solutions to add atomicity across the entire file while reading a multi-page file would require some nontrivial kernel change. This was good point to bring up, but AFAICT handling 1 and handling 2 are mutually exclusive, in which case it's probably better to choose 1 right?

LalitMaganti: So all that said is it actually safe to do what you're doing here, even on older kernel versions? I cannot quite parse that out of your message.

Also @primiano FYI.

primiano: we once looked it up with @rsavitski 
IIRC pread on seq files has some complicated logic so that if two threads pread concurrently it rewinds and regenerates the file to maintain the idempotency.

The real question I have here is: do we really need the pread?
For sure we don't do concurrent reads (also we are single threaded)
I think the only reason I did it back then was to avoid an extra lseek syscall to reset the pointer 

But reached this point I think it would be better to just use a loop with a standard read (without _p_) and a lseek at the beginning, rather than all this. 
I think it will be cheaper if we end up hitting page boundaries, as otherwise the kernel needs to emulate it.

THe other comment i have is: in any case this must be behind a flag (either perfetto_flags.aconfig or a field in the data source config for this data source, which we deprecate after it sticks). this is subtle enough that we might be getting it wrong, and if we do it will ruin tracing for everybody

wusamuel6: > So all that said is it actually safe to do what you're doing here, even on older kernel versions?

Yes, ultimately the intent is to adhere to the semantics of `read()/pread()` syscall wherein it's valid for them to return less bytes than requested. I believe this behavior predates Linux (since it's a POSIX spec), so this multi-read thing is backwards compatible.

> The real question I have here is: do we really need the pread?

Agree- if there are no concurrency issues, `read()` is perfectly fine. `read()` also seem to be more performant in the case where we need to read multiple pages.

> in any case this must be behind a flag

Added to perfetto_flags.aconfig

perfetto_flags.aconfig
```diff
@@ -56,3 +56,10 @@ flag {
   bug: "494498282"
   is_fixed_read_only: true
 }
+flag {
+  name: "sys_stats_large_read"
+  namespace: "kernel_memory_management"
```
primiano: can you please keep it into namespace: perfetto, so they are all listedin the same gantry section when I search. Otherwise if something goes wrong we'll ahve to remember that it's in some other namespace

src/traced/probes/sys_stats/sys_stats_data_source.cc
```diff
@@ -749,6 +750,36 @@ void SysStatsDataSource::Flush(FlushRequestID, std::function<void()> callback) {
 size_t SysStatsDataSource::ReadFile(base::ScopedFile* fd, const char* path) {
   if (!*fd)
     return 0;
+
+  if (PERFETTO_FLAGS_SYS_STATS_LARGE_READ) {
```
primiano: `if constextpr (PERFETTO...) {`

this suppresses the compiler warning about dead code below (otherwise the compiler sees if(True) and says the code below is unreachable)

primiano (approved):
LGTM % 1 comment on the namespace

LalitMaganti (approved):

PR #5531, opened 2026-04-22
ui: MenuTitle - use children instead of label attr

stevegolton:
MenuTitle - Use children to inject the title content instead of the label attr.

Before:
```ts
m(MenuTitle, {label: 'Foo'})
```

After:
```ts
m(MenuTitle, 'Foo')
```


aMayzner (approved):

stevegolton:
Closing due to consistency with `MenuItem`.

`MenuItem` cannot use children for its label content as children are used for sub menu items.

```ts
m(MenuItem, {...},
  m(MenuItem), // <-- appears in a sub-menu
),
```

PR #5539, opened 2026-04-22
tp: Add gpu_api arg for tagging track events with GPU API type

dreveman:
Add a GpuApi enum to gpu_track_event.proto that allows producers to
tag host-side API call slices with the GPU API they represent (e.g.
CUDA, Vulkan, OpenGL).

The gpu_api extension field (3001) on GpuTrackEvent is automatically
parsed into the args table by the trace processor's proto reflection
system via the registered gpu_track_event.descriptor. No custom
importer code is needed.

Example query:
  SELECT s.name
  FROM args a
  JOIN slice s ON s.arg_set_id = a.arg_set_id
  WHERE a.key = 'gpu_api' AND a.string_value = 'GPU_API_CUDA'

Related: https://github.com/google/perfetto/issues/5097

### DIFF
--- a/protos/perfetto/trace/gpu/gpu_track_event.proto
+++ b/protos/perfetto/trace/gpu/gpu_track_event.proto
@@ -18,6 +18,18 @@ package perfetto.protos;
 
 import public "protos/perfetto/trace/track_event/track_event.proto";
 
+// GPU API type for tagging track events.
+// Values intentionally match InternedGraphicsContext.Api in
+// gpu_render_stage_event.proto.
+enum GpuApi {
+  GPU_API_UNDEFINED = 0;
+  GPU_API_OPEN_GL = 1;
+  GPU_API_VULKAN = 2;
+  GPU_API_OPEN_CL = 3;
+  GPU_API_CUDA = 4;
+  GPU_API_HIP = 5;
+}
+
 // Correlates a track event with GPU render stage events.
 message GpuCorrelation {
   // Event IDs of GpuRenderStageEvents that this track event submitted.
@@ -32,5 +44,8 @@ message GpuTrackEvent {
   extend TrackEvent {
     // Correlates this track event with GPU render stage events.
     optional GpuCorrelation gpu_correlation = 3000;
+
+    // GPU API that this track event represents a call to.
+    optional GpuApi gpu_api = 3001;
   }
 }
--- a/test/trace_processor/diff_tests/parser/graphics/gpu_api_slice.textproto
+++ b/test/trace_processor/diff_tests/parser/graphics/gpu_api_slice.textproto
@@ -1,0 +1,62 @@
+# Test gpu_api_call table populated from GpuTrackEvent.gpu_api extension.
+
+# Interned data for track event sequences.
+packet {
+  trusted_uid: 9999
+  trusted_packet_sequence_id: 1
+  timestamp: 0
+  interned_data {
+    event_categories { iid: 1 name: "gpu" }
+    event_names { iid: 1 name: "cuLaunchKernel" }
+    event_names { iid: 2 name: "vkCmdDispatch" }
+  }
+  sequence_flags: 1  # SEQ_INCREMENTAL_STATE_CLEARED
+}
+
+# CUDA API call slice.
+packet {
+  trusted_uid: 9999
+  trusted_packet_sequence_id: 1
+  timestamp: 100
+  track_event {
+    type: TYPE_SLICE_BEGIN
+    category_iids: 1
+    name_iid: 1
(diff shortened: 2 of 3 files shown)


src/trace_processor/importers/common/gpu_tracker.cc
```diff
@@ -20,12 +20,35 @@
...
+#include "src/trace_processor/tables/gpu_tables_py.h"
 #include "src/trace_processor/tables/metadata_tables_py.h"
 
 namespace perfetto::trace_processor {
 
+namespace {
+
+const char* GpuApiToString(int32_t api) {
+  switch (api) {
+    case 1:
```
LalitMaganti: OPlease use the proto enum values instead of hardcoding like this.

Also UI think you can use the EnumToString helper function autogenerated by pbzero.

src/trace_processor/perfetto_sql/stdlib/std/gpu/api_slice.sql
```diff
@@ -0,0 +1,31 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- Maps track event slices to the GPU API they target.
+--
+-- Each row links a host-side API call slice to the GPU API it
+-- targets, populated from the gpu_api extension field on
+-- GpuTrackEvent.
+CREATE PERFETTO VIEW gpu_api_slice (
```
LalitMaganti: I think the naming of this table is non-ideal.

dreveman:
@LalitMaganti It feels a bit overkill to me with this table when api is the only thing we have in it but it's useful when I want to identify all host track slices that represent some GPU API. I can't rely on the correlation ids here as there are important API calls that doesn't have GPU side work. E.g. GPU memory allocation. The hacky workaround I've used in the past is to assume a specific track event category, e.g. "cuda" but that always felt too fragile.

LalitMaganti:
Can you not make it an argument and add to args table instead?

dreveman:
> Can you not make it an argument and add to args table instead?

@LalitMaganti Latest version is putting this in the args table instead. I guess it's questionable if we should have GpuTrackEvent.gpu_api at all or just rely on a normal track event arg with specific key/value for this. I guess this makes it more official and a bit more efficient storage wise but I'm fine just abandoning this if you prefer?

LalitMaganti:
> @LalitMaganti Latest version is putting this in the args table instead. I guess it's questionable if we should have GpuTrackEvent.gpu_api at all or just rely on a normal track event arg with specific key/value for this. I guess this makes it more official and a bit more efficient storage wise but I'm fine just abandoning this if you prefer?

Up to you on this. You should see https://perfetto.dev/docs/instrumentation/extensions though I landed this just this morning. Basically you don't need to add code to trace processor to have it understand extensions - you can just emit the descriptor into the trace, or, alternatrively similar to Chrome track event, you can compile the extension int oa descriptor file and have it automatically show up in the args table.

I am thinking the latter might be what you want here (also makes it more future proof if you continue on adding things to this proto).

src/trace_processor/importers/common/gpu_tracker.cc
dreveman: Removed the string usage from latest version

src/trace_processor/perfetto_sql/stdlib/std/gpu/api_slice.sql
dreveman: Also gone from latest version of change

dreveman:
> > @LalitMaganti Latest version is putting this in the args table instead. I guess it's questionable if we should have GpuTrackEvent.gpu_api at all or just rely on a normal track event arg with specific key/value for this. I guess this makes it more official and a bit more efficient storage wise but I'm fine just abandoning this if you prefer?
> 
> Up to you on this. You should see https://perfetto.dev/docs/instrumentation/extensions though I landed this just this morning. Basically you don't need to add code to trace processor to have it understand extensions - you can just emit the descriptor into the trace, or, alternatrively similar to Chrome track event, you can compile the extension int oa descriptor file and have it automatically show up in the args table.
> 
> I am thinking the latter might be what you want here (also makes it more future proof if you continue on adding things to this proto).

Ok, that seems good here. I'll take a look at how the Chrome track event stuff works and update this change accordingly.

dreveman:
@LalitMaganti I just dropped the tp changes from this commit and rely on reflection system. Makes it very simple but nice that it's part of the GpuTrackEvent extension. ptal.

LalitMaganti (approved):

PR #5541, opened 2026-04-22
Allow trace_processor extract screenshot track events.

m--koma:
A follow on https://github.com/google/perfetto/pull/5348, making screenshot data available via trace processor.

Internal tracker: b/498545104.


src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -1248,6 +1249,33 @@ class TrackEventEventImporter {
     return base::OkStatus();
   }
 
+  base::Status ParseScreenshotArgs(ConstBytes screenshot_bytes,
+                                   BoundInserter* inserter) {
+    protos::pbzero::Screenshot::Decoder screenshot(screenshot_bytes);
+    if (screenshot.has_jpg_image()) {
```
LalitMaganti, line 1256: We'll need to base64 encode screenshot content. We cannot store it raw in the string pool, everything will break silently if there's an embedded null.

m--koma: Thank you! Great point, done.

LalitMaganti (approved):

src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -1248,6 +1250,33 @@ class TrackEventEventImporter {
...
```
sashwinbalaji, line 1256: LGTM. Sorry, the below is just for my understanding.

Can we have all 3 image types in one msg ? Like why separate if rather than else if.

I did check https://crsrc.org/c/third_party/perfetto/protos/perfetto/trace/track_event/screenshot.proto and it seems like it's possible but just want to clarify if that's the case or not.

sashwinbalaji (approved):

src/trace_processor/importers/proto/track_event_event_importer.h
m--koma: Thank you! The primary reason is that Perfetto UI already supports screenshot.jpg_image, and splitting them up by type makes sure we don't break it by adding .ppm and .pam.

LalitMaganti: I mean you could also just change the UI code at the same time of that was a concern?

m--koma: Do you mean that we should've made it a single field in https://github.com/google/perfetto/pull/5348? How do we distinguish the image format in this case?

LalitMaganti: I agree that we want to make it diffierent args for the type. That's a convincing argument. I just think using "but we have to change the UI" is a not a good reason for not doing a right thing.

I agree that what is implemented in this PR is the right thing to do.

m--koma: Ah, sorry, I misinterpreted the comment. Agreed.

m--koma:
@Owners, thank you for the review! I don't appear to have write access to the repository - could you merge this PR when you get a chance?

PR #5542, opened 2026-04-22
tp: emit stdlib SQL bundle as uint8_t arrays

LalitMaganti:
The amalgamated SQL bundle for stdlib and metrics previously emitted
each module as a raw string literal split into ~1000-char chunks.
For very large SQL modules the resulting chain of concatenated string
literals pushes C++ compilers to their limits.

Switch to a plain uint8_t byte array initializer, which compilers
handle reliably regardless of size. The FileToSql struct now carries
a pointer + size, plus an inline sql_view() accessor so callers get
a std::string_view without having to spell out the reinterpret_cast.
The kFileToSql table stays constant-initialized: a const char* member
would require reinterpret_cast in the initializer, which is not a
constant expression and would demote the table to dynamic init.

### DIFF
--- a/src/trace_processor/trace_processor_impl.cc
+++ b/src/trace_processor/trace_processor_impl.cc
@@ -424,7 +424,7 @@ sql_modules::NameToPackage GetStdlibPackages() {
     std::string module_name = sql_modules::GetIncludeKey(file_to_sql.path);
     std::string package_name = sql_modules::GetPackageName(module_name);
     packages.Insert(package_name, {})
-        .first->push_back({module_name, file_to_sql.sql});
+        .first->emplace_back(module_name, file_to_sql.sql_view());
   }
   return packages;
 }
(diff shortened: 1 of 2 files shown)


src/trace_processor/trace_processor_impl.cc
```diff
@@ -424,7 +424,7 @@ sql_modules::NameToPackage GetStdlibPackages() {
     std::string module_name = sql_modules::GetIncludeKey(file_to_sql.path);
     std::string package_name = sql_modules::GetPackageName(module_name);
     packages.Insert(package_name, {})
-        .first->push_back({module_name, file_to_sql.sql});
+        .first->push_back({module_name, std::string(file_to_sql.sql_view())});
```
sashwinbalaji: nit: emplace_back and remove std::string cast

sashwinbalaji (approved):

Issue #5498, opened 2026-04-20
Coarse CPU usage counter


belregionsoft:
Hello! I need help understanding the lack of CPU usage monitoring data in my trace. I'm using the web tool Perfetto to record a trace of an Android 11 Go device with the Coarse CPU usage counter switch enabled. All other CPU metrics are recorded, except for CPU usage.

LalitMaganti:
CPU usage polling relies on the `/proc/stat` kernel interface being enabled on device. Fundamentally, the WebUI sets these parts of the config: https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/protos/perfetto/config/sys_stats/sys_stats_config.proto;l=48?q=protos%2Fperfetto%2Fconfig%2Fsys_stats%2Fsys_stats_config.proto&ss=chromium

We set that in the UI at https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/ui/src/plugins/dev.perfetto.RecordTraceV2/pages/cpu.ts;l=47?q=statperiodms%20f:perfetto%20f:ui&ss=chromium

There's few possibilities for why this is not working:
1) /proc/stat is not accessible to Perfetto for some reason (e.g. OEM modification, permissions etc)
2) Android 11 at this point is quite old so there might be some historical bug which has long since been fixed in that version
3) There's something specifcally wrong with the import of these events due to some change in how they were represented

It would be useful if you could help narrow down this:
1) Provide us access to the trace file you're seeing so we can investigate if the data is in the trace or not
2) Look at logcat to see if there are any error messages with the "perfetto:" tag
3) Poke at /proc/stat on device to see if it's returning sensible results
4) Record the trace with the command line to see if it might be a web-only issue (seems unlikely but possible)

belregionsoft:
[android-202633-92047.zip](https://github.com/user-attachments/files/26959885/android-202633-92047.zip)

LalitMaganti:
I see the counters just fine in the trace you linked:

<img width="3350" height="1860" alt="Image" src="https://github.com/user-attachments/assets/c7589209-8aa9-4a98-9725-c6d378df6d59" />

This maps precisely to what is available from us with Linux interface of /proc/stat (https://man7.org/linux/man-pages/man5/proc_stat.5.html)

[...]

PR #5544, opened 2026-04-22
ui: add CoarseCpu plugin for per-CPU /proc/stat counters

LalitMaganti:
The cpustat tracks parsed from /proc/stat (cpu.times.user_ns,
cpu.times.system_mode_ns, etc.) were previously all piled under a
generic "CPU Stat" group with duplicate-looking names — one track per
(CPU, metric) with the same name across CPUs, leaving users unable to
tell which CPU each track belonged to.

Introduce a dedicated dev.perfetto.CoarseCpu plugin (enabled by
default) that pivots these tracks by metric under the standard CPU
group — "User Time", "Kernel Time", "Idle Time", "IO Wait Time",
"Hardware IRQ Time", "Soft IRQ Time", "Steal Time", "Nice Time" —
with per-CPU tracks named e.g. "User Time (CPU 0)". Tracks render
in delta mode (more intuitive than the raw monotonic counter) and
carry descriptions sourced from proc_stat(5).

Also drop the redundant "CPU " prefix from other group names already
nested under the CPU standard group (Max Frequency, Capacity, etc.).

Bug: https://github.com/google/perfetto/issues/5498


ui/src/plugins/dev.perfetto.CoarseCpu/index.ts
```diff
@@ -0,0 +1,191 @@
...
+      .getPlugin(StandardGroupsPlugin)
+      .getOrCreateStandardGroup(ctx.defaultWorkspace, 'CPU');
+
+    const metricGroups = new Map<string, TrackNode>();
+
+    for (; it.valid(); it.next()) {
+      const {trackId, cpu, metric, machineId, ucpu} = it;
+
+      const info = METRICS[metric];
+      if (info === undefined) continue;
```
sashwinbalaji: nit: not sure of the UI convention, but can we have a warn for this ?

sashwinbalaji (approved):

PR #5545, opened 2026-04-22
Capture ftrace events in traced perf

abhmen01:
PMU events captured through perf are not reliably aligned with the Linux
system clocks used by ftrace. This is especially visible at the start of a
trace, where PMU samples can appear before scheduler activity such as
sched_switch. The mismatch makes it harder to correlate hardware counters
with context switches and can introduce inaccuracies in Perfetto timelines.

Fix this by allowing traced_perf to capture ftrace events through the perf
subsystem instead of the direct ftrace endpoint. In this mode, scheduler
tracepoints and PMU events are all opened through perf_event_open and share
the same perf timebase, while tracepoint samples are still emitted as
standard ftrace events in the trace. This significantly improves timestamp
alignment between scheduling activity and PMU data.

This change supports two configurations under the linux.perf data source:

1. Using ftrace_config in linux.perf data source.
   traced_perf translates the requested ftrace events into perf tracepoint
   events. Only the ftrace_events and has_comp_sched fields are honored;
   other fields in the ftrace_config are ignored.

2. Using new field - perf_event_config.ftrace_events
   Tracepoints configured in perf_event_config are already captured through
   perf, but their raw sample data is not captured. With this change,
   raw tracepoint data can also be captured and emitted as
   ftrace events in the trace, allowing tracepoints and PMU events to be
   collected together in the same data source with aligned timestamps.

PerfEventConfig.ftrace_events has two optional settings.
- read_counter (default: true): read the perf counter value together with
  the raw ftrace sample data.
- compact_sched (default: true): use compact encoding for scheduler events
  such as sched_switch and sched_wakeup.

With this change, Perfetto produces perf counter samples for all configured
events and ftrace events for all configured tracepoints from the same perf
capture path.

In the two commits, already a seperate pull request #5526  has been created for [traced_perf - support setting period/frequency for follower events.](https://github.com/google/perfetto/commit/9e2f449e9370ec1ec121bda67a7bf80e73662171). It is included here only to meet the dependency.


primiano:
i'm sorry this is a quite big architectural change for the project if we start capturing ftrace events via traced_perf

why can you not chagne the clock in ftrace and capture that way?

there must be a better way. sorry this creates too much chaos and mainteinance cost for the project this way. having two ways to capture the same data ... just to work around some clock issue... can't fly

abhmen01:
I agree that a generic “capture ftrace via traced_perf” path would be a significant architectural change, and I understand the concern about adding a second way to capture the same data.

On the clock question: using the same clock in ftrace does not solve the alignment problem, because ftrace and perf still use different capture paths. In practice, the timestamp recorded for the same event occurrence can differ between the two paths even when the clock source is the same. I ran experiments on multiple devices to verify this, and the misalignment is still observable.

That said, I think there are really two different pieces in my proposal:

1. `perf_event_config.ftrace_events`
   This is the narrower change. When a tracepoint is already configured in `perf_event_config`, traced_perf is already capturing it through perf. The proposal here is only to also decode the raw tracepoint payload and emit it as an ftrace event in the trace.

2. `linux.perf { ftrace_config { ... } }`
   I agree this is the broader architectural step, because it effectively introduces a new entry point for ftrace-style capture on top of traced_perf.

The use case we are most interested in is the first one. In particular, we want to support cases where `sched:sched_switch` is the leader and a PMU event is configured as a follower. If we can emit the `sched_switch` tracepoint as an ftrace event from the same perf capture path, it becomes much easier to visualize and correlate the PMU data with scheduling activity.

For example:

data_sources: {
  config {
    name: "linux.perf"
    perf_event_config {
      cpu: 8
      timebase {
        period: 1
        tracepoint {
          name: "sched:sched_switch"
        }
        name: "SCHED_SWITCHES"
      }
      followers: {
        counter: INSTR_PER_CYCLE
        name: "INSTR_PER_CYCLE"
      }
    }
  }
}

Would you be open to the narrower `perf_event_config` approach on its own, even if we drop or defer the `ftrace_config` entry point? Or can you think of any other solution to support this use case?

LalitMaganti:
Personally, I think even 1) is problematic. I don't think this is a good fit for the direction of the project, we don't want to add *yet another* source of ftrace events and people expect us to maintain it. It's a very core feature and there is high likliehood that there is some delta between how tracepoints works through ftrace API and how they work through the perf API. And I don't want to have to deal with these issues.

If we really want to "align data between perf and ftrace", we're more thinking about leverage thigns like eBPF as a central tracing mechanism and it's something in our longer term (O(years)) plans. But for now, I don't think we want to support this upstream.

PR #5551, opened 2026-04-22
Make trace redactor run at low priority

edgararriagag:
Trace redactor has been designed so far to be ran at low priority, so we should assume this as the default, there is currently no need to run it at a higher priority so we take a global decision to lower its priority during startup, but this could change if requirements change for redactor in the future.


src/trace_redaction/main.cc
```diff
@@ -26,6 +27,12 @@ namespace perfetto::trace_redaction {
 static base::Status Main(std::string_view input,
                          std::string_view output,
                          std::string_view package_name) {
+  // Redaction is a low priority task as users are asynchronously informed
+  // when redaction finishes so we can keep it as a lower priority for now, if
+  // this requirement ever changes, then consider modifying redactor to let
+  // caller specify process level thread priority.
+  setpriority(PRIO_PROCESS, 0, 19);
+
```
LalitMaganti, line 44: Can you please guard this with PERFETTO_BUILD_WITH_ANDROID like other parts of the codebase please? Unconditionally enabling this will be annoying for host tools which don't have pemrissions to do this and also standalone Android where I don't know nif we're have perms from shell.

edgararriagag: Done

src/trace_redaction/main.cc
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include <sys/resource.h>
```
LalitMaganti: This should be gated behind the same.

src/trace_redaction/main.cc
```diff
@@ -26,6 +28,14 @@ namespace perfetto::trace_redaction {
 static base::Status Main(std::string_view input,
                          std::string_view output,
                          std::string_view package_name) {
+  // Redaction is a low priority task as users are asynchronously informed
+  // when redaction finishes so we can keep it as a lower priority for now, if
+  // this requirement ever changes, then consider modifying redactor to let
+  // caller specify process level thread priority.
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
```
LalitMaganti: RThis is OS_ANDROID *and* OS_BUILD_WTIH_ANDROID

src/trace_redaction/main.cc
edgararriagag: Done

src/trace_redaction/main.cc
edgararriagag: Done

LalitMaganti (approved):

primiano:
To be honest it's a bit unusual that a utility process by itself changes its priority.
That decision typically belongs to the caller.

IMHO:

1) Why isn't the caller (the code that creates the subprocess for this) do this
2) If for whatever reason the caller cannot, make it a cmdline argument 

In any case, I think this change does only in part what you want.

As per our meeting, I think what you really want here is to set the CPU affinity to little/efficiency cores.
BUt you can't just to a sched_setaffinity here because you don't know which cores are the little/efficient ones. Only the BSP knows. There must be some API somewhere to pin a process to efficiency cores and even cap its frequency.

THe other thing that is going to happen here is that because this thing does a lot of computation, it's going to bump up the scaling governor. If this is in a cgroup with clamping it could prevent this, as it's not interactive-critical.

But then again, there must be some Android way to put the process in the right cgroup  .

In summary: I'm not opposed to this (if you put behind a cmdline flag, or keep it os_android only). but also I don't think it does what you really want.

edgararriagag:
Replying to Primiano's concern, we are also working on the CPU set in parallel on the framework side of things.

So you are correct, this is only part of the overall change, this is the priority change, the rest is happening in the framework. Also a bit of background, the priority I would be happy to have it in the framework entirely handled, however, we had some limitation where there was the need to perform reflection because the OpenJDK used by system_server does not expose the PID and the deadline is too tight, so we had to do it in the Native process because the only available way found was via reflection in system_server (process that runs the redactor) which is prohibited in a mainline module.

This kind also got mixed with a tight deadline to land this and some other changes for Friday to get this into our mainline. That being said, I do agree that its better to add the plumbing so this goes through a cmdline argument so I went ahead and implemented as you suggested.

src/trace_redaction/main.cc
```diff
@@ -35,21 +52,54 @@ static base::Status Main(std::string_view input,
...
   constexpr int kSuccess = 0;
   constexpr int kFailure = 1;
   constexpr int kInvalidArgs = 2;
 
-  if (argc != 4) {
-    PERFETTO_ELOG(
-        "Invalid arguments: %s <input file> <output file> <package name>",
-        argv[0]);
+  std::vector<const char*> positional_args;
+  int nice = 0;
```
LalitMaganti: Shouldn't this be std::optional so if the argument isn't specified you don't do anything?

src/trace_redaction/main.cc
```diff
@@ -14,18 +14,35 @@
...
 
 // Builds and runs a trace redactor.
 static base::Status Main(std::string_view input,
                          std::string_view output,
-                         std::string_view package_name) {
+                         std::string_view package_name,
+                         int nice) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID) && \
+    PERFETTO_BUILDFLAG(PERFETTO_ANDROID_BUILD)
+  // This will only occur for builds from android repository.
```
LalitMaganti, line 41: This should be wrapped in an if nice once std::optonal below.

src/trace_redaction/main.cc
```diff
@@ -14,18 +14,35 @@
...
+  setpriority(PRIO_PROCESS, 0, nice);
+#else
+  base::ignore_result(nice);
```
LalitMaganti: Ifn nice is specified you should log an error message and return saying it's not supported on this platform.

src/trace_redaction/main.cc
```diff
@@ -14,18 +14,35 @@
...
+
 namespace perfetto::trace_redaction {
 
 // Builds and runs a trace redactor.
 static base::Status Main(std::string_view input,
                          std::string_view output,
-                         std::string_view package_name) {
+                         std::string_view package_name,
+                         int nice) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID) && \
```
LalitMaganti, line 39: If you're going to make this a command line argument, you should now do this on both linux *and* Android and remove the PERFETTO_ANDROID_BUILD so it's easier to test. I was only suggesting PERFETTO_ANDROID_BUILD becuase you were doing it uncojditionally.

edgararriagag:
As agreed offline, we will land first the unconditional priority change to de-risk the module launch. We will create a follow up CL that will plumb the cmdline arguments.

LalitMaganti (approved):

primiano (approved):

PR #5554, opened 2026-04-23
tp: Display all messages involved in hash collisions

srapiteanu:
Improves ProtoLog handling by showing all messages that result in the same hash, even across buildtime-hashed and runtime-hashed messages. This ensures no ProtoLog data is hidden due to hash collisions.


src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -75,10 +75,17 @@ class ProtoLogMessageDecoder {
...
+                            const std::vector<int64_t>& sint64_params,
+                            const std::vector<double>& double_params,
+                            const std::vector<bool>& boolean_params,
+                            const std::vector<std::string>& string_params);
+
  private:
   TraceProcessorContext* const context_;
   base::FlatHashMap<uint64_t, TrackedGroup> tracked_groups_;
-  base::FlatHashMap<uint64_t, TrackedMessage> tracked_messages_;
+  base::FlatHashMap<uint64_t, std::vector<TrackedMessage>> tracked_messages_;
```
rukkal: I assume that 99.9% of the time the std::vector will contain only one element. If so, this should rather be a `SmallVector<TrackedMessage, 1>`

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -35,18 +35,96 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+      collision_indicator += "'" + formatted_message + "'";
+      if (i < messages.size() - 1) {
+        collision_indicator += "\n ";
+      }
+    }
+    collision_indicator += "'>";
+
+    return DecodedMessage{ProtoLogLevel::WARN, nullptr, collision_indicator,
+                          std::nullopt};
+  }
```
rukkal, line 68: nit: factor out the code handling the exceptional case into something like
```
if (messages.size() > 1) {
   DecodeCollidingMessageIds(...);
}
```
?

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -176,32 +245,102 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+                          collision_message, potential_matches[0].location};
+  } else {
+    if (potential_matches.empty()) {
+      collision_message += "NO TYPE MATCH >";
+    } else {
+      collision_message += "MULTIPLE TYPE MATCHES : ";
+      for (size_t i = 0; i < potential_matches.size(); ++i) {
+        auto formatted_message =
+            FormatMessage(potential_matches[i].message, sint64_params,
+                          double_params, boolean_params, string_params);
```
PabiGamito, line 380: does this fail/crash if the message doesn't have the right parameter types?

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -75,10 +83,28 @@ class ProtoLogMessageDecoder {
...
                     const std::optional<std::string>& location);
 
+ private:
+  std::string FormatMessage(const std::string& message,
+                            const std::vector<int64_t>& sint64_params,
+                            const std::vector<double>& double_params,
+                            const std::vector<bool>& boolean_params,
+                            const std::vector<std::string>& string_params);
+
+  ProtoLogParamCounts GetParameterSignature(const std::string& message);
```
PabiGamito: we can be more precise than just counts, the order of arguments also matters, and can help resolve conflicts

PabiGamito (changes requested):

src/trace_processor/storage/stats.h
```diff
@@ -526,6 +526,9 @@ namespace perfetto::trace_processor::stats {
   F(winscope_protolog_view_config_collision,                                   \
                                           kSingle,  kInfo,     kAnalysis,      \
       "Got a viewer config collision!"),                                       \
+  F(winscope_protolog_view_config_collision_resolved,                                   \
+                                          kSingle,  kInfo,     kAnalysis,      \
+      "Got a viewer config collision!"),                                       \
```
PabiGamito: doesn't match the stats name fully

PabiGamito (changes requested):

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -75,10 +78,34 @@ class ProtoLogMessageDecoder {
...
+
+  base::SmallVector<char, 8> GetParameterSignature(const std::string& message);
+
+  bool MatchesParameterSequence(const TrackedMessage& tracked_message,
+                                const std::vector<int64_t>& sint64_params,
+                                const std::vector<double>& double_params,
+                                const std::vector<bool>& boolean_params,
+                                const std::vector<std::string>& string_params);
+
+  std::optional<DecodedMessage> DecodeCollidiongMessageIds(
```
PabiGamito: nit: typo

PabiGamito (changes requested):

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -176,32 +245,140 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+  for (size_t i = 0; i < messages.size(); ++i) {
+    if (MatchesParameterSequence(messages[i], sint64_params, double_params,
+                                 boolean_params, string_params)) {
+      potential_matches.emplace_back(messages[i]);
+    }
+  }
+
+  std::string collision_message =
+      "<PROTOLOG COLLISION (id=0x" + base::Uint64ToHexString(message_id) + ") ";
+  if (potential_matches.size() == 1) {
```
PabiGamito, line 347: if we have a single match I don't think we should even mention anything about the collision we should gracefully recover and just format the message normally using that match string

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -35,18 +35,87 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+  }
+
+  base::SmallVector<TrackedMessage, 1>& existing_messages =
+      *existing_messages_ptr;
+  bool message_already_tracked = false;
+
+  for (const auto& existing_msg : existing_messages) {
+    if (existing_msg.message == new_tracked_message.message &&
+        existing_msg.level == new_tracked_message.level &&
+        existing_msg.group_id == new_tracked_message.group_id) {
```
PabiGamito: this might actually have already been a bug before but now I'm thinking if we should consider the message's code location as well when detecting if a message is already tracked or not? so that when we decode we make sure we also show the correct `location` for the case where two different messages happen to be the same log message and log group, which realistically could happen.

PabiGamito (approved):

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -35,18 +35,86 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+  }
+}
+
+void ProtoLogMessageDecoder::TrackGroup(uint32_t id, const std::string& tag) {
+  auto tracked_group = tracked_groups_.Find(id);
+  if (tracked_group != nullptr && tracked_group->tag != tag) {
+    context_->storage->IncrementStats(
+        stats::winscope_protolog_view_config_collision);
   }
+  tracked_groups_.Insert(id, TrackedGroup{tag});
```
PabiGamito: don't know if there is a super good solution, so feel free to resolve, but wondering if we can think of a way to make sure it's clear to the end user that there was a group collision instead of just showing the last added group in the case of a collision, thoughts?

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -35,18 +35,86 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+                                     double_params, boolean_params,
+                                     string_params);
+  }
+}
+
+void ProtoLogMessageDecoder::TrackGroup(uint32_t id, const std::string& tag) {
+  auto tracked_group = tracked_groups_.Find(id);
+  if (tracked_group != nullptr && tracked_group->tag != tag) {
+    context_->storage->IncrementStats(
+        stats::winscope_protolog_view_config_collision);
```
PabiGamito: nit: should we report message configs and group config collisions separately? might help us better track and figure out issues quicker

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -176,32 +244,139 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+        stats::winscope_protolog_view_config_collision_resolved);
+
+    auto formatted_message =
+        FormatMessage(potential_matches[0].message, sint64_params,
+                      double_params, boolean_params, string_params);
+
+    auto group = tracked_groups_.Find(potential_matches[0].group_id);
+    std::string group_tag;
+    if (group == nullptr) {
+      group_tag = "UNKNOWN TAG";
```
PabiGamito: thoughts on reporting some error stat about this case?

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -176,32 +244,139 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+    return DecodedMessage{potential_matches[0].level, group_tag,
+                          formatted_message, potential_matches[0].location};
+  } else {
     context_->storage->IncrementStats(
         stats::winscope_protolog_view_config_collision);
+    std::string collision_message = "<PROTOLOG COLLISION (id=0x" +
+                                    base::Uint64ToHexString(message_id) + ") ";
+
+    if (potential_matches.empty()) {
+      collision_message += "NO TYPE MATCH >";
```
PabiGamito: nit: extra space? since we have none at the start should we also have none at the end?

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -29,6 +29,8 @@
 
 namespace perfetto::trace_processor::winscope {
 
+inline constexpr char kCollisionGroupTag[] = "COLLISION_GROUP";
```
PabiGamito: nit: `inline constexpr std::string_view kCollisionGroupTag = "COLLISION_GROUP";`

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
srapiteanu: Code updated: in case of group tag collision the tag is updated to "COLLISION_GROUP" and a new stat to count the instances is increased. Thank you!

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -40,20 +43,21 @@ enum ProtoLogLevel : int32_t {
 
 struct DecodedMessage {
   ProtoLogLevel log_level;
-  std::string group_tag;
+  std::string_view group_tag;
```
PabiGamito: let's revert these changes from string to string_view, otherwise the ownership of the strings is not so well defined and I'm not sure if when we clear it. You can convert from string_view to string implicitly by just wrapping it in `std::string(my_string_view)` so that we copy over the string and the structs have ownership of these strings, it's very possible the tag strings underlying the string views outlive these structs anyway but I think it make sense to have the structs properly own the strings anyways. You can keep using string_views elsewhere though wherever they are passed around.

src/trace_processor/importers/proto/winscope/protolog_message_decoder.h
```diff
@@ -67,18 +71,42 @@ class ProtoLogMessageDecoder {
       const std::vector<bool>& boolean_params,
       const std::vector<std::string>& string_params);
 
-  void TrackGroup(uint32_t id, const std::string& tag);
+  void TrackGroup(uint32_t id, const std::string_view tag);
```
PabiGamito: nit: const is no really needed here now that we are passing std::string_view by value.

src/trace_processor/storage/stats.h
```diff
@@ -523,9 +523,18 @@ namespace perfetto::trace_processor::stats {
   F(winscope_protolog_message_decoding_failed,                                 \
                                           kSingle,  kInfo,     kAnalysis,      \
       "Failed to decode ProtoLog message."),                                   \
-  F(winscope_protolog_view_config_collision,                                   \
+  F(winscope_protolog_message_collision,                                       \
                                           kSingle,  kInfo,     kAnalysis,      \
-      "Got a viewer config collision!"),                                       \
+      "Got a message collision!"),                                             \
```
PabiGamito: nit: I know we were not explicit about this before, but can we make it clearer in the messages that it's a protolog message collision?

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -35,18 +35,96 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+    single_message_vector.emplace_back(new_tracked_message);
+    tracked_messages_.Insert(message_id, std::move(single_message_vector));
+    return;
+  }
+
+  base::SmallVector<TrackedMessage, 1>& existing_messages =
+      *existing_messages_ptr;
+  bool message_already_tracked = false;
+
+  for (const auto& existing_msg : existing_messages) {
```
PabiGamito: nit: minor performance improvement, but we could check this first before even allocating the `TrackedMessage` and doing any other work, and return early if it's already tracked, and only after do any other work including allocating the `TrackedMessage`.

PabiGamito (approved):
Very minor nit, and otherwise rest looks good to me and good to merge!

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -176,32 +248,141 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
-void ProtoLogMessageDecoder::TrackGroup(uint32_t id, const std::string& tag) {
-  auto tracked_group = tracked_groups_.Find(id);
-  if (tracked_group != nullptr && tracked_group->tag != tag) {
-    context_->storage->IncrementStats(
-        stats::winscope_protolog_view_config_collision);
+base::SmallVector<char, 8> ProtoLogMessageDecoder::GetParameterSignature(
+    const std::string& message) {
+  base::SmallVector<char, 8> signature;
+  for (size_t i = 0; i < message.length(); ++i) {
+    if (message.at(i) == '%' && i + 1 < message.length()) {
```
rukkal: In out case (with exceptions disabled) I am not sure what is the generated code of std::string::at(). But the general idea is that at() performs some extra bounds checking, whereas operator[] doesn't. Let's use operator[] as we are already taking care of the bound checks.

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -35,18 +35,90 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
+    single_message_vector.emplace_back(TrackedMessage{
+        level, group_id, message, location, GetParameterSignature(message)});
+    tracked_messages_.Insert(message_id, std::move(single_message_vector));
+    return;
+  }
+
+  base::SmallVector<TrackedMessage, 1>& existing_messages =
+      *existing_messages_ptr;
+
+  for (const auto& existing_msg : existing_messages) {
```
rukkal: consider using the STL (e.g. std::find) instead of hand-written loops

src/trace_processor/importers/proto/winscope/protolog_message_decoder.cc
```diff
@@ -176,32 +248,141 @@ std::optional<DecodedMessage> ProtoLogMessageDecoder::Decode(
...
-  auto tracked_group = tracked_groups_.Find(id);
-  if (tracked_group != nullptr && tracked_group->tag != tag) {
-    context_->storage->IncrementStats(
-        stats::winscope_protolog_view_config_collision);
+base::SmallVector<char, 8> ProtoLogMessageDecoder::GetParameterSignature(
+    const std::string& message) {
+  base::SmallVector<char, 8> signature;
+  for (size_t i = 0; i < message.length(); ++i) {
+    if (message.at(i) == '%' && i + 1 < message.length()) {
+      switch (message.at(i + 1)) {
```
rukkal: ditto

rukkal (approved):
lgtm. Just some minor comments

LalitMaganti:
So couple of meta points:
1) Please fix the PR title and description. The commit messages don't matter once the PR lands and is squashed, it's all taken from the PR itself. Right now the PR title is not in a good state
2) I'm honestly very confused as to why we are even trying to work around hash collisisons instead of just preventing them with a better hash function. This seems like a lot of code to workaround an issue which should be extraordinarily rare. But I don't want to start investigating myself so I'll leave it at that.

LalitMaganti (approved):

PR #5555, opened 2026-04-23
tp: Add tables with stdlib documentation data

aMayzner:


src/trace_processor/perfetto_sql/intrinsics/operators/stdlib_docs_operator.h
```diff
@@ -0,0 +1,220 @@
...
+
+#include "src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h"
+#include "src/trace_processor/sqlite/bindings/sqlite_module.h"
+#include "src/trace_processor/util/stdlib_doc_parser.h"
+
+namespace perfetto::trace_processor {
+
+// Eponymous vtab: __intrinsic_stdlib_modules
+// Returns all modules registered in the engine, one row per module.
+struct StdlibDocsModulesOperator : sqlite::Module<StdlibDocsModulesOperator> {
```
LalitMaganti: There's absolutely no reason this needs to be an operator. This can be a table function quite simply.

src/trace_processor/util/stdlib_doc_parser.cc
```diff
@@ -0,0 +1,466 @@
...
+      pos++;
+    }
+  }
+  return line.substr(pos);
+}
+
+// Given the full SQL source and a byte offset of a statement, extracts the
+// contiguous block of line comments (-- ...) immediately preceding the
+// statement. Returns the joined description text.
+std::string ExtractDescriptionAbove(const char* sql, uint32_t stmt_offset) {
```
LalitMaganti: This is completely unnecessary. Comment will be attached to the statement node. You don't need to do any of this.

LalitMaganti: Like the whole reason we are using syntaqlite is to avoid horrible things like this from needing to happen.

src/trace_processor/util/stdlib_doc_parser.cc
```diff
@@ -0,0 +1,466 @@
...
+#include "src/trace_processor/perfetto_sql/syntaqlite/syntaqlite_perfetto.h"
+
+namespace perfetto::trace_processor::stdlib_doc {
+
+namespace {
+
+// Returns the text for a SyntaqliteTextSpan as a std::string.
+std::string SpanText(SyntaqliteParser* p, SyntaqliteTextSpan span) {
+  uint32_t len;
+  const char* text = syntaqlite_parser_span_expanded_text(p, &span, &len);
```
LalitMaganti: WHy expanded_text and not text? Why make a copy instead of using a string_view?

src/trace_processor/util/stdlib_doc_parser.cc
```diff
@@ -0,0 +1,466 @@
...
+      }
+
+      default:
+        // INCLUDE PERFETTO MODULE, DROP PERFETTO INDEX, plain SQL, etc.
+        // We don't document these.
+        break;
+    }
+  }
+
+  syntaqlite_parser_destroy(p);
```
LalitMaganti: Please properly wrap this in a RAII type rather than using it directly like this. If someone accidentally adds a return above, you'll end up leaking lots of memory.

src/trace_processor/util/BUILD.gn
```diff
@@ -26,6 +26,18 @@ source_set("stdlib") {
   sources = [ "sql_modules.h" ]
 }
 
+source_set("stdlib_doc_parser") {
```
LalitMaganti: sql_module_doc_parser

src/trace_processor/perfetto_sql/engine/perfetto_sql_engine.h
```diff
@@ -245,6 +245,20 @@ class PerfettoSqlEngine {
   sql_modules::RegisteredPackage* FindPackage(const std::string& name) {
     return packages_.Find(name);
   }
+  const sql_modules::RegisteredPackage* FindPackage(
+      const std::string& name) const {
+    return packages_.Find(name);
+  }
+
+  // Calls |fn(package_name, module_key)| once for every registered module.
```
LalitMaganti: Why do this instead of just returning an std::vector? We try to avoid callbacks+templates in TP where possible

src/trace_processor/perfetto_sql/intrinsics/table_functions/stdlib_docs_table_function.cc
```diff
@@ -0,0 +1,301 @@
...
+#include "src/trace_processor/util/simple_json_serializer.h"
+#include "src/trace_processor/util/sql_modules.h"
+#include "src/trace_processor/util/stdlib_doc_parser.h"
+
+namespace perfetto::trace_processor {
+
+namespace {
+
+StringPool::Id Intern(StringPool* pool, const std::string& s) {
+  return pool->InternString(base::StringView(s));
```
LalitMaganti: This seems very questionable as a function.

src/trace_processor/util/stdlib_doc_parser.cc
```diff
@@ -0,0 +1,432 @@
...
+
+namespace perfetto::trace_processor::stdlib_doc {
+
+namespace {
+
+// Returns the authored source text for a span as a string_view into the
+// original input — no allocation. Uses span_text (not expanded_text) so
+// macro-expanded names resolve back to the call-site source, not the
+// internal expansion buffer.
+std::string_view SpanText(SyntaqliteParser* p, SyntaqliteTextSpan span) {
```
LalitMaganti: This is a copy of a function that already exists. We should refactor this into syntaqltie/utils.{h, cc}

src/trace_processor/util/stdlib_doc_parser.cc
```diff
@@ -0,0 +1,432 @@
...
+  uint32_t len = 0;
+  const char* text = syntaqlite_parser_span_text(p, &span, &len, nullptr);
+  if (!text) {
+    return {};
+  }
+  return {text, len};
+}
+
+// Strips leading "-- " or "--" from a single comment line.
+std::string StripCommentPrefix(const char* text, uint32_t len) {
```
LalitMaganti: This can return an std::string_view.

src/trace_processor/util/sql_module_doc_parser.cc
```diff
@@ -0,0 +1,432 @@
...
+std::string StripCommentPrefix(const char* text, uint32_t len) {
+  std::string_view s = base::TrimWhitespace(std::string_view(text, len));
+  if (s.size() >= 2 && s[0] == '-' && s[1] == '-') {
+    s.remove_prefix(2);
+  }
+  return std::string(base::TrimWhitespace(s));
+}
+
+// Joins all line comments (kind == 0) from |comments| into a single string,
+// space-separated. |stmt_ptr| is the base pointer for comment offsets.
```
LalitMaganti, line 45: This can be made more efficent by sharing the vector and in general reducing the allocations here. With the amount of SQL we have these days, I can easily see the amount of work we're doing here actually being a problem.

src/trace_processor/util/stdlib_doc_parser.cc
```diff
@@ -0,0 +1,432 @@
...
+    std::string s =
+        StripCommentPrefix(stmt_ptr + comments[i].offset, comments[i].length);
+    if (!s.empty()) {
+      parts.push_back(std::move(s));
+    }
+  }
+  return base::Join(parts, " ");
+}
+
+// Returns the last contiguous block of line comments (no blank line between
```
LalitMaganti: AFAIK syntaqlite will already do this. If it doesn't please file an issue against me so I can fix this upstream.

src/trace_processor/util/sql_module_doc_parser.cc
```diff
@@ -0,0 +1,432 @@
...
+  }
+  return JoinLineComments(stmt_ptr, comments + block_start,
+                          count - block_start);
+}
+
+// Maps a source pointer (returned by a syntaqlite span/node accessor) to the
+// index of the corresponding token in the per-statement token array. Returns
+// UINT32_MAX if not found.
+// O(token_count) per call. Stdlib modules are small (hundreds of tokens), so
+// this is acceptable even when called once per arg.
```
LalitMaganti: You don't need to do this, there's already a function which does this `syntaqlite_node_token_range`.

src/trace_processor/util/sql_module_doc_parser.cc
```diff
@@ -0,0 +1,432 @@
...
+template <typename Entry>
+std::vector<Entry> ExtractArgDefList(SyntaqliteParser* p,
+                                     uint32_t list_id,
+                                     const char* stmt_ptr,
+                                     const SyntaqliteParserToken* tokens,
+                                     uint32_t token_count) {
+  std::vector<Entry> result;
+  if (!syntaqlite_node_is_present(list_id)) {
+    return result;
+  }
```
LalitMaganti, line 105: You can use the SYNTAQLITE_NODE/SYNTAQLITE_LIST_ITEM/SYNTAQLITE_LIST_FOREACH macros which will amke this code nicer.

src/trace_processor/util/sql_module_doc_parser.cc
```diff
@@ -0,0 +1,432 @@
...
+std::vector<Arg> ExtractMacroArgs(SyntaqliteParser* p,
+                                  uint32_t list_id,
+                                  const char* stmt_ptr,
+                                  const SyntaqliteParserToken* tokens,
+                                  uint32_t token_count) {
+  std::vector<Arg> result;
+  if (!syntaqlite_node_is_present(list_id)) {
+    return result;
+  }
+  const auto* list = static_cast<const SyntaqlitePerfettoMacroArgList*>(
```
LalitMaganti, line 167: As above.

src/trace_processor/util/sql_module_doc_parser.cc
```diff
@@ -0,0 +1,417 @@
...
+#include "src/trace_processor/perfetto_sql/syntaqlite/syntaqlite_perfetto.h"
+#include "src/trace_processor/perfetto_sql/syntaqlite/utils.h"
+
+namespace perfetto::trace_processor::stdlib_doc {
+
+namespace {
+
+// Strips leading "-- " or "--" from a single comment line. Returns a
+// string_view into the original source buffer — no allocation.
+std::string_view StripCommentPrefix(const char* text, uint32_t len) {
```
LalitMaganti: Passing in as a string_view would be neater.

src/trace_processor/util/sql_module_doc_parser.cc
LalitMaganti: This seems not done?

LalitMaganti (commented):
Doesn't seem like any opf my comments from before were addressed.

src/trace_processor/util/sql_module_doc_parser.cc
aMayzner: the standalone function is replaced with syntaqlite_node_token_range where it works (return type description), but for arg names the node extents aren't recorded for IdentName sub-nodes, so a span-based offset scan is still needed there

src/trace_processor/perfetto_sql/intrinsics/table_functions/stdlib_docs_table_function.cc
LalitMaganti: Sorry this is still not solved.

LalitMaganti (changes requested):

LalitMaganti (approved):

PR #5559, opened 2026-04-23
Remove depth filter from critical blocking calls

nicomazz:
Top level handler tracing doesn't work after the introduction of Message queue tracing because the depth is no longer 0. Removing the depth filter allows these slices to be captured again.

diff tests are updated accordingly. 

Bug: 505313258


LalitMaganti (approved):

nicomazz:
the formatter complained during upload about the `.out` files. they're now formatted properly. I would have split it into 2 changes but the change is trivial enough ..

PR #5562, opened 2026-04-23
sdk: Guard activateTrigger in Java on sdk registration

zezeozue:
Calling activateTrigger before register() SIGTRAPs in TracingMuxerFake. Callers worked around this by pre-checking getAttempedSystemRegistration(), but that check was stricter than it needed to be and silently dropped triggers against in-process registrations, which the native muxer handles safely.

Move the guard inside activateTrigger and key it on any register() call. The return value now tells the caller whether the trigger reaches the system backend, so callers can skip the pre-check and only fall back to trigger_perfetto when it returns false.

### DIFF
--- a/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
+++ b/src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
@@ -46,6 +46,7 @@ public final class PerfettoTrace {
   private static final PerfettoNativeMemoryCleaner sNativeMemoryCleaner =
       new PerfettoNativeMemoryCleaner();
 
+  private static final AtomicBoolean sAttemptedRegistration = new AtomicBoolean(false);
   private static final AtomicBoolean sAttemptedSystemRegistration = new AtomicBoolean(false);
 
   /** For fetching the next flow event id in a process. */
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Uhhhh this seems like a very wrong fix in my exeprience. The SDK automatically "buffers" ActivateTriggers calls until the connection is made. Working around this seems like very much the *wrong* thign to do.

Not to mention I don't understand the use of TracingMuxerFake - why is that running on the production codepath at all?

Overall something is smelling very wrong in this change.

zezeozue:
> Uhhhh this seems like a very wrong fix in my exeprience. The SDK automatically "buffers" ActivateTriggers calls until the connection is made. Working around this seems like very much the _wrong_ thign to do.
> 
> Not to mention I don't understand the use of TracingMuxerFake - why is that running on the production codepath at all?
> 
> Overall something is smelling very wrong in this change.

For context, here's the crash: b/504755548

My understanding is that we hit TracingMuxerFake when not registered yet?

LalitMaganti:
> My understanding is that we hit TracingMuxerFake when not registered yet?

This is imprecise. We hit TracingMuxerFake when we are not *initialized* yet. Initialization != registration, initialization is syncrhonus, registration is async.

All you are doing is hiding a problem by doing this. The same thing would happen if you e.g. had a PerfettoTrace.emit() call as well.

The correct fix is to *initialize* the SDK earlier. Once iniitlaized, triggers will be buffered and everything else should work.

zezeozue:
Will paste your suggestion on the bug and cc you.

PR #5563, opened 2026-04-23
trace_processor: Add Windows ETW disk I/O event parsing

davidbienvenu:
Add support for disk i/o ETW events recorded from Chrome tracing, including duration., thread, etc.


davidbienvenu:
This PR adds Windows ETW disk io event parsing to perfetto tracing.

protos/perfetto/trace/etw/etw.proto
```diff
@@ -278,3 +278,17 @@ message StackWalkEtwEvent {
   optional string trigger = 1;
   optional uint64 callstack_iid = 2;
 }
+
+// Proto definition based on the `DiskIo_TypeGroup1` class definition.
+message DiskIoEtwEvent {
+  optional uint32 disk_number = 1;
+  optional uint32 irp_flags = 2;
+  optional uint32 transfer_size = 3;
+  optional uint32 reserved = 4;
```
seanptmaher: I wouldn't bother adding this field, unless you know how it is actually used by the kernel and can tag it appropriately.

protos/perfetto/trace/etw/etw.proto
```diff
@@ -278,3 +278,17 @@ message StackWalkEtwEvent {
...
+  optional int64 byte_offset = 5;
+  optional uint64 file_object = 6;
+  optional uint64 irp_ptr = 7;
```
seanptmaher: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_irp

Do we care about the contents of this packet?

it's sorta useless to record this but not record the fields in here.

could you take a look in here and figure out whether you care about these fields, and if yes, record them directly, and if not, remove this line?

src/trace_processor/importers/etw/disk_io_tracker.h
```diff
@@ -0,0 +1,88 @@
...
+
+namespace perfetto::trace_processor {
+
+// A class to keep track of disk I/O events recorded by Event Tracing for
+// Windows (ETW).
+class DiskIoTracker {
+ public:
+  explicit DiskIoTracker(TraceProcessorContext* context);
+
+  void ParseDiskIo(int64_t timestamp, protozero::ConstBytes);
```
seanptmaher: it doesn't look like this is ever called?

src/trace_processor/importers/etw/etw_parser.cc
```diff
@@ -110,9 +111,25 @@ base::Status EtwParser::ParseEtwEvent(uint32_t cpu,
...
 
+  if (decoder.has_disk_io()) {
+    context_->disk_io_tracker->ParseDiskIo(ts, decoder.disk_io());
+  }
+
   return base::OkStatus();
 }
 
+void EtwParser::ParseDiskIo(int64_t timestamp, ConstBytes blob) {
+  protos::pbzero::DiskIoEtwEvent::Decoder disk_io(blob);
```
seanptmaher: it doesn't look like this is ever called?

protos/perfetto/trace/etw/etw.proto
davidbienvenu: Removed.

protos/perfetto/trace/etw/etw.proto
davidbienvenu: Do you mean the irp_ptr? We don't care about its contents, but we do care about the ptr itself, in order to match start and end events (e.g., which kReadInit corresponds with a kRead).

src/trace_processor/importers/etw/disk_io_tracker.h
davidbienvenu: It's called from etw_parser.cc, but EtwParser::ParseDiskIo isn't called, and I've removed that.

src/trace_processor/importers/etw/etw_parser.cc
davidbienvenu: Removed.

protos/perfetto/trace/etw/etw.proto
seanptmaher: Yeah, irp_ptr. In that case, hash the pointer, because we don't want to leak kernel structure pointers.

src/trace_processor/importers/etw/disk_io_tracker.cc
```diff
@@ -0,0 +1,229 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
```
seanptmaher: 2026

src/trace_processor/importers/etw/disk_io_tracker.cc
```diff
@@ -0,0 +1,229 @@
...
+#include "src/trace_processor/importers/common/slice_tracker.h"
+#include "src/trace_processor/importers/common/track_compressor.h"
+
+namespace perfetto::trace_processor {
+
+namespace {
+
+using protozero::ConstBytes;
+using std::nullopt;
+using std::optional;
```
seanptmaher: this seems pretty odd for something you only use in one function. maybe move these to that function if you insist, but i'd remove them.

protos/perfetto/trace/etw/etw.proto
davidbienvenu: Yes, we hash it when we record the trace, just like the file_io does, e.g., EtwConsumer::DecodeFileIoInfoEvent - so we're only recording a hashed version of the irp_ptr.

src/trace_processor/importers/etw/disk_io_tracker.cc
```diff
@@ -0,0 +1,229 @@
...
+
+  switch (opcode) {
+    case kReadInit:
+    case kWriteInit:
+    case kFlushInit: {
+      const char* event_type = GetEventTypeString(opcode);
+      if (!event_type)
+        return;
+      StringId name = context_->storage->InternString(event_type);
+      StartEvent(irp, name, timestamp, utid, std::move(set_args));
```
seanptmaher: why are you passing all the bound variables and the closure itself by value?

seanptmaher (approved):

src/trace_processor/importers/etw/disk_io_tracker.cc
davidbienvenu: It gets stored in started_events_, so they outlast the calling function.

src/trace_processor/importers/etw/disk_io_tracker.cc
davidbienvenu: Done.

src/trace_processor/importers/etw/disk_io_tracker.cc
davidbienvenu: done.

src/trace_processor/importers/etw/disk_io_tracker.h
```diff
@@ -0,0 +1,87 @@
...
+
+  void OnEventsFullyExtracted();
+
+ private:
+  struct StartedEvent {
```
LalitMaganti: Why do you need to store these? this is totally unnecessary.

src/trace_processor/importers/etw/disk_io_tracker.cc
```diff
@@ -0,0 +1,230 @@
...
+  // End the slice for this event.
+  const auto track_id = context_->track_compressor->InternEnd(
+      kBlueprint, tracks::Dimensions(utid),
+      /*cookie=*/static_cast<int64_t>(irp));
+  context_->slice_tracker->End(end_timestamp, track_id, kNullStringId,
+                               event_name, std::move(args));
+  started_events_.erase(started_event_it);
+}
+
+void DiskIoTracker::RecordEventWithoutIrp(StringId name,
```
LalitMaganti: This is highly abnormal and breaks the semantic of how *all* other events work. I would strongly reconsider this.

src/trace_processor/importers/etw/disk_io_tracker.cc
```diff
@@ -0,0 +1,230 @@
...
+      "etw_diskio",
+      tracks::DimensionBlueprints(tracks::kThreadDimensionBlueprint));
+
+  const auto event_name = started_event_it->second.name;
+
+  // End the slice for this event.
+  const auto track_id = context_->track_compressor->InternEnd(
+      kBlueprint, tracks::Dimensions(utid),
+      /*cookie=*/static_cast<int64_t>(irp));
+  context_->slice_tracker->End(end_timestamp, track_id, kNullStringId,
```
LalitMaganti: You can just look at the id returned by end and use that instead of storing anything in started_events_. None of that is necessary.

LalitMaganti (commented):
Please rework the PR title and description which will become the commit title/description.

src/trace_processor/importers/etw/disk_io_tracker.cc
davidbienvenu: I've removed `started_events_`, which was modelled after the file_io_tracker's calculation of elapsed time for operations. It turns out that there's new documentation for the disk io events on MSDN, and there's a high_res_response_time for the read/write/flush events, so I've switched to using that.  PTAL

src/trace_processor/importers/etw/disk_io_tracker.h
davidbienvenu: I've removed the code that stored these. I was using it to calculate the latency between an init event and the corresponding event finishing, but I've removed that.

src/trace_processor/importers/etw/disk_io_tracker.cc
davidbienvenu: I've removed started_events_.

src/trace_processor/importers/etw/disk_io_tracker.cc
```diff
@@ -0,0 +1,171 @@
...
+
+  const char* event_type = GetEventTypeString(opcode);
+  if (!event_type)
+    return;
+  StringId name = context_->storage->InternString(event_type);
+  HandleEvent(name, issuing_thread_id.value_or(utid), timestamp,
+              response_time.value_or(0), std::move(set_args));
+}
+
+void DiskIoTracker::HandleEvent(StringId name,
```
LalitMaganti: What's the point in having this separate function which is only used form one place?

src/trace_processor/importers/etw/disk_io_tracker.cc
```diff
@@ -0,0 +1,171 @@
...
+        }
+        if (issuing_thread_id) {
+          inserter->AddArg(issuing_thread_id_arg_,
+                           Variadic::UnsignedInteger(*issuing_thread_id));
+        }
+      };
+
+  const char* event_type = GetEventTypeString(opcode);
+  if (!event_type)
+    return;
```
LalitMaganti, line 153: Generally in situations like this we add an import log about why this message was dropped to avoid mysterious debugging in the future where you don't see events which are in the trace appear in the UI.

davidbienvenu: Used PERFETTO_DLOG - hope that's the right one.

src/trace_processor/importers/etw/disk_io_tracker.cc
davidbienvenu: pulled inline

src/trace_processor/importers/etw/disk_io_tracker.cc
LalitMaganti: No it's not correct. Import logging is a specific API in trace processor to make the logs avaialble as an SQL table so it can e.g. be queiried in the UI. Please follow the example of https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/importers/proto/gpu_event_parser.cc;l=528?q=import_log%20f:perfetto&ss=chromium

davidbienvenu: Ah, thanks, done. It seems simple enough not to warrant a description, but I can add one if you want.

src/trace_processor/types/trace_processor_context.h
```diff
@@ -226,6 +227,7 @@ class TraceProcessorContext {
   PerTraceAndMachinePtr<TrackTracker> track_tracker;
   PerTraceAndMachinePtr<TrackCompressor> track_compressor;
   PerTraceAndMachinePtr<SliceTracker> slice_tracker;
+  PerTraceAndMachinePtr<DiskIoTracker> disk_io_tracker;
```
LalitMaganti: This doens't need to be on the context, this is an internal implementation detail of ETW. Please move it insdie. Honestly even the FileIoTracker should not be there, it might have been grandfathered in from before.

src/trace_processor/storage/stats.h
```diff
@@ -138,6 +138,7 @@ namespace perfetto::trace_processor::stats {
   F(app_wakelock_unknown_id,              kSingle,  kError,    kAnalysis, Scope::kMachineAndTrace,      \
        "Interning ID not found. Should never happen."),                        \
   F(meminfo_unknown_keys,                 kSingle,  kError,    kAnalysis, Scope::kMachineAndTrace, ""), \
+  F(missing_disk_io_event_name,           kSingle,  kError,    kAnalysis, Scope::kMachineAndTrace, ""), \
```
LalitMaganti: Please add a description

src/trace_processor/types/trace_processor_context.h
davidbienvenu: Inside ETWParser?

LalitMaganti: Yes

src/trace_processor/storage/stats.h
davidbienvenu: done.

src/trace_processor/types/trace_processor_context.h
davidbienvenu: Done

davidbienvenu:
PTAL

src/trace_processor/trace_processor_storage_impl.cc
```diff
@@ -39,6 +39,7 @@
 #include "src/trace_processor/importers/common/stack_profile_tracker.h"
 #include "src/trace_processor/importers/common/symbol_tracker.h"
 #include "src/trace_processor/importers/common/trace_file_tracker.h"
+#include "src/trace_processor/importers/etw/disk_io_tracker.h"
```
LalitMaganti: Please remove.

davidbienvenu: done.

davidbienvenu:
PTAL

LalitMaganti (approved):

PR #5577, opened 2026-04-24
perfetto: Remove trace_to_hprof converter

LalitMaganti:
This was an abandoned experiment that never went anywhere. Drops the
traceconv `hprof` subcommand along with the now-unused
trace_to_hprof.{h,cc} sources and their BUILD.gn entries.


ilkos (approved):
thanks!

PR #5588, opened 2026-04-25
tp: stop mutating PacketSequenceStateGeneration on packet loss

LalitMaganti:
`PacketSequenceStateGeneration::OnPacketLoss()` previously mutated `this` (flipping `is_incremental_state_valid_` and the track-event delta state's `timestamps_valid_`) and returned `RefPtr(this)`. The TraceSorter holds RefPtrs to the same generation for buffered packets that were tokenized while the sequence was valid, so the in-place flip was retroactively visible from those RefPtrs. No current reader observes the mutation through a buffered RefPtr, but that is incidental — anything added in the future that consults the generation's validity from the parser side would see a stale, retroactively-invalidated view.

Make `OnPacketLoss()` return a new generation instead, sharing the per-interval state (interned data, custom states, persistent thread descriptor) with the previous one but with the validity bits cleared and delta-encoded track-event state reset. Buffered packets retain the view they had at tokenization time; the Builder swaps its current generation to the new instance so subsequent tokenizations see the cleared state.

To support this, replace `TrackEventSequenceState::OnPacketLoss()` (which mutated `timestamps_valid_` in place) with a value-returning form that produces a fresh successor with persistent state preserved and delta state reset. Incremental counter values and running reference timestamps cannot be carried across the lost run of packets, so they are dropped along with `timestamps_valid_`.

The full PSSG constructor's `trace_packet_defaults` parameter changes from `TraceBlobView` to `std::optional<InternedMessageView>` so OnPacketLoss can pass the existing defaults through without having to reach inside the optional and reconstruct the view.

---

**Stack:**
- **#5588 — tp: stop mutating PacketSequenceStateGeneration on packet loss** (this PR)
- #5590 — tp: split TrackEventSequenceState into descriptor + delta state
- #5593 — tp: extract IncrementalState to fix CustomState UAF

### DIFF
--- a/src/trace_processor/importers/proto/track_event_sequence_state.h
+++ b/src/trace_processor/importers/proto/track_event_sequence_state.h
@@ -32,11 +32,17 @@ class TrackEventSequenceState {
     return TrackEventSequenceState(PersistentState());
   }
 
-  TrackEventSequenceState OnIncrementalStateCleared() {
+  // Returns a successor for use after packet loss. Persistent state (pid/tid)
+  // is preserved; delta-encoded state (running reference timestamps,
+  // incremental counter values, the timestamps_valid flag) is reset because
+  // those values are tied to the run of packets that we just lost.
+  TrackEventSequenceState OnPacketLoss() const {
     return TrackEventSequenceState(persistent_state_);
   }
 
-  void OnPacketLoss() { timestamps_valid_ = false; }
+  TrackEventSequenceState OnIncrementalStateCleared() const {
+    return TrackEventSequenceState(persistent_state_);
+  }
 
   bool pid_and_tid_valid() const { return persistent_state_.pid_and_tid_valid; }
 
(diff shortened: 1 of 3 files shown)


src/trace_processor/importers/proto/track_event_sequence_state.h
```diff
@@ -32,11 +32,17 @@ class TrackEventSequenceState {
...
+  // Returns a successor for use after packet loss. Persistent state (pid/tid)
+  // is preserved; delta-encoded state (running reference timestamps,
+  // incremental counter values, the timestamps_valid flag) is reset because
+  // those values are tied to the run of packets that we just lost.
+  TrackEventSequenceState OnPacketLoss() const {
     return TrackEventSequenceState(persistent_state_);
   }
 
-  void OnPacketLoss() { timestamps_valid_ = false; }
+  TrackEventSequenceState OnIncrementalStateCleared() {
```
sashwinbalaji: nit this also const ?

sashwinbalaji (approved):

PR #5595, opened 2026-04-25
ui: GpuCompute plugin: performance improvements and bug fixes

dreveman:
This PR addresses rendering performance issues and correctness bugs in the GPU Compute plugin that become apparent with large traces containing many kernel launches (10s of thousands).

  Performance fixes

  - Fix O(n²) rendering in toolbar dropdown — The kernel selector rebuilt thousands of <option> vnodes on every Mithril redraw. Memoize the component to skip vdom diffing
  when the options list and selected value haven't changed.
  - Memoize kernel selector to skip unchanged redraws — Further reduces unnecessary vdom work in the toolbar by tracking previous state and short-circuiting onbeforeupdate.
  - Skip summary table redraw when nothing changed — Add a dirty flag to KernelSummarySection so the table skips vdom diffing unless data or sort state actually changed.
  - Paginate summary table — Limit the summary table to 100 rows per page with prev/next navigation, preventing the browser from rendering thousands of table rows at once.
  - Deduplicate kernel list and summary queries — fetchKernelLaunchList and fetchKernelSummaryRows both scanned all compute slices. The launch list was a strict subset.
  Remove it and call fetchKernelSummaryRows once at trace load, deriving the dropdown options from the result and passing pre-fetched rows to the summary component.

  Bug fixes

  - Fix details tab for kernels without compute counters — The details query filtered counters through gpu_counter_group.group_id = 6 in the WHERE clause, which discarded
  the entire row when non-compute counters (e.g. Power, Temperature) overlapped in time with a kernel slice. This left the query with zero rows, making the kernel appear to
   have no data. Move the compute counter group filter into the LEFT JOIN condition so non-compute counters simply don't match the join, preserving launch args for the
  Launch Statistics section. Also replace the throw when no metric sections are visible with a user-facing message.
  - Eliminate tab flicker when selecting compute kernels — Three issues caused visible flicker: (1) the tab switch waited for an async metric query before calling
  showTab(), letting the Current Selection tab appear; (2) the async result could fall back to summary for known kernels without compute counters; (3) render() is called
  multiple times per Mithril redraw cycle and the deselection block briefly set activeInfoTab to summary between calls. Fix by using the known kernel ID set for synchronous
   tab switching, guarding the summary fallback, removing the eager summary switch from deselection, and guarding setSliceId() against stale async completions.

### DIFF
--- a/ui/src/plugins/com.meta.GpuCompute/summary.ts
+++ b/ui/src/plugins/com.meta.GpuCompute/summary.ts
@@ -33,7 +33,7 @@ import type {GpuComputeContext} from './index';
 import {adjustSeconds} from './humanize';
 
 // Per-kernel row returned by {@link fetchKernelSummaryRows}.
-type SummaryRow = {
+export type SummaryRow = {
   id: number;
   demangledName: string;
   durationNSecNum: number | string | null;
@@ -178,6 +178,7 @@ export interface SummarySectionAttrs extends m.Attributes {
   engine: Engine;
   sliceId?: number;
   openSliceInDetail?: (sliceId: number) => void;
+  prefetchedRows?: SummaryRow[];
 }
 
 // Column keys that the table can be sorted by.
@@ -266,7 +267,9 @@ export const KernelSummarySection: m.Component<
     state.sortDescending = false;
     state.pageOffset = 0;
 
-    const rows = await fetchKernelSummaryRows(attrs.ctx, attrs.engine);
+    const rows =
+      attrs.prefetchedRows ??
+      (await fetchKernelSummaryRows(attrs.ctx, attrs.engine));
 
     // Build launch-order map so the ID column shows 0, 1, 2, …
     rows.forEach((opt, zeroBasedIndex) =>
--- a/ui/src/plugins/com.meta.GpuCompute/summary.ts
+++ b/ui/src/plugins/com.meta.GpuCompute/summary.ts
@@ -300,8 +300,6 @@ export const KernelSummarySection: m.Component<
       rows.map((r) => Number(r.registersPerThread)),
     );
     state.maxGridSize = finiteMax(rows.map((r) => Number(r.gridSize)));
-
-    m.redraw();
   },
 
   view({state, attrs}) {
@@ -355,7 +353,6 @@ export const KernelSummarySection: m.Component<
         state.sortDescending = true;
       }
       state.pageOffset = 0;
-      m.redraw();
     };
 
     // Up/down arrow indicator for the active sort column
@@ -491,8 +488,6 @@ export const KernelSummarySection: m.Component<
             disabled: state.pageOffset === 0,
             onclick: () => {
               state.pageOffset = Math.max(0, state.pageOffset - PAGE_SIZE);
-
-              m.redraw();
             },
           }),
           m(
@@ -507,8 +502,6 @@ export const KernelSummarySection: m.Component<
                 state.pageOffset + PAGE_SIZE,
                 sortedRows.length - PAGE_SIZE,
               );
-
-              m.redraw();
             },
           }),
         ]),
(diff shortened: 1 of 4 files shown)


LalitMaganti:
Given you are here, can I ask you please change the OWNERS file in this folde rto OWNERS.github. Right now this is breaking our internal merge (I was planning on sending a CL myself but if you can just do it, that would save me the trouble :))

dreveman:
> Given you are here, can I ask you please change the OWNERS file in this folde rto OWNERS.github. Right now this is breaking our internal merge (I was planning on sending a CL myself but if you can just do it, that would save me the trouble :))

Done

default.profraw
stevegolton, line 1: Did you mean to add this file?

ui/src/plugins/com.meta.GpuCompute/toolbar.ts
```diff
@@ -28,6 +28,58 @@ import {Switch} from '../../widgets/switch';
...
+  onChange: (id: number | undefined) => void;
+}
+
+class KernelSelect implements m.ClassComponent<KernelSelectAttrs> {
+  private prevValue = '';
+  private prevOptionsLength = -1;
+  private prevFirstId = -1;
+  private prevLastId = -1;
+
+  onbeforeupdate({attrs}: m.CVnode<KernelSelectAttrs>): boolean {
```
stevegolton: onbeforeupdate receives two arguments: the current vnode and the previous vnode. You could probably avoid storing the previous state and use this old version as long as you can ensure the old and new attrs don't point to the same array reference (e.g. the array is immutable - which it looks like it is). Could add readonly to its type at the source to enforce this.

If the options list is immutable, you can also check for equality using a simple reference check.

`new.attrs.options === old.attrs.options`

ui/src/plugins/com.meta.GpuCompute/summary.ts
```diff
@@ -256,6 +257,17 @@ export const KernelSummarySection: m.Component<
   SummarySectionAttrs,
   SummaryState
 > = {
+  onbeforeupdate({state}) {
```
stevegolton: I'm a little wary of this. I usually try to follow [this advice](https://mithril.js.org/lifecycle-methods.html#avoid-premature-optimizations) in regards to onbeforeupdate.

In particular since adding pagination I suspect this can be avoided.

ui/src/plugins/com.meta.GpuCompute/summary.ts
```diff
@@ -346,6 +367,8 @@ export const KernelSummarySection: m.Component<
         state.sortKey = key;
         state.sortDescending = true;
       }
+      state.pageOffset = 0;
+      state.dirty = true;
       m.redraw();
```
stevegolton: I think you can omit all of the m.redraw() calls across this entire plugin - mithril autoredraw should handle it.

stevegolton (changes requested):
Left a few comments but generally looks fine. I'd be wary of using onbeforeupdate as it can lead to mysterious bugs when things stop updating properly, though for tactical usage in the <select> + <options> list with thousands of elements I think it makes sense.

Also I did the owners file rename in #5602 to avoid blocking while we iterate on this PR.

ui/src/plugins/com.meta.GpuCompute/summary.ts
dreveman: Done

ui/src/plugins/com.meta.GpuCompute/summary.ts
dreveman: Yep. I just dropped the commit that did this from latest version of this change.

ui/src/plugins/com.meta.GpuCompute/toolbar.ts
dreveman: Done

default.profraw
dreveman: nope. good call. removed

stevegolton (approved):
LGTM

PR #5596, opened 2026-04-25
gpu: Add structured compute kernel encoding and launch parameters

dreveman:
Add structured protobuf encoding for GPU compute kernel identity and launch parameters, replacing the flat
string-based ExtraData key/value pairs with typed fields. This improves encoding efficiency, enables interning
for repeated values, and provides a standard schema for GPGPU compute profiling across CUDA, HIP, OpenCL, and
other APIs.

Proto changes

- InternedComputeKernel — new interned message for kernel identity (name, demangled name, architecture),
referenced by kernel_iid on GpuRenderStageEvent
- ComputeKernelLaunch — structured grid/workgroup dimensions via Dim3, plus extensible typed args
- ExtraComputeArg — typed key/value arguments (int, uint, double, string) with interning support for both keys
(InternedComputeArgName) and string values
- name_iid on GpuRenderStageEvent — interned event name via EventName, grouped into oneof name_field with the
existing name string field
- InternedComputeArgName — dedicated interned type for compute arg names, with its own compute_arg_names field
(49) in InternedData

Trace processor changes

- Parse InternedComputeKernel fields into kernel_name, kernel_demangled_name, arch args
- Parse ComputeKernelLaunch grid/workgroup dimensions into structured args (launch.grid_size.x,
launch.workgroup_size.x, etc.)
- Parse ExtraComputeArg from both kernel and launch into typed slice args
- Resolve name_iid via interned EventName lookup

UI changes

- Update GpuCompute plugin to use new structured arg names (launch.grid_size.x instead of launch__grid_size_x,
workgroup_size instead of block_size)


src/trace_processor/importers/proto/gpu_event_parser.cc
```diff
@@ -844,6 +913,72 @@ void GpuEventParser::ParseGpuRenderStageEvent(
             }
           }
 
+          if (event.has_kernel_iid()) {
+            auto kernel_iid = static_cast<size_t>(event.kernel_iid());
```
LalitMaganti: nit: would be nicer to pull this out into separate functions. Here and below.

protos/perfetto/trace/gpu/gpu_render_stage_event.proto
```diff
@@ -18,7 +18,44 @@ syntax = "proto2";
 
 package perfetto.protos;
 
-// next id: 19
+// 3D unsigned integer vector for grid and workgroup dimensions.
+message Dim3 {
```
LalitMaganti: Here and belwo: please avoid making these top level fields. This will all appear in the "perfetto.protos namesapce". Put this inside the render stage event instead.

dreveman: Done

src/trace_processor/importers/proto/gpu_event_parser.cc
dreveman: Done

dreveman:
@LalitMaganti, I'm been looking at workloads with 10s of thousands of kernel launches, which is not unlikely for AI training. The same kernel is frequently launched for these workloads so being able to intern everything that is guaranteed to be constant for the kernel as very useful. More efficient encoding of everything that can change between launches helps too.

protos/perfetto/trace/interned_data/interned_data.proto
```diff
@@ -103,6 +103,9 @@ message InternedData {
   // Description of a GPU hardware queue or render stage.
   repeated InternedGpuRenderStageSpecification gpu_specifications = 24;
 
+  // Description of a compute kernel (name, demangled name, architecture).
```
LalitMaganti: I'm getting a bit uncomfortable by how many things we are dumping here. Just like TrackEvent extension I think we need to start thinking about splitting this up into extensions and interning there.

In fact, I'm not sure if at this point we really need to use InternedData at all: from [1], I think extensions *should* just work. Can we please switch to doing this?

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/importers/proto/proto_trace_reader.cc;l=642;drc=c17b46567656b9909581440163d197847bccb47f;bpv=0;bpt=1

dreveman: Makes sense. Done.

dreveman:
@LalitMaganti would be nice to get this merged before a v55 release goes out so v55 can be used to produce traces that work in the GpuCompute UI plugin. Let me know if there's anything else you like me to address here.

LalitMaganti (approved):

Issue #5591, opened 2026-04-25
v54.0 SDK example fails to build, linking error


EngineerSmith:
**TLDR**: SDK (cpp) v54.0 fails to link in example and my code. v53.0's repo's SDK same link error. v50.1's SDK complies without error for example and my code

---

I've been trying to using the (cpp) SDK provided in the latest release, v54.0 - but whenever I tried to build [my code](https://github.com/EngineerSmith/perfetto-lua/tree/75f1642664111a0b3666b0e7d57cece89882c821), it always failed to link:
```
perfetto_lua.obj : error LNK2019: unresolved external symbol "public: static class perfetto::internal::DataSourceType &
 __cdecl perfetto::DataSourceHelper<class perfetto::internal::TrackEventDataSource,struct perfetto::internal::TrackEven
tDataSourceTraits>::type(void)" (?type@?$DataSourceHelper@VTrackEventDataSource@internal@perfetto@@UTrackEventDataSourc
eTraits@23@@perfetto@@SAAEAVDataSourceType@internal@2@XZ) referenced in function "public: static __cdecl `public: stati
c bool __cdecl perfetto::internal::TrackEventDataSource::AddRegistry(class perfetto::internal::TrackEventCategoryRegist
ry const *)'::`5'::<lambda_2>::<lambda_invoker_cdecl>(class perfetto::protos::gen::DataSourceDescriptor const &)" (?<la
mbda_invoker_cdecl>@<lambda_2>@?4??AddRegistry@TrackEventDataSource@internal@perfetto@@SA_NPEBVTrackEventCategoryRegist
ry@45@@Z@SA@AEBVDataSourceDescriptor@gen@protos@5@@Z) [C:\git\perfetto-lua\build\perfetto-lua.vcxproj]
C:\git\perfetto-lua\build\Release\perfetto-lua.dll : fatal error LNK1120: 1 unresolved externals [C:\git\perfetto-lua\b
uild\perfetto-lua.vcxproj]
```
I've traced it down to being the `PERFETTO_TRACK_EVENT_STATIC_STORAGE` macro in [my code](https://github.com/EngineerSmith/perfetto-lua/blob/75f1642664111a0b3666b0e7d57cece89882c821/perfetto_categories.cpp#L5), despite trying to recreate the example code base.

The next day, I've tried to build the repo's SDK example with the same SDK provided in release 54.0, and got the same issue (Note, there was multiple linking errors, I'm picking the first):
```
[...]

PR #5599, opened 2026-04-26
tracing: fix MSVC LNK2019 in DataSourceHelper::type() (#5591)

LalitMaganti:
Move the primary-template definition of `DataSourceHelper::type()` out of
the class body so it is not implicitly inline.

`DataSourceHelper::type()` was added in 06486b8b6010 ("Fix component build
for client library on Windows", 2023-05-05) as a workaround for an MSVC
C2086 redefinition bug affecting an earlier static-data-member design.
That commit defined `type()` inline inside the class body
([dcl.fct.spec]/4 then makes it implicitly inline). MSVC propagates that
inline attribute to every explicit specialization of `type()` defined
out-of-line by `PERFETTO_DEFINE_DATA_SOURCE_STATIC_MEMBERS` in a `.cc`,
emits a C4506 ("no definition for inline function"), and silently drops
the specialization body. Any TU that does not see the body but calls
`Helper::type()` -- transitively via `DataSource<...>::Register<>()`,
`Trace()`, etc. -- is then left with an unresolved external symbol.

The bug stayed dormant until a5f062d78a60 ("[API] Single data source for
all track events", 2025-06-24, in v52.0+) made `TrackEventDataSource`
a concrete class with a `PERFETTO_DECLARE_DATA_SOURCE_STATIC_MEMBERS`
visible in the public header, so user TUs that invoke
`PERFETTO_TRACK_EVENT_STATIC_STORAGE()` started instantiating
`Register<>` and producing LNK2019:

```
  example.cc.obj : error LNK2019: unresolved external symbol
  "public: static class perfetto::internal::DataSourceType & __cdecl
   perfetto::DataSourceHelper<class perfetto::internal::TrackEventDataSource,
                              struct perfetto::internal::TrackEventDataSourceTraits>::type(void)"
  referenced in function "perfetto::DataSource<...>::Register<>(...)".
```

GCC and Clang are unaffected: they treat the out-of-line explicit
specialization as a regular non-inline function and emit it normally.

Fix: declare `type()` in the class body and define it out-of-line. The
primary template's member is then a non-inline function template, and
explicit specializations are unambiguously non-inline regular functions
on every compiler. No call sites change; the only file touched is
`include/perfetto/tracing/data_source.h`.

Verified on Windows / MSVC v143 (14.50.35717) for x64 Debug and Release
builds of `examples/sdk` and the v54.0 amalgamated SDK. GCC/Clang
unaffected because the change is a no-op semantic refactor for them.

Bug: #5591

### DIFF
--- a/include/perfetto/tracing/data_source.h
+++ b/include/perfetto/tracing/data_source.h
@@ -237,21 +237,29 @@ struct DefaultDataSourceTraits {
   static bool ClearIncrementalState(IncrementalStateType*) { return false; }
 };
 
-// Holds the type for a DataSource. Accessed by the static Trace() method
-// fastpaths. This allows redefinitions under a component where a component
-// specific export macro is used.
-// Due to C2086 (redefinition) error on MSVC/clang-cl, internal::DataSourceType
-// can't be a static data member. To avoid explicit specialization after
-// instantiation error, type() needs to be in a template helper class that's
-// instantiated independently from DataSource. See b/280777748.
+// Holds the DataSourceType for a DataSource, accessed by Trace()'s fastpaths.
+// Lives on a separate helper (not as a static member on DataSource) so
+// per-component export macros can redefine it, and because MSVC/clang-cl reject
+// the static-data-member design with a redefinition error (C2086) and an
+// explicit-specialization-after-instantiation error. See b/280777748.
+//
+// type() must be defined out-of-line: defining it in the class body makes it
+// implicitly inline ([dcl.inline]/4), which MSVC then propagates to every
+// explicit specialization from PERFETTO_DEFINE_DATA_SOURCE_STATIC_MEMBERS and
+// silently drops the body (C4506), so callers fail to link with an unresolved
+// external symbol (LNK2019). GCC/Clang are unaffected.
 template <typename DerivedDataSource,
           typename DataSourceTraits = DefaultDataSourceTraits>
 struct DataSourceHelper {
-  static internal::DataSourceType& type() {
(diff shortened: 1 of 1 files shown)


include/perfetto/tracing/data_source.h
```diff
@@ -247,12 +247,28 @@ struct DefaultDataSourceTraits {
 template <typename DerivedDataSource,
           typename DataSourceTraits = DefaultDataSourceTraits>
 struct DataSourceHelper {
-  static internal::DataSourceType& type() {
-    static perfetto::internal::DataSourceType type_;
-    return type_;
-  }
+  // NOTE: type() is intentionally declared here and defined out-of-line below.
+  // Defining it inside the class body would make it implicitly inline
+  // ([dcl.fct.spec]/4), and MSVC propagates that inline attribute to every
```
sashwinbalaji: https://eel.is/c++draft/dcl.fct.spec#4 seems to about something else. Maybe https://eel.is/c++draft/dcl.inline#4 ?

include/perfetto/tracing/data_source.h
sashwinbalaji: nit: can we combine this and below comment.

sashwinbalaji (approved):

include/perfetto/tracing/data_source.h
LalitMaganti: Done

include/perfetto/tracing/data_source.h
LalitMaganti: Done.

PR #5601, opened 2026-04-27
Fix perfetto build with C++23 (for Chromium)

victorvianna:
6fca04fc6689d55094dc11135d322f3ea018db61 broke the Chromium build because WorkItem needs to be a complete type by the time ProtoToArgsParser's constructor and destructor are instantiated.


victorvianna:
@LalitMaganti

LalitMaganti (approved):

PR #5613, opened 2026-04-27
tp: amalgamated SQL via cpp_blob_header primitive

LalitMaganti:
Migrate `perfetto_amalgamated_sql_header` to the generic
`perfetto_cpp_blob_header` build primitive. The amalgamated SQL header
is now a single `constexpr std::array<uint8_t, N>` blob in the SqlBundle
wire format (u32 count + per-entry varint-prefixed path/sql), parsed at
runtime by `src/trace_processor/util/sql_bundle.h`.

This removes the SQL-specific translator handlers + Bazel rule, with
only a small chrome-glob exception in the Soong sql_source_set handler
(mirroring the precedent of `target_overrides` in tools/gen_bazel; the
chrome stdlib BUILD.gn is autorolled from Chromium so the glob can't
live there).

Build wiring changes:
- gen_amalgamated_sql.py: rewritten to pack the bundle and hand off to
  `cpp_blob_emitter.emit_array`. CLI now matches the cpp_blob_header
  convention (`--output`, `--gen-dir`, positional inputs).
- perfetto_cpp_blob_header.gni: new `extra_inputs` invoker arg for
  scripts that import other Python modules (threaded into Soong
  `tool_files`; excluded from Bazel genrule srcs since Bazel wires
  Python imports via py_binary deps).
- gen_sql_source_set (Bazel): filegroups now include dep filegroups so
  Bazel resolves the SQL set transitively.
- gen_android_bp: new `create_sql_source_set_module` emits a Soong
  filegroup per sql_source_set (chrome glob hardcoded as a known
  exception).
- Deleted: `gen_cc_amalgamated_sql` (gen_bazel),
  `create_amalgamated_sql_module` (gen_android_bp),
  `perfetto_cc_amalgamated_sql` (bazel/rules.bzl).


LalitMaganti:
@sashwinbalaji FYI

aMayzner (approved):

PR #5625, opened 2026-04-28
ui: Fix bigtrace TypeScript build errors

zezeozue:
Update bigtrace code to use getCommands() instead of the removed .commands property, and move fuzzyFilterCommands() into the omnibox component to match the main UI pattern.


LalitMaganti:
Already fixed by Steve in https://github.com/google/perfetto/pull/5624

PR #5634, opened 2026-04-28
build: re-run gen_android_bp and gen_bazel

dreveman:


dreveman:
@LalitMaganti Bazel CI run seems to be failing in ToT after my compute kernel change. Not sure how that made it through CI but I'm guessing the CI run was a bit old and the patch still applied.

LalitMaganti (approved):

LalitMaganti:
It was a race with my change

PR #5638, opened 2026-04-29
ui: Fix SQL syntax error for dotted launch arg names

dreveman:
EXTRACT_ARG metric IDs like "launch.workgroup_size.x" were used directly as SQL column aliases, but unquoted identifiers cannot contain dots. Replace dots with underscores in the AS alias and use the same transformation when reading columns back from the query result.

### DIFF
--- a/ui/src/plugins/com.meta.GpuCompute/details.ts
+++ b/ui/src/plugins/com.meta.GpuCompute/details.ts
@@ -218,6 +218,10 @@ type QueryCapable = {
 // SQL query building
 // =============================================================================
 
+function sqlAlias(id: string): string {
+  return id.replace(/\./g, '_');
+}
+
 // Launch metrics required for display and toolbar regardless of which
 // section plugins are loaded.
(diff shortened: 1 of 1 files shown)


LalitMaganti (approved):

ui/src/plugins/com.meta.GpuCompute/details.ts
```diff
@@ -218,6 +218,10 @@ type QueryCapable = {
 // SQL query building
 // =============================================================================
 
+function sqlAlias(id: string): string {
+  return id.replace(/\./g, '_');
```
stevegolton, line 222: You could also wrap in quotes if you wanted to keep the period.

stevegolton (approved):

ui/src/plugins/com.meta.GpuCompute/details.ts
dreveman: Ack. Ill keep this but good to know

PR #5643, opened 2026-04-29
perfetto: fix bogus chunks_overwritten in TraceBufferV2

LalitMaganti:
DeleteNextChunksFor() unconditionally bumped chunks_overwritten /
bytes_overwritten for any non-padding chunk it walked, while data_loss
was correctly gated on has_cleared_unconsumed_fragments. The two should
move together: a chunk that the reader already drained (payload_avail
== 0) but hadn't yet "stepped past" via the follow-up
ReadNextTracePacket() call is still non-padding, and clearing it on a
subsequent wrap is not a real overwrite.

Move the stat increments inside the same gate as data_loss so they
only fire when there were actually unread fragments at clear time.
This matches v1 (trace_buffer_v1.cc only counts when num_fragments_read
< num_fragments).

Also adds NoOverwriteCountIfReaderCatchesUp, which mirrors the
existing NoDataLossIfReaderCatchesUp pattern but asserts on the stats
the latter never checked.

### DIFF
--- a/src/tracing/service/trace_buffer_v2.cc
+++ b/src/tracing/service/trace_buffer_v2.cc
@@ -971,18 +971,21 @@ void TraceBufferV2::DeleteNextChunksFor(size_t bytes_to_clear) {
       has_cleared_unconsumed_fragments = true;
     }
 
-    // In future this branch should become "&& !protovm_has_consumed_packet"
-    // We shouldn't report a data loss if ProtoVM merged the outgoing packet.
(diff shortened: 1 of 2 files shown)


primiano:
I believe https://github.com/google/perfetto/pull/5676 is a bit more reliable and correct. your version doesn't deal with the case of overwriting a sequence of out-of-order chunks, as it only counts the one where we started from

PR #5644, opened 2026-04-29
perfetto: fix false data_loss on re-admit in TraceBufferV2

LalitMaganti:
ChunkSeqReader's ctor checks iter_->chunk_id against
last_chunk_consumed.chunk_id + 1 to detect a gap in the sequence. But
when a previously-evicted incomplete chunk is re-admitted via the
re-admit path in CopyChunkUntrusted (same chunk_id, was_incomplete=true,
strictly more payload), the new iter_ lands on the same chunk_id as
last_consumed — so the +1 check fires spuriously and seq.data_loss is
set even though the re-admit recovered the deferred fragments without
losing any data.

The next packet delivered for that sequence then carries
previous_packet_dropped=true, which trace_processor surfaces as
traced_buf_sequence_packet_loss. This bites real traces with heavy
scraping/recommit traffic (chunks_rewritten dominated) and can show up
as nonzero packet loss even on under-utilized buffers with no
trace_writer_packet_loss and no overwrites.

Mirror the re-admit acceptance condition in the gap check: skip data
loss when iter_->chunk_id == last.chunk_id and last.was_incomplete.
This is the symmetric counterpart of the re-admit path; any other
equality case is impossible because CopyChunkUntrusted would have
discarded the chunk before it reached the buffer.

Adds RescrapeAfterEviction_NoFalseDataLoss alongside the existing
rescrape tests, which exercised the same path but didn't assert on
previous_packet_dropped.

### DIFF
--- a/src/tracing/service/trace_buffer_v2.cc
+++ b/src/tracing/service/trace_buffer_v2.cc
@@ -289,9 +289,18 @@ ChunkSeqReader::ChunkSeqReader(TraceBufferV2* buf,
                            : 0,
                        iter_->chunk_id, end_->chunk_id);
 
-  if (seq_->last_chunk_consumed.has_value() &&
-      iter_->chunk_id != seq_->last_chunk_consumed->chunk_id + 1) {
-    seq_->data_loss = true;
+  if (seq_->last_chunk_consumed.has_value()) {
+    const auto& last = *seq_->last_chunk_consumed;
+    // Re-admit: an incomplete chunk that was evicted has been re-committed
+    // with strictly more payload. The chunk_id is unchanged, so the +1
+    // successor check below would fire spuriously even though the re-admit
+    // recovered the deferred fragments without losing data. Keep this
+    // predicate in sync with the re-admit acceptance check in
+    // TraceBufferV2::CopyChunkUntrusted.
+    bool readmit = last.was_incomplete && iter_->chunk_id == last.chunk_id;
+    if (!readmit && iter_->chunk_id != last.chunk_id + 1) {
+      seq_->data_loss = true;
+    }
   }
 }
 
@@ -754,7 +763,8 @@ void TraceBufferV2::CopyChunkUntrusted(
   // Don't allow re-commit of chunks that have been consumed already, unless
   // the chunk was evicted while incomplete (scraped) and the new commit has
   // strictly more payload. In that case we re-admit it so the previously-
-  // dropped fragments can be recovered.
+  // dropped fragments can be recovered. Keep this acceptance condition in
+  // sync with the re-admit branch of the gap check in ChunkSeqReader's ctor.
   //
   // When re-admitting, |previously_consumed_payload| records how many payload
   // bytes were already consumed before eviction, so we can skip them below.
--- a/src/tracing/service/trace_buffer_v2_unittest.cc
+++ b/src/tracing/service/trace_buffer_v2_unittest.cc
@@ -2441,13 +2441,20 @@ TEST_F(TraceBufferV2Test, RescrapeAfterEviction_FullyRead) {
       .PadTo(512)
       .CopyIntoTraceBuffer(/*chunk_complete=*/false);
 
-  // Drain and verify: 'b' recovered, 'a' not duplicated.
+  // Drain and verify: 'b' recovered, 'a' not duplicated. The recovered 'b'
+  // must not be flagged with previous_packet_dropped — the re-admit landed
+  // on the same chunk_id as last_consumed, and the gap check in
+  // ChunkSeqReader must treat that as gapless rather than firing spuriously.
   trace_buffer()->BeginRead();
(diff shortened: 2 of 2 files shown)


src/tracing/service/trace_buffer_v2.cc
```diff
@@ -289,9 +289,17 @@ ChunkSeqReader::ChunkSeqReader(TraceBufferV2* buf,
...
-      iter_->chunk_id != seq_->last_chunk_consumed->chunk_id + 1) {
-    seq_->data_loss = true;
+  if (seq_->last_chunk_consumed.has_value()) {
+    const auto& last = *seq_->last_chunk_consumed;
+    // Re-admit: an incomplete chunk that was evicted has been re-committed
+    // with strictly more payload (see the re-admit path in
+    // CopyChunkUntrusted). The chunk_id is unchanged, so the +1 successor
+    // check below would fire spuriously even though the re-admit recovered
+    // the deferred fragments without losing data.
+    bool readmit = last.was_incomplete && iter_->chunk_id == last.chunk_id;
```
primiano, line 300: rather than replicating here, can we have a function that we call in both places so we avoid further out of sync in future?

src/tracing/service/trace_buffer_v2_unittest.cc
```diff
@@ -2586,6 +2586,69 @@ TEST_F(TraceBufferV2Test, RescrapeAfterEviction_CompleteCommit) {
...
+// recovered the deferred fragments cleanly.
+TEST_F(TraceBufferV2Test, RescrapeAfterEviction_NoFalseDataLoss) {
+  ResetBuffer(4096);
+  SuppressClientDchecksForTesting();
+
+  // Initial scrape of chunk 0 for sequence (P=1, W=1). Scraping drops the
+  // trailing kFragBegin 'b', so only 'a' is visible.
+  CreateChunk(ProducerID(1), WriterID(1), ChunkID(0))
+      .AddPacket(20, 'a')
+      .AddPacket(10, 'b', kContOnNextChunk)
```
primiano: remove the kCont here, it's incorrect (given what you do below) and I dont' think it's necessary

primiano (commented):
```
diff --git a/src/tracing/service/trace_buffer_v2.cc b/src/tracing/service/trace_buffer_v2.cc
index 5a38c6ff45..e06de9e05c 100644
--- a/src/tracing/service/trace_buffer_v2.cc
+++ b/src/tracing/service/trace_buffer_v2.cc
@@ -79,6 +79,15 @@ int ChunkIdCompare(ChunkID a, ChunkID b) {
   return (static_cast<int32_t>(a - b) < 0) ? -1 : 1;
 }
 
+bool IsGapless(const internal::SequenceState& seq, ChunkID chunk_id) {
+  if (!seq.last_chunk_consumed.has_value())
+    return true;
+
+  const auto& last_chunk = *seq.last_chunk_consumed;
+  return chunk_id == last_chunk.chunk_id + 1 ||
+         (chunk_id == last_chunk.chunk_id && last_chunk.was_incomplete);
+}
+
 // The number of the latest empty SequenceState(s) to keep around.
 constexpr size_t kKeepLastEmptySeq = 1024;
 
@@ -754,19 +763,20 @@ void TraceBufferV2::CopyChunkUntrusted(
   // Don't allow re-commit of chunks that have been consumed already, unless
   // the chunk was evicted while incomplete (scraped) and the new commit has
   // strictly more payload. In that case we re-admit it so the previously-
-  // dropped fragments can be recovered.
+  // dropped fragmentts can be recovered.
   //
   // When re-admitting, |previously_consumed_payload| records how many payload
   // bytes were already consumed before eviction, so we can skip them below.
   uint16_t previously_consumed_payload = 0;
   if (seq.last_chunk_consumed.has_value()) {
-    int cmp = ChunkIdCompare(chunk_id, seq.last_chunk_consumed->chunk_id);
-    if (cmp == 0 && seq.last_chunk_consumed->was_incomplete &&
+    if (IsGapless(seq, chunk_id) &&
         seq.last_chunk_consumed->payload_size < all_frags_size) {
       previously_consumed_payload = seq.last_chunk_consumed->payload_size;
       TRACE_BUFFER_V2_DLOG("  Re-admitting chunk %u (prev_payload=%u)",
                            chunk_id, previously_consumed_payload);
-    } else if (cmp <= 0) {
+
+    } else if (ChunkIdCompare(chunk_id, seq.last_chunk_consumed->chunk_id) <=
+               0) {
       // Older or same chunk with no new data to recover.
       stats_.set_chunks_discarded(stats_.chunks_discarded() + 1);
       PERFETTO_DCHECK(suppress_client_dchecks_for_testing_);

```

src/tracing/service/trace_buffer_v2.cc
LalitMaganti: Discussed this offline, got worse than just having the simple code.

primiano (approved):

PR #5645, opened 2026-04-29
Monitor Render Thread blocking calls during Android CUJs

nicomazz:
Until now the blocking calls were only considering what was happening in the ui thread associated with the CUJ. After this change, we're also considering the render thread of the process reporting the CUJ.

The render thread is a single one for each process, and is identified with the thread name.

CUJ boundaries considered are left unchanged.

bug:
Change-Id: I1d68bb712eb8860436c330b4db28d2e91d0a8710

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/render_thread.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/render_thread.sql
@@ -1,0 +1,26 @@
+--
+-- Copyright 2026 The Android Open Source Project
+--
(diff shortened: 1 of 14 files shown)


src/trace_processor/perfetto_sql/stdlib/android/render_thread.sql
```diff
@@ -0,0 +1,10 @@
+CREATE PERFETTO TABLE _render_thread_per_process (
```
LalitMaganti, line 17: This needs a copyright header.

LalitMaganti (approved):

PR #5649, opened 2026-04-29
ui: search and highlight chip text in omnibox commands

LalitMaganti:
The command omnibox shows a "source" chip (e.g. plugin name) on the
left of each result but neither searched nor highlighted it. Index
both name + source so users can filter by either, compute fuzzy
highlight segments for the chip, and invert chip colours on the
highlighted row so the primary-intent chip stays visible against the
primary-coloured selection background.

FuzzyFinder grows an optional displayLookup so a caller can index
combined fields while restricting highlight segments to one of them.

### DIFF
--- a/ui/src/base/fuzzy.ts
+++ b/ui/src/base/fuzzy.ts
@@ -96,7 +96,10 @@ function camelCaseTokenize(text: string): string[] {
 // Given a query (possibly multi-word) and candidate text, compute highlight
 // segments. Each query token is first tried as a substring match, then falls
 // back to sequential character matching.
-function computeHighlightSegments(query: string, text: string): FuzzySegment[] {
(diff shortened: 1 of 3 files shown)


ui/src/base/fuzzy.ts
```diff
@@ -60,10 +60,10 @@ export class FuzzyFinder<T> {
   find(searchTerm: string): FuzzyResult<T>[] {
     if (searchTerm === '') {
       return this.items.map((item) => {
-        const normalisedTerm = this.keyLookup(item);
+        const text = this.keyLookup(item);
```
stevegolton: This file hasn't changed functionally right? Maybe restore to avoid churn.

stevegolton (approved):
LGTM % churn nit

PR #5654, opened 2026-04-29
gpu: Add GenericGpuFrequencyEvent and move GpuMemTotalEvent

dreveman:
Add GenericGpuFrequencyEvent proto in generic_kernel/ following the
same pattern as GenericKernelCpuFrequencyEvent. This provides a
stable public API for non-ftrace data sources to emit GPU frequency
data. The trace processor reuses kGpuFrequencyBlueprint so the
resulting tracks are identical to ftrace gpu_frequency events and
the UI discovers them automatically.

Move GpuMemTotalEvent from android/ to gpu/ alongside the other
GPU event protos.

Reorganize the TracePacket oneof into three sections: stable
TrackEvent API, stable generic kernel events, and unstable
context-specific fields.

Generate Rust SDK bindings for both protos and add them to the
TracePacket extension in perfetto-protos-gpu.


protos/perfetto/trace/trace_packet.proto
```diff
@@ -192,6 +193,7 @@ message TracePacket {
     TracingServiceEvent service_event = 69;
     InitialDisplayState initial_display_state = 70;
     GpuMemTotalEvent gpu_mem_total_event = 71;
+    GpuFrequencyEvent gpu_frequency_event = 129;
```
LalitMaganti: So if we're going to make this an API we need to follow the example of "generic kernel" events. That is, we have to place it in this perfetto/trace/generic_kernel, put it in a separate dedicated section of TracePacket etc.

The goal is that there are only two sets of things in Perfetto we want to stably commit to supporting in our public API: generic level protos and track event. Everything else is unstable and you depend on at your own risk.

dreveman: Makes sense. Done.

protos/perfetto/trace/trace_packet.proto
LalitMaganti: Can I ask you to do the following: 1) separate this oneof into three segments: trackevent, generic kernel and then everything else. 2) clearly mark each section with comments 3) mark the third sectio nwith the fact that these are not "public API" and only are valid in given context and you depend on it at your own risk.

LalitMaganti (approved):

protos/perfetto/trace/trace_packet.proto
dreveman: Took at stab at this in latest version PR. Let me know if it's not what you had in mind.

protos/perfetto/trace/android/gpu_mem_event.proto
```diff
@@ -14,12 +14,7 @@
  * limitations under the License.
  */
 
+// Moved to protos/perfetto/trace/gpu/gpu_mem_event.proto.
+// This file re-exports the message for backwards compatibility.
```
LalitMaganti: This seems unideal. Can we avoid moving this proto please?

LalitMaganti: Or if we move it, let's just it and not worry about backcompat.

protos/perfetto/trace/trace_packet.proto
```diff
@@ -150,16 +151,41 @@ message TracePacket {
...
 
   oneof data {
+    //
+    // == STABLE PUBLIC API: TrackEvent ==
+    //
+    // TrackEvent is the stable, extensible API for recording trace events.
+    // See docs/instrumentation/track-events.md for details.
+    TrackEvent track_event = 11;
+    TrackDescriptor track_descriptor = 60;
+    TrackEventRangeOfInterest track_event_range_of_interest = 90;
```
LalitMaganti: This is not part of the stable API. Please remove from here.

protos/perfetto/trace/trace_packet.proto
```diff
@@ -150,16 +151,41 @@ message TracePacket {
...
+
+    //
+    // == GENERIC KERNEL EVENTS ==
```
LalitMaganti: We should makr this as stable public also.

dreveman: Done

protos/perfetto/trace/trace_packet.proto
dreveman: Done

protos/perfetto/trace/android/gpu_mem_event.proto
dreveman: Removed backcompat

LalitMaganti (approved):

PR #5656, opened 2026-04-29
tp: stdlib: Fix lock_name in android_monitor_contention

ivan-chong:
Previously, `monitor contention*` events with a child `Lock contention*` slice were not properly showing a `lock_name` in `android_monitor_contention` as the query uses lag and lead to look for `*_lock_acquire` and `*_lock_held` slices respectively. This meant that the leading slice would be seen to be the child `Lock contention*` as opposed to the adjacent `*_lock_held`.

This PR changes the logic to fetch both the next two slice names with an extra LEAD() statement, and uses CASE to account for both scenarios.

Bug: 507922080

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/monitor_contention.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/monitor_contention.sql
@@ -205,7 +205,8 @@ WITH
       s.ts,
       s.name,
       lag(s.name) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS prev_name,
-      lead(s.name) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS next_name
+      lead(s.name, 1) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS next_name_1,
+      lead(s.name, 2) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS next_name_2
     FROM slice AS s
     WHERE
       s.track_id IN (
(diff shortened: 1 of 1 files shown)


src/trace_processor/perfetto_sql/stdlib/android/monitor_contention.sql
```diff
@@ -208,7 +208,8 @@ WITH
       lead(s.name) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS next_name
```
LalitMaganti: Can you use a filter clause on the window function instead of doing the not glob below? Not glob is a *very* expensive operation and one you should try to avoid as much as possible.

ivan-chong: I'm not sure if it's possible to use a filter on lag/lead. I instead changed the logic to calculate the next two names and updated the CASE to check both these scenarios., to avoid doing negative checks and also replaced GLOB with substr() and equality checks, wdyt?

LalitMaganti (approved):

gignat-dev (commented):
From the definition of ` _valid_android_monitor_contention` table I can see that a 'monitor contention*' slice can only have a 'Lock contention*' child or no child.

What do you think about filtering out the 'Lock contention*' slices before applying the window functions?

Something like: 
```
   lead(s.name) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS next_name
FROM slice AS s
    WHERE
       s.name NOT GLOB 'Lock contention*'
       AND s.track_id IN ( .. )
```

ivan-chong:
> From the definition of ` _valid_android_monitor_contention` table I can see that a 'monitor contention*' slice can only have a 'Lock contention*' child or no child.
> 
> What do you think about filtering out the 'Lock contention*' slices before applying the window functions?
> 
> Something like:
> 
> ```
>    lead(s.name) OVER (PARTITION BY s.track_id ORDER BY s.ts) AS next_name
> FROM slice AS s
>     WHERE
>        s.name NOT GLOB 'Lock contention*'
>        AND s.track_id IN ( .. )
> ```

@gignat-dev Although this semantic makes more sense, we are intentionally not using a `NOT GLOB` due to the operation being expensive, as per Lalit's comment above.

gignat-dev (approved):

PR #5659, opened 2026-04-30
ui: add slice duration histogram when single selecting

LalitMaganti:
Adds a reusable distribution panel and shows a duration histogram for matching slices from the single-slice details panel.

The inline details view defaults to the selected slice track, supports switching to whole-trace scope, and links out to a full matching-slices tab.


ui/src/widgets/tree.ts
```diff
@@ -29,13 +29,17 @@ import {hasChildren} from '../base/mithril_utils';
 interface TreeAttrs {
   // Space delimited class list applied to our tree element.
   className?: string;
+  // When true, render the tree as a bordered card with a divider between
+  // every label/value row and a tinted label column. Use for stats/key-value
+  // tables where rows should look like proper cells.
+  bordered?: boolean;
```
stevegolton, line 35: This mode looks pretty broken when we have nested tree nodes:

Image

Also plz add a opt in the tree demo on the widgets page for this option for easier testing.

stevegolton (approved):
Looks great!

One suggestion - I think the args and flows sections should show above this new section to avoid surprises. Perhaps this new section should go below the details section on the left?

PR #5661, opened 2026-04-30
ui: Add dependencies and dependants to plugins page

stevegolton:
- Add list of dependencies (requires) to each plugin.
- Add list of dependants (required by) to each plugin.
- Each plugin in the list is a clickable link to that plugin.
- If a plugin is disabled but transitively enabled by another plugin it now has a badge indicating this.
- Each plugin link is colored green or gray dependeing on whether it's enabled or disabled, allowing the user to easily see which plugin is keeping this plugin enabled through a transititive dependency.

image


LalitMaganti:
Looks fantastic!

LalitMaganti (approved):

PR #5671, opened 2026-04-30
rust-sdk: Add tracing-perfetto-sdk crate

dreveman:
Add a tracing-subscriber Layer that bridges the Rust tracing
ecosystem to Perfetto via the perfetto-sdk crate. Spans become
duration slices (begin/end) and events become instant events,
with debug annotations for fields and source locations.

Uses the SDK's track event macros for proper interning,
incremental state, and integration with both in-process tracing
and the system tracing service.

Also replace symlinks to the main crate README with dedicated
README files for each crate. Each README describes the crate's
purpose, basic usage, and links to related crates.

### DIFF
--- a/contrib/rust-sdk/perfetto-protos-trace-processor/README.md
+++ b/contrib/rust-sdk/perfetto-protos-trace-processor/README.md
@@ -1,1 +1,37 @@
-../perfetto/README.md
+# perfetto-sdk-protos-trace-processor
+
+Trace processor protobuf bindings for the [Perfetto](https://perfetto.dev)
+Rust SDK.
+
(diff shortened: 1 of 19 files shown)


contrib/rust-sdk/perfetto-protos-trace-processor/README.md
LalitMaganti: Please use serve http instead of -D as that's the new API we're moving to. (Old one is still supported for backwards compatibility)

LalitMaganti (approved):

contrib/rust-sdk/perfetto-protos-trace-processor/README.md
dreveman: Done

PR #5677, opened 2026-04-30
ui: add histogram action to slice aggregation rows

LalitMaganti:
Taking the same thing we built for flamegraph and single selection but also linking it from the slice aggregator.

### DIFF
--- a/ui/src/components/aggregation_adapter.ts
+++ b/ui/src/components/aggregation_adapter.ts
@@ -25,6 +25,7 @@ import {Trace} from '../public/trace';
 import {Track} from '../public/track';
 import {UnionDataset, Dataset, DatasetSchema} from '../trace_processor/dataset';
 import {Engine} from '../trace_processor/engine';
+import {SqlValue} from '../trace_processor/query_result';
 import {EmptyState} from '../widgets/empty_state';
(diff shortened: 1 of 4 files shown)


stevegolton (commented):
I like the utility but I don't love the way this is implemented or the UX I'm afraid.

- A similar but subtly different drilldown-like option is confusing, especially as it only works for slice aggregations - perhaps we should build configurable histogram support into the datagrid itself.
- The extraPivotRowMenuItems adds yet another very specific slot to the already busy DataGrid attrs and is kinda hard to reason about. Why does it appear on every single groupby column? (I'm not sold on the fact that the drilldown appears on every col either - I think we could remove it).
- I don't love all the extra plumbing pushing extraPivotRowMenuItems through from the aggregator through to the datagrid + it looks like we do a lot of the same dataset aggregation work twice..?

I don't know how to to make fix any of these problems elegantly - aside from building the histogram into the datagrid, and adding a preset config for it in the aggregators, but that's more involved.

LalitMaganti:
I am just going to drop this in that case pending whatever other improvement we make here. I will say that I do think we need an approach which doesn't try to do everything inside datagrid and gives a flexible way for us to combine multiple different widgets together similar to how the histogram + table view works today.

PR #5680, opened 2026-04-30
ui: show slice arguments before flows

noah-CAL:
### Details
Perfetto UI's slice details panel currently renders Preceding Flows and Following Flows before the Arguments section. For slices with many preceding/following flows, this pushes arguments and debug annotations below the fold and makes the common inspection path require scrolling.

This changes the slice details panel ordering so Arguments appears before the flow sections.

This is a UI-only change, and does not change trace proto, trace processor behavior, or how arguments/flows are emitted.

### Tested
  - Manual: opened an example Perfetto trace and verified Arguments appears above Preceding/Following Flows
  - `eslint src/components/details/thread_slice_details_tab.ts`
  - `prettier --check src/components/details/thread_slice_details_tab.ts`
  - `git diff --check`

### Screenshot (After PR)
image

### DIFF
--- a/ui/src/components/details/thread_slice_details_tab.ts
+++ b/ui/src/components/details/thread_slice_details_tab.ts
@@ -323,9 +323,9 @@ export class ThreadSliceDetailsPanel implements TrackEventDetailsPanel {
     ) {
       return m(
         GridLayoutColumn,
+        args,
         precFlows,
         followingFlows,
-        args,
         distribution,
         additionalSections,
       );


LalitMaganti (approved):

stevegolton (commented):
I would be very wary of changing UI that's been stable for years at this point. People don't tend to like change.

I'm not saying this change is wrong or that this isn't an improvement, but there's a lot of momentum behind the status quo.

Edit:

A counter argument here might be that long argument lists now push the flows sections down below the fold.

We only have limited screen real estate. Trying to decide which sections are more important than others across all users is difficult.

noah-CAL:
> I would be very wary of changing UI that's been stable for years at this point. People don't tend to like change.
>
> We only have limited screen real estate. Trying to decide which sections are more important than others across all users is difficult.

Thanks Steve, I see the concern. I agree this is a tradeoff given the limited vertical space and the fact that the current ordering has been stable for a long time.

If changing the default order is too disruptive, would you be open to an alternative such as making the ordering configurable as a settings preference? Our team works with slices that have many flows, so the current ordering makes slice-local metadata hard to inspect. I'd love to solve this upstream if there's a version of the change that fits Perfetto's UI direction.

primiano:
My take on how to settle this:

Prework: compute via a bigtrace query  (A) % of slices have args, and (F) % of slices that have flows, within each trace

### If A > F (args are more frequent than flows):
Go with this PR. Showing args first makes the UX more consistent across clicks, as you tend to get the frequent thing (args) in the same place. The current state is such that args are "usually" there, but sometimes flows can push them quite far below.

### If F > A (flows are more frequent than flows):
keep as is, same rationale of above

### If A ~= F (they are roughly the same)
keep as is, as UX consistency (i.e. not breaking people's muscle memory) wins.

PR #5682, opened 2026-04-30
ui: allow 'data:' URIs in ?url=

zezeozue:
Allows embedding trace bytes (e.g. a small pprof) directly in the URL as base64, instead of hosting the file. fetch() already supports data: URIs; only the protocol assertion in TraceHttpStream blocked them.

### DIFF
--- a/ui/src/frontend/trace_url_handler.ts
+++ b/ui/src/frontend/trace_url_handler.ts
@@ -20,6 +20,7 @@ import {loadAndroidBugToolInfo} from './android_bug_tool';
 import {Route, Router} from '../core/router';
 import {taskTracker} from './task_tracker';
 import {AppImpl} from '../core/app_impl';
+import {isTrustedOrigin, saveUserTrustedOrigin} from './post_message_handler';
 
 function getCurrentTraceUrl(): undefined | string {
   const source = AppImpl.instance.trace?.traceInfo.source;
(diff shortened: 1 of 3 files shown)


LalitMaganti:
Decision on this should be Steve's call.

primiano:
My only comment on this, if we end up doing this: make sure that the dialog that asks "do you trust this origin" (based on the opener) shows up in the same way we do for postMessage. otherwise that becomes an easy attack vector (window.open(ui.perfetto.dev?data=malicious_trace) in a hidden popup)

zezeozue:
added the prompt gating

ui/src/frontend/trace_url_handler.ts
```diff
@@ -235,6 +242,29 @@ function loadTraceFromUrl(url: string) {
...
+  const originTxt = origin === 'null' ? 'An unknown origin' : origin;
+  showModal({
+    title: 'Open trace?',
+    content: m(
+      'div',
+      m('div', `${originTxt} is trying to open an embedded trace.`),
+      m('div', 'Only continue if you trust this URL.'),
+    ),
+    buttons: [
+      {text: 'No', primary: true},
```
LalitMaganti, line 267: We should add a "trust and rememebr:

zezeozue:
> My only comment on this, if we end up doing this: make sure that the dialog that asks "do you trust this origin" (based on the opener) shows up in the same way we do for postMessage. otherwise that becomes an easy attack vector (window.open(ui.perfetto.dev?data=malicious_trace) in a hidden popup)

added the prompt gating

ui/src/frontend/trace_url_handler.ts
```diff
@@ -235,6 +243,44 @@ function loadTraceFromUrl(url: string) {
   }
 }
 
+function promptDataUrlTrust(url: string) {
```
primiano, line 246: You created a function, but this is also duplicated from the one in post_message_handler

Imho this needs to be rationalized better:

- isTrustedOrigin & co at this point are better placed here 
- promptDataUrlTrust should be an exported function in here
- both places should use the same logic. otherwise this is super confusing when somebody tried to codesearch adn figure out how it works

primiano (changes requested):

PR #5686, opened 2026-05-01
tp: import Firefox profiler markers as slices

LalitMaganti:
Parse the `markers` table from preprocessed Gecko/Firefox profiler
profiles and emit them as slices in trace_processor:

- Pull `meta.categories` once at the profile level so marker `category`
  indices resolve to readable names.
- Capture each per-thread `markers.{name,startTime,endTime,phase,category,
  data}` array; the `data` payload is captured as raw JSON via a new
  `Iterator::CollectCurrentScope` helper on the JSON parser and
  flattened into `data.*` args at parse time using
  `AddJsonValueToArgs`.
- Phase 0 (Instant) becomes dur=0 slices; phase 1 (Interval) uses
  `endTime - startTime`; phase 2/3 (IntervalStart/End) are LIFO-matched
  per (utid, name) via a cookie stack.
- Markers route through `TrackCompressor` with a blueprint dimensioned
  by `(utid, marker_name)` so each marker name on each thread gets its
  own track (matching the Firefox Profiler's marker chart layout) and
  same-name overlaps fan out across compressed lanes instead of being
  dropped.
- Fix a pre-existing importer bug exposed while developing: when the
  top-level `shared` block carries only the string table (no shared
  frames), per-thread frames/stacks were silently dropped. Threading
  the strings through `Process{Legacy,Preprocessed}FramesAndStacks`
  fixes this.
- Refactor `gecko_trace_tokenizer.cc` parsing into single-purpose
  helpers (`ParseFrameTable`, `ParseFuncTable`, `ParseStackTable`,
  `ParseSamplesTable`, `ParseMarkersTable`, `ParseMeta`,
  `ParseMetaCategories`, `ParseThreads`, `ParseMarkerDataArray`) so
  `ParseThread` and `ParseGeckoProfile` become a flat dispatch.
- Add a dedicated `dev.perfetto.FirefoxProfilerMarkers` UI plugin
  (enabled by default) that lays out `firefox_marker` tracks under
  each thread, sub-grouped by category when a thread spans multiple
  categories. A single SQL query computes per-(utid, name) track
  groups with their dominant category and the depth needed for
  intra-name overlap.
- Add diff tests covering all four phases, payload flattening, and
  same-name overlap fanning out across lanes.
- Update other-formats.md to reflect markers being supported, and add
  generation instructions for samply and Python's `profiling.sampling
  --gecko` (3.15+).


stevegolton (approved):
UI part LGTM

sashwinbalaji (approved):

PR #5696, opened 2026-05-01
tp,ui: add PROFILE_FROM_TREE aggregate and expose it in flamegraphs

zezeozue:
PROFILE_FROM_TREE(id, parent_id, frame_name, self_value, sample_type, unit) is a new SQL aggregate that converts any tree-shaped table into a serialized pprof Profile. The stdlib module `pprof.from_tree` exposes it as `_pprof_from_tree!`, composing on top of the `std.trees.*` operators so users can filter, propagate, and emit pprof through the same primitives.

The Flamegraph widget gains a "Download as pprof" toolbar button. Encoding lives in the engine; the widget delegates via a callback so the bytes always match the SQL surface.

The SQL query results panel and SQL table tab gain an "Add Flamegraph" control mirroring "Add debug track": the user picks the id/parent_id/name/value columns of any query, optionally overrides sample type and unit, and the result opens as an ephemeral flamegraph tab that inherits the toolbar Download button.

Documentation under analysis/pprof-from-tree.md describes the SQL surface and walks through recipes for heap dumps, callstack samples, message-queue stacks, and Java heap dominator trees. Diff tests cover the aggregate and the stdlib macro.

### DIFF
--- a/ui/src/plugins/dev.perfetto.QueryPage/results_table.ts
+++ b/ui/src/plugins/dev.perfetto.QueryPage/results_table.ts
@@ -16,6 +16,7 @@ import m from 'mithril';
 import {classNames} from '../../base/classnames';
 import {Icons} from '../../base/semantic_icons';
 import {AddDebugTrackMenu} from '../../components/tracks/add_debug_track_menu';
+import {AddFlamegraphMenu} from '../../components/tracks/add_flamegraph_menu';
 import type {DataSource} from '../../components/widgets/datagrid/data_source';
 import {DataGrid, renderCell} from '../../components/widgets/datagrid/datagrid';
 import type {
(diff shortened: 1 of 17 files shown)


ui/src/plugins/dev.perfetto.QueryPage/results_table.ts
```diff
@@ -198,6 +199,22 @@ export class ResultsTable implements m.Component<ResultsTableAttrs> {
...
+        trigger: m(Button, {
+          label: 'Add flamegraph',
+          icon: 'local_fire_department',
+        }),
+        position: PopupPosition.Top,
+      },
+      m(AddFlamegraphMenu, {
+        trace: attrs.trace,
+        query: data.lastStatementSql,
+        availableColumns: data.columns,
```
stevegolton, line 214: Maybe add an onAdd handler here that navigates to the timeline page, otherwise when this is called from the query page, nothing happens.

See AddDebugTrackMenu above.

Something like: `onAdd: () => attrs.trace.navigate('#!/viewer'),`

stevegolton (approved):
All good from the UI side (% one suggestion)

LalitMaganti (changes requested):
Already spoke to Zim, this cannot land from the TP side until PerfettoSQL next is a thing.

PR #5697, opened 2026-05-01
tests: guard remaining integrationtests inside enable_perfetto_integration_tests

svmhdvn:
This fixes the standalone build with enable_perfetto_integration_tests=false due to test_helper.

### DIFF
--- a/src/trace_processor/BUILD.gn
+++ b/src/trace_processor/BUILD.gn
@@ -396,27 +396,29 @@ perfetto_cc_proto_descriptor("gen_cc_test_messages_descriptor") {
   descriptor_target = "../protozero:testing_messages_descriptor"
 }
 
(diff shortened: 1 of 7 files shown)


LalitMaganti (approved):

LalitMaganti:
Nice thanks for contribution!

svmhdvn:
Sorry for the PR spam, but it looks like there are other similar changes I need to make to guard unittests against the same thing. Should I just keep this PR as is and start a new one for unittests? Or just draft this PR, add commits to it, then undraft when ready for review?

LalitMaganti:
I think separating into a different PR for unit vs integrationtests makes sense!

sashwinbalaji (approved):

svmhdvn:
Sounds good, will do.

PR #5706, opened 2026-05-02
traced_perf: add support for in-kernel frame pointer unwinding

### LINKED ISSUE #3300: Add the ability to choose in-kernel frame pointer unwinding of the user stack
The ability to use frame pointer unwinding landed in ab3bf51c8300eb01b00c5d0029a87f1d34dfba40. But as far as I can tell this still copies the user stack bytes and does the unwinding in user-land, which is slow (copies a lot of memory) and which leads to truncated stacks because it can only unwind the top 64KB of the stack.

https://github.com/google/perfetto/blob/0ce5e90e79f04dbd589146d06b83af4a3a3f9975/src/profiling/perf/event_config.cc#L623-L638

I would like to have a path which asks the kernel to do the unwinding of the user stack, i.e. which uses `PERF_SAMPLE_CALLCHAIN`, and which does not drop the user frames (`exclude_callchain_user = false`). This path would not set `PERF_SAMPLE_STACK_USER`, i.e. it would skip the copying of the stack bytes.

Could such a path be added? cc @LalitMaganti

LalitMaganti:
This CL adds support for unwinding frame pointers in the kernel allowing for
higher sampling rates and less overhead.

Fixes: https://github.com/google/perfetto/issues/3300


src/profiling/perf/event_config.h
```diff
@@ -142,6 +142,15 @@ class EventConfig {
   bool sample_callstacks() const { return user_frames() || kernel_frames_; }
   bool user_frames() const { return IsUserFramesEnabled(unwind_mode_); }
   bool kernel_frames() const { return kernel_frames_; }
+  // True iff the kernel walks the userspace frame pointer chain for us
+  // (UNWIND_KERNEL_FRAME_POINTER). The producer still needs the process's
+  // /proc/<pid>/maps (and /proc/<pid>/mem) to fill in mapping info and build
+  // IDs for the IPs the kernel returns - it just doesn't need to do the
+  // stack unwind itself.
```
rsavitski: nit: drop the entire comment. It's not wrong, but this struct is not the place to explain the implementation details. It ends up distracting the reader since such an overdocumented accessor stands out compared to the rest, yet its name (`kernel_unwinds_user_frames`) is already sufficiently descriptive.

src/profiling/perf/perf_producer.cc
```diff
@@ -266,6 +266,61 @@ uint32_t TimeToNextReadTickMs(DataSourceInstanceID ds_id, uint32_t period_ms) {
...
+  for (uint64_t ip : kernel_ips) {
+    if (ip == PERF_CONTEXT_USER)
+      return true;
+  }
+  return false;
+}
+
+// Translates the proto-level unwind mode into the unwinder's internal enum.
+// Unknown / SKIP / DWARF all collapse to the libunwindstack path, which is
+// the historical default.
```
rsavitski: nit: delete comment, it adds nothing to the implementation. Also, the latter might be easier with a "using" declaration for enum namespace prefixes.

src/profiling/perf/perf_producer.cc
```diff
@@ -266,6 +266,61 @@ uint32_t TimeToNextReadTickMs(DataSourceInstanceID ds_id, uint32_t period_ms) {
...
+//    signal here.
+//
+//  * UNWIND_KERNEL_FRAME_POINTER does not sample user regs. We instead use
+//    PERF_SAMPLE_CALLCHAIN, which includes a PERF_CONTEXT_USER marker iff
+//    the task has a userspace context. Note: this is independent of
+//    |cpu_mode|, since a userspace task sampled in kernel mode (e.g. during
+//    a syscall) still gets a PERF_CONTEXT_USER section.
+//
+// In both cases the signal is something the kernel synthesises specifically
+// to indicate "this task has no userspace", not a heuristic on our side.
```
rsavitski: This is too verbose and uses too many implementation constants that are only useful if you already know the answer. How about:

// Returns true if this sample is for a kernel thread:
// * if we're performing unwinding in this profiler, then we requested userspace registers in the samples. No registers means it's a kernel thread.
// * if the kernel is unwinding userspace, look for a PERF_CONTEXT_USER marker in the callchain as that mode doesn't sample userspace registers.

src/profiling/perf/perf_producer.cc
```diff
@@ -266,6 +266,61 @@ uint32_t TimeToNextReadTickMs(DataSourceInstanceID ds_id, uint32_t period_ms) {
   return period_ms - ((now_ms - ds_period_offset) % period_ms);
 }
 
+// True if the kernel-supplied callchain contains a userspace section. The
+// kernel inserts a PERF_CONTEXT_USER marker before the userspace frames iff
+// the sampled task has a userspace context, regardless of whether the FP
+// walk produces any user IPs.
```
rsavitski, line 272: note: unlike elsewhere, please keep this comment as is, the part about context being inserted even if the callchain is empty is useful.

src/profiling/perf/unwinding.cc
```diff
@@ -473,39 +520,71 @@ CompletedSample Unwinder::UnwindSample(const ParsedSample& sample,
...
+
+  // Use libunwindstack's canonical helper to compute |rel_pc| (handling
+  // |load_bias|, |elf_offset|, and architecture-specific PC adjustment for
+  // return addresses), and to populate |map_info|. This is the same path the
+  // libunwindstack/FP unwinders rely on for emission, just driven directly
+  // by the IPs the kernel returned.
+  for (; it != sample.kernel_ips.end(); ++it) {
+    unwindstack::FrameData frame = unwindstack::Unwinder::BuildFrameFromPcOnly(
+        *it, unwindstack::Regs::CurrentArch(), &user_state->fd_maps,
+        /*jit_debug=*/nullptr, user_state->fd_mem,
```
rsavitski, line 557: Hmm, please stick an obvious "TODO(rsavitski): add vdso redirect and remove fd_mem" comment above this.

I want to move away from requiring /proc/pid/mem for native processes (at least by default), but I need to first land vdso handling (as it's not a file on disk). In the future, this should become a nullptr so that libunwindstack only tries on-disk libraries (see MapInfo::GetElf -> MapInfo::CreateMemory -> MapInfo::CreateMemory).

src/profiling/perf/unwinding.h
```diff
@@ -196,9 +196,24 @@ class Unwinder {
...
+  // with their corresponding build IDs. Only relevant when sampling with
+  // |UnwindMode::kKernelFramePointer|. For kernel threads (no user section)
+  // this is a no-op.
+  //
+  // |user_state| (if non-null) is used to resolve mapping info, |rel_pc|, and
+  // build IDs for each user IP. If null, the user IPs are skipped (they would
+  // be uninterpretable without mapping context).
+  void AppendKernelSuppliedUserFrames(const ParsedSample& sample,
+                                      UnwindingMetadata* user_state,
+                                      CompletedSample* out);
```
rsavitski: The function signatures for both (and comments) are really odd and seem specialised for the implementation, not readability.

1) what was wrong with `SymbolizeKernelCallchain` signature? Is the vector becoming an output pointer an attempt to minimise intermediate allocations?
2) `AppendKernelSuppliedUserFrames` now takes the overly-specific `CompletedSample*`, but logically is also just something that builds a vector of FrameData.

Unless the allocation churn is significant, it makes more sense to have two functions that extract-and-return-as-a-vector the relevant parts of the callchain.

src/profiling/perf/unwinding.cc
```diff
@@ -34,6 +34,41 @@ constexpr uint32_t kDataSourceShutdownRetryDelayMs = 400;
...
+
+// Iterator to the start of the userspace section of the kernel-supplied
+// callchain (the element after the PERF_CONTEXT_USER marker), or end() if no
+// user section is present (e.g. kernel thread sample, or callchain
+// configured to exclude user frames).
+std::vector<uint64_t>::const_iterator UserCallchainBegin(
+    const std::vector<uint64_t>& kernel_ips) {
+  auto it = std::find(kernel_ips.begin(), kernel_ips.end(), PERF_CONTEXT_USER);
+  return it == kernel_ips.end() ? it : it + 1;
+}
```
rsavitski: nit: I'd inline this given there's only one caller for 2 lines of code.

src/profiling/perf/unwinding.cc
```diff
@@ -276,9 +311,10 @@ base::FlatSet<DataSourceInstanceID> Unwinder::ConsumeAndUnwindReadySamples() {
     // TODO(rsavitski): start tracking process exits more accurately, either
     // via PERF_RECORD_EXIT records or by checking the validity of the procfs
     // descriptors.
-    if (PERFETTO_UNLIKELY(!entry.sample.regs &&
-                          proc_state.status ==
-                              ProcessState::Status::kFdsResolved)) {
+    bool proc_fds_resolved =
+        proc_state.status == ProcessState::Status::kFdsResolved;
+    if (PERFETTO_UNLIKELY(IsLikelyKthreadPidReuse(
```
rsavitski: I think it'll be sufficient to add a ( mode =! kKernelFramePointer) to the existing check, since that hack exists to avoid dealing with the mistaken thread in the libunwindstack-based unwinding.

Whereas with kernel-based unwinding, the only thing the profiler does is use procfs descriptors for symbolisation, and those should fail cleanly.

LalitMaganti:
PTAL: the diff to main should now be significantly reduced, making it much easier to review. I'll leave any optimizations for future patches.

src/profiling/perf/perf_producer.cc
rsavitski, line 918: nit: let's revert the comments here since the logic is unchanged, and the original phrasing is still valid.

rsavitski (commented):
LGTM with one more nit.

rsavitski (approved):

LalitMaganti:
Thanks! Removed the unnecessary comment changes.

PR #5708, opened 2026-05-02
tp: layer the critical-path walker into enumerate + flatten

fiveapplesonthetable:
## Summary
- Refactor `_critical_path_by_roots!` from a monolithic walk-and-flatten into two composable PerfettoSQL macros: **layered** (every viable depth) and **flatten** (collapse to one blocker per `(root_id, ts)`). The composition `flatten(layered(...))` is bit-identical to the previous implementation; existing critical-path diff tests pass unchanged.
- Two complementary user-facing SQL functions:
  - `_thread_executing_span_critical_path_layered(root_utid, ts, dur)` — at each `(ts, dur)` slice, one row for the on-CPU thread (deepest-only).
  - `_thread_executing_span_critical_path_all_layers(root_utid, ts, dur)` — same shape but with the full ancestor stack at each slice (depth N is on-CPU, depth N-1 is the thread waiting for it, …, depth 0 is the root).
- New UI commands `dev.perfetto.CriticalPathLayered` and `dev.perfetto.CriticalPathLayeredAll` with two buttons on the thread-state details panel — "Critical path layered (deepest)" and "Critical path layered (all chains)". Each pins a collapsible `TrackNode` group at the top of the workspace.

## Drill-down UX

Each command pins **one collapsed group** (single row of UI footprint). The parent track shows the canonical flattened single-chain critical path — "answer at a glance". Expand to see the per-depth breakdown: one child slice track per active layer (`◀ depth N`), capped at 8 layers. Slices are coloured by `<thread> [<process>]` so the same on-CPU thread keeps the same colour across the summary row and every layer it appears in.

The **deepest** view's depth tracks are sparse (depth 0 only where root is on-CPU). The **all chains** view's depth tracks are dense (depth 0 = root everywhere, depth 1 = immediate waker chain everywhere, …) so you can read the full ancestor stack at every instant rather than just "who's on-CPU".

## Design

The C++ `__intrinsic_critical_path_walk` already emits a multi-depth DAG annotation (one row per `(root_id, depth, ts, dur, id, parent_id)` frame). The previous SQL pipeline collapsed it to one blocker per `(root_id, ts)` immediately, hiding every layer above the leaf from callers.

Splitting into `layered` and `flatten` exposes the full DAG to SQL callers. The deepest-wins selection is intrinsic to `flatten` (it walks `parent_id` inside `_intervals_flatten!`); no separate selection macro is needed.

```
_critical_path_by_roots!(roots, nodes)
  ≡ _critical_path_flatten!(_critical_path_layered_by_roots!(roots, nodes))
```

The all-chains view is a recursive CTE on top of the layered output that walks `_wakeup_graph.waker_id` one hop per recursion to surface every ancestor at the same `(ts, dur)`.

## Performance (sched_wakeup_trace.atr, trace_processor_shell)

| query                     | lite (existing) | layered (new) |
|---------------------------|----------------|---------------|
| narrow per-utid window    | 334 ms         | 324 ms (~3% faster) |
| whole-trace               | 528 ms         | 349 ms (~34% faster — skips the deepest-wins flatten step entirely) |

## Test plan
- [x] All 24 `TablesSched` diff tests pass (22 existing + 2 new).
- [x] New `test_thread_executing_span_critical_path_layered` covers multi-depth deepest-only output for the existing `sched_wakeup_trace.atr` fixture.
- [x] New `test_thread_executing_span_critical_path_all_layers` covers full-ancestor expansion (each slice with depth=N also has rows at depths 0..N-1).
- [x] E2E verified in browser on a synthetic A→B→main lock-contention trace: both commands pin a collapsed group whose summary shows the canonical chain; expanding them reveals the per-depth breakdown — sparse for "deepest", dense for "all chains".

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/critical_path.cc
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/critical_path.cc
@@ -36,14 +36,11 @@
 #include "src/trace_processor/sqlite/bindings/sqlite_result.h"
 #include "src/trace_processor/sqlite/bindings/sqlite_value.h"
 
-// SQLite intrinsics backing `sched.thread_executing_span`'s critical
-// path: an aggregate that materialises the wakeup graph and a walk that
-// turns it into per-root blocker frames. Each root is walked
-// independently as iterative DFS over the wakeup chain; every frame is
-// bounded by its node's [ts - idle_dur, ts + dur] window intersected
-// with the caller's recursion window. Each push descends to a node
-// with a strictly smaller `ts`, bounding work at one (node, sub-window)
-// per reachable causal predecessor.
+// SQLite intrinsics for `sched.thread_executing_span`'s critical path:
+// an aggregate that materialises the wakeup graph and an iterative DFS
+// that emits per-root blocker frames. Each push descends to a node with
+// a strictly smaller `ts`, bounding work at one (node, sub-window) per
+// reachable causal predecessor.
 namespace perfetto::trace_processor {
 namespace {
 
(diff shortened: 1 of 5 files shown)


fiveapplesonthetable:
Closing — moving this work to the fork branch only.

PR #5711, opened 2026-05-04
ui: show trace start and duration in timeline header

LalitMaganti:
Replace the time-axis origin gutter (which previously showed the axis
origin in various forms — sometimes a date, sometimes a two-line
timecode) with a consistent two-line "Start" and "Duration" pair that
applies to every timestamp format.

- Start: the trace's wall-clock start, formatted per the active
  timestamp format (HH:MM:SS for realtime/timecode, numeric for
  Seconds/Milliseconds/Microseconds, raw for TraceNs/TraceNsLocale).
  Sub-second precision is dropped from realtime renders since it's
  noisy.
- Duration: end - start, formatted via the shared formatDuration so
  it matches duration formatting elsewhere in the UI.

Screenshot 2026-05-04 at 21 58 16


stevegolton:
> which previously showed the axis origin in various forms — sometimes a date, sometimes a two-line timecode

There's a good reason for the way this works. The general idea is that the origin offset label and the tick labels work in tandem to ground the user.

image

Different timestamp formats have different origin offsets and tick offsets:

| Format           | Origin    | Tick offset       |
|------------------|------------|-------------------|
| Timecode         | TP origin  | Trace start       |
| Seconds/ms/us/ns | TP origin  | Trace start       |
| Real time        | Unix epoch | Previous midnight |

- The origin offset is what we consider t=0. For real time formats this is the unix epoch, for all other formats this is TP origin (usually boot time).
- The tick offset is where we consider 0 to be for the tick labels - this is at what point the '0' tick would be positioned. For real tie formats this is the previous midnight (because that's how date-less time works) and for other formats this is the start of the trace.

This has the nice property that if you want to find the absolute timestamp of any tick on the timeline, simply add the offset to the tick label (as indicated by the little plus symbol to the right of the offset). I don't expect folks aren't doing this in their head, but it's a simple rule to follow for how these two labels relate.

There's a few problems I can see with the approach in this PR.

1. It breaks the neat relationship between the offset and the ticks (in real time format modes, the tick offset != trace start).
2. When we have dual line timestamps printed in the tick labels (timecode, or realtime timestamps) the second line is sub-seconds, which now doesn't line up with the second line of the offset label.

I realise these are nits, but I want to convey that quite a bit of thought has gone into the design of what you're changing and I don't think we should change it without considering why.

From what I understand, the reason behind this PR in the first place was simply to add the absolute trace duration somewhere (the start offset was not part of the initial requirements, but correct me if I'm wrong). Let's focus on finding somewhere to put that. I don't want to overload this little rectangle of real estate too much. Perhaps we could put it down in the RHS of the status bar?

image

Edit: reworded and added more context for clarity.

LalitMaganti:
If that unblocks getting the duration somewhere on screen sure. Though I think it fits better on the timeline if this gets something in then I can move it to the status bar. Would you like me to make that change?

stevegolton:
> If that unblocks getting the duration somewhere on screen sure. Though I think it fits better on the timeline if this gets something in then I can move it to the status bar. Would you like me to make that change?

Go for it!

LalitMaganti:
As discussed offline where we all agreed status bar didn't look too great, I've reverted back to what we had before, I'm just not dropping the last half of the timecode/realtime anymore, being conisstent with how it's displayed everywhere else.

stevegolton (approved):

PR #5712, opened 2026-05-04
ui: generalize Embedder hooks for third-party deployments

LalitMaganti:
Carve out the build-time and runtime surfaces that branding-customising
embedders previously had to fork-patch:

Build-time (new ui/config/default.json + --config flag on build.js):
  appTitle, errorMessage, faviconPath are substituted into index.html;
  includePlugins points to an allowlist file forwarded to
  tools/gen_ui_imports via the new --include-list flag. build.js
  cross-checks the allowlist against the union of //ui/src/plugins and
  //ui/src/core_plugins so a typo cannot silently drop a plugin.

Runtime (Embedder interface):
  - homePage replaces the route '/' component when set.
  - brandLogo swaps the sidebar wordmark; undefined keeps the bundled
    Perfetto brand asset so default deployments are unchanged.
  - CookieConsent is now mounted only when embedder.analyticsId is
    defined, so non-analytics deployments never instantiate it.

### DIFF
--- a/tools/gen_ui_imports
+++ b/tools/gen_ui_imports
@@ -59,9 +59,23 @@ def plugin_has_css(dir):
   return os.path.isdir(dir) and os.path.exists(os.path.join(dir, 'styles.scss'))
 
 
-def gen_imports(input_dir, output_path):
+def parse_include_list(path):
+  """Reads a plugin allowlist file. One basename per line; blank lines and
+  lines starting with '#' are ignored. Returns the set of allowed names."""
+  allowed = set()
+  with open(path, 'r') as f:
+    for line in f:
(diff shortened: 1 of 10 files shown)


primiano:
Can we discuss this on Wed? I'm not sure I want to add much more complexity to build.js (Especially as i'm in the process of moving it over to vite) so that somebody can change the title and co.
I need some time to digest this, figure out what you are trying to do and propose alternatives.
I'm not even sure we should put effort to make it easier to rebrand perfetto (it's more... to which extent).
Also at very least before doing any of this we should check with David to see if this would actually work for that use case, rather than doing this in the vacuum and hope it will work for others.

PR #5716, opened 2026-05-04
tp: drop non-root prev_id descent in critical-path walker

fiveapplesonthetable:
## Summary
- Walker descended `prev_id` at non-root frames at the same depth, attributing arbitrary prior episodes of every waker thread to T's critical path
- Those rows are not on T's wakeup ancestry — only the specific waker episode T was waiting on belongs at depth-(N+1)
- Fix: drop the `prev_id` stack push at non-root frames; root is unaffected (its `window_start == node_idle_start` already gated the descent off)

## Probe — app-launch trace, root utid 1128, 131 chains
- chains with multi-node depth-1 groups: **55 / 131 (42%)**
- max prior episodes per group: **59**
- extra layered rows from this path: **240**

## Performance — same trace, full window
- layered rows: 2786 → 475
- lite rows: 2791 → 475
- query: 7523 → 3162 ms (**2.4× faster**)

## Tests
The `thread_executing_span_critical_path_stack` golden is repointed to a timestamp whose chain has multi-depth coverage in the new walker. The original `ts` drew its coverage from the prev descent and is now correctly empty.

## Test plan
- [x] `tools/diff_test_trace_processor.py out/linux/trace_processor_shell --name-filter TablesSched` (23/23)
- [x] `tools/diff_test_trace_processor.py --name-filter Fuchsia` (11/11)
- [x] E2E recorded UI walk on app-launch trace via the CP-tree drill-down plugin: chain ancestry only, no phantom prev episodes

### DIFF
--- a/src/trace_processor/perfetto_sql/intrinsics/functions/critical_path.cc
+++ b/src/trace_processor/perfetto_sql/intrinsics/functions/critical_path.cc
@@ -18,6 +18,7 @@
 
 #include <algorithm>
 #include <cstdint>
+#include <limits>
 #include <memory>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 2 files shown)


fiveapplesonthetable:
Closing — reverting on author review. The walker's existing prev_id descent is correct: it is gated by the ancestor's idle window (`f.window_start < node_idle_start`), so prev attribution only fires when prev's run actually overlaps an ancestor's idle. Root's own prev is naturally excluded. The duplicate stacking I observed was a UI-side issue with the per-track `_anc` JOIN and was fixed at the SQL layer in the cp-tree branch (collapse `_anc` per chain).

PR #5721, opened 2026-05-05
perfetto: fix ScrapeEmulatedSharedMemoryBuffer never committing

sashwinbalaji:
ScrapeEmulatedSharedMemoryBuffer declared a local CommitDataRequest named |commit_req|, but the lambda passed to ForEachScrapableChunk appended scrape entries to the member |commit_data_req_| instead of the local. 

As a result:
- |commit_req.chunks_to_move_size()| was always zero, so the function always returned early without ever calling |producer_endpoint_->CommitData|. Producer-side SMB scraping for remote (emulated SMB) producers was therefore a silent no-op.
- The scrape entries were instead leaked onto the pending |commit_data_req_|, where the FlushPendingCommitDataRequests drain treats them as arbiter-owned chunks and would ReleaseChunkAsComplete + ReleaseChunkAsFree them while their writers were still mid-write, corrupting the SMB.

Adds a unit test (ScrapeEmulatedSharedMemoryBuffer) that drives the emulated-SMB scrape path with a chunk in kChunkBeingWritten state and asserts that CommitData is invoked with the scraped chunk and that the chunk is left in kChunkBeingWritten afterwards.

### DIFF
--- a/src/tracing/core/shared_memory_arbiter_impl.cc
+++ b/src/tracing/core/shared_memory_arbiter_impl.cc
@@ -974,7 +974,7 @@ void SharedMemoryArbiterImpl::ScrapeEmulatedSharedMemoryBuffer(
     if (writer == buffer_for_writers.end())
       return;
     BufferID target_buffer_id = writer->second;
-    auto* ctm = commit_data_req_->add_chunks_to_move();
+    auto* ctm = commit_req.add_chunks_to_move();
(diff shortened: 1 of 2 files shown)


primiano (approved):

primiano:
@jahdiel-alvarez FYI, you might want to cherry-pick this wherever needed. TUrns out that a silly typo in one of the last iterations turned your thing into a no-op

PR #5730, opened 2026-05-05
ui: Extract FlamegraphPanel widget from QueryFlamegraph

zezeozue:
Mithril component that owns the inner QueryFlamegraph and disposes it on unmount or when trace/dependencies identity changes. Migrates all consumers and drops QueryFlamegraphWithMetrics.

Also adds optional `displayLabel` on the PIVOT view; SQL match still uses `pivot`. This will be used in the future by the 
HeapDumpExplorer to provide a better UX when it auto filters a flamegraph to path hash


ui/src/components/flamegraph_panel.ts
```diff
@@ -0,0 +1,55 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import m from 'mithril';
+import {Trace} from '../public/trace';
+import {QueryFlamegraph, QueryFlamegraphMetric} from './query_flamegraph';
+import {FlamegraphState} from '../widgets/flamegraph';
+
+export interface FlamegraphPanelAttrs {
```
LalitMaganti, line 20: Some comments on these thigns would be good IMO.

LalitMaganti (approved):

PR #5732, opened 2026-05-05
perfetto: add metatrace to power and kernel-wakelocks data sources

sashwinbalaji:
These data sources do periodic synchronous HAL calls on the main task runner thread and maybe causing traced_probes watchdog hangs.

Add PERFETTO_METATRACE_SCOPED at the top of Tick() and each Write* helper in:
  - AndroidCpuPerUidDataSource
  - AndroidKernelWakelocksDataSource
  - AndroidPowerDataSource
  - LinuxPowerSysfsDataSource


src/traced/probes/power/android_power_data_source.cc
```diff
@@ -403,6 +415,9 @@ void AndroidPowerDataSource::WriteEnergyEstimationBreakdown() {
 }
 
 void AndroidPowerDataSource::WriteEntityStateResidency() {
+  PERFETTO_METATRACE_SCOPED(TAG_PROC_POLLERS,
```
rsavitski: I'd probably push the metatrace macro below the next if() for this data source, since it calls all of these functions if /any/ of the functionality is active, and the thin metatrace slices won't give us any insights:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/traced/probes/power/android_power_data_source.cc;l=265;drc=440b6caf63f21cd4682825936ee7a3c3cb81e9a5

include/perfetto/ext/base/metatrace_events.h
```diff
@@ -70,7 +70,14 @@ enum Tags : uint32_t {
...
   F(PROFILER_MAPS_REPARSE), \
-  F(PROFILER_UNWIND_CACHE_CLEAR)
+  F(PROFILER_UNWIND_CACHE_CLEAR), \
+  F(ANDROID_KERNEL_WAKELOCKS_TICK), \
+  F(ANDROID_POWER_TICK), \
+  F(ANDROID_POWER_WRITE_BATTERY_COUNTERS), \
+  F(ANDROID_POWER_WRITE_POWER_RAILS_DATA), \
+  F(ANDROID_POWER_WRITE_ENERGY_ESTIMATION_BREAKDOWN), \
+  F(ANDROID_POWER_WRITE_ENTITY_STATE_RESIDENCY), \
+  F(LINUX_POWER_SYSFS_TICK)
```
rsavitski: optional: things that are often collected in android and might be nice to instrument:
* StatsdBinderDataSource
* AndroidCpuPerUidDataSource

src/traced/probes/power/android_power_data_source.cc
sashwinbalaji: True.
my original reasoning was having the info that call happened is better than us questioning wheter the call we skipped or if data loss. But having ANDROID_POWER_TICK should be conclusive enough + we can look into config.

src/traced/probes/statsd_client/statsd_binder_data_source.cc
```diff
@@ -241,6 +242,8 @@ void StatsdBinderDataSource::Start() {
 void StatsdBinderDataSource::OnData(uint32_t reason,
                                     const uint8_t* data,
                                     size_t sz) {
+  PERFETTO_METATRACE_SCOPED(TAG_PROC_POLLERS, STATSD_BINDER_ON_DATA);
```
rsavitski: Sorry, on second thought and having seen this diff, let's drop this one. This is just the serialisation of a proto and the actual work happens elsewhere. Plus this seems to be O(atoms), so might become spammy in some traces.

rsavitski (approved):

PR #5744, opened 2026-05-06
protovm: extend merge operation with del_if_src_empty

rukkal:
This change implements the "delete submessage" semantics during
merge operations: when the flag delete_if_src_empty is set, an empty
src field means "delete the corresponding dst field".

This functionality is required to replicate in the ProtoVM the removal
of a subobject within the producer's state.

E.g.:
This event on the producer side: mProducerState.mSubobject = nullptr
can be represented with this patch: { producer_state { subobject {} }}

### DIFF
--- a/protos/perfetto/protovm/vm_program.proto
+++ b/protos/perfetto/protovm/vm_program.proto
@@ -125,6 +125,15 @@ message VmOpMerge {
   // the deepest submessages first, then moving up to the parent
   // message and merging it with skip_submessages = true.
   optional bool skip_submessages = 1;
+
+  // When true, the merge operation will remove the dst field if the src field
+  // is empty.
+  //
+  // This can be used to replicate the removal of a subobject within the
+  // producer's state. E.g.
+  //  This event on the producer side: mProducerState.mSubobject = nullptr
+  //  can be represented by this patch: { producer_state { subobject {} }}
+  optional bool del_if_src_empty = 2;
 }
 
 // Copies SRC into DST. If a message, replaces the DST node, discarding any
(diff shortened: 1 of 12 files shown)


rukkal:
@primiano note that this PR contains a commit with a small preliminary refactoring. I ended up not needing the refactoring for this change, but I'd keep it anyways as it makes things slightly cleaner IMO.

protos/perfetto/protovm/vm_program.proto
```diff
@@ -125,6 +125,15 @@ message VmOpMerge {
   // the deepest submessages first, then moving up to the parent
   // message and merging it with skip_submessages = true.
   optional bool skip_submessages = 1;
+
+  // When true, the merge operation will remove the dst field if the src field
+  // is empty.
+  //
+  // This can be used to replicate the removal of a subobject within the
+  // producer's state. E.g.
+  //  This event on the producer side: mProducerState.mSubobject = nullptr
```
primiano: nit: extra space/indent in the last 2 lines.

src/protovm/executor.h
```diff
@@ -59,7 +59,9 @@ class Executor {
   virtual StatusOr<uint64_t> ReadRegister(uint8_t reg_id) const;
   virtual StatusOr<void> WriteRegister(const Cursors& cursors, uint8_t reg_id);
   virtual StatusOr<void> Delete(RwProto::Cursor* dst) const;
-  virtual StatusOr<void> Merge(Cursors* cursors, bool skip_submessages) const;
+  virtual StatusOr<void> Merge(Cursors* cursors,
+                               bool skip_submessages,
```
primiano: probably a bit more error-prone and readable if instead of passing 2 bools, you pass a MergeFlags which can be `kSkipSubmessages | kDelIfSrcEmpty` and the like.

primiano (approved):
Thanks for splitting it. LGTM with a minor comment.

src/protovm/executor.h
rukkal: Done. And Yes! That's better

PR #5745, opened 2026-05-06
tp+ui: make stats per machine and trace

sashwinbalaji:
stats are now tagged with `machine_id`/`trace_id` and exposed through a real columnar `tables::StatsTable`

## Commits

1. **tp: stats per (machine, trace) — core types and plumbing**
   - New `tables::StatsTable` with `name`, `key`, `idx`, `severity`,
     `source`, `value`, `description`, `machine_id`, `trace_id`. Replaces
     `StatsModule`; `__intrinsic_stats` is now a real table accessed via
     `storage->stats_table()`.
   - `GlobalStatsTracker` (writer over the table; `FlatHashMap` side-index
     for O(1) row lookup) + `StatsTracker` (per-context wrapper that
     injects `machine_id`/`trace_id` from the current
     `TraceProcessorContext`).
   - `stats::Scope` enum (`kGlobal`, `kMachine`, `kTrace`,
     `kMachineAndTrace`) on every stat key. Pre-finalization
     materialization preserves the legacy "every kSingle stat visible at
     value=0" contract per context.
   - `kGlobal` convenience helpers (`SetGlobalStats`,
     `IncrementGlobalStats`, `TraceExecutionTimeIntoStats`) for root-context
     callers (forwarding parser, sorter, profiler util).
   - JSON export pinned to a single `(MachineId(0), TraceId(0))` context
     (multi-machine JSON has no consumer spec today; explicit CHECK guards
     against silently dropping data).
   - SQL prelude `stats` view + `metadata` view exposed with `JOINID`
     columns for `machine_id`.

2. **tp: migrate importers, sorter, viewcapture and tests to StatsTracker**
   - Removes `TraceStorage::IncrementStats`/`SetStats` etc. Every
     importer/parser/tokenizer/tracker call site goes through
     `context_->stats_tracker` (per-context) or
     `context_->global_stats_tracker` (root-context). Unittest fixtures
     wire `GlobalStatsTracker(storage.get())` and per-context
     `StatsTracker`s.

3. **ui: stats per (machine, trace) — show machine/trace columns where relevant**
   - `TraceInfoPage` Stats / Import Errors / Trace Errors / Data Losses
     tabs render a "Trace" and/or "Machine" column when more than one of
     each appears in the data. Single-trace single-machine traces are
     unchanged.
   - SQL tier-2 fixes: `stdlib/traced/stats.sql` now joins on
     `(machine_id, trace_id)` per bucket; `metrics/sql/trace_stats.sql`
     aggregates with `SUM(value) GROUP BY name, idx, source, severity`;
     `metrics/sql/trace_metadata.sql` uses `COUNT(DISTINCT name || '|' ||
     ifnull(idx, ''))` for `data_loss_count` / `error_count`.


src/trace_processor/importers/common/global_stats_tracker.h
```diff
@@ -0,0 +1,194 @@
...
+  //
+  // In the new columnar StatsTable, rows only exist when something is
+  // written. To preserve the legacy contract we eagerly insert every
+  // kSingle stat at value=0 the first time a context of the matching scope
+  // is touched. Cost: O(kNumKeys) one-time inserts per touched context
+  // (typically <300 rows per bucket).
+  //
+  // kIndexed stats are NOT pre-materialized — the legacy IndexMap was
+  // sparse too (default-empty), so absence in the new table matches.
+  void MaterializeKSingleRowsForContext(stats::Scope scope,
```
LalitMaganti: This is a bit different to how it worked before. Even if a single stat was *not* touched before, this would still be populated. Now this is not the case. If you wanted to do this properly, I would instead say "ZeroSingleStatsForContext" and then do this in the StatsTracker consructor rather than the first time it ws touched.

src/trace_processor/metrics/sql/trace_metadata.sql
```diff
@@ -85,13 +85,16 @@ SELECT TraceMetadata(
   'suspend_count', (
     SELECT COUNT() FROM android_suspend_state WHERE power_state = 'suspended'
   ),
+  -- The `stats` view can have multiple rows per (name, idx) when a session
```
LalitMaganti: I'm not convinced this is actually the right behaviour?

sashwinbalaji: kept the logic as before

LalitMaganti (approved):

PR #5748, opened 2026-05-06
ui: Add CloneSession support to recording targets

stevegolton:
Adds an optional cloneSession to the RecordingTarget interface and implement cloning in all but the chrome target.

This allows for the currently recording trace to be snapshotted, cloned in readonly mode and extracted without interrupting the session.

This feature is designed to be used in the upcoming Memscope plugin.

### DIFF
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/tracing_protocol/consumer_ipc_tracing_session.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/tracing_protocol/consumer_ipc_tracing_session.ts
@@ -64,7 +64,7 @@ export class ConsumerIpcTracingSession implements TracingSession {
       ipcFactory,
       traceConfig.uniqueSessionName ?? undefined,
     );
-    session.startLive(traceConfig);
+    session.start(traceConfig);
     return okResult(session);
   }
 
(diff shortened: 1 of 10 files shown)


primiano (commented):
Discussed offline. I think the method should only be plumbed in ConsumerIpcTracingSession, and should jsut work everywhere else

ui/src/plugins/dev.perfetto.RecordTraceV2/tracing_protocol/consumer_ipc_tracing_session.ts
```diff
@@ -21,31 +21,69 @@ import {
...
+    this.uniqueSessionName = uniqueSessionName;
+    this.ipcFactory = ipcFactory;
   }
 
   get state(): TracingSessionState {
     return this._state;
   }
 
-  private async start(traceConfig: protos.ITraceConfig): Promise<void> {
+  private async startLive(traceConfig: protos.ITraceConfig): Promise<void> {
```
primiano: Imho startLive is a bit confusing because when I see it I wonder "so where is the non-live start?(i.e. i expect there is another start())?" What does "live" mean?

I'd keep that as start() to avoid causing too many unnecessary questison to the reader tbh.

primiano (approved):
LG with one comment

ui/src/plugins/dev.perfetto.RecordTraceV2/tracing_protocol/consumer_ipc_tracing_session.ts
stevegolton: Good catch. This was left over from another patch that I (mostly) ended up backing out of but evidently forgot to remove this final rename. Reverted.

PR #5755, opened 2026-05-06
traced_probes: add wdog handler to finalize the trace gracefully

primiano:
This change introduces an option to install a watchdog handler.
Rather than killing the process directly, it allows the process
to gracefully handle the wathdog for up to 60 seconds.
We need this in traced_probes to give us a chance to serialize the
trace which will be picked up by traced to better debug these
watchdog failures.

Bug: b/496506220
Flag: trigger_perfetto_on_traced_probes_disconnect

### DIFF
--- a/include/perfetto/ext/base/metatrace_events.h
+++ b/include/perfetto/ext/base/metatrace_events.h
@@ -87,6 +87,9 @@ enum Tags : uint32_t {
 //   tracing_service_impl.cc for the format.
 // * PROFILER_UNWIND_CURRENT_PID represents the PID that is being unwound.
 // * WATCHDOG_CRASH_REASON represents the base::WatchdogCrashReason enum.
+// * WATCHDOG_CRASH_TIME_BOOT_MS The time in ms since boot when the watchdog
+//   timeout alarm triggered (i.e. when it would have caused a crash, without
+//   counting for the optional handler time).
 // * WATCHDOG_CRASH_ACTUAL_{MONO,BOOT,CPU} are the elapsed seconds (since the
 //   offending fatal timer was created) sampled from CLOCK_MONOTONIC,
 //   CLOCK_BOOTTIME and CLOCK_PROCESS_CPUTIME_ID respectively.
@@ -99,6 +102,7 @@ enum Tags : uint32_t {
   F(PROFILER_UNWIND_QUEUE_SZ), \
   F(PROFILER_UNWIND_CURRENT_PID), \
   F(WATCHDOG_CRASH_REASON), \
+  F(WATCHDOG_CRASH_TIME_BOOT_MS), \
   F(WATCHDOG_CRASH_ACTUAL_MONO), \
   F(WATCHDOG_CRASH_ACTUAL_BOOT), \
   F(WATCHDOG_CRASH_ACTUAL_CPU), \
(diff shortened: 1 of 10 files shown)


include/perfetto/ext/base/metatrace_events.h
```diff
@@ -86,14 +86,23 @@ enum Tags : uint32_t {
 // * FTRACE_SERVICE_COMMIT_DATA is a bit-packed representation of an event, see
 //   tracing_service_impl.cc for the format.
 // * PROFILER_UNWIND_CURRENT_PID represents the PID that is being unwound.
+// * WATCHDOG_CRASH_REASON represents the base::WatchdogCrashReason enum.
+// * WATCHDOG_CRASH_ACTUAL_{MONO,BOOT,CPU} are the elapsed seconds (since the
+//   offending fatal timer was created) sampled from CLOCK_MONOTONIC,
+//   CLOCK_BOOTTIME and CLOCK_PROCESS_CPUTIME_ID respectively.
```
sashwinbalaji, line 95: nit: Add FTRACE_SESSIONS for completeness

src/traced/probes/probes_producer.cc
```diff
@@ -688,6 +692,57 @@ void ProbesProducer::OnFlushTimeout(FlushRequestID flush_request_id) {
...
+               ds_to_flush->size());
+  auto flush_callback = [ds_to_flush](ProbesDataSource* ds) {
+    ds_to_flush->erase(ds);
+    if (ds_to_flush->empty()) {
+      PERFETTO_FATAL("Done flushing data sources. Crashing now.");
+    }
+  };
+
+  for (ProbesDataSource* ds : *ds_to_flush) {
+    const FlushRequestID kFlushId = UINT64_MAX;
```
sashwinbalaji: can be moved out of the for loop, nit as compiler will optimise anyway

include/perfetto/ext/base/watchdog.h
```diff
@@ -41,6 +41,28 @@ enum class WatchdogCrashReason {
...
 
+// Information about an imminent watchdog-triggered crash. Passed to the fatal
+// handler installed via Watchdog::Start() so embedders can record it (e.g. via
+// metatrace counters) before the process is force-killed.
+struct WatchdogCrashInfo {
+  int thread_id = 0;
+  WatchdogCrashReason reason = WatchdogCrashReason::kUnspecified;
+
+  // The time in CLOCK_BOOTTIME when the watchdog triggered.
+  int64_t alarm_time_ms = 0;
```
sashwinbalaji, line 52: not used anywhere ?

src/traced/probes/probes_producer.cc
primiano: made it constexpr, but there is 0 chance that even before that would have yielded any actual instruction.
i prefer to keep it next so the code is more readable.

include/perfetto/ext/base/watchdog.h
primiano: ah actually i ment to plumb it in metatrace, i forgot .added now.

include/perfetto/ext/base/metatrace_events.h
primiano: Hmm no that is actually used as a counter. the list, as the comment says is to explain "Values that aren't used as counters" (where we abuse counters to convey a payload)

src/base/watchdog_posix.cc
```diff
@@ -50,6 +51,10 @@ namespace {
 
 constexpr uint32_t kDefaultPollingInterval = 30 * 1000;
 
+// Grace period given to a fatal handler (passed to Start()) to
+// crash the process on its own before the watchdog force-kills it.
+constexpr uint32_t kFatalHandlerGraceMs = 60 * 1000;
```
sashwinbalaji, line 56: shall we set increase it more maybe 2/3 min ? just to 100% ensure we'll get useful data and anyway we can lower it soon after.

sashwinbalaji (approved):

PR #5756, opened 2026-05-06
ui: Use the tp generated stdlib documentation

aMayzner:
Replace the monolithic stdlib docs JSON (generated by Python-parsing SQL files) with a two-source approach: TraceProcessor's `__intrinsic_stdlib_*` table functions serve as the authoritative doc source, and a slimmed-down metadata JSON provides the supplementary bits that TP doesn't expose (module tags, includes, data-availability checks, table importance).


LalitMaganti (commented):
LGTM from my side.

stevegolton (approved):

Issue #5760, opened 2026-05-07
traced_relay data from Guest not visualized in PerfettoUI


SergejIsbrecht:
## Issue

Guest data from `traced_relay` not visualized in PerfettoUI, but is queryable with SQL

PerfettoUI

* procs from GUEST visible
* procs from HOST visible
* `sched_switch` proc-state for HOST procs visible
* `sched_switch` proc-state for GUEST  procs _NOT_ visible
* GUEST CPU tracks empty
<img width="2260" height="1318" alt="Image" src="https://github.com/user-attachments/assets/80eaf4be-8433-419c-b33a-bcdf1f0241bd" />

## Setup
* [Perfetto v54.0](https://github.com/google/perfetto/releases/tag/v54.0)
* Linux Ubuntu (x64) Guest (traced_relay) with Perfetto v54 tracebox
* Android Emulator (x64) Host (traced) with Perfetto v54 tracebox running in the same Linux (Guest)
* PerfettoUI v55.0-086179353 with Auto-Push and `dev.perfetto.MultiTraceOpen` enabled

NOTE: also tested with Perfetto MASTER -- same result

| id | raw_id | sysname | release | version | arch | num_cpus | android_build_fingerprint | android_device_manufacturer | android_sdk_version | system_ram_bytes | system_ram_gb |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 0 | 0 | Linux | 6.6.30-android15-8-gdd9c02ccfe27-ab11987101 | #1 SMP PREEMPT Tue Jun 18 20:50:32 UTC 2024 | x86_64 | 4 | Android/sdk_phone64_x86_64/emu64x:15/AE3A.240806.019/12368160:userdebug/test-keys | unknown | 35 | 3121680384 | 3 |
| 1 | 3578731344 | Linux | 6.14.0-37-generic | #37-Ubuntu SMP PREEMPT_DYNAMIC Fri Nov 14 22:10:32 UTC 2025 | x86_64 | 8 | null | null | null | 16047329280 | 16 |

## Action

Android `traced` and data source `traced_probes`
```
sergej@sergej-ThinkPad-P50 ~/D/android-x64> adb shell -- 'PERFETTO_PRODUCER_SOCK_NAME=127.0.0.1:38593 /data/local/tmp/tracebox traced --enable-relay-endpoint'
[...]

PR #5762, opened 2026-05-07
perfetto: add multi-machine information to --query

LalitMaganti:
When multiple machines are connected (e.g. using traced_relay), add
support in --query and --query-raw to indicate which machine the source
came from. This is backwards compatible as host-only traces are left
completely unchanged (i.e. we don't pollute the output of --query
unnecessarily).

Fixes the follow-up raised in #5760.

**Output**

Multi-machine `--query`:

```
PRODUCER PROCESSES CONNECTED:

ID     PID      UID      FLAGS  MACHINE              NAME                                       SDK
==     ===      ===      =====  =======              ====                                       ===
1      0        501             guest-B              example_system_wide                        Perfetto v55.0
2      0        501             guest-A              example_system_wide                        Perfetto v55.0
3      -1       0               (host)               example_system_wide                        Perfetto v55.0


DATA SOURCES REGISTERED:

NAME                                     PRODUCER                     MACHINE              DETAILS
===                                      ========                     =======              ========
track_event                              example_system_wide (1)      guest-B              ...
track_event                              example_system_wide (2)      guest-A              ...
track_event                              example_system_wide (3)      (host)               ...
```

Host-only `--query` output is byte-identical to before.

### DIFF
--- a/protos/perfetto/common/tracing_service_state.proto
+++ b/protos/perfetto/common/tracing_service_state.proto
@@ -48,6 +48,19 @@ message TracingServiceState {
     // Returns true if the process appears to be frozen (Android only).
     // Introduced in Perfetto V49 / Android 24Q4.
     optional bool frozen = 6;
+
+    // Identifies the machine this producer is connected from in multi-machine
+    // tracing setups (e.g. when going through traced_relay). Unset (or zero)
+    // means the producer is connected from the local/host machine where the
+    // tracing service is running.
+    // Introduced in Perfetto V56 / Android 26Q3.
+    optional uint32 machine_id = 7;
+
+    // Human-readable name of the machine, as advertised by the relay (usually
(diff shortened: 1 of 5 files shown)


protos/perfetto/common/tracing_service_state.proto
```diff
@@ -48,6 +48,17 @@ message TracingServiceState {
     // Returns true if the process appears to be frozen (Android only).
     // Introduced in Perfetto V49 / Android 24Q4.
     optional bool frozen = 6;
+
+    // Identifies the machine this producer is connected from in multi-machine
+    // tracing setups (e.g. when going through traced_relay). Unset (or zero)
+    // means the producer is connected from the local/host machine where the
+    // tracing service is running.
+    optional uint32 machine_id = 7;
```
primiano, line 57: Added: Introduced in v27 (and add a note in CHANGELOG)

primiano (approved):

PR #5766, opened 2026-05-07
tp: refactor *Engine -> Connection and split out Database

LalitMaganti:
This CL does some prep work to make it much easier to distinguish
between connection local objects and database global objects. Right
now this doesn't matter os much because TP only has one connection
and one database but this will not apply with multi-threaded trace
processor.


src/trace_processor/sqlite/sqlite_database.cc
```diff
@@ -0,0 +1,41 @@
+/*
```
sashwinbalaji: nit: do we need this file or can wrap it in .h only and split it out if it grow ?

src/trace_processor/sqlite/module_state_manager.cc
```diff
@@ -30,10 +29,12 @@ namespace perfetto::trace_processor::sqlite {
 void ModuleStateManagerBase::OnCommit() {
   std::vector<std::string> to_erase;
   for (auto it = state_by_name_.GetIterator(); it; ++it) {
-    if (auto& state = it.value(); state->active_state) {
-      state->committed_state = state->active_state;
-      state->savepoint_states.clear();
+    auto& state = *it.value();
```
sashwinbalaji: nit: initialization in if as before

sashwinbalaji (approved):

PR #5767, opened 2026-05-07
profiling: add new parser for /proc/pid/smaps

rsavitski:
Parses /proc/pid/smaps into a structure-of-arrays style packet with packed repeated fields. Supports collecting only config-specified set of fields, and two modes: aggregated by path name and not.

The implementation is self-contained so that it can be used without dependencies on libmeminfo/libprocinfo on android.

Followup patches will:
* add config options for including/excluding path names.
* add this to libperfetto_client_experimental on android so that it can be used within the perfetto_hprof ART plugin.
* use this from perfetto_hprof (configured via java_hprof_config.proto)

A test 5 MB smaps file (system_server from an android device), serialises to 95 KB aggregated, 130 KB unaggregated (of which 78 KB are strings that will compress well). On a desktop, this runs in <30ms, but excludes kernel seqfile writing duration, which I'm yet to measure.


protos/perfetto/trace/profiling/smaps.proto
```diff
@@ -39,9 +39,42 @@ message SmapsEntry {
   optional uint64 shared_clean_resident_kb = 13;
   optional uint64 locked_kb = 14;
   optional uint64 proportional_resident_kb = 15;
-};
+}
+
+// Structure of arrays style smaps packet of a single process. All packed
+// repeated fields, if set, must have the same length.
+// Quirks:
+// * if aggregating, we always emit an entry the empty-named mapping.
```
primiano: nit: missing  "for" (for the empty-named)

protos/perfetto/trace/profiling/smaps.proto
```diff
@@ -39,9 +39,42 @@ message SmapsEntry {
...
+// * if aggregating, name_id is currently 1:1 with string_table, but still
+//   written for potentially sharing the table between packets in the future.
```
primiano: not sure I undertstand what this comment means

protos/perfetto/trace/profiling/smaps.proto
```diff
@@ -39,9 +39,42 @@ message SmapsEntry {
...
+message PackedSmaps {
+  repeated string string_table = 1;
+
+  // Interned file paths as an index into |string_table|.
+  // Note: the "(deleted)" suffix is kept, and is to be handled during trace
```
primiano: I feel it's a bit more future proof doing this:
1) remove the "(deleted") it from the mapping name. For most analysis purposes it's going to be more an inconvenience
2) Let's transform "name_id" into "name_id_and_flags". Shift it by 3 bits left and let's use the bottom 3 bits for flags. For now we only have one flag that is "deleted". If we ever need more flags we can always use the 3rd bit as an "flag extension", which means it's shifted by 8 and we have more space for future flags.

So trace processor will have to shift-right name_id_and_flags to get the index in the string_table.

Alternatively you can have a `repeated uint8 flags [packed]` column and omit it if all flags are 0, up to you

src/profiling/smaps/smaps.cc
```diff
@@ -0,0 +1,357 @@
...
+    std::string_view stable_sv(storage_.back());
+    map_.Insert(stable_sv, index);
+    return index;
+  }
+
+  const std::deque<std::string>& OrderedStrings() const { return storage_; }
+
+ private:
+  base::FlatHashMap<std::string_view, StringId> map_;
+  std::deque<std::string> storage_;
```
primiano, line 58: I think rather than this here you could have a HeapBuffered<PackedSmaps> so you build the string_table proto as you go, without churning heap too much.
You can't take StringView onto a heap buffer as they can be scattered across chunks, but I'd do this:


  base::FlatHashMap<uint64_t /*hash(string)*/, StringId, base::AlreadyHashed<uint64_t>> map_;
  protozero::HeapBuffered<PackedSmaps> string_table_;


Then in Intern() you do pretty much the same other than you have to invoke the hasher (use MurMur) yourself.
You can rely on the fact that if the hash matches you don't need to check for collisions, because I promise you you won't get any within 1000 strings

src/profiling/smaps/smaps.cc
```diff
@@ -0,0 +1,357 @@
...
+  uint64_t shared_dirty_kb = 0;
+  uint64_t private_clean_kb = 0;
+  uint64_t private_dirty_kb = 0;
+  uint64_t locked_kb = 0;
+  uint64_t pss_kb = 0;
+  uint64_t pss_dirty_kb = 0;
+  uint64_t swap_pss_kb = 0;
+};
+
+// clang-format off
```
primiano, line 79: minor thing, what I found is that rather than playing with clang-format off/onn (which can break if people move things around), you can achieve this stuff by just appending a tailing `//` after each line.
(also feel free to ignore if for some reason doens't work)

src/profiling/smaps/smaps.cc
```diff
@@ -0,0 +1,357 @@
...
+struct SmapsFieldDef {
+  SmapsField flag;
+  uint64_t Vma::* member_ptr;
+  int32_t config_pb_enum;
+  uint32_t trace_field_id;
+};
+using SC = protos::gen::SmapsConfig;
+using SP = protos::pbzero::PackedSmaps;
+constexpr SmapsFieldDef kSmapsFieldDefs[] = {
+  {kSize,         &Vma::size_kb,          SC::VMA_FIELD_SIZE,          SP::kSizeKbFieldNumber},
```
primiano, line 108: given that you have this nice table, why don't you:
1. Add a column with the prefix here (e.g. "Pss:", "SwapPss:")
2. when parsing create a hashmap from prefix to entry in this table
3. So you can make the parsing code more generic and less bug-prone, as it will be all based on the table?

src/profiling/smaps/smaps_unittest.cc
```diff
@@ -0,0 +1,309 @@
...
+ShmemPmdMapped:        0 kB
+FilePmdMapped:         0 kB
+Shared_Hugetlb:        0 kB
+Private_Hugetlb:       0 kB
+Swap:                  0 kB
+SwapPss:               0 kB
+Locked:                0 kB
+THPeligible:    0
+VmFlags: mr mw me 
+)";
```
primiano, line 238: let's add a (deleted) entry (See my other comment)

primiano (commented):
Great stuff. i have a few comments but other than that looks in great shape

src/profiling/smaps/smaps_unittest.cc
primiano: Disussed offline, let's deal with this in TP instead

src/profiling/smaps/smaps.cc
primiano: Discussed offline. keep as is as this is a good tradeoff between perf and maintainability

src/profiling/smaps/smaps.cc
rsavitski: The trailing "//" usually helps for forcing a linebreak, but here I'm trying to keep the columnar layout, while clang-format ruins that by packing the elements densely.

I'm keeping the clang-format directives, but made them tighter.

rsavitski:
PTAL.

primiano (approved):

PR #5769, opened 2026-05-07
ui: Add rollup-plugin-visualizer option to rollup config

stevegolton:
Add option to generate a report showing which files contribute to the package when running ui/build with BUNDLE_STATS=true configured. I.e. `BUNDLE_STATS=true ui/build`.

The report is generated alongside the various *_bundle.js files in the `out/ui/ui/dist_version` directory.

For example:
image


primiano (approved):
LGTM but make it a cmdline argument rather than an env var?

PR #5783, opened 2026-05-08
ui: Add custom SVG-based line chart component

stevegolton:
Adds LineChartSvg - a Mithril+SVG drop-in replacement for the ECharts backed LineChart.

Adds features such as series hover & tooltip emphasis, and handles updates much more cleanly without losing state compared to the echarts approach.

Also, add to the widget demo page, and add features to CursorTooltip (popup position and offset).

### DIFF
--- a/ui/src/base/array_utils.ts
+++ b/ui/src/base/array_utils.ts
@@ -86,14 +86,28 @@ export function intersperse<T, S>(arr: T[], separator: S): (T | S)[] {
   );
 }
 
-export function min(xs: ReadonlyArray<number>): number {
+// Returns the smallest number in `xs`, or `undefined` if `xs` is empty.
+// Propagates NaN: if any element is NaN the result is NaN (matches
+// Math.min semantics).
+export function min(xs: ReadonlyArray<number>): number | undefined {
+  if (xs.length === 0) return undefined;
   let r = Infinity;
-  for (const x of xs) if (x < r) r = x;
-  return r === Infinity ? 0 : r;
+  for (const x of xs) {
+    if (Number.isNaN(x)) return NaN;
+    if (x < r) r = x;
+  }
+  return r;
 }
 
-export function max(xs: ReadonlyArray<number>): number {
+// Returns the largest number in `xs`, or `undefined` if `xs` is empty.
+// Propagates NaN: if any element is NaN the result is NaN (matches
+// Math.max semantics).
+export function max(xs: ReadonlyArray<number>): number | undefined {
+  if (xs.length === 0) return undefined;
   let r = -Infinity;
-  for (const x of xs) if (x > r) r = x;
-  return r === -Infinity ? 1 : r;
+  for (const x of xs) {
+    if (Number.isNaN(x)) return NaN;
+    if (x > r) r = x;
+  }
+  return r;
 }
--- a/ui/src/base/array_utils.ts
+++ b/ui/src/base/array_utils.ts
@@ -85,3 +85,15 @@ export function intersperse<T, S>(arr: T[], separator: S): (T | S)[] {
     i < arr.length - 1 ? [item, separator] : [item],
   );
 }
+
+export function min(xs: ReadonlyArray<number>): number {
+  let r = Infinity;
+  for (const x of xs) if (x < r) r = x;
+  return r === Infinity ? 0 : r;
+}
+
+export function max(xs: ReadonlyArray<number>): number {
+  let r = -Infinity;
+  for (const x of xs) if (x > r) r = x;
+  return r === -Infinity ? 1 : r;
+}
--- a/ui/src/components/widgets/charts_svg/line_chart_svg.ts
+++ b/ui/src/components/widgets/charts_svg/line_chart_svg.ts
@@ -38,6 +38,7 @@ import {
   pointMarker,
   rangeWithFixedBounds,
 } from './common';
+import {assertIsInstance} from '../../../base/assert';
 
 export type {
   LineChartAttrs,
@@ -115,10 +116,10 @@ export class LineChartSvg implements m.ClassComponent<LineChartAttrs> {
   onupdate({dom, attrs}: m.CVnodeDOM<LineChartAttrs>) {
     // Re-query: Mithril may have replaced the container node if siblings
     // changed (e.g. legend appeared/disappeared).
-    const container = dom.querySelector(
-      '.pf-chart-svg__container',
-    ) as HTMLElement | null;
-    if (!container) return;
+    const container = assertIsInstance(
+      dom.querySelector('.pf-chart-svg__container'),
+      HTMLElement,
+    );
     this.container = container;
     this.currentAttrs = attrs;
     const {width, height} = container.getBoundingClientRect();
@@ -171,9 +172,14 @@ export class LineChartSvg implements m.ClassComponent<LineChartAttrs> {
     } else if (isEmpty) {
       inner = m('.pf-chart-svg__empty', 'No data to display');
     } else {
-      inner = this.renderChart(attrs, data!, width, height);
+      inner = this.renderChart(attrs, data, width, height);
     }
 
+    // Mithril's render() function actually takes a hidden third parameter which
+    // is a function called whenever an event triggers a redraw from within the
+    // rendered vnodes. Here we can use this to hook up the events triggered
+    // from within the rendered SVG code to trigger a global redraw. This is the
+    // same technique used in raf.ts.
     const renderWithRedraw = m.render as (
       el: Element,
       vnodes: m.Children,
@@ -194,14 +200,14 @@ export class LineChartSvg implements m.ClassComponent<LineChartAttrs> {
     // measured dimensions on the very first paint and on every attrs change.
     // The legend is plain Mithril and lives outside the imperative subtree.
     const legend =
-      showLegend && data !== undefined
-        ? renderLegend(
-            data,
-            attrs.formatYValue ?? defaultFmt,
-            this.hiddenSeries,
-            (name: string) => this.toggleSeries(name),
-          )
-        : null;
+      showLegend &&
+      data !== undefined &&
+      renderLegend(
+        data,
+        attrs.formatYValue ?? defaultFmt,
+        this.hiddenSeries,
+        (name: string) => this.toggleSeries(name),
+      );
     const chart = m('.pf-chart-svg__container', {key: 'chart'});
 
     const seriesHover = data?.series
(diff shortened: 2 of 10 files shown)


ui/src/base/array_utils.ts
```diff
@@ -85,3 +85,15 @@ export function intersperse<T, S>(arr: T[], separator: S): (T | S)[] {
...
+export function min(xs: ReadonlyArray<number>): number {
+  let r = Infinity;
+  for (const x of xs) if (x < r) r = x;
+  return r === Infinity ? 0 : r;
+}
+
+export function max(xs: ReadonlyArray<number>): number {
+  let r = -Infinity;
+  for (const x of xs) if (x > r) r = x;
+  return r === -Infinity ? 1 : r;
```
primiano: This is weird

ui/src/base/array_utils.ts
```diff
@@ -85,3 +85,15 @@ export function intersperse<T, S>(arr: T[], separator: S): (T | S)[] {
     i < arr.length - 1 ? [item, separator] : [item],
   );
 }
+
+export function min(xs: ReadonlyArray<number>): number {
+  let r = Infinity;
+  for (const x of xs) if (x < r) r = x;
+  return r === Infinity ? 0 : r;
```
primiano: This is questionable (min([]) == 0)

ui/src/components/widgets/charts_svg/line_chart_svg.ts
```diff
@@ -0,0 +1,794 @@
...
+    // The legend is plain Mithril and lives outside the imperative subtree.
+    const legend =
+      showLegend && data !== undefined
+        ? renderLegend(
+            data,
+            attrs.formatYValue ?? defaultFmt,
+            this.hiddenSeries,
+            (name: string) => this.toggleSeries(name),
+          )
+        : null;
```
primiano: can we avoid nulls and use undefined everywhere?

ui/src/components/widgets/charts_svg/line_chart_svg.ts
```diff
@@ -0,0 +1,794 @@
...
+
+    const tooltip =
+      this.hover !== undefined && seriesHover !== undefined
+        ? renderTooltip(
+            attrs.stacked ? [...seriesHover].reverse() : seriesHover,
+            this.hover.index,
+            attrs.formatXValue ?? defaultFmt,
+            attrs.formatYValue ?? defaultFmt,
+          )
+        : null;
```
primiano: ditto

primiano (approved):
LG with some comments

primiano:
Also a few more comments from AI

  1. Variable shadowing: stacked in buildSeriesPlots (line_chart_svg.ts:~line 1169 of diff)                                                                                                                                                       
                                                                  
  function buildSeriesPlots(series: ..., stacked: boolean): ... {                                                                                                                                                                                 
    // ...                                                                                                                                                                                                                                        
    const stacked: {x: number; y: number; raw: number}[] = [];  // shadows parameter
  The local array stacked shadows the function parameter of the same name. Rename the local to e.g. stackedPoints.                                                                                                                                
                                                                                                                                                                                                                                                  
  2. Variable shadowing: m in estimateLabelWidth (common.ts)                                                                                                                                                                                      
                                                                                                                                                                                                                                                  
  import m from 'mithril';                                        
  // ...                                                                                                                                                                                                                                          
  export function estimateLabelWidth(labels: ReadonlyArray<string>): number {
    let m = 0;  // shadows the mithril import                                                                                                                                                                                                     
  This shadows the m (mithril) import at module scope. Use maxW or similar.                                                                                                                                                                       
                                                                                                                                                                                                                                                  
  3. min()/max() defaults are surprising for a general utility (array_utils.ts)                                                                                                                                                                   
                                                                                                                                                                                                                                                  
  min([]) returns 0 and max([]) returns 1 — these are chart-specific sentinel values (ensuring a non-degenerate axis range), but they live in a general-purpose utility file. A caller expecting mathematical semantics would be surprised.       
  Consider either:                                                                                                                                                                                                                                
  - Moving these into charts_svg/common.ts where the chart-specific defaults make sense, or                                                                                                                                                       
  - Making the empty-array behavior return NaN or -Infinity/Infinity and handling defaults at the call site.                                                                                                                                      
                                                                                                            
  4. Removed filterPointsWithInterpolation changes ECharts demo behavior (charts_demo.ts)                                                                                                                                                         
                                                                                                                                                                                                                                                  
  The brush-filter path for the existing LineChart (ECharts) demo now uses xAxisMin/xAxisMax instead of filtering points with interpolation. This changes existing behavior — the old code produced boundary-interpolated subsets so the line     
  didn't abruptly start/end inside the chart. Worth confirming this doesn't regress the ECharts chart's brush experience.                                                                                                                         
                                                                                                                                                                                                                                                  
  5. O(n) nearest-point search in handlePointerMove (line_chart_svg.ts)                                                                                                                                                                           
   
  The linear scan to find the nearest X index on every mouse move works fine for small datasets but could jank with thousands of points. Since points are sorted by X, a binary search would be a straightforward improvement:                    
  // Binary search for nearest x                                  
  let lo = 0, hi = pts.length - 1;                                                                                                                                                                                                                
  while (lo < hi) {                                                                                                                                                                                                                               
    const mid = (lo + hi) >> 1;
    if (pts[mid].x < xVal) lo = mid + 1; else hi = mid;                                                                                                                                                                                           
  }                                                                                                                                                                                                                                               
   
  6. Imperative m.render() + redraw cast (line_chart_svg.ts)                                                                                                                                                                                      
                                                                                                                                                                                                                                                  
  const renderWithRedraw = m.render as (
    el: Element, vnodes: m.Children, redraw?: () => void,                                                                                                                                                                                         
  ) => void;                                                                                                                                                                                                                                      
  renderWithRedraw(container, inner, m.redraw);                                                                                                                                                                                                   
  The type cast to expose Mithril's undocumented third redraw parameter is fragile — it could break on a Mithril upgrade. Worth a comment noting why this is necessary and which Mithril version it targets. Alternatively, consider              
  requestAnimationFrame(() => m.redraw()) as a more documented approach if the redraw parameter is only to ensure event handlers trigger redraws.                                                                                                 
                                                                                                                                                                                                                                                  
  7. Missing SVG key attributes on repeated elements                                                                                                                                                                                              
                                                                                                                                                                                                                                                  
  Gridlines, tick marks, and tick labels are generated via .map() without key attributes. Mithril's keyed diff is more predictable for lists that change length (e.g., when the axis range updates after a brush). Adding {key: t} (the tick      
  value) would avoid potential DOM reuse bugs.                                                                                                                                                                                                    
                                                                                                                                                                                                                                                  
  8. data! non-null assertion (line_chart_svg.ts)                                                                                                                                                                                                 
   
  inner = this.renderChart(attrs, data!, width, height);                                                                                                                                                                                          
  The control flow above guarantees data is defined here, but the ! is still a smell. A simple const d = data; if (d !== undefined) guard would be clearer.

ui/src/base/array_utils.ts
stevegolton: Modified to return undefined when passed an empty array. Added unit tests.

ui/src/base/array_utils.ts
stevegolton: As above.

ui/src/components/widgets/charts_svg/line_chart_svg.ts
stevegolton: As above

ui/src/components/widgets/charts_svg/line_chart_svg.ts
stevegolton: Avoided having to specify null or undefined in favour of the short circuit pattern:

```ts
condition && m(...)
```

stevegolton:
> Also a few more comments from AI
> 
> 1. Variable shadowing: stacked in buildSeriesPlots (line_chart_svg.ts:~line 1169 of diff)
> 
> function buildSeriesPlots(series: ..., stacked: boolean): ... { // ... const stacked: {x: number; y: number; raw: number}[] = []; // shadows parameter The local array stacked shadows the function parameter of the same name. Rename the local to e.g. stackedPoints.

This seems very nitpicky, but I guess it's valid. I've renamed the variable.

> 2. Variable shadowing: m in estimateLabelWidth (common.ts)
> 
> import m from 'mithril'; // ... export function estimateLabelWidth(labels: ReadonlyArray): number { let m = 0; // shadows the mithril import This shadows the m (mithril) import at module scope. Use maxW or similar.

I'm not sure this is a real issue. It's a tiny function, and if one tried to use m(...) as a function typescript would immediately complain. The real offense here is the single char variable names which I've fixed.

> 3. min()/max() defaults are surprising for a general utility (array_utils.ts)
> 
> min([]) returns 0 and max([]) returns 1 — these are chart-specific sentinel values (ensuring a non-degenerate axis range), but they live in a general-purpose utility file. A caller expecting mathematical semantics would be surprised. Consider either:
> 
> * Moving these into charts_svg/common.ts where the chart-specific defaults make sense, or
> * Making the empty-array behavior return NaN or -Infinity/Infinity and handling defaults at the call site.

Valid - as discussed above, I've made these more generic returning undefined for an empty array.

> 4. Removed filterPointsWithInterpolation changes ECharts demo behavior (charts_demo.ts)
> 
> The brush-filter path for the existing LineChart (ECharts) demo now uses xAxisMin/xAxisMax instead of filtering points with interpolation. This changes existing behavior — the old code produced boundary-interpolated subsets so the line didn't abruptly start/end inside the chart. Worth confirming this doesn't regress the ECharts chart's brush experience.

This is intentional - it was very weird they way points are being interpolated and it obscures how the charts truly behave when zoomed in.

> 5. O(n) nearest-point search in handlePointerMove (line_chart_svg.ts)
> 
> The linear scan to find the nearest X index on every mouse move works fine for small datasets but could jank with thousands of points. Since points are sorted by X, a binary search would be a straightforward improvement: // Binary search for nearest x let lo = 0, hi = pts.length - 1; while (lo < hi) { const mid = (lo + hi) >> 1; if (pts[mid].x < xVal) lo = mid + 1; else hi = mid; }

Smells like preemptive optimization to me. If we have a large number of points, I suspect we'll run into Mithril rendering / DOM issues long before we need a binary search here. Also, this assumes that points are sorted by x which I'm not sure I want to make a requirement right now. Will address in a future update if it becomes a problem.

> 6. Imperative m.render() + redraw cast (line_chart_svg.ts)
> 
> const renderWithRedraw = m.render as ( el: Element, vnodes: m.Children, redraw?: () => void, ) => void; renderWithRedraw(container, inner, m.redraw); The type cast to expose Mithril's undocumented third redraw parameter is fragile — it could break on a Mithril upgrade. Worth a comment noting why this is necessary and which Mithril version it targets. Alternatively, consider requestAnimationFrame(() => m.redraw()) as a more documented approach if the redraw parameter is only to ensure event handlers trigger redraws.

This is intentional and required for auto redraw system to work from events triggering from within the SVG. Adding a comment makes sense though.

> 7. Missing SVG key attributes on repeated elements
> 
> Gridlines, tick marks, and tick labels are generated via .map() without key attributes. Mithril's keyed diff is more predictable for lists that change length (e.g., when the axis range updates after a brush). Adding {key: t} (the tick value) would avoid potential DOM reuse bugs.

Valid, but keying in mithril can be messy - mithril can be very picky when certain children are keyed and others are not and it's very difficult to track down issues. I generally avoid unless we're seeing real issues.

> 8. data! non-null assertion (line_chart_svg.ts)
> 
> inner = this.renderChart(attrs, data!, width, height); The control flow above guarantees data is defined here, but the ! is still a smell. A simple const d = data; if (d !== undefined) guard would be clearer.

Valid, the bang "!" was not actually required. Removed.

PR #5784, opened 2026-05-08
protovm: add compiler

rukkal:
protovm: add compiler

Introduces a compiler for ProtoVM that translates high-level configuration commands into
low-level bytecode instructions (VmProgram).

CompileConfig proto:
- Defines a sequence of high-level commands.

CLI:
- Reads CompileConfig textproto from stdin and outputs the compiled binary VmProgram
  to stdout.

Internal Compiler and InstructionEmitter classes:
- Parse compile configs (textproto).
- Resolve human-readable paths (field names) into field IDs using TracePacket and
  WinscopeExtensions descriptors.
- Convert high-level commands into the corresponding sequence of VmInstruction.

### DIFF
--- a/protos/perfetto/protovm/compile_config.proto
+++ b/protos/perfetto/protovm/compile_config.proto
@@ -18,10 +18,13 @@ syntax = "proto2";
 
 package perfetto.protos;
 
+// The configuration for the ProtoVM compiler.
+// It defines a sequence of high-level compilation commands that the compiler
+// translates into low-level instructions (VmProgram/VmInstruction).
 message CompileConfig {
-  // The fully qualified name of the root message. E.g.
-  // "perfetto.protos.TracePacket". All the paths in the compiler commands are
-  // relative to this message.
+  // The fully qualified name of the root message. The paths in the compiler
+  // commands are relative to this message. This is tipically set to
+  // "perfetto.protos.TracePacket".
   optional string root_message = 1;
   repeated CompileCommand commands = 2;
 }
--- a/protos/perfetto/protovm/compile_config.proto
+++ b/protos/perfetto/protovm/compile_config.proto
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+message CompileConfig {
+  repeated Command commands = 1;
+}
+
+message Command {
+  oneof command {
+    DelCommand del = 1;
+    EnterScopeCommand enter_scope = 2;
+    SetCommand set = 3;
+    MergeCommand merge = 4;
+  }
+
+  enum AbortLevel {
+    // Skip current command but execute following ones
+    SKIP_CURRENT_COMMAND = 1;
+    // Skip current command as well as following ones (default)
+    SKIP_CURRENT_COMMAND_AND_BREAK_OUTER = 2;
+    // Abort whole program
+    ABORT = 3;
+  };
+  optional AbortLevel abort_level = 5;
+}
+
+message DelCommand {
+  repeated string src = 1;
+  repeated string dst = 2;
+  optional bool if_src_present = 3;
+  optional string dst_key_field = 4;
+}
+
+message EnterScopeCommand {
+  repeated string src = 1;
+  repeated string dst = 2;
+  repeated Command commands = 3;
+}
+
+message SetCommand {
+  repeated string src = 1;
+  repeated string dst = 2;
+}
+
+message MergeCommand {
+  repeated string src = 1;
+  repeated string dst = 2;
+  optional bool recursive = 3;
+  optional string key_field = 4;
+}
--- a/protos/perfetto/protovm/compile_config.proto
+++ b/protos/perfetto/protovm/compile_config.proto
@@ -19,15 +19,19 @@ syntax = "proto2";
 package perfetto.protos;
 
 message CompileConfig {
-  repeated Command commands = 1;
+  // The fully qualified name of the root message. E.g.
+  // "perfetto.protos.TracePacket". All the paths in the compiler commands are
+  // relative to this message.
+  optional string root_message = 1;
+  repeated Command commands = 2;
 }
 
 message Command {
   oneof command {
     DelCommand del = 1;
     EnterScopeCommand enter_scope = 2;
-    SetCommand set = 3;
-    MergeCommand merge = 4;
+    MergeCommand merge = 3;
+    SetCommand set = 4;
   }
 
   enum AbortLevel {
@@ -54,14 +58,14 @@ message EnterScopeCommand {
   repeated Command commands = 3;
 }
 
-message SetCommand {
-  repeated string src = 1;
-  repeated string dst = 2;
-}
-
 message MergeCommand {
   repeated string src = 1;
   repeated string dst = 2;
   optional bool recursive = 3;
   optional string key_field = 4;
+}
+
+message SetCommand {
+  repeated string src = 1;
+  repeated string dst = 2;
 }
--- a/protos/perfetto/protovm/compile_config.proto
+++ b/protos/perfetto/protovm/compile_config.proto
@@ -23,15 +23,15 @@ message CompileConfig {
   // "perfetto.protos.TracePacket". All the paths in the compiler commands are
   // relative to this message.
   optional string root_message = 1;
-  repeated Command commands = 2;
+  repeated CompileCommand commands = 2;
 }
 
-message Command {
+message CompileCommand {
   oneof command {
-    DelCommand del = 1;
-    EnterScopeCommand enter_scope = 2;
-    MergeCommand merge = 3;
-    SetCommand set = 4;
+    CompileCommandDel del = 1;
+    CompileCommandEnterScope enter_scope = 2;
+    CompileCommandMerge merge = 3;
+    CompileCommandSet set = 4;
   }
 
   enum AbortLevel {
@@ -45,27 +45,27 @@ message Command {
   optional AbortLevel abort_level = 5;
 }
 
-message DelCommand {
+message CompileCommandDel {
   repeated string src = 1;
   repeated string dst = 2;
   optional bool if_src_present = 3;
   optional string dst_key_field = 4;
 }
 
-message EnterScopeCommand {
+message CompileCommandEnterScope {
   repeated string src = 1;
   repeated string dst = 2;
-  repeated Command commands = 3;
+  repeated CompileCommand commands = 3;
 }
 
-message MergeCommand {
+message CompileCommandMerge {
   repeated string src = 1;
   repeated string dst = 2;
   optional bool recursive = 3;
   optional string key_field = 4;
 }
 
-message SetCommand {
+message CompileCommandSet {
   repeated string src = 1;
   repeated string dst = 2;
 }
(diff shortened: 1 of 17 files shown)


protos/perfetto/protovm/compile_config.proto
```diff
@@ -0,0 +1,67 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
```
primiano, line 20: add some comments to explain what this is about?

src/protovm/compiler/BUILD.gn
```diff
@@ -0,0 +1,86 @@
...
+    "../../base",
+  ]
+}
+
+perfetto_cc_proto_descriptor("gen_cc_trace_descriptor") {
+  descriptor_name = "trace.descriptor"
+  descriptor_target = "../../../protos/perfetto/trace:descriptor"
+}
+
+perfetto_cc_proto_descriptor("gen_cc_winscope_descriptor") {
```
primiano: isn't this going to be included in trace_descriptor above? or isn't because of the lazy imports?

src/protovm/compiler/compiler.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "perfetto/protozero/scattered_heap_buffer.h"
+#include "src/protovm/compiler/compile_config.descriptor.h"
+#include "src/protovm/compiler/trace.descriptor.h"
+#include "src/protovm/compiler/winscope.descriptor.h"
+#include "src/protozero/text_to_proto/text_to_proto.h"
+
+namespace perfetto::protovm {
+
+Compiler::Compiler() : emitter_(&pool_) {
+  pool_.AddFromFileDescriptorSet(perfetto::kTraceDescriptor.data(),
```
primiano: rather than building the descriptors in the tool, I think it woudl be more flebible if we depend on //gn:protoc_lib and we allow passing .proto files from the cmdline.
This makes the compiler more flexible, because you can just point it to the proto you need, rather than assuming "I'm going to need winscope so i am going to bundle them inside the tool"

src/protovm/compiler/compiler.h
```diff
@@ -0,0 +1,86 @@
...
+#include "perfetto/base/status.h"
+#include "perfetto/ext/base/status_or.h"
+#include "protos/perfetto/protovm/compile_config.pbzero.h"
+#include "protos/perfetto/protovm/vm_program.pbzero.h"
+#include "src/protovm/compiler/instruction_emitter.h"
+#include "src/trace_processor/util/descriptors.h"
+
+namespace perfetto::protovm {
+
+// Implements the compiler front-end.
```
primiano, line 33: can we have README.md (or a /docs/design-doc) that explains what this compiler does. Just ask AI to turn your doc + code here into a README.md for here.

src/protovm/compiler/compiler.h
```diff
@@ -0,0 +1,86 @@
...
+//
+// Parses high-level configuration commands (from CompileConfig) and delegates
+// the conversion and emission of low-level bytecode instructions (VmProgram)
+// to the InstructionEmitter.
+class Compiler {
+ public:
+  Compiler();
+
+  base::StatusOr<std::string> Compile(std::string_view config_textproto);
```
primiano: add a comment saying  which message config_textproto is supposed to match

protos/perfetto/protovm/compile_config.proto
primiano, line 20: can you add a comment / explain what these protos are for? and how they are different from vm_program.proto?

src/protovm/compiler/BUILD.gn
```diff
@@ -0,0 +1,86 @@
...
+perfetto_host_executable("protovm_compiler") {
+  sources = [ "main.cc" ]
+  deps = [
+    ":compiler",
+    "../../../gn:default_deps",
+    "../../base",
+  ]
+}
+
+perfetto_cc_proto_descriptor("gen_cc_trace_descriptor") {
```
primiano, line 51: can we avoid all tehse descriptors and instead import .proto files at runtime? see my other comment about using protoc_lib

src/protovm/compiler/compiler.cc
primiano: Rather than assuming we have the descriptors built into the binary, pass them on the cmdline, and use protoc_lib to dynamically turn them into descriptros.

This is an offline tool and there is no problem depending on the full protobuf library.
THe pro is that it works with any descriptor whatsoever, rather than baking into assumptions on which proto you'll need

src/protovm/compiler/compiler.h
primiano: add a comment saying which protobuf message this config textproto is for

src/protovm/compiler/compiler.h
```diff
@@ -0,0 +1,86 @@
...
+                                      it->size()));
+    }
+    return path;
+  }
+
+  AbortLevel GetAbortLevel(
+      const protos::pbzero::Command::Decoder& command) const;
+
+  perfetto::trace_processor::DescriptorPool pool_;
+  const perfetto::trace_processor::ProtoDescriptor* root_proto_;
```
primiano: = nullptr

src/protovm/compiler/main.cc
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+#include <cstdio>
+#include <string_view>
+
+#include "perfetto/ext/base/file_utils.h"
+#include "src/protovm/compiler/compiler.h"
+
+int ProcessArgs(int argc, char** argv);
```
primiano: put this in the anon namespace

src/protovm/compiler/main.cc
```diff
@@ -0,0 +1,67 @@
...
+    return 1;
+  }
+
+  std::fwrite(status_or_program->data(), 1, status_or_program->size(), stdout);
+
+  return 0;
+}
+
+int ProcessArgs(int argc, char** argv) {
+  if (argc == 1) {
```
primiano: hmm no this should be ==2 as you dereferencing argv[1] below

src/protovm/compiler/main.cc
```diff
@@ -0,0 +1,67 @@
...
+
+int ProcessArgs(int argc, char** argv);
+
+// Implements the ProtoVM compiler CLI
+//
+// Reads a CompileConfig textproto from stdin and outputs the compiled binary
+// VmProgram to stdout.
+int main(int argc, char** argv) {
+  if (argc > 1) {
+    ProcessArgs(argc, argv);
```
primiano: isn't this missing a return? like return ProcessARgs(argc, argv)?

But also not sure the split function adds more, IMHO  only adds  confusion while reading (and bugs liek this)

src/protovm/compiler/BUILD.gn
```diff
@@ -0,0 +1,86 @@
...
+    "../../../protos/perfetto/protovm:zero",
+    "../../base",
+    "../../protozero",
+    "../../protozero/text_to_proto",
+    "../../trace_processor/util:descriptors",
+  ]
+}
+
+perfetto_host_executable("protovm_compiler") {
+  sources = [ "main.cc" ]
```
primiano, line 41: if you make it testonly=true you can depend on protoc lib

primiano (commented):
THe code makes sense, my main suggestion is:
1. use protoc_lib to pass the .proto file from cmdline rather than baking assumptions on which protos/descriptors you are going to need (or find a way to bundle them all, not just the winscope descriptor)

2. Some more comments would help for future reading

the rest looks great

src/protovm/compiler/compiler.cc
rukkal: I pushed a (dirty to be cleaned up) commit that implements this CLI:
```
Usage: protovm_compiler [-I <proto_path>] <file1.proto> ...

Reads ProtoVM CompileConfig textproto from stdin, compiles it and
outputs a VmProgram binary on stdout.
```

main.cc uses protoc_lib to load the descriptors. The descriptors are then passed to Compiler as opaque bytes. Compiler still uses `trace_processor::DescriptorPool` to load the descriptors, i.e. Compiler doesn't depend on protoc_lib so the unit tests (part of perfetto_unittests) don't depend on protoc_lib either.

Lmk if it all makes sense and then I'll polish the implementation

src/protovm/compiler/main.cc
rukkal: got rid of it

src/protovm/compiler/compiler.cc
rukkal: As discussed offline, I kept this solution and polished it.

rukkal:
@primiano I pushed a few commits with:

- Dynamic proto loading: specify proto files as CLI args and use protoc_lib to load descriptors (used to have hardcoded TracePacket and WinscopeExtensions descriptors).
- README.md
- More comments here and there where requested
- Better prefixes in the proto message names to avoid clashes

src/protovm/compiler/compiler.cc
```diff
@@ -0,0 +1,159 @@
...
+  return emitter_.Set(scope, src_path, dst_path, GetAbortLevel(command));
+}
+
+base::Status Compiler::ParseDel(
+    const InstructionEmitter::Scope& scope,
+    const protos::pbzero::CompileCommand::Decoder& command) const {
+  protos::pbzero::CompileCommandDel::Decoder del(command.del());
+  auto src_path = ParsePath(del.src());
+  auto dst_path = ParsePath(del.dst());
+  if (!(del.if_src_present() ^ del.has_dst_key_field())) {
```
primiano, line 100: is this correct?
del.if_src_present() reads the value of the bool field (defaults to false when absent), while del.has_dst_key_field() checks presence.
If neither is set, false ^ false == false, so the validation passes by accident. 
FUrthermore, if if_src_present is explicitly set to false and dst_key_field is set, the XOR becomes false ^ true == true which passes validation, then falls
  through both if branches to return "Unsupported Del command variant" no?

src/protovm/compiler/instruction_emitter.cc
```diff
@@ -0,0 +1,297 @@
...
+}
+
+base::Status InstructionEmitter::MergeByKey(
+    const Scope& scope,
+    const std::vector<std::string_view>& src_relative_path,
+    const std::vector<std::string_view>& dst_relative_path,
+    std::string_view key_field_name,
+    bool is_recursive,
+    AbortLevel abort_level) const {
+  static constexpr uint32_t REGISTER_ID = 0;
```
primiano, line 193: You always use register 0 both here and delete by key. what happens if these instructions are nested in other ops? can'tt hey clobber each other? or are they mutually exclusive?

primiano (approved):
LGTM with a vew questions

src/protovm/compiler/compiler.cc
rukkal: I think it's correct, but maybe worth to break it down into two distinct checks to make it clearer:
```
if (del_if_src_present && has_dst_key_field) {
   error "can't specify both"
}
if (!del_if_src_present && !has_dst_key_field) {
   error "must specify at least one"
}
```

> If neither is set, false ^ false == false, so the validation passes by accident.

Note that the xor is negated. So !(false ^ false) == true -> error is raised

> FUrthermore, if if_src_present is explicitly set to false and dst_key_field is set, the XOR becomes false ^ true == true which passes validation, then falls

In this case the validation passes, correct. But then only the branch `if (del.has_dst_key_field())` is executed, which is fine.

src/protovm/compiler/instruction_emitter.cc
rukkal: So far it's fine, because we execute these blocks "atomically":
```
src cursor = <some path>
{
   src cursor = <field>
   load R0
}
dst cursor = <some path reading R0>
// Note that now R0 can be overwritten
```

"atomically" in the sense that there won't be other instructions interleaved which could interfere with R0

src/protovm/compiler/compiler.cc
primiano: ah ack then no need to cange, I think i was misreading the code. It's a bit confusing because has_dst_key_field seems to act only on the presence of the proto field while the other one was acting on the value.

src/protovm/compiler/instruction_emitter.cc
primiano: ack!

PR #5785, opened 2026-05-08
ui: Memscope - Live dashboard page

stevegolton:
This patch adds a new `dev.perfetto.Memscope` UI plugin: a live memory dashboard that connects to an Android (ADB / WDP) or Linux (websocket) target, starts a lightweight trace, and renders periodically refreshed snapshots across four tabs (Processes, System, File Cache, Pressure & Swap).

NOTE: In order to keep the PRs as small as possible, this is just the first part of the memscope plugin which only features the dashboard - no profiling mode is implemented yet.

## Key bits

- Live tracing session (sessions/live_session.ts): low-overhead monitoring config (process_stats, sys_stats, ftrace). Polls snapshots on an interval, parsing and extracting key counter information.
- Connection view (views/connection.ts): Landing page when no session is currently running. Allows the user to select a device protocol, and, if relevant, presents a device list to allow device selection.
- Dashboard view: (views/dashboard.ts): Shows up when the user connects to a device and starts recording a live tracing session. Displays a toolbar showing the status of the live session and buttons to control the session, and the four tabs with a tab switcher (SegmentedButton).
- Tabs:
  - Processes tab (views/tabs/processes.ts): Shows a few header cards showing high level system memory usage, a chart showing the sum of all processes grouped by category (arbitrary list defined in process_categories.ts), and a table of all processes with various stats spread across various columns.
  - System tab (views/tabs/system.ts): Shows various system level memory stats across billboards and a chart.
  - Page cache tab (views/tabs/page_cache.ts): Show page cache stats including usage and activity charts.
  - Pressure, swap and LMK tab (views/tabs/pressure_swap.ts): Displays swap usage, memory pressure and LMK events.
- Process categorization (process_categories.ts): maps Android process names → high-level categories (Framework, HAL, System UI, …). Colors are assigned positionally from the chart palette so adding categories doesn't require picking hex.

## Usage / testing:
- Enable the dev.perfetto.Memscope plugin & restart the page.
- Click the memscope sidebar entry.
- Connect to your Android/Linux device of choice.
- Verify stats look correct. Press pause and play again to ensure the snapshots stop updating when paused.
- Switching tabs should stop the periodic updates (and period redraws).
- Stop & Open Trace produces a working trace in the main UI.
- Verify category chips and OOM-bucket chips read well in both light and dark themes

### DIFF
--- a/ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
+++ b/ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
@@ -1,0 +1,646 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import protos from '../../../protos';
+import {uuidv4} from '../../../base/uuid';
+import {deferChunkedTask} from '../../../base/chunked_task';
+import {NUM, NUM_NULL, STR} from '../../../trace_processor/query_result';
+import {WasmEngineProxy} from '../../../trace_processor/wasm_engine_proxy';
+import {AdbDevice} from '../../dev.perfetto.RecordTraceV2/adb/adb_device';
+import {createAdbTracingSession} from '../../dev.perfetto.RecordTraceV2/adb/adb_tracing_session';
+import {TracingSession} from '../../dev.perfetto.RecordTraceV2/interfaces/tracing_session';
+import {TracedWebsocketTarget} from '../../dev.perfetto.RecordTraceV2/traced_over_websocket/traced_websocket_target';
+import {ConnectionResult} from '../views/connection';
+
+const SNAPSHOT_INTERVAL_MS = 3_000; // How over to take a snapshot of the runnign trace and extract data.
+const INITIAL_SNAPSHOT_INTERVAL_MS = 1_000; // Use a shorter interval for the first snapshot to get data on screen faster.
+const POLLING_INTERVAL_MS = 1_000; // Recording config polling interval for process stats and system stats.
+
+let engineCounter = 0;
+
+export interface TsValue {
+  ts: number;
+  value: number;
+}
+
+export interface LmkEvent {
+  ts: number;
+  pid: number;
+  processName: string;
+  oomScoreAdj: number;
+}
+
+export interface ProcessInfo {
+  upid: number;
+  pid: number;
+  processName: string;
+  startTs: number | null;
+  debuggable: boolean;
+}
+
+export interface SnapshotData {
+  // counterName → sorted array of {ts, value}
+  systemCounters: Map<string, TsValue[]>;
+
+  // upid → counterName → sorted array of {ts, value}
+  processCountersByUpid: Map<number, Map<string, TsValue[]>>;
+
+  // upid → metadata
+  processInfo: Map<number, ProcessInfo>;
+
+  // LMK events sorted by ts.
+  lmkEvents: LmkEvent[];
+
+  // Whether the device is a userdebug build.
+  isUserDebug: boolean;
+
+  // Trace start timestamp from trace_bounds, used as the x=0 origin for all charts.
+  ts0: number;
+
+  // X-axis bounds in seconds relative to ts0, derived from the earliest/latest
+  // counter sample. Use these as xAxisMin/xAxisMax on all line charts.
+  xMin: number;
+  xMax: number;
+}
+
+export type OnSnapshotCallback = (data: SnapshotData) => void;
+
+/**
+ * Connects to a device or Linux host, starts a memory tracing session, and
+ * polls snapshots on a timer. Each snapshot clones the running trace, loads
+ * it into a reusable TraceProcessor engine, extracts structured data, and
+ * notifies registered callbacks.
+ */
+export class LiveSession {
+  private session?: TracingSession;
+  private engine?: WasmEngineProxy;
+  private readonly sessionName: string;
+  private readonly device?: AdbDevice;
+  private readonly linuxTarget?: TracedWebsocketTarget;
+  private timer?: ReturnType<typeof setTimeout>;
+  private snapshotInFlight = false;
+  private isDisposed = false;
+  private readonly onSnapshotCallbacks: OnSnapshotCallback[] = [];
+
+  readonly deviceName: string;
+  snapshotCount = 0;
+  status = 'Starting tracing session...';
+  isPaused = false;
+  data?: SnapshotData;
+  lastTraceBuffer?: ArrayBuffer;
+
+  // Timing breakdown for the last snapshot (ms).
+  lastSnapshotMs = 0;
+  lastSnapshotSizeKb = 0;
+  lastBufferUsagePct?: number;
+  lastCloneMs = 0;
+  lastParseMs = 0;
+  lastQueryMs = 0;
+  lastExtractMs = 0;
+  earliestEventTs = 0;
+
+  // True when the last snapshot took longer than the configured interval.
+  snapshotOverrun = false;
+
+  constructor(conn: ConnectionResult) {
+    this.device = conn.device;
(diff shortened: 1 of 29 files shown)


ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
```diff
@@ -0,0 +1,606 @@
...
+      this.lastSnapshotMs = Math.round(performance.now() - t0);
+      this.lastSnapshotSizeKb = this.lastTraceBuffer.byteLength / 1024;
+      this.lastBufferUsagePct = bufferUsagePct;
+      this.snapshotCount++;
+      this.status = `Snapshot #${this.snapshotCount} OK (${this.lastSnapshotMs}ms)`;
+    } catch (e) {
+      this.status = `Snapshot error: ${e}`;
+    } finally {
+      this.snapshotInFlight = false;
+      if (!this.isPaused) {
```
primiano: out of pure curiosity why did you did this (i.e. manually triggering each snapshot) rather than going with setInterval (with an asyncguard / already-running check)?
Not a big deal but it feels you are reinventing setInternal in some way

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
```diff
@@ -0,0 +1,606 @@
...
+              protos.MeminfoCounters.MEMINFO_DIRTY,
+              protos.MeminfoCounters.MEMINFO_WRITEBACK,
+            ],
+          },
+        },
+      },
+      {
+        config: {
+          name: 'linux.sys_stats',
+          targetBuffer: 3,
```
primiano: why this is in a different buffer? In my mind all the things polled at same rate shoudld go into the same ring buffer, otherwise you get asymmetries in the timelines of the two buffers

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
```diff
@@ -0,0 +1,606 @@
...
+}
+
+function createMonitoringConfig(
+  uniqueSessionName: string,
+): protos.ITraceConfig {
+  return {
+    uniqueSessionName,
+    buffers: [
+      {
+        sizeKb: 4 * 1024,
```
primiano, line 267: pro-tip you can set a name here (e.g. name: "initial_pkg_list_and_stats")
and then down below next to targetBuffer:0 you can say targetBufferName: "initial_pkg_list_and_stats"

In order to be backward compatible with both older versions of android an newever versions you need to set both the targetBuffer index and the targetBufferByName.

But I think it makes the code a bit easier to read and reason about.

ui/src/plugins/dev.perfetto.Memscope/views/connection.ts
```diff
@@ -0,0 +1,576 @@
...
+  ADB_DEVICE_FILTER,
+  getAdbWebUsbInterface,
+} from '../../dev.perfetto.RecordTraceV2/adb/webusb/adb_webusb_utils';
+import {TracedWebsocketTarget} from '../../dev.perfetto.RecordTraceV2/traced_over_websocket/traced_websocket_target';
+import {AsyncWebsocket} from '../../dev.perfetto.RecordTraceV2/websocket/async_websocket';
+
+export interface ConnectionResult {
+  device?: AdbDevice;
+  deviceName: string;
+  linuxTarget?: TracedWebsocketTarget;
```
primiano, line 43: why this is called "Linux" target? or is it as in "Linux and Android"?
a bit confusing because generally in our codebase when we say linux tends to be "linux desktop but not android" (we inherited this from chromium)

ui/src/plugins/dev.perfetto.Memscope/views/connection.ts
```diff
@@ -0,0 +1,576 @@
...
+        icon: 'computer',
+        variant: ButtonVariant.Filled,
+        intent: Intent.Primary,
+        disabled: this.linuxConnecting || this.linuxUrl.trim() === '',
+        onclick: () => this.connectLinux(attrs),
+      }),
+    ];
+  }
+
+  private async connectDevice(attrs: ConnectionPageAttrs) {
```
primiano, line 262: Hmm I feel a lot of code here and below could be avoided if you reuse recording_manager.ts to listTargets, listProvidersForCurrentPlatform and so on.
How does this even work if you have >1 devices?

IMHO there should be some more consistency with recording_manager (unless there is  a reason why you did it this way i'm not thinking about?)

primiano (approved):
THanks a lot for this. So most of the code LGTM, i only have some question on connection.ts

BUt i'd be fine with landing this and addressing them in a separate PR to avoid too much churn.

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
stevegolton: It's because we do async work (queries) in the callback. If the callback ends up taking longer than the interval, then the setInterval callback would fire again before we've completed processing the previous snapshot, which would end up running engines in parallel, overlapping snapshot handling, etc...

This way, we wait until all queries have completed before calling setTimeout so there's no chance of overlapping.

This was more of a problem previously before I improved the query efficiency. Nowadays, I don't expect the queries to take longer than the timeout under normal conditions. However, there are many variables that can affect this such as activity on the target device and the client machine performance. If someone is recording a busy phone on a celeron chromebook, I still want the UI to be somewhat usable.

Edit: Just read your comment properly re asyncguard. Yeah, could do - though I feel like what I'm doing here is more correct as it avoids inconsistent timing intervals between snapshots, and is about the same level of complexity as setInterval with a flag

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
stevegolton: Nice tip! I'm not a fan of having to manually index into these buffers to reason about the config. I'll make the change.

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
stevegolton: IIRC - it led to one datasource pushing the other one out of the buffer entirely, but I could be wrong. I'll test it out in the same buffer and see if I can reproduce the problem.

ui/src/plugins/dev.perfetto.Memscope/views/connection.ts
stevegolton: 'Linux' in this case does not mean android - it means the user has selected a linux target (i.e. not ADB) in the connection panel. The connection panel returns either an `AdbDevice` OR a `TracedWebsocketTarget` depending on what the user selected. IIRC these are handled slightly differently by the caller so we cannot return a simple generic 'device' regardless of what the user selected - though it's possible that you actually can and I've massively over complicated this.

The interface here is misleading - given the 'either or' relationship. The correct shape for this type would be something like a tagged union:

```ts
type ConnectionResult = {
  kind: 'adb',
  device: AdbDevice;
  ...
 } | {
  kind: 'linux',
  linuxTarget: TracedWebsocketTarget;
  ...
};
```

ui/src/plugins/dev.perfetto.Memscope/views/connection.ts
stevegolton: Fair comment. I'll look at sharing more code with the recording page.

I haven't testing with multiple devices. I didn't think this was a valid use case for the dashboard? Or are you referring to the case where you have multiple devices connected and you want to select one of them to record from?

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
stevegolton: Testing now, it seems to be fine. Good spot, thanks.

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
stevegolton: Done

ui/src/plugins/dev.perfetto.Memscope/sessions/live_session.ts
stevegolton: Tentatively marking this as resolved for now. If it's a problem I'll address in a future patch.

ui/src/plugins/dev.perfetto.Memscope/views/connection.ts
stevegolton: Deferring to a future patch.

ui/src/plugins/dev.perfetto.Memscope/views/connection.ts
stevegolton: Deferring to a future patch.

PR #5786, opened 2026-05-08
ui: Memscope - Profiling support

stevegolton:
This patch adds profiling support to the memscope dashboard.

- Addition of a new profile 'action' button along side each debuggable process in the process table.
- Profile session (sessions/profile_session.ts): Starts a new Java + native heap tracing session with an opinionated configuration recording into ring buffers.
- Profiling page (views/profile_page.ts): Displays the current status of the profiling session as well as some process stats taken from the live session data pertinent to the process being profiled. Features a cancel and stop & open button which stops and opens the trace in perfetto, switching back to the main timeline.

image

image

### DIFF
--- a/ui/src/plugins/dev.perfetto.Memscope/sessions/profile_session.ts
+++ b/ui/src/plugins/dev.perfetto.Memscope/sessions/profile_session.ts
@@ -1,0 +1,188 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import protos from '../../../protos';
+import {AdbDevice} from '../../dev.perfetto.RecordTraceV2/adb/adb_device';
+import {createAdbTracingSession} from '../../dev.perfetto.RecordTraceV2/adb/adb_tracing_session';
+import {TracingSession} from '../../dev.perfetto.RecordTraceV2/interfaces/tracing_session';
+import {TracedWebsocketTarget} from '../../dev.perfetto.RecordTraceV2/traced_over_websocket/traced_websocket_target';
+
+const DUMP_INTERVAL_MS = 10_000;
+const PROC_STATS_BUFFER_SIZE_KB = 4 * 1024;
+const HEAPPROFD_BUFFER_SIZE_KB = 128 * 1024;
+const JAVA_HPROF_BUFFER_SIZE_KB = 256 * 1024;
+const STATS_POLL_INTERVAL_MS = 3000;
+
+export type ProfileState = 'recording' | 'stopping' | 'finished' | 'error';
+
+export class ProfileSession {
+  readonly pid: number;
+  readonly processName: string;
+  readonly startX: number;
+
+  private inner?: TracingSession;
+  private intervalHandle?: ReturnType<typeof setInterval>;
+  private _state: ProfileState = 'recording';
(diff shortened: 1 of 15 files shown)


ui/src/plugins/dev.perfetto.Memscope/sessions/profile_session.ts
```diff
@@ -0,0 +1,188 @@
...
+        },
+      },
+      {
+        config: {
+          name: 'android.heapprofd',
+          targetBufferName: 'heapprofd',
+          heapprofdConfig: {
+            pid: [pid],
+            samplingIntervalBytes: 32 * 1024, // Slightly larger than default to reduce overhead.
+            shmemSizeBytes: 16 * 1024 * 1024, // Arbitrary, could use default.
```
primiano: remove the comment. 16mb is a good size for this scenario

primiano (approved):
Code looks great (% the pending comment in my past CL about reusing more recording classes to avoid divergencies).

Some overall suggestions that come from my playing with it (We can address in a separate PR):

1. (minor) the MEMSCOPE logo at the top feels a bit too blatantly AI-Generated and doesn't fit well with the rest of the UI imho. I'd jsust remove it
2. it's taking some quite long time to stop recordings, feels a bit more than expected (~10 seconds?). we shoudl debug and see what's going on
3. When clicking on a process, we should start the graph from scratch, otherwise it creates the false perception that we are observing the past
4. Sort the table by RSSAnon+Swap for consistenc with the default chart
5. Ask AI to give you a list of package names for SYSTEM_UI for other manufacturers for the most popular OEMs. there is probably ~10 total, and I think it can do a good guess work.
6. I'd simplify a bit the categories in process_categories.ts as follows, as there are a bit too many:

- Fold Bluetooth Wifi Nfc and Connectivity -> Connectivity
- Fold Audio -> Media
- Fold SYSTEM_SERVICE, Init -> NATIVE_DAEMON (and call it NATIVE_SERVICE)
- THen add an "APPS" category that matches xxx.yyy.zzz (i.e. at least 3 groups of [\w_-]+ seprated by dots). Note that you have optional trailer suffix after `:` like "com.google.blah:sandboxed_process_0" which should still count under the app category
- If possible I'd make NATIVE_SERVICE a regex matching: (optional prefix /system/bin/, /vendor/bin) followed by [\w_-]+

stevegolton:
Thanks for the review. Discussed offline that some of these we can punt.

> 1. (minor) the MEMSCOPE logo at the top feels a bit too blatantly AI-Generated and doesn't fit well with the rest of the UI imho. I'd jsust remove it

Removed.

> 2. it's taking some quite long time to stop recordings, feels a bit more than expected (~10 seconds?). we shoudl debug and see what's going on

I think I'll punt this one. I put the stop time down to the fact that these traces can be huge but maybe there's something else going on.

> 3. When clicking on a process, we should start the graph from scratch, otherwise it creates the false perception that we are observing the past

Agreed. I actually had this in but it was causing some unrelated issues so I removed it. Punting to another PR.

> 4. Sort the table by RSSAnon+Swap for consistenc with the default chart

Done. Changed the default sorted column to 'RSS Anon + Swap'.

> 5. Ask AI to give you a list of package names for SYSTEM_UI for other manufacturers for the most popular OEMs. there is probably ~10 total, and I think it can do a good guess work.

Done.

> 6. I'd simplify a bit the categories in process_categories.ts as follows, as there are a bit too many:
> 
> * Fold Bluetooth Wifi Nfc and Connectivity -> Connectivity
> * Fold Audio -> Media
> * Fold SYSTEM_SERVICE, Init -> NATIVE_DAEMON (and call it NATIVE_SERVICE)
> * THen add an "APPS" category that matches xxx.yyy.zzz (i.e. at least 3 groups of [\w_-]+ seprated by dots). Note that you have optional trailer suffix after `:` like "com.google.blah:sandboxed_process_0" which should still count under the app category
> * If possible I'd make NATIVE_SERVICE a regex matching: (optional prefix /system/bin/, /vendor/bin) followed by [\w_-]+

DONE

ui/src/plugins/dev.perfetto.Memscope/sessions/profile_session.ts
stevegolton: Done. Removed the comment.

PR #5787, opened 2026-05-08
ui: Prefix HeapDumpExplorer setting ID with com.android

zezeozue:


stevegolton (changes requested):
Please just put the com.android prefix, not the plugin name. Settings, commands, etc shouldn't have the plugin attached so we have some flexibility to rename plugins. The reason for the com.android "team" suffix at all is to avoid naming collisions between teams.

Also could you de-dupe the name by expecting and reusing the variable, or by plumbing the setting handle where it needs to be?

ui/src/plugins/com.android.HeapDumpExplorer/index.ts
```diff
@@ -26,8 +26,8 @@ export default class implements PerfettoPlugin {
   static readonly dependencies = [HeapProfilePlugin];
 
   async onTraceLoad(ctx: Trace): Promise<void> {
-    ctx.settings.register({
-      id: 'hideHeapDumpExplorerDefaultChangedHint',
+    const hideDefaultChangedHint = ctx.settings.register({
+      id: 'com.android#hideHeapDumpExplorerDefaultChangedHint',
```
stevegolton: Just make this `com.android.HideHeapDumpExplorerDefaultChangedHint` and you're good to go

stevegolton (approved):

PR #5789, opened 2026-05-08
ui: Fix GpuByProcess plugin and nest tracks under process groups

dreveman:
The previous GpuByProcess plugin assumed each process had a single hw queue on a single GPU and rendered all GPU slices on one combined track per process. When a process spanned multiple hw queues or multiple GPUs, slices from different queues collided at depth 0 on that combined track and got dropped or mis-rendered by the slice mipmap, leaving render stage events visible on the global hw queue tracks but missing from the per-process view.

Fix this by partitioning per-process events by (upid, hw_queue_id) and registering one SliceTrack per tuple. hw_queue_id is per-sequence so two processes can reuse the same value for distinct queues; the (upid, hw_queue_id) tuple disambiguates them. Sub-track names mirror the global hw queue track names ("Channel #1", etc.).

Nest the resulting hw queue tracks under a per-process "GPU" summary group, attached to the process group created by ProcessThreadGroups (or created lazily if no group exists yet). When a process has events on more than one GPU, insert an additional per-GPU layer ("GPU 0", "GPU 1", ...) mirroring the layout used by dev.perfetto.Gpu under the global GPU section. Tracks and groups are only created for (process, queue) combinations that actually have events.

### DIFF
--- a/ui/src/plugins/dev.perfetto.GpuByProcess/index.ts
+++ b/ui/src/plugins/dev.perfetto.GpuByProcess/index.ts
@@ -14,7 +14,6 @@
 
 import {
   LONG,
-  LONG_NULL,
   NUM,
   NUM_NULL,
   STR,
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.GpuByProcess/index.ts
```diff
@@ -26,67 +25,186 @@ import {TrackNode} from '../../public/workspace';
...
+    for (const row of rows) {
+      const {upid, hwqId, trackName, gpu} = row;
+      const uri = `dev.perfetto.GpuByProcess#${upid}#${hwqId}`;
       ctx.tracks.registerTrack({
         uri,
         renderer: SliceTrack.create({
           trace: ctx,
           uri,
           dataset: new SourceDataset({
-            src: 'gpu_slice',
```
LalitMaganti, line 68: Can we please keep the filtering at the dataset level? It is really important for aggregation performance.

dreveman: Ptal. I had to extend the filtering mechanism a bit for this

LalitMaganti:
Will defer entirely to Steve on the Dataset change.

dreveman:
I updated this to include the full stack of commits I'm interested in landed for the GpuByProcess plugin. This plugin is useful at as allows us to surface information that is process specific for APIs where that makes sense. Eg. for CUDA it's useful to know the logical device and stream that is associated with work and not only the global hw queue as the former is what the application is aware of and the latter is what actually happens on the hw. Happy to split this into different PRs but given the current topic of conversion it might be good to see the full picture.

ui/src/plugins/dev.perfetto.GpuByProcess/index.ts
stevegolton: Dataset optimization gets a bit iffy when you have `and` conditions, so I purposely avoided it.

Indeed, the way you've implemented it in the dataset optimizer is actually no better than just putting the conditions in the dataset `src`. The result is the same - union of subqueries rather than combined into an single query's `in` condition.

I'm also slightly confused as I see nowhere in this patch where the new 'and' condition is actually being used? Am I missing something?

For this patch, I'd suggest:
- Restore the dataset changes.
- Embed the conditions in the dataset `src` (I think you're doing this anyway).
- If aggregation perf is an issue, consider pre-materializing a table for all gpu events which partitions events into tracks based on a single column (e.g. a track_id for gpu events), and adjust the datasets to match.

To be clear, what I mean by aggregation perf is: area select over all gpu tracks to see the aggregated stats from that selection. The time taken to generate this graph is what may suffer when you have many gpu tracks.

stevegolton (changes requested):

ui/src/plugins/dev.perfetto.GpuByProcess/index.ts
dreveman: Sorry for the confusion. The `and` logic was only available in the first commit and I added the follow up changes to this PR so you could review the full context as I figured that would be useful to see as the `and` extension to the filter would not be enough for that. Dropped the `and` filter from the latest PR and now back to the original code.

Did some local testing of aggregation perf using an extreme trace I have with 10k kernel launches. Doesn't seem to be a problem as I see instant results when selecting all kernels in this trace. Happy to follow up with a change to make aggregation perf better if this turns out to be a problem later.

stevegolton: Thanks for clarifying and making the changes!

ui/src/core/flow_manager.ts
```diff
@@ -303,10 +303,42 @@ export class FlowManager {
       }
     }
 
-    // Fill in the track uris if available
+    // Per-slice trackUri resolution. Some slices are owned by a more
```
stevegolton: The code above is very old and very fragile, with special cases for specific known track types. We need to solve flows between arbitrary tracks holistically and efficiently, but I don't know what that looks like yet.

I fear this change completely ignores any difference in layout (slice depths) differences that arbitrary tracks might have (leading to flows pointing to the wrong place), and also feels very inefficient since calling resolveSqlEvents duplicates a lot of the work that we've already done here.

If you can proceed with this patch without this part, please do. Otherwise we'll have to work something else out.

stevegolton (changes requested):

ui/src/core/flow_manager.ts
dreveman: Dropped the flow commit from this PR. PTAL

stevegolton (approved):

PR #5795, opened 2026-05-09
protovm: fix UAF in protovm::ParseSelect

LalitMaganti:
ParseSelect snapshots `cursors_` (which holds raw Node* pointers in
RwProtoCursor::node_ and parent_link_.map_node) before running nested
instructions and unconditionally restores it afterwards. A nested `del`
that frees the same node aliased by the snapshot would leave a dangling
pointer on restore; a subsequent `del` would then double-free the node
and corrupt the SlabAllocator freelist.

Track one extra Node* in the parser: the dst.GetNode() of the innermost
in-flight ParseSelect frame. When `del` is asked to free a non-root
node whose address matches that snapshot, abort the program instead of
producing a stale cursor on restore. Snapshots form a non-strict
descending chain (each new ParseSelect either keeps dst at the same
node or moves it deeper), so checking only the innermost is sufficient:
if any outer snapshot aliases the deletion target, this one does too.
Auxiliary space is O(1).

Legitimate `del` patterns continue to work: top-level del; del under a
select that walks dst to the target via path components; iterate-and-
delete on a repeated field. Only `del` wrapped in a select that does
not move dst is rejected, and those forms are always expressible by
hoisting the del out of the wrapper:

  select(cursor=DST, path=[]) { del }       -> del
  select(cursor=SRC, path=[X]) { del }      -> select(SRC,X){...}; del
  select(path=[X]) { select(path=[]){del} } -> select(path=[X]){del}

As defense in depth, also reserve TracePacket.trace_provenance (124)
and TracePacket.protovms (125) in the SMB packet validator. Both
fields are emitted only by TracingServiceImpl and have no legitimate
producer-side use, so dropping them on the producer-incoming path
prevents a compromised producer from injecting protovm programs or
spoofing sequence provenance.

Regression tests added in vm_unittest.cc for both the empty-path and
SRC-cursor-with-non-empty-path variants of the UAF, and in
packet_stream_validator_unittest.cc for the two newly reserved fields.

### DIFF
--- a/src/protovm/error_handling.h
+++ b/src/protovm/error_handling.h
@@ -92,6 +92,18 @@ void LogStacktraceMessage(Stacktrace& stacktrace,
     }                                                                   \
   } while (0)
 
+#define PROTOVM_INTERNAL_CONCAT_IMPL(x, y) x##y
+#define PROTOVM_INTERNAL_MACRO_CONCAT(x, y) PROTOVM_INTERNAL_CONCAT_IMPL(x, y)
+
+// Evaluates |rhs| which should return a StatusOr<T> and assigns this
+// to |lhs|. If the status is an error status, returns the status from the
+// current function.
+#define PROTOVM_ASSIGN_OR_RETURN(lhs, rhs)                       \
+  PROTOVM_INTERNAL_MACRO_CONCAT(auto status_or, __LINE__) = rhs; \
+  PROTOVM_RETURN_IF_NOT_OK(                                      \
+      PROTOVM_INTERNAL_MACRO_CONCAT(status_or, __LINE__));       \
+  lhs = std::move(PROTOVM_INTERNAL_MACRO_CONCAT(status_or, __LINE__).value())
+
 enum class Status : uint8_t {
   kOk,
   kError,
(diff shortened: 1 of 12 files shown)


rukkal:
Oh nice, thanks for doing this @LalitMaganti !

I have to admit, I didn't really have the white hat on here. Did you ask Claude to hunt for vulnerabilities in the ProtoVM stuff? So now we are safe? 😅

The fix looks fine, but I have one architectural note:

The new `const Node* RwProtoCursor::GetNode()` violates the RwProtoCursor abstraction, as Node is meant to be an internal implementation detail.

I guess we can get around this by adapting Parser to store a `const RwProtoCursor* innermost_saved_dst_cursor_` (instead of the raw Node*).

Mind if I give this a spin on Monday and push/append a commit to this PR?

LalitMaganti:
> I have to admit, I didn't really have the white hat on here. Did you ask Claude to hunt for vulnerabilities in the ProtoVM stuff?

There is an internal bug for this, found by another team. Was just assigned to me as this is executed in trace processor. 

> Mind if I give this a spin on Monday and push/append a commit to this PR?

Please go ahead. I was not super convinced by this PR either but I figured was better to discuss on an actual PR rather than in the abstract.

rukkal (approved):

primiano (approved):

PR #5796, opened 2026-05-10
traced_probes: Add linux.journald data source (1/3)

### LINKED ISSUE #3288: Linux journald logs datasource

staticfloat:
Add a new linux.journald data source to traced_probes that reads log entries from the systemd journal via libsystemd. The library is loaded at runtime via dlopen("libsystemd.so.0") so there is no compile-time dependency on libsystemd.

This adds a few new protos to configure and store the journald events.  It creates a data source registered under the name 'linux.journald', opens the journal with `sd_journal_open()`, and emits one `JournaldEventPacket` per entry containing: timestamp, pid, priority, tag (SYSLOG_IDENTIFIER), message (MESSAGE), uid, comm (_COMM), systemd_unit (_SYSTEMD_UNIT), hostname (_HOSTNAME), and transport (_TRANSPORT).  It by default captures from every journal it has access to, so it will include both system and user journals if running as `root`, for instance.

This is part 1 of a split-up version of #5331 and will close #3288.

This PR was co-written by AI; while I iterated on it until the look and feel felt correct, there may be context or use cases that I am unaware of. I am open to all review comments, and will do my best to address them.


protos/perfetto/config/perfetto_config.proto
```diff
@@ -1937,6 +1937,26 @@ message InterceptorConfig {
 
 // End of protos/perfetto/config/interceptor_config.proto
```
LalitMaganti, line 1944: Please revert changes to their file, this is now auto synced

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -1937,6 +1937,26 @@ message InterceptorConfig {
```
LalitMaganti, line 1943: As above

protos/perfetto/trace/linux/journald_event.proto
```diff
@@ -0,0 +1,82 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// A batch of journald log events, collected from the Linux systemd journal.
```
LalitMaganti: Is the frequency of these events so much that we need to batch them like this? I'm honestly very unconvinced.

protos/perfetto/trace/linux/journald_event.proto
```diff
@@ -0,0 +1,82 @@
...
+// Emitted by the "linux.journald" data source in traced_probes.
+// Next field id: 3
+message JournaldEventPacket {
+  // A single journald log entry.
+  // Next field id: 15
+  message JournaldEvent {
+    // __REALTIME_TIMESTAMP: wall-clock time in microseconds since epoch.
```
LalitMaganti: I'd really prefer if we made use of the existing clock sync functionality in Perfetto rather than creating new fields for things like realtime.

protos/perfetto/trace/linux/journald_event.proto
```diff
@@ -0,0 +1,82 @@
...
+    optional string systemd_unit = 11;
+
+    // _HOSTNAME: hostname of the machine (trusted).
+    optional string hostname = 12;
+
+    // _TRANSPORT: how the entry was received by journald (trusted).
+    // Values: "audit", "driver", "syslog", "journal", "stdout", "kernel".
+    optional string transport = 13;
+
+    // __MONOTONIC_TIMESTAMP: monotonic time in microseconds since boot.
```
LalitMaganti: As above, not a fan of having out of band clock sync instead of using timestamp on TracePacket + timestamp_clock_id + clock_snapshot

Having two fields per event is also wasteful and will only cause confusion about which one is the source of truth. I'd strongly prefer to keep just monotonic and allow trace processor to figure out the realtime using clock snapshots.

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,354 @@
...
+
+    auto api = std::make_unique<SdJournalApi>();
+    auto load_sym = [&](const char* name) -> void* {
+      void* sym = dlsym(handle, name);
+      if (!sym)
+        PERFETTO_ELOG("dlsym(%s) failed: %s", name, dlerror());
+      return sym;
+    };
+
+#define LOAD(fn)                                                          \
```
LalitMaganti: This macro is honestly more confusing than helpful. I would just write it out explicitly, it's much easier to understand that way.

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,354 @@
...
+  sd_journal_seek_tail_t seek_tail;
+  sd_journal_previous_t previous;
+  sd_journal_next_t next;
+  sd_journal_get_fd_t get_fd;
+  sd_journal_process_t process;
+  sd_journal_get_realtime_usec_t get_realtime_usec;
+  sd_journal_get_monotonic_usec_t get_monotonic_usec;
+  sd_journal_get_data_t get_data;
+
+  static std::unique_ptr<SdJournalApi> Load() {
```
LalitMaganti: This doesn't need to be static if in the anonymous namespace.

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,354 @@
...
+    void* handle = dlopen("libsystemd.so.0", RTLD_NOW | RTLD_LOCAL);
+    if (!handle) {
+      PERFETTO_ELOG("Failed to dlopen libsystemd.so.0: %s", dlerror());
```
LalitMaganti: Afaik, this might well be Wai on a system without systems but this data source compiled. Imo it would be nicer if this was a more user friendly message.

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,354 @@
...
+    LOAD(get_realtime_usec)
+    LOAD(get_monotonic_usec)
+    LOAD(get_data)
+#undef LOAD
+
+    return api;
+  }
+};
+
+// Constants from <systemd/sd-journal.h> — defined locally so we don't need
```
LalitMaganti: Please place in anonymous namespace.

LalitMaganti (changes requested):
Some intuial comments.

protos/perfetto/trace/linux/journald_event.proto
staticfloat: I considered these events to be similar in volume to the android logcat, which is why I based my implementation on that.  Anecdotally, in the scenarios where I'm looking at this, we're generating ~1000 events per second in short bursts (as many related services all start logging at once in a flurry of activity, then stop).  If you think it's unnecessary to batch at this volume, I'm happy to simplify the implementation, although keeping things as similar to the android logcat case does make some sense to me.

staticfloat: Reading your comments about the clock fields, I think I'll go ahead and drop the batching, embedding only the monotonic clock field.

src/traced/probes/journald/journald_data_source.cc
staticfloat: This is the only comment that I'm not sure what to do with; I would consider it quite a serious error if I listed a datasource and it silently failed to collect any data.  I defer to you as to what the correct log level for something like this should be, for now I have improved the error message a bit.

LalitMaganti: To be clear, I agree with completely that if there was a failure it should be logged. It was more the phrasing of the message I had a problem with 

The new phrasing is significantly better thanks!

LalitMaganti (approved):

protos/perfetto/trace/linux/journald_event.proto
```diff
@@ -0,0 +1,80 @@
...
+// limitations under the License.
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// A journald log event packet, collected from the Linux systemd journal.
+// Emitted by the "linux.journald" data source in traced_probes.
+// Next field id: 3
+message JournaldEventPacket {
```
LalitMaganti: Could we remove this intermediate layer and inline both the event and stats directly into the Packet indirection?

staticfloat: Yes.  For my own education, is there a penalty we're paying for this organization, or is this more a matter of taste?

staticfloat:
I also see that [CI is unhappy with me](https://github.com/google/perfetto/actions/runs/25779343718/job/75768925641?pr=5796#step:7:23) because I added [this `builtin_clock.pbzero.h` include](https://github.com/google/perfetto/pull/5796/changes#diff-78d86f89338803bd7b1a062c8239a882702ed022f4bf47bf926653f9350f7e46R31).  Is there a better way for me to specify that this packet's timestamp comes from `CLOCK_MONOTONIC` than `packet->set_timestamp_clock_id(protos::pbzero::BUILTIN_CLOCK_MONOTONIC);`?

LalitMaganti:
> Is there a better way for me to specify that this packet's timestamp comes from CLOCK_MONOTONIC than packet->set_timestamp_clock_id(protos::pbzero::BUILTIN_CLOCK_MONOTONIC);?

You just need to update the BUILD.gn to add a reference to the protos targbet. You're missing that right now.

protos/perfetto/trace/linux/journald_event.proto
LalitMaganti: You are paying probably a couple of bytes per event. Not a huge deal. This is more about making it a bit cleaner: we generally like separating the stats from the event espeically in cases like you have here and I feel this doesn't meet the bar for that. It also makes it harder to reason about conceptually (where does the stats appear: can it appear *with* an event or is it a oneof).

If  you had a lot more different types of events under this banner, IMO it would make sense to have a layer of indirection. With two it doesn't and I also don't expect this to really grow.

staticfloat:
Hmmm, the anonymous namespace is [fighting the fact](https://github.com/google/perfetto/actions/runs/25811317345/job/75844048613?pr=5796#step:8:5504) that I have a persistent `sd_` member, despite the fact that it is private.  I could make that member an opaque pointer (e.g. `void * sd_`, then `auto* sd = static_cast<SdJournalApi*>(sd_);`), but it's a little ugly.  Do you have a preference on how to deal with this?

protos/perfetto/trace/linux/journald_event.proto
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// A journald log event packet, collected from the Linux systemd journal.
+// Emitted by the "linux.journald" data source in traced_probes.
+// Next field id: 13, statistics starting at 20
+message JournaldEventPacket {
```
LalitMaganti: Just JournalEvent. Packet doesnt add anything here.

src/traced/probes/journald/journald_data_source.h
```diff
@@ -0,0 +1,91 @@
...
+#include "perfetto/tracing/core/forward_decls.h"
+#include "src/traced/probes/probes_data_source.h"
+
+// Forward-declare the C struct to avoid pulling in <systemd/sd-journal.h>
+// into every TU that includes this header.
+struct sd_journal;
+
+namespace perfetto {
+
+// Opaque wrapper around dynamically loaded libsystemd symbols. The full
```
LalitMaganti: Why not make this just a private forward declaration in the class? You can still just declare it here and then define it in the cc file?

protos/perfetto/trace/linux/journald_event.proto
LalitMaganti: I'd actually maybe a bit more specific and call it SystemdJournalEvent to make it crysal clear.

protos/perfetto/config/linux/journald_config.proto
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Configuration for the "linux.journald" data source.
+// Next field id: 4
+message JournaldConfig {
```
LalitMaganti: Can wer keep this consistent with the config (SystemdJournaldConfig).

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,391 @@
...
+  api->get_data =
+      reinterpret_cast<sd_journal_get_data_t>(load_sym("sd_journal_get_data"));
+  if (!api->get_data)
+    return nullptr;
+
+  return api;
+};
+
+const ProbesDataSource::Descriptor JournaldDataSource::descriptor = {
+    /*name*/ "linux.journald",
```
LalitMaganti: systemd_journald for consistency.

protos/perfetto/config/data_source_config.proto
```diff
@@ -305,6 +306,9 @@ message DataSourceConfig {
   // Data source name: android.aflags
   optional AndroidAflagsConfig android_aflags_config = 140 [lazy = true];
 
+  // Data source name: linux.journald
+  optional JournaldConfig journald_config = 141 [lazy = true];
```
LalitMaganti: systemd_journald_config for consistency.

LalitMaganti (approved):

src/traced/probes/journald/journald_data_source_unittest.cc
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
```
sashwinbalaji: nit  for all journald/ files: 2026

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  // Emit a stats packet.
+  {
+    auto packet = writer_->NewTracePacket();
+    packet->set_timestamp(static_cast<uint64_t>(base::GetBootTimeNs().count()));
+    auto* stats = packet->set_journald_event();
+    stats->set_num_total(stats_.num_total);
+    stats->set_num_skipped(stats_.num_skipped);
+    stats->set_num_failed(stats_.num_failed);
```
sashwinbalaji, line 381: don't see any increments for this field

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,392 @@
...
+                                     : static_cast<uint32_t>(std::strtoul(
+                                           prio_str.c_str(), nullptr, 10));
+    if (prio > min_prio_) {
+      stats_.num_skipped++;
+      continue;
+    }
+
+    uint64_t monotonic_us = 0;
+    sd_id128_t boot_id{};
+    sd_->get_monotonic_usec(journal_, &monotonic_us, &boot_id);
```
sashwinbalaji: should we verify the return value is 0, here and other places ?

https://man7.org/linux/man-pages/man3/sd_journal_get_monotonic_usec.3.html

src/traced/probes/journald/journald_data_source.cc
staticfloat: Thank you, I neglected to do a pass of error handling, this and your other comment are now addressed.

LalitMaganti (approved):

LalitMaganti:
Still LGTM

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,406 @@
...
+    if (!gid_str.empty())
+      ev->set_gid(
+          static_cast<uint32_t>(std::strtoul(gid_str.c_str(), nullptr, 10)));
+
+    stats_.num_total++;
+    n++;
+  }
+}
+
+std::string JournaldDataSource::GetField(const char* field) {
```
sashwinbalaji: maybe need to benchmark would a single sd_journal_enumerate_data be better than these many separate sd_journal_get_data calls ?

https://man7.org/linux/man-pages/man3/sd_journal_enumerate_data.3.html

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,406 @@
...
+void JournaldDataSource::Flush(FlushRequestID, std::function<void()> callback) {
+  if (journal_ && sd_) {
+    sd_->process(journal_);
+    ReadJournalEntries();
+  }
+
+  // Emit a stats packet.
+  {
+    auto packet = writer_->NewTracePacket();
+    packet->set_timestamp(static_cast<uint64_t>(base::GetBootTimeNs().count()));
```
sashwinbalaji, line 377: confused why boottime used here but monotonic used in ReadJournalEntries

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,406 @@
...
+    uint32_t prio = prio_str.empty() ? min_prio_
+                                     : static_cast<uint32_t>(std::strtoul(
+                                           prio_str.c_str(), nullptr, 10));
+    if (prio > min_prio_) {
+      stats_.num_skipped++;
+      continue;
+    }
+
+    uint64_t monotonic_us = 0;
+    sd_id128_t boot_id{};
```
sashwinbalaji: should we check this variable ? if not why pass it ?

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,406 @@
...
+
+void JournaldDataSource::ReadJournalEntries() {
+  uint32_t n = 0;
+
+  while (sd_->next(journal_) > 0 && n < kMaxEventsPerRead) {
+    std::string prio_str = GetField("PRIORITY");
+    uint32_t prio = prio_str.empty() ? min_prio_
+                                     : static_cast<uint32_t>(std::strtoul(
+                                           prio_str.c_str(), nullptr, 10));
+    if (prio > min_prio_) {
```
sashwinbalaji: i'm not sure if this if block will be ever true.

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,406 @@
...
+  // establish the per-file n_entries baseline in the sd_journal internals.
+  // Without it the EOF short-circuit in sd_journal_next() never clears on
+  // subsequent wakeups, so new entries written after Start() would never
+  // be seen.
+  while (sd_->next(journal_) > 0) {
+  }
+}
+
+void JournaldDataSource::OnJournalReadable() {
+  sd_->process(journal_);
```
sashwinbalaji: nit: return value not considered

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,406 @@
...
+    min_prio_ = cfg.min_prio();
+  for (auto id = cfg.filter_identifiers(); id; ++id)
+    filter_identifiers_.push_back(id->as_std_string());
+  for (auto u = cfg.filter_units(); u; ++u)
+    filter_units_.push_back(u->as_std_string());
+}
+
+JournaldDataSource::~JournaldDataSource() {
+  if (journal_ && sd_) {
+    task_runner_->RemoveFileDescriptorWatch(sd_->get_fd(journal_));
```
sashwinbalaji: it's possibe AddFileDescriptotWatch was never invoked due to early return by SD_CHECKs which can on a debug build might result in crash as we have PERFETTO_DCHECK(watch_tasks_.count(fd));

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  // Register the fd watch before draining so no wakeups are missed.
+  auto weak = weak_factory_.GetWeakPtr();
+  task_runner_->AddFileDescriptorWatch(fd, [weak] {
+    if (weak)
+      weak->OnJournalReadable();
+  });
+
+  // Drain once after seek_tail(). sd_journal_next() returns 0 immediately
```
sashwinbalaji: I think there's a small race here worth a look.

If journald writes a new entry between previous() above and this loop, next() returns 1, the cursor moves onto that entry, and the empty loop body drops it. 

OnJournalReadable can only run after this function is completed, so when it runs next() returns 0 because we're already past it, so that entry never makes it into the trace.

Probably rare but I dont have much info on journald so would ask you to consider once.

sashwinbalaji (commented):
I'm not that familiar with journald, so please use your best judgement on these.

Also feel free to resolve anything that doesn't make sense as just flagging things that looked a bit off to me.

src/traced/probes/journald/journald_data_source.cc
staticfloat: In this case, it's because the two choices of timestamps for journal [events are monotonic or realtime](https://manpages.debian.org/stretch-backports/libsystemd-dev/sd_journal_get_monotonic_usec.3.en.html), I chose monotonic for simplicity.  However, for this `stats` packet, we don't have a journal event to use the time from, so I'm just taking the current time, which seems easiest to do with `base::GetBootTimeNs()`.  Is it a problem to switch packets clock id's per-packet like this?

Alternatively, is it worth it to try and construct a CLOCK_MONOTONIC timestamp here?

src/traced/probes/journald/journald_data_source.cc
staticfloat: The default initialization of `NULL` means to use the current boot (rather than a previous boot).  However, looking at this from a different angle, there's no reason to use `CLOCK_MONOTONIC` here instead of `CLOCK_REALTIME`, which doesn't require this parameter at all (because the times are non-ambiguous across multiple boots).  We don't really care about this distinction because we're only interested in live events, not historical ones, but it makes the code here slightly shorter, so I'll switch to `CLOCK_REALTIME`.

src/traced/probes/journald/journald_data_source.cc
LalitMaganti: On desktop it won't matter too much because there is no suspend so monotonic == boottime.

I would add timestamp_clock_id here explicitly as well FWIW.

src/traced/probes/journald/journald_data_source.cc
LalitMaganti: I would prefer to keep it CLOCK_MONOTONIC honestly. REALTIME has monotonicity problems so it's not always possible to convert to trace time. MONOTONIC doesn't have that problem.

src/traced/probes/journald/journald_data_source.cc
staticfloat: Agreed, the disjunction above _should_ handle this completely, so I'm going to remove it (and the `skipped` stat).

src/traced/probes/journald/journald_data_source.cc
staticfloat: I looked into this more, and I don't actually think this draining is needed.  I ran both sparse logging tests and a torture test generating many events, and I think my initial diagnosis (which led me to this draining solution) was incorrect.  I've dropped it and compared against similar projects (which don't have this drain) and it seems to still be working fine.

src/traced/probes/journald/journald_data_source.cc
staticfloat: Thank you, I did indeed benchmark it and it looks like `enumerate_data()` is better in every case I tested.

src/traced/probes/journald/journald_data_source.cc
staticfloat: I tried to trigger this again after my refactor, and I did get into a situation where no events fired, I believe because journald received an event between the time I requested the `fd` and when I added a watch upon it.  I fixed this by explicitly calling `OnJournalReadable()` at the end of `Start()` to properly process those events.

staticfloat:
It looks like a CI run was attempted, but it's totally unclear to me why the CI failed.

LalitMaganti:
I think there was a scheduling failure, kicked off a rerun.

staticfloat:
Another scheduling error?

LalitMaganti:
All passing now :) 

GitHub seems to be having a problematic time, had to send a bunch of retries.

src/trace_processor/storage/stats.h
```diff
@@ -175,6 +175,9 @@ namespace perfetto::trace_processor::stats {
   F(stackprofile_invalid_frame_id,        kSingle,  kError,    kTrace, Scope::kMachineAndTrace,    ""), \
   F(stackprofile_invalid_callstack_id,    kSingle,  kError,    kTrace, Scope::kMachineAndTrace,    ""), \
   F(stackprofile_parser_error,            kSingle,  kError,    kTrace, Scope::kMachineAndTrace,    ""), \
+  F(systemd_journal_num_failed,           kSingle,  kError,    kTrace, Scope::kMachineAndTrace,    ""), \
+  F(systemd_journal_num_skipped,          kSingle,  kInfo,     kTrace, Scope::kMachineAndTrace,    ""), \
```
sashwinbalaji: redundant we removed this from proto.

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,382 @@
...
+      EXTRACT_UINT_FIELD(PRIORITY, set_prio)
+      EXTRACT_UINT_FIELD(_PID, set_pid)
+      EXTRACT_UINT_FIELD(_TID, set_tid)
+      EXTRACT_UINT_FIELD(_UID, set_uid)
+      EXTRACT_UINT_FIELD(_GID, set_gid)
+    }
+
+    stats_.num_total++;
+  }
+}  // namespace perfetto
```
sashwinbalaji: misplaced comment ?

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,382 @@
...
+void JournaldDataSource::OnJournalReadable() {
+  SD_CHECK(sd_->process(journal_));
+  ReadJournalEntries();
+}
+
+static inline std::string extract_field(const char* field,
+                                        const char* key,
+                                        size_t len) {
+  size_t key_len = strlen(key);
+  if (len > strlen(key) + 2 && memcmp(field, key, key_len) == 0 &&
```
sashwinbalaji: maybe I'm misunderstanding how does this work for priority=n, isn't len = key len + 2 ?

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,382 @@
...
+      field[key_len] == '=') {
+    return std::string(field + strlen(key) + 1, len - (strlen(key) + 1));
```
sashwinbalaji: we can reuse key_len from above

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,382 @@
...
+  // `poll()`.
+  OnJournalReadable();
+}
+
+void JournaldDataSource::OnJournalReadable() {
+  SD_CHECK(sd_->process(journal_));
+  ReadJournalEntries();
+}
+
+static inline std::string extract_field(const char* field,
```
sashwinbalaji: can we move this to unknown namespace and also align namign style pls.

src/traced/probes/journald/journald_data_source.cc
```diff
@@ -0,0 +1,382 @@
...
+    auto* ev = packet->set_journald_event();
+
+    const void* data = nullptr;
+    size_t len = 0;
+    sd_->restart_data(journal_);
+    while (sd_->enumerate_available_data((journal_), &data, &len) > 0) {
+      const char* field = static_cast<const char*>(data);
+
+// Extract a `std::string` field, pass it to `setter()` if it exists
+#define EXTRACT_STRING_FIELD(key, setter)                    \
```
sashwinbalaji: can we instead use a lambda ?

staticfloat: Maybe I'm misunderstanding, but the way I would make this a lambda would be to have:

```
auto extract_string = [&](const char* key, auto setter) {
  std::string value = ExtractField(field, key, len);
  if (!value.empty())
    setter(value);
};
```

And then I would need to make each invocation also pass in another lambda, in order to pass along the object's `this` pointer;
```
extract_string("MESSAGE", [&](const auto& val) { ev->set_message(val); });
```

Which seems much more verbose than the current macro approach.  Do you have a better suggestion?

staticfloat: I decided to just go with this and if you don't like it, I'll change it back.

sashwinbalaji (approved):
Thanks so much for the iterations!!

src/traced/probes/journald/journald_data_source.cc
staticfloat: You're right, thank you for catching that.  I think this was previously masked by pulling `PRIORITY` out as a separate get data request.

LalitMaganti (approved):

LalitMaganti:
Thanks a lot for the patience here while we got this reviewed! Please let me know when the next change in the chain is rebased and ready for me to look at :)

PR #5798, opened 2026-05-10
ui: Add dev.perfetto.JournaldLog plugin and update AndroidLog (3/3)

### LINKED ISSUE #3288: Linux journald logs datasource

staticfloat:
Add a new JournaldLog plugin (dev.perfetto.JournaldLog) that displays
journald log entries from the linux.journald data source. The plugin
shares UI components with the existing AndroidLog plugin, extracted into:

- ui/src/components/tracks/log_track.ts: generic log track factory
- ui/src/components/widgets/log_panel/log_panel.ts: shared LogPanel
  component with filtering, search, and per-source extensibility

The AndroidLog plugin is refactored to use these shared components.
Register dev.perfetto.JournaldLog in default_plugins.ts.

This is part 3 of a split-up version of #5331 and will close #3288.

This PR was co-written by AI; while I iterated on it until the look and feel felt correct, there may be context or use cases that I am unaware of. I am open to all review comments, and will do my best to address them.

### DIFF
--- a/ui/src/plugins/dev.perfetto.JournaldLog/index.ts
+++ b/ui/src/plugins/dev.perfetto.JournaldLog/index.ts
@@ -1,0 +1,158 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import m from 'mithril';
+import {
+  type JournaldLogFilteringCriteria,
+  JournaldLogPanel,
+} from './logs_panel';
+import type {Trace} from '../../public/trace';
+import type {PerfettoPlugin} from '../../public/plugin';
+import {NUM, STR_NULL} from '../../trace_processor/query_result';
+import {createJournaldLogTrack} from './logs_track';
+import {TrackNode} from '../../public/workspace';
+import {escapeSearchQuery} from '../../trace_processor/query_utils';
+import {exists} from '../../base/utils';
+
+const VERSION = 1;
+const JOURNALD_LOGS_TRACK_KIND = 'JournaldLogTrack';
+
+const DEFAULT_STATE: JournaldLogPluginState = {
+  version: VERSION,
+  filter: {
+    // Show all levels by default (7 = DEBUG, the least severe).
+    minimumLevel: 7,
+    tags: [],
+    isTagRegex: false,
+    textEntry: '',
+    hideNonMatching: true,
+  },
+};
+
+interface JournaldLogPluginState {
+  version: number;
+  filter: JournaldLogFilteringCriteria;
+}
+
+function isJournaldLogFilteringCriteria(
+  value: unknown,
+): value is JournaldLogFilteringCriteria {
+  if (!exists(value) || typeof value !== 'object') {
+    return false;
+  }
+  const candidate = value as Record<string, unknown>;
+  return (
+    typeof candidate.minimumLevel === 'number' &&
+    Array.isArray(candidate.tags) &&
+    candidate.tags.every((tag) => typeof tag === 'string') &&
+    typeof candidate.textEntry === 'string' &&
+    typeof candidate.hideNonMatching === 'boolean' &&
+    (candidate.isTagRegex === undefined ||
+      typeof candidate.isTagRegex === 'boolean')
+  );
+}
+
+function migrateJournaldPluginState(init: unknown): JournaldLogPluginState {
+  if (!exists(init) || typeof init !== 'object') {
+    return DEFAULT_STATE;
+  }
+  const candidate = init as Record<string, unknown>;
+  if (
+    candidate.version === VERSION &&
+    isJournaldLogFilteringCriteria(candidate.filter)
+  ) {
+    return {
+      version: VERSION,
+      filter: candidate.filter,
+    };
+  }
+  return DEFAULT_STATE;
+}
+
+function migrateJournaldFilter(value: unknown): JournaldLogFilteringCriteria {
+  return isJournaldLogFilteringCriteria(value) ? value : DEFAULT_STATE.filter;
+}
+
+export default class implements PerfettoPlugin {
+  static readonly id = 'dev.perfetto.JournaldLog';
+  async onTraceLoad(ctx: Trace): Promise<void> {
+    const store = ctx.mountStore<JournaldLogPluginState>(
+      'dev.perfetto.JournaldLogFilterState',
+      migrateJournaldPluginState,
+    );
+
+    const result = await ctx.engine.query(`
+      INCLUDE PERFETTO MODULE linux.systemd_journald;
+      select count(1) as cnt from linux_systemd_journald_logs;
+    `);
+    const logCount = result.firstRow({cnt: NUM}).cnt;
+    const uri = 'perfetto.JournaldLog';
+    if (logCount > 0) {
+      ctx.tracks.registerTrack({
+        uri,
+        tags: {kinds: [JOURNALD_LOGS_TRACK_KIND]},
+        renderer: createJournaldLogTrack(ctx, uri),
+      });
+      const track = new TrackNode({
+        name: 'Journald logs',
+        uri,
+      });
+      ctx.defaultWorkspace.addChildInOrder(track);
+    }
+
+    const journaldLogsTabUri = 'perfetto.JournaldLog#tab';
+
+    const filterStore = store.createSubStore(['filter'], migrateJournaldFilter);
+
+    ctx.tabs.registerTab({
+      isEphemeral: false,
+      uri: journaldLogsTabUri,
+      content: {
+        render: () => m(JournaldLogPanel, {filterStore, trace: ctx}),
+        getTitle: () => 'Journald Logs',
+      },
+    });
+
+    if (logCount > 0) {
+      ctx.tabs.addDefaultTab(journaldLogsTabUri);
+    }
+
+    ctx.commands.registerCommand({
+      id: 'dev.perfetto.ShowJournaldLogsTab',
+      name: 'Show journald logs tab',
+      callback: () => {
+        ctx.tabs.showTab(journaldLogsTabUri);
+      },
+    });
+
+    ctx.search.registerSearchProvider({
+      name: 'Journald logs',
+      selectTracks(tracks) {
+        return tracks
+          .filter((track) =>
+            track.tags?.kinds?.includes(JOURNALD_LOGS_TRACK_KIND),
+          )
+          .filter((t) =>
+            t.renderer.getDataset?.()?.implements({msg: STR_NULL}),
+          );
+      },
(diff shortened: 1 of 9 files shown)


staticfloat:
Note that because `gh stack` says that this repository doesn't have stacks enabled, I wasn't able to use that new feature, so this branch contains the two commits before it.

staticfloat:
Now that the previous two branches in this series have been merged, this is now ready for review.  Please note that this branch represents the code area I'm least confident in; I'm really not a frontend developer, and I've let the LLM drive more than I did in the C++ code. From a high level perspective, the intent here is to factor out the shared logging infrastructure code into a shared module (`widgets/log_panel`), derive both the Android logcat and Linux systemd log panels from that, and add a web recording page section for Linux-specific recording types, similar to that of the Android recording types.

Note that I added a dropdown element as well, since having a slider for log level seemed less user-friendly.

staticfloat:
@LalitMaganti pinging you in case this has fallen through the cracks.

LalitMaganti:
This is mainly on @stevegolton. I'm out most of this week but I can take a look at the end.

stevegolton:
This is on me, I'll take a look today, hopefully this morning.

ui/src/public/track_kinds.ts
```diff
@@ -20,3 +20,4 @@ export const THREAD_STATE_TRACK_KIND = 'ThreadStateTrack';
 export const SLICE_TRACK_KIND = 'SliceTrack';
 export const COUNTER_TRACK_KIND = 'CounterTrack';
 export const ANDROID_LOGS_TRACK_KIND = 'AndroidLogTrack';
+export const JOURNALD_LOGS_TRACK_KIND = 'JournaldLogTrack';
```
stevegolton: Since this const is only used in the journald plugin, it should be moved there. This file is really to support legacy features like legacy search, and is messy.

ui/src/components/widgets/log_panel/log_panel.ts
```diff
@@ -0,0 +1,663 @@
...
+import type {Store} from '../../../base/store';
+import type {Trace} from '../../../public/trace';
+import {Icons} from '../../../base/semantic_icons';
+import {SerialTaskQueue, QuerySlot} from '../../../base/query_slot';
+import {Anchor} from '../../../widgets/anchor';
+import {getThreadUriPrefix} from '../../../public/utils';
+
+export const ROW_H = 24;
+
+// ---------------------------------------------------------------------------
```
stevegolton: I'd remove these sections, they are classic LLM generated bloat and don't really add much.

If you wanted to document each interface/class/function - use jsdoc instead.

ui/src/plugins/dev.perfetto.JournaldLog/logs_panel.scss
stevegolton, line 1: Is this file being imported anywhere? I think it should be imported in `./logs_panel.ts`

ui/src/plugins/dev.perfetto.JournaldLog/index.ts
```diff
@@ -0,0 +1,129 @@
+// Copyright (C) 2024 The Android Open Source Project
```
stevegolton: Nit: 2026 (and below)

ui/src/components/widgets/log_panel/log_panel.ts
```diff
@@ -0,0 +1,663 @@
...
+    priorities.push(row.prio as number);
+    tags.push(row.tag as string);
+    messages.push(row.msg as string);
+    isHighlighted.push(
+      row.isMsgHighlighted === 1 || row.isProcessHighlighted === 1,
+    );
+    processName.push(row.processName as string);
+    utids.push((row.utid as number | null) ?? null);
+    upids.push((row.upid as number | null) ?? null);
+    config.collectExtraRow?.(row, extra);
```
stevegolton: I'm slightly concerned about all these type assertions - the point of the schema is that you shouldn't have to do this.

stevegolton (changes requested):
While I appreciate the commitment to DRY here by pulling out the common functionality between the logs panel and tracks, I think it actually muddies the water somewhat as I'm finding it difficult to reason about how the new journald functionality works, and I believe we've lost a bit of functionality from android logs panel.

I suspect it might be clearer to simply duplicate the code across both plugins for now, and we can run a cleanup job later if there's enough commonality between the two. There's not a whole lot of complexity in each implementation (especially the track, which leans heavily on the SliceTrack helper anyway).

Other than that, it would be great if you could supply a test trace (if you haven't already in previous PRs - I might have missed it!)

staticfloat:
> I suspect it might be clearer to simply duplicate the code across both plugins for now, and we can run a cleanup job later if there's enough commonality between the two.

Okay, I will revert back to something more similar to [my original PR](https://github.com/google/perfetto/pull/5331) where I had duplicated the code rather than factored it out into a shared logs panel.

> it would be great if you could supply a test trace

[Here's a test trace](https://github.com/user-attachments/files/28515095/linux-202652-73640.pftrace.gz) with only the `linux.systemd_journald` datasource enabled.  I have a bash loop generating a bunch of different priority journald messages via:

```bash
while [ true ]; do
    sleep 1
    for prio in $(seq 0 7); do
        echo "Priority ${prio}" | logger -p ${prio} --id=$$
    done
done
```

staticfloat:
Some design questions you may have an opinion on:

* Currently, due in part to the shared history with the Android track, the colors/heights used in the log track for different log priority levels are not unique.  I think this may partly be a feature to avoid overloading the user with fully seven different colors for the log priority levels, but perhaps not.  Do you have an opinion?

* I added a `Linux` section in the web UI recording interface.  This seems like the right place to put this, although I could also see maybe we add a "System logs" section and group the android and journald logs together in that location; with potential other logging sources (e.g. a future custom datasource that grabs lines out of a bare log file) in the future, that could be useful for integrating logs from simpler services that don't have explicit perfetto integration).

staticfloat:
Ah, I realized I accidentally dropped the "click on PID/TID" to scroll to that process in the rest of the track; so [here's a new example trace](https://github.com/user-attachments/files/28519421/linux-202652-10218.pftrace.gz) with the scheduling datasource enabled so that you can see that working as well:

stevegolton:
> Okay, I will revert back to something more similar to [my original PR](https://github.com/google/perfetto/pull/5331) where I had duplicated the code rather than factored it out into a shared logs panel.

Apologies for the churn here. I didn't realize Lalit already asked you to merge the two. I still think separating the two is the right call. There's some commonality for sure but where we draw the line isn't entirely clear to me right now.

> Currently, due in part to the shared history with the Android track, the colors/heights used in the log track for different log priority levels are not unique. I think this may partly be a feature to avoid overloading the user with fully seven different colors for the log priority levels, but perhaps not. Do you have an opinion?

You're referring to this existing android log track? Looking at it now, it seems levels 0 and 1 are not used in practice on android, so each of the log levels 2 through 7 inclusive get their own color. It does seem DEBUG (3) and VERBOSE (2) share a lane... though I suspect that might be a bug.

I see you've merged log levels 0-2 and 5-6 into the same lane and color in the journald plugin. I don't really have an opinion here, if you wanted to make the track taller so you can squeeze more lanes than I wouldn't be against that. Either way I'd imagine each event should have a unique color even if they do share lanes?

> I added a Linux section in the web UI recording interface. This seems like the right place to put this, although I could also see maybe we add a "System logs" section and group the android and journald logs together in that location; with potential other logging sources (e.g. a future custom datasource that grabs lines out of a bare log file) in the future, that could be useful for integrating logs from simpler services that don't have explicit perfetto integration).

I can see both sides, though I think calling it Linux makes sense for now seeing as logcat is under the Android section. If we created a System logs section and don't move logcat into it, it might cause some confusion. Though I'd like to hear @primiano 's thoughts on this.

> [Here's a test trace](https://github.com/user-attachments/files/28515095/linux-202652-73640.pftrace.gz) with only the linux.systemd_journald datasource enabled. I have a bash loop generating a bunch of different priority journald messages via:

Thanks for the test trace, though FYI I got this error opening opening it on the tip of your branch. I think we need to use LONG_NULL for the pid and tid fields.

```
Error @ row: 238 col: 'pid': SQL value is NULL but that was not expected (expected type: LONG). Did you mean NUM_NULL, LONG_NULL, STR_NULL or BLOB_NULL?
EngineTag: dev.perfetto.JournaldLog
Query:

        select
          id,
          ts,
          pid,
          tid,
          prio,
          ifnull(tag, '[NULL]') as tag,
          ifnull(msg, '[NULL]') as msg,
          is_msg_highlighted as isMsgHighlighted,
          is_process_highlighted as isProcessHighlighted,
          ifnull(process_name, '') as processName,
          ifnull(systemd_unit, '') as systemdUnit
        from filtered_linux_systemd_journald_logs
        where ts >= 21573343482549799 and ts <= 21573349212743224
        order by ts
        limit 0, 54
 - Error: Error  ( row: 238 col: 'pid': SQL value is NULL but that was not expected (expected type: LONG). Did you mean NUM_NULL, LONG_NULL, STR_NULL or BLOB_NULL?)
 - RowIteratorImpl.makeError (/trace_processor/query_result.ts:484:12)
 - RowIteratorImpl.tryMoveToNextBatch (/trace_processor/query_result.ts:603:20)
 - RowIteratorImpl.next (/trace_processor/query_result.ts:506:17)
 - new RowIteratorImpl (/trace_processor/query_result.ts:478:10)
 - new RowIteratorImplWithRowData (/trace_processor/query_result.ts:617:18)
 - QueryResultImpl.iter (/trace_processor/query_result.ts:230:18)
 - updateLogEntries (/plugins/dev.perfetto.JournaldLog/logs_panel.ts:319:25)
 -  (async /base/query_slot.ts:81:26)
 - async SerialTaskQueue.runNext (/base/query_slot.ts:34:7)
```

staticfloat:
My apologies, I pushed an updated version of the code a little after I wrote my message, which fixes that error.

stevegolton:
The CI failure is down to a brittle test which I've fixed in https://github.com/google/perfetto/pull/6109

You should be good to go if you rebase on main.

ui/src/plugins/dev.perfetto.JournaldLog/logs_panel.ts
```diff
@@ -0,0 +1,602 @@
...
+
+export interface JournaldLogFilteringCriteria {
+  readonly minimumLevel: number;
+  readonly tags: string[];
+  readonly isTagRegex?: boolean;
+  readonly textEntry: string;
+  readonly hideNonMatching: boolean;
+}
+
+export interface JournaldLogPanelCache {}
```
stevegolton: It doesn't look like this is used?

stevegolton (approved):

stevegolton:
Looks good to me. Deferring to @primiano for the second stamp and opinions on the recording page layout.

staticfloat:
Thank you Steve; I've gone ahead and broken up the track into 8 rows so that each priority level has its own sub-track, and also given them disctinct colors (except for critical, alert and emergency, which are all magenta, and are quite rare in my experience).  I'm unsure if the current rainbow palette is actually good, I'll have to play around with it with real traces to see how I feel about it; I tried to adapt the same palette used by the Android logs to start.

stevegolton (approved):
LGTM!

LalitMaganti (approved):

staticfloat:
Thank you all for your guidance, glad to have this in!

LalitMaganti:
Thanks for the contribution and the patience while we got this in! Very much appreciated!

Issue #5799, opened 2026-05-10
Reduce amalgamated C SDK source size


dreveman:
`tools/gen_amalgamated --sdk c` currently produces ~12 MB / 290k lines of source. Most of it is dead weight: per-domain trace + config protos and a handful of C++ files that the C ABI under `include/perfetto/public/abi/` never references. They're pulled in transitively via the GN dep graph (`recurse_in_header_deps` walks `:zero|cpp` even when nothing `#include`s them) and via `cppgen_plugin`'s `[lazy = true]` skip not being transitive.

I have a working set of changes that:
- adds `enable_perfetto_minimal_trace_protos` (default off, auto-on for `--sdk c`)
- gates per-domain trace/config/common proto deps behind it
- extends `cppgen_plugin`'s lazy-import skip to walk transitively
- gates `src/tracing/{console_interceptor,track_event_state_tracker,track_event_legacy}.cc` and `src/base/{cpu_info,ctrl_c_handler,getopt_compat,subprocess*}.cc` (all unreachable from the C ABI)

Result on the C SDK build (no public API change, no impact on `--sdk cpp` / `--sdk all` / chromium / Android / standalone):

|  | Before | After |
|---|---:|---:|
| `perfetto_c.{cc,h}` | 12.0 MB | **5.6 MB (-54%)** |
| Stripped `.so` (`-O2 -fvisibility-inlines-hidden`) | 2.69 MB | **2.22 MB (-17%)** |
| `clang++ -O2 -c` wall time | 37.1 s | 30.1 s (-19%) |

@LalitMaganti @primiano Are we interested in making this improvement? If yes I'll prepare them for review (6 small commits).

LalitMaganti:
As long as it's not big changes I'm in favor.

PR #5800, opened 2026-05-10
sdk: Reduce amalgamated C SDK source size by ~54%

dreveman:
`tools/gen_amalgamated --sdk c` produces 12 MB / 290k lines today. Most is dead weight: per-domain trace + config protos and a handful of C++ files that the C ABI under `include/perfetto/public/abi/` never references — pulled in transitively via the GN dep graph and via `cppgen_plugin`'s `[lazy = true]` skip not being transitive.

This PR adds a new GN arg `enable_perfetto_minimal_trace_protos` (default off, auto-enabled by `gen_amalgamated --sdk c` only) and uses it to gate the unused subgraphs. `--sdk cpp` / `--sdk all` / chromium / Android / standalone builds are unaffected. Public C ABI is unchanged.

|  | Before | After |
|---|---:|---:|
| `perfetto_c.{cc,h}` | 12.0 MB | 5.6 MB (-54%) |
| Stripped `.so` (`-O2 -fvisibility-inlines-hidden`) | 2.69 MB | 2.22 MB (-17%) |
| `clang++ -O2 -c` wall time | 37.1 s | 30.1 s (-19%) |

The stack is six small commits, each independently revertable:

1. Drop unused per-domain trace pbzero headers via the new flag.
2. Make `cppgen_plugin`'s `[lazy = true]` import-skip transitive (strict bugfix of an existing partial behaviour) and use it to drop per-domain config + common protos.
3. Drop a few more chrome / etw / stress_test config protos that the cppgen fix newly unblocks.
4. Drop unused top-level trace protos (`test_event`, `memory_graph`, `ui_state`, `evdev`, `test_extensions`).
5. Drop `sys_stats_counters` from common (importers all dropped above).
6. Drop C++ implementation files unused by the C ABI: `console_interceptor.cc`, `track_event_state_tracker.cc`, `track_event_legacy.cc`, and `src/base/{cpu_info,ctrl_c_handler,getopt_compat,subprocess*}.cc`.

See per-commit messages for the audit method (`grep` for `#include` sites under `src/{shared_lib,tracing}/` and `include/perfetto/{public,tracing,ext/tracing}/`, excluding tests) and the source/binary delta of each step.

Issue: #5799


LalitMaganti:
I think Primiano is best placed on whether this would make sense on the future direction of the SDK.

primiano:
Hmm so I am sympathetic with the overall goal, but not the with the way this is being done.
I don't like the fact that this is achieved by creating a variable in the build system, which is set only by the SDK generator, because this effectively creates a new build "flavour" which is impossible to make sure it builds and is consistent until we build the SDK.
It also prevents from the future to depend on the SDK (without amalgamantion) at the build system level, because this effectively requires a different invocation of the buildystem, setting that `enable_perfetto_minimal_trace_protos`

Instead I think the proper way of doing this would be more splitting targets inside the build system, i.e. define a target with the minimal protos (and making sure the C sdk only depends on those) and keep the beefier target with all the protos for the existing cases. 
Essentially the way trace_processor_minimal vs trace_processor_shell work today. You can get both from the same checkout without having to specify a different GN config.

dreveman:
@primiano I let an agent run with this and after some back and forth we ended up with the current set of changes. take a look and let me know if this is something you like me to move forward with or abandon? either way is fine with me.

PR #5802, opened 2026-05-11
ui: Add SVG based histogram chart component

stevegolton:
This patch introduces a new SVG based histogram implementation (not using echarts), similar to the newly introduced SVG based line chart component, which (compared to the echart impl):

- Is stable in screenshot tests.
- Looks better (no sub-pixel gaps between bars).
- Gives us a lot more control to do simple yet custom things like highlighting the background of a bar, which echarts makes it difficult to do.
- Fits with the UI's theme and and existing widgets much more closely.

This patch contains the following changes:
- New chart implementation `HistogramSVG`.
- Refactor out some common reusable chart logic into common modules shared with `LineChartSvg`.
- Refactor `LineChartSVG` a little to use the new common code and remove a previous hack based on keying.
- Use the new svg based histogram in the slice distribution panels.


ui/src/components/widgets/charts_svg/svg_chart_frame.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (!this.container) return;
+    let inner: m.Children;
+    if (attrs.isLoading) {
+      inner = m('.pf-chart-svg__loading', m(Spinner));
+    } else if (attrs.isEmpty) {
+      inner = m('.pf-chart-svg__empty', 'No data to display');
+    } else {
+      inner = attrs.renderChart(width, height);
+    }
+    // m.render's hidden third parameter is the redraw hook used by
```
LalitMaganti, line 97: Not a huge fan of relying on internal impl details like this. This feels like it's a hack?

LalitMaganti (approved):

ui/src/components/widgets/charts_svg/svg_chart_frame.ts
stevegolton: It's not actually an internal implementation - it's documented here: https://mithril.js.org/render.html#signature

It's just not in the typescript signature (I'm not sure why). FWIW we use this technique in raf.ts as well.

Going forward we could either:
- Override the type globally for the project.
- Submit a patch upstream.

Either way, I'd suspect this is out of scope for this patch.

PR #5805, opened 2026-05-11
protovm: optimize node memory footprint

rukkal:
Decouple MapNode and Node types, so that for MapNode we don't pay the
extra memory cost of Node (std::variant internal state, bool is_being_merged).

This reduces a MapNode's memory footprint from 64 to 48 bytes on Linux x64.

### DIFF
--- a/src/protovm/allocator.h
+++ b/src/protovm/allocator.h
@@ -17,6 +17,7 @@
 #ifndef SRC_PROTOVM_ALLOCATOR_H_
 #define SRC_PROTOVM_ALLOCATOR_H_
 
+#include <algorithm>
 #include <cstdlib>
 
 #include "perfetto/protozero/field.h"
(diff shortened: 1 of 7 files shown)


src/protovm/allocator.h
```diff
@@ -33,13 +34,34 @@ namespace protovm {
...
+    PROTOVM_RETURN_IF_NOT_OK(p);
+
+    // 1. Construct node's inner value T passing args (struct aggregate
+    // initialization)
+    // 2. Construct node passing T (struct aggregate initialization)
+    new (*p) Node{T{std::forward<Args>(args)...}};
+
+    return OwnedPtr<Node>(static_cast<Node*>(*p));
+  }
+
```
primiano, line 55: how come you are no longer updating used_memory_bytes_?  is that intended or accidental?

primiano (approved):
LGTM with one question on used_memory_bytes_

src/protovm/allocator.h
rukkal: No worries, the used_memory_bytes_ update is still there. It was only moved into the Allocate() helper function.

PR #5817, opened 2026-05-12
trace_processor: Parse adreno_cmdbatch_retired/sync events

rossning92:
Add custom parsing for kgsl adreno_cmdbatch_sync and adreno_cmdbatch_retired ftrace events to produce GPU timeline slices in the trace processor.
image

This is a rework of #5519 (reverted in #5793) which fixes the GPU event drift in the timeline by using per-cmdbatch sync points instead of a single global clock snapshot. 
Confirmed this PR fixes all drift issues that @batesj has observed on his end.

Test run result:

```
$ tools/ninja -C out/host
[16/16] stamp obj/default.stamp

$ .venv/bin/python3 tools/diff_test_trace_processor.py out/host/trace_processor_shell --name-filter '.*adreno.*'
[==========] Running 1 tests.
[ RUN      ] AdrenoCmdbatch:adreno_cmdbatch_retired_slice
[       OK ] AdrenoCmdbatch:adreno_cmdbatch_retired_slice (ingest: 738.62 ms query: 0.44 ms)
[==========] Name filter selected 1 tests out of 1357.
[==========] 1 tests ran out of 1 total. (1115 ms total)
[  PASSED  ] 1 tests.
```

### DIFF
--- a/src/trace_processor/importers/ftrace/ftrace_parser.cc
+++ b/src/trace_processor/importers/ftrace/ftrace_parser.cc
@@ -542,6 +542,7 @@ FtraceParser::FtraceParser(TraceProcessorContext* context,
       gpu_power_state_off_id_(context->storage->InternString("OFF")),
       gpu_power_state_pg_id_(context->storage->InternString("PG")),
       gpu_power_state_on_id_(context->storage->InternString("ON")),
+      gpu_cmdbatch_slice_name_id_(context->storage->InternString("GPU")),
       ddic_underrun_id_(context_->storage->InternString("ddic_underrun")),
       panel_settings_full_id_(
           context_->storage->InternString("panel_settings_full")),
@@ -947,6 +948,14 @@ base::Status FtraceParser::ParseFtraceEvent(uint32_t cpu,
       }
       case FtraceEvent::kKgslGpuFrequencyFieldNumber: {
         ParseKgslGpuFreq(ts, fld_bytes);
+        break;
+      }
+      case FtraceEvent::kKgslAdrenoCmdbatchQueuedFieldNumber: {
+        ParseKgslAdrenoCmdbatchQueued(pid, fld_bytes);
+        break;
+      }
+      case FtraceEvent::kKgslAdrenoCmdbatchRetiredFieldNumber: {
+        ParseKgslAdrenoCmdbatchRetired(ts, fld_bytes);
         break;
       }
       case FtraceEvent::kCpuIdleFieldNumber: {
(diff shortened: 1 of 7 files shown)


batesj:
I'll try this out shortly. Give me a few days though - hopefully this is not in a rush.

rossning92:
@batesj no rush many thanks for the help!

batesj:
@rossning92 I'm getting a build error:
```
../../src/trace_processor/importers/ftrace/ftrace_tokenizer.cc:551:24: error: no viable conversion from 'TracePacketData' to 'FtraceData'
```

rossning92:
@batesj My bad. I did a rebase on to main branch, but apparently some structure name changed.
I just fixed it and retested. All GPU trace event is still fully aligned with the sql query you have shared earlier.

batesj:
@rossning92 This is working much better! I think there's only one remaining issue where initial clipped events are interpreted wrong. Looks like if an event was clipped at the start of ftrace events (like there's only an end event with no begin event), this CL will turn it into an even that starts at the end time instead. You can see in this screenshot the start of ftrace events there's an offset GPU event in Ctx=25. In my kgsl query I get no event for that:

image

Is there a way to drop that instead when there's no begin event?

batesj:
Other examples of additional events showing at the start of the trace that don't all appear to be related to clipped events. Is it possible that this CL is being a bit too flexible in interpreting events with partial data or something like that?

image

The red square above is zoomed in below:

image

rossning92:
@batesj thanks for trying it! Indeed, the initial clipped/drifted events are because they are not paired with `adreno_cmdbatch_sync` (the trace misses a few initial sync event). For such cases, we should simply skip generating GPU slices to avoid confusion.

I updated the PR and I'm pretty confident that it should now match the parity of your SQL query. On top of that, I also attached the process name to the track name (whenever `adreno_cmdbatch_queue` events are available)
image

Thanks so much again for such helpful feedback!

batesj:
LGTM, thanks @rossning92 !

batesj:
@LalitMaganti this works with my test cases now, PTAL.

src/trace_processor/importers/ftrace/ftrace_parser.cc
```diff
@@ -1931,6 +1944,80 @@ void FtraceParser::ParseKgslGpuFreq(int64_t timestamp, ConstBytes blob) {
...
+    base::StackString<64> name("GPU (Ctx=%u, Prio=%u)", context_id,
+                               static_cast<uint32_t>(evt.prio()));
+    track_name = context_->storage->InternString(name.string_view());
+  }
+
+  TrackId track_id = context_->track_compressor->InternScoped(
+      kBlueprint,
+      tracks::Dimensions(context_id, static_cast<uint32_t>(evt.prio())),
+      gpu_start_ts, duration, tracks::DynamicName(track_name));
+  context_->slice_tracker->Scoped(gpu_start_ts, track_id, kNullStringId,
```
LalitMaganti: This doesn't work the the way you want it to unfortunately: you should never insert a timestamp here which didn't pass through the sorter queues or you'll silently break all downstream tables which expect slice to be sorted by timestamp.

This has to be passed through the tokenizer as you had before.

rossning92: @LalitMaganti ah, that does make sense to me  (I chose this approach since updating the timestamp has a side effect where those original ftrace events will also be shifted (which may be confusing if the user directly selects the ftrace event). 

Curious Lalit if there is an approach that works without shifting the original ftrace events yet won't break the global sort order, i can't seem to find existing code path that does this.. otherwise I am happy to revert back to the original approach.

LalitMaganti: You need to build upon https://github.com/google/perfetto/pull/5809 which did this for dropping events. We should use the same "raw ftrace timestamp" for all events where it's avialable when adding to the ftrace_event table. 

I was going to do this but didn't get round to it. I would ask you do that in a separate dependent CL before this one can land. That way it solves your problem and in general imrpoves the behaviouir for the other ~6 events which are encountering this same problem.

rossning92: @LalitMaganti thanks! will give it a shot

rossning92: > use the same "raw ftrace timestamp" for all events where it's avialable when adding to the ftrace_event table.

@LalitMaganti  I had one question after some code dive - seems like ftrace_event table is also declared SORTED. If we directly `mutable_ftrace_event_table()->Insert({raw_ts...})`, that will also break the binary search right unless we remove the ColumnFlag.SORTED flag which hurts the perf.. sorry if i misinterpret.

LalitMaganti: Ahhhhhhh nice catch. 

Ok in that case I think what we actually need to do is at tokenization time, split the event into 2 with a flag and then pass it through the sorter queues twice. One purely for raw ftrace adding and one for event parsing. Does that make sense?

rossning92: Updated the PR using the ftrace dual push approach implemented in https://github.com/google/perfetto/pull/6058. Re-tested and confirmed all generated GPU events are well sorted, all tests passed.

LalitMaganti (approved):

sashwinbalaji (approved):

PR #5818, opened 2026-05-12
tp: merge Wattson interconnect power into cpu_2d curves

runpinglai-creator:
Background: In wattson, suspend is always mixed up with deep idle, both modeled as using 0 power. A lot of wattson regressions are due to suspend, either suspending longer or shorter than expected. Either way, wattson should be able to differentiate between suspend vs deep idle.

This change introduces active baseline power modeling for the interconnect and system bus in Wattson. This power is modeled under a new `interconnect` column in `cpu_2d.csv` and integrated directly into the "DSU/SCU estimate"
track.

The interconnect baseline represents the public power overhead of the SOC whenever the system is awake (!suspended), keeping the power domain model unified and physically accurate.


src/trace_processor/perfetto_sql/stdlib/wattson/cpu/estimates.sql
```diff
@@ -130,7 +133,12 @@ LEFT JOIN _filtered_curves_2d AS lut7
 LEFT JOIN _filtered_curves_l3 AS l3_lut
   ON l3_lut.freq_khz = base.freq_0
   AND l3_lut.dep_policy = base.dep_policy_0
-  AND l3_lut.dep_freq = base.dep_freq_0;
+  AND l3_lut.dep_freq = base.dep_freq_0
+LEFT JOIN _filtered_curves_interconnect AS interconnect_lut
+  ON interconnect_lut.freq_khz = base.freq_0
+  AND interconnect_lut.dep_freq = base.dsu_freq
+  AND interconnect_lut.policy = 0
+  AND interconnect_lut.dep_policy = 255;
```
wusamuel6: AND interconnect_lut.dep_policy = _dsu_dep!()
instead of 255.

src/trace_processor/perfetto_sql/stdlib/wattson/cpu/estimates.sql
```diff
@@ -150,8 +158,13 @@ SELECT
   + (coalesce(slices.l3_hit_count * base.l3_hit, 0)
   + coalesce(slices.l3_miss_count * base.l3_miss, 0))
   * 1000
-  / slices.dur AS dsu_scu_mw
+  / slices.dur
+  + iif(suspend.suspended = 1, 0, coalesce(base.interconnect_mw, 0)) AS dsu_scu_mw
 FROM _w_independent_cpus_calc AS slices
 JOIN _unique_estimates_mw AS base USING (config_hash)
+LEFT JOIN _gapless_suspend_slices AS suspend
```
wusamuel6: This is weird logic, why does it need to check for time window overlap?
And if we do need time window overlap, we should use `interval_intersect!()`

src/trace_processor/perfetto_sql/stdlib/wattson/cpu/estimates.sql
runpinglai-creator: done

src/trace_processor/perfetto_sql/stdlib/wattson/cpu/estimates.sql
runpinglai-creator: This was because upstream tables didn't pass "suspend" state down here. So I had to check the suspend overlap here to determine whether it's suspended here. Now I've passed the "suspend" states from upstream tables. Is this the concern?

src/trace_processor/perfetto_sql/stdlib/wattson/cpu/estimates.sql
```diff
@@ -130,7 +131,12 @@ LEFT JOIN _filtered_curves_2d AS lut7
 LEFT JOIN _filtered_curves_l3 AS l3_lut
   ON l3_lut.freq_khz = base.freq_0
   AND l3_lut.dep_policy = base.dep_policy_0
-  AND l3_lut.dep_freq = base.dep_freq_0;
+  AND l3_lut.dep_freq = base.dep_freq_0
+LEFT JOIN _filtered_curves_interconnect AS interconnect_lut
+  ON interconnect_lut.freq_khz = base.freq_0
+  AND interconnect_lut.dep_freq = base.dsu_freq
+  AND interconnect_lut.policy = 0
+  AND interconnect_lut.dep_policy = _dsu_dep!();
```
wusamuel6: This is hard-coded to always have interconnect dependent on DSU.

The code in `pivot.sql` should be able to dynamically figure out this. The `_cpu_lut_dependencies` table will figure out a CPU's dependency, so we should be able to do the same here with the interconnect.

src/trace_processor/perfetto_sql/stdlib/wattson/cpu/pivot.sql
```diff
@@ -41,7 +41,8 @@ AS (
     t1.curve_value AS $curve_col,
     iif($cpu IN _device_policies, coalesce(t1.static, 0), 0) AS $static_col,
     coalesce(t1.freq, 0) AS $freq_col,
-    coalesce(t1.idle, deepest.idle) AS $idle_col
+    coalesce(t1.idle, deepest.idle) AS $idle_col,
+    t1.suspended
```
wusamuel6: I think adding suspended column to every CPUx table is a little awkward, as it's a property of the system instead of per CPU. There might be performance and memory regression doing this too.

I think best way might be to:
1. Remove making a per CPU rows in`_gapless_suspend_slices`
2. interval_intersect the nominal suspend table after per CPU calculations, in `_all_stats`

src/trace_processor/perfetto_sql/stdlib/wattson/cpu/estimates.sql
runpinglai-creator: now it's dynamically looking up for CPU dependencies in src/trace_processor/perfetto_sql/stdlib/wattson/cpu/estimates.sql line 135

src/trace_processor/perfetto_sql/stdlib/wattson/cpu/pivot.sql
runpinglai-creator: Great performance insight! 
I've refactored the logic exactly as suggested:
1. Reverted `_gapless_suspend_slices` to a nominal CPU-less table.
2. Removed it from the CPU-partitioned intersect in `freq_idle.sql` to prevent unpivoting performance regressions.

runpinglai-creator: Also, in order to support more than 10 tables passed into interval-intersect. I had to insert another commit to upgrade `_interval_intersect!` to support 15 tables.

wusamuel6 (approved):

LalitMaganti (approved):

PR #5824, opened 2026-05-12
stdlib: Add bitmap specific table with bitmap metadata.

zezeozue:
Decoded the `Bitmap.mId` and `Bitmap.mSourceId` into pid + pixel-storage-type components and resolved the parcel sender's upid + process name when the trace contains process information.


src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/bitmap.sql
```diff
@@ -0,0 +1,270 @@
...
+      MAX(
+        CASE
+          WHEN p.field_name = 'android.graphics.Bitmap.mSourceId' THEN p.long_value
+        END
+      ) AS source_id_raw
+    FROM bitmap_objects AS b
+    LEFT JOIN heap_graph_primitive AS p
+      ON p.field_set_id = b.field_set_id
+    GROUP BY
+      b.object_id
```
ilkos, line 200: Do we need this aggregation in the first place? I'd expect these are all 1:1

ilkos (approved):
Thanks!

src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/bitmap.sql
zezeozue: removed

zezeozue: actually. needed to bring back. sqlite picks aribtrary rows without the max

PR #5826, opened 2026-05-12
DNS: ui: Heap Dump Explorer diffing

zezeozue:
Adds a baseline-vs-current diff mode to the Heap Dump Explorer, introducing per-tab diff views (Overview, Classes, Objects, Dominators, Bitmaps, Strings, Arrays). This supports both same-trace workflows (multiple snapshots in one file) and cross-trace workflows (separate primary and baseline files).

A key feature is the same-trace flamegraph diff, which paints each node by its Δ direction while preserving the standard per-name palette hue. Grown nodes saturate and darken, shrunk nodes lighten and desaturate, unchanged nodes stay neutral, and new nodes receive a saturation boost. This ensures natural pprof-style color continuity between diff and non-diff modes rather than relying on absolute red/blue diff coloring.


LalitMaganti:
Wow exciting stuff! I'd love to test this but CI build seems to be failing. Will take a look immediately after the bot gives me a build!

matteozanettii (approved):

matteozanettii (commented):
Suggestion: the checks may have failed due to the the file's size

lukaville:
Looking forward to have this available!

> A key feature is the same-trace flamegraph diff, which paints each node by its Δ direction while preserving the standard per-name palette hue

Do you think it's important to keep the name-based palette in diffing mode? Imho when using single color it is much easier to spot increased areas. Maybe we could have an option to switch between single color/saturation based color?

Btw I have also just tried the diffing on a 80mb hprof dump, and the page sometimes hangs and renders blank page (I got the heap dumps from internal bug 513156264), not sure if there is anything we can do to optimize it

image

zezeozue:
> Looking forward to have this available!
> 
> > A key feature is the same-trace flamegraph diff, which paints each node by its Δ direction while preserving the standard per-name palette hue
> 
> Do you think it's important to keep the name-based palette in diffing mode? Imho when using single color it is much easier to spot increased areas. Maybe we could have an option to switch between single color/saturation based color?
> 
> Btw I have also just tried the diffing on a 80mb hprof dump, and the page sometimes hangs and renders blank page (I got the heap dumps from internal bug 513156264), not sure if there is anything we can do to optimize it
> 
> image

Thanks for the test. In my latest patchset locally, I've switched to color based diff. I'll upload that soon. I'm also aware of the hangs. And I'm investigating.

PR #5830, opened 2026-05-12
ui: Rename HeapDumpExplorer CSS classes from ah- to pf-hde-

zezeozue:
The HeapDump Explorer plugin is now in line with the repo-wide pf- CSS namespace convention documented in docs/AGENTS-ui.md

Pure rename of every "ah-" / "--ah-" token to "pf-hde-" / "--pf-hde-" across styles.scss and all .ts class strings (no behavioural change).


stevegolton (approved):
Thanks for this!

PR #5833, opened 2026-05-12
[ChromeScrollJank plugin] Add a command for visualizing frame timelines

petrcermak:
This commit adds a new "Chrome Scroll Jank: Show frame timelines" command (accessible via Ctrl/Cmd+Shift+P) which adds a new track that visualizes possible frame timelines provided by Android to Chrome.

### DIFF
--- a/ui/src/plugins/org.chromium.ChromeScrollJank/index.ts
+++ b/ui/src/plugins/org.chromium.ChromeScrollJank/index.ts
@@ -36,6 +36,15 @@ import {
   SCROLL_TIMELINE_V4_TRACK,
 } from './tracks';
 import {createEventLatencyModel} from './event_latency_model';
+import {
+  createFrameTimelineModel,
+  FrameTimelineModel,
+} from './frame_timeline_model';
+import {createFrameTimelineTrack} from './frame_timeline_track';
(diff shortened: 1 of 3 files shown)


ui/src/plugins/org.chromium.ChromeScrollJank/index.ts
```diff
@@ -277,4 +281,25 @@ export default class implements PerfettoPlugin {
       }
     }
   }
+
+  private registerShowFrameTimelinesCommand(ctx: Trace, group: TrackNode) {
```
kartarsingh-google, line 291: Is there a way to add some error message as well, in case "correct" categories are not enabled in trace config?

petrcermak: I added some code which shows a warning dialog when the resulting track contains no timelines. We should consider this for the other tracks as well, though it might be better to show the warning directly on the track.

kartarsingh-google: Thanks Petr, looks good to me.

kartarsingh-google (approved):

PR #5834, opened 2026-05-12
ui: Restrict Wattson GPU track to supported devices

runpinglai-creator:
Dynamically check if the current device has GPU curves before enabling
the Wattson GPU estimate track. This replaces the hardcoded SoC checks
with a query to `_gpu_filtered_curves` in `wattson.curves.utils`. The
GPU track will only show up if the trace contains GPU frequency events,
GPU power state events, and the device has compatible GPU curves.

Bug: 512481646
Test: UI test

### DIFF
--- a/ui/src/plugins/org.kernel.Wattson/index.ts
+++ b/ui/src/plugins/org.kernel.Wattson/index.ts
@@ -215,12 +215,14 @@ async function hasWattsonGpuSupport(engine: Engine): Promise<boolean> {
   const result = await engine.query(`
     INCLUDE PERFETTO MODULE android.gpu.frequency;
     INCLUDE PERFETTO MODULE wattson.gpu.freq_idle;
+    INCLUDE PERFETTO MODULE wattson.curves.utils;
     SELECT
       EXISTS (SELECT 1 FROM android_gpu_frequency) as freq,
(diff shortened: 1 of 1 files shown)


ui/src/plugins/org.kernel.Wattson/index.ts
```diff
@@ -215,12 +215,17 @@ async function hasWattsonGpuSupport(engine: Engine): Promise<boolean> {
...
     INCLUDE PERFETTO MODULE android.gpu.frequency;
     INCLUDE PERFETTO MODULE wattson.gpu.freq_idle;
+    INCLUDE PERFETTO MODULE wattson.device_infos;
     SELECT
       EXISTS (SELECT 1 FROM android_gpu_frequency) as freq,
-      EXISTS (SELECT 1 FROM _gpu_power_state) as idle
+      EXISTS (SELECT 1 FROM _gpu_power_state) as idle,
+      EXISTS (
+        SELECT 1 FROM _wattson_device
+        WHERE name = 'Tensor' OR name = 'Tensor G5'
```
wusamuel6: We shouldn't hard code a device here, the scalable solution can be a check for if there are GPU curves present for the device

wusamuel6 (approved):

PR #5839, opened 2026-05-13
perfetto: declare empty permissions for pr-branch-check workflow

arpitjain099:
The single job in `pr-branch-check.yml` only inspects `github.base_ref` and exits 1 if a PR targets a `dev/` branch. No checkout, no API calls, no writes. `permissions: {}` matches the pattern that `cut-canary.yml` and `promote-stable.yml` already use in this repo.

YAML parses.

### DIFF
--- a/.github/workflows/pr-branch-check.yml
+++ b/.github/workflows/pr-branch-check.yml
@@ -24,6 +24,8 @@ on:
     branches:
       - dev/**/*
 
+permissions: {}
+
 jobs:
   pr-branch-check:
     if: startsWith(github.base_ref, 'dev/')


LalitMaganti:
Can you explain what the purpose of this change is? It seems like a noop to me?

arpitjain099:
@LalitMaganti `permissions: {}` isn't a noop, it's an explicit deny-all: it caps the `GITHUB_TOKEN` scope to zero instead of inheriting the repo default (which on this repo grants broader scopes). The workflow only reads `github.base_ref` from the event payload and exits with a status, no API calls or git writes, so deny-all is the actual minimum it needs.

Two existing workflows in this repo use the same pattern: `cut-canary.yml` and `promote-stable.yml`. So it's matching the in-repo convention rather than introducing a new one.

That said, if you'd rather rely on the repo default token scope for this workflow specifically, happy to close.

LalitMaganti (approved):

sashwinbalaji (approved):

PR #5849, opened 2026-05-13
Wrap around when using the keyboard to navigate the command palette

petrcermak:
If you press the "arrow up" key when the first command is selected in the command palette, select the last command. Conversely, if you press the "arrow down" key when the last command is selected in the command palette, select the first command.


stevegolton (approved):
Nice, thanks!

PR #5860, opened 2026-05-13
ui: Move distribution histogram to the left column in slice details

anakryiko:
Move the histogram/distribution panel to the bottom of right column, so that additional pieces of information like callstacks are rendered before histogram panels. 

Before:
Screenshot 2026-05-13 at 9 23 59 AM

Unless you made your details tab super high, you'd never notice that there is callstack.

After:
Screenshot 2026-05-13 at 2 59 39 PM

### DIFF
--- a/ui/src/components/details/thread_slice_details_tab.ts
+++ b/ui/src/components/details/thread_slice_details_tab.ts
@@ -326,8 +326,8 @@ export class ThreadSliceDetailsPanel implements TrackEventDetailsPanel {
         precFlows,
         followingFlows,
         args,
-        distribution,
         additionalSections,
+        distribution,
       );
     } else {
       return undefined;


LalitMaganti:
Not a fan of moving to the left. It was intentional that it was on the right, it fits better conceptually on the right.

if we want, I'm happy to move it down below the extra sections.

anakryiko:
> Not a fan of moving to the left. It was intentional that it was on the right, it fits better conceptually on the right.
> 
> if we want, I'm happy to move it down below the extra sections.

one idea I entertained (but I won't be able to pull this off implementation-wise) would be to render this histogram panel initially collapsed and let user uncollapse/collapse it back. (I think it generally makes sense to allow user to collapse/uncollapse any pane in general). WDYT?

LalitMaganti:
Not a huge fan of collapsing things either hoenstly, at least not unless we make this a common thing in the UI in generally. 

We would also need to figure out a way to persist the collapsed state, potentially even across traces as otherwise it would just be annoying to constantly have to expand/collapse.

anakryiko:
ok, let's at least move it to after callstacks? should I update my PR with that, or you'd like to do this on your own?

re: "common thing in the UI" for collapsing, that's what I was thinking and yes that makes it a bigger changes, unfortunately.

LalitMaganti:
I'm fine if we say in this patch we move post call stacks

stevegolton (approved):

LalitMaganti (approved):

Issue #3118, opened 2025-09-28
Enable compression with perfetto c++ sdk


lh35344:
Hi, I am currently integrating the perfetto sdk into my c++ project and I see the API has compression support. 

But when I try to explicitly set the compression type with `TraceConfig::set_compression_type(COMPRESSION_TYPE_DEFALTE)`, I get a runtime message that says: `COMPRESSION_TYPE_DEFALTE is not supported in the current build configuration. Skipping compression`

May I know what are the necessary steps to enable compression? Do I need to link the executable with some sort of third party compression libs like zlib or something? Currently I do not think there are clear documentations on this.

LalitMaganti:
Unfortunately the amalgamated SDK currently does not support compression: as you inferred, the problem is indeed that there is a dependency on zlib to do the actual compression. However, when the amalgamated SDK is generated, we strip out all usage of ZLIB. So it's not as simple as "just link against zlib" as the Perfetto side code which uses that has been removed.

We would need to reconfigure our build files and ifdefs to avoid doing this stripping and instead allow for SDK users to set some ifdef which would turn on zlib (with the responsibility to link against it).

Not sure when we will get to this, requires some surgery of the build files of some code which has not been touched for a long time. Maybe in a fixit which we tend to do every few months?

If you want to send a PR for this, we'd very much be in favour.

LalitMaganti:
This will be fixed by https://github.com/google/perfetto/pull/6034

PR #5862, opened 2026-05-13
sdk: honor TraceConfig.compression_type on the in-process backend

dreveman:
Mirror what `src/traced/service/service.cc` does for the system backend: when the SDK is built with `enable_perfetto_zlib=true`, construct the in-process backend's TracingService with `init_opts.compressor_fn = &ZlibCompressFn` so that `TraceConfig.compression_type = COMPRESSION_TYPE_DEFLATE` actually takes effect. Without this, the in-process service is created with default `init_opts` (compressor_fn == nullptr) and the `compression_type` field on the TraceConfig is silently ignored — embedders writing a local `.pftrace` via `TracingSession::Setup(cfg, fd)` got an uncompressed file regardless of what they set in the config.

The compressor is wired only when PERFETTO_ZLIB is enabled, matching the existing pattern in `service.cc:136-139`. No new public API; the behavior is opt-in via the TraceConfig field that already exists.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -954,6 +954,7 @@ cc_library {
         ":perfetto_src_tracing_ipc_producer_producer",
         ":perfetto_src_tracing_ipc_service_service",
         ":perfetto_src_tracing_service_service",
+        ":perfetto_src_tracing_service_zlib_compressor",
         ":perfetto_src_tracing_system_backend",
     ],
     host_supported: true,
@@ -1033,15 +1034,22 @@ cc_library {
     cflags: [
         "-DPERFETTO_REGEX_FORCE_STD",
         "-DPERFETTO_SHLIB_SDK_IMPLEMENTATION",
+        "-DZLIB_IMPLEMENTATION",
     ],
     min_sdk_version: "30",
     target: {
         android: {
             shared_libs: [
                 "liblog",
+                "libz",
             ],
             static_libs: [
                 "perfetto_flags_c_lib",
+            ],
+        },
+        host: {
+            static_libs: [
+                "libz",
             ],
         },
         windows: {
@@ -1168,6 +1176,7 @@ cc_library_static {
         ":perfetto_src_tracing_ipc_producer_producer",
         ":perfetto_src_tracing_ipc_service_service",
         ":perfetto_src_tracing_service_service",
+        ":perfetto_src_tracing_service_zlib_compressor",
         ":perfetto_src_tracing_system_backend",
     ],
     shared_libs: [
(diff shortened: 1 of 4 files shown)


LalitMaganti:
I think this is not as simple as this: we also need to do what I said in https://github.com/google/perfetto/issues/3118 - basically define some new #defines you can add to the SDK enabling ZLIB in the amalgmated SDK (but you are responsible for making zlib available to us).

dreveman:
> I think this is not as simple as this: we also need to do what I said in #3118 - basically define some new #defines you can add to the SDK enabling ZLIB in the amalgmated SDK (but you are responsible for making zlib available to us).

Works fine for me as long as I build the SDK with the following and add zlib dependency to my target:

```
tools/gen_amalgamated \
     --sdk cpp \
     --gn_args 'enable_perfetto_ipc=true enable_perfetto_zlib=true enable_perfetto_re2=false enable_perfetto_pcre2=false is_debug=false is_perfetto_build_generator=true is_perfetto_embedder=true perfetto_enable_git_rev_version_header=true use_custom_libcxx=false' \
     --output /tmp/perfetto-sdk/perfetto \
     //:libperfetto_client_experimental \
     //include/perfetto/protozero:protozero \
     //protos/perfetto/config:zero \
     //protos/perfetto/trace:zero \
     //src/tracing/service:zlib_compressor
```

I can work on adding a simpler flag to gen_amalgamated for this as a follow up if that's something we want? Either way, it would be nice to make this change so someone (like me) who is OK with maintaining a custom generated version of the SDK doesn't have to manually patch the code.

LalitMaganti:
Hmmmmm I thought the amalgamator stripped out the buildflags (at least that's what I say in the issue). I'm quite confused by this, will have to look into this a bit.

dreveman:
> Hmmmmm I thought the amalgamator stripped out the buildflags (at least that's what I say in the issue). I'm quite confused by this, will have to look into this a bit.

@LalitMaganti Did you have a chance to look at this? It'd like to get this fixed for v56 if possible

LalitMaganti:
v56 has already been cut so this will need to be v57. Let me take a look now.

LalitMaganti:
So I looked into this and I actually think https://github.com/google/perfetto/pull/6034 is a better fix. Without needing to change how the SDK is generated or people having to regenerate it themselves, it exposes cflags for people to turn on/off sdk features, fully implementing the request in https://github.com/google/perfetto/issues/3118

LalitMaganti:
https://github.com/google/perfetto/pull/6034 has landed so this has now been superseded.

PR #5867, opened 2026-05-14
tp: migrate winscope importer to be a plugin

LalitMaganti:
This a straightshot move to integrate with the plugin system. Followup
changes will consolidate all the winscope code into a single plugin
for simple interaction with the rest of trace processor.


PriyankaSPX (approved):
so iiuc we don't need to change anything in the way we build TP to include the winscope plugin? since it's still included based on PERFETTO_ENABLE_WINSCOPE in trace_processor_impl.cc

LalitMaganti:
No changes to how you need to build in this CL.

sashwinbalaji (approved):

PR #5898, opened 2026-05-15
stdlib: android_lmk_events use evt.pid instead of process.pid

gignat-dev:
This change replace process.pid with evt.pid, which is guaranteed to be set even when the LEFT JOIN fails.

We also changed the order of the columns in the SELECT.


zezeozue (approved):

src/trace_processor/perfetto_sql/stdlib/android/memory/lmk.sql
```diff
@@ -121,9 +121,9 @@ CREATE PERFETTO TABLE android_lmk_events(
   -- timestamp of the kill being requested by lmkd
   ts TIMESTAMP,
   -- upid of the process being killed
-  upid JOINID(process.id),
+  upid JOINID(process.upid),
   -- pid of the process being killed
-  pid LONG,
+  pid JOINID(process.pid),
```
LalitMaganti: Please revert this. Pid cannor be a join id because it can be reused.

src/trace_processor/perfetto_sql/stdlib/android/memory/lmk.sql
```diff
@@ -136,8 +136,8 @@ CREATE PERFETTO TABLE android_lmk_events(
 AS
 SELECT
   ts,
-  process.pid,
   process.upid,
+  cast_int!(evt.pid) AS pid,
   process.name AS process_name,
```
LalitMaganti, line 141: I'm very confused why you need to cast fo int here...

LalitMaganti (changes requested):

src/trace_processor/perfetto_sql/stdlib/android/memory/lmk.sql
gignat-dev: Done

src/trace_processor/perfetto_sql/stdlib/android/memory/lmk.sql
gignat-dev: Fixed. The int conversion needed be done earlier on table `_kill_one_process_events` which gets the pid from a counter.

Changed to

```SQL
CREATE PERFETTO TABLE _kill_one_process_events AS
WITH
  kills AS (
    SELECT c.ts, cast_int!(c.value) AS pid
```

LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/memory/lmk.sql
```diff
@@ -116,12 +116,41 @@ FROM _kill_one_process_events
 WHERE
   (SELECT s FROM selector) = '_kill_one_process_events';
 
+-- Resolves the upid for a given pid and timestamp.
+-- Uses the same logic as _android_bitmap_resolve_sender_upid to handle ambiguities.
+CREATE PERFETTO FUNCTION _android_lmk_resolve_upid(
```
gignat-dev: Added a function that finds the upid given a pid and ts.

LalitMaganti: My suspicion is this method is very fragile. Start/end ts will not be available for long lived processes nor are they guaranteed to be there on all traces (e.g. If process scraping is not present)

gignat-dev: agreed, reverted

PR #5908, opened 2026-05-15
perfetto: scope RelayPort to explicitly opted-in producer sockets

LalitMaganti:
With --enable-relay-endpoint set and multiple producer sockets bound,
traced exposed the RelayPort IPC service on every one of them,
including local AF_UNIX sockets reachable by unprivileged apps. A local
caller could bind RelayPort and cache forged SystemInfo /
RemoteClockSync data that traced later emitted with its trusted packet
sequence ID, defeating PacketStreamValidator.

Gate RelayPort exposure per ListenEndpoint via a new
expose_relay_endpoint bit honoured by ServiceIPCHostImpl. Add
--enable-relay-endpoint-on=<sock> as a narrower variant of
--enable-relay-endpoint that turns RelayPort on only for the named
producer sockets, which must already appear in
PERFETTO_PRODUCER_SOCK_NAME. On Android the init-bound local producer
socket is never relay-capable; traced.relay_producer_port keeps adding
its own relay socket. --enable-relay-endpoint retains its existing
semantics for the common single-socket multi-machine setup.


LalitMaganti:
Friendly ping on this.

docs/learning-more/multi-machine-tracing.md
```diff
@@ -55,6 +55,25 @@ UNIX path to a TCP listener that remote machines can reach.
 `--enable-relay-endpoint` makes that socket accept `traced_relay`
 connections in addition to ordinary local producers.
 
+For deployments that need to keep a local AF_UNIX producer socket
+listening alongside the TCP relay socket — so host-side unprivileged
+producers and remote `traced_relay` clients use separate sockets —
+list both in `PERFETTO_PRODUCER_SOCK_NAME` and use the narrower
+`--enable-relay-endpoint-on` form, which names the socket that should
```
sashwinbalaji, line 62: let's add it to changelog also pls

sashwinbalaji (approved):

src/traced/service/service.cc
```diff
@@ -57,9 +58,19 @@ Options and arguments
...
         <cons_group> is the group name for chgrp the consumer socket, and
         <cons_mode> is the mode bits (e.g. 0660) for chmod the consumer socket.
-    --enable-relay-endpoint : enables the relay endpoint on producer socket(s)
-        for traced_relay to communicate with traced in a multiple-machine
-        tracing session.
+    --enable-relay-endpoint : exposes the RelayPort service (used by
+        traced_relay in multi-machine tracing) on every producer socket named
+        by PERFETTO_PRODUCER_SOCK_NAME / the default producer socket. This
+        is the standard switch for multi-machine setups.
+    --enable-relay-endpoint-on <socket> : narrower variant of the above,
```
primiano, line 65: Some thoughts on this, but also I feel we have bee spent enough time on this and I'd go with your version anyways as it's too late. So read.. but also ignore the below.

rather than doing this, which sounds rather clunky, can we make it so that --enable-relay-endpoint exposes the relay only on the NON-AF-UNIX socket (ie. on TCP and VSOCK).

Yes in theory your option allows to have UNIX; TCP and VSOCK and enable relay only on VSOCK and not TCP but .... do we really care that deeply?

There is also a question of: does it even make sense to have VSOCK without relay? I can't think of a legitimate case where one might want VSOCK without relay in the first place. I'd even argue that should be implicit in vsock.

So after writing all this, I figured maybe we should go with your option, although makes setting up traced quite clunky (but also you do one-off so maybe not a big deal)

src/traced/service/service.cc
```diff
@@ -154,17 +172,14 @@ int PERFETTO_EXPORT_ENTRYPOINT ServiceMain(int argc, char** argv) {
...
-  // If a guest producer port is defined, then the relay endpoint should be
-  // enabled regardless. This is used in cases where perf data is passed
-  // from guest machines or the hypervisor to Android.
-  if (!relay_producer_socket.empty())
-    init_opts.enable_relay_endpoint = true;
+  // On Android, a relay-only producer socket can be configured via system
+  // property — e.g. when perf data is forwarded from guest VMs or the
+  // hypervisor into the host's traced. The named socket is appended to the
+  // producer endpoint list below and marked relay-capable.
+  android_relay_socket = base::GetAndroidProp("traced.relay_producer_port");
```
primiano: rather than this I'd move this code below and make it so that if the prop is set you directly add to producer_eps.
Not sure there is a lot of value to keep this assignment here and split this logic into two places (here and below)

primiano (approved):

PR #5913, opened 2026-05-16
protos: include GPU TrackEvent/InternedData extensions in trace:zero

dreveman:
The base GPU protos (gpu_counter_event.proto, gpu_render_stage_event.proto, etc.) are already pulled in via gpu:@TYPE@. The two extension protos (gpu_track_event.proto, gpu_interned_data.proto) live in stand-alone sub-targets so trace_processor can build per-extension descriptors with exclude_imports = true, but they're producer-side protos too — pull them into the main trace group so the SDK amalgamation ships their pbzero / cpp / lite headers alongside the base GPU messages.


LalitMaganti:
I'm not sure we want this to be honest. If anything we should be going in the other direction and *reducing* what we are shipping in the SDK nof adding more targets.

dreveman:
> I'm not sure we want this to be honest. If anything we should be going in the other direction and *reducing* what we are shipping in the SDK nof adding more targets. 

@LalitMaganti Ok. Sounds good. Have we documented how to generate bindings for things like this in downstream projects? I just depend on the protozero protoc plugin and use that manually to generate bindings right now but unclear to me if that's the best way to do it.

LalitMaganti:
@sashwinbalaji is working on this problem right now with the whole "out of trees protos" work. We should certainly have documentation for this when we've fully landed the work for Android.

dreveman:
> @sashwinbalaji is working on this problem right now with the whole "out of trees protos" work. We should certainly have documentation for this when we've fully landed the work for Android.

Great! @sashwinbalaji, is there a way I can track this work? I couldn't find an issue for it.

sashwinbalaji:
> Great! @sashwinbalaji, is there a way I can track this work? I couldn't find an issue for it.

I will update the RFC https://github.com/google/perfetto/discussions/4783 when it's complete

PR #5915, opened 2026-05-16
gn: Simplify logic for finding Clang resource directory

pcc:
This code was using a convoluted way of determining the resource directory that doesn't always work (for example, the resource directory is not guaranteed to have a /clang/ component). Replace it with a call to `clang -print-resource-dir`.

### DIFF
--- a/gn/standalone/toolchain/linux_find_llvm.py
+++ b/gn/standalone/toolchain/linux_find_llvm.py
@@ -46,18 +46,12 @@ def main():
   for clang in candidate_clangs():
     if shutil.which(clang) is None:
       continue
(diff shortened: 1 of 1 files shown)


LalitMaganti:
What version of clang does this work for? Because I'm not certain this was necessarily always supported.

pcc:
It's been supported since Clang 5: https://github.com/llvm/llvm-project/commit/51208a38de9ea04e3e36022c02db68d911e96d1a

LalitMaganti (approved):

sashwinbalaji (approved):

PR #5916, opened 2026-05-16
gn: Stop passing -Werror in non-hermetic builds

pcc:
Passing -Werror when the compiler version is not
controlled is considered a bad practice. See e.g.
https://blog.schmorp.de/2016-02-27-tidbits-for-the-love-of-god-dont-use-werror.html

Therefore, stop doing so in non-hermetic builds.

### DIFF
--- a/gn/standalone/BUILD.gn
+++ b/gn/standalone/BUILD.gn
@@ -267,8 +267,8 @@ config("default") {
     ]
   }
 
-  # Treat warnings as errors, but give up on fuzzer builds.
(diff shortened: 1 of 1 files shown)


LalitMaganti (approved):

sashwinbalaji (approved):

primiano (approved):
Ah yes well spotted. TBH haven't used the non-hermetic variants for ages, which I guess made that code path a bit bitrotten.
But I can also make guesses on why you are using that :)

PR #5923, opened 2026-05-17
traced_perf: Include kernel return addresses in stack traces

pcc:
These are sometimes useful for identifying the call site (e.g. in case of an indirect call or multiple calls to the same function).

### DIFF
--- a/src/profiling/perf/unwinding.cc
+++ b/src/profiling/perf/unwinding.cc
@@ -526,6 +526,7 @@ std::vector<unwindstack::FrameData> Unwinder::SymbolizeKernelCallchain(
     unwindstack::FrameData& frame = ret.emplace_back();
     frame.function_name = kernel_map->Lookup(ip);
     frame.map_info = kernel_map_info.ref();
+    frame.rel_pc = ip;
   }
   return ret;
 }


pcc:
P.S. To investigate my actual scenario I also had this diff:
```
-    frame.function_name = std::move(function_name);
+    frame.function_name = std::move(function_name) + " " +
+                          base::Uint64ToHexString(sample.kernel_ips[i]);
```
so that the flame graph entries would be discriminated by call site. We may consider adding an option to the UI to do something similar, which may also be useful for userspace stack traces.

primiano:
I'm not too familiar with this code, but I have one question: is this going to leak absolute kernel addresses? That cannot happen (As it would disclose KASLR). I'm asking because the field is called "rel"_pc but the field seems absolute (I cannot fully tell)?

> P.S. To investigate my actual scenario I also had this diff:

Hmm that cannot happen by default because it is going to break a lot of pipelines that rely on this data to populate go/apc

Also i'm not sure that's a good default behaviour as it is going to change the behaviour of analysis (e.g. pivoting on frames) if the same function can have N identities based on where the call is made

If anything this needs to be plumbed as a configuration through the perf_event_config.proto

pcc:
> I'm not too familiar with this code, but I have one question: is this going to leak absolute kernel addresses? That cannot happen (As it would disclose KASLR). I'm asking because the field is called "rel"_pc but the field seems absolute (I cannot fully tell)?

I think the kernel subtracts the KASLR offset before exposing addresses to userspace (otherwise symbol lookup would not work). That being said, I don't think I've tried it on a KASLR kernel yet.

> Hmm that cannot happen by default because it is going to break a lot of pipelines that rely on this data to populate go/apc

Right, that diff would clearly break things that are expecting an actual symbol name, so I'm not proposing that we would actually apply that diff. It was just the easiest way for me to achieve my immediate goal.

> If anything this needs to be plumbed as a configuration through the perf_event_config.proto

My idea was that we would change the UI to do the moral equivalent of that diff if some checkbox is checked (by using the now filled in rel_pc field). So I don't think any configuration would be needed at collection time.

primiano:
> My idea was that we would change the UI to do the moral equivalent of that diff if some checkbox is checked (by using the now filled in rel_pc field). So I don't think any configuration would be needed at collection time.

Ah hmm I see.
That is a setting that should be plumbed through trace processor init args through, as those stringifications are done by TP in C++ (Wasm), not the end-leaft UI typescript code.

LalitMaganti:
> That is a setting that should be plumbed through trace processor init args through, as those stringifications are done by TP in C++ (Wasm), not the end-leaft UI typescript code.

No it doesn't need parser side changes if all you want is for flamegraph to work better: you can just do in SQL in the flamegraph.

rsavitski:
> > I'm not too familiar with this code, but I have one question: is this going to leak absolute kernel addresses? That cannot happen (As it would disclose KASLR). I'm asking because the field is called "rel"_pc but the field seems absolute (I cannot fully tell)?
> 
> I think the kernel subtracts the KASLR offset before exposing addresses to userspace (otherwise symbol lookup would not work). That being said, I don't think I've tried it on a KASLR kernel yet.

We can't expose these specific kernel ips. They are the actual runtime addresses, and are symbolised directly against /proc/kallsyms.

Afaik relativising everything ourselves against one address in vmlinux (e.g. _text) is incorrect since modules (and bpf?) have their own kaslr. One approach could be to compute the offset relative to the symbol's address (i.e. offset into the function), but we'd need to separate it from rel_pc and take care not to leak the full address as a delta in corner cases.

So while the request is valid, it's not this simple, and there are higher priority features to add to the profiler atm.

PR #5930, opened 2026-05-18
tp: fix potential UAF in tree operations code

LalitMaganti:
We were holding onto pointers from SQLite which could be invalidated
at any time and so we have to make a copy. We were not doing so however
leading to potential UAF situation (though I didn't see one in practice).

Refactor out the OwnedSqlValue struct which we were already using for
functions for use here as well.

### DIFF
--- a/src/trace_processor/util/owned_sql_value.h
+++ b/src/trace_processor/util/owned_sql_value.h
@@ -83,7 +83,7 @@ class OwnedSqlValue {
       case base::variant_index<Data, std::vector<uint8_t>>():
         return SqlValue::Type::kBytes;
     }
-    PERFETTO_FATAL("Unreachable");
+    PERFETTO_FATAL("For GCC");
   }
 
   int64_t AsLong() const { return base::unchecked_get<int64_t>(data_); }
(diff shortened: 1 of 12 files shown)


src/trace_processor/util/owned_sql_value.h
```diff
@@ -0,0 +1,126 @@
...
+      case base::variant_index<Data, int64_t>():
+        return SqlValue::Type::kLong;
+      case base::variant_index<Data, double>():
+        return SqlValue::Type::kDouble;
+      case base::variant_index<Data, std::string>():
+        return SqlValue::Type::kString;
+      case base::variant_index<Data, std::vector<uint8_t>>():
+        return SqlValue::Type::kBytes;
+    }
+    PERFETTO_FATAL("Unreachable");
```
sashwinbalaji: nit: follow the existing convention of [PERFETTO_FATAL](https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/include/perfetto/base/logging.h;drc=ce2ceffaf54641bb40994094c3fb6a680bf1d3cc;l=197)("For GCC");

sashwinbalaji (approved):

PR #5932, opened 2026-05-18
Handle invalid utids in export_json.cc

betasheet:
Theoretically, trace data can override the `passthrough_utid` param in args, which is later used to lookup a tid/pid. Work around invalid values by returning invalid tid/pids when an invalid utid is supplied.

See also http://crbug.com/513155149.


LalitMaganti:
Wouldn't a better fix here be to disallow debug annotations from specifying a field with the name legacy_event? Basically make that name "privileged"?

betasheet:
> Wouldn't a better fix here be to disallow debug annotations from specifying a field with the name legacy_event? Basically make that name "privileged"?

Yeah, let's do that instead. I wasn't sure how many other places could potentially leak a utid that would be translated, but I can't spot any.

LalitMaganti (approved):

PR #5936, opened 2026-05-18
shared_lib: Expose shutdown

jmacnak:
Example usage in https://ui.perfetto.dev/#!/?s=315a987bda469405966eb46797201e11b12d247c captured using https://github.com/google/android-cuttlefish/pull/2502.

The Cuttlefish launcher host tooling uses `fork()` ... a lot ... so we introduce a wrapper that handles shutting down tracing before `fork()`ing and re-initializing after `fork()`ing. With this, we need shutdown to be available via the c shared lib.

Bug: b/378560215

### DIFF
--- a/contrib/rust-sdk/perfetto-sys/src/bindings.rs
+++ b/contrib/rust-sdk/perfetto-sys/src/bindings.rs
@@ -394,6 +394,9 @@ unsafe extern "C" {
     pub fn PerfettoProducerInProcessInit(args: *const PerfettoProducerBackendInitArgs);
 }
 unsafe extern "C" {
+    pub fn PerfettoProducerShutdown();
+}
+unsafe extern "C" {
     pub fn PerfettoProducerActivateTriggers(
         trigger_names: *mut *const ::std::os::raw::c_char,
         ttl_ms: u32,
--- a/include/perfetto/public/abi/producer_abi.h
+++ b/include/perfetto/public/abi/producer_abi.h
@@ -66,6 +66,20 @@ PERFETTO_SDK_EXPORT void PerfettoProducerSystemInit(
 PERFETTO_SDK_EXPORT void PerfettoProducerInProcessInit(
     const struct PerfettoProducerBackendInitArgs* args);
 
+// Shutdown the global in-process perfetto producer releasing any allocated
+// OS resources (threads, files, sockets, etc.)
+//
+// Note that Perfetto cannot be reinitialized again in the same process [1].
+// Instead, this function is meant for shutting down all Perfetto-related
+// code so that it can be safely unloaded, e.g., with dlclose().
+//
+// It is only safe to call this function when all threads recording trace
+// events have been terminated or otherwise guaranteed to not make any further
+// calls into Perfetto.
+//
+// [1] Unless static data is also cleared through other means.
+PERFETTO_SDK_EXPORT void PerfettoProducerShutdown(void);
+
 // Informs the tracing services to activate any of these triggers if any tracing
 // session was waiting for them.
 //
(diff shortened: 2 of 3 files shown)


primiano:
Hmm sorry I don't want to commit to maintain shutdown paths. if that code exist then people will expect that it works.
And getting shutdown works properly is an utter nightmare as it's a code path that is never used in production (outside of testing, which is very controlled)

Can you initialize the sdk only after fork?
Doing this game of shutting down before fork and re-initializing seems a recipe for bugs that I don't want anybody in our project to spend any time to debug.

it's also a questionable approach as your comment says rightly thatn once the SDK is shut down it cannot be reinitialized. so what would be the plan if you fork N times?

Substantially, I don't believe in graceful teardown/dlclose and the like . It's an extermely hard problem and it's a code path that almost never gets exercised.

I don't want to create an expectation that this is a supported use case, because then people will rightly file bugs, and those bugs usually end up requiring adding locks and synchronization that would hurt the hot paths

jmacnak:
edit: sorry, hit enter too soon

> Can you initialize the sdk only after fork?

The Cuttlefish host tooling unfortunately `fork()`s many many many times in a single launch.

> so what would be the plan if you fork N times?

https://github.com/google/android-cuttlefish/pull/2502 has the working example. TL;DR: it uses `pthread_atfork()` to register handlers to shutdown tracing (`PerfettoProducerShutdown();` and `dlclose()` to unload) before every fork and restart tracing (`dlopen()`, `PerfettoProducerSystemInit()`, etc) after every fork.

> I don't want to commit to maintain shutdown paths.

IIUC, `perfetto::Tracing::Shutdown();` already exists to support being able to shutdown in libraries that are being `dlclose()`ed (from b/237740008) so I don't believe this is introducing anything fundamentally new.

primiano:
> https://github.com/google/android-cuttlefish/pull/2502 has the working example. TL;DR: it uses pthread_atfork() to register handlers to shutdown tracing (PerfettoProducerShutdown(); and dlclose() to unload) before every fork and restart tracing (dlopen(), PerfettoProducerSystemInit(), etc) after every fork.

I think that is a recipe for troubles, I am not suggesting you do that and certainly don't want to support officially.

If you tell me what's the problem you are trying to solve, rather than the specific solution you come up with, maybe we can figure out something better.

I have a substantial disbelief on the ability of any threaded library to be able to dlopen/close and re-initialize. In my experience that will lead only to leaks, blood, agony and sadness. I don't even want to try to go there.
This is like when I see somebody trying to gracefully support exit() (rather than going nuclear with _exit()). It's not a thing, stop even trying ;) 

> IIUC, perfetto::Tracing::Shutdown(); already exists to support being able to shutdown in libraries that are being dlclose()ed (from b/237740008) 

Yes that made me sad already.. but at least they are not trying to re-initialize within the same process. you are IIUC

> The Cuttlefish host tooling unfortunately fork()s many many many times in a single launch.

Yeah but why do you initialize tracing in the pre-fork process? why do you fork midway after having initialized a threaded library?

Sorry but if you "fork many many times" without an exec(), in a process which (depends on libraries like perfetto) that create threads, I have a terrible news for you: you are in a terrible state regardless of tracing initialization: forking a multuthreaded process means that if another thread is in the middle of an allocation (or any other resource acquisition or semaphore), the forked process will be  in an inconsistent state and will deadlock on the next allocation/resource acquisition.
fork should be used:
1) In purely single-threaded processes, if you have full control of the execution flow
2) In multi-threded processes, only if followed by an exec (which will replace the whole process image anyways). In that case actually it should be a vfork()... and also you have to be very careful with FD closure (but it's doable)

anything else it's a recipe for troubles. And I neither recommend it, nor want to  be part of it

PR #5939, opened 2026-05-18
android_sdk: make libperfetto_jni host_supported

zezeozue:
Follow-up to #5924. JNIHelp.h calls __android_log_*, so liblog has to link on host too.

Add a host_supported_shared_libs map so liblog is
emitted as top-level shared_libs for libperfetto_jni.


tools/gen_android_bp
```diff
@@ -188,6 +189,12 @@ needs_libfts = [
 # All module names are prefixed with this string to avoid collisions.
 module_prefix = 'perfetto_'
 
+# host_supported modules whose listed shared libs must be linked on host
+# too (e.g. libperfetto_jni pulls __android_log_* via nativehelper/JNIHelp.h).
+host_supported_shared_libs = {
```
LalitMaganti: This seems wrong and unnecessary. The way this should work is on Android this should be shared library and on host this should be static library. Please see how other things work and follow that example.

LalitMaganti (changes requested):

LalitMaganti (approved):

PR #5941, opened 2026-05-18
ui: speed up trace_processor worker startup

LalitMaganti:
Three coupled wasm-bundle changes that together cut worker startup
from `new Worker(engine_bundle.js)` to "ready for RPCs":

1. Precompile + share the wasm module on the main thread.
   WasmEngineProxy.compileStreaming()s trace_processor_*.wasm once at
   module load and forwards the Module to every worker via the
   bootstrap message. V8's wasm code cache then reuses tiered-up code
   across workers (which also speeds up the C++ TraceProcessor ctor).
   Requires WASM_ASYNC_COMPILATION=1 and a startInit() refactor of
   WasmBridge to honour Module.instantiateWasm.

2. wasm_lib environment/filesystem opt-in.
   Defaults flip to worker-only + FILESYSTEM=0. proto_utils and
   traceconv opt back into web,worker (main-thread loader) and FS
   (MEMFS + WORKERFS) respectively. trace_processor stays on the lean
   defaults — strips Emscripten's web-detect branches and the MEMFS
   init from the worker bundle.

3. --closure 1 on the Emscripten JS glue in release builds.
   Reduces JS parse cost per worker spawn. Disabled in debug.
   Closure will rename any dynamic Module.* accesses; current JS-side
   callers only use exported runtime methods.

Aggregate browser-bench measurement is on the TP follow-up commit.


ui/src/trace_processor/wasm_engine_proxy.ts
```diff
@@ -18,6 +18,44 @@ import {EngineBase} from '../trace_processor/engine';
...
+function precompileTraceProcessorWasm(): Promise<
+  WebAssembly.Module | undefined
+> {
+  if (
+    typeof WebAssembly === 'undefined' ||
+    typeof WebAssembly.compileStreaming !== 'function'
+  ) {
+    return Promise.resolve(undefined);
+  }
+  const wasmUrl = USE_MEMORY64
```
stevegolton: Loading the wasm blobs like this feels a little nosey into the inner workings the emscripten glue code. IFAIK we just use the emscripten generated JS module everywhere else, relying on the public interface that that exposes, but now we're hard coding the wasm blob names and loading them manually. If something changes like the host directory or the file names then this will break.

It's not terrible and I realize there is some difficulty as we're on a different thread, and if it's unavoidable then that's fair enough, but I thought I'd check whether this is achievable via the module's API somehow.

stevegolton (approved):
Looks good % one comment

stevegolton (approved):

PR #5948, opened 2026-05-18
descriptors: Match by type and fields rather than name

sashwinbalaji:


src/trace_processor/util/descriptors.cc
```diff
@@ -114,9 +128,80 @@ std::optional<uint32_t> DescriptorPool::ResolveShortType(
...
+  }
+
+  // If both sides are enums, they match only if they have the same set of
+  // values (same numbers and same value names).
+  if (lhs.type() == ProtoDescriptor::Type::kEnum) {
+    return lhs.HasSameEnumValues(rhs);
+  }
+
+  // A different number of fields means different wire shape.
+  if (std::distance(lhs.fields().begin(), lhs.fields().end()) !=
```
LalitMaganti: You can just do .size here.

src/trace_processor/util/descriptors.cc
```diff
@@ -114,9 +128,80 @@ std::optional<uint32_t> DescriptorPool::ResolveShortType(
...
+  using FieldDescriptorProto = protos::pbzero::FieldDescriptorProto;
+
+  // Same descriptor: trivially equal.
+  if (lhs_idx == rhs_idx) {
+    return true;
+  }
+
+  // Coinductive cycle guard: if this pair is already being compared higher
+  // in the recursion, assume equal so recursive message graphs terminate.
+  std::pair<uint32_t, uint32_t> pair_key =
```
LalitMaganti: Please wrap this in a struct to make this more understandable, right now this is super mysterious. The constructor should do the reordering.

src/trace_processor/util/descriptors.cc
```diff
@@ -114,9 +128,80 @@ std::optional<uint32_t> DescriptorPool::ResolveShortType(
...
+    }
+
+    std::optional<uint32_t> lhs_sub_idx =
+        FindDescriptorIdx(lhs_field.resolved_type_name());
+    std::optional<uint32_t> rhs_sub_idx =
+        FindDescriptorIdx(rhs_field->resolved_type_name());
+    if (!lhs_sub_idx.has_value() || !rhs_sub_idx.has_value()) {
+      return false;
+    }
+    if (!DescriptorsStructurallyEqual(lhs_sub_idx.value(), rhs_sub_idx.value(),
```
LalitMaganti: Please rewrite using iterative stack.

src/trace_processor/util/descriptors.cc
```diff
@@ -114,9 +128,95 @@ std::optional<uint32_t> DescriptorPool::ResolveShortType(
...
+    const ProtoDescriptor& candidate_desc = descriptors_[candidate_idx];
+    if (existing_desc.type() != candidate_desc.type()) {
+      return false;
+    }
+
+    // If both sides are enums, they match only if they have the same set of
+    // values (same numbers and same value names). Enums have no sub-types to
+    // queue.
+    if (existing_desc.type() == ProtoDescriptor::Type::kEnum) {
+      if (!existing_desc.HasSameEnumValues(candidate_desc)) {
```
LalitMaganti: So actually it's not that it's exactly the same but one has to be a strict subset of the other.

src/trace_processor/util/descriptors.cc
```diff
@@ -114,9 +128,95 @@ std::optional<uint32_t> DescriptorPool::ResolveShortType(
...
+        return false;
+      }
+      continue;
+    }
+
+    // A different number of fields means different wire shape.
+    if (existing_desc.fields().size() != candidate_desc.fields().size()) {
```
LalitMaganti: I think this is fine. Again we should be looking at subsets not for exact equality.

src/trace_processor/util/descriptors.cc
```diff
@@ -69,23 +70,36 @@ FieldDescriptor CreateFieldFromDecoder(
...
-      return base::ErrStatus(
-          "Field %s is re-introduced with different type %s (was %s)",
-          field.name().c_str(), field.raw_type_name().c_str(),
-          existing_field->raw_type_name().c_str());
-    }
+  if (!existing_field) {
+    return base::OkStatus();
+  }
+
+  if (field.type() != existing_field->type()) {
```
LalitMaganti, line 85: Its also fine to *expand* the type (i.e. increase integer type etc). We should also allow for this

LalitMaganti (approved):

PR #5954, opened 2026-05-18
ui: Support viewing screenshots on mouse hover over track

betasheet:
Redesigned the screenshot track to support viewing screenshots on mouse hover.

The track now shows the most recent screenshot up until the timestamp under the cursor anywhere along the track.

- Added unit tests for findMostRecentScreenshot.
- Added integration test for hover behavior.
- Added a valid trace file for testing and updated snapshots.

PR #5958, opened 2026-05-19
ui: DNS: Screenshot track tooltip proposed example

stevegolton:
- Render screenshots as slices to the next screenshot rather than instants (which helps reduce tooltip flicker when hovering).
- Reuse SliceTrack for rendering tooltips.
- Make use of QuerySlot to handle caching.


stevegolton:
Pushed patch to original PR: https://github.com/google/perfetto/pull/5954

PR #5960, opened 2026-05-19
ui: Add new side panel extension point

stevegolton:
Gives plugins a place to surface their own UI alongside the main page. Plugins can register tabs against a new SidePanelManager (exposed on both App and Trace) and call showTab() open them; tabs registered via Trace are automatically torn down with the trace.

The panel itself lives to the right of the page content in a resizable split, with a tab strip across registered tabs so multiple plugins can coexist without fighting over the space.

Also add com.example.SidePanel example plugin which demonstrates using this panel extension point.

image

### DIFF
--- a/ui/src/core/app_impl.ts
+++ b/ui/src/core/app_impl.ts
@@ -40,6 +40,7 @@ import {PluginManagerImpl} from './plugin_manager';
 import {raf} from './raf_scheduler';
 import {Router} from './router';
 import type {SettingsManagerImpl} from './settings_manager';
+import {SidePanelManagerImpl} from './side_panel_manager';
 import {SidebarManagerImpl} from './sidebar_manager';
 import type {SerializedAppState} from './state_serialization_schema';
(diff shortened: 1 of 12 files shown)


LalitMaganti (approved):

primiano:
Didn't look at the code but I like the idea.
My main reccomendation is to make the API as close as possible semantic-wise to the bottom details panel, so if one day we create "just tabs" it's easier to reconcile these apis.
Ship it!

stevegolton:
> Didn't look at the code but I like the idea. My main reccomendation is to make the API as close as possible semantic-wise to the bottom details panel, so if one day we create "just tabs" it's easier to reconcile these apis. Ship it!

I added an [experimental JSdoc tag](https://tsdoc.org/pages/tags/experimental/) to the public API to indicate that it's new and to caution folks that it might not be stable, which gives us some wiggle room later.

To your point, the interface is very similar - render() function, title etc.

PR #5961, opened 2026-05-19
ui: Allow external hosts in Vite dev server

gignat-dev:
When a custom httpServerListenHost is provided, via `ui/run-dev-server --serve-host my.host`, disable Vite's strict Host header checks by setting allowedHosts to true. This
  allows accessing the dev server, preventing Vite from blocking the requests due to DNS rebinding
  protections.

### DIFF
--- a/ui/build.mjs
+++ b/ui/build.mjs
@@ -826,6 +826,7 @@ async function startViteDevServer() {
     configFile: pjoin(ROOT_DIR, 'ui/vite.config.mjs'),
     server: {
       host: cfg.httpServerListenHost,
+      allowedHosts: cfg.httpServerListenHost ? true : undefined,
       port,
       strictPort: false,
       headers,


zezeozue (approved):

ui/build.mjs
```diff
@@ -826,6 +826,7 @@ async function startViteDevServer() {
     configFile: pjoin(ROOT_DIR, 'ui/vite.config.mjs'),
     server: {
       host: cfg.httpServerListenHost,
+      allowedHosts: cfg.httpServerListenHost ? true : undefined,
```
stevegolton, line 829: This will allow access from **any** hostname, which sort of defeats the point of this protection.

If you add a custom hostname via the `host` prop, vite should allow this host as well. Does that not work?

stevegolton (changes requested):

PR #5971, opened 2026-05-19
tp: PerfettoSqlConnection hot-path rewrite + benchmarks

LalitMaganti:
Initialize() wraps the ~100 static-table CREATEs in one BEGIN/COMMIT,
and RegisterStaticTable collapses the 4-statement SAVEPOINT script to
2 plain statements now that the outer transaction does the batching.

ProcessFrame's hot path is rewritten to inline the SqliteSql case (the
common case for vanilla SQLite statements) and detour PerfettoSQL
extensions through ResolveExtensionStatement, keeping per-branch
locals out of the hot frame. ExecutionFrame is split into a small
hot-path struct plus an aux pointer for kInclude/kWildcard-only
fields. execution_stack_ becomes a SmallVector<..., 1> so the common
single-Execute() call has no heap alloc. A cached_parser_ on the
connection avoids the syntaqlite create/destroy round-trip per
Execute() and a PushIncludeFrame helper deduplicates the kInclude
frame construction.

sqlite3_expanded_sql is now computed lazily on first PreparedStatement::
sql() call (non-const, caches result) instead of in the constructor.

Adds perfetto_sql_connection_benchmark.cc with three Execute()
microbenches (TrivialSelect, NestedFunction, PerfettoSqlExtension).


src/trace_processor/perfetto_sql/engine/perfetto_sql_connection_benchmark.cc
```diff
@@ -0,0 +1,77 @@
...
+
+#include <memory>
+
+#include <benchmark/benchmark.h>
+
+#include "perfetto/base/logging.h"
+#include "src/trace_processor/containers/string_pool.h"
+#include "src/trace_processor/sqlite/sql_source.h"
+
+namespace perfetto::trace_processor {
```
sashwinbalaji, line 25: can we also have a benchmark for intialize optimsiations, just curious about how much gain this leads to

sashwinbalaji (approved):

src/trace_processor/perfetto_sql/engine/perfetto_sql_connection_benchmark.cc
LalitMaganti:
```
┌─────────────────────────────────────────┬─────────┬─────────┬────────┐
  │                 variant                 │  mean   │ median  │ stddev │
  ├─────────────────────────────────────────┼─────────┼─────────┼────────┤
  │ BEFORE (no outer txn, SAVEPOINT script) │ 5.84 ms │ 5.85 ms │ 35 µs  │
  ├─────────────────────────────────────────┼─────────┼─────────┼────────┤
  │ AFTER (outer txn + 2-stmt script)       │ 5.40 ms │ 5.37 ms │ 69 µs  │
  └─────────────────────────────────────────┴─────────┴─────────┴────────┘
```
Add benchmark: ~7.6% reduction. And that's for the whole of TraceProcessorImpl creation, not just for PerfettoSqlConnection creation :)

PR #5973, opened 2026-05-19
android_sdk: Host JVM test runner for PerfettoTraceTest

zezeozue:
Companion to tools/run_android_test (which runs the same test under ART on a device or emulator). tools/run_android_sdk_host_test uses $JAVA_HOME, the Maven jars fetched by `tools/install-build-deps --java-sdk`, and the host stubs landed in #5967 (plus a small AndroidJUnit4 stub added here for the test runner) to compile the perfetto Java SDK and the PerfettoTraceTest suite for host, links against libperfetto_jni.so from the gn+ninja pipeline, and runs the JUnit suite.

* src/android_sdk/jni/BUILD.gn: new jdk_include_path declare_arg + a conditional include_dirs / cflags block. Defaults empty so AOSP / device builds are untouched; the runner sets it from $JAVA_HOME/include.
* src/android_sdk/jni/host_stubs/android/log.h: stderr-mapping shim so the JNI sources can compile against the standard JNI ABI without liblog.
* src/android_sdk/java/host_stubs/androidx/test/ext/junit/runners/ AndroidJUnit4.java: stub for the test's @RunWith annotation; delegates to JUnit 4's BlockJUnit4ClassRunner.
* tools/install-build-deps --java-sdk: SHA256-pinned Maven jars (JUnit, Hamcrest, Truth, Guava, protobuf-java 4.31.1, error_prone_annotations) into buildtools/java_sdk/. Opt-in; standard build is unaffected.
* tools/run_android_sdk_host_test: runs gn gen with jdk_include_path, builds libperfetto_jni + protoc, generates Java protos for the closure of TEST_PROTOS, javac-compiles SDK + host_stubs + protos and the test, invokes JUnitCore. Picks jars by globbing buildtools/java_sdk/*.jar so version bumps live only in install-build-deps.

Usage:
  export JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64 tools/install-build-deps tools/install-build-deps --java-sdk tools/run_android_sdk_host_test # -> "OK (20 tests)" on host OpenJDK 21

 Follow-up: the 16-flag -Wno-* block in BUILD.gn suppresses standalone
 -Weverything warnings the upstream SDK sources trip. A separate CL will
 clean up the SDK sources and drop the suppression list.

### DIFF
--- a/src/android_sdk/jni/BUILD.gn
+++ b/src/android_sdk/jni/BUILD.gn
@@ -3,16 +3,6 @@ import("../../../gn/perfetto_android_sdk.gni")
 
 assert(enable_perfetto_android_java_sdk)
 
-declare_args() {
-  # Path to the host JDK include directory that contains jni.h. When set,
-  # libperfetto_jni's source set adds this and its platform-specific
-  # subdirectory to its include path so the .so can be built against a
-  # host JVM. tools/run_android_sdk_host_test sets it from $JAVA_HOME.
-  # Empty (default) on Android / AOSP builds, where jni.h comes from the
-  # NDK / jni_headers.
-  jdk_include_path = ""
-}
-
 SOURCE_SET_SOURCES = [
   "dev_perfetto_sdk_PerfettoNativeMemoryCleaner.cc",
   "dev_perfetto_sdk_PerfettoNativeMemoryCleaner.h",
--- a/src/android_sdk/jni/BUILD.gn
+++ b/src/android_sdk/jni/BUILD.gn
@@ -3,6 +3,16 @@ import("../../../gn/perfetto_android_sdk.gni")
 
 assert(enable_perfetto_android_java_sdk)
 
+declare_args() {
+  # Path to the host JDK include directory that contains jni.h. When set,
+  # libperfetto_jni's source set adds this and its platform-specific
+  # subdirectory to its include path so the .so can be built against a
+  # host JVM. tools/run_android_sdk_host_test sets it from $JAVA_HOME.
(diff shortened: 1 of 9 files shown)


src/android_sdk/jni/BUILD.gn
```diff
@@ -3,6 +3,16 @@ import("../../../gn/perfetto_android_sdk.gni")
...
 assert(enable_perfetto_android_java_sdk)
 
+declare_args() {
+  # Path to the host JDK include directory that contains jni.h. When set,
+  # libperfetto_jni's source set adds this and its platform-specific
+  # subdirectory to its include path so the .so can be built against a
+  # host JVM. tools/run_android_sdk_host_test sets it from $JAVA_HOME.
+  # Empty (default) on Android / AOSP builds, where jni.h comes from the
+  # NDK / jni_headers.
+  jdk_include_path = ""
```
LalitMaganti: All args should be declared in gn/ not one random src files

tools/run_android_sdk_host_test
```diff
@@ -0,0 +1,212 @@
...
+    visited.add(rel)
+    with open(os.path.join(ROOT_DIR, rel)) as f:
+      for line in f:
+        m = _IMPORT_RE.match(line)
+        if m and m.group(1).startswith('protos/perfetto/'):
+          stack.append(m.group(1))
+  return sorted(visited)
+
+
+def GenJavaProtos():
```
LalitMaganti: This feels like something which should be done at the GN layer not here.

LalitMaganti (approved):

PR #5976, opened 2026-05-19
Add stdlib function to fill data from trace start to end.

rzuklie:
We often want to ensure a state-like table is fully populated from trace_start to trace_end so that downstream consumers can safely use SPAN_JOIN or _interval_intersect without losing data. This handles:
- Filling gaps from trace start to first event
- Filling gaps between events
- Filling gaps from last event to trace end
- Filling the whole start to end, if no valid slices are present

As a complication, this handles partitioned data in order to support use cases such as a process state table (uid partitioned) or the suspend state table (machine_id partitioned).


src/trace_processor/perfetto_sql/stdlib/intervals/fill_gaps.sql
```diff
@@ -0,0 +1,110 @@
...
+--
+-- Example:
+-- ```
+-- intervals_fill_gaps!(
+--   (uid, package_name),
+--   (process_state),
+--   process_state_table
+-- )
+-- ```
+CREATE PERFETTO MACRO intervals_fill_gaps(
```
LalitMaganti: I'd like this to be private until perfetto SQL next lands at which point this can be a first class feature b

rzuklie: Sounds good, done.

LalitMaganti (approved):

PR #5982, opened 2026-05-20
perfetto: Add trace analysis skill for AI agents

YasserDbeis:
Adds a new trace analysis skill under ai/skills/perfetto_workflow_trace_analysis. This skill guides AI agents through analyzing Perfetto traces to identify performance bottlenecks (CPU, Graphics, I/O, IPC, Memory, Power) in Android apps.

It includes domain-specific hints and delegates all query execution mechanics, tool setup, and standard library schema verification to the perfetto-infra-querying-traces skill, completely avoiding SQL and stdlib duplication. It also adds build, evaluation, and testing suites conforming to the repository's structural standards.

This change also introduces some sample evals that are more outcome oriented. These are just examples for discussion but not meant to be merged at this point as they are rooted in internal traces.

Evals referred to in the newly added `/perfetto_workflow_trace_analysis/EVAL.txtpb `:
https://docs.google.com/document/d/1VNL3vr_mpyMwMXeU6f4PNNTN-cqzGheX9Kc1PQWTi8U/edit?tab=t.p23ovvtr6ipx

### DIFF
--- a/ai/skills/perfetto_workflow_android_trace_analysis/BUILD
+++ b/ai/skills/perfetto_workflow_android_trace_analysis/BUILD
@@ -1,0 +1,31 @@
+# Copyright (C) 2026 The Android Open Source Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@perfetto_cfg//:perfetto_cfg.bzl", "PERFETTO_CONFIG")
+load("@perfetto//bazel:rules.bzl", "perfetto_agent_skill", "perfetto_validate_skill_test")
+
+package(
+    default_applicable_licenses = [PERFETTO_CONFIG.root + ":license"],
+    default_visibility = ["//visibility:public"],
+)
+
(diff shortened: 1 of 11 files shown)


YasserDbeis:
@LalitMaganti Included the sample evals from our doc in the EVALS file per the PR message, but just to show you what we have. I can remove them once this is ready for merge.

LalitMaganti:
Few meta thoughts on this skill: unfortunately I don't think this skill, as it stands, is a good 1:1 fit for how I'm envisaging skills with Perfetto to work. Concretely, I don't think a good "analyze this trace" skill makes sense. What we want to really do is layer the skills and fit into:
1) "infra skills" which are to do with the mechanics of writing queries, code styles, best practices, how to fetch the trace processor, how to fetch things from the standard library 
2) "workflow skills" which are about how you debug a rpoblem end to end using trace procesor, standard library, queries you want to run, how you cshould interpret the output etc.

Right now this skill is really tying to do both. the rpoblem is with *composition* of these skills: basically you want the agent to pick 1-2 infra skills and 1 workflow skill and putting those together, understand how best to query the trace. The problem with doing it this way is:
1) it overwhelms the context window of the agent as it takes in a lot of information which might well be irrelevant to it
2) it makes it difficult to update as now you have a M x N

Maybe what we can do is present this more as a "fallback" workflow skill which is, if you don't have a concrete workflow or it's unclear what exactly is being asked for, this can be the last resort. But really then I would still think a lot of this skill should be broken up into 3 or even more other skills, each debugging a given vertical + then the fallback which links out to the other skills.

Also from my epxerientation, sills work best if they are prescriptive (i.e. run the following set of queries, debug in this way) rather than just "hints" which honestly a) rot quickly b) can mislead the agent rather than help it c) overwheln the agent's context window.

LalitMaganti:
See also http://chat/app/chat/AAQAm1lMzUQ/topic/HGWCzlibIwo where I commented more about my philosophy here.

PR #5983, opened 2026-05-20
perfetto: enhance trace querying skill with guidelines and execution protocol

harshitrawatjgd:
Integrates high-fidelity analysis guidelines into the infra querying skill to improve query reliability and agent methodology during complex trace investigations.

Key changes:
- Added an execution protocol (Research -> Draft -> Validate -> Output) to ensure agents systematically locate schemas before drafting.
- Add more SQL best practices.
- Expanded TEST and EVAL suites to include complex wakeup chain tracing, ensuring the skill maintains high quality for advanced analytical intents.

Discussion Points:
Would love guidance on the following:
**Perfetto-stdlib Reference:** I am looking for feedback on the best way to reference the perfetto-stdlib from this skill. To prevent the agent from sometimes skipping necessary web searches for the documentation (which is the current structure), would it be more efficient to check in perfetto-stdlib.md as a local reference, or perhaps maintain a versioned perfetto-stdlib.md in the Perfetto repository for the agent to download on the fly? What is your preferred precedent for linking or referencing this?

Evals referred in the updated `EVALS.txtpb`: https://docs.google.com/document/d/1Dkgchsx4e4g5tenPjSTAflHC0Qfijsm4OPKRRznIIac/edit?tab=t.2dw8nkqw4y4k

Welcome to Perfetto!

For more details please see


ai/skills/perfetto_infra_querying_traces/SKILL.md
```diff
@@ -92,15 +92,43 @@ Notes:
...
-column schema of any specific table, view, or query result.
+these to find what's available.
+
+**Mandatory Schema and Module Search:** For every table or view you plan to use,
+you MUST find its schema in [Perfetto
+Documentation](https://perfetto.dev/docs/analysis/stdlib-docs/).
+**Don't read the entire documentation file** --- it consumes the context window.
+Follow this precise workflow:
+
+-   **Discovery and Search:** Use available web fetching or browsing tools to
```
LalitMaganti: I personally don't think relying on the stdlib docs on the website is the way to go here. Over time, whatever you will have access to (e.g. studio, Perfetto UI) *can* and *does* drift from the docs website as the docs website is served from head while the trace processor in each of those tools is in a fixed version. 

Like this skill said before, the correct way is for the agent to use trace processor to directly query the things which are important to it. In fact in https://github.com/google/perfetto/pull/5756, Anna, before she left Google, was changing it so that the documentation was generated *from* querying trace processor.

ai/skills/perfetto_infra_querying_traces/SKILL.md
```diff
@@ -173,29 +201,152 @@ The module name maps directly to the file path under the stdlib root:
...
+    library modules (for example, `sched.runnable`, `linux.perf.counters`,
+    `intervals.overlap`) which safely handle trace gaps and preemptions.
+-   **Working with Identifiers:**
+    -   **Use Unique Identifiers for Joins:** When writing SQL queries in
+        Perfetto, you must join tables using `utid` (unique thread ID) or `upid`
+        (unique process ID) instead of the regular `tid` or `pid`. **Why it's
+        useful**: The operating system recycles `TIDs` and `PIDs`, while `UTIDs`
+        and `UPIDs` remain unique for the lifetime of the trace, which prevents
+        incorrect joins.
+    -   **Don't expose raw IDs in summaries:** Columns like `id`, `utid`,
```
LalitMaganti: I'm not sure this is good advice, it can be really useful to be able to jump to a slice by id etc in the UI.

ai/skills/perfetto_infra_querying_traces/SKILL.md
```diff
@@ -173,29 +201,152 @@ The module name maps directly to the file path under the stdlib root:
...
+        `upid`, `track_id` are not stable across traces or even runs of
+        trace_processor on the same trace. They are fine to use *inside* a query
+        as join keys, but join out to a stable name (`thread.name`,
+        `process.name`, `slice.name`) before reporting results to the user.
+-   **Materialise expensive intermediate results.** `CREATE PERFETTO TABLE foo
+    AS SELECT ...` caches the result so subsequent queries don't redo the work.
+    Use `CREATE PERFETTO VIEW` if you want lazy evaluation.
```
LalitMaganti: I would not recommend CREATE PERFETTO VIEW for this honestly, it's rarely ever a good idea for interactive exploarion like this.

ai/skills/perfetto_infra_querying_traces/SKILL.md
```diff
@@ -173,29 +201,152 @@ The module name maps directly to the file path under the stdlib root:
...
+    -   For standard SQLite indexes, prepend `DROP INDEX IF EXISTS index_name;`.
+-   **`SPAN_JOIN` safety.** `SPAN_JOIN` will crash if intervals **within the
+    same input table** overlap. Always use the `PARTITIONED {column}` (for
+    example, `PARTITIONED track_id`) clause to isolate intervals.
+-   **Avoid `SELECT *` in saved queries.** Trace processor table schemas can
+    gain columns; pin the columns you actually use.
+-   **Use `EXPLAIN QUERY PLAN` if a query is slow.** It shows whether SQLite is
+    using indexes. Counter and slice tables have built-in indexes on `ts` and
+    `track_id`; queries that don't filter on either will scan the whole table.
+-   **Safe Argument Extraction.** Use `EXTRACT_ARG(arg_set_id, 'key')` to
```
LalitMaganti: I understand telling it about extract_arg, I don't understand why you would tell it about string parsing, why would an agent even do this?

ai/skills/perfetto_infra_querying_traces/SKILL.md
```diff
@@ -173,29 +201,152 @@ The module name maps directly to the file path under the stdlib root:
...
+        between two different sets of time intervals `[start1, end1]` and
+        `[start2, end2]`:
+        -   **Condition:** The intervals overlap if `start1 < end2` and `start2
+            < end1`.
+        -   **Duration:** The overlap duration is calculated as `MIN(end1, end2)
+            - MAX(start1, start2)`.
+        -   **Important:** Incomplete Perfetto slices have a duration of -1
+            (`dur = -1`). Always calculate the effective end time using `ts +
+            IIF(dur = -1, trace_end() - ts, dur)` before applying this logic.
+-   **App Startups:** Query `android_thread_slices_for_all_startups` for app
```
LalitMaganti: I think this is now getting out of generic Perfetto advice and into specific workflows. I think and below these should be separated into a separate startup/cpu  etc skill

ai/skills/perfetto_infra_querying_traces/SKILL.md
harshitrawatjgd: Thanks for sharing #5756!

I completely agree with your point about the documentation drifting from the actual trace processor version being used. In fact, in my PR description, I mentioned looking for guidance on this exact problem.

Looking at the direction in #5756, I have a few quick questions on how to best align with it:
* Is there a plan for someone else to pick up and land #5756? I just want to make sure I know how to best handle this.
* I noticed that the PR discusses a slimmed-down metadata JSON. Do we plan to keep this lightweight JSON versioned? If so, we could potentially rely on that as our local, version-matched fallback which the agent can download on the fly based on the TP version (alongside querying the intrinsic tables).
* How should we handle environments running older versions of the Trace Processor engine? Since some __intrinsic_* tables won't be available when the skill interacts with older TP binaries, should we have some fallback in the skill?

The aim of the explicit fallback instruction (if it does not find __intrinsic_* tables) is primarily to provide a fallback mechanism to the agent so that it can reference structured documentation instead of guessing modules/tables or falling back on writing manual arithmetic.

Would love your thoughts on how we can best implement the direct trace processor query approach here, and how I can adjust this PR to align with it!

ref:
Screenshot 2026-05-21 at 2 24 20 PM

ai/skills/perfetto_infra_querying_traces/SKILL.md
harshitrawatjgd: Removed this instruction, thanks.

ai/skills/perfetto_infra_querying_traces/SKILL.md
harshitrawatjgd: Makes perfect sense. I've removed the domain specific hints to keep this skill generalized. This could be added as separate references or within a dedicated skill (perhaps trace analysis).

ai/skills/perfetto_infra_querying_traces/SKILL.md
harshitrawatjgd: Updated the instruction.

ai/skills/perfetto_infra_querying_traces/SKILL.md
harshitrawatjgd: This was more of a guardrail for the agents to avoid writing messy SUBSTR or regex logic to extract keys manually if they see dictionary or JSON-like data, and instead rely on functions like `EXTRACT_ARG`

LalitMaganti: I'm clearly missing something: extract_arg cannot help you with dictionary data at all. It prevents agents from joining with the args table only. So imo this is more confusing than helpful. 

Id you have a JSON argument you need to use the SQLite JSON functions to navigate that but that has nothing to do with extract _arg

ai/skills/perfetto_infra_querying_traces/SKILL.md
LalitMaganti: 1) I might pick up that CL at some point but it's a noop refactor really, just rebasing the docs onto the TP based system 
2) exposing that hson imo has exactly the same problems with the markdown - it's too much content for an agent. We might keep that internally in TP but again it's IMO the wrong thing to expose. I could consider exposing subsets of that on demand though, it's something I want to see how this skill evolved and make a decision 
3) from the Perfetto perspective the first versions skills will be usable will be the version of TP with the intirisc functionality. I don't particularly want to support older versions because no matter what we do they will have a bad experience. I would rather say it's not supported rather than people try and then have a bad experience.

ai/skills/perfetto_infra_querying_traces/SKILL.md
harshitrawatjgd: Ah, fair enough. I mixed up `EXTRACT_ARG` and JSON parsing in a single instruction.

The core intent was to guide agents to use standard functions than attempting raw parsing. But I see how this instruction is confusing or mixed up rather than being helpful.

Should we update it to something like:
```
- **Argument Extraction:** Use `EXTRACT_ARG(arg_set_id, 'key')` to fetch event properties instead of manually joining the `args` table.
- **JSON Parsing:** When dealing with JSON text, use standard SQLite JSON functions (for example, `json_extract()`) to extract values.
```

WDYT?

ai/skills/perfetto_infra_querying_traces/SKILL.md
```diff
@@ -92,15 +92,44 @@ Notes:
 
 PerfettoSQL ships with **intrinsic table-functions** for browsing the
 loaded standard library — modules, tables/views, functions, macros. Use
-these to find what's available; use a plain `LIMIT 0` query to read the
-column schema of any specific table, view, or query result.
+these to find what's available.
+
+**Mandatory Schema and Module Search:** For every table or view you plan to use,
```
LalitMaganti: I think, as discussed, we should not suggest to use the stdlib docs except as a last resort and ask it to instead focus on using the builtin functions/tables for doing so.

ai/skills/perfetto_infra_querying_traces/SKILL.md
```diff
@@ -92,15 +92,44 @@ Notes:
...
+  [Perfetto Documentation](https://perfetto.dev/docs/analysis/stdlib-docs/) to
+  read the column schema of any specific table or view.
 
 > **Intrinsic surface — not stable API.** The `__intrinsic_*` names below
 > are an implementation detail of trace processor. They're fair game for
 > an agent to use during a session because this skill is loaded, but
 > **don't bake `__intrinsic_*` names into committed scripts, dashboards,
 > or stdlib modules** — they can change without notice.
 
+> CRITICAL: If the __intrinsic_* tables do not exist, you MUST search the
```
LalitMaganti: This is good, we shouldn't need to specify the above.

LalitMaganti (commented):
LGTM % nits.

ai/skills/perfetto_infra_querying_traces/SKILL.md
LalitMaganti: Looks good!

ai/skills/perfetto_infra_querying_traces/SKILL.md
harshitrawatjgd: Updated instructions.

LalitMaganti (approved):

PR #5985, opened 2026-05-20
smaps: add path name redaction rules

rsavitski:
Adds optional path name redaction rules to the new /proc/pid/smaps
encoder. A config can specify an ordered set of rules, where the first
matching rule will either keep the name as is, or redact it.

Matching can be done using prefixes and fnmatch (i.e. glob patterns).
Redaction is constrained to keeping leading N path elements and file
extensions, or combinations thereof.

Note that all mappings and values are still emitted, it is only the name
of the mapping that potentially gets changed. In other words, this is
*not* a filter for conditionally collecting only specific mappings.

The redaction rules are implemented separately from the existing trace
filtering bytecode logic for efficiency. However if the need arises for
regex redaction, the bytecode can still be applied to the string_table
field.


src/profiling/smaps/smaps.cc
```diff
@@ -271,6 +281,164 @@ void Parse(FILE* file,
     callback(vma);
   }
 }
+
+bool EndsWith(std::string_view s, std::string_view suffix) {
```
primiano: can you not use base::StringView(s).EndsWith(suffix)?

src/profiling/smaps/smaps.cc
```diff
@@ -271,6 +281,164 @@ void Parse(FILE* file,
...
+  // keep = 3 for /x/y/z -> /x/y/z
+  // keep = 4 for /x/y/z -> /x/y/z
+  std::string_view keep_prefix;
+  if (rule->keep_path_elements() > 0) {
+    size_t pos = 0;
+    size_t max_elems = rule->keep_path_elements();
+    for (size_t i = 0; i < max_elems && pos != std::string_view::npos; ++i) {
+      pos = name.find('/', pos + 1);
+    }
+    size_t prefix_len = (pos != std::string_view::npos) ? pos + 1 : name.size();
```
primiano, line 365: should the condition also include `&& pos < name.size() - 1` for the edge case where the string is '/foo/bar/baz/' (which would be unlikely, probably impossible, but if it happens somehow I think it's an OOB memory access here)

src/profiling/smaps/smaps_unittest.cc
```diff
@@ -303,6 +313,254 @@ TEST(SmapsParserTest, UnaggregatedMode) {
   EXPECT_THAT(packed.aggregate_count(), IsEmpty());
 }
 
+TEST(SmapsParserTest, NameRedactionPrefixMatch) {
```
primiano, line 316: Ask AI to write some tests to cover  aggregated mode + redaction rules.
the redaction mode is way more complex and delicate (But also what we will use)

I'm not sure how much of this is true, but this is what I got by cross-checking with my agent:

```
In aggregated mode, the string table
  ordering is load-bearing (vmas are indexed by string table position).
  ConsumeStringsAndReset() now moves the strings out before redaction potentially changes
   them, but the vma vector was built with the original interned IDs. If redaction causes
   two previously distinct names (e.g., /usr/lib/a.so and /usr/lib/b.so) to collapse into
   the same redacted string (e.g., <pf_redacted>), the aggregated vector will have two
  separate entries pointing at different (now-identical) string IDs — no aggregation
  happens. This is arguably correct behavior (the PR description says "this is not a
  filter"), but it should have a test documenting this, since it's a subtle semantic that
   future maintainers might trip over.
```

primiano (approved):

src/profiling/smaps/smaps.cc
rsavitski: std::string.find does size bounds checking (xpos + str.size() <= size()): https://en.cppreference.com/cpp/string/basic_string/find

src/profiling/smaps/smaps.cc
rsavitski: I didn't want to pull in a second string view implementation just for a single suffix match. Decided to just inline into the single callsite.

src/profiling/smaps/smaps_unittest.cc
rsavitski: Added a test to show that distinct entries remain distinct even when redacted to the same name.

PR #5995, opened 2026-05-20
Migrate battery modules to _intervals_fill_gaps.

rzuklie:
This helper guarantees data from trace_start to trace_end which is important for state tables we join against.

In order to better support the common case, the macro was also updated to return a single row for the full trace when the input is totally empty.


rzuklie:
A lot of the diff is indenting a CTE. I'd recommend using the "Hide whitespace" option in the diff, or this URL: https://github.com/google/perfetto/pull/5995/changes?w=1

LalitMaganti (approved):

PR #6006, opened 2026-05-21
tp: add support for forwards looking GPU counters

### LINKED ISSUE #5683: Discarded zero values appearing as misleading large values
Looking at the combination of changes from:
- https://github.com/google/perfetto/pull/1631
- [pps: intel: Various fixes for Android CTS](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/40233/diffs?commit_id=c0c49e2b5479e6b4dab7482d10a5f5518162f1ee)

I observed that the values discarded from the pps changes are now causing the backwards-looking functionality in perfetto to pull in the next non-zero value, which becomes the value used for the entire block of now-null packets, which looks very misleading to anyone who isn't familiar with the mechanisms underneath.

Image

From the supplied image, we can see an extremely long metric counter (among other flat values) at value 75,776, but the values are actually all zero. The timestamp is also no longer correct, since the value has been shifted earlier due to the backwards-looking mechanism in the trace import.

Is there a preferred way of handling this so we don't break other usages (i.e. CTS)? Perhaps we could revert the backwards-looking mechanism if we change the pps behavior to accomodate?

Unsure where the fix for this should go without impacting others here.
@LalitMaganti, any thoughts?

LalitMaganti:
Fixes: https://github.com/google/perfetto/issues/5683

### DIFF
--- a/protos/perfetto/common/gpu_counter_descriptor.proto
+++ b/protos/perfetto/common/gpu_counter_descriptor.proto
@@ -50,6 +50,35 @@ message GpuCounterDescriptor {
     repeated MeasureUnit denominator_units = 8;
     optional bool select_by_default = 9;
     repeated GpuCounterGroup groups = 10;
+
+    // Direction of the time interval that each emitted sample applies to.
+    // A GPU counter is sampled at discrete timestamps, but its value really
+    // represents a measurement over some interval around that timestamp
+    // (e.g. average GPU utilization over the sampling window). This field
+    // tells trace processor on which side of the timestamp that interval
+    // lies, which controls where the value is drawn in the UI.
+    //
+    // Example: samples at t=10, 20, 30 with values 5, 7, 0.
+    //   BACKWARDS_LOOKING: 5 covers (-inf, 10], 7 covers (10, 20], 0 covers
+    //     (20, 30]; the UI draws 7 between t=10 and t=20.
+    //   FORWARDS_LOOKING:  5 covers [10, 20), 7 covers [20, 30), 0 covers
+    //     [30, +inf); the UI draws 5 between t=10 and t=20.
+    enum ValueDirection {
+      // Treated as BACKWARDS_LOOKING (the historical default).
+      VALUE_DIRECTION_UNSPECIFIED = 0;
+      // The value applies to the interval ending at this event's timestamp.
+      // This is the convention used by AGI and most existing GPU producers,
+      // where a sample reports what happened during the window leading up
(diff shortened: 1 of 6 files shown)


dreveman (approved):

protos/perfetto/common/gpu_counter_descriptor.proto
```diff
@@ -50,6 +50,35 @@ message GpuCounterDescriptor {
...
+      // The value applies to the interval ending at this event's timestamp.
+      // This is the convention used by AGI and most existing GPU producers,
+      // where a sample reports what happened during the window leading up
+      // to the timestamp.
+      VALUE_DIRECTION_BACKWARDS_LOOKING = 1;
+      // The value applies to the interval starting at this event's
+      // timestamp, matching standard Perfetto counters. Producers that emit
+      // explicit zero samples to denote "no activity from here on" should
+      // use this so the zero is not back-propagated over earlier non-zero
+      // samples (b/issues/5683).
```
sashwinbalaji: issue link invalid

sashwinbalaji (approved):

ccallawa-intel:
@LalitMaganti what's the best way to incorporate this change into the Mesa project? Seems the Perfetto meson.build that Mesa consumed previously is invalid now that the SDK is no longer part of the main source tree.

- https://github.com/google/perfetto/blob/d211be3e8dc2001d6cb3be8884a4533ecee9d6b7/meson.build#L40
- https://github.com/google/perfetto/blob/d211be3e8dc2001d6cb3be8884a4533ecee9d6b7/sdk/MOVED_TO_GITHUB_RELEASES_ARTIFACTS.md

LalitMaganti:
Is it possible to switch to the SDK artifact we're releasing on GitHub releases? This will be available in a week or so when we cut v56.

ccallawa-intel:
> Is it possible to switch to the SDK artifact we're releasing on GitHub releases? This will be available in a week or so when we cut v56.

Yes, pulling the SDK artifact should work fine if a release is coming in a week or so.

LalitMaganti:
Great! It should be out next Wednesday ideally :)

LalitMaganti:
@ccallawa-intel just in case you didn't see, v56 (and a fixup 56.1) has been released :)

ccallawa-intel:
@LalitMaganti yes, thanks, I saw. MR to enable this in Mesa is here: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/42035

PR #6009, opened 2026-05-22
tp: fix android input end-to-end latency app frame matching

cervantesyu:
This revision fixes the implementation of end_to_end_latency_dur in android_input_events: the first non-dropped frame of the input event needs to match the upid of the input event receiver. This fixes a bug that input event delivered to app A matches the first non-dropped frame of app B in calculating the end_to_end input latency.

Bug: b/515625953
Change-Id: Id159357c123226fb58d05b4ec62763161eea3705


LalitMaganti:
Ivan is the owner here so passing to him for review.

ivan-chong (approved):

LalitMaganti (approved):

PR #6011, opened 2026-05-22
ui: Enable MemoryViz by default and improve load performance

gignat-dev:
This change enables the com.android.MemoryViz plugin by default, exposing the Kswapd, Direct Reclaim, Memcg Reclaim, and LMK memory tracks to all users under the MEMORY group.

Additionally, to mitigate the impact on trace loading times, the track initialization functions in onTraceLoad are now executed concurrently using Promise.all.


zezeozue (approved):

ui/src/plugins/com.android.MemoryViz/index.ts
```diff
@@ -54,10 +54,12 @@ export default class MemoryViz implements PerfettoPlugin {
       INCLUDE PERFETTO MODULE android.memory.lmk;
     `);
 
-    await this.addKswapdTrack(ctx, memoryGroup);
-    await this.addDirectReclaimTracks(ctx, memoryGroup);
-    await this.addMemcgReclaimTracks(ctx, memoryGroup);
-    await this.addLmkTracks(ctx, memoryGroup);
+    await Promise.all([
```
stevegolton: Have you benchmarked this? TP is single threaded so it shouldn't be massively faster - the only thing it saves on is the enqueueing overhead so that TP always has something to process, but I'm curious to see how much that impacts things in this case.

stevegolton (approved):

ui/src/plugins/com.android.MemoryViz/index.ts
LalitMaganti: It would be nice if there was a fast query you could run here which would do an early return if this wasn't an android trace and/or it didn't have any events enabled to avoid people with non android traces spending a lot of time.

ui/src/plugins/com.android.MemoryViz/index.ts
gignat-dev: I benchmarked it on a few traces with LMKs, kswapd and direct reclaim, and indeed the gain in speed is not that big, in the order of tens of milliseconds. Should I revert this change?

ui/src/plugins/com.android.MemoryViz/index.ts
gignat-dev: I was thinking to check this android_build_fingerprint metadata
```
select extract_metadata('android_build_fingerprint') as fingerprint
```

This should be present only to android traces, right?

LalitMaganti: I would instead suggest checking for the presence of some events in the raw tables (e.g. slice, counter etc) as that's the most reliable.

zezeozue: Maybe check if there's a lmkd process (or system_server process?)

gignat-dev: Great idea, I think checking for `system_server` process should suffice

gignat-dev: Done, added `isAndroidTrace` method that checks for `system_server` process

LalitMaganti: I think this should be centralized in a plugin such that it can be reused. We should also followup and make all the Android plugins use it.

ui/src/plugins/com.android.MemoryViz/index.ts
stevegolton: Don't revert, all good, just curious.

ui/src/plugins/com.android.MemoryViz/index.ts
gignat-dev: Agreed, will do that as a followup

PR #6023, opened 2026-05-22
tp: bump SQLite per-connection lookaside slab

LalitMaganti:
Raise the lookaside slab from the SQLite default (1200,40 ~ 48 KB) to
1200x1000 (~1.2 MB) per connection. At startup we measured ~10K
lookaside hits and ~1.4K spillovers to malloc with the default; the
spilled allocations all fit in a 1200-byte slot, so growing the slot
count (not size) is what helps. Trades RAM for ~1400 fewer mallocs per
CreateInstance and gives more headroom for user queries.

### DIFF
--- a/src/trace_processor/sqlite/sqlite_connection.cc
+++ b/src/trace_processor/sqlite/sqlite_connection.cc
@@ -125,14 +125,15 @@ SqliteConnection::SqliteConnection(std::shared_ptr<SqliteDatabase> database)
   EnsureSqliteInitialized();
   PERFETTO_CHECK(sqlite3_open_v2(database_->shared_filename().c_str(), &db,
                                  kSqliteOpenFlags, nullptr) == SQLITE_OK);
-  // Bump the per-connection lookaside slab well above the SQLite default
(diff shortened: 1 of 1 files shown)


src/trace_processor/sqlite/sqlite_connection.cc
```diff
@@ -125,6 +125,14 @@ SqliteConnection::SqliteConnection(std::shared_ptr<SqliteDatabase> database)
   EnsureSqliteInitialized();
   PERFETTO_CHECK(sqlite3_open_v2(database_->shared_filename().c_str(), &db,
                                  kSqliteOpenFlags, nullptr) == SQLITE_OK);
+  // Bump the per-connection lookaside slab well above the SQLite default
+  // (1200,40 ≈ 48 KB). At startup we measured ~10K lookaside hits and
```
sashwinbalaji: nit: can we explcitly menion 1200 is the size and 40 is the count

sashwinbalaji (approved):

PR #6024, opened 2026-05-22
ui: Fix query formatting on prod

stevegolton:
Query formatting on the query page is broken on prod. This patch fixes it.

The syntaqlite wasm blobs were being placed in `dist/`, not `dist/<version>` so were never uploaded to GCS. Fix in build script and adjust URL path to match.

This should be cp'd to canary before stable release.


LalitMaganti (approved):

LalitMaganti:
You might want to cherry-pick this change into canary since I added formatting as a feature there!

PR #6025, opened 2026-05-22
ui: Show oom_score_adj on heap dump explorer overview

zezeozue:
Added an "OOM score" row under Process in the Heapdump Explorer overview, showing the process oom_score_adj at the dump instant as "bucket (score)" from the stdlib oom adjuster module. Omitted when the trace has no oom data.


zezeozue:
image

ui/src/plugins/com.android.HeapDumpExplorer/queries.ts
```diff
@@ -261,6 +261,35 @@ export async function getOverview(
...
+        AND c.ts <= ${activeDump.ts}
+      ORDER BY c.ts DESC
+      LIMIT 1
+    `);
+    const oomIt = oomRes.iter({score: NUM, bucket: STR});
+    if (oomIt.valid()) {
+      oomScore = oomIt.score;
+      oomBucket = oomIt.bucket;
+    }
+  } catch (_e) {
```
LalitMaganti: Would be nice instead of this if you looked for nulls instead.

ui/src/plugins/com.android.HeapDumpExplorer/queries.ts
```diff
@@ -261,6 +261,35 @@ export async function getOverview(
...
+    await engine.query('INCLUDE PERFETTO MODULE android.oom_adjuster;');
+    const oomRes = await engine.query(`
+      SELECT
+        CAST(c.value AS INT) AS score,
+        android_oom_adj_score_to_bucket_name(CAST(c.value AS INT)) AS bucket
+      FROM counter c
+      JOIN process_counter_track t ON c.track_id = t.id
+      WHERE t.name = 'oom_score_adj'
+        AND t.upid = ${activeDump.upid}
+        AND c.ts <= ${activeDump.ts}
```
ilkos: Unfortunately for heap dumps it's possible that the heap dump data source completes before the process stats data source which would break. APC actually does some handling for this edge case (accepts some millis of delay). Not sure if you want to handle this

zezeozue: In the worst case we'll miss a sample? I'm hesitant about adding a delay especially since in the worst case you can check the timeline

ilkos (approved):

PR #6030, opened 2026-05-25
ui: Fix column sort, reorder, and hide in QueryPage results grid

gignat-dev:
https://github.com/google/perfetto/pull/5872 fixed a bug by fixing the list of columns passed to DataGrid in the various query page results tables, however, since sorting is defined in the column list, this also means that sorting doesn't work (the buttons are present but no one is home). Less important but still notable is that column reordering and hiding is also no longer possible, however the DataGrid correctly hides these options from the UX so this isn't a bug, it's more of a missing feature which has questionable value anyway.

There are two issues here:
1. DataGrid should not show sorting options if sorting is controlled by no column callback is specified, as there is no way that sorting can be persisted in this case - it already correctly hides column re-ordering options when this callback is missing.
2. Regardless, we should allow sorting in the query results as this is a genuinely useful feature.

This patch fixes the issue by:
- Reverting DataGrid to uncontrolled mode - allow it to store the column state internally. This enables sorting and column management.
- Assign a unique id to each query result, and use it to key the ResultsTable component.
- While in the area, also move the DataSource construction (which caches the intermediate lookups) to the oninit function of the ResultsTable component as a new one will be created for each result set anyway now, which removes a lot of boilerplate caching code in the parent.

Notes:
- Switched from happy-dom to jsdom in the jest unit tests - as jsdom more faithfully represents the real browser DOM (ariaLabel -> aria-label) properly.
- This means we can un-skip a test which was previously skipped due to it not working with happy-dom.
  
Fixes: https://b.corp.google.com/issues/516394557


stevegolton (commented):
Hey @gignat-dev, thanks for the fix. I think the whole datasource / grid state management was a bit clumsy anyway, so I've pushed a new commit that:
- Assigns a unique ID (just an incrementing counter as we only need this to be session unique) to each query result produced.
- Use this ID as a key for the ResultsTable - causing it to be torn down and recreated for each query result.
- Let the DataGrid handle the sorting and column order state internally (uncontrolled mode). It will be recreated for every query result anyway.
- While we're here - move the DataSource management into the ResultsTable too, which removes a lot of useless bookkeeping code.

Let me know what you think!

gignat-dev:
> Hey @gignat-dev, thanks for the fix. I think the whole datasource / grid state management was a bit clumsy anyway, so I've pushed a new commit that:
> 
> * Assigns a unique ID (just an incrementing counter as we only need this to be session unique) to each query result produced.
> * Use this ID as a key for the ResultsTable - causing it to be torn down and recreated for each query result.
> * Let the DataGrid handle the sorting and column order state internally (uncontrolled mode). It will be recreated for every query result anyway.
> * While we're here - move the DataSource management into the ResultsTable too, which removes a lot of useless bookkeeping code.
> 
> Let me know what you think!

Hi, Steve! Yes, I think having IDs for each query execution makes results management much cleaner, and we should go with this.

stevegolton (approved):

stevegolton:
Please hold off merging for now, I'm going to fix a couple things in the datagrid on top of this.

PR #6031, opened 2026-05-25
ui: QueryPage collapsible right sidebar with persistence

gignat-dev:
Add a toggle button to the editor tab strip (right-aligned via a flex spacer) that hides/shows the History/Tables right sidebar. When hidden, the editor takes the full page width.

Visibility persists across sessions via a new `dev.perfetto.QueryPage#sidebarVisible` setting (defaultValue: true), registered next to the existing queryTabPersistence one.

image


stevegolton (approved):

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -83,9 +84,18 @@ export interface QueryPageAttrs {
   // draggedTabId is the tab being moved, beforeTabId is the tab it should be
   // placed before (or undefined if moved to the end).
   onTabReorder?(draggedTabId: string, beforeTabId: string | undefined): void;
+
+  readonly sidebarVisibleSetting: Setting<boolean>;
 }
 
 export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
+  // Whether the right-hand History/Tables sidebar is shown.
+  private sidebarVisible = true;
```
stevegolton, line 93: I just realized a bug - caching this here means that if the user modifies the setting in the settings page it won't get updated.

I'd avoid caching the setting's value here at all - just read it every time. It's designed to be efficient (reads from a cache).

gignat-dev: Thank you! 
I proposed a fix in https://github.com/google/perfetto/pull/6083

PR #6032, opened 2026-05-25
[SDK] Avoid DCHECK Track::FromPointer in uninitialized perfetto

etiennep-chromium:
It's common for perfetto to not be initialized by the test harness in Chrome.
This prevents (Named)Track::FromPointer from being used outside of TRACE_EVENT macro.

This is a common pattern:
```
auto track = perfetto::Track::FromPointer(this);
TRACE_EVENT_BEGIN(..., track);
TRACE_EVENT_END(..., track);
```
To allow doing this without extra hoops (e.g. guarding behind perfetto::Tracing::IsInitialized()), this PR avoids triggering the
DCHECK when perfetto is not initialized (process_uuid = 0).

Some thoughts: this poses a risk of *not* triggering the check if the track is created before perfetto is initialized. However: (1) this wasn't the purpose of this check anyway and it's possible to do so with other Tack api and (2) in chrome, perfetto is initialized early enough that it'd be hard to achieve.

### DIFF
--- a/include/perfetto/tracing/track.h
+++ b/include/perfetto/tracing/track.h
@@ -117,7 +117,8 @@ struct PERFETTO_EXPORT_COMPONENT Track {
     // per-proccess and the same pointer value can be used in different
     // processes. If you hit this check but are providing no |parent| track,
     // verify that Tracing::Initialize() was called for the current process.
-    PERFETTO_DCHECK(parent.uuid != Track().uuid);
+    PERFETTO_DCHECK(parent.uuid != Track().uuid ||
+                    MakeProcessTrack().uuid == 0);
 
     return Track(static_cast<uint64_t>(reinterpret_cast<uintptr_t>(ptr)),
                  parent);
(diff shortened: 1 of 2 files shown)


include/perfetto/tracing/track.h
```diff
@@ -117,7 +117,8 @@ struct PERFETTO_EXPORT_COMPONENT Track {
     // per-proccess and the same pointer value can be used in different
     // processes. If you hit this check but are providing no |parent| track,
     // verify that Tracing::Initialize() was called for the current process.
```
primiano, line 119: You are effectively invalidating this comment with he code below. The comment literally says "verify that Tracing::Initialize() was called for the current process" which  is the opposite of what you want in your PR.

IMHO either

1) We remove both the comment and the DCHECK altogether and agree to live with the risk and make this sort of bugs in production code less debuggable.

2) We create a boolean that is used to skip the checks in tests (is there some common test code where you could set this boolean?)

@betasheet thoughts?

LalitMaganti (approved):

primiano:
Sorry for the delay . I am 100% convinced I submitted this comment when you sent the PR the other day but for some reason was never sent out and was left as a draft comment

include/perfetto/tracing/track.h
etiennep-chromium: > We remove both the comment and the DCHECK altogether and agree to live with the risk and make this sort of bugs in production code less debuggable.

Why not just remove / update the comment? The DCHECK still provide value: once perfetto is initialized, it prevents `perfetto::Track::FromPointer(this, parent)` from being called with a global parent.

PR #6033, opened 2026-05-25
[SDK] Add missing TrackEventSessionObserver::OnFlush

etiennep-chromium:
This PR exposes TrackEventSessionObserver::OnFlush similar to other TrackEventSessionObserver methods,
matching DataSource::OnFlush.

FlushArgs::HandleFlushAsynchronously is not supported (PERFETTO_IMMEDIATE_CRASH)
because having multiple observers each calling it would require non trivial chaining.

Some thoughts: StopArgs::HandleStopAsynchronously exists but it's broken; it doesn't actually implement
chaining, and will only work for the first observer that tries to use it. IMO there's no good case for it and it's not worth trying to support.

### DIFF
--- a/include/perfetto/tracing/internal/track_event_data_source.h
+++ b/include/perfetto/tracing/internal/track_event_data_source.h
@@ -279,6 +279,8 @@ class PERFETTO_EXPORT_COMPONENT TrackEventDataSource
       std::move(inner_stop_args.async_stop_closure)();
   }
 
+  void OnFlush(const DataSourceBase::FlushArgs& args) override;
+
   void WillClearIncrementalState(
       const DataSourceBase::ClearIncrementalStateArgs& args) override {
     TrackEventInternal::WillClearIncrementalState(args);
--- a/include/perfetto/tracing/internal/track_event_internal.h
+++ b/include/perfetto/tracing/internal/track_event_internal.h
@@ -95,6 +95,7 @@ class PERFETTO_EXPORT_COMPONENT TrackEventSessionObserver {
   // to emit track events from this callback.
   virtual void OnStop(const DataSourceBase::StopArgs&);
   // Called when the tracing service requests a Flush.
+  // FlushArgs::HandleFlushAsynchronously is not supported.
   virtual void OnFlush(const DataSourceBase::FlushArgs&);
   // Called when tracing muxer requests to clear incremental state.
   virtual void WillClearIncrementalState(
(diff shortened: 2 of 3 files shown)


primiano:
Can you expand and explain why should we expose the OnFlush hook at the track event level?
One is not supposed to know / think about when flush happens on track event.
That is a concept that should be exposed only when one implements a full data source themselves.


I'm not completely opposed, but I need to understand what you are trying to do end-to-end so we can see if there is a better alternative

etiennep-chromium:
> Can you expand and explain why should we expose the OnFlush hook at the track event level?

The use case is to emit data when a trace snapshot is taken (with the TracingSession::CloneTrace API) similar to OnStop(). In the clone scenario, OnStop is not called, but OnFlush() is.

> One is not supposed to know / think about when flush happens on track event.

I also learned about go/perfetto-frozen because of failing tests, so maybe OnFlush() is not the way to go.

primiano:
> The use case is to emit data when a trace snapshot is taken 

Okay can you tell me more and expand. If you say this, my obvious next question is "but why?". also "which snapshot?" (are you referring to CLONE_SNAPSHOT or something else?).


Please try to explain the actual problem you have, it's a better way to have this discussion

my main reaction in lack of this context is: why don't you dump this data at the beginning of the trace in a dedicated buffer? that's what we do in android

BUt again, without a proper problem statement it's impossible for me to help

overall feels a bit odd to expose the notion of flush to track event, because the whole track event system is supposed to be one-shot emission of events

etiennep-chromium:
Context: https://chromium-review.git.corp.google.com/c/chromium/src/+/7748848

> "which snapshot?" (are you referring to CLONE_SNAPSHOT or something else?).

Chrome drives its tracing session and makes direct calls to TracingSession::CloneTrace() to achieve what I called a snapshot, but it's functionally equivalent to CLONE_SNAPSHOT AFAIU.

> why don't you dump this data at the beginning of the trace in a dedicated buffer? 

iOS is emitting system logs in the trace. The logs persist for days on disk, so we can dump it at any time. To get the most recent logs though, we need to dump the data OnStop().

In spirit, this is the job of a Data Source. But in practise, emitting the data as TRACE_EVENT is sufficient and saves a lot of boiler plate of writing track event and descriptor protos manually.

LalitMaganti:
I'm personally not a huge fan of reinventing the data source API inside of track event. We're just asking for trouble by doing this.

primiano:
> I'm personally not a huge fan of reinventing the data source API inside of track event. We're just asking for trouble by doing this.

Yeah +1 to this.

Note you can still write a TrackEvent proto from a data source (so you don't need a lot of boilerplate after all).
 But that still feels reasonable.

Essentially TrackEvent is a DataSource, but not viceversa.
So adding things to TrackEvent to make it behave like a data source feels really weird to me.

Can you create a DS and fill whatever proto you find convenient from there?

PR #6034, opened 2026-05-25
sdk: allow zlib and re2 to be enabled with the amalgamated SDK

### LINKED ISSUE #3118: Enable compression with perfetto c++ sdk

LalitMaganti:
This CL adds support for the SDK exposing cflag knobs for enabling
optional features such as use of zlib for compression and re2 for
regex usage. This depends on these headers being avaialble and the
libraries being linkable which its up to callers to enable.

We define new cflags for this purpose which feed into to buildflags
transparently so it just works.

Fixes: https://github.com/google/perfetto/issues/3118


dreveman:
Lgtm

Don't know if worth mentioning but this is not limited to the amalgamated SDK. Non-amalgamated usage of the in-process service is also gaining the compression feature.

dreveman (approved):

sashwinbalaji (approved):

primiano:
Hmm I like the end result, but I think the plumbing added is adds too much complexity vs what this actually does. I don't really like how all this is plumbed as now in order to reason about the fact that gen_amalgamated says `enable_perfetto_re2=true` but that is not really true, because turns out that `is_perfetto_amalgamated_sdk=True` makes the former a "maybe yes, maybe not", which then passes -DPERFETTO_SDK_USE_SYSTEM_ZLIB_FOR_COMPRESSION which then is picked up by the buildflag-header which gets a specific override that turns the header into an alias for PERFETTO_SDK_USE_SYSTEM_ZLIB_FOR_COMPRESSION and so on.

Try to write down an explanation of how this works, and you'll see how awkard this is.

I'll think of some counter-proposal soon, but I don't like this as-is as it solves the problem by sprinkling layers of magic in too many different places

primiano:
So I think what needs to happen here is that things like `PERFETTO_BUILDFLAG(PERFETTO_PCRE2)` remain the source of truth, and the game is to make it so when the SDK overrides, those flage evaluate to the correct thing.

There are a few ways we could achieve this. One option is that in the generated perfetto_build_flags.h let's just make it so that after the various

```
#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ENABLE_SOCK_INOTIFY() (1)
#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_RE2() (1)
#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_PCRE2() (0)
```
we add immediately below

```
#if __has_include("perfetto_config.h")
#  include "perfetto_config.h"
#endif
```

So we instruct users to just create a perfetto_config.h that looks like this:

```
#undef PERFETTO_BUILDFLAG_DEFINE_PERFETTO_RE2
#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_RE2() (1)
```

which is way more reliable than hoping to get cflags right.

We can make this more or less petty, but this is the spirit.

Furthermore, no sorry that thing you did in 
```
    # The amalgamation build picks a single regex backend at compile time,
    # but the SDK consumer chooses theirs via PERFETTO_SDK_USE_SYSTEM_*. Force
    # both the active and fallback backend headers (regex_re2.h, regex_std.h)
    # to be inlined regardless of which one the build picked.
```

is a gross hack. The way to make that proper is that:

1. base/regex.cc always includes unconditionally the various regex_xxx.h (with a comment that explains that it's needed for the amalgamated SDK)
2. The various `#if PERFETTO_BUILDFLAG(PERFETTO_PCRE2) ` get pushed down inside each regex_xxx, (like you did for zlib_compressor).  
3. You still need to keep the `#define PERFETTO_REGEX_BACKEND` in regex.cc with the existing switch matrix (just the includes become unconditional)


I think if you do this there is no need for `is_perfetto_amalgamated_sdk` nor PERFETTO_SDK_USE_SYSTEM_ZLIB_FOR_COMPRESSION and so on

LalitMaganti:
PTAL now: I didn't quite go with what you suggested since, as discussed offline, I really dislike *config.h headers a lot, they make integration really painful.

Instead, it's sort of a hybrid of what I had before and what you suggested.

gn/BUILD.gn
```diff
@@ -447,7 +448,9 @@ config("system_zlib_config") {
 # Zlib is used both by trace_processor and by perfetto_cmd.
 if (enable_perfetto_zlib) {
   group("zlib") {
-    if (perfetto_use_system_zlib) {
+    if (perfetto_amalgamated_sdk) {
```
primiano: I don't think that you need changes like this and the one below. Can we reuse the `system_library_map` that exists already in gen_amalgamated? Otherwise there is inconsistency in the way we deal with the same problem

gn/perfetto.gni
```diff
@@ -472,6 +472,9 @@ declare_args() {
   perfetto_use_system_sqlite = false
 
   perfetto_use_system_zlib = false
+
+  # Drops zlib/re2 link deps; consumer brings their own.
+  perfetto_amalgamated_sdk = false
```
primiano, line 478: see other comment. I think we shoudl deal with them by tweaking system_library_map in gen_amalgamated, not in the gn files

LalitMaganti:
PTAL again: changed as discussed offline.

gn/BUILD.gn
```diff
@@ -517,7 +520,9 @@ if (enable_perfetto_pcre2) {
 
 if (enable_perfetto_re2) {
   group("re2") {
-    if (perfetto_root_path == "//") {
+    if (perfetto_amalgamated_sdk) {
```
primiano: Can we reuse the system_library_map that exists already in gen_amalgamated? Otherwise there is inconsistency in the way we deal with the same problem.
Remove this diff here and do the same thing we do for other cases in gen_amalg

include/perfetto/base/build_config.h
```diff
@@ -225,4 +225,29 @@
...
+#include "perfetto_sdk_config.h"  // no-include-violation-check
+#endif
+#endif
+#undef PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ZLIB
+#if defined(PERFETTO_SDK_ENABLE_ZLIB) && (PERFETTO_SDK_ENABLE_ZLIB == 1)
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ZLIB() 1
+#else
+#define PERFETTO_BUILDFLAG_DEFINE_PERFETTO_ZLIB() 0
+#endif
+#undef PERFETTO_BUILDFLAG_DEFINE_PERFETTO_RE2
```
primiano, line 248: can you add some \n to parse these blocks a bit better (as a human)?

include/perfetto/base/build_config.h
```diff
@@ -225,4 +225,29 @@
 //   added to the amalgamated headers.
 #include "perfetto_build_flags.h"  // no-include-violation-check
 
+// In the amalgamated SDK, optional features (zlib, re2) are opt-in. The
+// consumer drops a header named "perfetto_sdk_config.h" somewhere on their
+// include path; inside it they `#define PERFETTO_SDK_ENABLE_ZLIB 1` /
+// `#define PERFETTO_SDK_ENABLE_RE2 1` to opt in. Setting the macro to 0 (or
+// omitting the file entirely) keeps the feature off.
+#if PERFETTO_BUILDFLAG(PERFETTO_AMALGAMATED_SDK)
```
primiano, line 233: do we actually need this?
I mean if somebody gets the header via some other way (e.g. one day with the debian package, if we expose them) i don't see much harm into doing this regardless no?

src/base/regex/regex.cc
```diff
@@ -26,23 +26,28 @@
 #include "perfetto/ext/base/flags.h"
 #include "perfetto/ext/base/status_macros.h"
 
+#include "src/base/regex/regex_std.h"
+
+#if PERFETTO_BUILDFLAG(PERFETTO_PCRE2)
```
primiano: hmm wait these includes should be unconditional. YOu moved the `#if` block in the headers, so this is not needed and actually defeats the point of letting the amalgamator see through

src/tracing/service/zlib_compressor.h
```diff
@@ -19,12 +19,15 @@
...
 
+#include "perfetto/base/build_config.h"
 #include "perfetto/ext/tracing/core/trace_packet.h"
 
 namespace perfetto {
 
+#if PERFETTO_BUILDFLAG(PERFETTO_ZLIB)
 // Matches TracingServiceImpl::kMaxTracePacketSliceSize. Exposed for testing.
 static constexpr size_t kZlibCompressSliceSize = 128 * 1024 - 512;
+#endif
```
primiano: if you really bother adding an `#if` that should cover also the function declaration, so code fails to compile rather than failing to link

tools/gen_amalgamated
```diff
@@ -101,6 +102,12 @@ define_allowlist = r'^(PERFETTO.*|GOOGLE_PROTOBUF.*)$'
 # Includes which will be removed from the generated source.
 includes_to_remove = r'^(gtest).*$'
 
+# Quoted #include lines matching this regex are kept verbatim in the
+# amalgamated output. Used for headers the SDK consumer is expected to
+# resolve from their own system install (e.g. re2/re2.h), and for the
+# optional perfetto_sdk_config.h opt-in header (guarded by __has_include).
+includes_to_leave_verbatim = r'^(re2/|perfetto_sdk_config\.h$)'
```
primiano: 2 options to avoid this (not a hufe fan as it's a surprsie factor we'll forget about):

1) Is there a reson why we use `#include "re2/..."` rather than `#include <re2/...> `  (same for the config.h)? IIRC the alamgamator already leaves verbatim the <> includes.
I can't think of a good reason. The only diffrence of "" should be that looks up first in the current directory. But we don't need that semantic.

2) If there is a good reason for NOT using <>, can we instead add a  add a `// gen_amalgamated:keep` next to each #include , e.g. `#include "re2/re2.h"  // gen_amalgamated:keep`. I think it's less of a surprise factor than having to remember that this place exists

include/perfetto/base/build_config.h
LalitMaganti: We need this because there is a desync between what the default for GN is (true) and what the default for the amalgamation is (false). I cannot think of a better way to do this other than that tbh.

primiano (approved):

PR #6040, opened 2026-05-26
perfetto: move AppWakelock fields out-of-tree

sashwinbalaji:
First subsystem migration on top of the unified extensions registry. Moves
  `app_wakelock_bundle` (TracePacket tag 116) and `app_wakelock_info`
  (InternedData tag 42) out of the in-tree perfetto protos and into a new
  `frameworks_base.proto` under `protos/third_party/android/`.

  The wire tags don't change, so existing trace files parse the same way.
  The wrapper messages (`FrameworksBaseTracePacket`, `FrameworksBaseInternedData`)
  are the proto2 way of saying "these tags belong to TracePacket / InternedData
  but live in a different file." `frameworks_base.proto` only carries
  AppWakelock for now.

  Trace-processor and traceconv are updated to read through the wrapper
  classes instead of the old direct accessors.


src/trace_processor/importers/proto/app_wakelock_module.cc
```diff
@@ -97,7 +101,12 @@ ModuleResult AppWakelockModule::TokenizePacket(
     TraceBlobView tbv =
         context_->blob_packet_writer->WritePacket([&](auto* pkt) {
           pkt->set_timestamp(static_cast<uint64_t>(real_ts));
-          auto* event = pkt->set_app_wakelock_bundle();
```
LalitMaganti, line 100: This is not the right way to do this. There is special handling of extensions.

src/trace_processor/importers/proto/app_wakelock_module.cc
```diff
@@ -115,8 +124,11 @@ void AppWakelockModule::ParseTracePacketData(
     const TracePacketData&,
     uint32_t field_id) {
   switch (field_id) {
-    case TracePacket::kAppWakelockBundleFieldNumber:
-      ParseWakelockBundle(ts, decoder.app_wakelock_bundle());
+    case FrameworksBaseTracePacket::kAppWakelockBundleFieldNumber:
+      ParseWakelockBundle(
+          ts,
+          decoder.Get(FrameworksBaseTracePacket::kAppWakelockBundleFieldNumber)
+              .as_bytes());
```
LalitMaganti: As above.

src/trace_processor/importers/proto/app_wakelock_module.cc
sashwinbalaji: thnx for the review. it's not complete yet it was for me to check forgot to make it draft

PR #6058, opened 2026-05-28
tp: use raw ftrace timestamp in ftrace_event table via dual-push

rossning92:
Custom tokenizers (gpu_work_period, thermal_exynos_acpm_bulk, param_set_value_cpm, fwtp_perfetto_counter, fwtp_perfetto_slice) forge a different sorter timestamp to reposition events on the timeline. This caused the ftrace_event table to show the forged timestamp instead of the original kernel time.

Fix this by having each custom tokenizer push the event through the sorter twice: once at the raw ftrace timestamp for the ftrace_event table, and once at the forged timestamp for event-specific parsing. Both ftrace_event.ts and slice.ts remain SORTED.

This builds on #5809 which introduced FtraceData::raw_ts for drop-window checks, and is a prerequisite for #5817 (adreno cmdbatch parsing) which needs to forge gpu_start_ts for correct slice ordering.

Bug: N/A

### DIFF
--- a/src/trace_processor/importers/common/parser_types.h
+++ b/src/trace_processor/importers/common/parser_types.h
@@ -153,6 +153,11 @@ struct alignas(8) FtraceData {
   TraceBlobView packet;
   RefPtr<PacketSequenceStateGeneration> sequence_state;
   int64_t raw_ts = kRawTsUnset;
+  // Custom tokenizers that forge a different sorter timestamp push each
+  // event twice: once at raw_ts for the ftrace_event table, and once at
+  // the forged timestamp for event-specific parsing.
+  bool insert_ftrace_event = true;
+  bool parse_event = true;
 };
 static_assert(sizeof(FtraceData) % 8 == 0);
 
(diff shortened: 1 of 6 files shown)


src/trace_processor/importers/common/parser_types.h
```diff
@@ -153,6 +153,11 @@ struct alignas(8) FtraceData {
   TraceBlobView packet;
   RefPtr<PacketSequenceStateGeneration> sequence_state;
   int64_t raw_ts = kRawTsUnset;
+  // Custom tokenizers that forge a different sorter timestamp push each
```
LalitMaganti, line 156: It would be nice if we could utilize the negative values in `raw_ts` instead of bloating this struct.

rossning92: Thanks @LalitMaganti  I thought about it, for example, to have a boolean `skip_parser` + a negative or special value in `raw_ts` can certainly distinguish 3 scenarios 1) insert ftrace + parse 2) only insert 3) only parse 

But having two booleans (and since they translate to 1-bit flags and not take up extra bytes) might just be clearer for readability and maintenance?

LalitMaganti: Well in this case they will not translate to 2 bits, they will actually still be 2 byts here but you're right it might be questuioable given these are short lived structs. Let's leave as is.

LalitMaganti (approved):

sashwinbalaji (approved):

PR #6059, opened 2026-05-28
perfetto: add proto2 extension support to protozero and trace_processor

sashwinbalaji:
Adds the shared infrastructure needed to define and decode out-of-tree (OOT) proto extensions, independent of any specific extension:

- protozero TypedProtoDecoderBase::GetExtensionSlowly() for resolving extension fields whose ids exceed the highest in-tree field id.
- DescriptorPool / FieldDescriptor support for extension full names (ExtensionInfo carries the enclosing scope), with matching protozero_to_text / protozero_to_json formatting.
- text_to_proto extension support and the tracing_proto_extensions tool, including a --base-descriptor flag so leaf extension protos can be compiled against a prebuilt FileDescriptorSet instead of the full source closure.
- Refactor existing track_event and winscope extension decoding to use GetExtensionSlowly (behaviour-preserving for in-range fields).

### DIFF
--- a/include/perfetto/protozero/proto_decoder.h
+++ b/include/perfetto/protozero/proto_decoder.h
@@ -344,6 +344,43 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
     return fields_[0];
   }
 
+  // Like Get(), but also resolves "extension" fields whose id exceeds the
+  // highest field id known in-tree at compile time (i.e. id >= num_fields_,
+  // which is MAX_FIELD_ID + 1). Such fields are NOT stored by ParseAllFields()
+  // -- e.g. fields carved out of TracePacket's `extensions 1000 to 1999`
+  // range. For in-tree ids this delegates to Get(); for extension ids it falls
+  // back to a linear re-scan of the buffer.
+  //
+  // NOTE: for *repeated* fields the slow path returns the FIRST occurrence
+  // (FindField semantics), whereas Get() returns the last. All current
+  // extension fields are singular submessages, so this is not observable today.
+  //
+  // TODO: make extension lookups fast without re-scanning the buffer. The
+  // blocker is that fields_ is indexed directly by field id (on_stack_storage_,
+  // or heap_storage_ after ExpandHeapStorage()), so it must be sized to the
+  // largest field id. That is fine for the densely-numbered in-tree fields, but
+  // extension ids are sparse and can be arbitrarily high (the OOT range starts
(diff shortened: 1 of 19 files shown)


include/perfetto/protozero/proto_decoder.h
```diff
@@ -344,6 +344,43 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
...
+  //   - Replace the flat array with a two-level sparse map. Level 1 is a set of
+  //     id ranges built dynamically while decoding (these need not match the
+  //     static extensions.json allocations), e.g. tracked with a bitset. Level
+  //     2 maps an id within a matched range to a compact slot index.
+  //   - Back the slots with an object pool so repeated decodes reuse storage
+  //     and allocate ~zero: only a few decoders are live at any time, so the
+  //     pool stays small.
+  // This costs memory proportional to the extension fields actually present
+  // rather than to the max field id, and keeps the decode hot path
+  // allocation-free.
```
LalitMaganti, line 373: There needs to be a version of this in TypedProtoDwcoder as well which is templated.

src/tools/tracing_proto_extensions.cc
```diff
@@ -50,6 +51,38 @@ using trace_processor::json::SimpleJsonParser;
 
 namespace {
 
+// Collects type-resolution errors from the DescriptorPool (e.g. "X is not
```
LalitMaganti: Sorry but I don't understand why you need this.

src/tools/tracing_proto_extensions.cc
```diff
@@ -637,13 +671,52 @@ base::StatusOr<std::vector<uint8_t>> GenerateExtensionDescriptors(
...
+      // A shared dependency may appear in more than one base set; keep the
+      // first definition and ignore later duplicates (Add() rejects dupes).
+      google::protobuf::FileDescriptorProto existing;
+      if (!base_db.FindFileByName(file.name(), &existing)) {
+        base_db.Add(file);
+      }
+    }
+  }
+
+  // Prefer the prebuilt base descriptors, fall back to compiling from source.
```
LalitMaganti: The comments here and above seem a bit unnecessary.

sashwinbalaji: Primiano had a better way: go/pgh/6062

src/tools/tracing_proto_extensions.cc
sashwinbalaji: Primiano had a better way: go/pgh/6062

LalitMaganti (approved):

PR #6068, opened 2026-05-29
[State] Add protos to describe state tracks and events [1/4]

etiennep-chromium:
Based on https://github.com/google/perfetto/blob/rfcs/0023-updatable-track-event.md

[1/4]: This PR adds state track and events proto, with no support, but minimal changes to make the build succeed.

For reference, this is the full implementation: https://github.com/google/perfetto/pull/6068


LalitMaganti:
LG but Primiano should also comment.

protos/perfetto/trace/track_event/track_event.proto
```diff
@@ -376,6 +382,19 @@ message TrackEvent {
   optional ChromeMessagePump chrome_message_pump = 35;
   optional ChromeMojoEventInfo chrome_mojo_event_info = 38;
 
+  // Represents the state value for a state track.
+  message State {
+    // The built-in string state value.
+    optional string string_value = 1;
```
etiennep-chromium: Come to think of it we might want to intern this, so either
- we add string_value_iid
- we reuse TrackEvent::name and TrackEvent::name_iid
WDYT?

protos/perfetto/trace/extensions.json
```diff
@@ -86,6 +86,25 @@
         }
       ]
     },
+    {
+      "scope": "perfetto.protos.TrackEvent.State",
```
sashwinbalaji: pls also update the top level comment in this file 

```
    "protos. Three scopes are tracked here:",
    "  * TrackEvent   -- typed-event payloads emitted on a track (`extend TrackEvent`)",
    "  * TracePacket  -- top-level trace packet types (`extend TracePacket`)",
    "  * InternedData -- per-sequence interned data types (`extend InternedData`)",
    "",
```

protos/perfetto/trace/extensions.json
```diff
@@ -86,6 +86,25 @@
...
+        {
+          "name": "perfetto_tests",
+          "range": [9000, 9999],
+          "contact": "perfetto-dev@googlegroups.com",
+          "description": "Reserved for Perfetto unit and integration tests",
+          "proto": "protos/perfetto/trace/test_extensions.proto"
+        }
+      ]
+    },
+
```
sashwinbalaji: nit: remove the line

protos/perfetto/trace/track_event/track_event.proto
etiennep-chromium: (I went with option 2)

protos/perfetto/trace/track_event/track_event.proto
```diff
@@ -159,6 +159,12 @@ message TrackEvent {
     // that most other TrackEvent fields (e.g. categories, name, ..) are not
     // supported for TYPE_COUNTER events. See also CounterDescriptor.
     TYPE_COUNTER = 4;
+
+    // Event that provides a state value update for a state track. |track_uuid|
```
primiano: Could be helpful to say here two things:

1. "Thinks of this as a symbolic counter (a TYPE_COUNTER with literals rather then numbers)"
2. A link to [the RFC](https://github.com/google/perfetto/discussions/5549)

protos/perfetto/trace/extensions.json
```diff
@@ -86,6 +87,24 @@
         }
       ]
     },
+    {
```
primiano: Not a huge fan of this. Can we just use TrackEvent extensions without creating extra nesting and a new scope here?
I don't want to set the precedent of "every new thing creates a new scope".

Is there a good reason for not just using TrackEvent extensions?

protos/perfetto/trace/track_event/track_event.proto
```diff
@@ -376,6 +382,16 @@ message TrackEvent {
   optional ChromeMessagePump chrome_message_pump = 35;
   optional ChromeMojoEventInfo chrome_mojo_event_info = 38;
 
+  // Represents the state value for a state track.
+  message State {
```
primiano: see my other comment in the JSON. why this is nested rather than just using standard extension ranges flattened below?

primiano (commented):
So overall makes sense, my main comment is about avoiding the nested message and creating yet another scope in the json extension. 
I'll update the RFC 23 to discuss there

protos/perfetto/trace/extensions.json
etiennep-chromium: Reverted to unblock, since this is being discussed in RFC

etiennep-chromium:
Updated to punt the typed state, so we can unblock the uncontroversial part.

primiano (approved):

PR #6077, opened 2026-05-30
tp: fix parsing of asterik/star in Pigweed messages

LalitMaganti:


src/trace_processor/importers/proto/pigweed_detokenizer.cc
```diff
@@ -139,13 +140,54 @@ base::StatusOr<DetokenizedString> PigweedDetokenizer::Detokenize(
...
+    if (old_ptr == ptr) {
+      return false;
+    }
+    *out = ::protozero::proto_utils::ZigZagDecode(raw);
+    return true;
+  };
+
   std::vector<std::variant<int64_t, uint64_t, double>> args;
   std::vector<std::string> args_formatted;
   for (Arg arg : format->args()) {
```
sashwinbalaji: nit: const ref

src/trace_processor/importers/proto/pigweed_detokenizer.cc
```diff
@@ -139,13 +140,54 @@ base::StatusOr<DetokenizedString> PigweedDetokenizer::Detokenize(
...
     char buffer[kFormatBufferSize];
-    const char* fmt = arg.format.c_str();
     size_t formatted_size;
 
+    // Resolve any '*' field width / precision wildcards. Each consumes an extra
+    // integer argument from the payload, which we substitute into the format
+    // string so that exactly one variadic value remains for SprintfTrunc. Not
+    // doing so would let vsnprintf read uninitialized memory.
+    std::string resolved_fmt = arg.format;
+    if (arg.width_star) {
```
sashwinbalaji: nit: since these if cases are rare can we structure it so that we dont creat a new string resolved_fmt when none of the if is true ?

sashwinbalaji (approved):

PR #6082, opened 2026-06-01
perfetto: attribute TraceBufferV2 data loss by cause

sashwinbalaji:
When investigating data loss in the field, previous_packet_dropped only
told us *that* packets were dropped, never *why*. This widens the field
from a bool to a uint32 cause bitmask so a reported drop also carries
its cause.

Bit 0 (value 1) is the generic "dropped" flag and preserves the old
bool semantics: any nonzero value means dropped, and it stays what
TraceBufferV1 and producer-signalled drops emit. TraceBufferV2
additionally sets a bit identifying the cause: read gap, chunk
corrupted, orphan continuation, reassembly gap, reassembly broken
chain, overwrite, or trace-writer drop.


src/tracing/core/trace_writer_impl.cc
```diff
@@ -183,8 +183,10 @@ TraceWriterImpl::TracePacketHandle TraceWriterImpl::NewTracePacket() {
     if (PERFETTO_UNLIKELY(was_dropping_packets)) {
       // We've succeeded to get a new chunk from the SMB after we entered
       // drop_packets_ mode. Record a marker into the new packet to indicate the
-      // data loss.
-      cur_packet_->set_previous_packet_dropped(true);
+      // data loss. The producer can't attribute a cause, so it sets the bare
+      // DATA_LOSS_PRESENT bit.
+      cur_packet_->set_previous_packet_dropped(
```
primiano, line 189: the comment is not correct, there is only one reason why we get here, it is when we cannot find a chunk in the SMB; the smb is full. we should 100% have a dedicated flag for this, and set it here in an OR.

src/tracing/service/trace_buffer_v1.cc
```diff
@@ -666,7 +666,8 @@ bool TraceBufferV1::ReadNextTracePacket(
         if (PERFETTO_LIKELY(result == ReadPacketResult::kSucceeded)) {
           *sequence_properties = {trusted_producer_id, client_identity,
                                   writer_id};
-          *previous_packet_on_sequence_dropped = previous_packet_dropped;
+          *previous_packet_on_sequence_dropped =
+              previous_packet_dropped ? 1 : 0;
```
primiano: !!prev_pack_dropped (But also doens't matter)

primiano (approved):

PR #6083, opened 2026-06-01
ui: QueryPage: read sidebar-visible setting inline

gignat-dev:
oninit cached sidebarVisibleSetting.get() into a private field, so a change to the setting made on the Settings page wasn't reflected by an already-mounted QueryPage. 

Read it into a local in view() each render instead; it's backed by an in-memory cache, so the per-render read is cheap, and the toggle writes through set() without a shadow field.

### DIFF
--- a/ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
+++ b/ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
@@ -89,13 +89,6 @@ export interface QueryPageAttrs {
 }
 
 export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
-  // Whether the right-hand History/Tables sidebar is shown.
-  private sidebarVisible = true;
-
-  oninit({attrs}: m.CVnode<QueryPageAttrs>) {
-    this.sidebarVisible = attrs.sidebarVisibleSetting.get();
-  }
-
   // Map of tab ID to DataSource for each tab's query results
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
```diff
@@ -182,13 +174,15 @@ export class QueryPage implements m.ClassComponent<QueryPageAttrs> {
...
-          },
-          active: this.sidebarVisible,
+          icon: attrs.sidebarVisibleSetting.get()
+            ? 'right_panel_close'
+            : 'right_panel_open',
+          title: attrs.sidebarVisibleSetting.get()
+            ? 'Hide sidebar'
+            : 'Show sidebar',
+          onclick: () =>
+            attrs.sidebarVisibleSetting.set(!attrs.sidebarVisibleSetting.get()),
```
stevegolton: Maybe assign attrs.sidebarVisibleSetting.get() to a const further up in the view function, not for efficiency, more for clarity/DRY seeing the code is duplicated 5 times.

stevegolton (approved):

ui/src/plugins/dev.perfetto.QueryPage/query_page.ts
gignat-dev: Done

stevegolton (approved):

PR #6084, opened 2026-06-01
[State] Support state tracks in trace_processor [2/4]

etiennep-chromium:
Based on https://github.com/google/perfetto/blob/rfcs/0023-updatable-track-event.md

[2/4]: This PR adds supports for state track in trace_processor, to export the protos into new 'state' SQL table

For reference, this is the full implementation: https://github.com/google/perfetto/pull/6068


src/trace_processor/importers/common/track_compressor.h
```diff
@@ -178,53 +235,18 @@ class TrackCompressor {
...
-                blueprint.description_blueprint,
-            };
-          }
-        },
-        typename BT::dimension_blueprints_t());
+  // Wrapper around tracks::StateBlueprint which makes the blueprint eligible
+  // for compression with TrackCompressor. Please see documentation of
+  // tracks::StateBlueprint for usage.
+  template <typename NB = tracks::NameBlueprintT::Auto, typename... D>
+  static constexpr auto StateBlueprint(
```
LalitMaganti: State should not need compression the same way counters don't need compression. Pleasse revert.

src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -520,6 +527,17 @@ class TrackEventEventImporter {
                         : std::nullopt));
   }
 
+  base::StatusOr<TrackId> ParseTrackAssociationState() {
+    if (!track_uuid_ && fallback_to_legacy_pid_tid_tracks_) {
```
LalitMaganti, line 531: Shouldn't it be an error to set `fallback_to_legacy_pid_tid_tracks_` with state?

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
```diff
@@ -31,6 +31,26 @@ CREATE PERFETTO VIEW counter(
 AS
 SELECT id, ts, track_id, value, arg_set_id FROM __intrinsic_counter;
 
+-- Contains updatable state tracking slices.
```
LalitMaganti: Let's not use the word "slice" here, that's just going to confuse everyone. Also this needs signifcantly more description here.

src/trace_processor/perfetto_sql/stdlib/viz/summary/states.sql
```diff
@@ -0,0 +1,25 @@
+--
+-- Copyright 2026 The Android Open Source Project
```
LalitMaganti: This should be updated with UI changes not here.

src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -658,6 +677,36 @@ class TrackEventEventImporter {
     return base::OkStatus();
   }
 
+  base::Status ParseStateEvent() {
+    ASSIGN_OR_RETURN(auto track_id, ParseTrackAssociationState());
+
+    StringId state_id = kNullStringId;
+    bool is_termination = true;
```
LalitMaganti: Why do we need this? Why is this not just if state_id == null?

LalitMaganti (approved):

PR #6092, opened 2026-06-01
ui: Wattson: Display track name in pinned track

wusamuel6:
The pinned track was displaying the utid instead of the thread name, making it difficult to immediately understand the pinned track. This was due to race with Sched plugin, so make Wattson threads logic dependent on Sched plugin, such that Sched plugin loads first and the requisite data is present before Wattson threads logic is calculated.

Wattson per thread logic is dependent on sched events, so this is anyways the proper dependency that should be explicitly listed out.

Test: localhost:10000/#!/viewer?local_cache_key=<MY_TRACE>&org.kernel.Wattson:metrics=perfetto_wattson_apps_rails-name-NOTIFICATION_SHADE_EXPAND_COLLAPSE::Expand-cpu_subsystem-estimated_mw-mean
Bug: 518793028


stevegolton (commented):
I'm not sure I quite follow the logic behind this one? How do the plugins interact?

wusamuel6:
> I'm not sure I quite follow the logic behind this one? How do the plugins interact?

Ahh, I could have been more clear- specifically this is for the case of pinning Wattson tracks during trace startup.

And then without this patch, Wattson can run before SchedPlugin has run. If that happens:
1. Wattson cannot find the target thread's track node and puts the per thread power track at bottom of screen instead of grouped with that thread
2. The name of the track (`const name = threadNode?.name ?? (utid === 0 ? 'swapper' : `utid ${utid}`);` from `wattson_thread_utils.ts`) uses the fallback of `utid 123` instead of the actual name like `RenderThread`, which is not very intuitive for a user (see b/518793028).

stevegolton:
Thanks for the explanation!

stevegolton (approved):

PR #6093, opened 2026-06-01
tp: rename blocking calls module and optimize for CUJs

gopi3587:
Renames the standard library module to `android.blocking_calls_during_cujs` and the internal table to `_android_blocking_calls_during_cujs`.

Restructures the stdlib table using strict CROSS JOIN top-down indexing to scope queries directly to process IDs with active CUJs. This completely bypasses unindexed scans over the global slice table.

By traversing the join path from active processes down to individual track slices, the query optimizer can leverage the B-Tree indices on thread tracks and slice track IDs, providing massive reductions in slices evaluated and cutting query latency.

Furthermore, this locks the optimal execution plan in-place, permanently preventing the SQLite query planner from degenerating into unindexed nested-loop cartesian scans, which previously caused severe wall-clock regressions on developer workstations.

All baseline and optimized outputs remain character-for-character binary-identical (verified via host trace_processor_shell benchmark sweeps).

Test: manual verification via trace_processor_shell benchmark sweeps

Welcome to Perfetto!

For more details please see


LalitMaganti:
Sorry but this is *not* the correct fix at all. The module was designed very intentionally to separately compute the blocking calls, the cujs and then a layer on top to compose them. This is because you can reason about blocking calls outside of CUJs perfectly fine.

You are hardcoding a hack to some particular problem you're having rather than solving the problem more correctly in a general way.

gopi3587:
Ack, working on the cleanup without impacting the existing tables and their dependencies.

gopi3587 (commented):
Created a new blocking calls during CUJ's without impact the previous generic implementation.

src/trace_processor/perfetto_sql/stdlib/android/blocking_calls_during_cujs.sql
```diff
@@ -0,0 +1,67 @@
...
+  ON process.upid = ru.upid
+CROSS JOIN thread
+  ON thread.upid = process.upid
+CROSS JOIN thread_track
+  ON thread_track.utid = thread.utid
+CROSS JOIN slice
+  ON slice.track_id = thread_track.id
+WHERE
+  _is_relevant_blocking_call(slice.name, slice.depth)
+UNION ALL
```
nicomazz, line 92: is there a way to avoid duplicating this logic the union logic with binder calls as it happens also in critical_blocking_calls.sql?

src/trace_processor/perfetto_sql/stdlib/android/blocking_calls_during_cujs.sql
```diff
@@ -0,0 +1,67 @@
...
+  slice.ts + slice.dur AS ts_end
+FROM relevant_upids AS ru
+CROSS JOIN process
+  ON process.upid = ru.upid
+CROSS JOIN thread
+  ON thread.upid = process.upid
+CROSS JOIN thread_track
+  ON thread_track.utid = thread.utid
+CROSS JOIN slice
+  ON slice.track_id = thread_track.id
```
nicomazz: does this give any advantage vs using `thread_slice` directly?

src/trace_processor/perfetto_sql/stdlib/android/blocking_calls_during_cujs.sql
```diff
@@ -0,0 +1,67 @@
...
+
+INCLUDE PERFETTO MODULE android.critical_blocking_calls;
+
+-- OPTIMIZED: Include CUJs module to filter by relevant processes
+INCLUDE PERFETTO MODULE android.cujs.sysui_cujs;
+
+-- Extract critical blocking calls from processes that have CUJs.
+-- Materialized as a table, heavily optimized using CROSS JOIN to force index
+-- and avoid scanning the global row slice table.
+CREATE PERFETTO TABLE _android_blocking_calls_during_cujs AS
```
nicomazz, line 72: if the issue is that we have too many blocking calls from processes we don't care about, would it make sense to instead not get atrace slices from them?

gopi3587: Are you referring to capture trace information only on the processes that we are interested in like systemui, launcher, surface flinger etc? I am also thinking from the power metrics perspective that disabling this at the process level might impact other metrics or some debugging capability.

src/trace_processor/perfetto_sql/stdlib/android/blocking_calls_during_cujs.sql
gopi3587: Yes, I moved this to separate view and reusing it now in both the places.

Since this change was month old, I had lot of merge conflicts. Resolved everything in this new pull request https://github.com/google/perfetto/pull/6582

src/trace_processor/perfetto_sql/stdlib/android/blocking_calls_during_cujs.sql
gopi3587: CROSS JOIN  chain is the very core of why we were able to slash the execution time. If we simply wrote  FROM
  thread_slice JOIN relevant_upids , the SQLite query planner is often blind to the optimal execution path

src/trace_processor/perfetto_sql/stdlib/android/blocking_calls_during_cujs.sql
gopi3587: I was playing around with trace config and not had much luck with reducing the trace file size. I am also not entirely sure about the additional details used for debugging. I would leave this to your team to do some deep dive on what can be optimized in the trace config to reduce the trace file size.

gopi3587:
I will close this pull request since I created a new one in https://github.com/google/perfetto/pull/6582

PR #6100, opened 2026-06-01
Add heap analysis skill to find caching opportunities in java heap dumps

edgararriagag:
This heap analysis skill finds opportunities in heap dumps to save memory via object caching. It skews towards finding opportunities for objects that are heavily repeated, potentially duplicated, and objects that are shared by many objects.


skills/java-heap-caching-optimizer/SKILL.md
```diff
@@ -0,0 +1,66 @@
+---
```
LalitMaganti: Could we please follow the naming scheme and style of other skills rather than doing this ad-hoc. We're trying to have a consistent system here.

skills/java-heap-caching-optimizer/SKILL.md
```diff
@@ -0,0 +1,66 @@
...
+name: java-heap-optimizer
+description: >-
+  Find caching optimization opportunities in heap dumps
```
LalitMaganti: This is too short. This should comprehensively explain to the AI *why* they would want to use the skill. Right now this basically explains nothing.

Also I think the wya this is worded, it's not clear to me if we need a "meta-skill" which is all about "what can I do with a heap dump" which collates both this and what Tim built.

skills/java-heap-caching-optimizer/SKILL.md
```diff
@@ -0,0 +1,66 @@
...
+---
+
+# Heap Optimizer
+
+You are an expert android performance engineer that specializes in analyzing heap dumps and finding caching opportunities.
```
LalitMaganti: This is an anti-pattern with modern AI. It does not help improve perofrmance and often makes it worse.

skills/java-heap-caching-optimizer/SKILL.md
```diff
@@ -0,0 +1,66 @@
...
+
+## When to Use
+
+Use this skill when the user wants to analyze a perfetto java heap dump and find potential caching optimizations that would generate memory savings and reduce memory churn.
```
LalitMaganti: This should really be in the description.

skills/java-heap-caching-optimizer/SKILL.md
```diff
@@ -0,0 +1,66 @@
...
+
+Some other useful tables to query in heap dumps are:
+heap_graph_object: Information about individual objects in the heap dump.
+heap_graph_class: Information about classes in the heap dump.
+heap_graph_reference: Information about references between objects in the heap dump.
+
+4. The goal with the data will be to find potential caching opportunities, perform any other queries that you think might help you find them, if these are not sufficient and show them in the output. In order to execute queries you will need to use the trace processor tool which can be downloaded from the Perfetto website. Use following command to download it:
+
+```
+curl -LO https://get.perfetto.dev/trace_processor
```
LalitMaganti: This is orthogonal and should be delegated to the getting trace processor skill. Please don't mix skills like this, keep them properly architected and use delegation.

It's really the same as code: you want to properly architect into functions and modules etc: the same applies to skills.

edgararriagag: Done, I made it refer to the `perfetto_infra_querying_traces` skill to read the heap dump for querying

skills/java-heap-caching-optimizer/SKILL.md
edgararriagag: Done, moved to description

skills/java-heap-caching-optimizer/SKILL.md
edgararriagag: Ack, Replaced with a paragraph that focuses on what the skill teaches the agent and what to expect from it.

skills/java-heap-caching-optimizer/SKILL.md
edgararriagag: heap dump is a source of data, there could different types of analysis performed in a heap dump, likely each one pulling into different queries and pieces of data to be performed. In this case, we look specifically for optimizations that will likely reduce memory footprint by reducing duplicates and memory churn. This is in contrast with the other skill which mostly focuses on memory leaks which are the largest objects in a heap. That being said, a meta-skill that decides which heap dump optimization skill to call may be an option. Though, I am in favor of starting out with independent skills, then an improvement could come later in the form of that meta-skill that chooses one or several among the different optimization techniques

skills/java-heap-caching-optimizer/SKILL.md
edgararriagag: Done, ptal and let me know if I need to further tweak anything. I also noticed I had the wrong folder, so I also corrected this.

ai/skills/perfetto_workflow_android_heap_dump_caching_optimizer/scripts/query_heap_references.sql
```diff
@@ -0,0 +1,21 @@
+-- This query returns all objects that are referenced (owned) by objects of class <owner_classname>
```
LalitMaganti, line 16: Here and below: we need copyright headers for source code.

ai/skills/perfetto_workflow_android_heap_dump_caching_optimizer/SKILL.md
```diff
@@ -0,0 +1,71 @@
+---
+name: perfetto-workflow-android-heap-dump-caching-optimizer
+description: >-
+  Use when the user asks to analyze a perfetto java heap dump and find potential caching optimizations that would generate memory savings or reduce memory churn.
```
LalitMaganti: I'm worried as if stands this might be too specific. If I think from a user perspective, I think it would be very likely I would think to ask the question "I want tocache objects to save memory". It would almost certainly be "I want to improve the memory usage" and it would be up to the skill to define that. Caching is just one technique were recommending now but it's possible we will expand over time. 

Maybe let's focus on the problem rather than on the solution? Will also make it more future proof as well.

ai/skills/perfetto_workflow_android_heap_dump_caching_optimizer/scripts/query_heap_references.sql
edgararriagag: Done

ai/skills/perfetto_workflow_android_heap_dump_caching_optimizer/SKILL.md
edgararriagag: I reworded and tested it still triggers when asking for caching optimizations (PTAL and let me know if you think further rewording is necessary).

Moving forward, I am leaning towards potentially having some meta-skill (which I believe is kind of what you suggested, but certainly outside the scope of this CL, would need to be a subsequent CL either by me or someone on perfetto team) that serves as triggering point that maybe asks questions to the user about what kind of optimizations to consider and then decide whether to call the memory leak skill (perfetto-workflow-android-heap-dump) or to call this one (the caching technique skill). That is one potential way this could be handled to reduce the need for specificity but also avoid collision with other skills when asking the general question of optimizing memory the meta-skill (would need to be built) being the deciding factor on how to move next (maybe the AI prompts where to focus, memory leaks or memory saving opportunities outside leaks).

LalitMaganti: Yeah I think we should add a meta skill when the third skill which tries to help with heap dumps comes in. Fine for now.

LalitMaganti (approved):

PR #6104, opened 2026-06-02
ui: Default to the Heap Dump Explorer on more traces

zezeozue:
Treat traces with a small number of slices as not having timeline data to handle cases where there are trace triggers.

### DIFF
--- a/ui/src/plugins/dev.perfetto.HeapProfile/index.ts
+++ b/ui/src/plugins/dev.perfetto.HeapProfile/index.ts
@@ -385,9 +385,11 @@ function matchingTracks(
 }
 
 export async function traceHasTimelineData(ctx: Trace): Promise<boolean> {
+  // We treat a small number of slices as not having timeline data cos
+  // there are some inevitable slices like trace triggers on oom etc.
   const res = await ctx.engine.query(`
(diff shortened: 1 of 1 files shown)


LalitMaganti (approved):

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
```diff
@@ -385,9 +385,11 @@ function matchingTracks(
 }
 
 export async function traceHasTimelineData(ctx: Trace): Promise<boolean> {
+  // We treat a small number of slices as not having timeline data cos
+  // there are some inevitable slices like trace triggers on oom etc.
   const res = await ctx.engine.query(`
     SELECT
-      EXISTS(SELECT 1 FROM slice) OR
+      (SELECT count(id) FROM slice) > 50 OR
```
stevegolton, line 392: 50 is pretty arbitrary no?

zezeozue: Yea. Was out of ideas. Lalit suggested 50 so i used it and left a comment. I don't have a better basis but I think 50 should be fine in practice.

PR #6105, opened 2026-06-02
tp: move winscope out-of-tree as proto2 extensions

sashwinbalaji:
Move the whole winscope schema out of the core trace protos into protos/third_party/android/winscope, so the Android platform owns it. Winscope payloads are now extensions on TracePacket, and trace_packet.proto no longer mentions winscope at all.

- Wire format is unchanged: the extensions keep the original field numbers, so old traces open on the new trace processor and new traces open on old ones.
- The winscope protos now use the com.android.internal package, like the rest of third_party/android. Their generated C++ namespace changes from perfetto::protos to com::android::internal, as do their type names in the descriptor. Source-level only; nothing on the wire changes.
- The winscope_extensions field (112) is also gone from the public TracePacket; it's an out-of-tree extension now.
- The schema is split across two build targets because protos that use custom options can't be built with the C++ proto backend.
- Diff tests were updated to read the out-of-tree descriptor and reference winscope payloads as extensions.
- gen_all outputs (Android.bp, bazel, public C SDK headers) are not regenerated here and still need the usual gen_all run.


LalitMaganti (approved):

protos/third_party/android/frameworks/base/proto/tracing/winscope/view/remote_animation_target.proto
```diff
@@ -16,12 +16,13 @@
...
-import "protos/perfetto/trace/android/graphics/rect.proto";
-import "protos/perfetto/trace/android/view/surfacecontrol.proto";
+import "protos/third_party/android/frameworks/base/proto/tracing/winscope/app/window_configuration.proto";
+import "protos/third_party/android/frameworks/base/proto/tracing/winscope/graphics/point.proto";
+import "protos/third_party/android/frameworks/native/tracing/winscope/common/rect.proto";
+import "protos/third_party/android/frameworks/base/proto/tracing/winscope/view/surfacecontrol.proto";
 
-package perfetto.protos;
+package com.android.internal;
+option java_package = "com.android.internal.tracing.winscope.protos";
```
sashwinbalaji, line 25: hold off on merging, until android presubmit becomes stable.

PR #6107, opened 2026-06-02
Fix ordering of global tracks

etiennep-chromium:
Force global root tracks (global tracks that themselves have children) to be
sorted before any process groups in the Perfetto UI timeline.

The query currently calculates track event order_ids by
partitioning all root-level tracks together (regardless of whether they
are global or process-scoped). This inflates the order_id of global tracks.
Because process groups have a fixed sortOrder of 50 in the UI, these
global tracks incorrectly appear at the very end of the timeline.

This CL fixes the issue in the UI layer by identifying global root
tracks (tracks with no parent, upid, or utid) and forcing their
sortOrder to 0. This ensures they always sort before process groups
(sortOrder 50) without needing to modify the SQL standard library
partitioning logic.


LalitMaganti:
All global tracks go into a "global track events" group so this should not matter. I don't get what problem this actually is solving.

LalitMaganti (approved):

etiennep-chromium:
Discussed offline, this is specific to global tracks with children.
Updated the fix to updating JS code instead of SQL query.

PR #6109, opened 2026-06-03
ui: Avoid using screenshots to assert omnibox behavior

stevegolton:
The omnibox test is currently based on screenshots and is fragile because the screenshots can change any time a new command is introduced.

For instance - take this test failure in a PR that doesn't even tough the omnibox - simply adds a new command: https://storage.googleapis.com/perfetto-ci-artifacts/gh-26835408928-2-ui/ui-test-artifacts/index.html
https://github.com/google/perfetto/pull/5798

This patch modifies the test to assert correctness by asserting the state of the DOM instead of taking screenshots, which:
a) Asserts correct behavior directly (rather than purely comparing against a 'golden' screenshot, which itself could become wrong through careless rebases).
b) Is much less subject to false positives as the test surfaces is significantly smaller and more targeted.

### DIFF
--- a/ui/src/test/omnibox.test.ts
+++ b/ui/src/test/omnibox.test.ts
@@ -14,33 +14,28 @@
 
 import {test} from '@playwright/test';
 import {PerfettoTestHelper} from './perfetto_ui_test_helper';
+import {expect} from '@playwright/test';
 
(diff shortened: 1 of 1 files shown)


LalitMaganti (approved):

petrcermak:
Thank you for making the test more robust 🙂

PR #6111, opened 2026-06-03
[SDK] Defer track descriptor writes until first event

etiennep-chromium:
Modify `TrackEvent::SetTrackDescriptor` to check if the track has
been seen by the data source instance before writing.
The descriptor will be written dynamically when the
track emits its first event.
This prevents trace buffers in "light" sessions from
being polluted with descriptors of idle threads.

### DIFF
--- a/include/perfetto/tracing/internal/track_event_data_source.h
+++ b/include/perfetto/tracing/internal/track_event_data_source.h
@@ -590,6 +590,8 @@ class TrackEvent {
     TrackRegistry::Get()->UpdateTrack(track, desc.SerializeAsString());
     Trace([&](TrackEventDataSource::TraceContext ctx) {
       auto* incr_state = ctx.GetIncrementalState();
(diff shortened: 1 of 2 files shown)


include/perfetto/tracing/internal/track_event_data_source.h
```diff
@@ -589,8 +589,12 @@ class TrackEvent {
     PERFETTO_DCHECK(track.uuid == desc.uuid());
     TrackRegistry::Get()->UpdateTrack(track, desc.SerializeAsString());
     Trace([&](TrackEventDataSource::TraceContext ctx) {
+      auto* incr_state = ctx.GetIncrementalState();
```
sashwinbalaji, line 592: nit: pls add a comment so explains for future.

sashwinbalaji (commented):
LGTM (but would let primiano/lalit to approve as still new to sdk)

primiano (approved):

PR #6115, opened 2026-06-03
ui: Persist Heap Dump Explorer state in shared links

zezeozue:
The Heap Dump Explorer kept its state (selected dump, active tab, open flamegraph and instance drill-down tabs, and the main flamegraph's filter/pivot/view state) in a per-trace session object rebuilt from scratch on every load, so a shared permalink dropped all of it and always reopened on the overview.

Mirror the session state into a plugin store via ctx.mountStore(), which the core serializes into the permalink and restores before the plugin runs. On load the session rehydrates from the store (keyed on the heap dump still existing in the trace)


ui/src/plugins/com.android.HeapDumpExplorer/session.ts
```diff
@@ -244,16 +282,19 @@ export class HeapDumpExplorerSession {
 
   setActiveInstanceTab(id: number): void {
     this._activeInstanceId = id;
+    this.persist();
```
stevegolton: These calls to persist seem rather fragile.

In fact it seems like the state for this plugin is stored in multiple places and there's no obvious single source of truth:
1) Session runtime fields
2) In the persistent store (synched up with persist() being called on every mutation).
2) In _activeFlamegraphId/_activeInstanceId which are derived from the url every render cycle IIUC.
3) In the URL bar itself.

I think we can simplify this a lot and remove a lot of potential brittleness by moving the state into a single source of truth. The story gets complicated around the URL bar - we **should** persist the URL when sharing a permalink but we currently don't - so for now its fine to duplicate the state locally and in the URL and keep them in sync - though we should clean this up when we fix the URL.

Can you try moving the session stat from session.ts to the store and just rendering that every frame? Then the restoration should just work automatically after that too.

Re the `count` field in the flamegraph tab - I'm not entirely sure what this is used for but it seems like cached/derived state to me.. it shouldn't belong in the store but in some cached area (perhaps mithril component member state?).

stevegolton (changes requested):
Left a comment relating to state management, PTAL.

ui/src/plugins/com.android.HeapDumpExplorer/session.ts
zezeozue: Thanks for the feedback. Reworked to avoid the duplication of state and have the store as the source of truth. lmk what you think.

ui/src/plugins/com.android.HeapDumpExplorer/session.ts
```diff
@@ -92,50 +85,91 @@ export class HeapDumpExplorerSession {
...
+  // Loads the dumps and reconciles them with the stored active dump. Returns
+  // true if a valid permalink was restored, otherwise resets to the first dump.
+  async loadDumps(): Promise<boolean> {
     this._dumps = await queries.loadDumpsList(this.engine);
-    this._activeDump = this._dumps.length > 0 ? this._dumps[0] : null;
+    const ref = this.store.state.activeDump;
+    const restored =
+      ref !== undefined &&
+      this._dumps.some((d) => d.upid === ref.upid && d.ts === BigInt(ref.ts));
+    if (!restored) {
```
stevegolton: Suggestion: Flip the branches to use a positive condition

ui/src/plugins/com.android.HeapDumpExplorer/persisted_state.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const DUMP_REF_SCHEMA = z.object({
+  upid: z.number(),
+  ts: z.string(),
+});
+
+const FLAMEGRAPH_TAB_SCHEMA = z.object({
+  pathHashes: z.string(),
+  isDominator: z.boolean(),
+  upid: z.number(),
```
stevegolton: Do we need upid and ts here as well? thought HDE these tabs were reset after selecting a new dump? In other words, this is redundant information as we always use the dump defined in DUMP_REF_SCHEMA

stevegolton (approved):

PR #6123, opened 2026-06-03
watchdog: Support ro.hw_timeout_multiplier on Android

tutan-chromium:
Multiply fatal timer durations by the integer multiplier read from the system property `ro.hw_timeout_multiplier` when building for Android.

This keeps Perfetto thread and task-runner timeouts safe from triggering false-positive crashes (wdog_reason: 3) during prolonged suspend/resume cycles or on slow emulated environments.

I use direct `__system_property_get` to scale the interval `ms` inside `CreateFatalTimer`, ensuring that we avoid pulling heavy dependencies like libbase into Perfetto base. Non-Android platforms are unaffected.

Bug: 6122

### DIFF
--- a/src/base/watchdog_posix.cc
+++ b/src/base/watchdog_posix.cc
@@ -67,7 +67,8 @@ base::CrashKey g_crash_key_actual_cpu("wdog_actual_cpu");
 
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
 // This system property is already used for all sorts of Android watchdogs.
-// Bringing the Perfetto watchdog behavior closer to that on Android reduces surprise.
+// Bringing the Perfetto watchdog behavior closer to that on Android reduces
+// surprise.
 uint32_t GetHwTimeoutMultiplier() {
   static uint32_t multiplier = []() {
     char prop_val[PROP_VALUE_MAX] = {0};
--- a/src/base/watchdog_posix.cc
+++ b/src/base/watchdog_posix.cc
@@ -66,6 +66,8 @@ base::CrashKey g_crash_key_actual_boot("wdog_actual_boot");
 base::CrashKey g_crash_key_actual_cpu("wdog_actual_cpu");
 
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+// This system property is already used for all sorts of Android watchdogs.
+// Bringing the Perfetto watchdog behavior closer to that on Android reduces surprise.
 uint32_t GetHwTimeoutMultiplier() {
   static uint32_t multiplier = []() {
     char prop_val[PROP_VALUE_MAX] = {0};
--- a/src/base/watchdog_posix.cc
+++ b/src/base/watchdog_posix.cc
@@ -30,12 +30,13 @@
 #include <unistd.h>
 
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
-#include <sys/system_properties.h>
+#include "perfetto/ext/base/android_utils.h"
 #endif
 
 #include <algorithm>
 #include <cinttypes>
 #include <fstream>
+#include <limits>
 #include <thread>
 
 #include "perfetto/base/build_config.h"
@@ -71,10 +72,12 @@ base::CrashKey g_crash_key_actual_cpu("wdog_actual_cpu");
 // surprise.
 uint32_t GetHwTimeoutMultiplier() {
   static uint32_t multiplier = []() {
-    char prop_val[PROP_VALUE_MAX] = {0};
-    if (__system_property_get("ro.hw_timeout_multiplier", prop_val) > 0) {
-      int val = atoi(prop_val);
-      return val > 1 ? static_cast<uint32_t>(val) : 1u;
+    std::string multiplier_prop = GetAndroidProp("ro.hw_timeout_multiplier");
+    if (!multiplier_prop.empty()) {
+      auto multiplier_opt = StringToUInt32(multiplier_prop);
+      if (multiplier_opt.has_value() && *multiplier_opt > 0) {
+        return *multiplier_opt;
+      }
     }
     return 1u;
   }();
@@ -156,7 +159,15 @@ Watchdog::Timer Watchdog::CreateFatalTimer(uint32_t ms,
     return Watchdog::Timer(this, 0, crash_reason);
 
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
-  ms *= GetHwTimeoutMultiplier();
+  uint32_t multiplier = GetHwTimeoutMultiplier();
+  if (ms > 0 && multiplier > 1) {
+    uint64_t scaled_ms = static_cast<uint64_t>(ms) * multiplier;
+    if (scaled_ms > std::numeric_limits<uint32_t>::max()) {
+      ms = std::numeric_limits<uint32_t>::max();
+    } else {
+      ms = static_cast<uint32_t>(scaled_ms);
+    }
+  }
 #endif
 
   return Watchdog::Timer(this, ms, crash_reason);
--- a/src/base/watchdog_posix.cc
+++ b/src/base/watchdog_posix.cc
@@ -29,6 +29,10 @@
 #include <sys/timerfd.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


src/base/watchdog_posix.cc
```diff
@@ -61,6 +65,23 @@ base::CrashKey g_crash_key_actual_mono("wdog_actual_mono");
...
 base::CrashKey g_crash_key_actual_cpu("wdog_actual_cpu");
 
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+// This system property is already used for all sorts of Android watchdogs.
+// Bringing the Perfetto watchdog behavior closer to that on Android reduces
+// surprise.
+uint32_t GetHwTimeoutMultiplier() {
+  static uint32_t multiplier = []() {
+    char prop_val[PROP_VALUE_MAX] = {0};
+    if (__system_property_get("ro.hw_timeout_multiplier", prop_val) > 0) {
```
sashwinbalaji: pls use the [GetAndroidProp](https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/src/base/android_utils.cc;bpv=1;bpt=1;l=49?q=__system_property_get&ss=h%2Fgoogleplex-android%2Fplatform%2Fsuperproject%2Fmain%2F%2B%2Frefs%2Fheads%2Fmain:external%2Fperfetto%2F&gsn=GetAndroidProp&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fgoogleplex-android.googlesource.com%2Fplatform%2Fsuperproject%2Fmain%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dexternal%2Fperfetto%2Fsrc%2Fbase%2Fandroid_utils.cc%23tmXTF6xeg03CaM7B91k5TKA_IYXiAc7if3w_CVcedzc)

src/base/watchdog_posix.cc
```diff
@@ -61,6 +65,23 @@ base::CrashKey g_crash_key_actual_mono("wdog_actual_mono");
...
+      int val = atoi(prop_val);
```
sashwinbalaji: [StringToUInt32](https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/include/perfetto/ext/base/string_utils.h;bpv=1;bpt=1;l=90?q=StringToUInt32&ss=h%2Fgoogleplex-android%2Fplatform%2Fsuperproject%2Fmain%2F%2B%2Frefs%2Fheads%2Fmain:external%2Fperfetto%2F&gsn=StringToUInt32&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fgoogleplex-android.googlesource.com%2Fplatform%2Fsuperproject%2Fmain%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dexternal%2Fperfetto%2Finclude%2Fperfetto%2Fext%2Fbase%2Fstring_utils.h%23f6bUGCP-Te3q703i5Fra14M8IKie6Jp7zzhRjIp-f0I) here pls

src/base/watchdog_posix.cc
```diff
@@ -61,6 +65,23 @@ base::CrashKey g_crash_key_actual_mono("wdog_actual_mono");
...
+      return val > 1 ? static_cast<uint32_t>(val) : 1u;
```
sashwinbalaji: nit/slight worry: AFAICT by code searching, the value of multiplier is usually small. But just to be defensive shall we clamp it ? Because if some erroneous value (should not happen), the below multiplication can overflow as timeout is also high (5min = 300000).

Also just curious on what are the possible values for your device ?

sashwinbalaji (commented):
Thanks for the PR!

FYI, we are also investigating watchdog issue on android (http://b/496506220). And to summarise the current status:
  - Added a trigger that captures a trace when traced_probes disconnects (#5299).
  - @primiano made the watchdog log the actual mono/boot/cpu clocks at crash time (#5749), and gave the process ~60s to flush the trace before crashing instead of dying immediately (#5755), so the captured traces are more useful.
  - The traces we got back were all kTraceDidntStop (wdog reason 4).
  - And on looking into them, the system was just busy immediately after boot (heavy ftrace load) and the timeout we compute was just too short. mono == boot in every case, so most likely suspend wasn't involved.
  - So we raised a PR to floor that timeout at 5 min (#6045).
  - This was done last week, waiting for more traces to investigate the watchdog:3 issue.

Happy to know if you any other insight.

src/base/watchdog_posix.cc
filmil: It's set to 500, for somewhat related reasons.

src/base/watchdog_posix.cc
tutan-chromium: Done

src/base/watchdog_posix.cc
tutan-chromium: Done

src/base/watchdog_posix.cc
tutan-chromium: As Filip said, we use 500 for our relatively slow HW. And I added uint32_t overflow check.

tutan-chromium:
> Thanks for the PR!
> 
> FYI, we are also investigating watchdog issue on android (http://b/496506220). And to summarise the current status:
> Happy to know if you any other insight.

Thank you so much for the information! In our case, we are likely dealing with an edge case on a relatively slow SoC with potential instabilities in suspend/resume state transitions and monotonic clock drifts (see b/460841705). I reviewed the changes you mentioned, and they will definitely be very useful for us as well. Thanks again!

src/base/watchdog_posix.cc
```diff
@@ -29,9 +29,14 @@
 #include <sys/timerfd.h>
 #include <unistd.h>
 
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
```
sashwinbalaji, line 32: nit: pls place it below along with other perfetto imports and after the stdlib ones.

src/base/watchdog_posix.cc
```diff
@@ -134,6 +158,18 @@ Watchdog::Timer Watchdog::CreateFatalTimer(uint32_t ms,
...
+  uint32_t multiplier = GetHwTimeoutMultiplier();
+  if (ms > 0 && multiplier > 1) {
+    uint64_t scaled_ms = static_cast<uint64_t>(ms) * multiplier;
+    if (scaled_ms > std::numeric_limits<uint32_t>::max()) {
+      ms = std::numeric_limits<uint32_t>::max();
+    } else {
+      ms = static_cast<uint32_t>(scaled_ms);
+    }
```
sashwinbalaji, line 169: nit optional: to redue the number of lines

```
    ms = static_cast<uint32_t>(
        std::min<uint64_t>(scaled_ms, std::numeric_limits<uint32_t>::max()));
```

sashwinbalaji (approved):

primiano (changes requested):
I am sorry folks but this is adding too much complexity to some piece of code that is extremely load bearing.
The watchdog has already very generous timeouts .
Multiplying them by 500 becomes ridiculous, at that point just disable that.

I suggest you either disable the wathdog, or do whatever multiplication you want, in the fuchsia downstream fork of perfetto.
I'm not keen of buying this surprise factor and complexity, as this change now requires having to think about, and taking the risk, of what happens selinux-wise to all the pieces of code all over the place that hit this code.

tutan-chromium:
Hi @primiano,

Thank you for the feedback. I completely understand your concerns about complexity, safety, and the "surprise factor" of adding this to core watchdog logic.

Regarding selinux related changes - The property lookup is implemented using `GetAndroidProp("ro.hw_timeout_multiplier")` and if SELinux blocks reading this property in restricted client contexts, the call fails gracefully by returning an empty string.

The `ro.hw_timeout_multiplier` is a standard system property in Android since v12  specifically to scale timeouts in slow environments. Currently, because Perfetto does not respect it, its watchdogs frequently trigger false-positive crashes (e.g., Reason 3 or 4 watchdog timeouts), while the rest of the OS successfully scales its timeouts and passes.

We agree that 500x is an extreme case. This value is a temporary configuration used in our downstream. But this just our specific case and we're planning to optimize this configuration to reduce it, but the property-driven scaling will still be >1.  Disabling the watchdog entirely  completely defeats the purpose of detecting genuine deadlocks/hangs which we also want to detect.

Thank you.

PR #6124, opened 2026-06-03
Update android blocking calls metrics and tests to add more RenderThread slices

XSJoJo:
See go/jank-metrics-2025

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
@@ -47,6 +47,12 @@ SELECT
   OR $name GLOB 'Compose:*'
   OR $name GLOB 'draw-VRI*'
   OR $name = 'CreateGraphicsPipeline'
+  OR $name GLOB 'drawLayer *'
+  OR $name GLOB 'DrawFrames*'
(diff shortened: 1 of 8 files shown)


src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
```diff
@@ -60,7 +66,11 @@ SELECT
 --Extract critical blocking calls from all processes.
 CREATE PERFETTO TABLE _android_critical_blocking_calls AS
 SELECT
-  android_standardize_slice_name(s.name) AS name,
+  CASE
+    WHEN s.name GLOB 'drawLayer *' THEN SUBSTR(s.name, 1, INSTR(s.name, ']'))
```
nicomazz: should this be inside android_standardize_slice_name as well? could you also add a small comment with how the original slice name would look like?

nicomazz (approved):

src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
XSJoJo: done

LalitMaganti (approved):

PR #6125, opened 2026-06-04
feat(trace_processor): parse legacy_sort_index for thread/process ordering

g122622:
## Summary
  Add support for parsing `legacy_sort_index` from `ThreadDescriptor` and `ProcessDescriptor` protobuf messages.

  ## Problem
  The `legacy_sort_index` field exists in the protobuf definitions but was not being parsed by the trace processor. This meant that SDK users setting `legacy_sort_index` on their TrackDescriptors would not see threads/processes ordered correctly in the UI.

  ## Solution
  Parse the `legacy_sort_index` field and store it as `thread_sort_index_hint` / `process_sort_index_hint` args, which the UI already uses for ordering (supports JSON format via `thread_sort_index` metadata).

  ## Testing
  Added diff tests for both thread and process `legacy_sort_index` parsing.

  ## Usage
  SDK users can now control thread display order:
  ```cpp
  auto desc = perfetto::ThreadTrack::Current().Serialize();
  desc.mutable_thread()->set_thread_name("MainThread");
  desc.mutable_thread()->set_legacy_sort_index(10);
  perfetto::TrackEvent::SetTrackDescriptor(perfetto::ThreadTrack::Current(), desc);
```
  Note: The UI sorts by thread_sort_index_hint in ascending order (lower values appear first).

### DIFF
--- a/src/trace_processor/importers/proto/track_event_parser.cc
+++ b/src/trace_processor/importers/proto/track_event_parser.cc
@@ -228,6 +228,10 @@ TrackEventParser::TrackEventParser(TraceProcessorContext* context,
       callsite_id_key_id_(context_->storage->InternString("callsite_id")),
       end_callsite_id_key_id_(
           context_->storage->InternString("end_callsite_id")),
+      thread_sort_index_hint_id_(
+          context_->storage->InternString("thread_sort_index_hint")),
+      process_sort_index_hint_id_(
+          context_->storage->InternString("process_sort_index_hint")),
       chrome_string_lookup_(context->storage.get()),
       active_chrome_processes_tracker_(context) {
   // Opt into DebugAnnotation handling: ParseMessage routes DebugAnnotation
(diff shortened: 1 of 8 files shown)


LalitMaganti:
Thanks for sending the PR but unfortunately we cannot accept this as it stands.

This is because:
1) I don't want to add a new "legacy" feature, if we're going to do it, I would like to build it as a new first class feature 
2) we need to think carefully about merged traces where there's both a chrome and JSON trace and they set inconsistent values - which one do we pick? This needs careful thought. 
3) we need to reason about how this sorting system interacts with the already existing sorting system for track events. 

All of these are complicated enough that if we were to send something here, I'd like the core team to do it.

Please feel free to open an issue with the problem you're facing instead and we can take a look. Having the motivation for this would help design this best.

LalitMaganti:
PTAL at https://github.com/google/perfetto/pull/6219 - it should fix the issue raised by this PR in a more sustainable way!

PR #6137, opened 2026-06-04
ftrace: Add `psci_domain_idle_{exit,enter}` ftrace events

staticfloat:
These events provide finer-grained information than just `cpu_idle` when attempting to see which power states are being entered.  This patch does not interpret these events at all, it merely allows them to be listed in the `ftrace events` log without needing to enable generic ftrace events.

X-ref: https://lkml.iu.edu/2501.2/03423.html


LalitMaganti:
If you're not parsing these events, there's basically no benefit to having these in the Perfetto repo. Since mid-2025, the cost of generic events is ~basically zero over proper typed protos.

staticfloat:
I was basing this off of this warning in the web UI:

image

Is that no longer true?  Is it valid to enable that by default?

That being said, with this patch I see the `psci_domain_idle_*` events in the `.systrace` export, but I don't see them in the `ftrace events` log view, which is what I was hoping for with this.  Do I need to do something more to cause those to be visible?

LalitMaganti:
> Is that no longer true? Is it valid to enable that by default?

Yes I think we should signifcantly soften the language there - as of mid-2025, there's a slight regresison in trace size but interms of performane it's basically on par with descriptors compiled in.

> Do I need to do something more to cause those to be visible?

Yes you need to rebuild the UI as well. This patch doesn't just affect how the events are collected but also how they are parsed in TP/UI.

LalitMaganti:
FWIW I'm not opposed to this, it'll make it easier if we do decide to parse this event in the future.

LalitMaganti (approved):

safayat-google (approved):

staticfloat:
FYI I did rebuild the UI and verified that the events do indeed then appear in the web UI.

PR #6143, opened 2026-06-04
tp: android.display.video data source, importer, table and stats

zezeozue:
Adds trace_processor support for the android.display.video data source, which captures device display frames as a hardware video elementary stream.

Wire format (protos/perfetto/trace/android/video_frame.proto):
  * VideoFrame (TracePacket.video_frame): a codec_config-only frame per stream (carrying display_name + RFC 6381 codec_string), then one VideoFrame per on-screen change with a single access unit (au_data). Self-identifying via a per-session display_id.
  * VideoFrameError (TracePacket.video_frame_error): per-stream failure with a Reason enum (size cap hit, codec error, display gone, no encoder, display not found, encoder setup failed, virtual display failed). Config: DataSourceConfig.display_video_config.

Importer (src/trace_processor/plugins/video_frame_importer/):
  * Parses video_frame into the __intrinsic_video_frames table. The encoded payload is stored zero-copy as a TraceBlobView, surfaced via the __INTRINSIC_VIDEO_FRAME_AU_DATA(id) scalar function (same pattern HPROF/ETM use for blob payloads).
  * Maps each VideoFrameError reason to its own kIndexed entry in the global stats table, keyed by display_id.
  * Enforces a parse-time 256 MB per-stream cap as a backstop against traces recorded without a producer cap: over-cap frames are dropped and android_video_size_cap_hit is reported once.

The table is intrinsic (not a stdlib module) since only the VideoFrames UI consumes it.

Tests: diff tests (single/multi display, au_data blob equivalence, all VideoFrameError reasons) and a unit test for the parse-time cap.


protos/perfetto/trace/android/video_frame.proto
```diff
@@ -0,0 +1,85 @@
...
+syntax = "proto2";
+
+package perfetto.protos;
+
+// A captured frame from a device display.
+//
+// Each display is captured as a hardware video elementary stream: a
+// codec_config-only VideoFrame first, then one VideoFrame per on-screen
+// change carrying a single access unit (au_data).
+message VideoFrame {
```
LalitMaganti: If this is Android specific, could we put it in a TracePacket extension in Android and copy it over here in third_party instead?

protos/perfetto/config/android/display_video_config.proto
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Config for the "android.display.video" data source. Unset/0 fields use
+// the producer's default.
+message DisplayVideoConfig {
```
LalitMaganti, line 23: Same here. I'm not sure if TraceConfig supports extensions yet but we should start thinking about it.

src/trace_processor/tables/android_tables.py
```diff
@@ -368,6 +368,49 @@
     ),
 )
 
+ANDROID_VIDEO_FRAMES_TABLE = Table(
+    python_module=__file__,
```
LalitMaganti: You can put this in the plugins folder.

src/trace_processor/plugins/video_frame_importer/video_frame_module.cc
```diff
@@ -0,0 +1,186 @@
...
+#include "protos/perfetto/trace/trace_packet.pbzero.h"
+
+namespace perfetto::trace_processor {
+
+using protos::pbzero::TracePacket;
+using protos::pbzero::VideoFrame;
+using protos::pbzero::VideoFrameError;
+
+namespace {
+
```
LalitMaganti: Why do you need to do this? ConstChars gas a function to convert to string view just use that.

src/trace_processor/plugins/video_frame_importer/video_frame_module.cc
```diff
@@ -0,0 +1,186 @@
...
+    return;
+  }
+
+  // Parse-time per-stream cap: drop frames once a stream exceeds it,
+  // reporting once via the same stat as the producer's cap.
+  if (info.emitted_bytes + static_cast<int64_t>(payload.size) >
+      max_stream_size_bytes_) {
+    if (!info.size_cap_hit) {
+      info.size_cap_hit = true;
+      context_->stats_tracker->IncrementIndexedStats(
```
LalitMaganti: This should use import logs not just stats.

LalitMaganti (approved):

PR #6156, opened 2026-06-05
ai: refactor multiple skills -> single skill

LalitMaganti:
This CL changes the approach of ahving N skills, each of which is
doing a specific thing with Perfetto to instead be a single skill
which internally has a tree of references/workflows guiding the
agent on how to work with traces.

The reason for this change is that we've found it's quite painful
for folks to juggle N skills and it's easy to accidentally only
install partial set of skills. This is problematic as then the
agent can get very confused and go in circles.

If we instead have everything packaged in a single skill and
then give strong guidance on how to "route" itself, it makes it a
lot easier. There's also nice advantages when it comes to
things like sharing common scripts and fragments - we can reference
other files using relative paths in a way which would be very
difficult if we were instead using N skills and couldn't count on
all skills necessarily being installed.


LalitMaganti:
@YasserDbeis FYI: we've changed approach slightly as we've gained experience in how these skills work internally in G3. While having N skills does work, you get into annoying situations where people might not actually get all the skills they need installed correctly and it also makes it fragile to refactor these things. Abstracting the routing complexity internally behind a single skill makes it a lot easier.

I would suggest if you want to build things on top, you take the contents of the individual workflows/references and you build your own top level skill router on top.

ai/skills/perfetto/environment-references/setup-standalone.md
```diff
@@ -122,4 +108,4 @@ kill $SERVER_PID
 ```
 
 If this prints a non-zero count and exits cleanly, the user is ready to
-follow the `querying-traces` skill.
+follow `../infra-references/querying.md`.
```
sashwinbalaji, line 111: should we give absolute paths ? Most likely the model will understand but for clarity.

sashwinbalaji (approved):

ai/skills/perfetto/environment-references/setup-standalone.md
LalitMaganti: No very much not because depending on the envirnonemtn the absolute path will be very very different.

PR #6158, opened 2026-06-05
tp: Add TrackEvent plugin registry + android_track_event_process table

zezeozue:
TrackEventParser gains a field-id-keyed plugin registry.

The frameworks/base plugin feeds the process table from process lifecycle events -- start sets name/uid via GetOrCreateProcess, binder-died EndThread()s the pid -- and keeps the framework start/end timestamps in the new __intrinsic_android_track_event_process table (upid, fw_start_ts, fw_end_ts).

Drops the aliased IMPORTANCE_BACKGROUND enumerator, which pbzero's switch-based ToString rejects under -Werror. This field is deprecated in favor of importance_cached anyways.


src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -1297,6 +1297,7 @@ class TrackEventEventImporter {
         parser_->AddActiveProcess(ts_, *it);
       }
     }
+
```
LalitMaganti, line 1300: Nit: revert

src/trace_processor/importers/proto/track_event_parser.cc
```diff
@@ -455,6 +457,8 @@ void TrackEventParser::ParseTrackEvent(int64_t ts,
         stats::track_event_dropped_packets_outside_of_range_of_interest);
     return;
   }
+  plugins_->ParseFields(blob, ts);
```
LalitMaganti, line 460: Shouldn't this return something? This feels very very as hoc here. We need to design this a bit better with various sets of callbacks: 
1) there's a layer here if you want to intercept the whole event and do parsing from scratch
2) then there's another layee which lives purely at the args level 

My theory is that you only want 2) right now in which case you should do this in trackeventeventimporter

src/trace_processor/importers/proto/track_event_plugin.h
```diff
@@ -0,0 +1,71 @@
...
+#include "perfetto/protozero/proto_decoder.h"
+
+namespace perfetto::trace_processor {
+
+// Parses an extension field nested inside a TrackEvent. Registered against one
+// or more TrackEvent field ids; ParseField runs for each matching field.
+class TrackEventPlugin {
+ public:
+  virtual ~TrackEventPlugin();
+  virtual void ParseField(uint32_t field_id,
```
LalitMaganti, line 36: This is too opaque. What is allowed, what is not allowed here? This need comments. I would also scope this down and say this is only about args associated with an event. So you can pass the slice id or counter I'd you've inserted.

zezeozue:
Will raise a new PR based on Lalit's new track event API

PR #6160, opened 2026-06-05
ui: chore: Bump deps to patch security issues

stevegolton:
This patch fixes the following security issues:
1. markdown-it - https://github.com/google/perfetto/security/dependabot/312
2. protobufjs - https://github.com/google/perfetto/security/dependabot/308

Note: These are not a major concern for us as this code for the most part runs client side, but worth keeping on top of this issues.

## List of packages updated

| Package | Previous Version | New Version | Type |
| :--- | :--- | :--- | :--- |
| `typescript` | `5.5.2` | `5.9.3` | devDependencies |
| `eslint` | `^9.6.0` | `^9.39.4` | devDependencies |
| `@typescript-eslint/eslint-plugin` | `^7.14.1` | `^8.60.1` | devDependencies |
| `@typescript-eslint/parser` | `^7.14.1` | `^8.60.1` | devDependencies |
| `protobufjs` | `^7.5.6` | `^7.6.2` | dependencies |
| `protobufjs-cli` | `1.2.1` | `^1.3.2` | devDependencies |
| `markdown-it` | `^13.0.0` | `^14.2.0` | dependencies |
| `@types/markdown-it` | `^13.0.0` | `^14.1.2` | devDependencies |
| `@types/node` | `^20.14.9` | `^20.19.41` | dependencies |
| `@types/w3c-web-usb` | `^1.0.10` | `^1.0.14` | dependencies |

## Updating markdown-it

Updating this package kicked off a cascade of issues.

- markdown-it@14 requires a newer typescript version than what we currently are using (5.5.2). For the sake of completeness this patch bumps typescript to 5.9.3 which is the latest version within the 5 major revision (moving to 6.X is another problem).
- Typescript 5.7 introduced a breaking change in TypedArrays and friends (such as DataView), making these generic over the type of ArrayBuffer they use - ArrayBuffer or SharedArrayBuffer. [TS Release Notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#typedarrays-are-now-generic-over-arraybufferlike). IIUC, this is to support ES2024 changes to the API of ArrayBuffer.
- A lot of libraries have been very slow to update their types to be more specific about what type of array they output, simply leaving it has Uint8Array or similar, or to allow ArrayBufferLike as inputs.
- This caused a cascade of type issues through the codebase which needed to be fixed. Mostly mechanical aside from one sticking point:
  - protobuf.Write.finish() returns a Uint8Array, the storage type is left purposely undefined. In reality it's always going to be an ArrayBuffer type unless patched but the types have not been tweaked to address this.
  - This causes type issues as we try to write this into a Blob as a BlobPart which the types specify as having a requirement on ArrayBuffer (not shared). This is bogus, all environments would accept a SharedArrayBuffer anyway.
- Tangentially related - a Uint8Array cannot be implicity cast to an ArrayBuffer any more - meaning the special cast in AppImpl is no longer required.
  - Argument of type 'Uint8Array<ArrayBuffer>' is not assignable to parameter of type 'ArrayBuffer'.
- Also what's changed is that calling map.keys().next().value now returns T | undefined, but before it returned T. See [TS release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-6.html#strict-builtin-iterator-checks-and---strictbuiltiniteratorreturn).


ui/src/core/metatracing.ts
```diff
@@ -121,9 +121,9 @@ function readMetatrace(): Uint8Array {
       .uint32(TRACE_PACKET_METATRACE_TAG)
       .bytes(protos.PerfettoMetatrace.encode(metatraceEvent).finish());
     wri.ldelim();
-    return wri.finish();
+    return wri.finish() as Uint8Array<ArrayBuffer>;
```
LalitMaganti: Are you sure this cast is safe?

ui/src/frontend/trace_actions.ts
```diff
@@ -138,8 +138,10 @@ async function finaliseMetatrace(engine: Engine) {
     throw new Error(`Failed to read metatrace: ${result.error}`);
   }
 
+  const metaTraceAsArrayBuffer = result.metatrace as Uint8Array<ArrayBuffer>;
```
LalitMaganti: As above

ui/src/plugins/dev.perfetto.RecordTraceV2/adb/webusb/adb_webusb_device.ts
```diff
@@ -324,12 +324,12 @@ export class AdbWebusbDevice extends AdbDevice {
...
     usb: AdbUsbInterface,
     len: number,
-  ): Promise<DataView> {
+  ): Promise<DataView<ArrayBuffer>> {
     const res = await usb.dev.transferIn(usb.rx, len);
     if (!exists(res.data) || res.status !== 'ok') {
       throw new WebusbTransportError(`res: ${res.status}, data: ${!!res.data}`);
     }
-    return res.data;
+    return res.data as DataView<ArrayBuffer>;
```
LalitMaganti: As above

LalitMaganti (approved):

ui/src/core/metatracing.ts
stevegolton: It is actually perfectly safe as protobufjs will always output an ArrayBuffer unless modified. However, to avoid the type assertion, I've updated this to a proper assert function which ensures the underlying buffer is definitely not shared or crash.

ui/src/frontend/trace_actions.ts
stevegolton: Fix is the same as above.

ui/src/plugins/dev.perfetto.RecordTraceV2/adb/webusb/adb_webusb_device.ts
stevegolton: I've actually managed to leave this type open and just change the interfaces of where it's used to allow the union type, which means we can omit the assertion.

PR #6167, opened 2026-06-08
heapprofd: add glob support to cmdline matching (only for running targets)

rsavitski:
Currently, heap profiling is the only data source that cannot be
configured with wildcard patterns for process names. This is because
"from startup" profiling on Android requires self-targeting code to
derive a unique sysprop name from the process cmdline, where the sysprop
value states whether to self-activate the profiling (and is set by
heapprofd). With wildcard matching, the startup case would need a
different scheme that can inspect all N patterns that are being
profiled.

This patch retrofits wildcard support into heapprofd, but makes it
explicitly incompatible with startup profiling to reduce surprise
factor. A config needs to explicitly state no_startup=true if its
|process_cmdline| includes wildcards.

The glob-aware matching codepath is identical to traced_perf and
java_hprof profilers.

Bug: 495877833

### DIFF
--- a/protos/perfetto/config/profiling/heapprofd_config.proto
+++ b/protos/perfetto/config/profiling/heapprofd_config.proto
@@ -59,13 +59,26 @@ message HeapprofdConfig {
   // this value.
   optional uint64 adaptive_sampling_max_sampling_interval_bytes = 25;
 
-  // E.g. surfaceflinger, com.android.phone
+  // Names of process cmdlines to profile. The semantics before perfetto v57 are
+  // as the following section (A), and on newer versions are both (A) and (B):
+  //
+  // (A) E.g. surfaceflinger, com.android.phone
   // This input is normalized in the following way: if it contains slashes,
   // everything up to the last slash is discarded. If it contains "@",
   // everything after the first @ is discared.
   // E.g. /system/bin/surfaceflinger@1.0 normalizes to surfaceflinger.
   // This transformation is also applied to the processes' command lines when
   // matching.
+  //
+  // (B) Perfetto v57+: a pattern may also contain a wildcard ('*'), in which
+  // case it is matched against running processes' command lines using glob
+  // semantics. If the pattern starts with '/' it is matched against the full
+  // argv0 (e.g. "/system/bin/*"), otherwise it is matched against the binary
+  // name part of argv0 (e.g. "system_*"). Wildcard patterns are *not*
+  // normalized as described above.
+  // Wildcard patterns are supported only for profiling already-running
+  // processes. A config must set no_startup=true if using wildcards, or it will
+  // be rejected.
   repeated string process_cmdline = 2;
 
   // For watermark based triggering or local debugging.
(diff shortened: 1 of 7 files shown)


src/profiling/memory/heapprofd_producer.h
```diff
@@ -321,7 +323,7 @@ class HeapprofdProducer : public Producer, public UnwindingWorker::Delegate {
   std::map<DataSourceInstanceID, DataSource> data_sources_;
 
   // Specific to mode_ == kChild
-  Process target_process_{base::kInvalidPid, ""};
+  Process target_process_{base::kInvalidPid, "", ""};
```
sashwinbalaji: nit: inline argument comment

src/profiling/memory/heapprofd_producer.cc
```diff
@@ -830,9 +863,13 @@ void HeapprofdProducer::SocketDelegate::OnNewIncomingConnection(
...
   Process peer_process;
   peer_process.pid = new_connection->peer_pid_linux();
-  if (!GetCmdlineForPID(peer_process.pid, &peer_process.cmdline))
+  if (!GetCmdlineForPID(peer_process.pid, &peer_process.cmdline)) {
     PERFETTO_PLOG("Failed to get cmdline for %d", peer_process.pid);
-
+  }
+  if (!glob_aware::ReadProcCmdlineForPID(peer_process.pid,
+                                         &peer_process.raw_cmdline)) {
+    PERFETTO_PLOG("Failed to get cmdline for %d", peer_process.pid);
```
sashwinbalaji, line 871: change the error log to distinghuish the error from above

protos/perfetto/config/profiling/heapprofd_config.proto
```diff
@@ -59,13 +59,26 @@ message HeapprofdConfig {
   // this value.
   optional uint64 adaptive_sampling_max_sampling_interval_bytes = 25;
 
-  // E.g. surfaceflinger, com.android.phone
+  // Names of process cmdlines to profile. The semantics before perfetto v57 are
```
sashwinbalaji, line 62: add it to changelog also pls

src/profiling/memory/heapprofd_producer.cc
```diff
@@ -830,9 +863,13 @@ void HeapprofdProducer::SocketDelegate::OnNewIncomingConnection(
     std::unique_ptr<base::UnixSocket> new_connection) {
   Process peer_process;
   peer_process.pid = new_connection->peer_pid_linux();
-  if (!GetCmdlineForPID(peer_process.pid, &peer_process.cmdline))
+  if (!GetCmdlineForPID(peer_process.pid, &peer_process.cmdline)) {
     PERFETTO_PLOG("Failed to get cmdline for %d", peer_process.pid);
-
+  }
+  if (!glob_aware::ReadProcCmdlineForPID(peer_process.pid,
```
sashwinbalaji: can we gate this behind if any glob pattern exists to avoid the /proc read

rsavitski: This is for incoming connections from processes that loaded the client library (e.g. by being signaled by heapprofd).
So this code doesn't yet know which DS this belongs to (in fact, it'll do the matching in HandleClientConnection()->GetDataSourceForProcess() as called below.

But a separate consequence is that this is only called once per process that's already been determined to be in scope for profiling, so the overheads are irrelevant.

PS: The O(nr_processes) loop in SignalRunningProcesses does "are there any patterns of this type?" check before reading from procfs.

src/profiling/memory/heapprofd_producer.cc
rsavitski: Folded into one PLOG.

sashwinbalaji (approved):

PR #6171, opened 2026-06-08
ai: remove eval files

LalitMaganti:
These aren't doing anything in our tree so just remove them


sashwinbalaji:
Can't we add some ci for these as they will keep growing ?

LalitMaganti:
Yes but we need to see how things settle internally to see if we need to write our own. There's the whole problem with getting the API key as well etc which is a whole can of worms I don't want to worry about in OSS today.

sashwinbalaji (approved):

PR #6175, opened 2026-06-09
tp: Parse and ingest ART OOME metadata into Trace Processor

BogdanLionte:
Adds support for extracting ART OOME crash details from the trace. When an OOME occurs, the ART VM emits metadata about the failed allocation and the thread stack trace where the allocation occurred.

This patch:
- Adds the `heap_graph` table to capture the list of heap dumps along with their timestamps, process PIDs, and dump reasons.
- Adds the `heap_graph_java_oome_details` table to capture the allocation size, free memory, and the VM error message.
- Adds the `heap_graph_thread_callsite` table to link the OOME heap graph to the thread ID (utid) and stack profile callsite.
- Populates these tables during trace ingestion of `ArtProcessMetadata` packets and sets the `dump_reason` in the `heap_graph` table to 'OOME'.
- Integrates the OOME callsite into `DeobfuscationTracker` to enable package guessing and proguard deobfuscation on OOME stack traces.
- Adds parser and deobfuscation diff tests to verify correctness.


src/trace_processor/importers/proto/proto_trace_parser_impl.cc
```diff
@@ -108,6 +113,10 @@ void ProtoTraceParserImpl::ParseTracePacket(int64_t ts, TracePacketData data) {
       module->ParseTraceConfig(config);
     }
   }
+
+  if (packet.has_art_process_metadata()) {
```
LalitMaganti: THe modern way to do this is via plugins. Can we please rework this into a new plugin [1]? You can use the RegisterProtoImporterModule functions and there are other examples using this.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/src/trace_processor/core/plugin/plugin.h?q=plugin.h%20f:perfetto&ss=chromium

src/trace_processor/importers/proto/proto_trace_parser_impl.cc
```diff
@@ -365,6 +374,185 @@ void ProtoTraceParserImpl::ParseMetatraceEvent(int64_t ts, ConstBytes blob) {
...
+  for (auto it = stack_decoder.frames(); it; ++it) {
+    raw_frames.push_back(*it);
+  }
+
+  std::optional<CallsiteId> current_callsite_id = std::nullopt;
+  uint32_t depth = 0;
+
+  if (!art_oome_mapping) {
+    art_oome_mapping =
+        &context->mapping_tracker->CreateDummyMapping("art_oome");
```
LalitMaganti: @ilkos is there not a mapping we can give this code? Do we keep track of the mappings or do we just have the frame names?

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/memory.sql
```diff
@@ -282,3 +282,57 @@ CREATE PERFETTO VIEW memory_snapshot_edge(
...
+  utid JOINID(thread.id),
+  -- The callsite of the leaf frame of the stacktrace.
+  callsite_id JOINID(stack_profile_callsite.id)
+)
+AS
+SELECT id, heap_graph_id, utid, callsite_id
+FROM __intrinsic_heap_graph_thread_callsite;
+
+-- Details of Java OutOfMemoryError exceptions that triggered heap dumps.
+CREATE PERFETTO VIEW heap_graph_java_oome_details(
```
LalitMaganti: I think this table specifically, I would put in the android stdlib module for heap graphs since it's *very* specific to Android. The ones above not so much because it could hypothetically work for other heap graph types as well.

LalitMaganti (commented):
Overall looks fantastic great work! Some small suggestions to imrpove the code a bit,

src/trace_processor/importers/proto/proto_trace_parser_impl.cc
ilkos: Not really, this is an in-process callstack so we never dealt with mappings in the first place.

src/trace_processor/tables/profiler_tables.py
```diff
@@ -592,6 +592,107 @@
...
+        C('ts', CppInt64()),
+        C(
+            'upid',
+            CppUint32(),
+            cpp_access=CppAccess.READ,
+            cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
+        ),
+        C('dump_reason', CppOptional(CppString())),
+    ],
+    tabledoc=TableDoc(
```
ilkos, line 622: Can you add an additional entry for HeapGraph.heap_bytes_allocated (perfetto/protos/perfetto/trace/profiling/heap_graph.proto)

src/trace_processor/importers/proto/heap_graph_tracker.cc
```diff
@@ -813,6 +817,31 @@ void HeapGraphTracker::FinalizeProfile(uint32_t seq_id) {
 
   PopulateSuperClasses(sequence_state);
   PopulateNativeSize(sequence_state);
+
+  auto& heap_graph_table = *storage_->mutable_heap_graph_table();
+  std::optional<tables::HeapGraphTable::Id> heap_graph_id;
+  for (auto it = heap_graph_table.IterateRows(); it; ++it) {
```
LalitMaganti: Instead of manually doing equality checks with IterateRows, I suggest you use FilterToIterator and add equality constraints instead. It's not a huge deal given this table is very small *but* it's good practice in the long term.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/memory.sql
```diff
@@ -282,3 +282,34 @@ CREATE PERFETTO VIEW memory_snapshot_edge(
...
+  upid JOINID(process.upid),
+  -- Reason why the heap graph was dumped (e.g. OOME, periodic, manual).
+  dump_reason STRING,
+  -- Total bytes allocated in the heap as reported by the VM.
+  heap_size LONG
+)
+AS
+SELECT id, ts, upid, dump_reason, heap_size FROM __intrinsic_heap_graph;
+
+-- Callstack profiles of threads at the time of heap graphs.
```
LalitMaganti: nit: at the time the heap graph was collected.

src/trace_processor/plugins/art_oome_importer/art_oome_module.cc
```diff
@@ -0,0 +1,247 @@
...
+  return upid;
+}
+
+void UpdatePackageList(TraceProcessorContext* context,
+                       base::StringView package_name,
+                       int64_t uid) {
+  StringId package_name_id = context->storage->InternString(package_name);
+  auto* package_list = context->storage->mutable_package_list_table();
+  bool found = false;
+  for (auto it = package_list->IterateRows(); it; ++it) {
```
LalitMaganti: As above.

src/trace_processor/plugins/art_oome_importer/art_oome_module.cc
```diff
@@ -0,0 +1,247 @@
...
+         /*profileable_from_shell*/ false, /*version_code*/ 0});
+  }
+}
+
+tables::HeapGraphTable::Id InsertHeapGraph(TraceProcessorContext* context,
+                                           int64_t ts,
+                                           UniquePid upid) {
+  auto& heap_graph_table = *context->storage->mutable_heap_graph_table();
+  std::optional<tables::HeapGraphTable::Id> heap_graph_id;
+  for (auto it = heap_graph_table.IterateRows(); it; ++it) {
```
LalitMaganti: As above.

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
```diff
@@ -0,0 +1,247 @@
...
+    }
+  }
+  if (!found) {
+    context->storage->mutable_package_list_table()->Insert(
+        {package_name_id, uid, /*debuggable*/ false,
+         /*profileable_from_shell*/ false, /*version_code*/ 0});
+  }
+}
+
+tables::HeapGraphTable::Id InsertHeapGraph(TraceProcessorContext* context,
```
LalitMaganti, line 75: @ilkos do we have some guarantee whether this packet will appear before/after the Heap graph itself? This code is currently wirtten such that it can appear either way but it's quite verbose for that reason.

src/trace_processor/tables/profiler_tables.py
```diff
@@ -592,6 +592,125 @@
...
+    python_module=__file__,
+    class_name='HeapGraphThreadCallsiteTable',
+    sql_name='__intrinsic_heap_graph_thread_callsite',
+    wrapping_sql_view=WrappingSqlView('heap_graph_thread_callsite'),
+    columns=[
+        C(
+            'heap_graph_id',
+            CppTableId(HEAP_GRAPH_TABLE),
+            cpp_access=CppAccess.READ,
+            cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
```
LalitMaganti, line 647: I don't think we actually need these post-finalization.

src/trace_processor/tables/profiler_tables.py
```diff
@@ -592,6 +592,125 @@
...
+HEAP_GRAPH_TABLE = Table(
+    python_module=__file__,
+    class_name='HeapGraphTable',
+    sql_name='__intrinsic_heap_graph',
+    wrapping_sql_view=WrappingSqlView('heap_graph'),
+    columns=[
+        C(
+            'ts',
+            CppInt64(),
+            cpp_access=CppAccess.READ,
```
LalitMaganti: Please check and see if we really need READ/POST_FINALIZATION here: if you copy/poasted, it might not actually be needed.

src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/oome.sql
```diff
@@ -0,0 +1,40 @@
...
+--
+
+-- Details of Java OutOfMemoryError exceptions that triggered heap dumps.
+CREATE PERFETTO VIEW android_heap_graph_java_oome_details(
+  -- Unique identifier for this details row.
+  id ID,
+  -- The heap graph instance this OOM trigger details belongs to.
+  heap_graph_id JOINID(heap_graph.id),
+  -- Number of bytes that triggered the OOME.
+  byte_count LONG,
```
ilkos: maybe name this allocation_size_bytes (byte_count in this context could mean a lot of things) ?

src/trace_processor/plugins/art_oome_importer/art_oome_importer.cc
```diff
@@ -0,0 +1,52 @@
...
+#include <memory>
+
+#include "perfetto/base/compiler.h"
+#include "src/trace_processor/core/plugin/plugin.h"
+#include "src/trace_processor/importers/proto/proto_importer_module.h"
+#include "src/trace_processor/plugins/art_oome_importer/art_oome_module.h"
+
+namespace perfetto::trace_processor::art_oome_importer {
+
+class ArtOomeImporter : public Plugin<ArtOomeImporter> {
```
ilkos: Why not ArtProcessMetadataImporter to match the proto name? This is not necessarily OOME specific (e.g. package names, process names etc apply to all heap dumps you can collect). Ditto for namespace, directory structure etc

src/trace_processor/plugins/art_oome_importer/art_oome_module.cc
```diff
@@ -0,0 +1,247 @@
...
+    case TracePacket::kArtProcessMetadataFieldNumber:
+      ParseArtProcessMetadata(ts, decoder.art_process_metadata());
+      return;
+  }
+}
+
+void ArtOomeModule::ParseArtProcessMetadata(int64_t ts,
+                                            protozero::ConstBytes blob) {
+  protos::pbzero::ArtProcessMetadata::Decoder decoder(blob.data, blob.size);
+  if (!decoder.has_oom_allocation_size()) {
```
ilkos: OOME is one of the modes heap dumps run as. It's possible that oom_allocation_size is not present, but other metadata is (e.g. process name, uid, etcetc)

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
ilkos: art_process_metadata will always come before the heap_graph packet (is that what you meant?).

src/trace_processor/tables/profiler_tables.py
```diff
@@ -592,6 +592,125 @@
...
+    sql_name='__intrinsic_heap_graph_java_oome_details',
+    wrapping_sql_view=WrappingSqlView('android_heap_graph_java_oome_details'),
+    columns=[
+        C(
+            'heap_graph_id',
+            CppTableId(HEAP_GRAPH_TABLE),
+            cpp_access=CppAccess.READ,
+            cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
+        ),
+        C('byte_count', CppInt64()),
```
ilkos: As above, allocation_size_bytes ?

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
BogdanLionte: Do you want me to remove the handling and assume that art_process_metadata always comes first?

ilkos: Yes, it will always come first so if it results in simpler code we should do it

src/trace_processor/importers/proto/heap_graph_tracker.cc
BogdanLionte: Done

src/trace_processor/plugins/art_oome_importer/art_oome_module.cc
BogdanLionte: Done

src/trace_processor/plugins/art_oome_importer/art_oome_module.cc
BogdanLionte: Done

src/trace_processor/tables/profiler_tables.py
BogdanLionte: I double checked and removed them where not necessary.

src/trace_processor/tables/profiler_tables.py
BogdanLionte: I double checked and removed them where not necessary.

src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/oome.sql
BogdanLionte: Done

src/trace_processor/plugins/art_oome_importer/art_oome_importer.cc
BogdanLionte: Done

src/trace_processor/tables/profiler_tables.py
BogdanLionte: Done

src/trace_processor/importers/proto/heap_graph_tracker.cc
```diff
@@ -813,6 +817,44 @@ void HeapGraphTracker::FinalizeProfile(uint32_t seq_id) {
 
   PopulateSuperClasses(sequence_state);
   PopulateNativeSize(sequence_state);
+
+  auto& heap_graph_table = *storage_->mutable_heap_graph_table();
+  std::optional<tables::HeapGraphTable::Id> heap_graph_id;
+  auto cursor = heap_graph_table.CreateCursor({
```
LalitMaganti: I would cache the cursor on the class the first time it's used (I.e. inside an optional).

src/trace_processor/importers/proto/heap_graph_tracker.cc
```diff
@@ -813,6 +817,44 @@ void HeapGraphTracker::FinalizeProfile(uint32_t seq_id) {
...
+          {},
+      },
+  });
+  cursor.SetFilterValueUnchecked(0, sequence_state.current_upid);
+  cursor.SetFilterValueUnchecked(1, sequence_state.current_ts);
+  cursor.Execute();
+  if (!cursor.Eof()) {
+    heap_graph_id = cursor.id();
+  }
+  if (heap_graph_id) {
```
LalitMaganti, line 845: I would next here and check that the cursor is actually empty.

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
```diff
@@ -0,0 +1,278 @@
...
+  return upid;
+}
+
+void UpdatePackageList(TraceProcessorContext* context,
+                       base::StringView package_name,
+                       int64_t uid) {
+  StringId package_name_id = context->storage->InternString(package_name);
+  auto* package_list = context->storage->mutable_package_list_table();
+  auto cursor = package_list->CreateCursor({
+      dataframe::FilterSpec{
```
LalitMaganti: As above, would cache this on the module class the first time you see it.

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
```diff
@@ -0,0 +1,278 @@
...
+         /*profileable_from_shell*/ false, /*version_code*/ 0});
+  }
+}
+
+tables::HeapGraphTable::Id InsertHeapGraph(TraceProcessorContext* context,
+                                           int64_t ts,
+                                           UniquePid upid) {
+  auto& heap_graph_table = *context->storage->mutable_heap_graph_table();
+  std::optional<tables::HeapGraphTable::Id> heap_graph_id;
+  auto cursor = heap_graph_table.CreateCursor({
```
LalitMaganti: As above

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
```diff
@@ -0,0 +1,278 @@
...
+  });
+  cursor.SetFilterValueUnchecked(0, upid);
+  cursor.SetFilterValueUnchecked(1, ts);
+  cursor.Execute();
+  if (!cursor.Eof()) {
+    heap_graph_id = cursor.id();
+  }
+  if (heap_graph_id) {
+    auto row_ref = heap_graph_table[*heap_graph_id];
+    row_ref.set_dump_reason(context->storage->InternString("OOME"));
```
LalitMaganti: I think we discussed this should never happen right because this always happens before?

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
```diff
@@ -0,0 +1,278 @@
...
+          1,
+          dataframe::Eq{},
+          {},
+      },
+  });
+  cursor.SetFilterValueUnchecked(0, upid);
+  cursor.SetFilterValueUnchecked(1, ts);
+  cursor.Execute();
+  if (!cursor.Eof()) {
+    heap_graph_id = cursor.id();
```
LalitMaganti: As above, please check cursor is empty.

src/trace_processor/importers/proto/heap_graph_tracker.cc
BogdanLionte: Done

src/trace_processor/importers/proto/heap_graph_tracker.cc
BogdanLionte: Done

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
BogdanLionte: Done

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
BogdanLionte: Deleted this entirely.

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
BogdanLionte: Yes, sorry, it got reverted somehow.

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
BogdanLionte: Deleted this entirely.

LalitMaganti (approved):

LalitMaganti:
Great work!

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
```diff
@@ -0,0 +1,255 @@
...
+    UpdatePackageList(context_, package_list_cursor_, decoder.package_name(),
+                      decoder.uid());
+  }
+
+  if (!decoder.has_oom_allocation_size()) {
+    return;
+  }
+
+  tables::HeapGraphTable::Id heap_graph_id =
+      InsertHeapGraph(context_, ts, upid);
```
ilkos: nit: I think we should run this for all heap dumps (not just OOME), and pass a boolean is_oome == decoder.has_oom_allocation_size()) to control the dump reason (right now there's no value set for non-OOME). So I'd move this call up a couple of lines. This way you can also set the size and other metadata-related entries here and remove the patching code from heap_graph_tracker

heap_graph_tracker of course should continue to insert if not present for backwards compatibility (older perfetto versions will not be emitting this metadata packet)

BogdanLionte: I agree with the first part of the comment, however,  for the second part, do we actually have heap_size (the total size of the heap dump) inside art_process_metadata_module?

ilkos: Yes I expect the heap size to be present for non oome heap dumps too

ilkos: Ignore the previous comment, I was replying to a different question (is it present for non oome), but as you pointed out it's in a different packet type

ilkos (approved):

PR #6177, opened 2026-06-09
[DNS] ui: SurfaceFlinger layers viewer (com.android.SurfaceFlinger)

zezeozue:
Adds a SurfaceFlinger layers viewer for the android.surfaceflinger.layers data source. Inspired by the Android Winscope tool.

trace_processor: the winscope importer plugin now reports the timestamp bounds of the winscope snapshot tables (SurfaceFlinger layers + transactions, WindowManager, shell transitions, IME, ViewCapture, ProtoLog). Previously a trace captured with only the winscope data sources had empty trace bounds and the UI treated it as empty; now it opens directly. Adds read access to those tables' `ts` column.

Surfaced two ways onto one shared session:
- A nested "SurfaceFlinger" timeline group with one track per display (real and virtual, e.g. a video-encoder/screen-recorder display, which are part of what SurfaceFlinger composited). Each track shows that display's snapshots as frame-numbered, colorized slices; selecting one shows a compact summary plus an inline layout preview and opens the page at that snapshot.
- A full-screen page (sidebar > SurfaceFlinger) with a display selector, snapshot time-slider, prev/next, and a jump-to-timeline button that selects the snapshot slice on that display's track. The snapshot index and timestamp are shown consistently in the page and the details panel.

Three panes:
- Surface: a rotatable, stacked 3D view of the selected display's layer rects, pure 2D-canvas projection (no WebGL). Camera angles follow the standalone tool's mapper3d (pitch = factor*PI/8, yaw = 1.5*pitch); gradient / opacity / wireframe shading; spacing slider; only-visible filter; per-rect hide + pin; click-to-select; and leader-line labels in a side gutter that never overlap and are themselves clickable.
- Hierarchy: the full cross-display layer tree with chips (visible / GPU / HWC / relative-Z / hidden / spy / duplicate), only-visible / flat / simplify-names / diff filters and search, and per-node hide + pin toggles.
- Properties: a header card (name + chips), a curated summary as a DataGrid (with clickable layer references) and the full proto dump as a filterable DataGrid; with diff on, layers are marked added/modified/deleted and the proto grid gains a Previous column.

A layer belongs to a display via its rect's group id, so the Surface scopes to the selected display's composition while the hierarchy/properties span all displays. Built from native widgets (Section, Tree, Chip, DataGrid, Checkbox, TextInput, Select) and themed with the --pf-color-* variables, so it follows the app's light and dark mode. Enabled by default.

### DIFF
--- a/src/trace_processor/plugins/winscope_importer/winscope_importer.cc
+++ b/src/trace_processor/plugins/winscope_importer/winscope_importer.cc
@@ -16,12 +16,17 @@
 
 #include "src/trace_processor/plugins/winscope_importer/winscope_importer.h"
 
+#include <cstdint>
+#include <limits>
 #include <memory>
+#include <utility>
 
 #include "perfetto/base/compiler.h"
 #include "src/trace_processor/core/plugin/plugin.h"
 #include "src/trace_processor/importers/proto/proto_importer_module.h"
 #include "src/trace_processor/plugins/winscope_importer/winscope_module.h"
+#include "src/trace_processor/storage/trace_storage.h"
+#include "src/trace_processor/types/trace_processor_context.h"
 
 namespace perfetto::trace_processor::winscope_importer {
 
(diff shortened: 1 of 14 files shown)


stevegolton:
Could you add the following:
- At least one representative trace (not just a minimal example - something that properly exercises the various code paths and UI elements).
- Some playwright integration tests for the new features.
- Some instructions on how to use and test the new features.

All plugins, especially default enabled plugins, submitted to the codebase carry a contract that we will continue to maintain them and host them on ui.perfetto.dev going forward. Without representative traces and comprehensive tests, this is next to impossible.

LalitMaganti:
We discussed offline, this code is not landing. We should make clear this is purely a prototype

PR #6184, opened 2026-06-10
smaps plugin: instants track, datagrid details panel

rsavitski:
For traces with smaps packets, we now show process-scoped tracks with
instants representing those smaps snapshots. When selected, the list of
mappings is shown in the details panel as a datagrid.

One complication is that typically a trace won't record all possible
value columns (such as rss, anon, etc), so I wanted to render only the
relevant columns in the details panel to make it less
cluttered/misleading. The underlying SQL table doesn't have nullability
(unset readings default to 0), and I am not convinced that adding
nullability is worth the increased effort to query the table by people.

What I've done instead: the entire smaps table is scanned and only
columns that have a nonzero reading are shown in all smaps details
panels for that trace. Similarly, the "aggregate_count" is only shown if
at least one snapshot has aggregation. So there is some ambiguity and a
particular corner case: if a value was recorded, but was truly zero for
all snapshots in the trace (right now that column would end up hidden).

Also, I'm deliberately not doing anything special for a hypothetical
trace with smaps packets recorded with different configs.


ui/src/plugins/dev.perfetto.Smaps/smaps_details_panel.ts
```diff
@@ -0,0 +1,158 @@
...
+  {name: 'locked_kb', title: 'Locked (KB)'},
+];
+
+// Computes the columns worth showing in the datagrid, using the entire table
+// of memory mapping snapshots. For the memory value columns, which might not
+// all be recorded: the sql columns are not nullable, so instead we show any
+// column that has non-zero values (which is indistinguishable from truly
+// all-zeroes readings). Similarly, the aggregate count is only shown if it's
+// ever >1.
+export async function computeSmapsSchema(
```
stevegolton: Small suggestion - the DataGrid has two separate concepts: 'schema' vs 'columns'.  The schema describes the tables and columns that are available to be added, and the columns defines the list of columns that are currently being displayed.

I'd leave the schema wide open - just include all the columns. Then, narrow down the useful columns and supply a separate list of columns which is a subset of the schema.

This will allow users to add any of the other columns manually if they so please, but still displays the reduced set of columns that are actually worth showing initially.

stevegolton (approved):
LGTM % one small suggestion, but I'm not going to block on it.

ui/src/plugins/dev.perfetto.Smaps/smaps_details_panel.ts
rsavitski: Ah, neat, migrated to initialColumns (with some assistance).

PR #6189, opened 2026-06-10
proto_merger: keep oneof fields and enum values deleted upstream

sashwinbalaji:
When a field still exists in our local proto but was removed upstream, proto_merger is supposed to keep it under a "not present upstream" comment. That worked for plain message fields, but fields inside a oneof and enum values were dropped from the output: the merge step tracked them as deleted, the serializer just never wrote them back.

- Write back deleted oneof fields and enum values, under the same comment used for the other deleted items.
- Split the sources into a lib target so tests can link against them.
- Add tests for the serializer and a merge round trip, hooked into perfetto_unittests.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -355,15 +355,8 @@ perfetto_cc_binary(
     name = "proto_merger",
     srcs = [
         ":src_protozero_multifile_error_collector",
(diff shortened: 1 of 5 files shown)


safayat-google (approved):
LGTM

PR #6206, opened 2026-06-11
ui: OomAdjScoreViz: improve track ordering, descriptions, and rendering

gignat-dev:
This commit consolidates several improvements to the OomAdjScoreViz plugin:

* Fix concurrency track cut: The per-bucket "processes count" track dropped its final return-to-zero point because `lead(ts) - ts` is NULL. Coalescing the trailing NULL duration to `trace_end() - ts` ensures the count drops to 0 at the right time when a bucket empties.
* Order bucket tracks: Bucket tracks are now ordered by `oom_score` descending (cached first, system last) matching the ranges in `oom_adjuster.sql`, rather than relying on arbitrary `SELECT DISTINCT` ordering.
* Describe score tracks: Added descriptions to per-process score tracks so their purpose is documented in the UI on hover and in track details.


ui/src/plugins/com.android.OomAdjScoreViz/index.ts
```diff
@@ -80,8 +83,14 @@ export default class OomAdjScoreViz implements PerfettoPlugin {
           window = await getTimeSpanOfSelectionOrVisibleWindow(ctx);
         }
 
+        // Order buckets by their oom_score range (per oom_adjuster.sql): the
+        // highest-score bucket (cached) first down to the lowest (system).
```
zezeozue: I think we should show it in opposite direction

zezeozue (approved):

ui/src/plugins/com.android.OomAdjScoreViz/index.ts
gignat-dev: I think it feels more natural to go from high to low (inspired from https://github.com/google/perfetto/blob/main/src/trace_processor/perfetto_sql/stdlib/android/oom_adjuster.sql ), but of course we can change it if the opposite seems better

zezeozue: You are probalby intersted in fg process more often than cached. There's usually a tonne of cached. But it probalby doesn't matter either ways. its a small handful of things.

stevegolton (approved):

ui/src/plugins/com.android.OomAdjScoreViz/index.ts
gignat-dev: Done, low to high.

PR #6213, opened 2026-06-11
protozero: speed up TypedProtoDecoder

LalitMaganti:
Replace the unconditional 1.6KB per-decode memset with a presence bitmap
(field storage left uninitialized, reads gated by a per-id bit), order the
wire-type dispatch as a frequency if-chain, and decode varints with an
unrolled fast path. ParseOneField() is force-inlined into ReadField() and
the ParseAllFields() decode loop: the ParseFieldResult is decomposed by the
compiler after inlining, so it does not round-trip through memory. Adds a
decode microbenchmark over real traces.

Since the field storage is no longer zero-initialized, an exhausted
RepeatedFieldIterator must not be dereferenced. The one caller that relied on
this (the structured query generator) is fixed to test the iterator instead.

Decode microbenchmark, linux_clang_release, pinned to one core, per iter
(wall = median of 7 repetitions; instructions via perf stat over a 1-rep vs
9-rep run so process overhead cancels), before -> after:

  FtraceEventScan/sched_and_ps      wall 16.94ms -> 11.59ms (1.46x)  instr 359.3M -> 237.6M (1.51x)
  TrackEventScan/chrome_rendering   wall  4.77ms ->  2.75ms (1.73x)  instr 101.5M ->  65.7M (1.54x)
  TracePacketScan/chrome_rendering  wall  5.44ms ->  3.44ms (1.58x)  instr 125.4M ->  83.9M (1.49x)

End-to-end trace_processor_shell parse (perf stat -r 3), before -> after:
  example_android_trace_30s.pb      instr 3.50G -> 3.32G (-5.1%)  cpu 325.6ms -> 329.1ms
  android_sched_and_ps.pb           instr 2.50G -> 2.36G (-5.5%)  cpu 212.3ms -> 205.1ms
  chrome_rendering_desktop.pftrace  instr 3.22G -> 2.93G (-9.2%)  cpu 247.7ms -> 230.5ms


include/perfetto/protozero/proto_decoder.h
```diff
@@ -334,14 +334,18 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
...
-    // so we return the 0th invalid field.
-    return fields_[0];
+    return kInvalidField;
+  }
+
+  // True if the field with the given id was seen while decoding. The presence
+  // bitmap replaces zero-initializing the whole field storage (a 1.6KB memset)
+  // with a clear of num_fields_/8 bytes.
+  bool HasField(uint32_t id) const {
+    return (presence_[id >> 6] >> (id & 63)) & 1u;
```
primiano: can we make this more redable?
```
    constexpr uint32_t kBitsPerWord = sizeof(*presence_*) * 8;
    const uint32_t word_idx = id / kBitsPerWord;
    const uint32_t bit_off = id % kBitsPerWord;
    return (presence_[word_index] & (1UL << bit_offset)) != 0;
```

include/perfetto/protozero/proto_decoder.h
```diff
@@ -464,30 +468,38 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
...
                       std::is_trivially_destructible<Field>::value &&
                       std::is_trivial<Field>::value,
                   "Field must be a trivial aggregate type");
-    memset(fields_, 0, sizeof(Field) * capacity_);
+    memset(presence_, 0, sizeof(uint64_t) * ((num_fields_ + 63) / 64));
     PERFETTO_DCHECK(capacity > 0);
   }
 
+  // Marks the field with the given id as present. See HasField().
+  void SetField(uint32_t id) { presence_[id >> 6] |= uint64_t(1) << (id & 63); }
```
primiano: ditto

```
presence_[id / 64] |= 1ULL << (id % 64);
```

include/perfetto/protozero/proto_decoder.h
```diff
@@ -464,30 +468,38 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
...
+
+  // Returned by Get()/at() for fields that were not seen while decoding. Always
+  // !valid() (id == 0).
+  static const Field kInvalidField;
```
primiano: this seems misplaced. move up at the beginning of the private/protected section

include/perfetto/protozero/proto_utils.h
```diff
@@ -309,6 +310,57 @@ inline const uint8_t* ParseVarInt(const uint8_t* start,
...
+// branch of ParseVarInt, which matters for long varints (e.g. ftrace
+// timestamps are 8-10 bytes). The caller must guarantee at least 10 readable
+// bytes at |p| and that p[0] has the continuation bit set. Returns nullptr on
+// an overlong (> 10 byte) varint. The 10th byte contributes only its low bit
+// (matching ParseVarInt).
+PERFETTO_ALWAYS_INLINE inline const uint8_t* ParseVarIntUnrolled(
+    const uint8_t* p,
+    uint64_t* out) {
+  uint64_t value = p[0] & 0x7Fu;
+#define PERFETTO_PZ_VARINT_STEP(i)                           \
```
primiano: can we make this a lambda rather than a #define?
I'm 90% sure it will emit the same code

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
```diff
@@ -560,11 +560,11 @@ base::StatusOr<std::string> GeneratorImpl::SqlSource(
...
-  if (sql.column_names()->size() != 0) {
-    std::vector<std::string> cols;
-    for (auto it = sql.column_names(); it; ++it) {
-      cols.push_back(it->as_std_string());
-    }
+  std::vector<std::string> cols;
+  for (auto it = sql.column_names(); it; ++it) {
+    cols.push_back(it->as_std_string());
+  }
+  if (!cols.empty()) {
```
primiano, line 567: is this diff intended? seems unrelated

include/perfetto/protozero/proto_decoder.h
```diff
@@ -610,9 +626,14 @@ class TypedProtoDecoder : public TypedProtoDecoderBase {
...
              sizeof(on_stack_storage_));
     }
+    // The presence bitmap is always on-stack; repoint and copy it.
+    presence_ = presence_storage_;
+    memcpy(presence_storage_, other.presence_storage_,
+           sizeof(presence_storage_));
   }
 
   Field on_stack_storage_[PROTOZERO_DECODER_INITIAL_STACK_CAPACITY];
+  uint64_t presence_storage_[(MAX_FIELD_ID + 1 + 63) / 64];
```
primiano, line 648: swap the order.
on_Stack_storage is large, but you use presence_storage_ more frequently.
likely swapping them gives you better dcache hit rate and also on arm allows using less instructions to compute the offset because it becomes near to |this|

include/perfetto/protozero/proto_decoder.h
```diff
@@ -334,14 +334,18 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
...
-    // If id >= size_ and <= num_fields, the id is valid but the field has not
-    // been seen while decoding (hence the stack storage has not been expanded)
-    // so we return the 0th invalid field.
-    return fields_[0];
+    return kInvalidField;
+  }
+
+  // True if the field with the given id was seen while decoding. The presence
+  // bitmap replaces zero-initializing the whole field storage (a 1.6KB memset)
+  // with a clear of num_fields_/8 bytes.
```
primiano: Add a DCHECK(id<num_field_); for documentation purposes. I'd add to the comment that the caller must check that id is within range

include/perfetto/protozero/proto_decoder.h
```diff
@@ -503,6 +515,10 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
   // case of a large number of repeated fields.
   Field* fields_;
 
+  // Presence bitmap, 1 bit per field id, sized to cover [0, num_fields_).
+  // Provided by the template specialization (always on-stack). See HasField().
+  uint64_t* presence_;
```
primiano, line 528: btw I think you need to update FindNextMatchingId to add a HasField() check, because that today that seems relies on zero-initialization of fields_ (probably some other places as well)

include/perfetto/protozero/proto_decoder.h
```diff
@@ -334,14 +334,18 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
...
+    // read. Out-of-range or never-seen ids resolve to a static invalid field.
+    if (PERFETTO_LIKELY(id < num_fields_) && HasField(id))
       return fields_[id];
-    // If id >= num_fields_, the field id is invalid (was not known in the
-    // .proto) and we return the 0th field, which is always !valid().
-    // If id >= size_ and <= num_fields, the id is valid but the field has not
-    // been seen while decoding (hence the stack storage has not been expanded)
-    // so we return the 0th invalid field.
-    return fields_[0];
+    return kInvalidField;
```
primiano, line 343: do you need kInvlidFieldId? isn't fields_[0] always invalid ? 
maybe we should just zero-initialize that one (only the 0th index)

src/protozero/proto_decoder.cc
```diff
@@ -178,66 +170,107 @@ Field ProtoDecoder::ReadField() {
...
+  // Inlined decode loop: unlike ReadField()/ParseOneField(), this writes each
+  // Field straight into its storage slot, avoiding the per-field
+  // ParseFieldResult round-trip.
+  const uint8_t* pos = begin_;
+  const uint8_t* const end = end_;
+  while (pos < end) {
+    const uint8_t* const field_start = pos;
+
+    // Tag.
+    uint64_t preamble;
```
primiano: don't you need 10 bytes to make parsevarintfast really fast?

src/protozero/proto_decoder.cc
```diff
@@ -178,66 +170,107 @@ Field ProtoDecoder::ReadField() {
...
-  const uint8_t* cur = begin_;
-  ParseFieldResult res;
-  for (;;) {
-    res = ParseOneField(cur, end_);
-    PERFETTO_DCHECK(res.parse_res != ParseFieldResult::kOk || res.next != cur);
-    cur = res.next;
-    if (PERFETTO_UNLIKELY(res.parse_res == ParseFieldResult::kSkip))
-      continue;
-    if (PERFETTO_UNLIKELY(res.parse_res == ParseFieldResult::kAbort))
+  // Inlined decode loop: unlike ReadField()/ParseOneField(), this writes each
```
primiano: hmm I am not convinced that this gains anything vs calling ParseOneField.
Surely that funcion gets inline (if not we should ALWAYS_INLINE it) and the compiler can see through that.

TBH this function became a bit of a mess now

include/perfetto/protozero/proto_decoder.h
LalitMaganti: Given this code runs on every proto parse in TP, I'm trying to reduce as much code on the hot path as possible. Given the low cost and complexity of having a static thing, I think it's worth it.

include/perfetto/protozero/proto_decoder.h
LalitMaganti: I audited the other places, it used Get, nothing else should be relying on zero init.

include/perfetto/protozero/proto_decoder.h
LalitMaganti: Done this is a nice 1-2% performance bump thanks!

src/trace_processor/perfetto_sql/generator/structured_query_generator.cc
LalitMaganti: Yes it's intentional. This change relied on fields being zero intiialized because it unconditionally dereferences the repeated field iterator. We no longer do this with the new code.

include/perfetto/protozero/proto_decoder.h
LalitMaganti: Done

LalitMaganti (commented):
Ptal now: I minimized the diff quite a bit while preserving the perf gains b

primiano (approved):

PR #6216, opened 2026-06-11
ui: BreakdownTracks: consolidate interval logic and improve BinderViz

gignat-dev:
This commit consolidates the following improvements to BreakdownTracks and AndroidBinderViz:

* Parallelize track creation: The server and client transaction track trees are
  now built concurrently via Promise.all, reducing trace engine idle time.
  Deterministic attach order (server first) is preserved.
* Single interval_self_intersect path: BreakdownTracks is now collapsed onto one
  code path built entirely on `interval_self_intersect`, deleting the legacy machinery
  and divergent code paths. `createTracks` now builds `_breakdown_intervals`,
  `_breakdown_segments`, and `_breakdown_projected` once. Lazy `CounterTrack.create`
  is used over the segments table.
* Group and describe breakdown tracks: The server and client transaction-count trees
  are nested under a single collapsible "Binder" summary group. A root-track description
  is added for each perspective, explaining the overlap-count grouping.


ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
+  private readonly pivotJoinClause: string;
+
+  // Unique per-instance materialized tables (see buildTables).
+  private readonly intervalsTableName: string;
+  private readonly segmentsTableName: string;
+  private readonly projectedTableName: string;
+
+  // Stable projected column names. The user-supplied column expressions are
+  // evaluated once into these names so every downstream filter is raw column
+  // equality (k0 = 'x') instead of re-evaluating the expression.
```
zezeozue: This comment is not very clear to me

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
-  private sliceJoinClause?: string;
-  private pivotJoinClause?: string;
+  private readonly props: BreakdownTrackProps;
+  private readonly uri: string;
+  private readonly modulesClause: string;
+  private readonly sliceJoinClause: string;
+  private readonly pivotJoinClause: string;
+
+  // Unique per-instance materialized tables (see buildTables).
+  private readonly intervalsTableName: string;
```
zezeozue, line 189: Can you preifily document what these tables do.

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
-  private modulesClause: string;
-  private sliceJoinClause?: string;
-  private pivotJoinClause?: string;
+  private readonly props: BreakdownTrackProps;
+  private readonly uri: string;
+  private readonly modulesClause: string;
+  private readonly sliceJoinClause: string;
+  private readonly pivotJoinClause: string;
+
+  // Unique per-instance materialized tables (see buildTables).
```
zezeozue: Per-instance of?

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
+ */
 export class BreakdownTracks {
-  private readonly props;
-  private uri: string;
-  private modulesClause: string;
-  private sliceJoinClause?: string;
-  private pivotJoinClause?: string;
+  private readonly props: BreakdownTrackProps;
+  private readonly uri: string;
+  private readonly modulesClause: string;
```
zezeozue, line 179: Short comment on these

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
+  // intervals end.
+  private aggValueExpr(): string {
+    switch (this.props.aggregationType) {
+      case BreakdownTrackAggType.MAX:
+        return 'MAX(IIF(interval_ends_at_ts = FALSE, agg_value, NULL))';
+      case BreakdownTrackAggType.SUM:
+        return 'SUM(IIF(interval_ends_at_ts = FALSE, agg_value, 0))';
+      case BreakdownTrackAggType.COUNT:
+      default:
+        return 'SUM(IIF(interval_ends_at_ts = FALSE, 1, 0))';
```
zezeozue, line 253: Do you need a default if you handled all the cases?

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
-        ), ts, dur)
-      `;
+  // The aggregate evaluated over each atomic segment's active (non-end-marker)
+  // intervals. COUNT counts them; MAX/SUM reduce the denormalized agg_value.
+  // End-markers contribute nothing so the counter naturally drops where
+  // intervals end.
+  private aggValueExpr(): string {
+    switch (this.props.aggregationType) {
+      case BreakdownTrackAggType.MAX:
+        return 'MAX(IIF(interval_ends_at_ts = FALSE, agg_value, NULL))';
```
zezeozue, line 249: Why is the aggregate value NULL for max as opposed to? MIN?

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
+          `LEFT JOIN ${joinTableName} USING(${joinColumns.join(', ')})`,
       )
       .join('\n');
   }
 
-  async createTracks() {
-    if (this.modulesClause !== '') {
-      await this.props.trace.engine.query(this.modulesClause);
-    }
+  // Projects per-category ts/dur plus the category's column expressions into
```
zezeozue: Whats a category? breakdown col?

zezeozue: Specifically, the use of 'category' here seems like a new concept so I'm trying to map it to existing concepts you already have if its not really new? Maybe there's a clearer name for projextXYZ?

ui/src/components/tracks/breakdown_tracks.ts
```diff
@@ -165,141 +154,238 @@ interface Filter {
...
-
-        ${this.getIntervals()}
-      `);
-    }
+  // Builds the three tables that drive every track:
+  //   _breakdown_intervals: one row per source interval (id, ts, dur, k*,
+  //     [agg_value]) from the aggregation table ONLY — no slice/pivot joins, so
+  //     the overlap count is never inflated by a 1:N join.
+  //   _breakdown_segments: per-(atomic segment, active interval) rows from
+  //     interval_self_intersect over the intervals, with k*/agg_value inline so
```
zezeozue: These comments coulld be clearer? breakdown columns innstead of k*?

zezeozue (approved):

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: In SQL, aggregate functions like  MAX()  and  SUM()  naturally ignore  NULL  values. If you use a dummy value (like  0 ,  MIN , or  -Infinity ) instead of  NULL  when the interval is ending, that dummy value could incorrectly become the  MAX  if all the valid, active values happen to be lower than the dummy value (e.g., negative numbers). Returning  NULL  is the safest, most standard way to completely exclude those non-active rows from the  MAX()  calculation.

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Done

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Done

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Done, updated the comment

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Done, added comments

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Done, removed the default

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Updated

ui/src/components/tracks/breakdown_tracks.ts
gignat-dev: Category was indeed a vague term. What it wanted to do was to differentiate columns from the `slices` and columns from `pivot`, so that they are renamed to `s0, s1, ..`, respectively `p0, p1, ..`.

Changed it to `columnGroup`.

zezeozue: I think columnGroup is also confusing. I still don't really understand whats projected from what and to what?

gignat-dev: You're right — `project` is the misleading part; the method only **aliases**. It
takes one column group (the `slice` or `pivots` config — each a set of SQL
expressions over the source table) and emits the SELECT-list entries that rename
those expressions to fixed column names, so downstream tracks/filters read
predictable columns instead of re-deriving the caller's SQL.

`from → to`:

- **slice** `{tsCol: 'client_ts', durCol: 'client_dur', columns: ['aidl_name', 'method_name']}`
  → `client_ts AS slice_ts`, `client_dur AS slice_dur`, `aidl_name AS s0`, `method_name AS s1`
- **pivot** `{tsCol: 'breakdown_ts', durCol: 'breakdown_dur', columns: ['reason']}`
  → `breakdown_ts AS pivot_ts`, `breakdown_dur AS pivot_dur`, `reason AS p0`

So it aliases *from* the caller's expressions *to* the fixed `slice_*`/`s0..` (or
`pivot_*`/`p0..`) columns of the projected table; the differing `s`/`p` prefixes
just keep the two groups from colliding in that shared row.

Wdyt about renaming `projectColumnGroup` → `aliasColumnGroup` or just `aliasColumns`?

stevegolton (approved):

PR #6218, opened 2026-06-11
tp: replace GetExtensionSlowly with SelectiveTracePacketDecoder

LalitMaganti:
Introduce SelectiveTracePacketDecoder, a wrapper around protozero's
SelectiveTypedProtoDecoder. An allowlist mask covers the TracePacket
metadata fields the pipeline reads by name; the data field and
out-of-tree extensions fall into unknown_fields(), which drives module
dispatch in one pass per packet instead of one buffer re-scan per
registered extension id. This deletes GetExtensionSlowly().

Module hooks now take an args struct and receive the dispatched field
as a TypedProtoField, read via generated field constants. The allowlist
must stay disjoint from module-registered ids (DCHECKed).

Wall time is neutral to ~+1% across the test traces, but extension
dispatch no longer scales with the number of registered extension
modules.


src/trace_processor/plugins/winscope_importer/winscope_module.cc
```diff
@@ -75,126 +76,111 @@ WinscopeModule::WinscopeModule(ProtoImporterModuleContext* module_context,
...
       surfaceflinger_layers_parser_.Parse(
-          timestamp,
-          decoder
-              .GetExtensionSlowly<
-                  FrameworksNativeWinscopeTracePacket::
-                      kSurfaceflingerLayersSnapshotFieldNumber>()
-              .as_bytes(),
+          args.ts,
+          args.field.Cast<
+              FrameworksNativeWinscopeTracePacket::kProtologViewerConfig>(),
```
sashwinbalaji: kSurfaceflingerLayersSnapshotFieldNumber

src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -388,29 +383,29 @@ ModuleResult TrackEventTokenizer::TokenizeTrackDescriptorPacket(
...
+    const TokenizePacketArgs& args) {
+  if (PERFETTO_UNLIKELY(!args.decoder.has_trusted_packet_sequence_id())) {
     context_->import_logs_tracker->RecordTokenizationError(
-        stats::thread_descriptor_missing_sequence_id, packet_blob->offset());
+        stats::thread_descriptor_missing_sequence_id, args.packet->offset());
     return ModuleResult::Handled();
   }
 
-  // TrackEvents will be ignored while incremental state is invalid. As a
+  // TrackEvents will be ignored while incremental args.state is invalid. As a
```
sashwinbalaji: state

src/trace_processor/importers/proto/proto_importer_module.h
```diff
@@ -106,6 +107,24 @@ class ModuleResult {
...
+  TraceBlobView* packet;
+  int64_t packet_timestamp;
+  RefPtr<PacketSequenceStateGeneration> state;
+  TracePacketField field;
+};
+
+// Arguments for ProtoImporterModule::ParseField(). See TokenizePacketArgs.
+struct ParseFieldArgs {
+  const SelectiveTracePacketDecoder& decoder;
+  int64_t ts;
```
sashwinbalaji, line 123: nit: can we call this also timestamp or update the above, just a bit off ot read

sashwinbalaji (approved):

PR #6221, opened 2026-06-12
tracing: fix TBv2 patching regression

rukkal:
Optimize TBv2 patching performance by fixing macro evaluation
overhead and adding a fast-fail range check.
    
In TraceBufferV2::TryPatchChunkContents, TRACE_BUFFER_V2_DLOG
was evaluating its arguments (specifically base::HexDump) even
in release builds.  This caused a ~10'000x slowdown during patch
writes. Redefine the macro to a real noop in non-verbose builds.

### DIFF
--- a/src/tracing/service/trace_buffer_benchmark.cc
+++ b/src/tracing/service/trace_buffer_benchmark.cc
@@ -230,6 +230,52 @@ static void BM_TraceBuffer_RD(benchmark::State& state) {
   state.SetBytesProcessed(static_cast<int64_t>(total_bytes_read));
 }
 
+// Benchmark 3: Patching performance
+template <typename BufferType>
+static void BM_TraceBuffer_Patch(benchmark::State& state) {
+  const size_t kBufferSize = 16 * 1024 * 1024;
+  const int patch_distance = static_cast<int>(state.range(0));
+
+  auto chunk_templates = GenerateChunkTemplates(10);
+  ClientIdentity client_identity(1000, 100);
+
+  auto buffer = BufferType::Create(kBufferSize);
+  PERFETTO_CHECK(buffer);
+
+  ChunkID chunk_id = 0;
+  size_t bytes_written = 0;
+  size_t template_idx = 0;
+  std::vector<ChunkID> written_chunk_ids;
+  while (bytes_written < kBufferSize) {
+    const auto& tmpl = chunk_templates[template_idx % chunk_templates.size()];
+    buffer->CopyChunkUntrusted(ProducerID(1), client_identity, WriterID(1),
+                               chunk_id, tmpl.num_fragments, tmpl.flags,
+                               /*chunk_complete=*/true, tmpl.data.data(),
+                               tmpl.data.size());
+    written_chunk_ids.push_back(chunk_id);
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Nice find! Though I don't think this is the right fix; I think the correct one is to wrap that section in a DCHECK_IS_ON ifdef. 

I also am shocked the compiler is not optimizing this out honestly.

primiano:
Wow thanks a lot for finding this, this was super subtle.

IN terms of fixing, as Lalit said, this is going to risk to fail build on release mode due to "unused variable ..."

THe two solutions I see:

1. For now, as Lalit suggested, wrap the code that has the base::HexDump into a `#if TRACE_BUFFER_V2_VERBOSE_LOGGING()`  ( it applies only to that specific callsite)

2. slightly more general, we could define 

```
#define TRACE_BUFFER_V2_DLOG(...) do { if constexpr(TRACE_BUFFER_V2_VERBOSE_LOGGING()) PERFETTO_DLOG((__VA_ARGS__) while(0)
```


3. One could ague we could do 2 at the PERFETTO_DLOG level.
But this one comes with a bit of a risk. If there is some code today that has some side-effects which happen to be load bearing, it would turn them into bugs. So I'd probably pass for now on this.

rukkal:
@primiano

> Wow thanks a lot for finding this, this was super subtle.
> 
> IN terms of fixing, as Lalit said, this is going to risk to fail build on release mode due to "unused variable ..."
> 
> THe two solutions I see:
> 
> 1. For now, as Lalit suggested, wrap the code that has the base::HexDump into a `#if TRACE_BUFFER_V2_VERBOSE_LOGGING()`  ( it applies only to that specific callsite)
> 2. slightly more general, we could define
> 
> ```
> #define TRACE_BUFFER_V2_DLOG(...) do { if constexpr(TRACE_BUFFER_V2_VERBOSE_LOGGING()) PERFETTO_DLOG((__VA_ARGS__) while(0)
> ```
> 
> 3. One could ague we could do 2 at the PERFETTO_DLOG level.
>    But this one comes with a bit of a risk. If there is some code today that has some side-effects which happen to be load bearing, it would turn them into bugs. So I'd probably pass for now on this.

Oh, I see. Why not going with 2. directly? See the follow up commit with
```
#if TRACE_BUFFER_V2_VERBOSE_LOGGING()
#define TRACE_BUFFER_V2_DLOG PERFETTO_DLOG
#else
#define TRACE_BUFFER_V2_DLOG(...)                    \
  do {                                               \
    if constexpr (TRACE_BUFFER_V2_VERBOSE_LOGGING()) \
      PERFETTO_DLOG(__VA_ARGS__);                    \
  } while (0)
#endif
```

primiano (approved):
This LGTM, but I think we can avoid the `#iif TRACE_BUFFER_V2_VERBOSE_LOGGING()` and instead unconditionally always define it with the constexpr no?

rukkal:
> This LGTM, but I think we can avoid the `#iif TRACE_BUFFER_V2_VERBOSE_LOGGING()` and instead unconditionally always define it with the constexpr no?

Sure! Done

PR #6223, opened 2026-06-12
ui: Synthesize version number and git hash at bundle time

stevegolton:
Currently the version info (scraped from the CHANGELOG plus the git hash) is written to a .ts file by the build script, and has to exist on disk before typescript could run cleanly. Stale or missing copies of it could be a source of confusion.

This patch generates a synthetic module containing these values on the fly in the bundler instead, with the types declared once in a static d.ts (they never change). The value now always reflects the tree being built, and we're a step closer to running tsc --noEmit over a raw checkout with no preparation step.


LalitMaganti (approved):
Nice this is very cool!

stevegolton:
Note: The formatting issues are pre-existing issues exposed here because this PR happens to touch the same files.

PR #6226, opened 2026-06-12
bt-ui: bigtrace add SQL formatting to the query editor

gignat-dev:
Format the editor tab's SQL with the same syntaqlite formatter the Query page uses: Alt+Shift+F in the editor (the option key on Mac) plus an icon-only toolbar button. The shortcut is documented in the help modal. No command — the chord can't be expressed in the command hotkey grammar, whose key matching breaks on option-composed characters on Mac; the editor binding matches by key code and works everywhere.

The engine init mirrors the Query page plugin's (it is private to the plugin and cannot be imported across bundles). The syntaqlite assets are now also copied into the bigtrace dist, which resolves assets against its own serving root, and bigtrace's CSP gains 'wasm-unsafe-eval' — Wasm compilation without JS eval().

### DIFF
--- a/ui/build.mjs
+++ b/ui/build.mjs
@@ -703,8 +703,22 @@ function copySyntaqliteRuntime() {
     'syntaqlite-sqlite.wasm',
   ]) {
     addTask(cp, [pjoin(srcDir, fname), pjoin(dstDir, fname)]);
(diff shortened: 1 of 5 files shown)


stevegolton (approved):
Nice

PR #6227, opened 2026-06-12
protozero: add selective decoding to TypedProtoDecoder

LalitMaganti:
Add SelectiveTypedProtoDecoder: ids whitelisted in a compile-time
SelectiveDecodeMask (built from a list of field ids) decode into the
dense storage as usual; everything else -- including ids beyond
MAX_FIELD_ID, which the normal decoder drops -- spills in wire order to
a small inline area (with heap fallback), iterated via unknown_fields().

- The dense window is capped at the mask's extent, so the decode loop's
  existing num_fields_ check also bounds the mask read: no extra
  per-field work, and the mask only spans the whitelisted ids.
- Subclass, not a TypedProtoDecoder constructor: spill members in the
  decoder itself cost ~1% instructions on an E2E trace_processor parse
  (decoders are constructed per nested message). The plain decoder is
  untouched.
- The decode loop is one implementation compiled into two
  specializations (~1.3KB of binary), so the plain path carries no
  trace of the mask support: verified at parity with the pre-selective
  decoder on both the decode microbenchmark and E2E parse.

TracePacket scan on chrome_rendering_desktop, 11 whitelisted metadata
fields (wall, median of 9):

  decode-everything (cannot see unknown/extension fields)      3.83ms
  + 4 GetExtensionSlowly() probes (extensions today)          10.93ms
  selective: 11 dense ids, data fields + extensions spilled    4.10ms

Supersedes the reverted SelectiveProtoDecoder (#6217) and enables
replacing GetExtensionSlowly()'s per-probe buffer re-scan in a
follow-up.


include/perfetto/protozero/proto_decoder.h
```diff
@@ -510,6 +550,17 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
 
   void ParseAllFields();
 
+  // Selective-decoding flavour, driven by SelectiveTypedProtoDecoder. The
+  // spill arguments are in/out and deliberately arguments rather than
+  // members: as members they would cost an initialization per decoder
+  // construction (decoders are constructed per nested message). |*spill| is
+  // repointed at |*heap_spill| if its capacity is exhausted.
+  void ParseAllFieldsSelective(const uint64_t* dense_mask,
```
sashwinbalaji, line 558: nit: how about a struct to wrap these params.

src/protozero/proto_decoder.cc
```diff
@@ -148,6 +148,23 @@ ParseOneField(const uint8_t* const buffer, const uint8_t* const end) {
...
 }
 
+// Grows the selective-decoding spill area once the inline slots are
+// exhausted, moving the spilled fields to the heap. The append itself is
+// open-coded in the decode loop; see the comment there.
+PERFETTO_NO_INLINE void ExpandSpill(Field** spill,
+                                    uint32_t* spill_capacity,
+                                    uint32_t spill_size,
+                                    std::unique_ptr<Field[]>* heap_spill) {
+  const uint32_t new_capacity = std::max(*spill_capacity * 2, 16u);
```
sashwinbalaji, line 161: we start spill_capacity at 16(kSpillStackCapacity) so the max seems unnecessary. If needed can we plumb kSpillStackCapacity instead of directly using 16u ?

src/protozero/proto_decoder.cc
```diff
@@ -188,17 +230,39 @@ void TypedProtoDecoderBase::ParseAllFields() {
...
+      if (!dense) {
+        // The spill append is open-coded from the field's scalar components:
+        // a function taking the whole Field (by value or reference) makes
+        // the compiler materialize the packed Field representation on every
+        // loop iteration, including for fields that are not spilled.
+        if (PERFETTO_UNLIKELY(*spill_size >= *spill_capacity))
+          ExpandSpill(spill, spill_capacity, *spill_size, heap_spill);
+        (*spill)[(*spill_size)++].initialize(
+            field_id, static_cast<uint8_t>(res.field.type()),
+            res.field.raw_int_value(), res.field.raw_size());
```
sashwinbalaji, line 254: wow, didn't expect this. so even doing direct assignment is not optimal ?

include/perfetto/protozero/proto_decoder.h
```diff
@@ -21,6 +21,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
+#include <initializer_list>
```
primiano: You don't seem to be using this

include/perfetto/protozero/proto_decoder.h
```diff
@@ -317,6 +318,43 @@ class UnifiedRepeatedFieldIterator {
...
+//       kMask{};
+//   SelectiveTypedProtoDecoder<MAX_FIELD_ID> decoder(data, size, kMask);
+//   for (const Field& f : decoder.unknown_fields()) { ... }
+template <uint32_t... kFieldIds>
+class SelectiveDecodeMask {
+ public:
+  static_assert(sizeof...(kFieldIds) > 0, "Empty field id list");
+  static constexpr uint32_t kMaxFieldId = std::max({kFieldIds...});
+
+  constexpr SelectiveDecodeMask() : words_{} {
```
primiano: uh how can this be constexpr? it modifies words_

include/perfetto/protozero/proto_decoder.h
```diff
@@ -317,6 +318,43 @@ class UnifiedRepeatedFieldIterator {
...
+    for (uint32_t id : {kFieldIds...})
+      words_[id / 64] |= uint64_t(1) << (id % 64);
+  }
+  const uint64_t* data() const { return words_; }
+
+ private:
+  uint64_t words_[kMaxFieldId / 64 + 1];
```
primiano: you need {} otherwise this is not default-initialized IIRC so you get random ones here depending on what you have on the stack

include/perfetto/protozero/proto_decoder.h
```diff
@@ -649,6 +725,109 @@ class TypedProtoDecoder : public TypedProtoDecoderBase {
   Field on_stack_storage_[PROTOZERO_DECODER_INITIAL_STACK_CAPACITY];
 };
 
+// TypedProtoDecoder with selective decoding. Ids whitelisted in the mask are
```
primiano: s/whitelisted/allowlisted/ otherwise i'm sure that some bot somewhere will be ouraged and file some bug

include/perfetto/protozero/proto_decoder.h
```diff
@@ -649,6 +725,109 @@ class TypedProtoDecoder : public TypedProtoDecoderBase {
...
+  }
+
+  SelectiveTypedProtoDecoder& operator=(
+      SelectiveTypedProtoDecoder&& other) noexcept {
+    if (this != &other) {
+      Base::operator=(std::move(other));
+      spill_ = other.spill_;
+      spill_size_ = other.spill_size_;
+      heap_spill_ = std::move(other.heap_spill_);
+      MaybeCopyOnStackSpill(other);
```
primiano: no need to maintain two copies for the same stuff. my usual pattern here is:

    this->~SelectiveTypedProtoDecoder();
    new (this) SelectiveTypedProtoDecoder(std::move(other));

src/protozero/proto_decoder.cc
primiano: I think this is more guarding against the case (not sure how taht would happen... std::move maybe) where you end up with 0 capacity, so you don't just double a zero

src/protozero/proto_decoder.cc
```diff
@@ -188,17 +230,39 @@ void TypedProtoDecoderBase::ParseAllFields() {
...
-    // `extensions 1000 to 1999` range). They are intentionally not stored here
-    // because fields_ is indexed directly by field id, and extension ids are
-    // sparse and potentially very high. Callers that need them must use
-    // TypedProtoDecoderBase::GetExtensionSlowly().
-    // TODO: store extensions in a sparse, object-pooled side table so they stay
-    // accessible without a buffer re-scan. See GetExtensionSlowly() for the
-    // proposed design.
-    if (PERFETTO_UNLIKELY(field_id >= num_fields_))
+    if constexpr (kSelective) {
+      // Selective decoding: only whitelisted in-range ids go to the dense
```
primiano: sitto s/whitelisted/allowlisted/

primiano (approved):
LGTM. mind the "white" vs "allow" list

include/perfetto/protozero/proto_decoder.h
LalitMaganti: Structs in general are scary when it comes to really perf sensitive code like this, you are relying on compiler to see through it. Prefer keeping as is.

include/perfetto/protozero/proto_decoder.h
LalitMaganti: For primitive fields initialization, you can still have it be constexpr and use the class in a constexpr setting.

PR #6237, opened 2026-06-15
ui: SQL editor syntaqlite Lsp integration 

gignat-dev:
Adds an opt-in dev.perfetto.SqlEditorIntelligence plugin that wires the
QueryPage editor to syntaqlite's in-process LSP server (wasm_lsp_message)
through the stock @codemirror/lsp-client. The stock extensions provide
completion, diagnostics, hover docs, signature help, go-to-definition,
references, and rename; our side is loading the WASM engine + dialect,
a small in-memory transport, and feeding the stdlib schema via the
syntaqlite/setSessionContext extension request.

- The WASM runtime + dialect load once; each trace gets its own WASM
  session, disposed on trace unload. An LSP server accepts exactly one
  initialize, so sharing a session would reject the next trace's client
  handshake.
- Editor widget: new generic `extensions` prop, held in a Compartment so
  late-registering plugins reconfigure already-mounted editors (plugin
  onTraceLoad can run seconds after the query page first renders).
- QueryPage: editor intelligence is a per-tab factory - every query tab
  stays mounted, and the LSP workspace requires one document per view.
  The factory memoizes per document; a fresh extension identity would
  close and reopen the server document on every redraw.
- SqlModules: expose waitForSqlModules() so consumers can await the
  module catalog instead of polling.
- @codemirror/autocomplete is version-overridden so basicSetup and the
  LSP client resolve one shared instance; two copies each render their
  own completion popup.

Known gap: unknown_function diagnostics flag every stdlib call (the
server's function catalog is SQLite built-ins only). Left visible
pending upstream support for declaring functions in the session
context.


LalitMaganti:
Question: why are we adding so much code here rather than switching to Monaco and then using the LSP layer of syntaqlite for autocomplete?

Would suggest looking at the editor experience at https://playground.syntaqlite.com/ and seeing if we can work with that before doing anything else.

stevegolton:
> Question: why are we adding so much code here rather than switching to Monaco and then using the LSP layer of syntaqlite for autocomplete?
> 
> Would suggest looking at the editor experience at https://playground.syntaqlite.com/ and seeing if we can work with that before doing anything else.

Codemirror has an LSP client plugin if that's all we're missing.

I'd be wary of switching editors just for the sake of it.

Granted, I have no experience with either monaco or the codemirror lsp editor so take this with a huge grain of salt.

LalitMaganti:
Personally I enjoyed my experience of working with Monaco in syntaqlite and certainly like the final result more than the look we get with codemirror (makes sense since it's just vscode's experience at the end of the day).

But I'm fine with whatever we do here: I'm more concerned about (poorly) reinventing LSP here which is what a lot of this change appears to be.

stevegolton:
> But I'm fine with whatever we do here: I'm more concerned about (poorly) reinventing LSP here which is what a lot of this change appears to be.

Agreed, let's not re-invent the wheel.

gignat-dev:
> Question: why are we adding so much code here rather than switching to Monaco and then using the LSP layer of syntaqlite for autocomplete?
> 
> Would suggest looking at the editor experience at https://playground.syntaqlite.com/ and seeing if we can work with that before doing anything else.

1. Although I really like how Monaco looks, from what I researched, it seems to add a really big bundle size.

2. I didn't use syntaqlite autocompletion because I wanted the autocompletion to be richer, e.g. definitions

image

image


Maybe we could look into improving the syntaqlite autocompletion first, or just use the bare autocompletion it provides. What do you think?

LalitMaganti:
syntaqlite auto omplete supports definitions just fine. You just need to feed it the schema for this. The playground shows how to do this. And if there's specific cases which are not working I would like to understand that so we can fix them. 

Also fwiw we should not be putting a lot of dedicated work in this space anyway because in the longer term with a combination of data explorer + AI I don't think many people will be writing SQL by hand really.

Also also there's another problem here which is that there appears to be a lot of hardcoded logic here with respect to PerfettoSQL which will change a lot when PerfettoSQL next comes in with pipe syntax and custom interval/tree/graph operators.

stevegolton:
I still think there's value to be had in having this autocomplete available, not only in the query page but anywhere where you can type in perfettosql directly - including in the DE page.

I won't pretend to understand how the built in syntaqlite completions work but I think we should just use whatever it provides + use the codemirror lsp plugin to hook it up.

gignat-dev:
I agree with Steve. 
I'll try to prepare a lighter version of this prototype (no custom autocompletion logic, no documentation tooltips, no bigtrace ui changes, etc.) and start again from there.

gignat-dev:
Hi, Steve! After some research I found out that syntaqlite doesn't have a LSP server that runs in the browser - the lsp command is stdio only, and the WASM build exposes direct functions instead (wasm_completions and wasm_diagnostics). So the **CodeMirror LSP plugin** would have nothing to connect to right now.

After consulting with Lalit, the correct solution for this is to **add a JSON-RPC entry point to syntaqlite-wasm so a normal LSP client can talk to it from a worker.**

So, the conclusion is: we'll need to wait for upstream syntaqlite changes. [issue](https://github.com/LalitMaganti/syntaqlite/issues/298)

gignat-dev:
Hey, I just pushed a much lighter version of the feature, which uses the syntaqlite LSP. 
Also updated the description.

Please lmk what you think.

LalitMaganti:
Overall looks a *lot* better. Not too sold on the "editor intelligence" thing but I'll leave that to Steve to look at.

ui/package.json
```diff
@@ -72,5 +73,10 @@
     "build": "node build.mjs",
     "test": "node build.mjs --run-unittests",
     "lint": "npx eslint . --ext .js,.ts"
+  },
+  "pnpm": {
+    "overrides": {
+      "@codemirror/autocomplete": "^6.20.3"
```
stevegolton: This feels avoidable

ui/src/plugins/dev.perfetto.SqlModules/index.ts
```diff
@@ -101,6 +103,11 @@ export default class SqlModulesPlugin implements PerfettoPlugin {
     return this.sqlModules;
   }
 
+  // Resolves once the modules have loaded and getSqlModules() is set.
+  waitForSqlModules(): Promise<SqlModules> {
```
stevegolton: I don't think we need this - the plugin dependency framework will ensure that ontraceload completes for all dependant plugins.

stevegolton (approved):
Honestly, looks great % a couple of nits!

Re the naming, i'd just call it what it is - 'language server' or similar

ui/src/plugins/dev.perfetto.SqlModules/index.ts
gignat-dev: Done, removed that waitForSqlModules.

ui/package.json
gignat-dev: You're right. I deleted that override.

gignat-dev:
I renamed the plugin to `dev.perfetto.SqlLsp`

Please lmk what you think.

stevegolton (approved):
LGTM!

PR #6238, opened 2026-06-15
tp: refactor IteratorImpl to be an abstract base for remote iterators

LalitMaganti:
Split IteratorImpl into an abstract base plus final subclass, so a
remote TraceProcessor can have its own implementation. The
public ExecuteQuery function returns the Iterator by value, so the
polymorphism has to live behind the pImpl pattern.

To avoid regressing existing callers from direct calls to virtual valls,
Iterator caches the SqliteIteratorImpl* directly and forwarders to it directly.
Because SqliteIteratorImpl is `final`, the compiler should devirtualizes the local path to
a direct call (verified in the disassembly). QueryResultSerializer is unchanged: it
still takes the concrete impl and devirtualizes per cell.

### DIFF
--- a/src/trace_processor/iterator_impl.h
+++ b/src/trace_processor/iterator_impl.h
@@ -17,149 +17,50 @@
 #ifndef SRC_TRACE_PROCESSOR_ITERATOR_IMPL_H_
 #define SRC_TRACE_PROCESSOR_ITERATOR_IMPL_H_
 
-#include <sqlite3.h>
-#include <cstddef>
 #include <cstdint>
 #include <string>
 
-#include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
-#include "perfetto/ext/base/scoped_file.h"
-#include "perfetto/ext/base/status_or.h"
 #include "perfetto/trace_processor/basic_types.h"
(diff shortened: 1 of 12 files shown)


src/trace_processor/iterator_impl.h
```diff
@@ -17,149 +17,42 @@
...
+  virtual std::string GetColumnName(uint32_t col) const = 0;
+  virtual base::Status Status() const = 0;
+  virtual uint32_t ColumnCount() const = 0;
+  virtual uint32_t StatementCount() const = 0;
+  virtual uint32_t StatementCountWithOutput() const = 0;
+  virtual std::string LastStatementSql() = 0;
+
+  // Returns |this| if this is the local iterator implementation,
+  // otherwise nullptr. Allows for fast-path optimizations.
+  virtual SqliteIteratorImpl* AsSqlite() { return nullptr; }
```
sashwinbalaji, line 53: this seems a bit off. ain't there a better solution (I'm currently thinking) ?

LalitMaganti: Unfortunately not something that gives us performance 😕

sashwinbalaji (approved):

PR #6248, opened 2026-06-15
tp: add unix socket protocol for RPCs

LalitMaganti:
Add a third RPC transport, `trace_processor_shell server unix <trace>`, that
serves the trace processor an AF_UNIX socket addressed by name, so
repeated `query --remote <name>` calls can reference the same trace.

Lifecycle (idle timeout, orphan detection, --daemonize) and the --remote
client land in follow-up CLs.


src/trace_processor/shell/server_subcommand.cc
sashwinbalaji: nit: misses unix

src/trace_processor/shell/server_subcommand.cc
```diff
@@ -154,8 +172,48 @@ base::Status ServerSubcommand::Run(const SubcommandContext& ctx) {
 #endif
   }
 
-  return base::ErrStatus("server: unknown mode '%s' (expected http or stdio)",
-                         mode.c_str());
+  if (mode == "unix") {
+    // Resolve the socket path: --path wins, else --name, else a generated name.
+    std::string socket_path = socket_path_;
+    std::string session_name = session_name_;
```
sashwinbalaji, line 183: in the else block when socket_path not empty and session_name not empty we dont invoke IsValidSessionName is that fine maybe a very advanced scenario?

src/base/file_utils.cc
```diff
@@ -273,6 +273,32 @@ bool Mkdir(const std::string& path) {
...
+
+  // Creates a single directory, treating an already-existing directory as
+  // success. `mode` is ignored on Windows.
+  auto mkdir_one = [mode](const std::string& p) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+    base::ignore_result(mode);
+    return _mkdir(p.c_str()) == 0 || errno == EEXIST;
+#else
+    return mkdir(p.c_str(), mode) == 0 || errno == EEXIST;
+#endif
```
sashwinbalaji: nit: can we updated base::Mkdir to take mode with default value = 0755 and invoke that to minimise cope duping?

src/trace_processor/rpc/session_paths.cc
```diff
@@ -0,0 +1,134 @@
...
+    return false;
+  for (char c : name) {
+    if (!std::isalnum(static_cast<unsigned char>(c)) && c != '_' && c != '-')
+      return false;
+  }
+  return true;
+}
+
+std::string GenerateSessionName(uint64_t seed) {
+  std::minstd_rand rng(static_cast<std::minstd_rand::result_type>(seed));
```
sashwinbalaji: found using AI (but it makes sense):

```
        uint64_t seed = static_cast<uint64_t>(base::GetWallTimeNs().count()) ^
                        (static_cast<uint64_t>(base::GetProcessId()) << 32);
        session_name = session::GenerateSessionName(seed);
```

std::minstd_rand::result_type is uint32_t so the pid stuff is gone from seed meaning a very slight chance of similar session_name

LalitMaganti:
Updated thanks!

sashwinbalaji (approved):

PR #6252, opened 2026-06-16
ui: insert 1px gap on right edge of slices

### LINKED ISSUE #6250: Boundary of same named scopes isn't visible.
Perfetto has lost the boundary of scopes with the same name.  These all just merge into one long color bar.   There used to be a gap or border.  Even when hovering over the scope, the color doesn't change at all.

Here's an example, where I've launched gnuplot 3x, and I want to visually see the split between each scope.  I can hover and see the timings, but I've lost the visual on this.

Image

LalitMaganti:
Allows two slices which are literally starting and ending at the same
timestamp to have a 1px gap which makes it possible to distinguish
between multiples instances of the same slice

Note we keep the minimum 1px width even for sub-pixel slices to avoid
regressing the zoomed-out behavior.

Screenshot of how this looks:

Screenshot 2026-06-16 at 01 31 35

Fixes: https://github.com/google/perfetto/issues/6250


stevegolton (approved):
Nice! LGTM.

Tested it zoomed out when the slices approach 1px in width and it looks fine.

PR #6259, opened 2026-06-16
ui: Added 'OOM callstack' track

TParaschiv:
Added UI options to display the OOM callstack traces

Screenshots:
https://screenshot.googleplex.com/4hJbxNeuEHp37jz
https://screenshot.googleplex.com/5hqXPG7GXtLYFx5


ui/src/plugins/dev.perfetto.OomCallstack/index.ts
```diff
@@ -0,0 +1,196 @@
+// Copyright (C) 2024 The Android Open Source Project
```
stevegolton: Nit: 2026

ui/src/core/embedder/default_plugins.ts
```diff
@@ -77,6 +77,7 @@ export const defaultPlugins = [
   'dev.perfetto.LinuxPerf',
   'dev.perfetto.MetricsPage',
   'dev.perfetto.Notes',
+  'dev.perfetto.OomCallstack',
```
LalitMaganti: I'm not sure if this should be a new pluign or be folded into the heap dump plugin honestly.

ui/src/plugins/com.android.HeapDumpExplorer/heap_dump_page.ts
```diff
@@ -191,6 +192,20 @@ function buildTabs(
     },
   ];
 
+  if (overview.hasOomCallstack) {
```
LalitMaganti: In general, I would always try and show all the tabs we have and just have a "this was not collected" rather than hiding the whole tab. It makes clear to people the *capability* is present.

ui/src/plugins/dev.perfetto.OomCallstack/oom_callstack_details_panel.ts
```diff
@@ -0,0 +1,169 @@
...
+    private state: FlamegraphState | undefined,
+    private readonly onStateChange: (state: FlamegraphState) => void,
+  ) {
+    this.metrics = [
+      {
+        name: 'OOM Callstack',
+        unit: '',
+
+        statement: `
+          WITH RECURSIVE callstack AS (
```
LalitMaganti: THis seems repeated. Can we please share it via the standard library?

ui/src/core/embedder/default_plugins.ts
stevegolton: As a default enabled plugin, would you mind adding a representative trace and a basic playwright test that simply loads that trace and takes a screenshot of the timeline

ui/src/plugins/com.android.HeapDumpExplorer/types.ts
```diff
@@ -63,6 +63,11 @@ export interface OverviewData {
   dmabufRssSize: bigint | null;
   /** The process uptime at the time of the heap dump. */
   processUptime: bigint | null;
+  /** True if the heap dump has an associated OOM callstack. */
+  hasOomCallstack: boolean;
+
```
stevegolton: Nit - whitespace

ui/src/plugins/dev.perfetto.OomCallstack/oom_callstack_details_panel.ts
```diff
@@ -0,0 +1,169 @@
+// Copyright (C) 2024 The Android Open Source Project
```
stevegolton: Nit 2026

ui/src/plugins/dev.perfetto.OomCallstack/oom_callstack_details_panel.ts
```diff
@@ -0,0 +1,169 @@
...
+          {
+            name: 'source_location',
+            displayName: 'Source Location',
+            mergeAggregation: 'ONE_OR_SUMMARY',
+          },
+        ],
+      },
+    ];
+
+    if (this.state === undefined) {
```
stevegolton: When is this.state ever not undefined in the constructor?

ui/src/plugins/dev.perfetto.OomCallstack/oom_callstack_details_panel.ts
```diff
@@ -0,0 +1,169 @@
...
+      );
+      this.onStateChange(this.state);
+      this.serialization.state = undefined;
+    }
+  }
+
+  render() {
+    return m(
+      FlamegraphProfile,
+      undefined,
```
stevegolton: Nit: You don't need the undefined in place of the props here

stevegolton (commented):
LGTM from the non-HDE side, % a handful of nits.

The main think I ask is about the representative trace and smoke test.

stevegolton:
+ @zezeozue for the HDE changes

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
```diff
@@ -0,0 +1,145 @@
...
+  ts: time,
+): ReadonlyArray<QueryFlamegraphMetric> {
+  return [
+    {
+      name: 'OOM Callstack',
+      unit: '',
+      statement: `
+        WITH RECURSIVE callstack AS (
+          SELECT c.id, c.parent_id as parentId, c.depth, c.frame_id
+          FROM stack_profile_callsite c
```
zezeozue: Could this be written with _graph_scan?

ui/src/plugins/com.android.HeapDumpExplorer/queries.ts
```diff
@@ -497,6 +497,23 @@ export async function getOverview(
     });
   }
 
+  const oomeRes = await engine.query(`
+    SELECT g.upid, g.ts
+    FROM heap_graph g
+    JOIN process p ON g.upid = p.upid
+    JOIN process p_dump ON p.pid = p_dump.pid
+    WHERE p_dump.upid = ${activeDump.upid} AND g.dump_reason = 'OOME'
```
zezeozue: Shouldnt you also filter for:         AND graph_sample_ts = ${activeDump.ts}

ui/src/plugins/dev.perfetto.OomCallstack/index.ts
```diff
@@ -0,0 +1,196 @@
...
+  private async createEventTable(trace: Trace) {
+    await trace.engine.query(
+      'INCLUDE PERFETTO MODULE android.memory.heap_graph.oome;',
+    );
+
+    await createPerfettoTable({
+      engine: trace.engine,
+      name: EVENT_TABLE_NAME,
+      as: `
+        SELECT
```
zezeozue: Can you extrac this into a stdlib function? it seems to have some assumptions/heuristics

ui/src/core/embedder/default_plugins.ts
TParaschiv: moved to `dev.perfetto.HeapProfile`

ui/src/plugins/dev.perfetto.OomCallstack/oom_callstack_details_panel.ts
TParaschiv: moved to "dev.perfetto.HeapProfile/oom_callstack_common.ts"

ui/src/plugins/dev.perfetto.OomCallstack/oom_callstack_details_panel.ts
TParaschiv: removed "ui/src/plugins/dev.perfetto.OomCallstack/oom_callstack_details_panel.ts"

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
```diff
@@ -181,8 +190,30 @@ export default class HeapProfilePlugin implements PerfettoPlugin {
...
-          'heap_profile:' || heap_name AS type
+          'heap_profile:' || heap_name AS type,
+          ts AS reference_ts
         FROM heap_profile_slices
+
+        UNION ALL
+
+        SELECT
+          g.id,
+          MIN(g.ts, (SELECT end_ts - 50000000 FROM trace_bounds)) AS ts,
```
LalitMaganti: This feels a bit odd. WHy are we doing this?

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
```diff
@@ -181,8 +190,30 @@ export default class HeapProfilePlugin implements PerfettoPlugin {
...
+          IFNULL((
+            SELECT p.upid 
+            FROM process p
```
LalitMaganti: This seems unnecessary, why not just g.upid?

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
TParaschiv: 2 parts:
1. `MIN(g.ts, (SELECT end_ts FROM trace_bounds))` -> Because the timestamp for the OOM callstack is outside the timeline of the track. So in this case, we use the last timestamp available
2. `end_ts - 50000000` -> because it's hard to click half of a chevron/diamond/icon that's right at the end of the track. So I moved it a bit to the left so that the whole icon is visible & clickable

LalitMaganti: 1) this should be fixed in trace processor by making the oom table participate in trace bounds logic
2) this is something that needs to be fixed centrally by @stevegolton not ad-hoc like this. We never lie about things in Perfetto, especially not timestamps as it makes debugging for Perfetto infinitely harder.

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
TParaschiv: Because the heap_graph with `dump_reason = OOME` has a different `upid` than all the other tracks. So if I use `g.upid` it will create a new entry for the main process and add just the "OOM Callstack" track under it.
 
`g.upid` is used as a fallback in case there are no other tracks available under the root process

LalitMaganti: This seems to be a bug on the device side: I don't understand why it would have a different upid in the first palce.

LalitMaganti: @ilkos can you please clarify?

ilkos: I'm not sure I understand (what are 'all the other tracks'? could you share a trace in the chat?). The heap graph only uses the profiled process pid everywhere (aka the parent pid of the forked process). That's what perfetto_hprof::WriteHeapPackets accepts and uses for all cases

ui/src/plugins/com.android.HeapDumpExplorer/types.ts
```diff
@@ -63,6 +63,10 @@ export interface OverviewData {
   dmabufRssSize: bigint | null;
   /** The process uptime at the time of the heap dump. */
   processUptime: bigint | null;
+  /** The upid of the process that triggered the OOM. */
+  oomUpid: number | null;
```
zezeozue: Showing the upid in the ui is usually not useful

ui/src/plugins/com.android.HeapDumpExplorer/types.ts
```diff
@@ -63,6 +63,10 @@ export interface OverviewData {
...
+  /** The timestamp of the OOM event. */
+  oomTs: bigint | null;
```
zezeozue: Is this any different to the timestamp of the heap graph? Is there any value in showing this?

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
TParaschiv: added `heap_graph_table` to the `start_ns` & `end_ns` calculations in "src/trace_processor/plugins/storage_tables/storage_tables.cc"

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
ilkos: AFAICT from your screenshot the pid is the same, so must be something in trace processor that ends up creating > 1 process table entries

ui/src/core/embedder/default_plugins.ts
TParaschiv: added the playwright test

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
TParaschiv: @ilkos added the timestamp to the OOM dump so it's no longer outside of the parent process duration.
Simplified the query

ui/src/plugins/dev.perfetto.HeapProfile/oom_callstack_common.ts
```diff
@@ -0,0 +1,87 @@
...
+          UNION ALL
+
+          SELECT c.id, c.parent_id as parentId, c.depth, c.frame_id
+          FROM stack_profile_callsite c
+          JOIN callstack ON c.id = callstack.parentId
+        )
+        SELECT
+          cs.id,
+          cs.parentId,
+          ifnull(f.name, '[Unknown]') as name,
```
ilkos: coalesce(f.deobfuscated_name, f.name) (also why would f.name ever be null)

ui/src/plugins/dev.perfetto.HeapProfile/oom_callstack_common.ts
```diff
@@ -0,0 +1,87 @@
...
+      SELECT oome.error_msg 
+      FROM heap_graph g
+      JOIN android_heap_graph_java_oome_details oome ON oome.heap_graph_id = g.id
+      WHERE g.ts = ${ts}
+      LIMIT 1
+    `);
+    if (errorMsgRes.numRows() > 0) {
+      return errorMsgRes.firstRow({error_msg: STR_NULL}).error_msg ?? undefined;
+    }
+  } catch (e) {
```
ilkos: Why catch? What can throw?

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
```diff
@@ -169,7 +177,8 @@ export default class HeapProfilePlugin implements PerfettoPlugin {
           upid,
           0 AS dur,
           0 AS depth,
-          'java_heap_graph' AS type
+          'java_heap_graph' AS type,
+          graph_sample_ts AS reference_ts
```
ilkos: There's already a ts, why are we introducing a separate one

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const OomCallstackView: m.ClosureComponent<OomCallstackViewAttrs> = () => {
+  let cachedMetrics: ReadonlyArray<QueryFlamegraphMetric> | undefined;
+  let cachedKey: string | undefined;
+  let oomErrorMsg: string | undefined;
+
+  async function loadErrorMsg(trace: Trace, ts: time) {
+    oomErrorMsg = await loadOomErrorMsg(trace.engine, ts);
+    m.redraw();
```
stevegolton: Redraws are triggered after every query has completed so you don't need to manually call it here.

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
```diff
@@ -0,0 +1,94 @@
...
+            state,
+            onStateChange: attrs.onStateChange,
+          }),
+        ),
+      );
+    },
+  };
+};
+
+export default OomCallstackView;
```
stevegolton: This isn't well documented anywhere, but we tend to avoid having default exports from modules.

Just export OomCallstackView in-place and import it like so: `import {OomCallstackView} from ...`.

The reason is it makes it easier to automatically import classes and functions from other files by name.

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
TParaschiv: Right, this was used when the ts did not match. Removed it

ui/src/plugins/dev.perfetto.HeapProfile/oom_callstack_common.ts
stevegolton: Agreed - I'd just let this one bubble if the query fails.

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
```diff
@@ -0,0 +1,94 @@
...
+import {Flamegraph, type FlamegraphState} from '../../../widgets/flamegraph';
+import {Stack} from '../../../widgets/stack';
+import {
+  buildOomCallstackMetrics,
+  loadOomErrorMsg,
+} from '../../dev.perfetto.HeapProfile/oom_callstack_common';
+
+interface OomCallstackViewAttrs {
+  readonly trace: Trace;
+  readonly upid: number | null;
```
stevegolton: Please prefer undefined over null

stevegolton (changes requested):

ui/src/plugins/dev.perfetto.HeapProfile/oom_callstack_common.ts
TParaschiv: replaced

ui/src/plugins/dev.perfetto.HeapProfile/oom_callstack_common.ts
TParaschiv: removed

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
TParaschiv: removed

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
TParaschiv: removed the `export` and changed the `import`

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
TParaschiv: replaced null with undefined

LalitMaganti (approved):

stevegolton (approved):

ui/src/plugins/com.android.HeapDumpExplorer/queries.ts
```diff
@@ -497,6 +497,22 @@ export async function getOverview(
     });
   }
 
+  const oomeRes = await engine.query(`
+    SELECT g.upid, g.ts
+    FROM heap_graph g
+    JOIN process p ON g.upid = p.upid
+    JOIN process p_dump ON p.pid = p_dump.pid
```
ilkos: Why join twice?

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
```diff
@@ -231,8 +251,26 @@ export default class HeapProfilePlugin implements PerfettoPlugin {
       }
 
       for (const upid of upids) {
-        const group = trackGroupsPlugin.getGroupForProcess(upid);
-        if (!group) continue;
+        let group = trackGroupsPlugin.getGroupForProcess(upid);
```
ilkos: Why do we need this in the first place? I am wondering if this is the right way to do this, or if this should be handled centrally in dev.perfetto.ProcessThreadGroups.

stevegolton: You shouldn't have to deal with the undefined case. You could `assertExists(group)` it if you wanted to be sure it exists or crash.

Are you seeing a case where this returns undefined?

TParaschiv: Yes. This was added because the trace I used for tests had multiple upids for the same pid. And only one upid has a TrackNode created. 
A similar use case exists in ui/src/plugins/dev.perfetto.GpuByProcess/index.ts

TParaschiv: Added `getOrCreateGroupForProcess` to [ui/src/plugins/dev.perfetto.ProcessThreadGroups/index.ts‎](https://github.com/google/perfetto/pull/6259/changes#diff-c829ed5b85fb66e40af64427e07065c3e307e91df417e6f78188e47f425b0deb)

ui/src/plugins/com.android.HeapDumpExplorer/queries.ts
TParaschiv: Because the trace that I used to test on had multiple upids for the same pid and I wanted to group every track under the same process.

Since https://googleplex-android-review.git.corp.google.com/c/platform/art/+/40462030, I'm not sure this is possible anymore. Removed he JOINS

ui/src/plugins/dev.perfetto.HeapProfile/index.ts
TParaschiv: Removed `getOrCreateGroupForProcess` and rolled back to using `if (!group) continue;`

ilkos (approved):

stevegolton (approved):

zezeozue (approved):

ui/src/plugins/com.android.HeapDumpExplorer/views/oom_callstack_view.ts
LalitMaganti: Because it's a single callstack, it's not worth it. Let's leave it like this and fix this up when we do PSQL Next.

ui/src/plugins/dev.perfetto.OomCallstack/index.ts
TParaschiv: moved the logic to dev.perfetto.HeapProfile

ui/src/plugins/com.android.HeapDumpExplorer/types.ts
TParaschiv: removed

ui/src/plugins/com.android.HeapDumpExplorer/types.ts
TParaschiv: removed

ui/src/plugins/com.android.HeapDumpExplorer/queries.ts
TParaschiv: but there is no activeDump when creating the heap_profile_events table. And the stack trace data is not stored in the `heap_graph_object` table.

It's enough to populate heap_profile_events with the  "oome_callstack" `type` & `ts`. The `ts` is then used in `oome_callstack_common.ts` to recreate the stack by querying `heap_graph_thread_callsite`.

PR #6264, opened 2026-06-16
converting.md: simplify examples a bit

noamraph:
I went over the getting-started/converting doc, and it's great!

I suggest a few small things, perhaps it will be useful:

1. Instead of defining TRUSTED_PACKET_SEQUENCT_ID in each example, define it once.

2. Instead of repeating the calculation of uuid.uuid4().int & ((1 << 63) - 1), define a function get_uuid().

3. I also removed the f from a string which has no `{}` parts, because I got a warning in my editor about it.

Thanks!
Noam

### DIFF
--- a/docs/getting-started/converting.md
+++ b/docs/getting-started/converting.md
@@ -88,7 +88,7 @@ from perfetto.trace_builder.proto_builder import TraceProtoBuilder
 from perfetto.protos.perfetto.trace.perfetto_trace_pb2 import TrackEvent, TrackDescriptor, ProcessDescriptor, ThreadDescriptor
 
 # Define a unique ID for this sequence of packets (generate once per trace producer)
-TRUSTED_PACKET_SEQUENCT_ID = 1001 # Choose any unique integer
+TRUSTED_PACKET_SEQUENCE_ID = 1001 # Choose any unique integer
 
 def get_uuid() -> int:
     return uuid.uuid4().int & ((1 << 63) - 1)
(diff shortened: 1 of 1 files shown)


docs/getting-started/converting.md
```diff
@@ -87,6 +87,12 @@ import uuid
 from perfetto.trace_builder.proto_builder import TraceProtoBuilder
 from perfetto.protos.perfetto.trace.perfetto_trace_pb2 import TrackEvent, TrackDescriptor, ProcessDescriptor, ThreadDescriptor
 
+# Define a unique ID for this sequence of packets (generate once per trace producer)
```
LalitMaganti, line 90: small nit: I would prefer if we didn't choose 1 as the example here because everyone will copy it :D

I suggest instead some high number similar to the sort we have below.

LalitMaganti (commented):
Thanks a lot for sending this! One small comment but otherwise LGTM!

docs/getting-started/converting.md
noamraph: Thanks! Done.

LalitMaganti (approved):
LGTM thanks!

docs/getting-started/converting.md
```diff
@@ -87,6 +87,12 @@ import uuid
...
+TRUSTED_PACKET_SEQUENCT_ID = 1001 # Choose any unique integer
```
safayat-google: `TRUSTED_PACKET_SEQUENCE_ID`?

noamraph: Oops! Fixed.

LalitMaganti (approved):

safayat-google (approved):
LGTM. Thanks!

PR #6286, opened 2026-06-17
tp: add RemoteTraceProcessor and the `--remote` client

LalitMaganti:
Add a RemoteTraceProcessor: a faithful, transport-only implementation of the
TraceProcessor interface whose every method marshals to the corresponding
TraceProcessorRpc message and talks to a `server unix` session.


src/trace_processor/shell/common_flags.cc
```diff
@@ -471,4 +477,81 @@ base::StatusOr<base::TimeNanos> LoadTraceFile(
...
+      {!opts.override_stdlib_path.empty(), "--override-stdlib"},
+      {!opts.register_files_dir.empty(), "--register-files-dir"},
+      {!opts.raw_metric_v1_extensions.empty(), "--metric-extension"},
+      {!opts.metatrace_path.empty(), "--metatrace"},
+  };
+  for (const auto& f : incompatible) {
+    if (f.is_set) {
+      return base::ErrStatus(
+          "%s cannot be combined with --remote: the trace is already loaded "
+          "and "
```
sashwinbalaji: formatting seems a bit off here and above

sashwinbalaji (approved):

src/trace_processor/rpc/remote_trace_processor.cc
```diff
@@ -0,0 +1,614 @@
...
+}
+
+base::Status RemoteTraceProcessor::ReadResponse(std::vector<uint8_t>* out) {
+  // Hoisted out of the loop: ProtoRingBuffer's fastpath can return a message
+  // pointing into the last buffer passed to Append(), so it must stay alive.
+  uint8_t buf[4096];
+  for (;;) {
+    auto msg = rxbuf_.ReadMessage();
+    if (msg.valid()) {
+      if (msg.fatal_framing_error)
```
sashwinbalaji, line 278: I think this if check should be outside the valid block.

(sorry, was going through this along with #6289 and saw now)

PR #6289, opened 2026-06-17
tp: add self-managing lifecycle (idle timeout + orphan detection)

LalitMaganti:
Make unix sessions clean themselves up so if an agent starts one and
backgrounds it, it will eventually free up the memory.

Done by combination of checking if the parent pid is alive (for agent
harnesses which have explicit support for background jobs) and
falling back to a plain idle timeout for agents which don't (i.e.
30 minutes since the last query).


src/trace_processor/rpc/session_lifecycle.cc
```diff
@@ -0,0 +1,149 @@
...
+#if !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+#include <unistd.h>
+#else
+#include <tlhelp32.h>
+#endif
+
+namespace perfetto::trace_processor {
+namespace {
+
+int64_t NowMs() {
```
sashwinbalaji, line 36: GetWallTimeMs

src/trace_processor/rpc/unixd.cc
```diff
@@ -114,9 +139,41 @@ const char* g_socket_path_for_signal = nullptr;
     unlink(g_socket_path_for_signal);  // async-signal-safe.
   _exit(0);
 }
-#endif
+
+// Forks into the background. The parent prints the startup record (it knows the
+// child pid) and exits; the child detaches (setsid + redirect std fds) and
+// returns to keep serving. The socket is already bound before this is called,
+// so the record the parent prints is accurate.
+void DaemonizeAndPrintRecord(const UnixServerArgs& args) {
```
sashwinbalaji, line 143: can we try to reuse base::Daemonize https://crsrc.org/c/third_party/perfetto/src/base/utils.cc;l=246

src/trace_processor/rpc/unixd.cc
```diff
@@ -74,6 +78,25 @@ std::string ShellQuoteIfNeeded(const std::string& value) {
...
 }
 
+// Prints the one-line, machine-parseable startup record to |f|.
+void PrintStartupRecord(FILE* f,
+                        int pid,
+                        const std::string& session,
+                        const std::string& socket_path,
+                        uint32_t idle_timeout_ms) {
+  if (idle_timeout_ms == 0) {
+    fprintf(f, "perfetto-session pid=%d session=%s socket-path=%s idle-timeout=never\n",
```
sashwinbalaji: minro nit: maybe local string to store idle-timeout so just one fprintf

src/trace_processor/rpc/unixd.cc
LalitMaganti: As discussed offline this is complicated because of Windows. I'm going to do in a followup

sashwinbalaji (approved):

PR #6294, opened 2026-06-17
tp: Place video frames at presentation time

zezeozue:
Frames were placed at the TracePacket timestamp, which is the encode-drain time and its later than when the frame was shown and didn't line up with other SurfaceFlinger events.

Placed each frame at its presentation time (pts_us) instead which is now much better aligned (and resilient to encoding delays) with the VSYNC-sf track and can be used to correlate with frametimeline events.

(Feedback from the Android platform team)


src/trace_processor/plugins/video_frame_importer/video_frame_module.cc
```diff
@@ -115,7 +118,20 @@ void VideoFrameModule::ParseVideoFrame(const TracePacket::Decoder& decoder,
     row.codec = frame.has_codec() ? static_cast<int32_t>(frame.codec()) : 0;
     row.is_key_frame = frame.is_key_frame() ? 1 : 0;
     if (frame.has_pts_us()) {
-      row.pts_us = static_cast<int64_t>(frame.pts_us());
+      int64_t pts_us = static_cast<int64_t>(frame.pts_us());
+      row.pts_us = pts_us;
```
LalitMaganti: Trace processor has one hyper important rule: the timestamp you place in tables must be the timestamp which went through the sorter queues :)

You are breaking that here. If you want to do this, you need to tokenize the packet, pull out the prs_ys at that point and push that through the sorter queue along with the translation below.

LalitMaganti (approved):

PR #6297, opened 2026-06-17
Add timestamp to ArtProcessMetadata proto

ilkos:
Bug: 511177236


LalitMaganti:
Why do you need this? Why not just use the one at the TracePacket level?

ilkos:
Indeed, ignore

PR #6299, opened 2026-06-17
sdk: support machine_id on the in-process backend

dreveman:
Add a machine_id to TracingInitArgs that the in-process backend attributes its trace data to, so separate in-process traces can be recorded under distinct machine ids.

The id flows from TracingInitArgs to ConnectProducerArgs (via the muxer) and into ClientIdentity(uid, pid, machine_id) in InProcessTracingBackend, so the service stamps TracePacket.machine_id on the trusted packets. The C ABI gains
PerfettoProducerBackendInitArgsSetMachineId, wired into PerfettoProducerInProcessInit.

It also fixes shared-memory mode selection: a non-default machine_id previously forced ShmemMode::kShmemEmulation, which is only valid for relay (remote-host) producers whose SMB is copied over IPC. An in-process producer has a real shared SMB, so emulation mode violated the SMB chunk-state invariant and crashed on first emit. Both SMB-setup paths now select kDefault for in-process producers regardless of machine id. The system backend is unaffected: it derives the machine id service-side and ignores this field.

### DIFF
--- a/include/perfetto/tracing/tracing.h
+++ b/include/perfetto/tracing/tracing.h
@@ -76,6 +76,17 @@ struct TracingInitArgs {
   uint32_t backends = 0;                     // One or more BackendTypes.
   TracingBackend* custom_backend = nullptr;  // [Optional].
 
+  // [Optional] Machine id to attribute this process's trace data to. Only
+  // honored by the in-process backend; the system backend derives the machine
+  // id service-side and ignores this. Lets separate in-process traces be
+  // recorded under distinct machine ids. 0 means the default (host) machine.
+  //
+  // For a non-default (non-host) machine id, the tracing session's TraceConfig
+  // must opt into matching non-host machines (set trace_all_machines, or a
+  // machine_name_filter that matches); otherwise the data sources are filtered
+  // out and record no data.
+  uint32_t machine_id = 0;
+
   // [Optional] Platform implementation. It allows the embedder to take control
   // of platform-specific bits like thread creation and TLS slot handling. If
   // not set it will use Platform::GetDefaultPlatform().
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -342,6 +342,16 @@ void AppendOwnedSlicesToPacket(std::unique_ptr<uint8_t[]> data,
   }
 }
 
+// Shmem emulation is only for relay (remote-host) producers whose SMB is copied
+// over IPC. An in-process producer always has a real shared SMB, so it must use
+// kDefault even when it carries a non-default machine id.
+SharedMemoryABI::ShmemMode GetShmemMode(const ClientIdentity& client_identity,
+                                        bool in_process) {
+  return (client_identity.machine_id() == kDefaultMachineID || in_process)
+             ? SharedMemoryABI::ShmemMode::kDefault
+             : SharedMemoryABI::ShmemMode::kShmemEmulation;
+}
+
 }  // namespace
 
 TracingServiceImpl::TracingServiceImpl(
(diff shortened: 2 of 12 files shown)


src/tracing/service/tracing_service_impl.cc
```diff
@@ -430,9 +430,13 @@ TracingServiceImpl::ConnectProducer(Producer* producer,
...
           "Adopting producer-provided SMB of %zu kB for producer \"%s\"",
           shm_size / 1024, endpoint->name_.c_str());
-      auto shmem_mode = client_identity.machine_id() == kDefaultMachineID
-                            ? SharedMemoryABI::ShmemMode::kDefault
-                            : SharedMemoryABI::ShmemMode::kShmemEmulation;
+      // Shmem emulation is only for relay (remote-host) producers whose SMB is
+      // copied over IPC. An in-process producer always has a real shared SMB,
+      // so it must use kDefault even when it carries a non-default machine id.
+      auto shmem_mode =
+          (client_identity.machine_id() == kDefaultMachineID || in_process)
```
sashwinbalaji: nit: can we create a small helper function for this. It would remove the duplicate and one single point of comment.

include/perfetto/tracing/tracing.h
```diff
@@ -76,6 +76,12 @@ struct TracingInitArgs {
   uint32_t backends = 0;                     // One or more BackendTypes.
   TracingBackend* custom_backend = nullptr;  // [Optional].
 
+  // [Optional] Machine id to attribute this process's trace data to. Only
+  // honored by the in-process backend; the system backend derives the machine
+  // id service-side and ignores this. Lets separate in-process traces be
+  // recorded under distinct machine ids. 0 means the default (host) machine.
+  uint32_t machine_id = 0;
```
sashwinbalaji, line 88: let's also add: for a id != host machine id(0), you need to set trace_all_machines / machine_name_filter to get the data.

context:

```
  // Check machine name filter for multi-machine (traced_relay) tracing.
  bool is_host_machine =
      producer->client_identity().machine_id() == kDefaultMachineID;
  if (cfg_data_source.machine_name_filter_size()) {
    // Explicit set of machines to match.
    const auto& cfg_names = cfg_data_source.machine_name_filter();
    bool filter_match =
        NameMatchesFilter(producer->machine_name_, cfg_names, {});
    // Special case: "host" also always matches the host machine.
    bool host_match =
        is_host_machine && NameMatchesFilter("host", cfg_names, {});
    if (!filter_match && !host_match) {
      PERFETTO_DLOG("Data source: %s is filtered out for machine: %s",
                    cfg_data_source.config().name().c_str(),
                    producer->machine_name_.c_str());
      return nullptr;
    }
  } else if (!tracing_session->config.trace_all_machines() &&
             !is_host_machine) {
    // Default matching behaviour starting from perfetto v54: match only host,
    // unless the config sets a top level flag.
    PERFETTO_DLOG("Data source: %s is filtered out for remote machine: %s",
                  cfg_data_source.config().name().c_str(),
                  producer->machine_name_.c_str());
    return nullptr;
  }
```

src/tracing/service/tracing_service_impl.cc
dreveman: Done

include/perfetto/tracing/tracing.h
dreveman: Done

src/tracing/service/tracing_service_impl.cc
```diff
@@ -162,6 +162,17 @@ std::unique_ptr<TracingService> TracingService::CreateInstance(
 namespace tracing_service {
 
 namespace {
+
+// Shmem emulation is only for relay (remote-host) producers whose SMB is copied
+// over IPC. An in-process producer always has a real shared SMB, so it must use
+// kDefault even when it carries a non-default machine id.
+SharedMemoryABI::ShmemMode GetShmemMode(const ClientIdentity& client_identity,
```
sashwinbalaji: nit: place it below the variables and with other functions pls.

sashwinbalaji (approved):

src/tracing/service/tracing_service_impl.cc
dreveman: Done

PR #6300, opened 2026-06-18
ui: compute tracing time bounds per (machine, trace)

dreveman:
getTracingMetadataTimeBounds aggregated tracing_started_ns,
all_data_source_started_ns and tracing_disabled_ns across the whole
metadata table as max(start) / min(end). These keys are recorded per
(machine, trace) (Scope::kMachineAndTrace), so for a single trace this
yields that trace's active window, but for several traces loaded
together (a multi-machine session, or per-machine traces opened as a
group) it computes the intersection of their windows. When the windows
are disjoint the intersection inverts (max(start) > min(end)), and
building the returned TimeSpan trips its start <= end assertion, failing
the whole trace load with "Span start ... cannot be greater than
end ...".

Compute each (machine, trace)'s window and take their union instead, so
the visible window spans all loaded traces. Single-trace behavior is
unchanged.

### DIFF
--- a/ui/src/core/load_trace.ts
+++ b/ui/src/core/load_trace.ts
@@ -678,25 +678,32 @@ async function getTraceErrors(engine: Engine): Promise<number> {
 }
 
 async function getTracingMetadataTimeBounds(engine: Engine): Promise<TimeSpan> {
-  const queryRes = await engine.query(`select
-       name,
-       int_value as intValue
-       from metadata
-       where name = 'tracing_started_ns' or name = 'tracing_disabled_ns'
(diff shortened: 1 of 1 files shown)


LalitMaganti:
This is a hack not the correct fix. We should make getting the tracing bounds machine aware rather than working around like this: afaik we now store in the metadata table the start/end for each trace.

dreveman:
> This is a hack not the correct fix. We should make getting the tracing bounds machine aware rather than working around like this: afaik we now store in the metadata table the start/end for each trace.

Done

LalitMaganti (approved):

PR #6307, opened 2026-06-18
tp: add detailed import logging for debugging overlapping slices

LalitMaganti:


src/trace_processor/importers/common/slice_tracker.cc
```diff
@@ -368,7 +389,7 @@ bool SliceTracker::MaybeCloseStack(int64_t new_ts,
                                    int64_t new_dur,
                                    const SlicesStack& stack,
                                    TrackId track_id,
-                                   bool* overlap_out) {
+                                   std::optional<OverlapInfo>* overlap_out) {
```
safayat-google, line 392: nits, should we add some tests to verify the returned value of this information?
For example, I have a quick try with [this](https://paste.googleplex.com/4780207981592576#l=3) and got a pass result.

safayat-google (approved):
LGTM.

just a small nits

src/trace_processor/importers/common/slice_tracker.cc
LalitMaganti: Done thanks! Added diff tests as suggested

PR #6310, opened 2026-06-19
ui: normalize postMessage trace buffers to a pure ArrayBuffer

LalitMaganti:
Traces opened via postMessage() from other dashboards carry an untyped
messageEvent.data, and senders routinely pass a Uint8Array rather than a
pure ArrayBuffer. AppImpl.openTrace() used to convert such a buffer into
a pure ArrayBuffer (the invariant TraceBufferStream and cache_manager
rely on), but that guard was removed in #6160.


stevegolton:
What's the context here?

PR #6311, opened 2026-06-19
ui: normalize postMessage trace buffers to a pure ArrayBuffer

LalitMaganti:
Traces opened via postMessage() from other dashboards carry an untyped
messageEvent.data, and senders routinely pass a Uint8Array rather than a
pure ArrayBuffer. AppImpl.openTrace() used to convert such a buffer into
a pure ArrayBuffer (the invariant TraceBufferStream and cache_manager
rely on), but that guard was removed in #6160.

### DIFF
--- a/ui/src/frontend/post_message_handler.ts
+++ b/ui/src/frontend/post_message_handler.ts
@@ -21,6 +21,7 @@ import {AppImpl} from '../core/app_impl';
 import type {SerializedAppState} from '../core/state_serialization_schema';
 import {parseAppState} from '../core/state_serialization';
 import {BUCKET_NAME, isValidGcsFileName} from '../base/gcs_uploader';
+import {toArrayBuffer} from '../base/utils';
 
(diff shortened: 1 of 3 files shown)


stevegolton:
What's the backstory for this? Which dashboard started failing and in what way?

I'm slightly confused about how passing a view ever worked, seeing as they are not instances of ArrayBuffer.

LalitMaganti:
b/525429245 is the context.

ui/src/frontend/post_message_handler.ts
stevegolton, line 205: I think the internal bug sends a wrapped trace which makes a bad assertion that the buffer is an ArrayBuffer.

I don't think the fix below will fix it.

We need both.

stevegolton (changes requested):

stevegolton (approved):

ui/src/frontend/post_message_handler.ts
stevegolton: Never mind - I missed the normalisation in the sanitise function

PR #6312, opened 2026-06-19
tp: record all clock graph edges in the clock_snapshot table

LalitMaganti:
Move the clock_snapshot table row-writing loop out of
ProtoTraceReader::ParseClockSnapshot into ClockTracker so that every
edge added to the clock graph is recorded in the table at the moment
it enters it, including the snapshots added by the ftrace and
instruments tokenizers, remote clock syncs and the identity edges
injected by the deferred identity sync. 

Also:
- The deferred identity sync now triggers on "cannot reach the trace
  time clock" instead of "not present in the graph", matching its
  documented contract: a clock can be in the graph yet disconnected
  from trace time.
- ToHostTraceTime uses Find() instead of operator[], which was
  inserting a spurious zero entry into remote_clock_offsets on every
  lookup from a remote machine.


src/trace_processor/importers/common/clock_tracker.cc
```diff
@@ -40,7 +41,32 @@
...
+#include "protos/perfetto/common/builtin_clock.pbzero.h"
+
 namespace perfetto::trace_processor {
+namespace {
+
+std::optional<StringId> GetBuiltinClockNameOrNull(TraceStorage* storage,
+                                                  int64_t clock_id) {
+  switch (clock_id) {
+    case protos::pbzero::BUILTIN_CLOCK_REALTIME:
+      return storage->InternString("REALTIME");
```
sashwinbalaji: should we create StringIds ?

src/trace_processor/importers/common/clock_tracker.cc
```diff
@@ -74,7 +100,70 @@ base::StatusOr<uint32_t> ClockTracker::AddSnapshot(
...
+  ASSIGN_OR_RETURN(uint32_t snapshot_id,
+                   active_sync_->AddSnapshot(clock_timestamps));
+  AddSnapshotToTable(snapshot_id, clock_timestamps);
+  return snapshot_id;
+}
+
+void ClockTracker::AddSnapshotToTable(
+    uint32_t snapshot_id,
+    const std::vector<ClockTimestamp>& clock_timestamps) {
+  std::optional<int64_t> trace_time_from_snapshot =
```
sashwinbalaji: nit: compute lazily in if opt_trace_ts block ?

src/trace_processor/importers/common/clock_tracker.cc
```diff
@@ -74,7 +100,70 @@ base::StatusOr<uint32_t> ClockTracker::AddSnapshot(
...
+    // |absolute_timestamp|.
+    int64_t ts_to_convert =
+        clock_timestamp.clock.is_incremental ? 0 : clock_timestamp.timestamp;
+    // Even if we have trace time from snapshot, we still convert to optimise
+    // future conversions. Don't pass byte_offset since we expect failures
+    // here (e.g., non-monotonic clocks). Convert directly rather than via
+    // ToTraceTime: recording is bookkeeping, so it must not flush the
+    // deferred identity sync or count as an event conversion.
+    auto* state = context_->trace_time_state.get();
+    std::optional<int64_t> converted = active_sync_->Convert(
```
sashwinbalaji, line 120: nit: can we explicitly set the  /*suppress_errors=*/ arg here pls

sashwinbalaji (approved):

PR #6318, opened 2026-06-19
GN: Add default protobuf src dir to standalone proto_library

Jwata:
In standalone builds, the proto_library template did not automatically include the protobuf source directory in import_dirs. This forced targets to explicitly pass it.

This change adds it by default in the standalone template, allowing us to remove the redundant import_dirs from individual targets, matching Chromium proto_library's behavior.

Bug: http://crbug.com/521299828

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -3759,6 +3759,7 @@ filegroup {
 genrule {
     name: "perfetto_protos_perfetto_common_cpp_gen",
     srcs: [
+        ":libprotobuf-internal-descriptor-proto",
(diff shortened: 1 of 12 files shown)


LalitMaganti:
I'm not a huge fan of adding all the deps on descriptor proto in bazel. I worry that will lead to problems in Google3. Could we make this opt-in via a boolean or something instead of we didn't want to spread the actual import path directly everywhere? The dispatching logic in the GN rule could decide whether that means we need to add the edge conditionally or unconditionally depending on wherher we're in chromium or not.

PR #6319, opened 2026-06-19
proto: smaps: add field to discriminate which impl wrote the snapshot

rsavitski:
We want to take different import actions based on which implementation wrote the packet, to compenstate for bugs or changes in semantics. Introduce an explicit enum to serve as an indicator, which we could make repeated in the future if necessary.

Bug: 476126931


protos/perfetto/trace/profiling/smaps.proto
```diff
@@ -72,6 +73,26 @@ message PackedSmaps {
...
+  enum RecordingType {
+    RECORDING_TYPE_UNKNOWN = 0;
+    // Normal recording.
+    RECORDING_TYPE_STANDALONE = 1;
+    // Android: recorded in the perfetto_hprof ART plugin as part a managed heap
+    // dump. Recording is done inside of a forked child process, which means
+    // private/shared and pss fields are misleading as there is COW page sharing
+    // with the forked child.
+    RECORDING_TYPE_ART_HPROF_POSTFORK = 2;
+    // Android: recorded in the perfetto_hprof ART plugin as part a managed heap
```
sashwinbalaji: nits: 
 - as part *of* a (above also)
 - also can we just say copy-on-write just easier to read

sashwinbalaji (approved):

PR #6326, opened 2026-06-19
bt-ui: keep SettingsPage mounted across route changes

gignat-dev:
Previously, navigating away from the settings route would unmount the SettingsPage, and returning would remount it. This caused its trace-list grid to rebuild its data source and refetch /trace_metadata every time the settings page was opened.

Instead, keep the SettingsPage mounted in the DOM at all times and hide it using the `pf-bt-route-pane--hidden` class when not on the settings route. This preserves the loaded rows in the trace-list grid across route changes, avoiding unnecessary refetches.

### DIFF
--- a/ui/src/bigtrace/index.ts
+++ b/ui/src/bigtrace/index.ts
@@ -18,6 +18,7 @@ import '../assets/bigtrace.scss';
 import '../frontend/ui_main.scss';
 import m from 'mithril';
 import {defer} from '../base/deferred';
+import {Gate} from '../base/mithril_utils';
 import {reportError, addErrorHandler, type ErrorDetails} from '../base/logging';
 import {initLiveReload} from '../core/live_reload';
 import {settingsStorage} from './settings/settings_storage';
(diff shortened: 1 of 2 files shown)


ui/src/bigtrace/index.ts
```diff
@@ -229,7 +229,16 @@ class BigTraceRoot implements m.ClassComponent {
         {className: route === Routes.QUERY ? '' : 'pf-bt-route-pane--hidden'},
         m(QueryPage, {useBigtraceBackend: true}),
       ),
-      route === Routes.SETTINGS && m(SettingsPage),
+      // SettingsPage likewise stays mounted so the trace-list grid keeps its
+      // loaded rows across route changes instead of refetching on each visit.
+      m(
```
stevegolton: Could use a Gate widget here - which makes some of the other components downstream from this play ball (like GateDetector).

But I'm not going to block on this.

stevegolton (approved):

ui/src/bigtrace/index.ts
gignat-dev: Thanks for the advice. I switched to the Gate widget

PR #6329, opened 2026-06-19
tp: add perfetto_manifest clock overrides

LalitMaganti:
Allow overriding the clocks chosen by TP and their offsets and
instead manually specify what the offset should be.

### DIFF
--- a/src/trace_processor/plugins/perfetto_manifest/perfetto_manifest_reader.cc
+++ b/src/trace_processor/plugins/perfetto_manifest/perfetto_manifest_reader.cc
@@ -18,10 +18,14 @@
 
 #include <cinttypes>
 #include <cstdint>
+#include <cstdio>
+#include <limits>
 #include <optional>
 #include <string>
+#include <utility>
 
 #include "perfetto/base/status.h"
+#include "perfetto/base/time.h"
 #include "perfetto/ext/base/status_macros.h"
 #include "perfetto/ext/base/status_or.h"
 #include "perfetto/trace_processor/trace_blob_view.h"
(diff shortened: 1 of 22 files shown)


src/trace_processor/plugins/perfetto_manifest/perfetto_manifest_reader.cc
```diff
@@ -60,6 +121,118 @@ base::StatusOr<uint32_t> ParseClockName(const json::Dom& value) {
                          name.c_str());
 }
 
+base::StatusOr<ClockOverride> ParseAnchor(const json::Dom& anchor) {
+  if (!anchor.HasMember("ts") || !anchor["ts"].IsNumeric()) {
```
sashwinbalaji: Can we have a helper fucntion to abstract away has + is + setting logic for time (I think there are 3 such places in the file). Also, isnumeric will let double, should it be isintegral ?

sashwinbalaji (approved):

src/trace_processor/plugins/perfetto_manifest/perfetto_manifest_reader.cc
LalitMaganti: Done. Extracted a `ParseAnchorTs(obj, key, label)` helper that abstracts the has + integral-check + non-negative + read logic; it now covers the two anchor `ts` sites (anchor `ts` and `is.ts`). The third site, `offset_ns`, is structurally different (optional, and negatives are allowed and remapped) so it does not share the helper, but I switched its check to integral too.

Good catch on the double — added `Dom::IsIntegral()` (`IsInt() || IsUint()`) to json_value.h and used it everywhere a timestamp/offset must be an integer, so `1.5` is now rejected instead of silently truncated.

PR #6344, opened 2026-06-23
tp: speed up and shrink SliceTracker

LalitMaganti:
Rework SliceTracker's hot path to cut per-event overhead and code size
while keeping behaviour identical:

- Per-slice ArgsTracker is now lazily acquired from a pool and SliceInfo
  holds a pointer instead of an embedded tracker, so arg-less slices
  allocate nothing and the slice stack is trivially movable.
- Begin/Scoped/End are templated only on the args callback and call a
  single out-of-line entrypoint that returns the BoundInserter, so each
  caller inlines just the args lambda.
- depth/parent are folded into the row before insert and the track is
  resolved once per event.

Net: ~1.5x faster on the new SliceTracker microbenchmark and the
trace_processor binary shrinks ~10KB versus the std::function baseline.
The syscall unittest now drives a real SliceTracker instead of a mock.

### DIFF
--- a/src/trace_processor/importers/common/slice_tracker.h
+++ b/src/trace_processor/importers/common/slice_tracker.h
@@ -19,10 +19,14 @@
 
 #include <stdint.h>
 #include <cstdint>
+#include <deque>
 #include <functional>
 #include <optional>
+#include <type_traits>
+#include <utility>
 #include <vector>
 
+#include "perfetto/base/compiler.h"
 #include "perfetto/base/logging.h"
 #include "perfetto/ext/base/flat_hash_map.h"
 #include "src/trace_processor/importers/common/args_tracker.h"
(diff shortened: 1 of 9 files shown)


src/trace_processor/importers/common/slice_tracker.h
```diff
@@ -37,9 +41,13 @@ class TraceProcessorContext;
 
 class SliceTracker {
  public:
-  using SetArgsCallback = std::function<void(ArgsTracker::BoundInserter*)>;
   using OnSliceBeginCallback = std::function<void(TrackId, SliceId)>;
 
+  // Sentinel default args callback; ArgsPresent() compiles arg handling away.
```
sashwinbalaji: nit: no function name argspresent

src/trace_processor/importers/common/slice_tracker.h
```diff
@@ -177,38 +199,125 @@ class SliceTracker {
...
+                          TrackId track_id,
+                          StringId category,
+                          StringId raw_name,
+                          bool want_args,
+                          std::optional<OverlapInfo>* overlap_out);
+
+  // First half of End(): translate, find the slice and set its duration. The
+  // caller invokes args (if any) on the returned inserter, then calls
+  // CompleteSliceFinalize before any other op on the track.
+  EndedSlice CompleteSliceBegin(int64_t timestamp,
```
sashwinbalaji, line 236: nit: let's make the order of params consistwent b/w start and completeslice functions

sashwinbalaji (approved):

PR #6345, opened 2026-06-23
tp: memoize and unify string-like interned messages

LalitMaganti:
Many interned message types are "string-like": their iid is field 1 and
their string value is field 2. This covers InternedString) as well as
EventName, EventCategory, DebugAnnotationName, LogMessageBody,
HistogramName and AndroidJobName. Each call site re-implemented the same
lookup + typed-decode + (re-)intern dance, and the value-resolving ones
re-decoded and re-interned the same few strings on every event - on a
funcgraph trace that is millions of re-interns for ~540 names.

Add two helpers on PacketSequenceStateGeneration that decode field 2
directly with a plain ProtoDecoder (so they are independent of the proto
type and don't disturb the entry's typed decoder cache):
 - InternedStringId(field_id, iid) -> StringId, memoized on the interned
   entry via a new generic cached-value slot on InternedMessageView. The
   cache-hit path is inline; the cache lives with the interned entry (per
   sequence-state generation) so it is invalidated on incremental reset,
   and is stored as a raw uint32 to keep InternedMessageView free of any
   StringPool dependency.
 - InternedStringView(field_id, iid) -> base::StringView, for sites that
   need the raw string (to compare, concatenate, or hand to something
   that interns itself) rather than a StringId.

Convert all the string-like call sites across ftrace, track_event, gpu,
stack-profile, winscope and protolog importers to these helpers, dropping
a `LookupInternedMessage<Field, Type>` template instantiation and its
decode block at each. trace_processor shrinks ~6.3KB stripped, the
value-resolving sites get the memoization speedup (funcgraph
--no-ftrace-raw ~1.88s -> ~1.64s; a large track-event trace ~1.00s ->
~0.96s), and several sites that previously dereferenced the lookup result
with no null check become safe via the nullopt-returning helpers.

### DIFF
--- a/src/trace_processor/importers/ftrace/ftrace_parser.cc
+++ b/src/trace_processor/importers/ftrace/ftrace_parser.cc
@@ -71,6 +71,7 @@
 #include "src/trace_processor/types/trace_processor_context.h"
 #include "src/trace_processor/types/variadic.h"
 #include "src/trace_processor/types/version_number.h"
+#include "src/trace_processor/util/interned_message_view.h"
 
 #include "protos/perfetto/trace/ftrace/android_fs.pbzero.h"
 #include "protos/perfetto/trace/ftrace/bcl_exynos.pbzero.h"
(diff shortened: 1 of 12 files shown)


safayat-google (approved):
LGTM, Thank!

PR #6348, opened 2026-06-23
tp: memoize proto arg keys to avoid re-interning per field

LalitMaganti:
ProtoToArgsParser rebuilt and re-interned the flat_key/key strings of
every arg on every field of every message. The flat_key for a given
proto field path is invariant, so give the parser a StringPool and a
memo trie keyed by field number.

The Delegate API now takes already-interned StringIds instead of Key
strings, so delegates are pure value sinks and never re-intern keys.

A new proto_to_args_parser_benchmark isolates the key path and shows
~1.7-2x; end-to-end trace load gains are ~1-2% as key interning is a
small slice of ingestion.


src/trace_processor/util/proto_to_args_parser.cc
```diff
@@ -179,12 +181,67 @@ ProtoToArgsParser::Delegate::~Delegate() = default;
...
+
+std::optional<uint32_t> ProtoToArgsParser::ResolveTypeDescriptorIdx(
+    const FieldDescriptor& descriptor,
+    uint32_t node) {
+  if (node != kNoPath && path_nodes_[node].resolved_done) {
+    return path_nodes_[node].resolved_idx;
+  }
+  std::optional<uint32_t> idx =
+      pool_.FindDescriptorIdx(descriptor.resolved_type_name());
+  if (node != kNoPath) {
```
sashwinbalaji, line 231: even if idx not found, we consider resolved_done = true ? what if it's an extension to be added later.

sashwinbalaji (approved):

src/trace_processor/util/proto_to_args_parser.cc
LalitMaganti: Yes we do: having an extension come in the middle of the trace is not something I want to think about right now. That would open a whole can of worms around cache invalidation etc which is just too much of a pain to deal with.

PR #6349, opened 2026-06-23
Fix trace processor crashes with JIT traces

khokhlov962:
Fix SIGSEGV in JitCache::MoveCode when the code being moved is not found in the JIT cache.

### DIFF
--- a/src/trace_processor/importers/common/address_range.h
+++ b/src/trace_processor/importers/common/address_range.h
@@ -407,24 +407,14 @@ class AddressRangeMap {
   // is, there is a point so for which `AddressRange::Contains` returns true for
   // both the entry and the given `range'
   template <typename Callback>
-  void ForOverlaps(AddressRange range, Callback cb) {
-    if (range.empty()) {
-      return;
-    }
-    for (auto it = ranges_.upper_bound(range.start());
-         it != ranges_.end() && range.end() > it->first.start(); ++it) {
-      cb(*it);
-    }
-  }
-
-  template <typename Callback>
   void ForOverlaps(AddressRange range, Callback cb) const {
     if (range.empty()) {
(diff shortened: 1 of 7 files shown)


src/trace_processor/importers/common/address_range.h
```diff
@@ -417,6 +417,18 @@ class AddressRangeMap {
...
   }
 
+  template <typename Callback>
+  void ForOverlaps(AddressRange range, Callback cb) const {
+    if (range.empty()) {
+      return;
+    }
+    for (auto it = ranges_.upper_bound(range.start());
+         it != ranges_.end() && range.end() > it->first.start(); ++it) {
+      cb(*it);
```
LalitMaganti, line 416: nit: would be cleaner to have this as a break instead of putting in loop condition.

src/trace_processor/importers/common/address_range.h
```diff
@@ -417,6 +417,18 @@ class AddressRangeMap {
     }
   }
 
+  template <typename Callback>
+  void ForOverlaps(AddressRange range, Callback cb) const {
```
LalitMaganti: There appears to be another "ForOverlaps" function. Why do we need both?

khokhlov962: Yeah, we don't really need the non-const version, the const one is enough. Fixed.

src/trace_processor/importers/common/address_range.h
khokhlov962: Done

LalitMaganti (approved):

betasheet:
Alternative at #6368 -- Sorry, didn't realize there was a patch for this in flight already.

PR #6352, opened 2026-06-23
tp: add cross-file clock sync to perfetto_manifest

LalitMaganti:
A file's clocks block can point its reference at another file with
is.file or a named machine with is.machine, and the top-level trace_time
becomes a {file, clock} pointer so trace time can be a specific file's
clock. Together these let captures from separate machines line up on one
timeline.

To support this the manifest reader is now the single place that
establishes cross-machine identity, before any file is parsed:

- It pre-allocates a machine-table row for every declared machine and
  names it, so clock references resolve to real rows at parse time and
  the later fork reuses the row (MachineTracker looks it up).

- It adds each cross-machine clock edge straight to the global clock
  graph. Files are parsed sequentially, so an edge a later file declares
  must already exist when an earlier file converts; adding the edges in
  the reader guarantees that.

AddSnapshotToTable becomes a static helper so the reader, which has no
per-machine ClockTracker, can still record these edges in the
clock_snapshot table. Each row is now tagged with its clock's own
machine.


src/trace_processor/importers/common/clock_tracker.cc
```diff
@@ -114,84 +101,75 @@ base::StatusOr<uint32_t> ClockTracker::AddQualifiedSnapshot(
...
+    // expected (non-monotonic clocks) and must not flush deferred syncs or
+    // count as event conversions.
+    int64_t ts_to_convert = ct.clock.is_incremental ? 0 : ct.timestamp;
+    std::optional<int64_t> opt_trace_ts =
+        sync->Convert(ct.clock.id, ts_to_convert, trace_time);
     if (!opt_trace_ts) {
-      // This can happen if |AddSnapshot| failed to resolve this clock, e.g.
-      // if clock is not monotonic. Try to fetch trace time from snapshot.
-      if (!trace_time_from_snapshot_computed) {
-        trace_time_from_snapshot = ToTraceTimeFromSnapshot(clock_timestamps);
```
sashwinbalaji, line 149: don't see amy more callers of ToTraceTimeFromSnapshot, so lets remove the function

sashwinbalaji (approved):

PR #6358, opened 2026-06-24
ui: switch flamegraph to regexp and improve help

LalitMaganti:
I did a bunch of calculations and I could not find *any* difference in
performance so this should be quite safe.

Also the hint improvements should make it more intuitive as to what
exactly we do here.


ui/src/components/query_flamegraph.ts
stevegolton, line 1: Could we also add a note in the popup explaining that this now uses regex. I think the issue was partially about the fact that it's not clear what syntax can be used do describe the filter.

stevegolton (approved):

sashwinbalaji (commented):
nit: update pr title to be ui:

PR #6360, opened 2026-06-24
ui: add keyboard shortcuts to flamegraph node actions

LalitMaganti:
While a node's tooltip is showing (hovered or pinned), a single key runs
that node's matching action: Z zoom, F focus, P pivot, S/H show/hide
stack, M merge, C copy. Each key is shown as a keycap beside its menu
item. Keystrokes aimed at an editable element are left alone.

### DIFF
--- a/ui/src/plugins/com.android.HeapDumpExplorer/views/flamegraph_view.ts
+++ b/ui/src/plugins/com.android.HeapDumpExplorer/views/flamegraph_view.ts
@@ -143,6 +143,9 @@ function buildHeapGraphMetrics(
     isDominator: boolean,
   ): FlamegraphOptionalAction => ({
     name: 'Show objects from this class',
+    icon: 'data_object',
+    category: 'DRILL',
+    description: 'List the individual objects of this class.',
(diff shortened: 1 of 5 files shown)


LalitMaganti:
@stevegolton just was going through old PRs and was remidned of this again. Any thoughts on this?

stevegolton:
> @stevegolton just was going through old PRs and was remidned of this again. Any thoughts on this?

Keen on this as a feature but implementation-wise I think we should lean on the document's native focus() to direct keyboard events to the active element. This is something (I'd imagine) that we can to Popup natively, and combined with HotkeyContext, it should be rather elegant.

I'll look into this, as I'm not sure how well it'll integrate with the existing Popup component.

I think other libraries do exactly this - e.g. https://www.radix-ui.com/primitives/docs/components/dropdown-menu#keyboard-interactions - note that you can navigate up and down using up/down keys and the active element _is_ the popup.

Edit: While in the area we should probably implement keyboard nav for popup menus too, but this is trivial given the correct foundation with Popup.

PR #6363, opened 2026-06-24
tp: process_tracker: treat parent pid changes as reparenting

rsavitski:
Right now, the process_tracker always assumes that the same pid under a
different parent must mean that this is a new process reusing the pid, and
therefore creates a new process to track. There is another underlying
possibility - the process' parent has exited and the original process got
reparented. Note that the reparenting isn't always to pid=1 (init), since things
like PR_SET_CHILD_SUBREAPER exist.

In retrospect, this heuristic is the wrong way around. Reparenting is way more
frequent than pid reuse, and the latter can be cleanly detected via
sched_process_free. (Note: the reverse isn't true, because sched_process_exit is
when a thread becomes reparentable, and only the most recent kernels include the
group_dead tracepoint field.)

Hence this patch makes the process_tracker assume reparenting instead. We also
discard the new parent pid since the original is more informative.

PS: I have considered merging the SetProcessParent with SetProcessMetadata, but
decided to keep them separate for readability.

### DIFF
--- a/src/trace_processor/importers/common/process_tracker.cc
+++ b/src/trace_processor/importers/common/process_tracker.cc
@@ -26,6 +26,7 @@
 #include "perfetto/ext/base/string_view.h"
 #include "perfetto/public/compiler.h"
 #include "src/trace_processor/importers/common/args_tracker.h"
+#include "src/trace_processor/importers/common/import_logs_tracker.h"
 #include "src/trace_processor/importers/common/stats_tracker.h"
 #include "src/trace_processor/storage/stats.h"
 #include "src/trace_processor/storage/trace_storage.h"
@@ -132,6 +133,11 @@ void ProcessTracker::EndThread(int64_t timestamp, int64_t tid) {
 
   // If the process pid and thread tid are equal then, as is the main thread
   // of the process, we should also finish the process itself.
+  // Note: this works on linux because even if the leader thread exits before
+  // the rest of the thread-group, the kernel keeps the leader in a zombie state
+  // until the last group member exits. So sched_process_free for the main
+  // thread is deferred until the thread-group is fully dead (see
+  // delay_group_leader() in kernel sources).
   PERFETTO_DCHECK(*td.is_main_thread());
   ps.set_end_ts(timestamp);
   pids_.Erase(tid);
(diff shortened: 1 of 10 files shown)


src/trace_processor/importers/common/process_tracker.cc
```diff
@@ -136,6 +136,20 @@ void ProcessTracker::EndThread(int64_t timestamp, int64_t tid) {
...
   pids_.Erase(tid);
   InvalidateProcessThreads(*opt_upid);
+  exited_upids_.erase(*opt_upid);
+}
+
+void ProcessTracker::MarkProcessExited(int64_t tid) {
+  std::optional<UniqueTid> opt_utid = GetThreadOrNull(tid);
+  if (!opt_utid)
+    return;
+  auto td = (*context_->storage->mutable_thread_table())[*opt_utid];
```
LalitMaganti: nit: here and below: you don't need mutable_thrad/mutable process tables here, you can just use the non-mutable ones.

src/trace_processor/importers/common/process_tracker.h
```diff
@@ -385,6 +392,10 @@ class ProcessTracker {
   };
   base::FlatHashMap<UniquePid, SortIndex> process_sort_indexes_;
   base::FlatHashMap<UniqueTid, SortIndex> thread_sort_indexes_;
+
+  // Processes which have exited (sched_process_exit) but not yet been freed
+  // (sched_process_free).
+  base::FlatSet<UniquePid> exited_upids_;
```
LalitMaganti: FlatSet is *not* going to scale well over large traces with lots of short lived processes. Please use unordered_set of flat_hash_map with a boolean value instead.

src/trace_processor/importers/common/process_tracker.cc
```diff
@@ -437,16 +451,30 @@ UniquePid ProcessTracker::UpdateProcessWithParent(UniquePid upid,
...
+                         process_table[*prev_parent_upid].end_ts());
+  if (parent_changed && !prev_parent_dead) {
     upid = StartNewProcessInternal(std::nullopt, pupid, prr.pid(),
                                    kNullStringId, ThreadNamePriority::kOther,
                                    associate_main_thread);
+  } else if (parent_changed) {
+    // Reparenting case. Keep existing info since the original parent pid is
+    // more useful.
+    context_->stats_tracker->IncrementStats(
+        stats::process_tracker_reparented_not_reused);
```
LalitMaganti: Please use the new "import_log_tracker" APIs to increment stats instead of directly incrementing stats: it increases the debuggability of this sort of thing tremendously.

rsavitski:
@LalitMaganti PTAL, after some thinking, I decided to flip the heuristic instead. I think this is both simpler and the expected interpretation of the events.

src/trace_processor/importers/common/process_tracker.cc
```diff
@@ -433,24 +439,28 @@ void ProcessTracker::AssociateCreatedProcessToParentThread(
...
-    upid = StartNewProcessInternal(std::nullopt, pupid, prr.pid(),
-                                   kNullStringId, ThreadNamePriority::kOther,
-                                   associate_main_thread);
-  } else {
+  if (!prev_parent_upid) {
     prr.set_parent_upid(pupid);
+  } else if (*prev_parent_upid != pupid) {
+    if (timestamp) {
+      context_->import_logs_tracker->RecordParserError(
+          stats::process_tracker_parent_pid_changed, *timestamp);
```
LalitMaganti, line 458: Do we want to log these as errors? Seems a bit agressive? How about we make them info instead?

LalitMaganti (approved):

src/trace_processor/importers/common/process_tracker.cc
rsavitski: They are already kInfo (the severity is part of the `stats` key), I agree that the import_logs_tracker method names could be made more general.

PR #6366, opened 2026-06-25
tp/ui: only surface machines that received packets

dreveman:
When a producer connects to the in-process service with a non-default
machine id, only the producer's own packets carried that id. The
service-emitted packets (clock snapshot, trace config, trace uuid,
system info, trace provenance, lifecycle events, stats, extension
descriptors and triggers) were written without a machine id and so
attributed to the default host machine. As a result the trace contained
a separate host machine with no data alongside the real machine, and the
real machine had no service clock snapshot of its own.

Stamp the service's own packets with the machine id of the connected
in-process producer so a single-machine in-process trace contains
exactly that machine and its timestamps resolve against the service
clock snapshot. Host and relay (multi-machine) sessions are unaffected:
the stamping is gated on there being no relay clients and a non-default
local machine id, and the relay paths keep setting their own machine id.

trace_processor creates a synthetic host machine (machine_id 0) per
trace file even when every packet is attributed to another machine, so
the UI showed an extra empty machine and stamped tracks with noisy
"(machine N)" suffixes for traces that really have a single machine.

Add a packet_count column to the machine table, incremented as each
packet is attributed to its machine; a machine that never receives a
packet keeps a count of zero. The UI then labels tracks by machine only
when more than one machine has a non-zero packet count, and omits
zero-packet machines from the Machines tab.

Single-machine traces show no machine suffixes and no empty machine;
multi-machine traces are unaffected.

### DIFF
--- a/src/trace_processor/importers/proto/proto_trace_reader.cc
+++ b/src/trace_processor/importers/proto/proto_trace_reader.cc
@@ -44,6 +44,7 @@
 #include "src/trace_processor/importers/common/clock_tracker.h"
 #include "src/trace_processor/importers/common/event_tracker.h"
 #include "src/trace_processor/importers/common/import_logs_tracker.h"
+#include "src/trace_processor/importers/common/machine_tracker.h"
 #include "src/trace_processor/importers/common/metadata_tracker.h"
 #include "src/trace_processor/importers/common/parser_types.h"
 #include "src/trace_processor/importers/common/process_tracker.h"
@@ -256,23 +257,30 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   // Any compressed packets should have been handled by the tokenizer.
   PERFETTO_CHECK(!decoder.has_compressed_packets());
 
-  // When the trace packet is emitted from a remote machine, parse it with a
-  // separate ProtoTraceReader bound to that machine's context.
-  if (PERFETTO_UNLIKELY(decoder.machine_id())) {
-    RETURN_IF_ERROR(CheckManifestSingleMachine());
-    if (is_machine_dispatcher_) {
-      auto [it, inserted] =
-          machine_to_proto_readers_.Insert(decoder.machine_id(), nullptr);
-      if (PERFETTO_UNLIKELY(inserted)) {
-        RETURN_IF_ERROR(CreateRemoteMachineReader(decoder.machine_id(), it));
+  // The top-level reader dispatches packets from other machines to a
+  // per-machine reader; host and adopted-machine packets are parsed here.
+  if (is_machine_dispatcher_) {
+    const uint32_t machine_id = decoder.machine_id();
+    if (PERFETTO_UNLIKELY(machine_id)) {
+      // Resolve adoption once, on the first remote-machine packet.
+      if (PERFETTO_UNLIKELY(!adopted_machine_id_.has_value())) {
+        RETURN_IF_ERROR(ResolveAdoptedMachine(machine_id));
       }
-      return it->get()->ParsePacket(std::move(packet));
+      // Route packets from a non-adopted machine to their own reader.
+      if (machine_id != *adopted_machine_id_) {
+        auto [reader, inserted] =
+            machine_to_proto_readers_.Insert(machine_id, nullptr);
+        if (PERFETTO_UNLIKELY(inserted)) {
+          RETURN_IF_ERROR(CreateRemoteMachineReader(machine_id, reader));
+        }
+        return reader->get()->ParsePacket(std::move(packet));
+      }
+    } else if (!host_machine_used_ && !decoder.has_synchronization_marker()) {
+      // The host owns this packet, so it can no longer be adopted away.
+      host_machine_used_ = true;
     }
   }
-  // Assert that the packet is parsed using the right instance of reader: the
-  // top-level dispatcher handles host packets (no machine_id), and the
-  // sub-readers it forks handle their own machine's packets (machine_id set).
-  PERFETTO_DCHECK(decoder.has_machine_id() == !is_machine_dispatcher_);
+  PERFETTO_DCHECK(is_machine_dispatcher_ || decoder.has_machine_id());
 
   // Execute ProtoVM logic right after the "machine ID fork" as detailed in
   // go/perfetto-proto-vm.
@@ -741,6 +749,31 @@ base::Status ProtoTraceReader::ParseClockSnapshot(ConstBytes blob,
   if (!snapshot_id.ok()) {
     PERFETTO_ELOG("%s", snapshot_id.status().c_message());
   }
+  return base::OkStatus();
+}
+
+PERFETTO_NO_INLINE base::Status ProtoTraceReader::ResolveAdoptedMachine(
+    uint32_t machine_id) {
+  RETURN_IF_ERROR(CheckManifestSingleMachine());
+  // Adopt onto the host context only when the host has no data of its own, its
+  // row is still unclaimed (files in one session share it), the machine has no
+  // context yet, and the file is not a multi-machine manifest.
+  const bool host_has_data = host_machine_used_;
+  const bool host_row_claimed =
+      context_->machine_tracker->raw_machine_id() != 0;
+  const bool is_manifest_machine =
+      context_->trace_state && context_->trace_state->machine_remap;
+  const bool already_has_context =
+      context_->forked_context_state->trace_and_machine_to_context.Find(
+          std::make_pair(context_->trace_id().value,
+                         static_cast<int64_t>(machine_id))) != nullptr;
+  if (host_has_data || host_row_claimed || is_manifest_machine ||
+      already_has_context) {
+    adopted_machine_id_ = 0;
+    return base::OkStatus();
+  }
+  context_->machine_tracker->SetRawMachineId(machine_id);
+  adopted_machine_id_ = machine_id;
   return base::OkStatus();
 }
 
--- a/src/trace_processor/importers/proto/proto_trace_reader.h
+++ b/src/trace_processor/importers/proto/proto_trace_reader.h
@@ -116,6 +116,9 @@ class ProtoTraceReader : public ChunkedTraceReader {
 
   using ConstBytes = protozero::ConstBytes;
   base::Status ParsePacket(TraceBlobView);
+  // Cold path: on the first remote-machine packet, decides whether to adopt it
+  // onto the host context and sets adopted_machine_id_ (machine_id or 0).
+  base::Status ResolveAdoptedMachine(uint32_t machine_id);
   // Out-of-line cold path: forks the sub-reader for a remote machine.
   base::Status CreateRemoteMachineReader(
       uint32_t machine_id,
@@ -165,6 +168,11 @@ class ProtoTraceReader : public ChunkedTraceReader {
   base::FlatHashMap<uint32_t, std::unique_ptr<ProtoTraceReader>>
       machine_to_proto_readers_;
   const bool is_machine_dispatcher_;
+  // Machine adopted onto the host context (relabelled), or nullopt until the
+  // first remote-machine packet resolves it; 0 means no adoption.
+  std::optional<uint32_t> adopted_machine_id_;
+  // Whether the host machine has parsed a packet of its own (blocks adoption).
+  bool host_machine_used_ = false;
   ProtoVmIncrementalTracing protovm_;
 
   // Temporary. Currently trace packets do not have a timestamp, so the
--- a/src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
@@ -756,10 +756,23 @@ CREATE PERFETTO VIEW machine(
   -- Total system RAM in bytes.
   system_ram_bytes LONG,
   -- Total system RAM in gigabytes (rounded).
-  system_ram_gb LONG
+  system_ram_gb LONG,
+  -- 1-based index of this machine among the non-host machines, used by the UI
+  -- to label tracks (e.g. "(machine 1)"). The host machine (raw_id 0) gets 0,
+  -- so it is never labelled.
+  label_index LONG
 )
 AS
-SELECT * FROM __intrinsic_machine;
+SELECT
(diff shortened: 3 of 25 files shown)


dreveman:
@LalitMaganti The problem I have is that I want to collect in-process traces that can optionally be merged with other in-process traces from the same workload but the common case is that you only look at one trace or some traces on the same host and I don't want the "(machine N)" tagging noise in the UI for these cases.

These changes allow me to collect in-process traces with machine-id so they can be merged when needed but when loaded independently (or only merged with other traces on the same host) the UI is not tagging all tracks with the "(machine N)" suffix when not needed. The UI change has a global variable that needs to be removed but interested in feedback on the approach before fixing that.

LalitMaganti:
I understand the motivation of this change but it feels *really* ugly to me, especially the SDK side. I'll leave this to Primiano and Ashwin to comment on.

dreveman:
> I understand the motivation of this change but it feels _really_ ugly to me, especially the SDK side. I'll leave this to Primiano and Ashwin to comment on.

@LalitMaganti Yes, it's the least ugly solution I've been able to come up with so far.

I'm fine to abandon this for now. I can mitigate it on the producer side by avoiding to set a machine-id unless I can determined that we're part of a multi-machine workload and for those cases maybe it's good to see the (machine 1) label even if I'm just examining the work from a single machine. It will be an indication that more machines were part of the workload..

I understand historically why we have this limitation but it seems a bit weird that we can't collect a trace with machine-id set without having be associated with two hosts.

LalitMaganti:
> Yes, it's the least ugly solution I've been able to come up with so far.

How about on service startup, you're able to provide the machine id you want to associate all service generated packets to? Would that work? I don't know if there's a way to configure the service today but maybe we should add it?

dreveman:
> > Yes, it's the least ugly solution I've been able to come up with so far.
> 
> How about on service startup, you're able to provide the machine id you want to associate all service generated packets to? Would that work? I don't know if there's a way to configure the service today but maybe we should add it?

That should be what my SDK change here does. It uses the machine-id that I'm now able to pass to the service for service packets in addition to producer packets. Are you suggesting something else?

The TP/UI changes are also needed on top of that as TP requires a row to exists in the machine table early before we've had a chance to populate the table from the trace. Maybe we can avoid the latter with some careful surgery to the TP code.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -418,6 +418,13 @@ TracingServiceImpl::ConnectProducer(Producer* producer,
       smb_scraping_enabled));
   auto it_and_inserted = producers_.emplace(id, endpoint.get());
   PERFETTO_DCHECK(it_and_inserted.second);
+
+  // Remember the machine of an in-process producer so the service can attribute
+  // its own packets to that machine, giving a single-machine in-process trace
+  // no separate host machine.
+  if (in_process && client_identity.machine_id() != kDefaultMachineID)
+    local_machine_id_ = client_identity.machine_id();
```
LalitMaganti, line 430: I would phrase this differently. Just keep track of local_machine_id always and move the kdefault check to the maybe emit code b

LalitMaganti: Also why do we need to gate this to in_process specifically. I could see this being useful even in non process mode.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -3886,6 +3892,18 @@ bool TracingServiceImpl::SnapshotClocks(
...
 
+void TracingServiceImpl::SetServiceTracePacketHeader(
+    protos::pbzero::TracePacket* tp) {
+  tp->set_trusted_uid(static_cast<int32_t>(uid_));
+  tp->set_trusted_packet_sequence_id(kServicePacketSequenceID);
+  // In a single-machine in-process session, attribute the service's own packets
+  // to that machine so the trace doesn't contain a separate host machine. In a
+  // relay (multi-machine host) session, leave them on the host machine; the
+  // relay paths stamp their own machine_id.
+  if (relay_clients_.empty() && local_machine_id_ != kDefaultMachineID)
```
LalitMaganti: I'm not understanding why relay clients matter here - surely if local machine id is set, you take advantage of it.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -274,6 +275,19 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   // sub-readers it forks handle their own machine's packets (machine_id set).
   PERFETTO_DCHECK(decoder.has_machine_id() == !is_machine_dispatcher_);
 
+  // Count the packet against this reader's machine. Forwarded remote-machine
+  // packets returned above, so each packet is counted once, on the machine that
```
LalitMaganti: Really not a fan of this, this is just a hack. What we should do instead of "adopt" the machine once we see the first packet with it set at the root level. It shouldn't matter what the packet is etc.

ui/src/public/utils.ts
```diff
@@ -134,10 +134,26 @@ export async function getTimeSpanOfSelectionOrVisibleWindow(
...
 }
 
+// Whether the active trace has more than one machine that actually carries
+// data. Set once per trace load (see setTraceHasMultipleMachines) and read by
+// maybeMachineLabel: with a single populated machine (the common case, including
+// a single in-process distributed rank) we don't disambiguate tracks by machine
+// at all, so no "(machine N)" suffix is shown.
+//
+// TODO: remove this module-global before merging.
+let gTraceHasMultipleMachines = false;
```
LalitMaganti: Remove this?

src/tracing/service/tracing_service_impl.cc
dreveman: Done. We need the `in_process` guard to prevent host machines in a relay setup from using the remote machine_id  for internal packets.

src/tracing/service/tracing_service_impl.cc
dreveman: Yes, dropped it from here

ui/src/public/utils.ts
dreveman: I added the proper plumbing for num machines in latest version so we can limit machine labels to when more than one machine exists. Matches how we handle GPUs now.

src/trace_processor/importers/proto/proto_trace_reader.cc
dreveman: Ack. Replaced this with logic to adopt a machine once we see it.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -261,18 +262,46 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
...
+      // Adopt the first machine seen onto the host context by relabelling its
+      // (otherwise empty) machine row, so a single-machine trace has no
+      // leftover host machine. Adopt only while: the host context is still
+      // empty (it has handled no data of its own); the host row has not already
+      // been adopted by another file's reader (files merged in one session
+      // share the host row, so its raw id is non-zero once claimed); the
+      // machine has no context of its own yet (a pre-declared machine keeps its
+      // own row); and the file is not a multi-machine manifest (whose embedded
+      // ids are remapped). Otherwise a context is forked below, keeping
+      // machines distinct.
```
LalitMaganti: I'd like id this was broken into multiple variables to make this easier to read. This is too difficult to review rn.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -261,18 +262,46 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   if (PERFETTO_UNLIKELY(decoder.machine_id())) {
     RETURN_IF_ERROR(CheckManifestSingleMachine());
     if (is_machine_dispatcher_) {
-      auto [it, inserted] =
-          machine_to_proto_readers_.Insert(decoder.machine_id(), nullptr);
-      if (PERFETTO_UNLIKELY(inserted)) {
-        RETURN_IF_ERROR(CreateRemoteMachineReader(decoder.machine_id(), it));
+      // Adopt the first machine seen onto the host context by relabelling its
+      // (otherwise empty) machine row, so a single-machine trace has no
```
LalitMaganti: I suggest out of lining this honestly.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -261,18 +262,46 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
   if (PERFETTO_UNLIKELY(decoder.machine_id())) {
     RETURN_IF_ERROR(CheckManifestSingleMachine());
```
LalitMaganti: Given now we will expect to run this on e eey packet I think we need to refactor how often this function is called to have an early check + return here to avoid paying the cost of the status+return every time.

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -261,18 +262,46 @@ base::Status ProtoTraceReader::ParsePacket(TraceBlobView packet) {
...
+      // The adopted machine is handled by this (host) context; fall through.
     }
-  }
-  // Assert that the packet is parsed using the right instance of reader: the
-  // top-level dispatcher handles host packets (no machine_id), and the
-  // sub-readers it forks handle their own machine's packets (machine_id set).
-  PERFETTO_DCHECK(decoder.has_machine_id() == !is_machine_dispatcher_);
+  } else if (is_machine_dispatcher_ && !decoder.has_synchronization_marker()) {
+    // A packet with no machine id is data the host machine owns and blocks
+    // adoption, so a real multi-machine trace (where the host has its own data)
```
LalitMaganti: Seems weird this is an else if here honestly. Why is it not just a standalone if. Also we should only set this on the first packet right? 

I think in general this code needs to be properly refactored into code which only runs once and the fast path. Right now it's too confusing what is run once vs not. The slow path stuff should be outlined from this function as well

src/trace_processor/importers/proto/proto_trace_reader.h
```diff
@@ -165,6 +165,17 @@ class ProtoTraceReader : public ChunkedTraceReader {
   base::FlatHashMap<uint32_t, std::unique_ptr<ProtoTraceReader>>
       machine_to_proto_readers_;
   const bool is_machine_dispatcher_;
+  // Dispatcher-only state for adopting a single machine onto the host context.
```
LalitMaganti: not: here and throughout the patch, please simplify the comments. This feels like ai massively over explaining things which are pretty self explanatory.

src/tracing/service/tracing_service_impl.cc
LalitMaganti: I would think we should allow for the case of non-in process mode to have this injected in (e.g. from the command line). We should structure this code for that to be possible whdih right now I don't think it is.

src/trace_processor/importers/proto/proto_trace_reader.cc
dreveman: Done

src/trace_processor/importers/proto/proto_trace_reader.cc
dreveman: Done

src/trace_processor/importers/proto/proto_trace_reader.cc
dreveman: Done

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -744,6 +744,49 @@ base::Status ProtoTraceReader::ParseClockSnapshot(ConstBytes blob,
   return base::OkStatus();
 }
 
+PERFETTO_NO_INLINE base::Status ProtoTraceReader::RouteMachinePacket(
+    uint32_t machine_id,
+    TraceBlobView& packet,
+    bool* forwarded) {
+  *forwarded = false;
```
LalitMaganti: I think this should be put in the above. For multi-mahine traces, this is a hot path. We should aslo add proper UNLIKELY/LIKELY annotations

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
```diff
@@ -756,10 +756,27 @@ CREATE PERFETTO VIEW machine(
...
+  -- 1-based index of this machine among the non-host machines, used by the UI
+  -- to label tracks (e.g. "(machine 1)"). The host machine (raw_id 0) gets 0,
+  -- so it is never labelled.
+  label_index LONG
 )
 AS
-SELECT * FROM __intrinsic_machine;
+SELECT
+  m.*,
+  iif(
```
LalitMaganti, line 770: Why is this not just id + 1? id will always be zero. Not a huge fan on this.

src/trace_processor/importers/proto/proto_trace_reader.cc
dreveman: Done

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/views.sql
dreveman: Adjusted this code and made it clear why it can't be just `id +1`

LalitMaganti (commented):
LG from the trace processor side and general overall patch. Will leave to Steve and Ashwin for service and UI changes.

src/tracing/service/tracing_service_impl.cc
sashwinbalaji, line 1: In the file there are couple more places calling set_trusted_packet_sequence_id:
- we should also update the ones in MaybeEmitProtoVmInstances and onflushtimeout to call the new function ?
- for the one in emitsyncmarker and others pls update a small comment explaining why it can't call SetServiceTracePacketHeader so in future we don't mistakenly update that.

Also pls add couple of unittests in src/tracing/service/tracing_service_impl_unittest.cc

dreveman: Done. Added two tests and updated the places that should use this and added short comments next to places where we shouldn't use this.

stevegolton (approved):

sashwinbalaji (approved):

PR #6368, opened 2026-06-25
tp: Fix crash when JIT range is subset of mapping

betasheet:
MappingTracker had a CHECK that JIT range must contain the memory mapping range. This is incorrect as JIT ranges (e.g. V8 JIT code space) are typically subsets of larger OS memory mappings (e.g. V8 heap cage). This caused crashes when loading merged traces with V8 JIT and perf data.

This CL refactors JIT mapping to do on-demand lookup of the JIT cache for a specific PC, instead of associating the JIT cache with the whole mapping. This allows supporting multiple JIT caches inside a single large mapping, and avoids treating non-JIT frames in the large mapping as failed JIT frames (significantly reducing jit_unknown_frame stats).

Also added a unit test to cover this scenario.

[Agent-created patch]

Bug: b/526490256


LalitMaganti:
This appears to be doing the same as https://github.com/google/perfetto/pull/6349. Can you please confirm which change we want to land?

betasheet:
Ah, cool -- hadn't realized that Mika was looking at this already. Comparing the two changes, I have a slight preference for this one (incl. the tests); but probably @khokhlov962 should have a look, too.

carlscabgro:
@betasheet IIRC the reason we had JitCache to contain the full mapping was because frames (STACK_PROFILE_FRAME) are associated to one mapping (STACK_PROFILE_MAPPING). This mapping table is not so much about the actual mappings but more about where to get the debug info for a bunch of frames. Which is important for all the symbolization, and flamegraphs to work correctly. Don't recall all the details but having "different things" in the same mapping might get you in trouble when symbolizing. Although not sure you would ever have regular executable code in those mappings. Anyway just FYI.

betasheet:
> @betasheet Don't recall all the details but having "different things" in the same mapping might get you in trouble when symbolizing

@carlscabgro Thanks for the context -- I did check how the lookups seem to happen here, but AFAICT it shouldn't break any assumptions if we look up by address; i.e. AFAIU the logic here now tries to identify "which" JitCache is the right one for the address that needs to be symbolized (by looking at the address range within the mapping). Do you think this could still be a problem?

khokhlov962:
> Ah, cool -- hadn't realized that Mika was looking at this already. Comparing the two changes, I have a slight preference for this one (incl. the tests); but probably @khokhlov962 should have a look, too.

I also like this one more. Let's land this one, and I'll repurpose my PR to just handling wrong addresses in JitCache::MoveCode().

LalitMaganti (approved):

betasheet:
Rereading @carlscabgro 's comment, I think the concern was around mixing native code / symbols and jit ranges inside the same mapping. I agree, that may need to be looked at further (e.g. checking other accesses of mapping->is_jitted()).

For now, however, this change should address the crashes we observe when loading Chrome profiles (which are likely just because we have smaller or multiple JitCaches in the same mapping, not necessarily mixed native/jitted code). I'll go ahead and land. Thanks all!

carlscabgro:
> Rereading @carlscabgro 's comment, I think the concern was around mixing native code / symbols and jit ranges inside the same mapping. I agree, that may need to be looked at further (e.g. checking other accesses of mapping->is_jitted()).
> 
Yes exactly. As I said I don't remember all details but might be worth thinking about this at some point. But I agree for now this should be good enough

PR #6374, opened 2026-06-25
ui: add support for bundling even with dev server

LalitMaganti:


ui/build.mjs
```diff
@@ -89,6 +89,8 @@ const cfg = {
   bigtrace: false,
   engineBench: false,
   startHttpServer: false,
+  bundle: false,
```
stevegolton: Is there any point in putting this bundle field in the config object? looks like it's only used to work out the useHmr field, which can be done from args.bundle instead.

stevegolton (approved):

ui/build.mjs
LalitMaganti: Good point. Done.

PR #6376, opened 2026-06-25
tp: store android_framework_track_event enum columns as names

zezeozue:
trigger_type and hosting_type are enums; store the resolved value name instead of the raw int. Resolution goes through the trace-loaded descriptor pool the same way proto_to_args_parser::AddEnum does for args (FindEnumString), falling back to the integer as a string when the descriptor isn't loaded or the value is unknown.


src/trace_processor/plugins/android_framework_track_event/android_framework_track_event.cc
```diff
@@ -152,7 +159,24 @@ class Parser : public TrackEventExtensionParser {
         ts, static_cast<uint32_t>(evt.pid()));
   }
 
+  StringId InternEnum(std::optional<uint32_t>& cached_idx,
```
LalitMaganti: I feel like this is a common enough problem to make this a function directly on the pool.

src/trace_processor/plugins/android_framework_track_event/android_framework_track_event.cc
```diff
@@ -152,7 +159,24 @@ class Parser : public TrackEventExtensionParser {
...
+    std::optional<std::string> name;
+    if (cached_idx) {
+      name = pool.descriptors()[*cached_idx].FindEnumString(value);
+    }
+    return trace_context_->storage->InternString(
+        base::StringView(name ? *name : std::to_string(value)));
+  }
+
   TraceProcessorContext* trace_context_;
+  std::optional<uint32_t> trigger_type_idx_;
```
LalitMaganti: It would be nice if this uint32 was abstracted a bit with an opaque id struct so we didn't expose that it was an index directly.

src/trace_processor/util/descriptors.h
```diff
@@ -238,6 +238,23 @@ class DescriptorPool {
...
+   private:
+    friend class DescriptorPool;
+    std::optional<uint32_t> descriptor_idx_;
+  };
+
+  // Returns the name of enum value |value| within enum |enum_name|, or nullopt
+  // if the enum or the value is unknown. |cache| stores the resolved descriptor
+  // so later calls skip the by-name lookup.
+  std::optional<std::string> FindEnumString(CachedDescriptor& cache,
+                                            const std::string& enum_name,
```
LalitMaganti: nit: std::string_view.

LalitMaganti (approved):

PR #6378, opened 2026-06-25
ui: add multi-trace "at the same time" merge configurator

LalitMaganti:
This CL introduces the first part of the final shape of the merge
trace configuration tool in the UI. Right now we're focusing on
exposing things who owns the global clock and alignment between
them. Followup CLs will deal with multi-machine clocks and all
the problems that come from that

### DIFF
--- a/ui/src/frontend/views/sidebar.ts
+++ b/ui/src/frontend/views/sidebar.ts
@@ -213,7 +213,7 @@ function getCurrentTraceItems(trace: TraceImpl): SidebarMenuItemInternal[] {
     section: 'current_trace',
     sortOrder: 51,
     text: 'Download',
-    action: () => downloadTrace(trace),
+    action: async () => await downloadTrace(trace),
     icon: 'file_download',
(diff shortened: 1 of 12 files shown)


ui/src/frontend/views/sidebar.ts
```diff
@@ -213,7 +213,7 @@ function getCurrentTraceItems(trace: TraceImpl): SidebarMenuItemInternal[] {
     section: 'current_trace',
     sortOrder: 51,
     text: 'Download',
-    action: () => downloadTrace(trace),
+    action: async () => await downloadTrace(trace),
```
stevegolton, line 216: Is this required?

stevegolton (approved):

ui/src/frontend/views/sidebar.ts
LalitMaganti: I think it just makes it consistent with everything else in the file which does this. No harm in it so keeping it here.

PR #6388, opened 2026-06-26
ui: Fix issue where headless tracks interfere with pinTrackByRegex command

### LINKED ISSUE #6253: Setting up startup commands can not show Thread of InputDispatcher
if I use the example command 
```
[
  {
    "id": "dev.perfetto.PinTracksByRegex",
    "args": [".*CPU [0-3].*"]
  }
]
```
Perfetto UI can  show pin Track well.

But I modify the command show the InputDispatcher
```
[
  {
    "id": "dev.perfetto.PinTracksByRegex",
    "args": [".*InputDispatcher.*"]
  }
]
```

It will can't show the Pin Track.

Image

stevegolton:
Pinning a track with children only pins that track only creates a shallow copy of the track - it doesn't pin the children as well. When using the pinTrackByRegex command, if a headless track matches that regex, the current algorithm pins that headless track but stops recursing its children.

This PR simply ignores all headless tracks when finding matching tracks, but still recurses their children.

Also add an option to show headless tracks, which can help with debugging.

Fixes: https://github.com/google/perfetto/issues/6253


ui/src/core_plugins/dev.perfetto.Timeline/index.ts
```diff
@@ -50,12 +50,22 @@ export default class TimelinePlugin implements PerfettoPlugin {
   }
 
   async onTraceLoad(trace: TraceImpl): Promise<void> {
+    const showHeadlessTracksSetting = trace.settings.register({
```
LalitMaganti: Shouldn't this be a flag?

LalitMaganti (approved):

ui/src/core_plugins/dev.perfetto.Timeline/index.ts
stevegolton: Yep, changed to a flag.

PR #6393, opened 2026-06-26
rust-sdk: Fix GPU counters example

### LINKED ISSUE #6323: Bug: Rust gpu_counters.rs example doesn't include timestamp for TracePacket
I have tried the example and noticed it doesn't show any GPU counter values in the PerfettoUI - GPU descriptors tracks are shown. Most likely reason is a missing timestamp in `TracePacket`.

https://github.com/google/perfetto/blob/e29c9cb0c575916522d88a0f92eb511db542a6a1/contrib/rust-sdk/perfetto-protos-gpu/examples/gpu_counters.rs#L126-L154

NOTE: Is there an official API to get the time (e.g. CLOCK_BOOTTIME) via the RUST SDK? I was forced to use the glibc crate to get this time. Alternatively I could have registered a "GPU clock" and use it instead.

I can fix it, but it would require to include the glibc crate and I am not sure what the guidelines regarding external dependencies are. 

CC @dreveman - FYI, because it is your code. I have already talked with @LalitMaganti about it.

dreveman:
Trace packets were missing timestamps.

Fixes #6323

### DIFF
--- a/contrib/rust-sdk/perfetto-protos-gpu/examples/gpu_counters.rs
+++ b/contrib/rust-sdk/perfetto-protos-gpu/examples/gpu_counters.rs
@@ -124,6 +124,7 @@ fn main() -> Result<(), Box<dyn Error>> {
                 }
             };
             ctx.add_packet(|packet: &mut TracePacket| {
+                packet.set_timestamp(start_time.elapsed().as_nanos() as u64);
                 packet.set_gpu_counter_event(|event: &mut GpuCounterEvent| {
                     for i in COUNTER_IDS.iter() {
                         event.set_counters(|counter: &mut GpuCounterEventGpuCounter| {


LalitMaganti (approved):

SergejIsbrecht:
The doc of `Instant::now` clearly describes the value returned as `CLOCK_MONOTONIC` on Linux

```
UNIX	[clock_gettime](https://pubs.opengroup.org/onlinepubs/9799919799/functions/clock_getres.html) with CLOCK_MONOTONIC
```
(https://doc.rust-lang.org/std/time/struct.Instant.html) 

I was under the impression, that the CLOCK must be `CLOCK_BOOTTIME`. 

```
Field 	Type 	Description
timestamp 	uint64 	The timestamp of the TracePacket.
By default this timestamps refers to the trace clock (CLOCK_BOOTTIME on Android). It can be overridden using a different timestamp_clock_id.
The clock domain definition in ClockSnapshot can also override:
- The unit (default: 1ns).
- The absolute vs delta encoding (default: absolute timestamp).
```

```
Field 	Type 	Description
timestamp_clock_id 	uint32 	Specifies the ID of the clock used for the TracePacket timestamp. Can be one of the built-in types from ClockSnapshot::BuiltinClocks, or a producer-defined clock id.
If unspecified and if no default per-sequence value has been provided via TracePacketDefaults, it defaults to BuiltinClocks::BOOTTIME.
```
(https://perfetto.dev/docs/reference/trace-packet-proto)

@LalitMaganti ,
Is passing `CLOCK_MONOTONIC` also ok, or does it depend on `timestamp_clock_id`?

LalitMaganti:
On Linux, CLOCK_MONOTONIC and CLOCK_BOOTTIME are really the same but on Android they will be different. We should probably set timestamp_clock_id to disambiguate and just make it really clear.

dreveman:
@SergejIsbrecht, this example is using neither clock right now. It's just using a relative timestamp for simplicity. In real code you might want to use something like what I do here: https://github.com/dreveman/perfetto-gpu-compute-injection/blob/main/perfetto-gpu-probes/src/poller.rs#L24

Or if your code is limited to Linux you can make it simpler and always use CLOCK_BOOTTIME.

We can update this example to be closer to what is recommended in a real program. I just made a minimal change to have a trace collected from this example load for now.

PR #6396, opened 2026-06-26
tp: inline traceconv conversion into the shell subcommands

LalitMaganti:
The `convert` subcommand used to rebuild an argv and hand it to
TraceconvMain. Call the trace_to_text functions directly instead.

While doing this, split the symbolization modes out of `convert`, which is
now only about turning a trace into another artifact:

  - bundle becomes a top-level subcommand.
  - symbolize and deobfuscate move under a new `util` subcommand.
  - profile and java_heap_profile stay in convert.

convert_helpers holds the stdin/stdout setup and the text-proto-to-binary
path shared by convert and util. Output is unchanged from traceconv, except
firefox and decompress_packets now exit 0 on success (they used to return
the bool result directly, so success exited 1).

TraceconvMain and the standalone binary are removed later in the stack.


src/trace_processor/shell/bundle_subcommand.cc
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const char* BundleSubcommand::detailed_help() const {
+  return R"(Create a self-contained bundle from a trace.
+
+Outputs a TAR containing the trace plus the symbols and deobfuscation
+mappings needed to make it self-contained. Both <input> and <output> must be
+real file paths (stdin/stdout are not supported).
+
+Options:
```
safayat-google: nits, seems like `$ trace_processor_shell help bundle` prints some duplicates lines since these are already printed under `Subcommand flags:`. So, should we remove this for a cleaner message?

```
$ out/linux_clang_release/trace_processor_shell help bundle
...
Options:
  --symbol-paths PATH1,PATH2,...   Additional paths to search for symbols.
  --no-auto-symbol-paths           Disable automatic symbol path discovery.
  --proguard-map [pkg=]PATH        ProGuard/R8 mapping.txt for Java/Kotlin
                                   deobfuscation (may be repeated). The pkg=
                                   prefix scopes the map to a package.
  --no-auto-proguard-maps          Disable automatic ProGuard/R8 mapping
                                   discovery (e.g. Gradle project layout).
  --verbose                        Print more detailed output.

Subcommand flags:
      --symbol-paths             Additional paths to search for symbols.
      --no-auto-symbol-paths     Disable automatic symbol path discovery.
      --proguard-map             ProGuard/R8 mapping.txt for deobfuscation (may be repeated).
      --no-auto-proguard-maps    Disable automatic ProGuard/R8 mapping discovery.
      --verbose                  Print more detailed output.
...
```

safayat-google (approved):
LGTM.

Just a small nits on the help output format.

src/trace_processor/shell/bundle_subcommand.cc
LalitMaganti: Nice catch! Makes sense, removed the "options" section.

PR #6406, opened 2026-06-29
[state] Fix overly strict state track tokenization

etiennep-chromium:
This PR fixes tokenization error when getting out of order tracks (child before parent) for state tracks.
Regular tracks don't check this, so this aligns behavior for StateTrack.
This moves the check to parsing, which would translate to a track_event_parser_errors.

### DIFF
--- a/src/trace_processor/importers/proto/proto_trace_parser_impl_unittest.cc
+++ b/src/trace_processor/importers/proto/proto_trace_parser_impl_unittest.cc
@@ -32,6 +32,7 @@
 #include "perfetto/trace_processor/basic_types.h"
 #include "perfetto/trace_processor/trace_blob.h"
 #include "perfetto/trace_processor/trace_blob_view.h"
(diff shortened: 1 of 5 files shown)


LalitMaganti:
We should add some tests here.

etiennep-chromium:
Tests added!

src/trace_processor/importers/proto/proto_trace_parser_impl_unittest.cc
```diff
@@ -1718,6 +1722,99 @@ TEST_F(ProtoTraceParserTest, TrackEventWithResortedCounterDescriptor) {
   EXPECT_EQ(rr_0.thread_dur(), 10000);
 }
 
+TEST_F(ProtoTraceParserTest, TrackEventStateOnNonStateTrack) {
```
LalitMaganti: Please make some diff tests rather than adding unittests. These are lot harder to audit and were a historical mistake. We should not be adding more here.

LalitMaganti (approved):

PR #6407, opened 2026-06-29
[state] Add state track tables and account for it in _process_available_info_summary

etiennep-chromium:
This PR adds state.sql to define state_track, process_state_track and thread_state_track with tests.

This also fixes a UI crash when a process contains only state tracks;
_process_available_info_summary needs to include process_state_track to be considered non-empty.


LalitMaganti:
I don't see any change to _process_available_info_summary

etiennep-chromium:
> I don't see any change to _process_available_info_summary

I added a comment; it's a 1 line change.

LalitMaganti (approved):

PR #6410, opened 2026-06-29
tp: remove the standalone traceconv binary

LalitMaganti:
traceconv is now trace_processor_shell, so delete the binary:
src/traceconv/{main,traceconv}.cc, the public header, and the
executable/main/traceconv_lib targets. The conversion code
(src/traceconv:lib) stays and is linked by the shell. Drop traceconv from
the release plumbing too: the LUCI recipe, the GitHub packager and
gen_android_bp.

The UI only needs format conversion, and pulling the whole shell into the
WASM would drag in rpc/httpd that doesn't build under Emscripten. So the
WASM keeps a small conversion-only entry point (wasm_main.cc) over
src/traceconv:lib; the worker args and the WASM output name don't change.

The bundle CLI test moves to bundle_integrationtest.cc and drives
TraceProcessorShellMain.

The recipe expectation files still need `recipes.py test train`, which
needs cipd and couldn't be run here.


infra/luci/recipes/perfetto.py
```diff
@@ -43,9 +43,6 @@
     {
         'name': 'trace_processor_shell'
     },
-    {
-        'name': 'traceconv',
```
LalitMaganti, line 47: Do not merge until v57 goes out as we will be missing traceconv binary from the prebuilts if we do because of this change.

safayat-google (approved):
LGTM.

I think the migration is completely backward compatible. We can still update some of the documentation to use the trace processor shell command instead.

1. `docs/contributing/build-instructions.md`:
  Remove `traceconv` from the list of standalone build targets.
  Direct developers to build `trace_processor_shell` instead.

2. [Optional] `docs/quickstart/traceconv.md` (or `docs/toc.md`):
  Update command syntax tables from `traceconv` `MODE` to `trace_processor` `SUBCOMMAND`.
  Explain that `traceconv` functionality is now built directly into `trace_processor`.

3. [Optional] `docs/getting-started/linux-cookbook.md` & `docs/getting-started/cpu-profiling.md`:
  Update references to `traceconv` bundle and `traceconv` profile.

LalitMaganti:
> LGTM.
> 
> I think the migration is completely backward compatible. We can still update some of the documentation to use the trace processor shell command instead.
> 
> 1. `docs/contributing/build-instructions.md`:
>    Remove `traceconv` from the list of standalone build targets.
>    Direct developers to build `trace_processor_shell` instead.
> 2. [Optional] `docs/quickstart/traceconv.md` (or `docs/toc.md`):
>    Update command syntax tables from `traceconv` `MODE` to `trace_processor` `SUBCOMMAND`.
>    Explain that `traceconv` functionality is now built directly into `trace_processor`.
> 3. [Optional] `docs/getting-started/linux-cookbook.md` & `docs/getting-started/cpu-profiling.md`:
>    Update references to `traceconv` bundle and `traceconv` profile.

Yes I have some followup CLs which will take care of all this.

PR #6415, opened 2026-06-29
perfetto: point the traceconv wrapper at trace_processor_shell

LalitMaganti:
There is no traceconv binary to download anymore, so the wrapper should
fetch trace_processor_shell and run it under the traceconv name.

roll-prebuilts gains a source_tool override: the traceconv manifest keeps
its name and cached filename (so the on-disk traceconv-<sha> trips the
shell's name detection) but downloads the trace_processor_shell artifact.

The checked-in manifest still pins v56.1, whose traceconv artifact is still
available, so nothing breaks until the next roll regenerates it.


sashwinbalaji (approved):

tools/release/roll-prebuilts
```diff
@@ -129,14 +135,19 @@ ARCH_TO_PYTHON = {
 }
 
 
-def make_manifest(git_revision, tool, arch, file_name=None):
+def make_manifest(git_revision, tool, arch, file_name=None, source_tool=None):
```
LalitMaganti, line 138: Do not submit until v57 - we're not ready for this until v57 is stable.

Issue #5534, opened 2026-04-22
Public C++ library API for symbolizing/deobfuscating traces


LalitMaganti:
### Feature request

The `traceconv bundle` command does an opaque "take a trace, produce an
enriched trace" operation: it symbolizes native frames, applies ProGuard/R8
deobfuscation, and emits a TAR containing the enriched output.

The implementation already exists as a C++ entry point
(`TraceToBundle()` in `src/traceconv/trace_to_bundle.h`, with `EnrichTrace()`
in `src/trace_processor/util/trace_enrichment/trace_enrichment.h` doing the
actual work), but it lives under `src/` and is not part of the public API.

### What we'd like

An `include/` C++ API that takes:

- A path to (or bytes of) an input trace.
- A `BundleContext`-equivalent (explicit symbol paths, ProGuard mapping
  specs, flags).

...and returns the enriched trace opaquely — without the caller having to
instantiate a `TraceProcessor`, execute queries, or know anything about
`TracePacket` internals.

If you hit this, please +1 so we can prioritise.

songsen:
+1

jrothfeder:
+1

LalitMaganti:
Some thoughts on this issue now that I came across this in a different context:
1) Right now we require llvm-symbolizer and fork/exec it as a subprocess. While this works, this is somewhat unideal and if we can avoid it would be good
2) We could use llvm is a library but that's honestly quite painful because of how many LLVM things it pulls in
3) Alternatives to this might be https://github.com/getsentry/symbolic, https://github.com/libbpf/blazesym or https://docs.rs/wholesym/latest/wholesym/ each of which try and solve these problems with various advantages and disadvantages.

All in all I don't think there's an easy improvement on "just assume they have llvm-symbolizer" here: anything we do would require a bunch of effort and/or new dependencies, getting them compiling everywhere etc.

LalitMaganti:
Some more thoughts after offline discussions: one approach we should take is that we make the symbolizer an interface (it largely already is, we just need to make it part of the public API) and then users can plugin their own symbolizer themselves. We could ship with a llvm-symbolizer executed as a subprocess by default but we could potentially also support it as a shared library or linking against a static library installed in the system etc for others.

PR #6430, opened 2026-06-30
Fix macOS crash when embedding trace processor in multithreaded apps

arvind-google:
#### Problem:
    
When the Trace Processor is embedded as a library or daemon inside a larger, multithreaded host application
  (such as Android Studio), attempting to symbolize native traces crashes the host process on macOS.

  This happens because `subprocess_posix.cc` relies on `fork()` followed by `execvp()`. On macOS, calling
  `fork()` from a multithreaded process is highly dangerous. If any other threads are active, Apple's system
  library (`libsystem_c.dylib`) detects the unsafe `fork()` -> `exec()` sequence and immediately terminates the
  host process with a `SIGTRAP`. While the standalone, single-threaded `trace_processor_shell` avoids this fate,
  any multithreaded embedder invoking the symbolizer will crash.

#### Solution:
This patch migrates `subprocess_posix.cc` from using `fork()` + `execvp()` to the safer `posix_spawnp()`
  POSIX standard.

### DIFF
--- a/src/trace_processor/util/symbolizer/subprocess_posix.cc
+++ b/src/trace_processor/util/symbolizer/subprocess_posix.cc
@@ -23,6 +23,11 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_APPLE)
+#include <spawn.h>
+#include <crt_externs.h>
+#endif
+
 #include "perfetto/ext/base/utils.h"
 
 namespace perfetto {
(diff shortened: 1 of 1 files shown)


LalitMaganti:
There are multiple problems with this change:
1) You appear to have incorrectly branched off main (maybe you're using a canary/stable branch accidentally?) meaning that the PR's diff is very wrong
2) You are not gating this code on macOS specifically meaning you will change the behavior on Linux and any other POSIX distribution where spawn is not available
3) I'm very confused as to why you're even trying to do this at all: isn't the fix here that you invoke trace processor as a subprocess meaning you don't fact any problems at all? Right now there's not even a public way you *can* symbolize using trace processor as a library hence https://github.com/google/perfetto/issues/5534

All in all I think this fix is quite in the wrong direction and conceptually not what we should be doing.

LalitMaganti:
As discussed offline, we cannot have this patch upstream and this issue is also coming from Studio using internal APIs of Perfetto in the first place. If Studio wants to keep this patch downstream, up to you folks but I would advise either switching to traceconv/trace_processor bundle and/or use https://github.com/google/perfetto/issues/5534 when it's available.

PR #6443, opened 2026-06-30
ui: add summary of allocation/free information to profiling slices

LalitMaganti:
For ART profiles and native profiles, add the number of bytes
allocated/unreleased to the slice, making it far more useful for
people to understand what is going on.

This is what it looks like:

image


ui/src/plugins/dev.perfetto.HeapProfile/index.ts
```diff
@@ -160,14 +160,26 @@ export default class HeapProfilePlugin implements PerfettoPlugin {
...
-            heap_name
+            heap_name,
+            -- Per-dump deltas scoped to this slice only (NOT cumulative from
+            -- the start of the trace). The native profiler is delta-encoded,
+            -- so a dump's rows are the change since the previous dump: bytes
+            -- allocated in the interval, the net still held afterwards
+            -- (unreleased = allocs - frees), and the number of allocations.
+            -- ART allocation samples have no frees, so allocated == net.
+            SUM(iif(size > 0, size, 0)) AS summary_alloc_size,
+            SUM(size) AS summary_net_size,
```
ilkos: Consider the following scenario
- ts1, allocate ptr1 (size sz)
- ts2, allocate ptr2 (size sz), free ptr1 (size sz)
- etcetc

For the slice corresponding to ts2, you probably want net size to be sz, and not 0 (as in, net during this window of time)? I am not sure this is handled here (there's some code for this in heap_profile_details_panel.ts)

ilkos (approved):

PR #6447, opened 2026-06-30
ui: add a "gutter" at left/right of timeline for selecting edge events

LalitMaganti:
There's been a historic problem in the UI where it's difficult to select
things close to the start/end boundaries of the UI. This is especially
problematic with instant events (especially java heap dumps/callstack
samples) which are mostly off screen.

Fix this by adding a fixed 1% of the screen as a "gutter" allowing
visibility into those edge events but still disallowing other things like
area selections. The same gutter is applied to the minimap and the
default zoomed-out view so the trace edges are never flush.


stevegolton (approved):
Nice!

PR #6449, opened 2026-06-30
ui: warn on an incomplete flamegraph in the Heap Dump Explorer

zezeozue:
Extract the Heap Profile plugin's incomplete-heap-graph check and warning modal into a shared helper (incomplete_flamegraph.ts), and reuse it: opening the Heap Dump Explorer flamegraph tab now shows the same dismissible modal, so the user explicitly continues before viewing a partial graph. The check runs only when the flamegraph tab is active and re-arms per dump. Heap Profile is refactored to call the shared helper too, keeping one source of truth.


ui/src/plugins/com.android.HeapDumpExplorer/views/flamegraph_view.ts
```diff
@@ -182,6 +196,19 @@ const FlamegraphView: m.ClosureComponent<FlamegraphViewAttrs> = () => {
...
+      // Check for an incomplete (broken) heap graph once per dump, re-arming
+      // the warning modal when the dump changes.
+      if (incompleteKey !== key) {
+        incompleteKey = key;
+        incomplete = undefined;
+        modalDismissed = false;
+        const trace = attrs.trace;
+        incompleteLimiter.schedule(async () => {
+          incomplete = await isHeapGraphIncomplete(trace);
+          m.redraw();
```
stevegolton: You shouldn't need this

ui/src/plugins/com.android.HeapDumpExplorer/views/flamegraph_view.ts
```diff
@@ -182,6 +196,19 @@ const FlamegraphView: m.ClosureComponent<FlamegraphViewAttrs> = () => {
       }
       const metrics = cachedMetrics;
 
+      // Check for an incomplete (broken) heap graph once per dump, re-arming
+      // the warning modal when the dump changes.
+      if (incompleteKey !== key) {
```
stevegolton: Please use a queryslot for this.

Should be something like this:

```ts
const incompleteSlot = new QuerySlot<{isIncomplete: boolean, dismissed: boolean}>();

view() {
  const incomplete = incompleteSlot.use({
    key: incompleteKey,
    queryFn: async () => {
      return {
        isIncomplete: await isHeapGraphIncomplete(),
        dismissed: false,
      };
    },
  });

  if (incomplete.isIncomplete && !incomplete.modalDismissed) {
    ... render modal ....
  }
}
```

stevegolton (changes requested):

stevegolton (approved):

PR #6450, opened 2026-07-01
tp: avoid duplicate machine row for an adopted machine

dreveman:
When an embedded machine id is adopted onto the host row, register it in the per-machine context map so a later fork for the same id reuses that row instead of inserting a second machine. Without this a single machine could show up twice in the UI.


src/trace_processor/importers/proto/proto_trace_reader_unittest.cc
```diff
@@ -337,5 +337,30 @@ TEST_F(ProtoTraceReaderTest, DeferredClockSnapshotSequenceScopedClock) {
                    stats::clock_sync_failure_unknown_source_clock));
 }
 
+// An embedded machine id first seen on the host dispatcher is adopted onto the
```
LalitMaganti: Please add diff tests not unit tests here.

dreveman: Done

LalitMaganti (approved):

PR #6455, opened 2026-07-01
Make boot tracing trigger robust

tudormot:
Resolve a race condition where perfetto_trace_on_boot fails to start if the traced daemon is slow to initialize and set traced_started=1. By using a property trigger gated by "sys.boot_completed=", we ensure the trace starts as soon as traced is ready, even under heavy boot load.


tudormot:
@primiano as discussed, submitting a PR here. b/527760919 is the internal bug, not sure if it is to be added to the commit message

primiano (approved):

PR #6459, opened 2026-07-01
perfetto: add Zstd as a trace compression codec

sashwinbalaji:
The tracing service could already compress its output with deflate. This
adds Zstd, which produces smaller traces at a similar speed.

- Traces opt in with the new TraceConfig.compression field: set
  compression.zstd for Zstd or compression.deflate for deflate. If both
  are set, the higher field number wins (so zstd over deflate).
- Zstd's level is configurable via compression.zstd.level (1 fastest to 22
  smallest; 0 or unset uses zstd's default).
- The old compression_type field is now deprecated but still honored, so
  services that predate `compression` keep compressing.
- The service picks the codec from the config and runs it directly. A
  codec's library is only linked when its build flag (PERFETTO_ZLIB /
  PERFETTO_ZSTD) is on, so libz/libzstd stay out of builds that don't use
  compression, such as the SDK.
- Each codec is configured by its own proto message, so the service never
  needs to know a codec's specific options.
- Adding a codec later (e.g. lz4) is a new compressor file, one branch in
  the service, and a proto sub-message - nothing else changes.

Bug: 529304320


bazel/deps.bzl
```diff
@@ -97,6 +97,16 @@ def perfetto_deps():
         build_file = "//bazel:zlib.BUILD",
     )
 
+    # Zstd, for the legacy WORKSPACE path (bzlmod uses the BCR module, see
```
LalitMaganti, line 100: Should we turn down WORKSPACE builds at this point?

protos/perfetto/config/perfetto_config.proto
```diff
@@ -5428,13 +5428,40 @@ message TraceConfig {
...
 
-  // Compress trace with the given method. Best effort.
+  // Deprecated: use |compression| (CompressionConfig) instead. Legacy selector
+  // that only ever selects deflate; kept so services predating |compression|
+  // keep compressing. Best effort.
   enum CompressionType {
     COMPRESSION_TYPE_UNSPECIFIED = 0;
     COMPRESSION_TYPE_DEFLATE = 1;
   }
   optional CompressionType compression_type = 24;
```
LalitMaganti: Set deprecated = true annotation.

bazel/deps.bzl
primiano: I have no strong opinion TBH. if the CI is green I'd go for it.

I might be wrong but I think we did this back then for Stadia back in the days (R.I.P)

protos/perfetto/config/trace_config.proto
```diff
@@ -488,13 +488,40 @@ message TraceConfig {
   // with this key.
   optional string unique_session_name = 22;
 
-  // Compress trace with the given method. Best effort.
+  // Deprecated: use |compression| (CompressionConfig) instead. Legacy selector
```
primiano, line 490: Can you add: this is needed for backwards compatibility with Perfetto <= v56.

src/tracing/service/zstd_compressor.cc
```diff
@@ -0,0 +1,182 @@
...
+  // the object is unusable (PushPacket should not be called) and must be
+  // destroyed.
+  TracePacket Finish();
+
+ private:
+  void PushData(const void* data, uint32_t size);
+  void NewOutputSlice();
+  void PushCurSlice();
+
+  ZSTD_CStream* cstream_ = nullptr;
```
primiano: This could be a ScopedResource<ZSTD_CStream*, ZSTD_freeCStream, nullptr, false>, so you can avoid worrying about dtor

src/tracing/service/zstd_compressor.cc
```diff
@@ -0,0 +1,182 @@
...
+  // Points into `cur_slice_`. Zero-initialized so that the first compression
+  // call observes a full output buffer and allocates the initial slice.
+  ZSTD_outBuffer out_ = {/*dst=*/nullptr, /*size=*/0, /*pos=*/0};
```
primiano: nit: no need of the =

src/tracing/service/zstd_compressor.h
```diff
@@ -0,0 +1,41 @@
...
+namespace gen {
+class TraceConfig_CompressionConfig_Zstd;
+}  // namespace gen
+}  // namespace protos
+
+#if PERFETTO_BUILDFLAG(PERFETTO_ZSTD)
+// Compresses `packets` with zstd, tuned by `zstd`.
+void ZstdCompressFn(
+    std::vector<TracePacket>*,
+    const protos::gen::TraceConfig_CompressionConfig_Zstd& zstd);
```
primiano: tbh i'm a bit lukewarm on passing the whole proto here. I'd just pass level unless we can prove in future we need more stuff.
It's unnecessary coupling.
My approach on coupling things is: when in doubt, avoid coupling, unless you have a real reason to.
An `int level` here makes code more readable, maintainable, easier to get refactored (e.g. if we need to move this class in a common place to use in TP)

it's a small thing in this case, so not going to die on this hill. but then all these small things pile up one after the other in a codebase.

primiano (approved):
Great work really!
LGTM with a few minor comments

src/tracing/service/zstd_compressor.h
sashwinbalaji: I was also debating on the same thing. Went with the entire config:
1) to avoid tracing_service_impl knowing about the various knobs of zstd or other algos
2) I can easily see us adding block size next month.

But yeah maybe ok to remove now and in future if it groups bring the struct back.

PR #6461, opened 2026-07-01
tp: add zstd decompression support

sashwinbalaji:
This adds zstd across the decode paths so it can open the traces the
tracing service now compresses with zstd.

- Introduce a codec-agnostic StreamDecompressor with gzip and zstd
  implementations. The old gzip-only GzipTraceParser becomes a
  DecompressingTraceReader that works with any codec.
- Whole-file compressed traces are detected by magic and streamed
  through the matching decompressor.
- In-trace compressed bundles are decoded by the field that carries
  them: compressed_packets for deflate, zstd_compressed_packets for
  zstd.
- zstd is only linked when PERFETTO_ZSTD is enabled, matching how the
  service gates the codec.

Bug: 245929954
Bug: 529304320


src/tools/protoprofile/main.cc
```diff
@@ -205,27 +206,27 @@ int Main(int argc, const char** argv) {
...
+  // magic bytes and decompress it whole.
+  namespace util = trace_processor::util;
+  const auto* bytes = reinterpret_cast<const uint8_t*>(s.data());
+  util::CompressionType codec = trace_processor::CompressionTypeForTraceType(
+      trace_processor::GuessTraceType(bytes, s.size()));
+  if (codec != util::CompressionType::kNone) {
+    if (!util::IsCompressionSupported(codec)) {
       PERFETTO_ELOG(
-          "Input (%s) is gzip-compressed but this build lacks zlib support",
+          "Input (%s) is compressed but this build lacks the required codec",
```
LalitMaganti: We should output a better error message explain g what the codec is. And how to fix it ideally

LalitMaganti (approved):

PR #6465, opened 2026-07-01
protovm: add support for proto descriptors in the compiler

BogdanFloris:
adds support for descriptors in the protovm compiler to make it easier to use in a soong genrule.

### DIFF
--- a/src/protovm/compiler/main.cc
+++ b/src/protovm/compiler/main.cc
@@ -76,19 +76,21 @@ int main(int argc, char** argv) {
   static const option long_options[] = {
       {"help", no_argument, nullptr, 'h'},
       {"proto_path", required_argument, nullptr, 'I'},
+      {"descriptor", required_argument, nullptr, 'd'},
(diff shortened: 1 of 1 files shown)


primiano (approved):

src/protovm/compiler/main.cc
```diff
@@ -98,22 +100,38 @@ int main(int argc, char** argv) {
...
+  if (!descriptor_files.empty()) {
+    for (const auto& file : descriptor_files) {
+      std::string content;
+      if (!perfetto::base::ReadFile(file, &content)) {
+        PERFETTO_ELOG("Failed to read descriptor file %s", file.c_str());
+        return 1;
+      }
+      descriptors += content;
+    }
+  } else {
```
rukkal: Ideally we should support both at the same time (proto files + descriptors) in the same invokation, no?

BogdanFloris: yeah good shout, updated the PR to support both.

rukkal (approved):

PR #6471, opened 2026-07-01
tp: Fix LogMaxDepthExceeded build failure in debug builds

safayat-google:
In debug builds (-DPERFETTO_IMPLEMENTATION / is_debug=true), PERFETTO_DFATAL expands to PERFETTO_FATAL which does not return. This caused clang's -Wmissing-noreturn compiler check to fail on LogMaxDepthExceeded.

Fix this by replacing PERFETTO_DFATAL/PERFETTO_DLOG with RecordParserError, recording the new stat key
slice_max_depth_exceeded along with the parent and current slice names as error arguments.

Bug: 530244645

### DIFF
--- a/src/trace_processor/importers/common/slice_tracker.cc
+++ b/src/trace_processor/importers/common/slice_tracker.cc
@@ -33,9 +33,6 @@
 #include "src/trace_processor/types/variadic.h"
 
 namespace perfetto::trace_processor {
-namespace {
-constexpr uint32_t kMaxDepth = 512;
-}
 
 SliceTracker::SliceTracker(TraceProcessorContext* context)
     : legacy_unnestable_begin_count_string_id_(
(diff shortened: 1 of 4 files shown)


src/trace_processor/importers/common/slice_tracker.cc
```diff
@@ -93,15 +93,24 @@ bool SliceTracker::PrepareStartSlice(TrackInfo& track_info,
...
-      context_->storage->GetString(name.is_null() ? kNullStringId : name);
-  PERFETTO_DLOG("Last slice: %s", parent_name.c_str());
-  PERFETTO_DLOG("Current slice: %s", current_name.c_str());
-  PERFETTO_DFATAL("Slices with too large depth found.");
+  StringId parent_name_id =
+      parent.row.ToRowReference(slices).name().value_or(kNullStringId);
+  StringId current_name_id = name.is_null() ? kNullStringId : name;
+
+  StringId parent_key = context_->storage->InternString("parent_slice_name");
+  StringId current_key = context_->storage->InternString("current_slice_name");
```
sashwinbalaji: we should not intern on each call and instead do it once in the constructor.

Follow:

```
  // Interned arg-name keys for the overlap import logs (see AddOverlapArgs).
  const StringId overlap_start_key_;
  const StringId overlap_end_key_;
  const StringId overlap_conflicting_name_key_;
  const StringId overlap_conflicting_ts_key_;
  const StringId overlap_conflicting_dur_key_;
```

LalitMaganti (approved):

src/trace_processor/importers/common/slice_tracker.cc
safayat-google: Done, thanks.

PR #6472, opened 2026-07-01
tp: expand (flags_enum)-annotated fields into per-flag args

zezeozue:
Adds DescriptorPool::FlagSetToViews, which expands a flag bitmask into the names of its set flags, appended as string_views into the pool's enum-name storage. Composite (multi-bit) values are skipped; the set bits with no single-bit name are returned to the caller.

A proto int field can carry a (flags_enum) FieldOptions annotation naming its flags enum. proto_to_args_parser reads it and emits one string arg per set flag via FlagSetToViews, instead of the raw int. Living in the args funnel, it covers base fields and both extension surfaces with no per-parser code. The annotation is surfaced on FieldDescriptor, extracted from the field options during descriptor finalization.


src/trace_processor/util/descriptors.cc
```diff
@@ -499,6 +504,14 @@ base::Status DescriptorPool::AddFromFileDescriptorSet(
         continue;
       }
       ResolveUninterpretedOption(descriptor, field, *field.mutable_options());
+      // Pull out the (flags_enum) annotation, now resolved to its field number.
+      protozero::ProtoDecoder opt(field.options().data(),
```
LalitMaganti: There's already a decoder in ResolveUninterpretedOption. No point duplicating. You can also use FindField instead of a ReadField loop.

src/trace_processor/util/descriptors.cc
```diff
@@ -40,6 +40,11 @@
 
 namespace perfetto::trace_processor {
 namespace {
+
+// FieldOptions extension field number for the (flags_enum) annotation: names
+// the flags enum whose bits a field's bitmask should be expanded with.
+constexpr uint32_t kFlagsEnumFieldOptionNumber = 60002;
```
LalitMaganti: Where is this coming from? If this is something we're hardcoding in descriptors.cc it must be some protobuf thing or something we're making a standardised thing in protozero. Otherwise, it shouldn't be here.

src/trace_processor/util/descriptors.h
```diff
@@ -256,6 +264,15 @@ class DescriptorPool {
                                             std::string_view enum_name,
                                             int32_t value) const;
 
+  // For each single-bit flag set in |mask| for the flags enum |enum_name|,
+  // appends its name into |*out| as a string_view into the pool's own storage.
+  // Composite values are ignored. Returns the set bits with no single-bit name
+  // (|mask| if |enum_name| is unknown). |cache| memoises the resolved enum.
+  int64_t FlagSetToViews(CachedDescriptor& cache,
```
LalitMaganti: This is a *weird* return value I'll be honest.

src/trace_processor/util/descriptors.h
```diff
@@ -51,6 +51,9 @@ class FieldDescriptor {
   uint32_t type() const { return type_; }
   const std::string& raw_type_name() const { return raw_type_name_; }
   const std::string& resolved_type_name() const { return resolved_type_name_; }
+  // Full name of the flags enum this field's bitmask should be expanded with,
```
LalitMaganti: Please cleanup comments like this. They don't add anything nor do they fit in with the style.

src/trace_processor/util/descriptors.h
```diff
@@ -256,6 +264,15 @@ class DescriptorPool {
                                             std::string_view enum_name,
                                             int32_t value) const;
 
+  // For each single-bit flag set in |mask| for the flags enum |enum_name|,
```
LalitMaganti: As above.

src/trace_processor/util/proto_to_args_parser.cc
```diff
@@ -752,6 +765,25 @@ base::Status ProtoToArgsParser::AddEnum(const FieldDescriptor& descriptor,
   return base::OkStatus();
 }
 
+base::Status ProtoToArgsParser::AddFlags(const FieldDescriptor& descriptor,
+                                         int64_t value,
+                                         Delegate& delegate) {
+  flag_views_.clear();
+  DescriptorPool::CachedDescriptor cache;
```
LalitMaganti: Uhhhh there's no point having a cached descriptor only to *immediately* throw it away.

src/trace_processor/util/descriptors.h
```diff
@@ -68,6 +71,10 @@ class FieldDescriptor {
     resolved_type_name_ = resolved_type_name;
   }
 
+  void set_flags_enum(const std::string& flags_enum) {
```
LalitMaganti: IMO I don't see any reason why you could not resolve it to an index right here.

LalitMaganti (commented):
I'm really not a fan of how this CL is being done. I feel like this is just taking weird shortcuts and design decisions and it's not coherent at all for code which is quite core to TP. If AI was used here, I don't think it's done a good job at all.

src/trace_processor/util/descriptors.h
zezeozue: It doesn't seem that weird to me. Its jut returning the unmatched bits. What alternative do you recommend?

src/trace_processor/util/descriptors.cc
```diff
@@ -520,14 +520,31 @@ base::Status DescriptorPool::AddFromFileDescriptorSet(
...
+    }
+  }
   for (ProtoDescriptor& descriptor : descriptors_) {
     for (auto& entry : *descriptor.mutable_fields()) {
       FieldDescriptor& field = entry.second;
       if (field.options().empty()) {
         continue;
       }
       ResolveUninterpretedOption(descriptor, field, *field.mutable_options());
+      if (flags_enum_option_number) {
```
LalitMaganti, line 538: Please refactor into a separate function.

src/trace_processor/util/descriptors.h
LalitMaganti: Any solution I come up with seems worse so fine.

LalitMaganti (approved):

PR #6479, opened 2026-07-02
tp: Strip non-canonical elements from Java stack frame names

lucigrigo:
When ingesting stack frames via the `art_process_metadata` packet, drop the non-canonical elements (such as return types and argument signatures) from the Java method name. 

This ensures that stack frame naming behavior is consistent across all collection mechanisms, matching the formatting used by the out-of-process unwinder.


src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
```diff
@@ -38,6 +38,18 @@ using protos::pbzero::TracePacket;
 
 namespace {
 
+base::StringView StripJavaSignature(base::StringView signature) {
```
ilkos: nit: I'd probably name this ExtractMethodName(base::StringView frame)

ilkos (approved):

src/trace_processor/plugins/art_process_metadata_importer/art_process_metadata_module.cc
lucigrigo: done

PR #6487, opened 2026-07-02
Fix AddressRange creation in JitCache::MoveCode

khokhlov962:
JitCache::MoveCode was incorrectly constructing AddressRange by passing
the size of the old range as the 'end' argument to the AddressRange constructor.
This CL fixes it by using AddressRange::FromStartAndSize to correctly
construct the range.


khokhlov962:
Friendly ping :)

LalitMaganti (approved):

PR #6488, opened 2026-07-02
proto: reorder fw/native trace_packet message

sashwinbalaji:
Move FrameTimelineEvent and EvdevEvent definitions above the FrameworksNativeTracePacket extension wrapper so every message is declared before it is used. Removes the forward reference and keeps the extend block at the bottom, matching the other OOT extension files.


LalitMaganti:
Shouldn't this be in Android repo?

safayat-google (approved):

sashwinbalaji:
> Shouldn't this be in Android repo? 

This is the third party mirror but in android it's fine http://ag/40356088

PR #6490, opened 2026-07-02
tp: ui: warn about dropped events before opening merged traces

LalitMaganti:
The multi-trace merge dialog's alignment dry-run only summed the
clock-sync drop stats, so a fixed offset which moves events before the
start of the merged timeline reported "All traces line up" and then
silently dropped those events at load, surfacing only as a post-load
data-loss badge.

Count trace_sorter_negative_timestamp_dropped in the dry-run and make
TraceSorter record it even in tokenize-only mode, which the dry-run
uses. Also generalize the dialog's warning copy and mention the merge
offset cause in the stat's description.

### DIFF
--- a/src/trace_processor/sorter/trace_sorter.h
+++ b/src/trace_processor/sorter/trace_sorter.h
@@ -364,12 +364,15 @@ class TraceSorter::Stream {
  public:
   // Public API for tokenizers.
   void Push(int64_t ts, T data) {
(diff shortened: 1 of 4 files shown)


LalitMaganti:
Superseded by #6491 (same change; the commit was amended and lost its spr commit-id, so spr opened a fresh PR).

PR #6504, opened 2026-07-03
ui: Google drive integration (Googlers only)

stevegolton:
Adds the following features for Googlers only (for now):

- Open a trace from google drive.
- Save a trace to google drive, share it with specific people/groups (modify ACLs), and generate a share link.
- Allow opening pftrace files in Perfetto directly from drive (after initial configuration).

The final point takes a bit of explanation. After connecting perfetto to drive for the first time - which happens after the first time the user opens or saves a trace to drive - perfetto installs itself as an app that can open pftrace files (.trace, .pftrace, .perfetto-trace, + .gz variants). These files then get an 'Open in perfetto' button when viewing them on the drive.


stevegolton:
Superseded by https://github.com/google/perfetto/pull/7188

PR #6512, opened 2026-07-03
ui: Add Intelletto assistant plugin

stevegolton:
Adds dev.perfetto.Intelletto, the conversational assistant that sits in the sidebar and uses the LLM gateway plugin. It owns the agent loop (multi-step tool use with lazy tool loading), the trace-scoped chat session and context registry, the core query/schema tools, and the sidebar chat panel.

The agent drives the gateway directly: it picks the first configured model advertising the 'agentic' role and streams a turn via gateway.createStream(). The chat panel header shows that active model.

RFC: https://github.com/google/perfetto/discussions/6240

Note: Plugin is not enabled while in development - so for testing, enable `dev.perfetto.Intelletto` and open the sidebar.

Tools:

1. run_query — run PerfettoSQL, return rows as JSON
2. get_schema — list tables/views or a table's columns
3. get_selection — read the current UI selection
4. show_query — open a query in the Query page (mutating)
5. navigate — switch to a route, e.g. /viewer (mutating)
6. list_tools / more_tools — lazy-loading meta-tools (off by default)

Context injections:

1. page — current page route
2. viewport — visible time range (start/end ns)
3. selection — current selection, if any

### DIFF
--- a/ui/src/core/app_impl.ts
+++ b/ui/src/core/app_impl.ts
@@ -16,7 +16,7 @@ import {AsyncLimiter} from '../base/async_limiter';
 import {defer} from '../base/deferred';
 import {ensureExists, assertTrue} from '../base/assert';
 import {ServiceWorkerController} from '../frontend/service_worker_controller';
-import type {App} from '../public/app';
+import type {App, Route} from '../public/app';
 import type {SqlPackage} from '../public/extra_sql_packages';
 import type {FeatureFlagManager, FlagSettings} from '../public/feature_flag';
 import type {Raf} from '../public/raf';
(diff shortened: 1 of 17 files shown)


ui/src/core/embedder/default_plugins.ts
```diff
@@ -73,6 +73,7 @@ export const defaultPlugins = [
   'dev.perfetto.JournaldLog',
   'dev.perfetto.HeapProfile',
   'dev.perfetto.InstrumentsSamplesProfile',
+  'dev.perfetto.Intelletto',
```
LalitMaganti: Do we want to enable this by default?

LalitMaganti (approved):

ui/src/core/embedder/default_plugins.ts
stevegolton: Good catch - forgot to push the latest changes

PR #6517, opened 2026-07-03
trace_processor: reject overflowing offset+length in TraceBlobViewReader::SliceOff

adilburaksen:
**trace_processor: reject overflowing offset+length in TraceBlobViewReader::SliceOff**

`TraceBlobViewReader::SliceOffImpl` bounds-checks a requested slice with `offset + length <= ...` (fast path)
and `offset + length > end_offset_`, computing `offset + length` in `size_t` without checking for overflow.

A caller can pass an attacker-controlled `offset`. For example the perf.data importer reads each attrs
`ids` section as a `PerfFile::Section {uint64 offset, uint64 size}` directly from the file and validates only
that `ids.size` is a multiple of 8 (`perf_data_tokenizer.cc`), then calls `SliceOff(ids.offset, ids.size)`.
With `offset + length` wrapping past `2^64`, the sum becomes a small value that spuriously passes the
fast-path check. `SliceOff` then returns a `TraceBlobView` pointing outside the backing buffer
(`TraceBlobView::slice_off` only `DCHECK`s, which is compiled out in release builds), and the subsequent
`Reader::ReadVector`/read performs an out-of-bounds heap access.

Reject `offset`/`length` pairs whose sum overflows, before the bounds checks. This fixes the issue for every
importer that reaches `SliceOff` with a file-controlled offset, not just perf.data.

### DIFF
--- a/src/trace_processor/importers/perf/perf_data_tokenizer.cc
+++ b/src/trace_processor/importers/perf/perf_data_tokenizer.cc
@@ -21,6 +21,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
+#include <limits>
 #include <memory>
 #include <optional>
 #include <string>
@@ -219,9 +220,9 @@ PerfDataTokenizer::ParseHeader() {
 static base::Status CheckSectionInBounds(uint64_t offset,
                                          uint64_t size,
                                          const char* name) {
-  if (offset + size < offset) {
+  if (size > std::numeric_limits<uint64_t>::max() - offset) {
     return base::ErrStatus("Invalid %s section: offset (%" PRIu64
-                           ") + size (%" PRIu64 ") overflows",
+                           ") + size (%" PRIu64 ") exceeds uint64 max",
                            name, offset, size);
   }
   return base::OkStatus();
--- a/src/trace_processor/importers/perf/perf_data_tokenizer.cc
+++ b/src/trace_processor/importers/perf/perf_data_tokenizer.cc
@@ -213,8 +213,24 @@ PerfDataTokenizer::ParseHeader() {
   return ParsingResult::kSuccess;
 }
 
+// Rejects a section whose attacker-controlled `offset + size` overflows. Such a
+// range can never be satisfied by more data, so it must reject the trace rather
+// than be treated as "more data needed" (an empty/absent slice).
+static base::Status CheckSectionInBounds(uint64_t offset,
+                                         uint64_t size,
+                                         const char* name) {
+  if (offset + size < offset) {
+    return base::ErrStatus("Invalid %s section: offset (%" PRIu64
+                           ") + size (%" PRIu64 ") overflows",
+                           name, offset, size);
+  }
+  return base::OkStatus();
+}
+
 base::StatusOr<PerfDataTokenizer::ParsingResult>
 PerfDataTokenizer::ParseAttrs() {
+  RETURN_IF_ERROR(CheckSectionInBounds(header_.attrs.offset,
+                                       header_.attrs.size, "attrs"));
   std::optional<TraceBlobView> tbv =
       buffer_.SliceOff(header_.attrs.offset, header_.attrs.size);
   if (!tbv) {
@@ -233,6 +249,8 @@ PerfDataTokenizer::ParseAttrs() {
       return base::ErrStatus("Invalid id section size: %" PRIu64,
                              entry.ids.size);
     }
+    RETURN_IF_ERROR(
+        CheckSectionInBounds(entry.ids.offset, entry.ids.size, "id"));
 
     tbv = buffer_.SliceOff(entry.ids.offset, entry.ids.size);
     if (!tbv) {
@@ -396,6 +414,9 @@ void PerfDataTokenizer::MaybePushRecord(Record record) {
 base::StatusOr<PerfDataTokenizer::ParsingResult>
 PerfDataTokenizer::ParseFeatureSections() {
   PERFETTO_CHECK(buffer_.start_offset() == header_.data.end());
+  RETURN_IF_ERROR(CheckSectionInBounds(feature_headers_section_.offset,
+                                       feature_headers_section_.size,
+                                       "feature headers"));
   auto tbv = buffer_.SliceOff(feature_headers_section_.offset,
                               feature_headers_section_.size);
   if (!tbv) {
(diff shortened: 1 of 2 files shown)


src/trace_processor/util/trace_blob_view_reader.cc
```diff
@@ -68,6 +69,15 @@ auto TraceBlobViewReader::SliceOffImpl(const size_t offset,
     return visitor.OneSlice(TraceBlobView());
   }
 
+  // Reject requests where `offset + length` would overflow. The bounds checks
+  // below compute `offset + length` in size_t arithmetic, so an overflowing
+  // (offset, length) pair -- e.g. an attacker-controlled section offset read
+  // from a perf.data file -- could wrap to a small value, spuriously pass the
+  // fast-path check and produce an out-of-bounds slice.
+  if (PERFETTO_UNLIKELY(offset > std::numeric_limits<size_t>::max() - length)) {
+    return visitor.NoData();
```
LalitMaganti: NoData is not sufficient - we should really just be rejecting the trace as a whole not just returning an empty slice.

adilburaksen:
Good point — thanks. Since `SliceOff` returns `std::optional<TraceBlobView>` where `nullopt` means "incomplete, wait for more data" to the streaming callers, the reader itself can't signal a hard error, and returning empty for an overflowing range would just stall as `kMoreDataNeeded` instead of rejecting.

I've moved the rejection to `PerfDataTokenizer`, which already returns `base::StatusOr`. It now validates the attacker-controlled section bounds (`attrs`, `id`, `feature-headers`, `feature`) and returns `base::ErrStatus` on an overflowing `offset + size`, so the trace as a whole is rejected. I kept the reader-level guard as defense-in-depth so the OOB read cannot happen even via a future caller that forgets to validate. Let me know if you'd prefer the reader guard dropped entirely.

src/trace_processor/importers/perf/perf_data_tokenizer.cc
```diff
@@ -213,8 +213,24 @@ PerfDataTokenizer::ParseHeader() {
   return ParsingResult::kSuccess;
 }
 
+// Rejects a section whose attacker-controlled `offset + size` overflows. Such a
+// range can never be satisfied by more data, so it must reject the trace rather
+// than be treated as "more data needed" (an empty/absent slice).
+static base::Status CheckSectionInBounds(uint64_t offset,
+                                         uint64_t size,
+                                         const char* name) {
+  if (offset + size < offset) {
```
LalitMaganti: I would prefer if we just checked against uint64 max instead of overflowing and trying to catch it. Would be more intuitirve.

adilburaksen:
Done — switched to an explicit `size > uint64_max - offset` check. Thanks.

src/trace_processor/util/trace_blob_view_reader.cc
LalitMaganti: Can we please remove this now? We can make this a CHECK instead because we should never be getting here with this property.

adilburaksen:
Done — replaced the reader's NoData early-return with `PERFETTO_CHECK(offset <= size_t_max - length)`, since the tokenizer now rejects overflowing section bounds before we ever slice. Thanks for the reviews.

LalitMaganti (approved):

LalitMaganti (approved):

src/trace_processor/importers/perf/perf_data_tokenizer.cc
```diff
@@ -213,8 +214,24 @@ PerfDataTokenizer::ParseHeader() {
   return ParsingResult::kSuccess;
 }
 
+// Rejects a section whose attacker-controlled `offset + size` overflows. Such a
+// range can never be satisfied by more data, so it must reject the trace rather
+// than be treated as "more data needed" (an empty/absent slice).
+static base::Status CheckSectionInBounds(uint64_t offset,
```
safayat-google: nits, assuming that the offset and size will always come from a `struct Section` object, let's pass the selection object directly.
```
static base::Status CheckSectionInBounds(const PerfFile::Section& section,
                                         const char* name) {
```

Then we can avoid passing the last parameter as well by defining a macro
```
#define RETURN_IF_SECTION_OUT_OF_BOUNDS(sec) \
  RETURN_IF_ERROR(CheckSectionInBounds(sec, #sec))
```

Finally it will look pretty cleaner.
```
- RETURN_IF_ERROR(CheckSectionInBounds(header_.attrs.offset, header_.attrs.size, "attrs"));
+ RETURN_IF_SECTION_OUT_OF_BOUNDS(header_.attrs);
```

adilburaksen:
Done — `CheckSectionInBounds` now takes the `PerfFile::Section` directly. Thanks!

safayat-google (approved):

LalitMaganti (approved):

adilburaksen:
@LalitMaganti @safayat-google — this has been approved since 8 July but never merged, and I found why while re-checking it: `repo-checks` was red the whole time, so `ensure-ci-is-green` blocked the merge. It was not waiting on either of you.

The failure was eight clang-format violations in lines this PR introduced — the two `RETURN_IF_ERROR(CheckSectionInBounds(...))` calls in `perf_data_tokenizer.cc` and the comment block above the new `PERFETTO_CHECK` in `trace_blob_view_reader.cc`. My apologies for leaving that red; I read the approvals and stopped watching CI.

Fixed in 2d874ec: reformatted with the pinned `third_party/clang-format` (linux64 `79a7b4e5`, sha256 verified against `tools/install-build-deps`). I confirmed both files are clean on `main` before the change and that the same eight errors reproduce on the previous head, so the violations were mine and the fix is limited to them — the diff only rewraps those lines, no behavior change.

Everything else was already green (asan/lsan, msan, tsan, x86, gcc9, android, ui, bazel, fuzzer, rust-sdk). If the new commit dismissed the approvals, could I get a re-stamp so it can land?

adilburaksen:
CI is now fully green on 2d874ec — `repo-checks` passes, and so does `ensure-ci-is-green`, along with asan/lsan, msan, tsan, debug, x86-release, gcc9-release, android, ui, bazel, fuzzer, rust-sdk and analyze. The clang-format failure that had been blocking the merge since 8 July is resolved.

The one remaining red is `enforce-reviews`, and it is purely procedural: fork PRs need two Google org member approvals **on the latest commit**, and pushing the format fix reset that count to zero:

```
Google org member approvals for latest commit: 0
❌ PR from fork requires 2 Google organization member approvals on the latest commit.
```

There was no way around it — the formatting could only be fixed with a new commit. @LalitMaganti @safayat-google, since you both approved ca20b86 already, would you mind re-stamping 2d874ec? The only delta between the two is the rewrapping of eight lines; there is no behavior change.

PR #6518, opened 2026-07-03
trace_processor: bounds-check id lookups in the instruments importer

adilburaksen:
**trace_processor: bounds-check id lookups in the instruments importer**

`RowDataTracker::GetThread/GetProcess/GetFrame/GetBacktrace` index
`threads_/processes_/frames_/backtraces_` at `[id - 1]`, guarded only by a
`PERFETTO_DCHECK(id != kNullId)` — which is compiled out in release builds
(`PERFETTO_DCHECK_IS_ON()` is 0 for `NDEBUG` non-standalone/chromium/android builds,
including the shipped WASM UI).

These ids come (transitively) from an untrusted Apple Instruments `.trace` file. A `<frame>`
emitted with no enclosing `<backtrace>`, or a `<backtrace>`/`<frame>`/`<thread>` referencing an
undefined id, leaves the id as `kNullId` (0) or out of range. In a release build
`GetBacktrace(0)` then evaluates `&backtraces_[0u - 1]` = `&backtraces_[0xFFFFFFFF]`, a wild
pointer that the following `frames.push_back(...)` (out-of-bounds write) or the field reads in
`RowParser` (out-of-bounds read) dereference — a memory-unsafe crash on a crafted trace
(confirmed under ASan).

`GetBinary` already guards this (`if (id == kNullId) return nullptr;`). Extend the same
fail-closed pattern to the other four getters (also rejecting out-of-range ids) and null-check at
the call sites so a malformed row/frame is skipped rather than dereferencing a wild pointer.

### DIFF
--- a/src/trace_processor/importers/instruments/row_data_tracker.cc
+++ b/src/trace_processor/importers/instruments/row_data_tracker.cc
@@ -35,9 +35,9 @@ IdPtr<Thread> RowDataTracker::NewThread() {
   return {id + 1, ptr};
 }
 Thread* RowDataTracker::GetThread(ThreadId id) {
-  // Ids are 1-based and come (transitively) from the parsed trace, so a malformed
-  // file can supply kNullId or an out-of-range id. Fail closed rather than indexing
-  // out of bounds (see GetBinary for the same pattern).
+  // Ids are 1-based and come (transitively) from the parsed trace, so a
+  // malformed file can supply kNullId or an out-of-range id. Fail closed rather
+  // than indexing out of bounds (see GetBinary for the same pattern).
   if (id == kNullId || id > threads_.size()) {
     return nullptr;
   }
@@ -51,9 +51,9 @@ IdPtr<Process> RowDataTracker::NewProcess() {
   return {id + 1, ptr};
 }
 Process* RowDataTracker::GetProcess(ProcessId id) {
-  // Ids are 1-based and come (transitively) from the parsed trace, so a malformed
-  // file can supply kNullId or an out-of-range id. Fail closed rather than indexing
-  // out of bounds (see GetBinary for the same pattern).
+  // Ids are 1-based and come (transitively) from the parsed trace, so a
+  // malformed file can supply kNullId or an out-of-range id. Fail closed rather
+  // than indexing out of bounds (see GetBinary for the same pattern).
   if (id == kNullId || id > processes_.size()) {
     return nullptr;
   }
@@ -67,9 +67,9 @@ IdPtr<Frame> RowDataTracker::NewFrame() {
   return {id + 1, ptr};
 }
 Frame* RowDataTracker::GetFrame(BacktraceFrameId id) {
-  // Ids are 1-based and come (transitively) from the parsed trace, so a malformed
-  // file can supply kNullId or an out-of-range id. Fail closed rather than indexing
-  // out of bounds (see GetBinary for the same pattern).
+  // Ids are 1-based and come (transitively) from the parsed trace, so a
+  // malformed file can supply kNullId or an out-of-range id. Fail closed rather
+  // than indexing out of bounds (see GetBinary for the same pattern).
   if (id == kNullId || id > frames_.size()) {
     return nullptr;
   }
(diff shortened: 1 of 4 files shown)


src/trace_processor/importers/instruments/row_parser.cc
```diff
@@ -50,7 +50,13 @@ void RowParser::Parse(int64_t ts, instruments_importer::Row row) {
   }
 
   Thread* thread = data_.GetThread(row.thread);
+  if (!thread) {
+    return;
```
LalitMaganti, line 58: We need to use the import log tracker to add import logs about here about what we're doing, why we're rejecting it etc.

src/trace_processor/importers/instruments/row_parser.cc
```diff
@@ -50,7 +50,13 @@ void RowParser::Parse(int64_t ts, instruments_importer::Row row) {
...
+  }
   Process* process = data_.GetProcess(thread->process);
+  if (!process) {
+    return;
```
LalitMaganti, line 64: As above.

src/trace_processor/importers/instruments/row_parser.cc
```diff
@@ -67,12 +73,18 @@ void RowParser::Parse(int64_t ts, instruments_importer::Row row) {
   auto& stack_profile_tracker = *context_->stack_profile_tracker;
 
   Backtrace* backtrace = data_.GetBacktrace(row.backtrace);
+  if (!backtrace) {
+    return;
```
LalitMaganti, line 85: As above.

src/trace_processor/importers/instruments/row_parser.cc
```diff
@@ -67,12 +73,18 @@ void RowParser::Parse(int64_t ts, instruments_importer::Row row) {
...
+  }
   std::optional<CallsiteId> parent;
   uint32_t depth = 0;
   auto leaf = backtrace->frames.rend() - 1;
   for (auto it = backtrace->frames.rbegin(); it != backtrace->frames.rend();
        ++it) {
     Frame* frame = data_.GetFrame(*it);
+    if (!frame) {
+      continue;
```
LalitMaganti, line 96: as above.

adilburaksen:
Thanks — agreed in principle, but there's a wrinkle here that differs from #6517. Unlike `PerfDataTokenizer` (which returns `base::StatusOr` and can reject directly), `RowParser` is a `TraceSorter::Sink` and its `Parse(int64_t, Row)` is a `void` callback invoked during the sorter flush, so there's no Status channel back to the tokenizer to reject the trace from this point.

To reject rather than skip, the options I see are:

- **(a)** widen the `TraceSorter::Sink` contract so `Parse` can return `base::Status` — but that touches every importer that uses the sorter;
- **(b)** record the first invalid lookup on the `RowParser` and surface it as `base::ErrStatus` after the sort completes (localized to the instruments importer);

Given the `void`-sink constraint, do you prefer (b), or is skipping the malformed rows with a stats counter acceptable here? Happy to implement whichever you'd like.

LalitMaganti:
Please read my message carefully, I've explained how to fix this and it has nothing to do with returning base::Status.

adilburaksen:
You're right, sorry — I misread. Done: each skip now records an import log via `context_->import_logs_tracker->RecordParserError(...)` explaining what was dropped (unknown thread / process / backtrace / frame) and keeps the timestamp, instead of dropping the row silently. Added the four corresponding `stats` keys.

src/trace_processor/importers/instruments/row_data_tracker.cc
safayat-google: also check  `id > binaries_.size()`?

src/trace_processor/storage/stats.h
```diff
@@ -140,6 +140,14 @@ namespace perfetto::trace_processor::stats {
   F(gpu_render_stage_parser_errors,       kSingle,  kError,    kAnalysis, Scope::kMachineAndTrace, ""), \
   F(graphics_frame_event_parser_errors,   kSingle,  kInfo,     kAnalysis, Scope::kMachineAndTrace, ""), \
   F(guess_trace_type_duration_ns,         kSingle,  kInfo,     kAnalysis, Scope::kGlobal, ""), \
+  F(instruments_row_missing_thread, kSingle, kError, kAnalysis, Scope::kMachineAndTrace,               \
```
safayat-google: For all of these lines [143, 150] please keep the same format/alignment with the other lines in this file

LalitMaganti (approved):

src/trace_processor/importers/instruments/row_parser.cc
safayat-google: since `GetBinary(` can return `nullptr` when `id > size`, should we check `binary` instead?

safayat-google (approved):

LalitMaganti (approved):

PR #6520, opened 2026-07-04
ui: Add intelletto tools for building DE graphs

stevegolton:
Inject the following tools and context injections to allow the agent to build, read and modify DE graphs.

Note 1: For now, these tools are defined in the Intelletto plugin. Keeping all tool calls in this plugin doesn't scale, and plugins in general should register their own tools with the intelletto plugin to expose more functionality to the assistant (or use commands when we adapt them for model use), but this introduces a dependency that transitively enables the intelletto plugin when depended on by an default enabled plugin. I want to avoid default enabling the intelletto plugin while in development, thus we flip the dependency for the DE tools - for now.

Note 2: The tools expect the model to build the DE JSON directly, so we teach it how to understand the JSON schema in the tool descriptions. Once we have Perfetto SQL Next we can transition to this.

Tools:

| Tool | Mutating | What it does |
|------|----------|--------------|
| `get_graph` | no | Returns the active Data Explorer graph as JSON (or `"<empty>"`). For the model to read what the user is exploring. |
| `set_graph` | yes | Replaces the graph with supplied JSON and navigates to the Data Explorer. Structural pre-check → apply → returns per-node runtime errors so the model can iterate. Its description embeds the whole `GRAPH_FORMAT` catalog. |
| `check_graph` | no | Runs the *current* graph against the trace, reports per-node errors, changes nothing. |
| `validate_graph` | no | Dry-runs a *candidate* graph JSON (structural + runtime) without applying it. |

Context injections:

| Id | Kind | What it injects |
|----|------|-----------------|
| `dev.perfetto.DataExplorer#selected_node` | context | The node the user has selected in the query builder: `nodeId`, `type`, `title`, serialized `state`, and output `columns`. Lets the model resolve "this node" / "the selected step" and edit the right one. |

### DIFF
--- a/ui/src/plugins/dev.perfetto.DataExplorer/index.ts
+++ b/ui/src/plugins/dev.perfetto.DataExplorer/index.ts
@@ -14,6 +14,7 @@
 
 import './styles.scss';
 import m from 'mithril';
+import {z} from 'zod';
 import type {PerfettoPlugin} from '../../public/plugin';
 import type {Trace} from '../../public/trace';
 import type {Store} from '../../base/store';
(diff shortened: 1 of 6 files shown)


ui/src/plugins/dev.perfetto.DataExplorer/index.ts
```diff
@@ -77,7 +90,11 @@ function isValidPersistedState(
 
 export default class implements PerfettoPlugin {
   static readonly id = 'dev.perfetto.DataExplorer';
-  static readonly dependencies = [QueryPagePlugin, SqlModulesPlugin];
+  static readonly dependencies = [
+    QueryPagePlugin,
+    SqlModulesPlugin,
+    IntellettoPlugin,
```
stevegolton: Hang on - this will transitively enable the intelletto plugin by default - but we're not ready yet...

I'll think about how to do this.

stevegolton: Flipped the dependency.

LalitMaganti (approved):

PR #6527, opened 2026-07-06
protovm: export compiler binary and blob emitter to Android.bp

BogdanFloris:
Exports the protovm compiler as a host binary so it can be used in soong genrules

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -24253,6 +24253,59 @@ cc_library_static {
     ],
 }
 
+// GN: //src/protovm/compiler:protovm_compiler
+cc_binary_host {
+    name: "protovm_compiler",
+    srcs: [
+        ":perfetto_base_default_platform",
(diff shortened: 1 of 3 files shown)


rukkal (approved):

Android.bp
```diff
@@ -26146,3 +26198,10 @@ filegroup {
     path: "src/trace_processor/plugins/wattson/data",
     visibility: ["//visibility:public"],
 }
+
+filegroup {
+    name: "perfetto_cpp_blob_emitter",
```
sashwinbalaji, line 26496: curious why is this needed ?

BogdanFloris: because of soong's sandbox rules, we cannot reference the script using relative paths from in `frameworks/native` using `../../external/perfetto`. to work around this, we can wrap it inside the filegroup.

sashwinbalaji (approved):

Issue #6260, opened 2026-06-16
Unable to add extensions to TrackEvent proto from UI extension server


carlscabgro:
I am unable to add extensions to the TrackEvent proto using an UI ExtensionServer that serves proto descriptors.

The proto descriptors sent by the ExtensionServer are applied to the descriptor data base before the TrackEvent descriptor has been added. Thus the parsing code will fail as it will be unable to find the proto definition for TrackEvent when parsing the extension fields in the descriptor. Sending the TrackEvent proto descriptor does not quite work either as the TraceProcessor initialisation will fail when adding the real TrackEvent descriptor (because one already exists).

I can see two options to fix this. Either do the processing of server extensions at the very end (seems fairly easy). Or delay name resolution of referenced protos in descriptors to the very end (probably trickier to implement). Is there a third approach?

Happy to send PRs your way if you tell me what approach you prefer.

LalitMaganti:
UI extensions should be processed by TP at the very end after baked in extensions. This is a bug. Yes please do send a PR for this if this a simple change! Otherwise let me know and I can look.

carlscabgro:
I have uploaded https://github.com/google/perfetto/pull/6528 to show this issue in action.

Note this test crashes, because the registry is left in a bad state because parsing the slim_descriptor fails, which means any further descriptor addition to the pool fails (silently). When we get to add the trace_summary descriptor the failure is CHECKED and thus the test crashed. But as I said, the original problem happens much earlier.

LalitMaganti:
Looking at the PR, I have a few thoughts:
1) longer term I need to work out a hierarchy of exactly what should happen when with respect to descriptor handling and write it down
2) for now we should simply defer extension server descriptors at the same time the track event module code runs as that will trivial solve the problem.

carlscabgro:
> 2. for now we should simply defer extension server descriptors at the same time the track event module code runs as that will trivial solve the problem.

IIRC trace_event descriptors are added as part of the reader initialisation (ProtoTraceReader) and this happens when a proto trace is detected (not sure additional descriptors are only needed for proto traces) + you would need to wire the descriptors from the config down to the reader.
If this is acceptable I can update the PR accordingly.

LalitMaganti:
Yes that's fine we should do that.

LalitMaganti:
https://github.com/google/perfetto/pull/6528 has landed.

PR #6528, opened 2026-07-06
tp: Defer parsing of extra descriptors until TrackEvent descriptors have been registered

carlscabgro:
# What
Defer parsing of extra descriptors until TrackEvent descriptors have been registered. This is an intermediate fix until we work out a hierarchy of exactly what should happen when with respect to descriptor handling.

# Why 
#6260 
- Extension descriptors are processed before the TrackEvent descriptor
- Descriptor processing skips allready processed files
- TrackEvent descriptor is added with merge=false

This means that in order to add a new extension descriptor, you need to also pass the TrackEvent description in the same FileSet. This means that when the "real" TrackEvent is added to the pool it is skipped, because the "track_event.proto" file has been already processed. So we might have a newer or (worse) older proto definition loaded. If the extension descriptor were to ship a stripped version fo the TrackEvent (with a dummy file name and defining only the extension range) once the real TrackEvent descriptor is added (no file name match this time) we get an error because the TrackEvent is already defined and it is being added with merge=false.


carlscabgro:
@LalitMaganti This is the test that shows the issue I mentioned in #6260. Lets discuss on the issue if this is really an issue and if so what is the best way to fix it.

carlscabgro:
@LalitMaganti PTAL

LalitMaganti (approved):

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -206,6 +206,15 @@ ProtoTraceReader::ProtoTraceReader(TraceProcessorContext* ctx,
   if (context_->register_additional_proto_modules) {
     context_->register_additional_proto_modules(&module_context_, context_);
   }
+  // This needs to happen after the TrackEvent descriptors have been registered,
+  // which happens in one of the modules.∫ (See
```
sashwinbalaji: ∫ is this intended ?

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -206,6 +206,15 @@ ProtoTraceReader::ProtoTraceReader(TraceProcessorContext* ctx,
...
+  // https://github.com/google/perfetto/issues/6260)
+  for (const std::string& raw_bytes :
+       context_->config.extra_parsing_descriptors) {
+    context_->descriptor_pool_->AddFromFileDescriptorSet(
```
sashwinbalaji: 1) can we add a dcheck to esnure track_event descriptors are loaded
2) should we check the return status of AddFromFileDescriptorSet?

sashwinbalaji (approved):

src/trace_processor/importers/proto/proto_trace_reader.cc
carlscabgro: > 1. can we add a dcheck to ensure track_event descriptors are loaded
Sure, will do
> 2. should we check the return status of AddFromFileDescriptorSet?
We could add a new stats to track failed descriptor parsing or some sort of error logging. Currently you will just see stat errors for not being able to parse track event extensions but the real problem happened here. thoughts?

carlscabgro: I could also use `ImportLogsTracker::RecordAnalysisError` to record the actual error.(Not really an import error  but AFAIK there is no general error logging mechanism right?

carlscabgro: I was trying this out, and I realised that there is no way to pass extra_parsing_descriptors from the command line. I am tempted to add a command line flag to TP `--add-extra-parsing-descriptors` WDYT?

LalitMaganti: Could we make it a dev flag instead? I don't think it makes sense for anyone except as a dev flag.

carlscabgro: Sure, but extra_parsing_descriptors and dev_flags are both members of the Config object, so you will have the same info in two places. I'll update the PR with all the changes and lets discuss based on that.

LalitMaganti: I think that's fine - I've done that with other things already. See for example the tokenize_only dev flag for example.

carlscabgro: PTAL

src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -206,6 +206,26 @@ ProtoTraceReader::ProtoTraceReader(TraceProcessorContext* ctx,
   if (context_->register_additional_proto_modules) {
     context_->register_additional_proto_modules(&module_context_, context_);
   }
+  // This needs to happen after the TrackEvent descriptors have been registered,
+  // which happens in one of the modules. (See
+  // https://github.com/google/perfetto/issues/6260)
+  PERFETTO_DCHECK(context_->descriptor_pool_->FindDescriptorIdx(
+      ".perfetto.protos.TrackEvent"));
```
LalitMaganti: I don't think this is adding much. Please remove.

LalitMaganti (approved):

src/trace_processor/importers/proto/proto_trace_reader.cc
carlscabgro: Done!

LalitMaganti (approved):

LalitMaganti (approved):

carlscabgro:
Could that be a flaky test?

PR #6540, opened 2026-07-06
ui: add ring buffer size control to stack sampling record page

LalitMaganti:
traced_perf suggests increasing ring_buffer_pages when the kernel
drops samples, but until now the only way to set it was to hand-edit
the trace config proto. Expose it as a slider in the Record > Stack
sampling page.

The control is denominated in KiB for clarity and converted to 4 KiB
pages internally, rounding up to the power of two that the kernel
requires. Leaving it at 0 keeps the traced_perf default (1 MiB per
cpu).

Bug: 407406366


rsavitski (approved):
LGTM. If you wish to go the extra distance, the "description" and the slider are clashing with a vertical monitor aspect ratio.

Image

LalitMaganti:
> LGTM. If you wish to go the extra distance, the "description" and the slider are clashing with a vertical monitor aspect ratio.

Done thanks!

PR #6544, opened 2026-07-06
ui: Add recently opened traces section to the home page

stevegolton:
This is a first attempt based purely on what data we have available in the cache. Thus, it only applies to traces that actually get cached (e.g. traces loaded from a url don't get cached).

Features:
- Show up to the 5 most recently cached traces.
- Click to open that trace.
- Click the x button to delete it from the cache.
- Use MiddleEllipsis widget to handle long titles.

Fixes: b/531725088

image

### DIFF
--- a/ui/src/core/cache_manager.ts
+++ b/ui/src/core/cache_manager.ts
@@ -38,7 +38,7 @@ async function getCache(): Promise<Cache | undefined> {
   return LAZY_CACHE;
 }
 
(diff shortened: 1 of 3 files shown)


stevegolton:
Hang on, need to handle title wrapping properly...

Edit: Switched to using MiddleEliipsis to handle long titles.

LalitMaganti (approved):

PR #6557, opened 2026-07-07
tools: support allowlisting options in proto merger

BogdanLionte:


src/tools/proto_merger/main.cc
```diff
@@ -77,6 +77,8 @@ const char kUsage[] =
 -r, --upstream-root-message: Root message in the upstream proto for which new
                               fields from the allowlist will be allowed.
 -o, --output:                Output path for writing the merged proto file.
+-w, --whitelisted-options:   Option/annotation keys that should be preserved
```
LalitMaganti: nit: allowlisted.

LalitMaganti: Here and throughout.

src/tools/proto_merger/main.cc
```diff
@@ -102,6 +104,8 @@ int Main(int argc, char** argv) {
       {"allowlist", required_argument, nullptr, 'a'},
       {"upstream-root-message", required_argument, nullptr, 'r'},
       {"output", required_argument, nullptr, 'o'},
+      {"allowlisted-options", required_argument, nullptr, 'w'},
```
LalitMaganti: Why are there too here? I also woudn't bother with comma-separated. Just make them repeated and it will make both your life easier and clearer for other people if they have e.g. spaces in their option names.

BogdanLionte: Done

LalitMaganti (approved):

dragosp95 (approved):

PR #6560, opened 2026-07-07
ui: ftrace plugin: support area selections of raw ftrace tracks

rsavitski:
Surface ftrace_explorer as a sub-tab of "current selection",
handling area selections. Both the standalone and the sub-tab versions
can exist at the same time, and share the event name filter.

Also:
* enlarge the [last] text-formatted "Args" column, similar to the
  logs panel.
* make the ftrace tab multimachine-aware with ucpu indices.
* I did not implement counting event totals within the selected bounds,
  so the event filter always shows all distinct event names seen in the
  trace.


stevegolton (changes requested):
Looks great! The new tab is perfect!

There is however a regression in the existing ftrace explorer tab - it no longer offers the ability to control the cpu filters, and this cpu filters are taken out of the plugin state and thus are not persisted over permalinks.

I'd probably at this point completely decouple FtraceExplorer from the filterStore and just hook up the information it needs in index.ts. This way the original tab can take the cpu filter state from the store, and the new area selection tab can take the cpu list from the are selection.

rsavitski:
Removing the cpu filter was an intentional choice on my part (see PR description). If we had ftrace track selections before the cpu filters in ftrace_explorer, would we have added the filters in their current state at all? I personally find the lack of visual feedback in the cpu selection quite subtle, and think that area selections are more obvious if someone needs to zoom in on a single cpu.

stevegolton:
Understood. However, removing features from the old UX is a regression. Removing features from existing workflows is pretty jarring, even if the area selection view is a better way to filter.

rsavitski:
I personally disagree, but have restored the CPU filter button in the standalone tab. I've changed its internals to be an inclusion filter. The exclusion makes sense for event names, but cpus are already known up-front. (I'm not jumping through hoops to migrate permalink-persisted cpu exclusion filters, those will reset in the new UI.)

I've also wired through multimachine ucpu indices and names, because the cpu filter dropdown always crashed (even at HEAD) with duplicate cpu indices. So the area selection is now multimachine-aware as well.

Does the shape of lifting state into index.ts make sense? In particular the addition of onChange callbacks for the state updates.

rsavitski:
(the ui difftests are a pain for me to update, so I'll update them once we agree on the shape of the change)

ui/src/plugins/dev.perfetto.Ftrace/index.ts
stevegolton: Just bump this version so that old versions of the UI won't crash when attempting to open permalinks from new versions.

stevegolton (approved):
Looks good to me - just one nit about the version

PR #6563, opened 2026-07-07
ui: improve error dialogs for corrupt, not present and unknown formats

LalitMaganti:
Instead of falling back to the generic error dialog, provide a nice
opinionated dialog which guides the user on the actual problem

### DIFF
--- a/ui/src/frontend/error_dialog.ts
+++ b/ui/src/frontend/error_dialog.ts
@@ -25,6 +25,8 @@ import {Button, ButtonVariant} from '../widgets/button';
 import {Intent} from '../widgets/common';
 import {Checkbox} from '../widgets/checkbox';
 import {Anchor} from '../widgets/anchor';
+import {Callout} from '../widgets/callout';
+import {CodeSnippet} from '../widgets/code_snippet';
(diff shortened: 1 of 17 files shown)


ui/src/frontend/error_dialog.ts
```diff
@@ -335,47 +350,168 @@ function showOutOfMemoryDialog() {
...
+          {icon: Icons.Info},
+          "The file opened doesn't look like a Perfetto trace or any " +
+            'other format recognized by the Perfetto trace processor.',
+        ),
+        m(
+          'div',
+          m('span.pf-error-dialog__list-label', 'Formats supported'),
+          m(
+            'ul.pf-error-dialog__list',
+            m('li', 'Perfetto protobuf trace'),
```
sashwinbalaji, line 367: nit: except perfetto make others alphabetical ?

sashwinbalaji (approved):

ui/src/frontend/error_dialog.ts
LalitMaganti: No I actually do want the explicit order here :)

PR #6570, opened 2026-07-07
ui: don't cancel timeline selection when dismissing a popup

### LINKED ISSUE #3629: ui: Pressing escape or clicking outside a popup (on the timeline) should not cancel the selection
Clicking outside a popup or pressing escape are both ways to close that popup, however they also have additional side effects as the event is propagated through the other elements that are clicked on / interact with hotkeys. The most grievous issue is this can cancel the selection on the timeline page.

More succinctly:
- Clicking outside a popup should have no other effect other than closing the popup.
- Pressing escape when a popup is open should have no other effect other than closing the popup.

See the attached video:

https://github.com/user-attachments/assets/82c6e3cf-a661-4634-9445-4b02df61a858

LalitMaganti:
Closing a popup by pressing Escape or clicking outside it leaked the
event to global document listeners, which then cleared the active
timeline selection. Escape reached the global Deselect hotkey in
HotkeyContext, and an outside click reached the timeline's
click-to-deselect gesture.

Register the popup's document mousedown and keydown listeners in the
capture phase so they run before those bubble-phase listeners, and call
preventDefault/stopPropagation when the event actually dismisses the
popup. Dismissing a popup now has no other side effect.

Fixes #3629.

### DIFF
--- a/ui/src/widgets/popup.ts
+++ b/ui/src/widgets/popup.ts
@@ -261,8 +261,10 @@ export class Popup implements m.ClassComponent<PopupAttrs> {
         );
         this.popupElement = popupElement;
         this.createOrUpdatePopper(attrs);
-        document.addEventListener('mousedown', this.handleDocMouseDown);
-        document.addEventListener('keydown', this.handleDocKeyPress);
+        // Capture phase so we can consume a dismissing event before global
+        // bubble-phase listeners (hotkeys, timeline deselect) see it.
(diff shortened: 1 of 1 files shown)


stevegolton (commented):
Sorry but this is a hack.

This will block *all* esc keypresses all the way up the tree when a popup is open.

I'm not sure but it might affect closing nested popups too.

I really think we should do something based on focus.

In hindsight this isn't a fixit bug, so I'll remove it from the list.

PR #6573, opened 2026-07-07
ui/tp: Fix empty Linux perf flamegraph counter options

### LINKED ISSUE #6552: Perf text area selections appear to be broken
It seems like they're looking for the cpu clock even though the script files don't have this. This might be something wrong at the UI level where we assume the CPU clock might be present. Or it might be TP where we're incorrectly putting the CPU counter.

Either way, we need to fix this.

safayat-google:
perfetto: Fix empty Linux perf flamegraph counter options
Prevent empty counter metric options (like "cpu-clock" for software clock
sources) from showing in the perf flamegraph dropdown when no counter
values exist in the trace.

- tp: Verify if the optional `timebase_count` field is present in the
  protobuf perf_sample packet before calling `PushCounter`. This stops
  writing dummy `0.0` counter entries to the database for software clock
  sources.
- ui: Update the pre-fetch query in `cacheCounterTypesPerSession` to
  exclude counter tracks that do not have at least one non-zero counter
  entry using a `WHERE EXISTS` check.
- test: Added `test_perf_sample_timebase_count` diff test to assert that
  sample packets with count fields populate counters, while packets
  without do not.

Fixes: #6552


ui/src/plugins/dev.perfetto.LinuxPerf/index.ts
```diff
@@ -94,6 +94,11 @@ export default class LinuxPerfPlugin implements PerfettoPlugin {
     const result = await trace.engine.query(`
       SELECT pct.perf_session_id, pct.name, MAX(pct.is_timebase) as is_timebase
       FROM perf_counter_track pct
+      WHERE EXISTS (
```
LalitMaganti: Is this necessary given your other change?

safayat-google: Yes, this is still necessary because the track configuration metadata is still registered in the `perf_counter_track` table at trace start. So, the query will still find the empty `cpu-clock` track and include it in the dropdown option.

LalitMaganti: I think this is better written as a cte with a distinct on teack_id and then join based on that. Or use the counter summary table. This particular join will be very expensive

safayat-google: Replaced with `join _counter_track_summary`.
Thanks for this suggestion!

LalitMaganti (approved):

PR #6576, opened 2026-07-07
tp: Add trace_diagnostics table to detect bad trace configs

primiano:
Introduce a new `trace_diagnostics` table that surfaces likely
misconfigurations in a trace's TraceConfig, so the UI (and any other
consumer) can warn users about configs that hurt trace quality. Each row
is a (name, description, remediation, confidence) tuple scoped to the
current trace and machine.

Detection runs in C++ because the TraceConfig is only reachable in a
structured form there: only the opaque pbtxt survives into SQL. A new
per-(trace, machine) TraceDiagnosticsTracker captures the raw config at
parse time (MetadataModule) and runs a growing corpus of detection rules
at finalization, once trace bounds and clocks are available. Rules are
plain functions; adding one is a single entry in kRules[]. Shared
machinery (emitting rows, iterating ftrace/data-source configs, deriving
values from the context, scanning stats) lives in TraceDiagnosticsHelper
to keep the rule file focused on the detection logic.

Initial rules cover the most common ftrace/profiling mistakes:
  - tiny ftrace buffer_size_kb (size-proportional confidence)
  - low ftrace drain bandwidth (buffer_size_kb / drain_period_ms)
  - extreme drain_period_ms
  - syscall tracing without a syscall_events filter
  - preserve_ftrace_buffer with tracing started long after boot
  - sched_switch without compact_sched
  - events needing symbolize_ksyms (and enable_function_graph)
  - DISCARD buffers backing ftrace/track_event while streaming to a file
  - atrace_apps: "*" with many categories (when data was lost)
  - heapprofd sampling_interval_bytes below 100 KB

Buffer/bandwidth rules escalate their confidence when the trace actually
recorded ftrace data loss, and the heapprofd rule escalates on heapprofd
errors.

```
Stack:
├─ origin/dev/primiano/ui-badconf
  ├─ **[This PR] (ui-badconf)**
    ├─ badconf-ui
```


src/trace_processor/perfetto_sql/stdlib/prelude/before_eof/trace_diagnostics.sql
```diff
@@ -0,0 +1,36 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- Diagnostics detecting poorly-written trace configs (and other trace-quality
+-- problems), one row per detected issue. Written by TraceDiagnosticsTracker.
+CREATE PERFETTO VIEW trace_diagnostics(
```
LalitMaganti: I'm not sure on this name tbh. We should really try and normalize between stats, metadata, import logs and trace diagnostics - all of these are trying to do similar but not same things and I'd like to have clear understanding (written down somewhere) on what should go where.

src/trace_processor/tables/metadata_tables.py
```diff
@@ -861,6 +861,65 @@
...
+            cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
+        ),
+        C(
+            'confidence',
+            CppDouble(),
+            cpp_access=CppAccess.READ,
+            cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
+        ),
+        C(
+            'machine_id',
```
LalitMaganti: Do we really want to make this per-machine or just per-trace?

src/trace_processor/tables/metadata_tables.py
```diff
@@ -861,6 +861,65 @@
...
+TRACE_DIAGNOSTICS_TABLE = Table(
+    python_module=__file__,
+    class_name='TraceDiagnosticsTable',
+    sql_name='__intrinsic_trace_diagnostics',
+    columns=[
+        C(
+            'name',
+            CppString(),
+            cpp_access=CppAccess.READ,
+            cpp_access_duration=CppAccessDuration.POST_FINALIZATION,
```
LalitMaganti: You don't actually need access duration of post_finalization. Please remove.

src/trace_processor/perfetto_sql/stdlib/prelude/before_eof/trace_diagnostics.sql
primiano: As discussed offline moved to directly use the __intrinstic

src/trace_processor/types/trace_processor_context.h
```diff
@@ -254,6 +255,7 @@ class TraceProcessorContext {
   PerTraceAndMachinePtr<SchedEventTracker> sched_event_tracker;
   PerTraceAndMachinePtr<MetadataTracker> metadata_tracker;
   PerTraceAndMachinePtr<StatsTracker> stats_tracker;
+  PerTraceAndMachinePtr<TraceDiagnosticsTracker> trace_diagnostics_tracker;
```
LalitMaganti: PerTrace

LalitMaganti (approved):

PR #6580, opened 2026-07-08
tp: add zip64 support to the streaming zip reader

### LINKED ISSUE #991: Request for Zip64 Support in Perfetto
I came across the following comment in the Perfetto UI source code:

`// - Does NOT support zip64, encryption and other advanced zip file features.`
`// - It is not suitable for security-sensitive contexts. E.g. it doesn't deal`
`//   with zip path traversal attacks (the same file showing up twice with two`
`//   different payloads).`

I understand that the current implementation does not support Zip64, and I was wondering if there are any plans to address this limitation in the future. The inability to handle large traces within zip files that require Zip64 support can be restrictive for certain workflows.

Could you please clarify:
Are there any plans to implement Zip64 support in upcoming releases?
If not, are there any recommended approaches or workarounds for handling large traces inside zip files, aside from extracting them manually beforehand?

Thank you for maintaining such an excellent tool, and I look forward to your insights on this!

Best regards,
Yohai

LalitMaganti:
The zip reader is a streaming parser that walks local file headers and
ignores the central directory, so zip64 support only needs to handle the
large-file encoding in the local headers rather than the zip64 end of
central directory structures.

Raise the accepted "version needed" from 2.0 to 4.5 and parse the local
header extra field instead of skipping it. When a Zip64 Extended
Information record (id 0x0001) replaces the 0xFFFFFFFF size sentinels,
read the 64-bit sizes it carries. For entries written in a streaming
fashion with a data descriptor, widen the descriptor to its 24-byte
zip64 form with 8-byte sizes. The compressed and uncompressed size
fields become 64-bit throughout.

This fixes issue #991 and applies to both trace_processor and the UI,
which parses zips via the same code compiled to Wasm. Encryption remains
unsupported.

Fixes: https://github.com/google/perfetto/issues/991


src/trace_processor/util/zip_reader.cc
```diff
@@ -211,7 +226,50 @@ base::Status ZipReader::TryParseFilename() {
...
+    uint16_t id = ReadAndAdvance<uint16_t>(&it);
+    uint16_t size = ReadAndAdvance<uint16_t>(&it);
+    if (end - it < size) {
+      break;
+    }
+    if (id == kZip64ExtraFieldId) {
+      cur_.is_zip64 = true;
+      const uint8_t* f = it;
+      const uint8_t* const fend = it + size;
+      if (cur_.hdr.uncompressed_size == kZip64SizeSentinel && fend - f >= 8) {
```
safayat-google: nits, if `cur_.hdr.uncompressed_size == kZip64SizeSentinel` but `fend - f < 8`, we can return a malformed ZIP error status.

safayat-google (approved):

src/trace_processor/util/zip_reader.cc
LalitMaganti: Really nice catch!

PR #6582, opened 2026-07-08
Optimize blocking calls cuj metrics with scoped materialization and i…

gopi3587:
Optimize blocking calls cuj metrics with scoped materialization and index push-down

The CUJ metrics originally relied on globally scanning all slices via `_android_critical_blocking_calls` and processed `thread_state` virtual tables using unoptimized inequality conditions, causing massive Cartesian nested loops on larger traces.

This CL dynamically accelerates query execution using two components: 1) Restricts materialization scopes cleanly replacing the global dependency with `_android_critical_blocking_calls_cuj`. This limits table mapping dynamically by employing nested CROSS JOIN logic on `relevant_upids`. Crucially, it attaches a structural Perfetto Index on (utid, ts) for instantaneous intersection verifications. 2) Replaces forced inner-loop table evaluations on virtual metrics mapping in `android_blocking_calls_cuj_per_frame_metric` via executing SQLite explicit `CROSS JOIN` conditions to force index-based evaluation on `utid`.

Execution on heavy traces cleanly drops delay times by 20x, without sacrificing precision (yields exact output parity).

Test: trace_processor_shell --run-metrics android_blocking_calls_cuj_metric,android_blocking_calls_cuj_per_frame_metric <trace>

Welcome to Perfetto!

For more details please see

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -18391,6 +18391,7 @@ filegroup {
         "src/trace_processor/perfetto_sql/stdlib/android/binder.sql",
         "src/trace_processor/perfetto_sql/stdlib/android/binder_breakdown.sql",
         "src/trace_processor/perfetto_sql/stdlib/android/bitmaps.sql",
+        "src/trace_processor/perfetto_sql/stdlib/android/blocking_calls_during_cujs.sql",
         "src/trace_processor/perfetto_sql/stdlib/android/broadcasts.sql",
         "src/trace_processor/perfetto_sql/stdlib/android/cpu/cluster_type.sql",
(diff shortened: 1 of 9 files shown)


LalitMaganti:
I'm opposed to the very concept of this PR for several reasons:
1) it's a hack, working around some slowness you are seeing rather than actually fixing the problem properly
2) the use of cross join is highly discouraged by SQLite developers *and* myself 
3) you might just be hiding a bug in query planning or a missing index in the cuj blocking code. Instead of fixing that we are just making a copy of the code which is the not the right thing to do. 

Please open a bug against me a with a trace and a slow query and I'll tackle doing the right fixes.

gopi3587:
Filed a follow up bug to come up with a better fix.

PR #6583, opened 2026-07-08
trace_processor: move stdlib tags and data-checks into module SQL

annamayzner:
The stdlib documentation is generated from the SQL, but module **tags** and
**data-availability check SQL** were still hand-maintained in Python dicts
(`MODULE_TAGS` in `stdlib_tags.py`; `MODULE_DATA_CHECK_SQL` / `TABLE_DATA_CHECK_SQL`
in `trace_data_checks.py`), keyed by module/table name away from the SQL they
describe. This moves them into each module's `.sql` as header comment directives
parsed by the stdlib parser:

```
-- @tags android, ipc
-- @data_check <sql>
-- @data_check_table <table> <sql>
```

- `docs_parse.py`: parse the directives onto `ParsedModule`
- `stdlib_parser.py`: read tags/checks from the parsed module
- `utils.py`: host `check_to_query`; delete `trace_data_checks.py`
- `check_sql_modules.py`: validate tags from the SQL; drop `check_orphaned_tags`
- `stdlib_tags.py`: keep only `VALID_TAGS` + table importance (importance stays hardcoded, out of scope here)

Intended to be a **noop**: the generated docs JSON, metadata JSON, entities, and
perfetto.dev markdown are byte-identical before/after.

Draft — open items:
- **Chrome stdlib** files are Copybara-rolled from Chromium and can't be edited here;
  their directives need to land in Chromium. (Handing this part off.)
- **Tag validation coverage:** the old import-time guard in `stdlib_tags.py` ran on every
  `check_sql_modules.py` invocation (including presubmit); tag validation now only runs
  under `--check-tags`, which presubmit doesn't pass. Adding `--check-tags` to presubmit
  would also surface ~23 pre-existing untagged modules.


petrcermak:
@annamayzner Would it be OK if we made the Chromium changes in the Chromium repository and let Copybara mirror them into Perfetto (instead of doing them as part of your change)?

LalitMaganti:
We're not landing this change yet anyway

petrcermak:
I see. Just let us know when you know what changes need to be made and in what timeframe and we'll make whatever changes are needed in Chromium 🙂.

PR #6587, opened 2026-07-08
trace_processor: fix OOB read in Instruments XML importer on unresolved ref=ids

lykan89:
## Problem

`InstrumentsXmlTokenizer` (the parser for Apple Instruments / `xctrace export`
`.trace` XML files) resolves `ref="N"` attributes back to previously-declared
`id="N"` entries via an id → vector-index mapping in `RowDataTracker`. This
mapping is derived directly from the trace file — untrusted input — but was
validated only with `PERFETTO_DCHECK`, which compiles out entirely in release
builds (`NDEBUG`).

Concretely: `<thread ref="999"/>` where `id="999"` was never declared (or any
tag missing its `id`/`ref` attribute, e.g. a bare `<thread/>`) resolves to
`id = 0` (`kNullId`). `RowDataTracker::GetThread()` then computes
`&threads_[id - 1]`, which underflows to a huge index and reads out of
bounds. The resulting wild pointer is dereferenced unconditionally both in
`RowParser::Parse()` and inside `InstrumentsXmlTokenizer::ElementStart()`, so
a single malformed element in a `.trace` file is enough to crash
`trace_processor`.

This is reachable from `trace_processor_shell`, the Python API, and the WASM
module behind ui.perfetto.dev — the Instruments trace type is auto-detected
purely from the first bytes of the file content (`trace_type.cc`), so no
special file extension or explicit opt-in is required.

The same unchecked-index pattern existed in `GetProcess`/`GetFrame`/
`GetBacktrace` (only `GetBinary` had a partial `kNullId` guard already), and
separately, `GetOrInsertByRef` could hit `strcmp(nullptr, "ref")` on an
attribute-less tag.

## Fix

- `RowDataTracker::GetThread/GetProcess/GetFrame/GetBacktrace/GetBinary` now
  go through a single bounds-checked helper (`At()`) and return `nullptr` for
  `kNullId` or any out-of-range id, instead of `PERFETTO_DCHECK` + unconditional
  indexing.
- Every call site in `row_parser.cc` and `instruments_xml_tokenizer.cc` that
  previously dereferenced these pointers unconditionally now checks for
  `nullptr` and skips just the malformed row/element, incrementing a new
  `stats::instruments_invalid_reference` counter instead of crashing.
- `GetOrInsertByRef` now explicitly handles missing/malformed `id`/`ref`
  attributes rather than relying on `PERFETTO_DCHECK`.
- Also guarded an adjacent `frames.rend() - 1` UB reachable when a
  `<backtrace>` has no `<frame>` children.

Well-formed traces are unaffected; only input that previously crashed the
process now results in the offending row/element being dropped and counted
in trace processor stats (queryable via the `stats` table), which matches
the existing convention used by other importers (systrace, hprof, etc.) for
handling malformed input gracefully.

## Testing

- Verified the failure mode and the fix against a standalone extraction of
  the exact id-resolution logic (same types/algorithm), built with `-DNDEBUG`
  to match a release build: reproduced an access violation before the fix,
  clean exit after, including a check that legitimate `id`/`ref` pairs still
  resolve correctly (no behavior change for well-formed traces).
- Not yet run through the full Perfetto build/test suite in this environment
  — will run `tools/ninja -C out/<cfg> trace_processor_shell
  perfetto_unittests` with `enable_perfetto_trace_processor_mac_instruments=true`
  before this is ready for review, and can add a `row_data_tracker_unittest.cc`
  covering these cases if useful.


LalitMaganti:
I believe https://github.com/google/perfetto/pull/6518 is trying to tackle the same issue. Can you please confirm if it is?

PR #6589, opened 2026-07-08
ui: Remap Mod+Shift+P to F1 on firefox

stevegolton:
The 'OpenCommandPalette' hotkey is currently bound to Mod+Shift+P which works fine in Chromium based browsers, but it's not overridable on Firefox - it always opens a new incognito window.

This patch introduces a a hotkey overlay subsystem. For now we simply define a single overlay if firefox is detected, which remaps the open palette command to F1 (similar to what the in-browser version of vscode does.

I.e.:

```js
const firefoxOverlay = {
  'dev.perfetto.OpenCommandPalette': '!F1',
};
```

### DIFF
--- a/ui/src/base/hotkeys.ts
+++ b/ui/src/base/hotkeys.ts
@@ -91,7 +91,20 @@ type Special =
   | ']'
   | ','
   | '.';
-export type Key = Alphabet | Number | Special;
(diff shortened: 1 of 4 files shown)


rsavitski (approved):
LGTM!

PR #6597, opened 2026-07-08
slice_track: ensure dataFrameSlot refetches for dynamic SQL (e.g. ftrace tracks)

rsavitski:
Change SliceTrack.useData() to fix staleness of raw ftrace tracks
whenever the event filter is changed. The tracks would continue showing
the old instants until the viewport was panned/zoomed.

This is because SliceTrack's loaded-slices cache (dataFrameSlot) does
not include sqlSource in its key, unlike the mipmap tablesSlot above it.
So because the ftrace track varies its SQL at runtime (and is the only?
track to do so), the mipmap tables were being updated, but cached data
was being reused for the instants themselves.


rsavitski:
PTAL

stevegolton (approved):
Good catch, thanks!

PR #6602, opened 2026-07-08
ui: restructure trace info page metadata and stats tabs

LalitMaganti:
The trace info page had a single "Info and Stats (advanced)" tab that
dumped the metadata table and the stats table together, with a hardcoded
priority allowlist for a handful of Chrome keys.

Split this into two purpose-built tabs:

- A new "Metadata" tab that groups the metadata table into curated
  sections (Trace, Device & OS, Benchmark, Chrome), each with a short
  description, and an "Other" catch-all for everything else. When
  nothing matches a curated section the "Other" header is dropped and
  the entries render as a plain list. Keys are matched by name or
  namespace prefix, replacing the previous cr-* allowlist.
  trace_config_pbtxt is hidden as it belongs on the Trace Config tab.

- The remaining stats table becomes its own "Statistics" tab.

Both tabs now render through the Grid widget with a dense styling
variant and a bold key column, replacing the hand-rolled HTML tables.
The combined "Info and Stats (advanced)" tab is removed.


ui/src/plugins/dev.perfetto.TraceInfoPage/tabs/metadata.ts
```diff
@@ -0,0 +1,255 @@
...
+  },
+  {
+    title: 'Chrome',
+    description: 'Metadata reported by Chrome.',
+    matches: (n) => n.startsWith('cr-'),
+    strip: 'cr-',
+  },
+];
+
+const OTHER_TITLE = 'Other';
```
sashwinbalaji, line 103: extract out ftrace_setup_errors from here

ui/src/plugins/dev.perfetto.TraceInfoPage/tabs/metadata.ts
```diff
@@ -0,0 +1,255 @@
+// Copyright (C) 2025 The Android Open Source Project
```
stevegolton: Nit: 2026

stevegolton (approved):

PR #6604, opened 2026-07-08
gn: Disable Zstd when building with Chromium

betasheet:
Disables enable_perfetto_zstd when build_with_chromium is true to unblock auto-roller.


betasheet:
Closing as #6595 already addresses this

sashwinbalaji:
Let's land it to unblock about size regression I will land a separate fix fior it.

sashwinbalaji (approved):

PR #6606, opened 2026-07-08
docs: add a guide for the display video frames feature

zezeozue:
Document the android.display.video data source end to end: the three ways to capture it (the on-device System Tracing toggle, the record page probe, and a raw trace config with the display_video_config options), and how to view it in the UI (the per-display track, hover preview, and playback with timeline sync and speed control), plus its cost on the device.

Also skip *.gif in the presubmit tab-character check, alongside the existing *.png/*.jpg, so the guide's animated screenshots don't trip it.


docs/data-sources/video-frames.md
```diff
@@ -0,0 +1,134 @@
...
+
+The **android.display.video** data source records what each physical
+display showed while a trace was being captured. Perfetto stores the
+frames as an encoded video stream inside the trace, and the UI adds a
+per-display timeline track that decodes them in the browser. You can
+hover the track to preview a frame, play the frames back like a video,
+and click any frame to line it up with the tracks below it.
+
+It records the actual contents of the screen, so it is meant for
+`userdebug` (debuggable) devices only, and any trace that contains it is
```
LalitMaganti, line 11: As discussed, would love if this restriction can be dropped.

docs/data-sources/video-frames.md
```diff
@@ -0,0 +1,134 @@
...
+  }
+}
+```
+
+| Option | Description |
+| --- | --- |
+| `scale` | Factor applied to each display's resolution before capture, e.g. `0.5` for half size or `0.25` for quarter. Lower scale means less encoder load and a smaller trace, at the cost of detail. |
+| `format` | `FORMAT_H264` (the default) or `FORMAT_HEVC`. HEVC produces a smaller stream at the same quality, but the device must support HEVC encoding to capture it and the browser must support HEVC decoding to preview it. |
+| `key_frame_interval_secs` | How often a keyframe is emitted. Smaller values make seeking snappier but grow the trace; larger values are more compact but slower to scrub. |
+| `max_stream_size_bytes` | A per-display cap on emitted bytes. When a display hits it, its stream is torn down (a size-cap error) rather than growing without bound. Left unset, the device applies a default cap of 256 MiB per display. |
```
LalitMaganti, line 100: Should we also have a section here explicitly explaining that due to technical limitations, we limit size in both on device and TP and pointing to a Github bug where people should comment and +1 if this should be prioritised to improve.

docs/toc.md
```diff
@@ -99,6 +99,7 @@
     - [Embedding the UI](visualization/embedding-the-ui.md) {.tag-android .tag-linux .tag-cpp-rust .tag-chrome .tag-perf}
     - [Debug Tracks](analysis/debug-tracks.md) {.tag-android .tag-linux .tag-cpp-rust .tag-chrome .tag-perf}
     - [Heap Dump Explorer](visualization/heap-dump-explorer.md) {.tag-android}
+    - [Display Video Frames](visualization/video-frames.md) {.tag-android}
```
LalitMaganti: I think this should actually be in the datasource category not in the UI. PLease also move the file there.

docs/toc.md
```diff
@@ -83,6 +83,7 @@
       - [Battery & Power](data-sources/battery-counters.md) {.tag-android}
       - [Android Game Interventions](data-sources/android-game-intervention-list.md) {.tag-android}
       - [Android Aflags](data-sources/android-aflags.md) {.tag-android}
+      - [Display Video Frames](data-sources/video-frames.md) {.tag-android}
```
LalitMaganti: I think this should be something like "Screen recording"

docs/data-sources/video-frames.md
```diff
@@ -0,0 +1,151 @@
+# Display Video Frames
```
LalitMaganti: This should also be called screen recording.

LalitMaganti (approved):

PR #6607, opened 2026-07-08
ui: allow vite to serve from any host

primiano:
Minor change: allows vite to be put behind a reverse proxy.
By default vite only accepts localhost as a host name,
and refuses any other host to prevent DNS rebind attacks.
This is not an interesting case for us (The code is open source,
and it's client only), so there is no risk, just annoyance.

Bug: b/532849066

### DIFF
--- a/ui/build.mjs
+++ b/ui/build.mjs
@@ -95,6 +95,7 @@ const cfg = {
   onlyWasmMemory64: false,
   wasmModules: [],
   crossOriginIsolation: false,
+  allowAllHosts: false,
   testFilter: '',
   noOverrideGnArgs: false,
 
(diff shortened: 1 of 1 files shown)


stevegolton:
I'd hesitate to add this policy by default. The [https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6](vuln) exists for a reason. I'd rather not allow any malicious websites to interact with my nor anyone else's dev server - even if it is open source, what folks are currently working on might not be.

I think a much better approach would be to either add a new `--allowed-hosts` option to the build script which you can set to a specific host or just '*' to permit all.

primiano:
> I'd hesitate to add this policy by default. The [https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6](vuln) exists for a reason. I'd rather not allow any malicious websites to interact with my nor anyone else's dev server - even if it is open source, what folks are currently working on might not be.
> 
> I think a much better approach would be to either add a new `--allowed-hosts` option to the build script which you can set to a specific host or just '*' to permit all.

Done, I added a cmdline argument, (which then I can override in the env file)

stevegolton (approved):

PR #6611, opened 2026-07-08
ui: video frame plugin fixes

zezeozue:
- Details panel: don't await player.seek() in load(). Awaiting held the
    panel in its loading state, which remounts and blanks the <canvas>;
    seek() cancels superseded decodes itself, so the preview stays painted
    during keyboard nav.
  - Track: render each frame as a bounded slice (dur = LEAD(ts) - ts, and 0
    for the last frame) instead of an unbounded, fading dur = -1 slice.
  - Player: don't decode a frame with no preceding key frame. nearestKey
    walked back to index 0 even when it was a delta frame (e.g. a ring buffer
    whose GOP start was overwritten); feeding a delta chunk after configure()
    throws "A key frame is required after configure()". nearestKey now returns
    -1 and the decode/playback paths bail, leaving a blank preview instead of
    throwing.

### DIFF
--- a/ui/src/plugins/dev.perfetto.VideoFrames/video_frame_details_panel.ts
+++ b/ui/src/plugins/dev.perfetto.VideoFrames/video_frame_details_panel.ts
@@ -48,7 +48,7 @@ export class VideoFrameDetailsPanel implements TrackEventDetailsPanel {
     // Don't await: while load() is pending the panel sits in its loading
     // state, which remounts and blanks the <canvas>. seek() cancels stale
     // decodes itself.
-    void this.player.seek(sel.eventId);
+    this.player.seek(sel.eventId);
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


ui/src/plugins/dev.perfetto.VideoFrames/video_frame_details_panel.ts
```diff
@@ -45,7 +45,10 @@ export class VideoFrameDetailsPanel implements TrackEventDetailsPanel {
     if (this.player.playing) return;
     await this.player.ensureFramesLoaded();
     if (this.player.frames[this.player.currentIdx]?.id === sel.eventId) return;
-    await this.player.seek(sel.eventId);
+    // Don't await: while load() is pending the panel sits in its loading
+    // state, which remounts and blanks the <canvas>. seek() cancels stale
+    // decodes itself.
+    void this.player.seek(sel.eventId);
```
stevegolton: I don't think we need the void here - eslint isn't configured to error on unhandled promises.

stevegolton (approved):

PR #6612, opened 2026-07-08
ui: Wrap errors in CommandError for command execution

safayat-google:
When a query crashes in the UI, it can be hard to know who to triage the issue to if the query was executed by a plugin or a custom macro.

This change catches any error thrown during command execution and wraps it in a CommandError containing the command's metadata (ID, name, source) and the original error as its cause.

If commands are nested (e.g. a macro running a plugin command), CommandError instances stack recursively. The toString() implementation formats the full execution chain down to the root query crash, revealing exactly which macro/plugin triggered the query failure in the crash report.

Bug: 504577537


ui/src/core/command_manager.ts
```diff
@@ -177,9 +187,22 @@ export class CommandManagerImpl implements CommandManager {
...
+          raf.scheduleFullRedraw();
+        });
+      }
+      this.activeCommandStack.pop();
+      raf.scheduleFullRedraw();
+      return res;
+    } catch (err) {
+      this.activeCommandStack.pop();
+      throw err;
+    }
```
primiano: Hmm do we need to maintain a stack ourselves?
COuld this just be a try/catch ardoun an await, somethiing like

```
  const commandId = cmd;
  try {
    await cmd.callback(args);   // await the whole returned chain
  } catch (e) {
    throw e; // Wrap this somehow adding the commandid
  } finally {
      raf.scheduleFullRedraw()
   }
}
```

primiano (commented):
Hmm would like to hear @stevegolton opinions on this as well.
My theory is that:

- If the plugin implementation does awaits correctly, we don' tneed to maintain a stack, we need only a try/catch around the async invocation
- If the plugin implementation forgets an await (so does accidentally a fire and forget promise), your thing will also pop() so won't be better

So maybe just async-try-catch shoudl do the trick?


The other option I can think about is that we store the last recent command/plugins id at the engine level

ui/src/core/command_manager.ts
safayat-google: Removed the command stack information and just included the one where the crash came from.

ui/src/trace_processor/query_result.ts
```diff
@@ -209,9 +209,23 @@ export function decodeInt64Varint(buf: Uint8Array, pos: number): bigint {
...
+  toString(): string {
+    const owner = this.source ? `\nSource/Owner: ${this.source}` : '';
+    return `Command/Macro: ${this.name} (${this.id})${owner}`;
+  }
+}
+
 export interface QueryErrorInfo {
   query: string;
   tag?: string; // The EngineProxy tag.
+  activeCommand?: ActiveCommandInfo;
```
stevegolton: Sources under `ui/src/trace_processor/` should not know anything about commands.

ui/src/core/command_manager.ts
```diff
@@ -172,14 +173,25 @@ export class CommandManagerImpl implements CommandManager {
...
     }
     const cmd = this.registry.get(id);
-    const res = cmd.callback(...args);
-    Promise.resolve(res).finally(() => raf.scheduleFullRedraw());
-    return res;
+    try {
+      return await cmd.callback(...args);
+    } catch (err) {
+      if (err instanceof QueryError) {
+        err.queryErrorInfo.activeCommand = new ActiveCommandInfo(
```
stevegolton: IIUC, if we have a macro calling a command, the activeCommand will be replaced as the error bubbles up the stack, meaning the outermost try-catch will 'win'. In this case this will be the macro invocation, so we'll only see the information about the macro in the error message.

This brings about an interesting issue of who's at fault if a command crashes? Is it the command itself, or is it the macro that invoked it. I think there's no way to actually be sure who we want to point the finger at, so it might make sense to print the entire stack.

So it might make sense to:
1: Catch all errors (not just QueryError).
2: Wrap the error inside a new CommandError, which takes the cmd details (id, name, source) and the original error, and throw the CommandError.
3: The CommandError should print all wrapped errors recursively - that way we'll see details about all commands invoked in the stack.

stevegolton (changes requested):
Concept looks good but I've suggested some ways to:

a) Attach reporting to all errors - not just query errors
b) Report nested command/macro invocations.

ui/src/core/command_manager.ts
safayat-google: Updated. Here's an example format.
```
Command/Macro: Test Error Parent (Runs Leaf) (dev.perfetto.TestErrorParent)
Caused by: Command/Macro: Test Error Leaf (dev.perfetto.TestErrorLeaf)
Caused by: QueryError: Query failed with syntax error
EngineTag: example_tag
Query:
SELECT * FROM invalid_table
```

ui/src/trace_processor/query_result.ts
safayat-google: removed it.

stevegolton (approved):

PR #6627, opened 2026-07-09
perfetto: list other tracing sessions in trace

sashwinbalaji:
Optionally record the lifecycle state changes of *other* tracing
sessions active while a trace is being recorded, to help diagnose
field issues caused by unexpected concurrent sessions.

- Add the ConcurrentSessionEvent packet: state (mirrors the service's
  session state machine), unique_session_name, service-local session
  id (deliberately not the trace UUID), consumer uid and number of
  data sources.
- Route every session state change through a single SetSessionState()
  helper. It broadcasts the change into the other opted-in sessions
  that are still recording (i.e. CONFIGURED, STARTED or
  DISABLING_WAITING_STOP_ACKS; read-only clones and already-disabled
  sessions are skipped), appending to a bounded per-session buffer
  that is drained on ReadBuffers().
- On creation, a session snapshots the current state of the other
  active sessions, so its trace records what was already running when
  it started. Each event is timestamped with when that session entered
  its current state.
- Cloned sessions inherit the source's events and record a terminal
  DISABLED on teardown.
- Opt-in via
  TraceConfig.BuiltinDataSource.enable_concurrent_session_events.
- Add service unittests.

### DIFF
--- a/protos/perfetto/trace/perfetto/concurrent_session_event.proto
+++ b/protos/perfetto/trace/perfetto/concurrent_session_event.proto
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Emitted by the tracing service to record a state change of another tracing
+// session that was active while this trace was being recorded. Opt-in via
+// TraceConfig.BuiltinDataSource.enable_concurrent_session_events.
+//
+// One packet per state change; its time is carried by the enclosing
+// TracePacket's timestamp (CLOCK_BOOTTIME). Events of the same session share
+// the same |session_id|. The trace UUID is deliberately not included.
+//
+// Introduced in v58 / Android 26Q3+.
+message ConcurrentSessionEvent {
+  // Mirrors the tracing service's internal session state machine
+  // (TracingSession::State in the service): keep these two enums in sync.
+  // These values are part of the trace wire format and must stay stable: never
+  // renumber or reuse an existing value, only append new ones at the end.
+  enum State {
+    STATE_UNSPECIFIED = 0;
+    // Not recording (initial and final state).
+    STATE_DISABLED = 1;
+    // Configured but waiting for a start signal (e.g. a trigger).
+    STATE_CONFIGURED = 2;
+    // Actively recording data.
+    STATE_STARTED = 3;
+    // Stopping, waiting for data sources to ack and drain their final data.
+    STATE_DISABLING_WAITING_STOP_ACKS = 4;
+    // A read-only snapshot of another session, created by a clone operation.
+    STATE_CLONED_READ_ONLY = 5;
+  }
+  // The state the session transitioned to at this timestamp.
+  optional State state = 1;
+
+  // unique_session_name from the session's TraceConfig. May be empty.
+  optional string session_name = 2;
+
+  // Service-local monotonic counter identifying the session. Shared by all
+  // state changes of one session, never reused within a service instance.
+  optional uint64 session_id = 3;
+
+  // Unix uid of the consumer that started the session.
+  optional int32 consumer_uid = 4;
+
+  // Number of data sources in the session at the time of the state change.
+  optional uint32 num_data_sources = 5;
+}
--- a/src/tracing/service/tracing_service_impl.cc
+++ b/src/tracing/service/tracing_service_impl.cc
@@ -1093,7 +1093,17 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
                                           weak_runner_.task_runner()))
            .first->second;
 
-  PopulateConcurrentTracingSessions(tracing_session);
+  // Snapshot the current state of every other session into the newly created
+  // one, so its trace records which sessions were already active when it
+  // started. Each snapshot is timestamped with when that session entered its
+  // current state.
+  if (cfg.builtin_data_sources().enable_concurrent_session_events()) {
+    for (auto& [src_id, src] : tracing_sessions_) {
+      if (src_id == tsid)
+        continue;
+      tracing_session->AddConcurrentSessionEventWithLimit(src);
+    }
+  }
 
   tracing_session->trace_uuid = uuid;
 
@@ -4296,7 +4306,6 @@ void TracingServiceImpl::SetSessionState(TracingSession* session,
   // Broadcast this state change into every other session that opted into
   // concurrent session events, so their trace logs that this session changed
   // state while they were running.
-  std::optional<TracingSession::ConcurrentSessionEvent> event;
   for (auto& [dst_id, dst] : tracing_sessions_) {
     if (!dst.config.builtin_data_sources().enable_concurrent_session_events())
       continue;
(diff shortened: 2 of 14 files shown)


protos/perfetto/trace/perfetto/concurrent_session_event.proto
```diff
@@ -0,0 +1,60 @@
...
+//
+// One packet per state change; its time is carried by the enclosing
+// TracePacket's timestamp (CLOCK_BOOTTIME). Events of the same session share
+// the same |session_id|. The trace UUID is deliberately not included.
+//
+// Introduced in v58 / Android 26Q3+.
+message ConcurrentSessionEvent {
+  // Mirrors the tracing service's internal session state machine.
+  enum State {
+    STATE_UNSPECIFIED = 0;
```
primiano, line 36: Add a comment saying these must be kept in sync with the matching enum in TracingServiceImpl. BUT at the same time we cannot renumber these, the numbering must be stable here

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4280,114 @@ void TracingServiceImpl::EmitLifecycleEvents(
     SerializeAndAppendPacket(packets, std::move(pair.second));
 }
 
+void TracingServiceImpl::SetSessionState(TracingSession* session,
+                                         TracingSession::State new_state) {
+  PERFETTO_DCHECK_THREAD(thread_checker_);
+  PERFETTO_DCHECK(session->state != new_state);
```
primiano: I'd remove this TBH, doesn't feel a programming error, can cause some issue later on. There is really no functional need you need this. WOrst case it will record the same event twice, which is no tthe end of the world

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4280,114 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+
+  const int64_t now_ns = clock_->GetBootTimeNs().count();
+  session->state = new_state;
+  if (new_state == TracingSession::STARTED) {
+    session->started_boot_ns = now_ns;
+    BackfillConcurrentSessionEvents(session);
```
primiano: I would do this in EnableTracing, soon after where we create the TracingSession* in &tracing_sessions_.emplace().

Also i'd call it something like PopulateConcurrentTracingSessions() or ReadConcurrentTracingSessions()

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4280,114 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+  }
+  RecordConcurrentSessionEvent(session, now_ns);
+}
+
+void TracingServiceImpl::RecordConcurrentSessionEvent(TracingSession* changed,
```
primiano: can you add a comment (here or just before the for loop below) to remeind us that what we want to do here is broadcast the insertion of the changed state into all the other sessions (if they are interested into the events).


Also can we inline RecordConcurrentSessionEvent inside the above?
If I codesearched right, there is only one callsite for this.
I feel it's easier to readon about the code if we reduce the number of functions in this case, as there are too many function with similar name (Record, Set, Backfill..)

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4280,114 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+                                                      int64_t timestamp) {
+  TracingSession::ConcurrentSessionEvent event;
```
primiano: As a habit: {} (`event{}`) so it gets zero initialized if you ever forget something below.
In this specific case it's not needed because you did correctly zero-initialize in the struct fields every member that is a POD.
But i feel this style is in a large scale safer, becuase then even if people forget to initialize an int field, the curly braces {} here forces zero initialziation of the struct anyways

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4280,114 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+
+void TracingServiceImpl::BackfillConcurrentSessionEvents(
+    TracingSession* session) {
+  if (!session->config.builtin_data_sources()
+           .enable_concurrent_session_events()) {
+    return;
+  }
+  for (auto& id_and_session : tracing_sessions_) {
+    TracingSession& peer = id_and_session.second;
+    if (peer.id == session->id || peer.state != TracingSession::STARTED)
```
primiano: I think here we should NOT filter these and list them all

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4280,114 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+}
+
+void TracingServiceImpl::BackfillConcurrentSessionEvents(
+    TracingSession* session) {
+  if (!session->config.builtin_data_sources()
+           .enable_concurrent_session_events()) {
+    return;
+  }
+  for (auto& id_and_session : tracing_sessions_) {
+    TracingSession& peer = id_and_session.second;
```
primiano: rather than peer, i feel this becomes a bit more readable if you call:
session -> dst
peer -> src

I got confused several times while reading this code, because "session" isn't really specific, everything is a session here

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4280,114 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+    if (peer.id == session->id || peer.state != TracingSession::STARTED)
+      continue;
+
+    TracingSession::ConcurrentSessionEvent event;
+    event.timestamp = peer.started_boot_ns;
+    event.session_id = peer.id;
+    event.state = TracingSession::STARTED;
```
primiano: change this to be src.state

src/tracing/service/tracing_service_session.h
```diff
@@ -212,6 +212,31 @@ struct TracingSession {
...
+    State state = DISABLED;           // The state the peer transitioned to.
+    uid_t consumer_uid = 0;           // Uid of the peer's consumer.
+    uint32_t num_data_sources = 0;    // Data sources in the peer session.
+    std::string name;                 // Peer unique_session_name (or empty).
+  };
+
+  // Appends |event|, dropping it if the buffer is full. The cap bounds
+  // memory, as the buffer is drained only on ReadBuffers().
+  void AddConcurrentSessionEvent(ConcurrentSessionEvent event) {
+    static constexpr size_t kMaxConcurrentSessionEvents = 256;
```
primiano: i'd make it 4096, as in 24h traces we can have many coming and going

src/tracing/service/tracing_service_session.h
```diff
@@ -212,6 +212,31 @@ struct TracingSession {
...
+    TracingSessionID session_id = 0;  // Peer session id.
+    State state = DISABLED;           // The state the peer transitioned to.
+    uid_t consumer_uid = 0;           // Uid of the peer's consumer.
+    uint32_t num_data_sources = 0;    // Data sources in the peer session.
+    std::string name;                 // Peer unique_session_name (or empty).
+  };
+
+  // Appends |event|, dropping it if the buffer is full. The cap bounds
+  // memory, as the buffer is drained only on ReadBuffers().
+  void AddConcurrentSessionEvent(ConcurrentSessionEvent event) {
```
primiano: can you call it "WithSaturation" (or WithLmit) so at the call site we remember that there is a magic limit in this function.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4283,131 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+  // state.
+  for (auto& [src_id, src] : tracing_sessions_) {
+    if (src_id == dst->id)
+      continue;
+
+    // Skip DISABLED sessions: they've already finished (a DISABLED session
+    // lingers in the map until the consumer calls FreeBuffers) and won't emit a
+    // follow-up transition, so snapshotting one would leave a dangling terminal
+    // event for a session |dst| never really overlapped with.
+    if (src.state == TracingSession::DISABLED)
```
primiano: i'd prob remove this and populate anyways so we can spot bugs in clients not freeing up the session

src/tracing/service/tracing_service_impl.cc
```diff
@@ -4269,6 +4283,131 @@ void TracingServiceImpl::EmitLifecycleEvents(
...
+
+    TracingSession::ConcurrentSessionEvent event{};
+    event.timestamp = src.current_state_start_ns;
+    event.session_id = src.id;
+    event.state = src.state;
+    event.consumer_uid = src.consumer_uid;
+    event.num_data_sources =
+        static_cast<uint32_t>(src.data_source_instances.size());
+    event.name = src.config.unique_session_name();
+    dst->AddConcurrentSessionEventWithLimit(std::move(event));
```
primiano: i'd probably pass the TracingSession* as argument to AddConcurrent, and do the serialization in there

primiano (approved):

PR #6628, opened 2026-07-09
proto_merger: allow passthrough fields

BogdanLionte:


src/tools/proto_merger/allowlist.cc
```diff
@@ -19,11 +19,19 @@
 #include <google/protobuf/descriptor.pb.h>
 
 #include "perfetto/ext/base/string_utils.h"
+#include "protos/perfetto/proto_filtering/proto_filter_options.pb.h"
```
LalitMaganti: This is a bit of a hack. Proto filtering and proto merging passthrough are sort of orthogonal to each other and while for our usage they are both the same semantically, they don't have to be. 

I suggest we introduce a new field options for automatic allow listing for merging instead of reusing the one for filtering. At the very least if we do end up using both, this field options needs to be refactored into a common place instead of merger depending on a filtering proto.

BogdanLionte: I extracted the passthrough option into a common location.

protos/perfetto/common/passthrough.proto
```diff
@@ -0,0 +1,26 @@
...
+
+package perfetto.protos;
+
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.FieldOptions {
+  // If true, this message field and its entire sub-tree are treated as fully
+  // allowed by default, bypassing individual field-level inspection or manual
+  // allowlisting. Only applies to message fields.
+  optional bool passthrough = 73400002;
```
LalitMaganti: proto_filter_merge_passthrough

protos/perfetto/common/passthrough.proto
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.FieldOptions {
+  // If true, this message field and its entire sub-tree are treated as fully
```
LalitMaganti, line 25: I would add some commentary here that this is realy designed for tooling which works with filtering of traces/merging protos together.

BogdanLionte: Done

protos/perfetto/common/passthrough.proto
BogdanLionte: Done

LalitMaganti (approved):

PR #6632, opened 2026-07-09
[chrome] Fix chrome_browser_main_thread_duration performance issue

etiennep-chromium:
This PR fixes performance issue in chrome_browser_main_thread_duration metric query.

Views (chrome_subprocess_types, chrome_process, chrome_thread): Added ORDER BY on upid/utid to allow SQLite to optimize joins across views containing DISTINCT and subqueries. Additionally, removed redundant duplicate joins on the thread table inside the chrome_thread view.
Metric (chrome_browser_main_thread_duration): Changed LEFT JOINs to JOINs, giving the query planner the flexibility to reorder tables and execute index searches on slice(track_id).


src/trace_processor/metrics/sql/chrome/chrome_browser_main_thread_duration.sql
```diff
@@ -23,14 +23,17 @@ SELECT MAX(len) AS duration_ms
...
-  LEFT JOIN thread_track tt ON s.track_id = tt.id
-  LEFT JOIN chrome_thread t ON t.utid = tt.utid
-  LEFT JOIN chrome_process p ON t.upid = p.upid
-  LEFT JOIN track tr ON tt.id = tr.id
-  LEFT JOIN args a ON tr.source_arg_set_id = a.arg_set_id
-  WHERE (t.canonical_name = "CrProcessMain" OR t.canonical_name = "CrBrowserMain")
-    AND p.process_type = "Browser"
-    AND a.key = "is_root_in_scope"
-    AND a.int_value = 1
+  WHERE s.track_id IN (
```
LalitMaganti: This rewrite should be unnecessary. If you just rewrite LEFT JOIN -> JOIN, you can just keep the joins and it should be essentially just as good.

etiennep-chromium: Very true, but also needed to fix issues with chrome_thread / chrome_process not being able to index properly

LalitMaganti (approved):

PR #6633, opened 2026-07-09
[UI] Collapse compressed track_event tracks by default

etiennep-chromium:
The UI for collapsing tracks exists already. 
This PR 
- makes expanding a collapse track easier with "click to expand" directly on the timeline (and changed cursor) 
- collapse compressed tracks by default (multiple track ids shown as a single track in UI, which are often very long in chrome)


LalitMaganti:
This cannot be a generic thing as it will break countless open source folks using the UI.

etiennep-chromium:
> will break countless open source folks

I think this is mitigated by making uncollapsing easier?

stevegolton:
As Lalit said, this is a big change that affects everyone who uses async tracks via track event. If you think it's the right direction to go for chrome then we need a way of only applying this to chrome traces, which is a larger refactor.

LalitMaganti:
With the change now, I'm more ambivalent though I'm still always scared of introducing backwards incompatible changes like this on such a core feature.

How about we make this an opt-in feature at the track descriptor level?

etiennep-chromium:
> How about we make this an opt-in feature at the track descriptor level?

Do you mean in track descriptor proto? I feel like I don't want to bake this in the trace.
Maybe a UI flag can suffice?

PR #6651, opened 2026-07-09
perfetto: unify trace attributes across config, packets and manifest

### LINKED ISSUE #3495: Feature Request: Metadata/notes data source

sashwinbalaji:
Trace attributes (key/value pairs describing a trace) could only be
written as TraceAttributes packets. Unify the other ways of attaching
them on the same message and query surface:

* Replace TraceConfig.notes and `perfetto --add-note` with
  TraceConfig.trace_attributes and `--add-attribute`, embedding the
  TraceAttributes message in the config. Notes were never parsed by
  trace_processor so nothing can depend on them; field 46 is reserved.
* Ingest config attributes through the same path as TraceAttributes
  packets; both become trace_attribute.<key> rows of the metadata
  table.
* Let perfetto_manifest files annotate an archive (top-level or
  per-file) with an `attributes` section, stored as
  manifest_attribute.<key> rows: a separate namespace because they
  annotate the archive, not the recorded trace.
* Make SetDynamicMetadata update duplicate keys in place so "last
  value wins" actually holds. Chrome importers keep their historical
  insert-always behavior via a new AppendDynamicMetadata.
* Show both namespaces as sections on the UI's trace info page;
  update docs; add unit and diff tests for all paths.

Fixes: https://github.com/google/perfetto/issues/3495


docs/reference/perfetto-manifest.md
```diff
@@ -225,6 +228,43 @@ merged timeline causes those events to be dropped, counted in the
...
+## {#attributes} attributes
+
+Arbitrary key/value pairs annotating the archive: a benchmark name, a run
+id, a build id, and so on. Each entry becomes a `manifest_attribute.<key>`
+row of the `metadata` table.
+
+The namespace is deliberately separate from `trace_attribute.*`
+([TraceAttributes](/protos/perfetto/common/trace_attributes.proto)): those
+are properties recorded in a trace itself, while manifest attributes are
+added when the archive is packed and say nothing about what happened at
```
LalitMaganti: Say nothing about what happened at recording time is not accurate. It's more they reflect attributes of the full archive itself.

docs/reference/perfetto-manifest.md
```diff
@@ -225,6 +228,43 @@ merged timeline causes those events to be dropped, counted in the
...
+```json
+{
+  "perfetto_manifest": {
+    "version": 1,
+    "attributes": {"benchmark": "startup", "run_id": 42},
+    "files": [
+      {
+        "path": "phone.pftrace",
+        "machine": {"name": "phone"},
+        "attributes": {"role": "client"}
```
LalitMaganti: I don't want file attributes right now. Maybe we can have this in the future but for now just confused things.

docs/reference/perfetto-manifest.md
```diff
@@ -246,6 +286,9 @@ After import, the manifest's effects are visible in the trace:
   2^32, deliberately outside the 32-bit space of ids embedded in trace
   packets.
 - `trace_time` sets the `trace_time_clock_id` key in the `metadata` table.
+- Every `attributes` entry becomes a `manifest_attribute.<key>` row in the
```
LalitMaganti, line 275: We can remove the second part

src/trace_processor/importers/common/metadata_tracker.h
```diff
@@ -45,10 +45,16 @@ class MetadataTracker {
...
   MetadataId AppendMetadata(metadata::KeyId key, Variadic value);
 
-  // Sets a metadata entry using any interned string as key.
-  // Returns the id of the new entry.
+  // Sets a metadata entry using any interned string as key, overwriting any
+  // existing entry with the same key (last value wins).
+  // Returns the id of the entry.
   MetadataId SetDynamicMetadata(StringId key, Variadic value);
 
+  // Appends a metadata entry using any interned string as key; repeated keys
```
LalitMaganti, line 53: Why do we need this? Imo repeated keys should be an error

src/trace_processor/types/trace_manifest_state.h
```diff
@@ -71,10 +72,17 @@ struct TraceManifestState {
...
 
+  // A parsed trace attribute: key plus a string or int64 value (see
+  // protos/perfetto/common/trace_attributes.proto).
+  using Attribute = std::pair<std::string, std::variant<int64_t, std::string>>;
+
   struct FileEntry {
     // Exact path of the member within the archive.
     std::string path;
     std::optional<ClockOverride> clock_override;
+    // The entry's `attributes` block, applied as manifest_attribute.* metadata
```
LalitMaganti: Here and throughout pr - I'd reduce the amount of comments like this not adding much

src/trace_processor/importers/common/metadata_tracker.h
sashwinbalaji: kept it only for legacy ones

LalitMaganti (approved):

PR #6655, opened 2026-07-10
ui: Support pinning tracks and drawing all flows for a selected app

jingjiewang-twn:
Previously, the Android Input Lifecycle plugin only allowed tracing the lifecycle of a single selected input event at a time, and did not have a way to automatically pin all input pipeline tracks for a specific app.

This PR adds two new capabilities to the Android Input Lifecycle plugin to improve trace analysis:

1) Drawing all input flows for a selected app:
   Adds commands "Android: Draw all input arrows (With speculative)" and
   "Android: Draw all input arrows (No speculative)". These commands
   display all input flow arrows for the currently selected app, helping
   the user visualize the global flow of inputs at once.

2) Pinning tracks for a selected app:
   Adds a command "Android: Pin input pipeline related tracks" which
   prompts the user to select an app, and automatically pins all related
   tracks in the input pipeline.

Additionally, this PR refactors index.ts to clean up internal state management and extract reusable helper methods:
- extractConnectionsFromRows(): Extracts arrow connections from row datasets.
- useRowState() & applyInitialSelection(): Simplifies rendering logic in the Mithril component.
- getUniqueInputApps() & getAllInputTrackUris(): Handles SQL retrieval of apps and tracks from trace processor.

### DIFF
--- a/ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
+++ b/ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
@@ -19,6 +19,7 @@ import {
   STR_NULL,
   type Row,
   type SqlValue,
+  type QueryResult as DbQueryResult,
 } from '../../trace_processor/query_result';
 import {type duration, Time, Duration} from '../../base/time';
 import {
@@ -146,7 +147,86 @@ export class AndroidInputEventSource {
     `;
 
     const result = await this.trace.engine.query(sql);
+    return this.mapResultToRows(result);
+  }
 
+  async getRowsForApp(
+    processName?: string,
+    excludeSpeculative = false,
+  ): Promise<InputChainRow[]> {
+    const processFilter = processName ? `'${processName}'` : 'NULL';
+    const speculativeFilter = excludeSpeculative ? '1' : '0';
+
+    const baseQuery = `
+      SELECT
+        e.input_event_id AS input_id,
+        e.event_channel AS channel,
+        e.end_to_end_latency_dur AS total_latency,
+        e.event_time AS event_time,
+        e.read_time AS ts_reader,
+        e.dispatch_ts AS ts_dispatch,
+        e.receive_ts AS ts_receive,
+        s_cons.ts AS ts_consume,
+        s_frame.ts AS ts_frame,
+        s_read.id AS id_reader,
+        s_read.track_id AS track_reader,
+        s_read.dur AS dur_reader,
+        s_disp.id AS id_dispatch,
+        e.dispatch_track_id AS track_dispatch,
+        s_disp.dur AS dur_dispatch,
+        s_recv.id AS id_receive,
+        e.receive_track_id AS track_receive,
+        s_recv.dur AS dur_receive,
+        s_cons.id AS id_consume,
+        s_cons.track_id AS track_consume,
+        s_cons.dur AS dur_consume,
+        s_frame.id AS id_frame,
+        s_frame.track_id AS track_frame,
+        s_frame.dur AS dur_frame,
+        e.is_speculative_frame
+      FROM android_input_events AS e
+      LEFT JOIN slice AS s_read
+        ON s_read.ts = e.read_time
+        AND s_read.track_id != 0
+        AND s_read.name GLOB 'UnwantedInteractionBlocker::notifyMotion*'
+      LEFT JOIN slice AS s_disp
+        ON s_disp.ts = e.dispatch_ts
+        AND s_disp.track_id = e.dispatch_track_id
+      LEFT JOIN slice AS s_recv
+        ON s_recv.ts = e.receive_ts
+        AND s_recv.track_id = e.receive_track_id
+      LEFT JOIN _input_consumers_lookup AS s_cons
+        ON s_cons.cookie = e.event_seq
+      LEFT JOIN _frame_choreographer_lookup AS s_frame
+        ON s_frame.frame_id = CAST(e.frame_id AS LONG)
+      WHERE e.event_channel LIKE '%/%'
+        AND (e.process_name = ${processFilter} OR ${processFilter} IS NULL)
+        AND (${speculativeFilter} = 0 OR e.is_speculative_frame = 0 OR e.is_speculative_frame IS NULL)
+    `;
+
+    let selectCols = 'a_evt.*';
+    let joinSql = '';
+
+    for (const ext of this.activeExtensions) {
+      const spec = ext.getSqlJoinSpec();
+      const alias = spec.tableAlias ?? spec.tableName;
+      selectCols += `, ${alias}.*`;
+      joinSql += ` LEFT JOIN ${spec.tableName}${spec.tableAlias ? ` AS ${spec.tableAlias}` : ''} ON ${spec.joinOn}`;
+    }
+
+    const sql = `
+      SELECT ${selectCols}
+      FROM (${baseQuery}) a_evt
+      ${joinSql}
+    `;
+
+    const result = await this.trace.engine.query(sql);
+    const rows = this.mapResultToRows(result);
+    await this.enrichAllDepths(rows);
+    return rows;
+  }
+
+  private mapResultToRows(result: DbQueryResult): InputChainRow[] {
     const rows: InputChainRow[] = [];
     let index = 0;
 
--- a/ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
+++ b/ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
@@ -124,6 +124,24 @@ export class AndroidInputEventSource {
     return AndroidInputEventSource.getStageSpecs(this.activeExtensions);
   }
 
+  /**
+   * Given a list of input lifecycle rows, extracts all track URIs and orders
+   * them chronologically according to stage sequence numbers.
+   */
+  getTrackUrisSortedByStage(rows: InputChainRow[]): string[] {
+    const specs = this.getStageSpecs();
+    const trackUris: string[] = [];
+    for (const spec of specs) {
+      for (const row of rows) {
+        const stageData = row.stagesData.get(spec.key);
+        if (stageData?.nav?.trackUri) {
+          trackUris.push(stageData.nav.trackUri);
+        }
+      }
+    }
+    return Array.from(new Set(trackUris));
+  }
+
   private async fetchRows(sliceId: number): Promise<InputChainRow[]> {
     const baseQuery = `SELECT * FROM _android_input_lifecycle_by_slice_id(${sliceId})`;
 
@@ -202,6 +220,7 @@ export class AndroidInputEventSource {
       WHERE e.event_channel LIKE '%/%'
         AND (e.process_name = ${processFilter} OR ${processFilter} IS NULL)
         AND (${speculativeFilter} = 0 OR e.is_speculative_frame = 0 OR e.is_speculative_frame IS NULL)
+      ORDER BY e.event_time ASC
     `;
 
     let selectCols = 'a_evt.*';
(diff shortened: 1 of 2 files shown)


ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -85,67 +93,200 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
...
+      this.allEventConnections = [];
+      this.activeExcludeSpeculative = undefined;
+      m.redraw();
+      return;
+    }
+
+    const rows = await source.getRowsForApp(this.pinnedApp, excludeSpeculative);
+    this.allEventConnections = this.extractConnectionsFromRows(rows);
+    this.activeExcludeSpeculative = excludeSpeculative;
+    m.redraw();
```
ivan-chong: Do we need this? From my understanding we generally avoid calls to m.redraw() unless absolutely necessary.

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -85,67 +93,200 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
...
-  }
+        let selectedApp = apps[0];
 
-  private applyInitialSelection(trace: Trace, rows: InputChainRow[]) {
-    const selection = trace.selection.selection;
-    if (selection.kind !== 'track_event') return;
+        await showModal({
+          title: 'Pin input pipeline tracks for app',
+          icon: 'help',
+          content: () => {
```
ivan-chong: Are we able to use existing Perfetto widgets for this?

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -85,67 +93,200 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
...
+    return apps;
+  }
+
+  private async getAllInputTrackUris(
+    trace: Trace,
+    processName?: string,
+  ): Promise<string[]> {
+    const processFilter = processName ? `'${processName}'` : 'NULL';
+    const query = `
+      INCLUDE PERFETTO MODULE android.input;
```
ivan-chong: Here and elsewhere: This is unnecessary as it is imported on trace load

ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
```diff
@@ -61,19 +62,82 @@ export class AndroidInputEventSource {
...
+    return this.mapIteratorToRows(result);
+  }
+
+  async getRowsForApp(
+    processName?: string,
+    excludeSpeculative = false,
+  ): Promise<InputChainRow[]> {
+    const processFilter = processName ? `'${processName}'` : 'NULL';
+    const speculativeFilter = excludeSpeculative ? '1' : '0';
+    const query = `
```
ivan-chong: Here and elsewhere in the other queries:

I think this makes sense for the "core" query - however, I think we should consider having additional logic which then loops over all extensions as per #6577 such that our logic will work when Pixel / frame extensions (and any future ones) land.

ivan-chong (changes requested):
Thanks for the PR! Given the recent introductions of extensions in #6577  I think this needs to be structured differently namely to make the logic more general as different lifecycle stages are present. 

Would it be possible to rebase on #6577 (or alternatively #6670 to have something concrete to test with) and then address the comments?

Also, can we split the refactoring changes which to a separate PR if they are not strictly necessary? Thanks!

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: Yes, it's necessary here because these actions are asynchronous (querying the trace processor) and/or triggered by commands. Since command execution and async callbacks run outside of Mithril's automatic redraw cycle, we need to call `m.redraw()` to ensure the UI updates immediately when the state changes (such as when drawing flow arrows or pinning tracks).

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: Yes, we are actually using the standard `Select` widget from `../../widgets/select` and `showModal` from `../../widgets/modal` here. If you had a different widget or design in mind for this type of selection dialog, please let me know!

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: Agreed. Since rebasing on the extension framework, we no longer perform manual SQL imports or setup queries in the plugin. The extension tables/views are now expected to be loaded by the core/stdlib. I have removed the redundant setup logic.

ui/src/plugins/com.android.AndroidInputLifecycle/android_input_event_source.ts
jingjiewang-twn: Done. I have updated the query generation in `android_input_event_source.ts` (specifically in `fetchRows` and `getRowsForApp`). It now dynamically loops over all registered active extensions (`this.activeExtensions`) to append their SQL joins and select their columns. This ensures that the Pixel extension (and any future extensions) are automatically integrated into the queries.

jingjiewang-twn:
Thanks for the feedback! I have rebased the PR on top of your extension changes (#6670) and restructured the code to adopt the new extension framework. 

Regarding the refactored code, I have kept only the necessary changes to support pinning and drawing flows. If there are still parts that you think should be split out, please let me know!

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
ivan-chong: I see, I meant in terms of minimising the use of `div` and `p`, and perhaps using the `pf-` style widgets (e.g. from `ui/src/widgets/modal.scss`) just to help maintain consistency in styling

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -186,6 +282,47 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
     return connections;
   }
 
+  private useRowState(
+    trace: Trace,
```
ivan-chong: It seems like these functions haven't changed - perhaps we can move them in a follow up PR to fix the diff?

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -186,6 +282,47 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
...
+    source: AndroidInputEventSource,
+  ): QueryResult<InputChainRow[]> {
+    const selection = trace.selection.selection;
+
+    if (selection.kind !== 'track_event') {
+      return {data: [], isPending: false, isFresh: true};
+    }
+
+    const track = trace.tracks.getTrack(selection.trackUri);
```
ivan-chong: Do we need this logic? I believe this was removed in the rebased changes - if it is necessary we would need to import `SLICE_TRACK_KIND`

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -102,62 +109,151 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
         trace.tabs.showTab('com.android.AndroidInputLifecycleTab');
       },
     });
-  }
 
-  // Fetch or reuse cached row data for the currently selected slice. Can call
```
ivan-chong, line 107: nit: Can we keep this comment?

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
ivan-chong: IIUC, this should be handled automatically by `CommandManager` when it calls `raf.scheduleFullRedraw()` after awaiting the result of the command's callback - can you check if this works without it?

ivan-chong (changes requested):
Thanks for the changes! looks great overall % some comments! (You will need to rebase again given my recent changes to add the resample latency offset)

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: You are correct. I've removed all manual `m.redraw()` calls, including the one in `toggleDrawAllArrows` and the one in the modal's `Pin Tracks` action.
I verified that the UI updates correctly without them. Since the `Pin Tracks` action queries the engine via `getRowsForApp`, the engine automatically schedules a redraw when the query resolves (via `engine.onResponseReceived`), ensuring the pinned tracks appear immediately.

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: I have refactored the modal content to use standard Perfetto form widgets (`Form` and `FormLabel` from `widgets/form.ts`). This structures the modal nicely without raw `div` or `p` tags, and establishes proper accessibility association via the `for` and `id` attributes.

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: Confirmed that `SLICE_TRACK_KIND` check was removed in the parent branch. I've removed it from here too.
I also reorganized the file to restore `useRowState` and `applyInitialSelection` to their original positions (including the JSDoc comment), which completely removes them from the diff.

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: Sorry, I accidentally removed it. It has been restored now in the latest commit.

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: I have reorganized the file to restore `useRowState` and `applyInitialSelection` to their original positions. This removes them from the diff and keeps it clean.

jingjiewang-twn:
Thanks for the review and the detailed feedback, Ivan!

The branch has been rebased onto `dev/ivankc/android-pixel-input-extension`. I've updated the SQL query in `android_input_event_source.ts` to expose `event_time` and match the `a_evt` alias, which resolved the join crash with the Pixel extension.

Regarding the refactoring: the helper `extractConnectionsFromRows` was introduced to share the line-drawing logic between the single-selection flow and the new "draw all flows" command. I've kept it here to avoid duplication.

ivan-chong:
LGTM, thanks!

ivan-chong (approved):

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
```diff
@@ -102,6 +110,97 @@ export default class AndroidInputLifecyclePlugin implements PerfettoPlugin {
...
+            ),
+          buttons: [
+            {
+              text: 'Pin Tracks',
+              primary: true,
+              action: async () => {
+                this.pinnedApp = selectedApp;
+                pinningManager.unpinTracks(this.pinnedTrackUris);
+
+                const rows = await source.getRowsForApp(selectedApp);
```
ivan-chong, line 183: Looking at this again, It would be nice if this sorted by stage automatically

ivan-chong (changes requested):

ui/src/plugins/com.android.AndroidInputLifecycle/index.ts
jingjiewang-twn: Done. Updated the track pinning logic to extract track URIs stage-by-stage according to their stage sequence numbers (`source.getTrackUrisSortedByStage(rows)`), ensuring pinned tracks are automatically sorted chronologically by input stage.

ivan-chong (approved):
Thanks! LGTM % updating the PR / commit message to reflect the reduced scope of changes

PR #6658, opened 2026-07-10
tp: add a Notices severity and surface ftrace setup errors on it

LalitMaganti:
Introduce a kNotice stat severity for conditions that are normal but
noteworthy: not a bug, not corruption, and not data loss, but still
worth showing to the user. It slots between kInfo and kDataLoss.

Reclassify to kNotice the stats that fire when packets are skipped
because incremental state is invalid (ftrace setup failures and the
thread-descriptor / track-event / interned-data skips caused by packet
loss at the start of a trace). These were previously either invisible
kInfo counters or, worse, kError counters that made trace_processor
raise on a normal recording condition.

Convert ftrace_setup_errors from a concatenated metadata blob into
per-error trace import logs, each carrying the message as an arg, and
drop the metadata::ftrace_setup_errors key. This matches how other
producer-reported errors are recorded and gives structured, queryable
rows instead of one string.

On the UI, add a Notices tab to the trace info page that renders
severity='notice' categories with their detailed import logs, extracting
the shared category+logs rendering out of the Import Errors tab into a
reusable component so both tabs share it. The topbar error badge counts
only error/data_loss stats so notices do not flag as errors, and the
logs grid sizes rows to their content so multi-line messages are not
clipped.


sashwinbalaji (approved):

stevegolton (commented):
LG from the UI side

PR #6664, opened 2026-07-10
tp: Add pixel.touch stdlib module for touch events

ivan-chong:
Introduces the pixel.touch module to parse Pixel driver-level touch events (IRQ top-half/bottom-half and Twoshay processing timestamps) from the trace.

Bug: 454761692


src/trace_processor/perfetto_sql/stdlib/pixel/touch.sql
```diff
@@ -0,0 +1,152 @@
...
+  s.dur AS th_dur,
+  s.track_id AS th_track_id,
+  c.value AS irq_idx
+FROM _interval_intersect!((irq_slices, irq_counters), ()) AS ii
+JOIN irq_slices AS s
+  ON s.id = ii.id_0
+JOIN irq_counters AS c
+  ON c.id = ii.id_1;
+
+-- Pixel-specific touch events, including top and bottom half IRQ, and Twoshay touch processing.
```
LalitMaganti, line 102: Can we have some diff tests for these tables?

ivan-chong: Good point, I've now added some diff tests

LalitMaganti (approved):

jingjiewang-twn:
Hi Ivan, thanks for the implementation! Could we consider renaming `input.sql` to `touch.sql`? 

Since the table defined inside is already named `pixel_touch_events` (and it handles Pixel-specific touch driver stages like Twoshay, Bottom Half, and Top Half IRQs), renaming the file to `touch.sql` would make it consistent with the table name and prevent potential confusion with generic core Android input events in the future. What do you think?

ivan-chong:
> Hi Ivan, thanks for the implementation! Could we consider renaming `input.sql` to `touch.sql`?
> 
> Since the table defined inside is already named `pixel_touch_events` (and it handles Pixel-specific touch driver stages like Twoshay, Bottom Half, and Top Half IRQs), renaming the file to `touch.sql` would make it consistent with the table name and prevent potential confusion with generic core Android input events in the future. What do you think?

Sure, makes sense I've renamed it to touch.sql! As per your comment on #6670, I've also exposed the resample latency offset here.

jingjiewang-twn:
LGTM! Verified the changes locally and everything looks good. Thanks!

Issue #6666, opened 2026-07-10
ui: Core plugin settings are not sorted to the top


ivan-chong:
In the plugin settings page, currently the settings appear ordered as all the "non-core" plugin settings begin with `dev.perfetto` hence sorting these alphabetically works. When adding settings for a plugin which begins with `com.android` for example, this appears above the Core section.

I was wondering whether this was WAI - is it supported for plugins to add their own settings to the settings page?

I've made a PR for this in #6667, please let me know if this makes sense, thanks!

stevegolton:
>  is it supported for plugins to add their own settings to the settings page?

Yes, absolutely, but core should come first.

Thanks for fixing - good catch.

PR #6667, opened 2026-07-10
ui: Ensure Core settings are at the top

ivan-chong:
It seems like currently Core is not ordered at the top, as the sorting logic currently checks for falsy values as opposed to the CORE_PLUGIN_ID after some refactoring. This PR adds back the comparison except to the name 'Core' through a new constant CORE_GROUP.

Bug: #6666

### DIFF
--- a/ui/src/core_plugins/dev.perfetto.SettingsPage/settings_page.ts
+++ b/ui/src/core_plugins/dev.perfetto.SettingsPage/settings_page.ts
@@ -38,6 +38,7 @@ import {GateDetector} from '../../base/mithril_utils';
 import {findRef} from '../../base/dom_utils';
 
 const SEARCH_BOX_REF = 'settings-search-box';
+const CORE_GROUP = 'Core';
 
 export interface SettingsPageAttrs {
   readonly subpage?: string;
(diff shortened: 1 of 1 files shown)


ui/src/core_plugins/dev.perfetto.SettingsPage/settings_page.ts
```diff
@@ -59,10 +59,10 @@ export class SettingsPage implements m.ClassComponent<SettingsPageAttrs> {
       : this.getAllSettingsGrouped(settingsManager);
     const groupedSettings = this.groupSettingsByPlugin(settings);
 
-    // Sort plugin IDs: CORE_PLUGIN_ID first, then alphabetically
+    // Sort plugin IDs: 'Core' first, then alphabetically
     const sortedPluginIds = Array.from(groupedSettings.keys()).sort((a, b) => {
-      if (!a) return -1;
-      if (!b) return 1;
+      if (a === 'Core') return -1;
```
stevegolton: Maybe make 'Core' a const so this doesn't drift.

stevegolton (commented):
Good catch, thanks!

stevegolton (approved):

ui/src/core_plugins/dev.perfetto.SettingsPage/settings_page.ts
ivan-chong: Good point, done!

stevegolton (approved):

PR #6668, opened 2026-07-10
perfetto: Stop failing native builds on stale UI test data

safayat-google:
If a developer runs native builds (without --ui), we shouldn't fail builds because UI screenshots or funcgraph traces are stale.

This adds a --skip-prefix flag to tools/test_data and uses it in tools/install-build-deps and tools/run_presubmit to skip checking those files when --ui is not specified or there are no UI changes.

Bug: 531632295


tools/install-build-deps
```diff
@@ -1143,10 +1143,18 @@ def Main():
...
       cur_python_interpreter, TEST_DATA_SCRIPT, 'status', '--quiet',
       '--ignore-new'
-  ])
+  ]
+  if not args.ui:
+    test_data_status_cmd += [
+        '--skip-prefix',
+        'ui-screenshots',
+        '--skip-prefix',
+        'ui-funcgraph.pftrace',
```
LalitMaganti: This is a very odd prefix to add :) AFAIK ui-funcgraph could also well be used for e.g. diff tests. If we're doing this, we should exclude test/data completely *but* we should cause diff tests to error out on startup if they're not up to date.

safayat-google: removed the `ui-funcgraph.pftrace` prefix. and excluded the test/data completely, finally added the check on diff tests if not updated.

PR #6669, opened 2026-07-10
ui: fix circular dependency build error printing

primiano:
The Rollup onwarn handler that turns circular-dependency warnings into
build errors read `warning.importer` and `warning.cycle`. Those fields
existed in older Rollup, but Rollup 4.x emits the CIRCULAR_DEPENDENCY
warning with `warning.ids` (the cycle path) and a pre-formatted
`warning.message` instead. As a result the error printed as:

    Circular dependency: undefined

The detection itself worked; only the message construction referenced
the now-undefined fields. Use `warning.message` (which already contains
the "a -> b -> a" chain) and read the cycle from `warning.ids`, keeping
`warning.cycle` as a fallback for older Rollup.

Bug: b/533388309


stevegolton (approved):
Nice,  thanks!

PR #6675, opened 2026-07-10
Add a tracker to write counters only when changed.

rzuklie:
SparseCounterTracker helps efficiently write counter events into a trace by omitting values not necessary to recreate the same track. Consider this:

```
A     B     C     D     E
^-----^-----^-----^-----^
1     2     2     2     3
```

It is possible to fully represent this sequence by ommitting the point `C`. While the value is unchanged at point `D`, if we don't emit the point `D`, then it would appear as though the value increase by 1 between `B` and `E`.

Due to the ordering requirements, this is designed to work during the tokenization phase and writes into the sorter to guarantee the events pushed to the track are pushed in order.


rzuklie:
This is a second attempt at https://github.com/google/perfetto/pull/6578, generalizing it and moving it to the tokenization phase.

Lalit, is this use of the sorter reasonable? Also would you rather this tracker be placed with the common trackers, or should each importer module instantiate one if they need it?

LalitMaganti (approved):

PR #6680, opened 2026-07-12
tp: add `report` subcommand with streaming packet protocol

LalitMaganti:
Introduce `trace_processor_shell report`, an opinionated zero-config
summary tool alongside `summarize`. Output is a stream of self-delimited
ReportPacket protos (report.proto), modeled on the trace format so it is
truncatable and incrementally parseable. The single schema is emitted in
four encodings via --format: text (human tables), jsonl (one packet per
line), binary (length-delimited), and sql (print the underlying queries
without running them).

Views implement a small ReportView interface (Sql + Emit) and are
registered in a noun table, so running, --format sql and the overview
fan-out apply to every view uniformly. This adds the framework and the
first view: slices aggregated by name.

### DIFF
--- a/protos/perfetto/trace_processor/BUILD.gn
+++ b/protos/perfetto/trace_processor/BUILD.gn
@@ -45,3 +45,12 @@ perfetto_proto_library("metrics_impl_@TYPE@") {
   proto_generators = [ "zero" ]
   sources = [ "metrics_impl.proto" ]
 }
+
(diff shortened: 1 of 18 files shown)


LalitMaganti:
@sashwinbalaji: I don't want you to do a thorough review on this, just go over it and from the high level check it makes sense. We can discuss in person.

PR #6682, opened 2026-07-12
Fix bug in Flow::ProcessScoped() with named scope

kroymann:
`Flow::ProcessScoped(flow_id, named_scope)` currently fails to xor the process track ID with the flow ID and instead is acting a pass-through wrapper on `Flow::Global(flow_id, named_scope)`.  Users of `Flow::ProcessScoped(flow_id, named_scope)` will currently get collisions if the same flow ID and named scope are used simultaneously across different processes.


LalitMaganti:
Nice find! Could you please add a test for this as well?

PR #6683, opened 2026-07-13
tp: show address @ mapping for unsymbolized callstack frames

LalitMaganti:
Previously an unsymbolized callstack frame was rendered as the placeholder
'[Unknown]' (or an empty name). Instead, resolve such frames to
'0x<rel_pc> @ <mapping basename>', e.g. '0x1a2b @ libart.so', so
unsymbolized native traces still convey the address and library.

Name resolution now lives in a single __intrinsic_frame_name C++ function
called from the callstacks.stack_profile forest, replacing the nested
coalesce/demangle/iif SQL. This flows to every flamegraph built on the
module (CPU profile, heap profile, perf, track event callstacks). Frames
that have source info but no name (e.g. anonymous JS functions) are kept
as-is rather than given a meaningless address. The intrinsic only attempts
demangling for names that can actually be mangled and hands owned buffers
to SQLite to avoid result copies.

The remaining paths are routed through the same intrinsic for consistency:
the pprof importer now stores an empty frame name (and empty dummy mapping
name) instead of '[unknown]', so its already-imported addresses and
mappings surface; and the OOME, Memscope native, and track-event callstack
UI paths call the intrinsic instead of their own name fallbacks.

### DIFF
--- a/src/trace_processor/importers/pprof/pprof_trace_reader.cc
+++ b/src/trace_processor/importers/pprof/pprof_trace_reader.cc
@@ -65,7 +65,6 @@ struct FunctionInfo {
 
 PprofTraceReader::PprofTraceReader(TraceProcessorContext* context)
     : context_(context),
-      unknown_string_id_(context->storage->InternString("[unknown]")),
       unknown_no_brackets_string_id_(context->storage->InternString("unknown")),
       count_string_id_(context->storage->InternString("count")),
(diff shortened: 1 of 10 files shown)


LalitMaganti:
We decided not to do this because of the concerns of noisy flamegraphs and the difficulty now in identifying unknown frames.

PR #6685, opened 2026-07-13
ui: Pin tracks for memory regressions investigations

Nublo:
ui: Refactor PinAndroidPerfMetrics handlers and add new memory handlers

Introduced `SimpleProcessMetricHandler` to reduce boilerplate for handlers
that match a process name in a metric URL and pin its tracks. This base
class uses a list of regexes for matching and prefix-matches track names
by default.

Implemented handlers for Heap Size, Bitmap, Dirty Memory,
GPU Memory, Leak (Activity/Binder), and Hardware Buffer memory metrics.


ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/simpleProcessMetricHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import {NUM} from '../../../trace_processor/query_result';
+import type {Trace} from '../../../public/trace';
+import {
+  expandProcessName,
+  type MetricHandler,
+  type ProcessMetricData,
+} from './metricUtils';
+
+export class SimpleProcessMetricHandler implements MetricHandler {
```
lukaville, line 23: nit: do we need 'Simple' prefix in the class? maybe it could just ProcessMetricHandler?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/simpleProcessMetricHandler.ts
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Base class for simple logic track pinning
+   * Use when you have a Regexp which can extract a process name from an url
+   * And pin tracks in the found process
+   *
+   * @param {RegExp[]} matchers List of matchers to check original url with
```
lukaville: I think this class should not know where the metric key comes from (the url), should we describe that 'matchers' is a list of matchers for metric keys?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler.ts
```diff
@@ -12,18 +12,9 @@
...
-export const STATE_TRACK_SCHEMAS: ReadonlyArray<StateTrackTypeSchema> = [
-  {
-    type: 'concurrent_tracing_sessions',
-    topLevelGroup: 'SYSTEM',
-    group: 'Concurrent tracing sessions',
-  },
-];
+export const pinBitmapMetricsInstance = new SimpleProcessMetricHandler(
+  [/^perfetto_android_bitmap_metric_max_val-(?<processName>.*)/],
+  ['Bitmap Memory', 'Bitmap Count'],
```
lukaville: nit: is the order respected here? should we move the count as the first track to match the metric name as this is the metric that is tracked

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinDirtyMemoryMetricsHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinDirtyMemoryMetricsInstance = new SimpleProcessMetricHandler(
+  [
+    /perfetto_android_mem-(?<processName>.*)-total_counters-anon_and_swap-max-mean/,
+  ],
+  ['mem.rss', 'mem.rss.anon'],
```
lukaville: should this be mem.rss.anon + mem.swap instead?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHardwareBufferMemoryMetricsHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinHardwareBufferMemoryMetricsInstance =
+  new SimpleProcessMetricHandler(
+    [
+      /perfetto_android_dmabuf_per_process_metric_max_val-(?<processName>.*)-p95/,
```
lukaville: it looks like we have dmabuf allocs track go/sysui-metric-dma-buf, should we pin that instead?
also mem.gralloc.allocations, and mem.gralloc.buffers

+maybe total dmabuf size

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHeapSizeMetricsHandler.ts
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinHeapSizeMetricsInstance = new SimpleProcessMetricHandler(
+  [
+    /perfetto_android_mem-(?<processName>.*)-total_counters-java_heap-max-mean/,
+    /^perfetto_java_heap_class_stats-instance_stats-name-(?<processName>.*?)-samples-type_count-type_name/,
+  ],
+  ['Heap size', 'Native heap profile', 'ART allocation profile'],
```
lukaville: let's not pin the 'Native heap profile' since this is a java heap size metric?
btw also does 'Heap size' pinning work? I think the track name is 'Heap size (kb)' or something like that, do we do wildcard checking?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinGPUMemoryMetricsHandler.ts
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinGPUMemoryMetricsInstance = new SimpleProcessMetricHandler(
+  [
+    /perfetto_android_gpu-(?<processName>.*)-mem_max-max/,
+    /perfetto_android_gpu-\/system\/bin\/(?<processName>surfaceflinger)-mem_avg-mean/,
+  ],
+  ['GPU Memory', 'GPU Completion'],
```
lukaville: I think there are multiple 'GPU compleition' tracks, will it pin all of them?

lukaville (changes requested):

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/simpleProcessMetricHandler.ts
Nublo: I added `Simple` due to it takes regexp for metric url extracting `process_name` and then pinning some tracks.

Probably not generic enough for just `ProcessMetricHandler`, will keep it as is if you don't mind

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinGPUMemoryMetricsHandler.ts
Nublo: Yes.

Matching for track names happening with `startWith`.

```
const isTargetTrack = this.trackMatchers.some((matcher) => {
    return name.startsWith(matcher);
});
```

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHeapSizeMetricsHandler.ts
Nublo: Will remove `Native heap profile`

> does 'Heap size' pinning work?

Yes, will work.
Matching for track names happening with `startWith`.
```
const isTargetTrack = this.trackMatchers.some((matcher) => {
    return name.startsWith(matcher);
});
```

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler.ts
Nublo: No guarantees about the order.
However I think it should match the order of the tracks which `currentWorkspace.flatTracks` returns.

```
// Find the tracks to pin.
const tracksToPin = ctx.currentWorkspace.flatTracks.filter((track) => {
```

I will swap the order, just in case if we will decide to guarantee the order based on `trackMatchers`

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHardwareBufferMemoryMetricsHandler.ts
Nublo: As per docs will replace with:
`['dmabuf allocs', 'mem.gralloc.allocations', 'mem.gralloc.buffers']`

Total dmabuf size would be irrelevant as regression is related to specific process.
We have another regression for that

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHeapSizeMetricsHandler.ts
Nublo: Btw, I used this pattern `name.startsWith` because I was thinking maybe we might have "Heap size (Mb)", and it seems we can't. Do you think it will be better to replace `Heap size` with `Heap size (KB)` everywhere?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinGPUMemoryMetricsHandler.ts
lukaville: I think just one of these tracks should be enough (as an indication where the gpu work is happenning), should we support regexp here too and e.g. limit to "^GPU Completion$"?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHeapSizeMetricsHandler.ts
lukaville: I would say it doesn't really matter, we should probably even change the name of the track to something like "Java heap size" and specify the units properly https://perfetto.dev/docs/instrumentation/track-events#counters

(so maybe we would need to change the name here anyway)

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler.ts
lukaville: would it be difficult to ensure track order? imho it could help a lot if we pin the most relevant tracks first, especially if there are many of them

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinGPUMemoryMetricsHandler.ts
Nublo: Sure, I will support regexp for those (also noticed that it should be lowercased for `^GPU completion$`

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler.ts
Nublo: Should be pretty simple. I will update

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler_unittest.ts
```diff
@@ -0,0 +1,67 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import type {ProcessMetricData} from './metricUtils';
+import {pinBitmapMetricsInstance} from './pinBitmapMetricsHandler';
+
+const validMetricsTest: {
```
lukaville: looks like all these tests just check that they correctly extract process name from the metric. not sure if it is very helpful to have them in separate files. what do you think about having just one file with tests for simpleProcessMetricHandler with a few real metric examples?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHardwareBufferMemoryMetricsHandler.ts
lukaville: nit: should we also be more lenient on the aggregations? e.g have .* instead of p95, in case we decide to change the aggregations in the future

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinGPUMemoryMetricsHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinGPUMemoryMetricsInstance = new SimpleProcessMetricHandler(
+  [
+    /perfetto_android_gpu-(?<processName>.*)-mem_max-max/,
+    /perfetto_android_gpu-\/system\/bin\/(?<processName>surfaceflinger)-mem_avg-mean/,
```
lukaville: nit: would it make sense to just have mapping /system/bin/surfaceflinger -> surfaceflinger inside SimpleProcessMetricHandler, so we don't need to create dulpicated regexp for each metric

this will also make all other exsiting metrics work in case we start reporting them for SF

also did you check if it works with just 'surfaceflinger' in the process name? I would expect it might be stored as '/system/bin/surfaceflinger' in the process table

Nublo: I think we can keep single regexp to:

```
/perfetto_android_gpu-(?<processName>.*)-mem_.*/
```

> also did you check if it works with just 'surfaceflinger' in the process name? I would expect it might be stored as '/system/bin/surfaceflinger' in the process table

It was working, because we use a function inside of `metricUtils`

```
export function expandProcessName(metricProcessName: string): string {
  if (metricProcessName.includes('systemui')) {
    return 'com.android.systemui';
  } else if (metricProcessName.includes('launcher')) {
    return 'com.google.android.apps.nexuslauncher';
  } else if (metricProcessName.includes('surfaceflinger')) {
    return '/system/bin/surfaceflinger';
  } else {
    return metricProcessName;
  }
}
```

However with this change should be even more clean

lukaville: Sounds good, thanks

lukaville (approved):

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler.ts
```diff
@@ -12,18 +12,9 @@
...
-export const STATE_TRACK_SCHEMAS: ReadonlyArray<StateTrackTypeSchema> = [
-  {
-    type: 'concurrent_tracing_sessions',
-    topLevelGroup: 'SYSTEM',
-    group: 'Concurrent tracing sessions',
-  },
-];
+export const pinBitmapMetricsInstance = new SimpleProcessMetricHandler(
+  [/^perfetto_android_bitmap_metric_max_val-(?<processName>.*)/],
+  ['Bitmap Count', 'Bitmap Memory'],
```
jernej-google: Pin also `dmabuf allocs` track, `mem.gralloc.allocations` and `mem.gralloc.buffers` tracks as well since they're directly related to bitmap memory.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler.ts
```diff
@@ -12,18 +12,9 @@
...
-
-export const STATE_TRACK_SCHEMAS: ReadonlyArray<StateTrackTypeSchema> = [
-  {
-    type: 'concurrent_tracing_sessions',
-    topLevelGroup: 'SYSTEM',
-    group: 'Concurrent tracing sessions',
-  },
-];
+export const pinBitmapMetricsInstance = new SimpleProcessMetricHandler(
+  [/^perfetto_android_bitmap_metric_max_val-(?<processName>.*)/],
```
jernej-google: The `max` aggregation should be templated as well, since it can be `min`, `avg` as well.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinDirtyMemoryMetricsHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinDirtyMemoryMetricsInstance = new SimpleProcessMetricHandler(
+  [
+    /perfetto_android_mem-(?<processName>.*)-total_counters-anon_and_swap-max-mean/,
```
jernej-google: This should be robust against different aggregations (`max` and `mean` can change).

Also, you want to pin `Heap size` track as well.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinActivityOrBinderLeaksMetricsHandler.ts
```diff
@@ -12,18 +12,16 @@
...
+export const pinActivityOrBinderLeaksMetricsInstance =
+  new SimpleProcessMetricHandler(
+    [
+      /(?<processName>.*)_Activities-last-first-diff/,
+      /(?<processName>.*)_View-last-first-diff/,
+      /(?<processName>.*)_ViewRootImpl-last-first-diff/,
+      /(?<processName>.*)_Local Binders-last-first-diff/,
+      /(?<processName>.*)_Proxy Binders-last-first-diff/,
+    ],
+    ['Heap size'],
```
jernej-google, line 26: I believe the heap size track is actually `[Heap Size (KB)`](https://g3doc.corp.google.com/company/teams/android-sysui/general_guides/performance/metrics/java_heap.md?cl=head). Will that make a difference?

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinGPUMemoryMetricsHandler.ts
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinGPUMemoryMetricsInstance = new SimpleProcessMetricHandler(
+  [/perfetto_android_gpu-(?<processName>.*)-mem_.*/],
```
jernej-google, line 18: Also pin `dmabuf allocs` and the `mem.gralloc.allocations` and `mem.gralloc.buffers` tracks which are relevant as well.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHardwareBufferMemoryMetricsHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinHardwareBufferMemoryMetricsInstance =
+  new SimpleProcessMetricHandler(
+    [
+      /perfetto_android_dmabuf_per_process_metric_max_val-(?<processName>.*)-.*/,
```
jernej-google: Again, make sure different aggregations are handled too.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHeapSizeMetricsHandler.ts
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+import {SimpleProcessMetricHandler} from './simpleProcessMetricHandler';
+
+export const pinHeapSizeMetricsInstance = new SimpleProcessMetricHandler(
+  [
+    /perfetto_android_mem-(?<processName>.*)-total_counters-java_heap-max-mean/,
+    /^perfetto_java_heap_class_stats-instance_stats-name-(?<processName>.*?)-samples-type_count-type_name/,
+  ],
+  ['Heap size', 'ART allocation profile'],
```
jernej-google: Never seen "ART Allocation profile", what is it?

Also pin anon rss.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinActivityOrBinderLeaksMetricsHandler.ts
Nublo: I am using prefix matchers, so this will work.

Originally I thought that we might have `Heap Size (KB)` / `Heap Size (MB)` (which we can't actually have).
But if in future something changes with the postfix ex: `Heap Size KB` => current pinning implementation will still work, so decided to keep prefix matching.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHeapSizeMetricsHandler.ts
Nublo: [ART Allocation profile](https://perfetto.dev/docs/visualization/heap-dump-explorer#heap-dumps-vs-heap-profiles)

Added "mem.anon.rss"

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinBitmapMetricsHandler.ts
Nublo: `dmabuf allocs` is a track created by plugin: `com.android.AndroidDmabuf`
Added proper tags to `com.android.AndroidDmabuf` implementation otherwise we can't pin it

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinActivityOrBinderLeaksMetricsHandler.ts
jernej-google: Understood.

ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/pinHeapSizeMetricsHandler.ts
jernej-google: Got it.

jernej-google (approved):

PR #6687, opened 2026-07-13
tp: fix dropped traces when merging same-remote-machine files

dreveman:
Merging independent traces that share a machine id dedups them onto one
machine, and each non-primary file isolates its clocks onto its own tag.
On a remote (non-trace-time) machine that isolation kept the file's
REALTIME off the cross-machine rendezvous, so every event failed clock
conversion and the whole trace was silently dropped.

Bridge a non-primary remote file's REALTIME to the machine-canonical
REALTIME so it joins the rendezvous, and flush remote-machine readers'
clock-deferred packets in OnPushDataToSorter.

### DIFF
--- a/src/trace_processor/importers/common/clock_tracker.cc
+++ b/src/trace_processor/importers/common/clock_tracker.cc
@@ -88,8 +88,31 @@ base::StatusOr<uint32_t> ClockTracker::AddSnapshot(
     }
     current_file_tag_ = own_file_id_;
   }
-  for (auto& ct : clock_timestamps)
+  // REALTIME is a single universal wall clock and the cross-machine rendezvous
+  // domain used to place files that share no other clock (see
+  // docs/concepts/merging-traces.md). A non-primary file on a *remote*
+  // (non-trace-time) machine reaches trace time only through that rendezvous,
+  // so relate each REALTIME clock it actually carries to the machine-canonical
+  // REALTIME at zero offset (a twin added to this snapshot). The trace-time
+  // machine is excluded: its events reach trace time directly through BOOTTIME.
+  const bool bridge_realtime =
+      !is_primary_ &&
+      machine_id_ != context_->trace_time_state->clock_id.machine_id;
+  std::vector<ClockTimestamp> canonical_realtime;
+  for (auto& ct : clock_timestamps) {
+    const uint32_t clock = ct.clock.id.clock_id;
     ct.clock.id = ClockId::Qualify(ct.clock.id, machine_id_, current_file_tag_);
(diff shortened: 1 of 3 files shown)


LalitMaganti:
I'm not sure this is actually the correct fix: shouldn't we instead add an edge from the per-file REALTIME clock and the machine REALTIME clock resolving this problem?

dreveman:
> I'm not sure this is actually the correct fix: shouldn't we instead add an edge from the per-file REALTIME clock and the machine REALTIME clock resolving this problem?

Good call — done, that's cleaner. I switched to adding a zero-offset REALTIME(machine,file) <-> REALTIME(machine,canonical) edge when a non-primary file isolates, instead of special-casing REALTIME in the qualification path (which kept the hot ToTraceTime/Convert path untouched). One subtlety: it has to be gated to remote (non-trace-time) machines. If added unconditionally it regresses multi_trace_single_machine_clock — for a non-primary file on the trace-time machine the BFS then has two equal-length paths for a REALTIME event (its own BOOTTIME snapshot vs. the primary's via the new edge) and picks the wall-clock one, shifting timestamps ~70ns. On the trace-time machine those events reach trace time directly through BOOTTIME, so the edge is only needed on remote machines.

src/trace_processor/importers/common/clock_tracker.cc
```diff
@@ -87,6 +87,27 @@ base::StatusOr<uint32_t> ClockTracker::AddSnapshot(
...
+    // files and, via NTP, across machines, where it is the cross-machine
+    // rendezvous domain (see docs/concepts/merging-traces.md). On a *remote*
+    // (non-trace-time) machine this file's events can only reach trace time
+    // through that rendezvous, so tie this file's now-isolated REALTIME to the
+    // machine-canonical REALTIME at zero offset; its own snapshots stay
+    // authoritative for its own conversions. On the trace-time machine we do
+    // NOT add this edge: those events reach trace time directly through
+    // BOOTTIME, and the extra REALTIME path would route them through the
+    // jittery wall clock.
+    if (machine_id_ != context_->trace_time_state->clock_id.machine_id) {
```
LalitMaganti: Instead of directly adding snapshots, why not just add to `clock_timestamps` so it's optimized below? You would have to be careful with double usage of Qualify though.

Also I would be careful of adding edges where the clocks don't exist - is it possible to verify that the clock *acutally* exists in the trace before adding this edge. I think this could cause bad behavior in some edge cases otherwise.

dreveman: Done. ptal.

LalitMaganti (approved):

PR #6692, opened 2026-07-13
tp: add experimental_slice_layout benchmark

LalitMaganti:


sashwinbalaji (approved):
can be made part of #6686 also

PR #6695, opened 2026-07-13
sdk: enable SHLIB export in amalgamated output

hfink-sc:
gen_amalgamated is unconditionally prepending

#define PERFETTO_SDK_DISABLE_SHLIB_EXPORT

...assuming that the C SDK will be used only in a static linking
scenario. However, for the use case of building a shared library from
the amalgamated C SDK, this prevents to control symbol visibility via
PERFETTO_SDK_EXPORT.

Change this to only define PERFETTO_SDK_DISABLE_SHLIB_EXPORT when
PERFETTO_SHLIB_SDK_IMPLEMENTATION is not set, so the default static
linkage keeps PERFETTO_SDK_EXPORT empty while shared library builds
get proper export annotations.

Also attach a drive-by documentation fix.

### DIFF
--- a/include/perfetto/public/abi/export.h
+++ b/include/perfetto/public/abi/export.h
@@ -31,7 +31,7 @@
 // line):
 // * PERFETTO_SDK_DISABLE_SHLIB_EXPORT: If this is defined, no export
 //   annotations are added. This might be useful when static linking.
-// * PERFETTO_SDK_SHLIB_IMPLEMENTATION: This must be defined when compiling the
+// * PERFETTO_SHLIB_SDK_IMPLEMENTATION: This must be defined when compiling the
 //   shared library itself (in order to export the symbols), but must be
 //   undefined when compiling objects that use the shared library (in order to
 //   import the symbols).
--- a/tools/gen_amalgamated
+++ b/tools/gen_amalgamated
@@ -51,8 +51,12 @@ c_sdk_header_preamble = [
     '#define _GNU_SOURCE',
     '#endif',
     '',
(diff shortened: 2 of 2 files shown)


sashwinbalaji:
Thanks for the PR! 

Unfortunately guarding on PERFETTO_SHLIB_SDK_IMPLEMENTATION won't work, gen_amalgamated copies the GN defines of [shared_lib](https://github.com/google/perfetto/blob/main/src/shared_lib/BUILD.gn#L46) into the generated source, so the amalgamated perfetto_c.cc already defines that macro at the top:

```
#if !defined(PERFETTO_SHLIB_SDK_IMPLEMENTATION)
#define PERFETTO_SHLIB_SDK_IMPLEMENTATION
#endif
#include "perfetto_c.h"
```

The guard is therefore always false inside perfetto_c.cc and the export annotations are always on. That happens to do the right thing for your Linux .so build, but:

* every normal static build of the amalgamated SDK also gets the export annotations: visibility("default") on Linux which defeats hidden visbility and re-exports the whole ABI from any binary that embeds the SDK 
* on Windows, consumers can never get __declspec(dllimport): defining PERFETTO_SHLIB_SDK_IMPLEMENTATION picks dllexport, the wrong side. 


I've sent https://github.com/google/perfetto/pull/6706 which fixes this with an explicit flag instead: 
- no macros = static linking (unchanged default)
- `-DPERFETTO_SDK_SHLIB` = shared library.

For your yocto build that just means adding `-DPERFETTO_SDK_SHLIB` when compiling the amalgamated perfetto_c.cc.

PR #6701, opened 2026-07-14
ui: pin relevant tracks for dma memory regression

Nublo:
Pin global `mem.dma_heap` and `mem.dma_buffer` when facing regression

Whenever we detect that url contained a global dma memory regression
We pin global (`mem.dma_heap` and `mem.dma_buffer`) tracks
to simplify investigation.

### DIFF
--- a/ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/metricUtils.ts
+++ b/ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/metricUtils.ts
@@ -88,7 +88,7 @@ export interface ProcessMetricData {
   process: string;
 }
 
-export interface DMAHeapMetricData {}
+export interface GlobalDmaHeapMetricData {}
 
 // Common MetricData for all handler. If new needed then add here.
 export type MetricData =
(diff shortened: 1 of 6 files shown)


ui/src/plugins/com.android.PinAndroidPerfMetrics/handlers/metricUtils.ts
```diff
@@ -88,14 +88,17 @@ export interface ProcessMetricData {
   process: string;
 }
 
+export interface DMAHeapMetricData {}
```
Nublo: We do not extract any data from url, but we have to provide a dummy object by the contract

lukaville: nit: maybe DmaHeapMetricData? for consistency (e.g. Cuj)

lukaville: Ack (note: it doesn't look like url is the only source for metric key, it could be also passed e.g. by manually invoking the pin tracks action, so here we extract data just from a metric key, not a url)

lukaville: also maybe it could be GlobalDmaHeapMetricData or DeviceDmaHeapMetricData to disambiguate with the process-specific metrics

lukaville (changes requested):

lukaville (approved):

jernej-google (approved):

PR #6706, opened 2026-07-14
perfetto: rework C SDK shared library export macros

sashwinbalaji:
Replace the negative-logic PERFETTO_SDK_DISABLE_SHLIB_EXPORT with a positive-logic gate, in the same shape as PERFETTO_EXPORT_COMPONENT:

* PERFETTO_SDK_SHLIB: defined when the SDK is built as a shared library, in both the library and its consumers.
* PERFETTO_SDK_SHLIB_IMPLEMENTATION (renamed from PERFETTO_SHLIB_SDK_IMPLEMENTATION): additionally defined when compiling the library itself.

This is a known follow-up from aosp/2132115, which introduced these macros: the negative logic was chosen so that copied headers work against the shared library out of the box, with the note that once the headers joined the amalgamated build a dedicated build flag would be needed. PERFETTO_SDK_SHLIB is that flag.

Benefits:
* Static linking, the default, needs no macros at all.
* The amalgamated C SDK can be built as a shared library by defining PERFETTO_SDK_SHLIB. Previously the generated header unconditionally disabled export annotations with no way to override: a Windows DLL build got no dllexport and its consumers no dllimport, so exported data symbols (e.g. perfetto_te_any_categories) failed to link.
* Checked-in headers and the amalgamated SDK now behave identically.
* GN owns the linkage macros: they are defined only in configurations that build libperfetto_c as a shared library, so gen_amalgamated harvests the build description verbatim; the generated .cc declares itself as the implementation.

Breaking changes:
* Windows consumers of a prebuilt libperfetto_c.dll relying on bare headers defaulting to dllimport must define PERFETTO_SDK_SHLIB.
* Out-of-tree builds defining PERFETTO_SHLIB_SDK_IMPLEMENTATION must switch to the new name.
* Defining PERFETTO_SDK_DISABLE_SHLIB_EXPORT is now a no-op with the same effect (annotations disabled by default).


src/shared_lib/BUILD.gn
```diff
@@ -14,16 +14,37 @@
...
+config("shlib_config") {
+  defines = [ "PERFETTO_SDK_SHLIB" ]
+}
+
+# Makes the public ABI headers annotate symbols for export from the shared
+# library, in the targets that implement it. Empty in the amalgamated SDK,
+# where the consumer chooses between static and shared linking when compiling
+# the generated sources. See include/perfetto/public/abi/export.h.
+config("shlib_impl_config") {
+  if (!perfetto_amalgamated_sdk) {
```
LalitMaganti, line 29: nit: can be removed.

LalitMaganti (approved):

src/shared_lib/BUILD.gn
sashwinbalaji: can't due to current design, filed http://b/534825262 to rework flags later.

PR #6707, opened 2026-07-14
sdk: Add PerfettoDsGetTimestamp to data source ABI

dreveman:
Track event C API has PerfettoTeGetTimestamp() which hides the
preferred trace clock selection. The data source C API lacked an
equivalent, so users had to duplicate the selection logic.

Add PerfettoDsGetTimestamp() and PerfettoDsGetDefaultClockId() to
data_source_abi.h. Make PerfettoTeTimestampType values reuse
PerfettoDsClockId and have TeGetTimestamp() reuse DsGetTimestamp().
Deprecate PERFETTO_I_CLOCK_INCREMENTAL_UNDERNEATH and update internal
C++ code to use the new function. Avoid platform ifdefs in public
headers by determining the default clock at runtime.

Expose DataSourceTimestamp::now() in the Rust SDK and update the
gpu_counters example. Bump perfetto-sdk-sys 1.2.0->1.3.0 and
perfetto-sdk 1.0.3->1.1.0.


include/perfetto/public/abi/track_event_abi.h
sashwinbalaji, line 121: this will set a diff clock then GetTraceClockId for OSes not windows,apple, linux and android.

src/shared_lib/data_source.cc
```diff
@@ -609,3 +610,13 @@ void PerfettoDsTracerImplFlush(struct PerfettoDsTracerImpl* tracer,
   }
   tls_inst->trace_writer->Flush(fn);
 }
+
+struct PerfettoDsTimestamp PerfettoDsGetTimestamp() {
+  struct PerfettoDsTimestamp ret;
+  ret.clock_id = perfetto::base::GetTraceClockId();
+  ret.value = static_cast<uint64_t>((ret.clock_id == 6
```
sashwinbalaji: let's use the PERFETTO_DS_CLOCK_BOOTTIME everywhere instead of 6

src/shared_lib/data_source.cc
```diff
@@ -609,3 +610,13 @@ void PerfettoDsTracerImplFlush(struct PerfettoDsTracerImpl* tracer,
   }
   tls_inst->trace_writer->Flush(fn);
 }
+
+struct PerfettoDsTimestamp PerfettoDsGetTimestamp() {
```
sashwinbalaji, line 623: pls also add tests and update changelog

include/perfetto/base/time.h
```diff
@@ -313,6 +313,17 @@ inline TimeNanos GetProcessCPUTimeNs() {
 }
 #endif
 
+// Returns clock id for the preferred trace clock. Values match
+// perfetto.common.BuiltinClock (3 = MONOTONIC, 6 = BOOTTIME).
+constexpr uint32_t GetTraceClockId() {
+#if !PERFETTO_BUILDFLAG(PERFETTO_OS_APPLE) && \
+    !PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  return 6;  // BOOTTIME
```
sashwinbalaji: this seems bit fragile and there are 3 copies of this now.  Since base can't see the protos, add some static_asserts in the other files to be safer

nit: rename the func to be GetDefaultTraceClockId as the clock value can be set at runtime also via args like use_monotonic_clock

include/perfetto/public/abi/track_event_abi.h
dreveman: I deprecated this in favor of using the new PerfettoDsGetDefaultClockId function I added. I kept the enum and updated it to match PerfettoDsGetDefaultClockId. I assume we can't remove it after having exposed it in the public API or was this always meant to only be for internal usage?

src/shared_lib/data_source.cc
dreveman: Done

src/shared_lib/data_source.cc
dreveman: Done

include/perfetto/base/time.h
dreveman: I moved this to src/shared_lib/data_source.cc so we can use `PERFETTO_DS_CLOCK_BOOTTIME/PERFETTO_DS_CLOCK_BOOTTIME/MONOTONIC` instead

include/perfetto/public/abi/track_event_abi.h
sashwinbalaji: haven't reviewed the change yet but as per https://r.android.com/2338023 anything with _I_ is internal so removing it should be fine

include/perfetto/public/abi/track_event_abi.h
```diff
@@ -104,14 +104,15 @@ PERFETTO_SDK_EXPORT void PerfettoTeCategoryImplDestroy(
...
+  PERFETTO_TE_TIMESTAMP_TYPE_MONOTONIC = PERFETTO_DS_CLOCK_MONOTONIC,
+  PERFETTO_TE_TIMESTAMP_TYPE_BOOT = PERFETTO_DS_CLOCK_BOOTTIME,
   PERFETTO_TE_TIMESTAMP_TYPE_INCREMENTAL = 64,
   PERFETTO_TE_TIMESTAMP_TYPE_ABSOLUTE = 65,
 };
 
+// Deprecated: Use PerfettoDsGetDefaultClockId() instead.
 enum {
-#ifdef __linux__
+#if !defined(__APPLE__) && !defined(_WIN32) && !defined(_WIN64)
```
sashwinbalaji: nit: _win64 not needed

```
_WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, x64, or ARM64EC. Otherwise, undefined.

_WIN64 Defined as 1 when the compilation target is 64-bit ARM, x64, or ARM64EC. Otherwise, undefined.
```

src/shared_lib/track_event/serialization.cc
```diff
@@ -101,7 +101,7 @@ void ResetIncrementalStateIfRequired(
   incr_state->was_cleared = false;
 
   auto sequence_timestamp = timestamp;
-  if (timestamp.clock_id != PERFETTO_I_CLOCK_INCREMENTAL_UNDERNEATH &&
+  if (timestamp.clock_id != PerfettoDsGetDefaultClockId() &&
```
sashwinbalaji: minor(optional) nit: store in local variable return of PerfettoDsGetDefaultClockId (ik it's a cheap call but just for clarity) instead of invoking it 3 times in this func.

CHANGELOG
```diff
@@ -43,6 +43,9 @@ Unreleased:
       PERFETTO_SDK_DISABLE_SHLIB_EXPORT is removed and
       PERFETTO_SHLIB_SDK_IMPLEMENTATION is renamed to
       PERFETTO_SDK_SHLIB_IMPLEMENTATION.
+    * Added PerfettoDsGetTimestamp() to the C data source ABI to get the
```
sashwinbalaji: nit: I interpreted " without handling OS differences" as the function doesn't handle it can we reword it a bit pls. also should cover PerfettoDsGetDefaultClockId as it's part of ABI

somethign like:
These return the current timestamp in the preferred trace clock, so callers n o longer need to pick the OS specific clock themselves.

sashwinbalaji (approved):

CHANGELOG
dreveman: Done

src/shared_lib/track_event/serialization.cc
dreveman: Done

include/perfetto/public/abi/track_event_abi.h
dreveman: Done

PR #6710, opened 2026-07-14
ui: Simplify DataGrid schema models

stevegolton:
Currently the schema is a record of tables keyed by name, like this:

```js
{
  slice: {
    id: {},
    ts: {},
    dur: {},
    track: {ref: 'track', ...}, // Ref to the foreign table
  },
  track: {
    id: {},
    name: {}
  },
  parent: {ref: 'slice', ...}, // Recursive reference
}
```

This works but is messy:
- Can reference non-existent tables - must check for undefined everywhere.
- The vast majority of schemas in practice are flat (single table) which means we have to nest every single schema inside a dummy table name when its totally redundant.
- Field paths don't specify a 'root' table name so we must pass the 'root' (default) table name around as a string everywhere.

## The fix

This patch replaces this nominal structure with a structured definition which solves all of these problems:

E.g.

```js
const sliceSchema = {
  id: {},
  ts: {},
  dur: {},
  table: {
    schema: { // <-- Nested table schema
       id: {},
       name: {}
    },
  },
  parent: {
    get schema() { // <-- recursive references can exist - you just have to use a getter.
      return sliceSchema;
    }
  },
}
```

Notes:
- The recursive via getter trick is borrowed from [zod](https://zod.dev/api?id=recursive-objects#recursive-objects).
- The sqlSchema (the separate schema passed to the SQLDataSource has been restructured in a similar way.
- This is a big change due to the number of DataGrid call-sites, but it's mainly mechanical. Should be a no-op.

Key changes:
- SchemaRef.ref (string) → SchemaRef.schema (ColumnSchema), with get schema() for self-referential tables
- SQLTableSchema.table → SQLTableSchema.tableOrSubquery, always wrapped in parens in generated SQL
- DataGridAttrs.schema is now ColumnSchema (was SchemaRegistry), rootSchema removed
- SQLDataSource config extends SQLTableSchema directly (was sqlSchema + rootSchemaName)
- Add getQuery() to SQLDataSource and sub-engines for SQL inspection

### DIFF
--- a/ui/src/bigtrace/pages/results_grid.ts
+++ b/ui/src/bigtrace/pages/results_grid.ts
@@ -19,10 +19,7 @@ import {linkify} from '../../widgets/anchor';
 import {Spinner} from '../../widgets/spinner';
 import {DataGrid} from '../../components/widgets/datagrid/datagrid';
 import type {DataSource} from '../../components/widgets/datagrid/data_source';
-import type {
-  ColumnSchema,
-  SchemaRegistry,
-} from '../../components/widgets/datagrid/datagrid_schema';
+import type {ColumnSchema} from '../../components/widgets/datagrid/datagrid_schema';
 import type {
   Column,
   SortDirection,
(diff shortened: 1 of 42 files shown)


stevegolton:
Note: The bigtrace CI failure is unrelated - should be fixed by https://github.com/google/perfetto/pull/6720

LalitMaganti (approved):

PR #6711, opened 2026-07-14
Add algorithm header for std::transform

Sharjeel-Khan:
Libc++ have started removing more transitive includes so we need the algorithm header in this file for std::transform.


Sharjeel-Khan:
We need this fix landed so Android can pull it back for our new compiler update.

LalitMaganti (approved):

sashwinbalaji (approved):

PR #6717, opened 2026-07-15
tp,ui,docs: Surface the display-video user-build sysprop in docs, record page and trace doctor

zezeozue:
Screen recording (`android.display.video`) is now capturable on `user`
  (production) builds, not just `userdebug`. On user builds it must be unlocked
  first, over ADB, until the next reboot:                              
 
      adb shell setprop debug.tracing_video_allowed true                                                                                      
                                                                       
  The property is cleared on reboot, so it has to be re-set after each restart                                                                
  before display video can be captured. Once set, capture works through any of the                                                            
  existing methods (the on-device toggle, the record page, or a raw config).                                                                  
                                                                       
  This makes that requirement discoverable in three places:                                                                                   
                                                                                                                                              
  **Docs** (`docs/data-sources/video-frames.md`)                                                                                              
  - Drop the "userdebug devices only" statement.                                                                                              
  - Add a "Prerequisite on `user` builds" section documenting the property, the                                                               
    command, and its reset-on-reboot behaviour; point the intro/TOC at it.
                                                                                                                                              
  **Record page** (`Display video frames` probe)                       
  - The probe description now spells out the user-build requirement and shows the
    `setprop` command as a code snippet, and links to the data-source doc.                                                                    
                                                                       
  **Trace doctor** (new `display_video_not_enabled` diagnostic)        
  - Fires when a trace requested `android.display.video` but captured no frames,
    on a `user` build, with no producer error stats — the usual cause being the                                                               
    property was not set. 
  - The video-frames table is owned by the `video_frame_importer` plugin and isn't                                                            
    reachable from the common diagnostics framework, so the plugin now bumps a new                                                            
    `android_video_frames_emitted` stat per inserted frame; the rule reads                                                                    
    emptiness through that stat, alongside the existing `android_video_*` error                                                               
    stats.

### DIFF
--- a/src/trace_processor/plugins/video_frame_importer/video_frame_module.cc
+++ b/src/trace_processor/plugins/video_frame_importer/video_frame_module.cc
@@ -175,10 +175,8 @@ void VideoFrameModule::ParseVideoFrame(protozero::ConstBytes bytes,
   info.emitted_bytes += static_cast<int64_t>(payload.size);
 
   auto id = table_->Insert(row).id.value;
-  // Count emitted frames so the trace-config diagnostics can tell "no frames
-  // captured" (e.g. debug.tracing_video_allowed not set on a user build) apart
-  // from a producer error, without reaching into this plugin-owned table.
-  context_->stats_tracker->IncrementStats(stats::android_video_frames_emitted);
+  // Signal frame output to trace doctor (see TraceStorage).
+  context_->storage->set_has_android_video_frames();
   // au_data is parallel to the table, indexed by row id.
   PERFETTO_DCHECK(id == au_data_->size());
   const TraceBlobView& packet = data.packet;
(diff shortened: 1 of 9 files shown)


zezeozue:
image
image
image

See images above

src/trace_processor/plugins/video_frame_importer/video_frame_module.cc
```diff
@@ -175,6 +175,10 @@ void VideoFrameModule::ParseVideoFrame(protozero::ConstBytes bytes,
   info.emitted_bytes += static_cast<int64_t>(payload.size);
 
   auto id = table_->Insert(row).id.value;
+  // Count emitted frames so the trace-config diagnostics can tell "no frames
+  // captured" (e.g. debug.tracing_video_allowed not set on a user build) apart
+  // from a producer error, without reaching into this plugin-owned table.
+  context_->stats_tracker->IncrementStats(stats::android_video_frames_emitted);
```
LalitMaganti: This is a bit of a hack: the thing is that trace doctor should *itself* be pluginized.

For now, I would suggest you move the tables you need to the core and leave a todo for me to fix.

LalitMaganti (approved):

Issue #6559, opened 2026-07-07
Include Extension Server modules in Data Explorer table search


jossiwolf:
Steps to repro:

1. Open https://ftl-prod-dot-androidx-dev-prod.appspot.com/tests/benchmarkArtifact/redirect?invocationId=I82300010568390486&resourceId=15588459%2FMacroMetricTests%2Finv_4521474933794414191%2FPokedexScrollBenchmark_scrollHomeCompose_compilation_Full_eSTS_false_eSET_false__iter000_2026-06-08-14-38-15_13035154417376973336_7054145665225701868.perfetto-trace.gz&redirect=true&signature=j6y%2FJ58wkoF2o667Nmo7%2F1G8%2F6nC3h0ERw2toWB12YQ%3D
2. Add https://github.com/jossiwolf/androidx-perfetto-extensions/ as extension server (Settings -> Extension Servers -> Add Server)
3. Open 'Data Explorer'
4. Click the 'Add' button -> 'Table'
5. Search for 'AndroidX'

Expected: Metrics form the extension appear
Actual: Only stdlib tables appear

This example SQL query works, so the extension is loaded correctly:

```SQL
INCLUDE PERFETTO MODULE androidx.benchmark.art;

SELECT * from androidx_art_metrics;
```

LalitMaganti:
Can you please check on autopush? There's been a bunch of changes here and I think this might now be working there.

jossiwolf:
Just checked, I don't see the tables on autopush.

LalitMaganti:
Are you sure? On autopush, I can see the Google3 tables for example from the Google internal extension server:

<img width="3058" height="1850" alt="Image" src="https://github.com/user-attachments/assets/233c12da-c834-45ff-8b5e-51be30fefdf8" />

jossiwolf:
Yeah I see those. I think there's a bug in the search actually:

<img width="3024" height="1724" alt="Image" src="https://github.com/user-attachments/assets/f5676468-57e3-4ca0-895f-50c181752648" />

I see a metric from the extension server when searching for 'jank':

<img width="3024" height="1724" alt="Image" src="https://github.com/user-attachments/assets/56bfb8ab-ca3c-48c2-908c-4645a9b3a224" />

LalitMaganti:
Got it. Thanks for filing, I'll take a look.

LalitMaganti:
https://github.com/google/perfetto/pull/6719 will fix this.

[...]

PR #6719, opened 2026-07-15
ui: rank table search by relevance then importance

LalitMaganti:
The Data Explorer "Choose a table" search ordered table-name matches
purely by table importance, discarding the fuzzy finder's relevance
order. A fuzzy but high-importance stdlib match like android_frames then
floated above an exact match like androidx_art_metrics, so searching
androidx_ surfaced only the stdlib android_* tables and buried the
extension-server androidx_* tables below them.

Rank each match group lexicographically: bucketed fuzzy relevance first,
importance second. A clearly better match always wins, and importance
only decides the order between matches of comparable relevance. Rounding
the fuzzy score into coarse buckets is what lets importance break ties
between similarly-relevant tables.

Expose the match score from FuzzyFinder and extract the search and
ranking out of the Mithril view into pure searchTables and
searchAndRankTables functions with unit tests.

Bug: #6559

### DIFF
--- a/ui/src/base/fuzzy.ts
+++ b/ui/src/base/fuzzy.ts
@@ -22,6 +22,9 @@ export interface FuzzySegment {
 export interface FuzzyResult<T> {
   item: T;
   segments: FuzzySegment[];
+  // Relevance score of the match, higher is better. Exact matches score higher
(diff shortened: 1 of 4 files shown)


stevegolton (approved):
LGTM

PR #6727, opened 2026-07-15
Add jank tooltip descriptions. 

adyabr:
Provide users with jank type description when hovering over janky frames

### DIFF
--- a/ui/src/plugins/dev.perfetto.Frames/actual_frames_track.ts
+++ b/ui/src/plugins/dev.perfetto.Frames/actual_frames_track.ts
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import m from 'mithril';
 import {HSLColor} from '../../base/color';
 import {makeColorScheme} from '../../components/colorizer';
 import type {ColorScheme} from '../../base/color_scheme';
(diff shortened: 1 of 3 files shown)


ui/src/plugins/dev.perfetto.Frames/actual_frames_track.ts
```diff
@@ -67,6 +99,39 @@ export function createActualFramesTrack(
...
+      const row = slice.row;
+      const tag = useExperimentalJankForClassification
+        ? row.jank_tag_experimental
+        : row.jank_tag;
+      const jankType = row.jank_type;
+
+      if (tag && tag !== 'No Jank' && tag !== 'None') {
+        const elements: m.Vnode[] = [];
+        elements.push(
+          m('div', {style: 'font-weight: bold; margin-bottom: 4px;'}, `${tag}`),
```
stevegolton: Please use classes and move styles into their own css file rather than putting them inline. Same below. You can import a css file directly from this typescript file like `import './frames_track.ts'` or similar.

stevegolton (approved):
LGTM % styles nit

ui/src/plugins/dev.perfetto.Frames/actual_frames_track.ts
stevegolton: What's the `${tag}` for? Can we just use `tag` in directly or am i missing something?

PR #6728, opened 2026-07-15
Add Jank Badges 

adyabr:
### Problem Statement
In long trace captures (spanning seconds or minutes), timeline frame slices shrink down to sub-pixel or fractional-pixel widths when zoomed out. Consequently, performance anomalies—such as severe Jank frames (`Self Jank` 🔴, `Other Jank` 🟡, `Dropped Frame` 🚫, `Non-perceivable Jank` ⚪)—become completely invisible. Developers are forced to manually pan and zoom repeatedly across thousands of frames to discover where jank occurred.

### Solution
Implement a high-performance, generic **Jank Badges Framework** in `SliceTrack`:
1. **Zoom-Invariant Visibility:** Renders fixed-screen-size canvas overlay badges (constant 16px badges) directly over timeline events so performance anomalies stand out immediately at macro zoom levels.
2. **Spatial Clustering & Priority Matrix:** When events cluster together when zoomed out, a render priority matrix resolves overlapping badges cleanly per depth row:
   `Self Jank (40) > Other Jank (30) > Dropped Frame (20) > Non-perceivable Jank (10)`
3. **Precision Alignment:** As the user zooms in, badges lock onto the exact start timestamp of the slice.
4. **Frame Timeline Registration:** Registers standard Android frame jank badges (`Self Jank` 🔴, `Other Jank` 🟡, `Dropped Frame` 🚫, `Non-perceivable Jank` ⚪) in `ActualFramesTrack`.

### DIFF
--- a/ui/src/components/tracks/slice_track.ts
+++ b/ui/src/components/tracks/slice_track.ts
@@ -74,6 +74,8 @@ import {
 } from '../../base/renderer';
 import {cropText} from '../../base/string_utils';
 
+import type {VisualMarkerStyle} from './visual_marker';
+
 const SLICE_MIN_WIDTH_FOR_TEXT_PX = 5;
 const CHEVRON_WIDTH_PX = 10;
 
@@ -277,6 +279,16 @@ export interface SliceTrackAttrs<T extends DatasetSchema> {
    * Define buttons displayed on the track shell.
    */
   shellButtons?(): m.Children;
+
+  /**
+   * Provides fixed-size overlay markers for events of interest (e.g. Jank, Cadence Drop).
+   */
+  markerProvider?(row: T): VisualMarkerStyle | undefined;
+
+  /**
+   * Priority score generator for row markers when clustering overlapping markers.
+   */
+  markerPriority?(row: T): number;
 
   /**
    * Called once per render cycle before drawing. Return an array of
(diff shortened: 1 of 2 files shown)


stevegolton (changes requested):
I'll be honest, I don't love how this looks, it doesn't scale very well, and it adds a lot of complexity to the slice track for one use case.

Here are some other approaches that lean on the existing infrastructure:
- Move the janky frames to a different layer, which makes them mipmapped separately and rendered over the top of the non janky frames. This is what thread state tracks do so that non-sleeping slices are never swallowed by surrounding sleeping slices - see https://github.com/google/perfetto/blob/09d113473237cc0325eb38db766a166c5c3e774c/ui/src/plugins/dev.perfetto.Sched/thread_state_track.ts#L63
- Colour the non-janky frames a neutral or semi transparent grey to make the janky frames pop out more visually.
- Add a separate track just showing the janky frames under the existing frame timeline track. These could be rendered as  either standard slices or even instant events which have a fixed pixel size on screen similar to your markers. We don't support custom instant markers properly yet for the webgl renderer but it's planned.

I'd suggest starting with the first two as they are simple changes.

I think the grey is actually pretty effective for such a simple change:

image

Something like this: https://github.com/google/perfetto/pull/6737

PR #6729, opened 2026-07-15
Add Jank hint for potential skipped frames

adyabr:
### Problem Statement
When debugging performance in Perfetto, developers frequently encounter subtle stutter—such as skipped frame callbacks—that are not flagged on-device and appear as normal green frames. Because heuristic detection can yield false positives depending on app state, this classification is designed strictly as an offline **UI debugging feature** in Perfetto. It overlays a speculative **`⚠️ Potential Skipped Frame`** warning badge over cadence drops without modifying on-device metrics or trace schema.

### Solution
- **Offline Heuristic Engine:** Post-processes presentation time deltas to detect skipped frame callbacks.
- **Overlay Debug Hint (`⚠️`):** Renders a warning badge (`⚠️`) and informational hover hint over flagged frames without altering native frame colors or OS metrics.

### DIFF
--- a/ui/src/components/tracks/slice_track.ts
+++ b/ui/src/components/tracks/slice_track.ts
@@ -281,7 +281,7 @@ export interface SliceTrackAttrs<T extends DatasetSchema> {
   shellButtons?(): m.Children;
 
   /**
-   * Provides fixed-size overlay markers for events of interest (e.g. Jank, Cadence Drop).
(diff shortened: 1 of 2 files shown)


stevegolton (changes requested):
Same concerns as in https://github.com/google/perfetto/pull/6728#issuecomment-4990420851 - let's sort how to to deal with those first. I suspect this is a good candidate for a new track to be honest.

PR #6743, opened 2026-07-16
perfetto: fix preventing duplicate comments for deleted elements in proto_merger

BogdanLionte:


src/tools/proto_merger/proto_merger.cc
```diff
@@ -28,12 +28,28 @@ namespace proto_merger {
...
-         leading_comments[1] ==
-             " The following enums/messages/fields are not present upstream" &&
-         leading_comments[2].empty()) {
-    leading_comments.erase(leading_comments.begin(),
-                           leading_comments.begin() + 3);
+  auto is_warning = [](std::string_view s) {
+    return base::TrimWhitespace(s) ==
+           "The following enums/messages/fields are not present upstream";
+  };
+  auto is_empty = [](std::string_view s) {
```
sashwinbalaji: don't think this lambda is needed and it can be inlined.

src/tools/proto_merger/proto_merger.cc
```diff
@@ -28,12 +28,28 @@ namespace proto_merger {
...
 void StripDeletedElementComment(std::vector<std::string>& leading_comments) {
-  while (leading_comments.size() >= 3 && leading_comments[0].empty() &&
-         leading_comments[1] ==
-             " The following enums/messages/fields are not present upstream" &&
-         leading_comments[2].empty()) {
-    leading_comments.erase(leading_comments.begin(),
-                           leading_comments.begin() + 3);
+  auto is_warning = [](std::string_view s) {
+    return base::TrimWhitespace(s) ==
+           "The following enums/messages/fields are not present upstream";
```
sashwinbalaji: let's move this to be a const in proto_file_serializer.h and use the const everywherre and if you do that I think we can also remove this wrapper func

src/tools/proto_merger/proto_merger.cc
```diff
@@ -28,12 +28,28 @@ namespace proto_merger {
...
+  };
+  auto is_empty = [](std::string_view s) {
+    return base::TrimWhitespace(s).empty();
+  };
+
+  auto it = std::find_if(leading_comments.begin(), leading_comments.end(),
+                         [&](const std::string& s) { return is_warning(s); });
```
sashwinbalaji: &](const std::string& s) this wrapping is not needed I think you can do std::find_if(..,..,is_warning).

But once you move to const I think can be directly inlined.

src/tools/proto_merger/proto_file_serializer_unittest.cc
```diff
@@ -223,6 +223,43 @@ TEST(ProtoFileSerializerTest, MergeDoesNotDuplicateDeletedComment) {
       << out;
 }
 
+TEST(ProtoFileSerializerTest, MergeDoesNotDuplicateDeletedCommentFormatted) {
```
sashwinbalaji: nit: this is almost identical to above func MergeDoesNotDuplicateDeletedComment. maybe have a helper func to avoid duplicate ?

src/tools/proto_merger/proto_merger.cc
```diff
@@ -28,12 +28,28 @@ namespace proto_merger {
...
+  while (it != leading_comments.end()) {
```
sashwinbalaji: nit: though the logic is absolutely correct but going back and forth with iterator and erasing while iterating  makes it a bit difficult to reason about and also this is slow path so performance doesn't matter that much so we can keep a seprate vector to make it simpler

so how about:

```
void StripDeletedElementComment(std::vector<std::string>& leading_comments) {
  std::vector<std::string> kept;
  kept.reserve(leading_comments.size());
  for (size_t i = 0; i < leading_comments.size(); i++) {
    if (base::TrimWhitespace(leading_comments[i]) !=
        kNotPresentUpstreamComment) {
      kept.push_back(std::move(leading_comments[i]));
      continue;
    }
    // |i| is the warning line: don't keep it, and also drop one blank
    // comment line before it (already in |kept|) and one after it.
    if (!kept.empty() && base::TrimWhitespace(kept.back()).empty())
      kept.pop_back();
    if (i + 1 < leading_comments.size() &&
        base::TrimWhitespace(leading_comments[i + 1]).empty())
      i++;  // Skips the blank line: the loop increment moves past it.
  }
  leading_comments = std::move(kept);
}
```

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file_serializer_unittest.cc

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file_serializer.h
```diff
@@ -22,6 +22,9 @@
 namespace perfetto {
 namespace proto_merger {
 
+constexpr char kDeletedCommentWarning[] =
```
sashwinbalaji: nit: add explicit inline keyword.

sashwinbalaji (approved):

src/tools/proto_merger/proto_file_serializer.h
BogdanLionte: Done

PR #6753, opened 2026-07-16
ui: Tint track shell when the track is selected

### LINKED ISSUE #6749: Enhancement idea: Highlight the track of the selected block
(I'm sorry for flooding you with issues. That's the downside to answering so amazingly fast 😉 - feel free to not answer so fast this time.)

I think it may be a good idea to add a highlight to the track of the selected block. Similar to what you have in spreadsheets:

Image

(It could be just the header, or the background of the entire track, as it is in code editors.)

The reason is that I find myself struggling to find the track header which describes a specific block. I need to carefully trace the line to the left. If it were highlighted, it would have been a bit easier.

For the X axis there's already a similar feature, where there's a marking of a time slice when I select a block.

Thanks again!
Noam

stevegolton:
Tracks that are part of the current selection - whether that's an area selection or a single track/event selection - now get a subtle accent tint on their shell, so it's easier to see at a glance which tracks the selection covers.

Fixes: https://github.com/google/perfetto/issues/6749

Area selection:

image

Single selection:

image

### DIFF
--- a/ui/src/core_plugins/dev.perfetto.Timeline/track_view.ts
+++ b/ui/src/core_plugins/dev.perfetto.Timeline/track_view.ts
@@ -201,6 +201,7 @@ export class TrackView {
         collapsible: collapsible && node.hasChildren,
         collapsed: collapsible && node.collapsed,
         highlight: this.isHighlighted(),
+        selected: this.isSelected(),
         summary: node.isSummary,
         reorderable,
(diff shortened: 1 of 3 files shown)


LalitMaganti (approved):

LalitMaganti:
Love it, thanks so much! Big usability improvement :)

PR #6755, opened 2026-07-16
Fix RepeatedFieldIterator packed fields

khokhlov962:
When a protozero-generated decoder (`RepeatedFieldIterator<T>`) reads a proto2 repeated scalar field that was serialized using packed (kLengthDelimited) format on the wire, it fails to unpack the repeated scalar field. Even worse, in release mode without PERFETTO_DCHECK, it dereferences the field memory address and returns the raw heap address rather than the valid data integer.

This change teaches `RepeatedFieldIterator<T>` to transparently unpack `kLengthDelimited` buffers for scalar types, seamlessly activating an internal packed iteration state without breaking the interface or requiring massive regeneration across the codebase.

TAG=agy

### DIFF
--- a/include/perfetto/protozero/proto_decoder.h
+++ b/include/perfetto/protozero/proto_decoder.h
@@ -118,10 +118,16 @@ class RepeatedFieldIterator {
   RepeatedFieldIterator()
       : field_id_(0u), iter_(nullptr), end_(nullptr), last_(nullptr) {}
 
-  explicit operator bool() const { return iter_ != end_; }
+  explicit operator bool() const {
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Nope this is working as intended: we don't want to pay this cost on every field access unless it's required.

In fact we even have a UnifiedRepeatedFieldIterator [1] which is designed *specifically* to dispatch to packed/non-packed fields in cases where we don't have control over how bytes are written and need to support both.

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/include/perfetto/protozero/proto_decoder.h;l=276

PR #6756, opened 2026-07-16
tp: add (pid)/(tid) field options and annotate frameworks_base

zezeozue:
Adds two custom field options (in field_options.proto) for integer fields that hold OS ids:

- (perfetto.protos.pid): the field is a process id. Alongside the raw value, trace_processor emits two extra args keyed by replacing the trailing "pid": "<name>upid" (the trace-unique upid) and "<name>process_name". e.g. caller_pid -> caller_upid + caller_process_name.
- (perfetto.protos.tid): the same for thread ids, emitting "<name>utid" and "<name>thread_name".

The upid/utid come from GetProcessOrNull/GetThreadOrNull, which are plain lookups: if the pid/tid was never seen, no extra arg is emitted and no process/thread is created.

Also annotates the pid, caller_pid, sender_pid and receiver_pid fields in frameworks_base_track_event.proto.

### DIFF
--- a/protos/perfetto/trace/field_options.proto
+++ b/protos/perfetto/trace/field_options.proto
@@ -26,4 +26,10 @@ extend google.protobuf.FieldOptions {
   // Fully-qualified flags enum for a bitmask int field; trace_processor expands
   // the mask into one string arg per set flag.
   optional string flags_enum = 73921;
+
+  // Marks an int field as a pid; emits companion upid + process_name args.
+  optional bool pid = 73922;
+
+  // Marks an int field as a tid; emits companion utid + thread_name args.
+  optional bool tid = 73923;
 }
--- a/protos/perfetto/trace/field_options.proto
+++ b/protos/perfetto/trace/field_options.proto
@@ -27,9 +27,11 @@ extend google.protobuf.FieldOptions {
   // the mask into one string arg per set flag.
   optional string flags_enum = 73921;
 
-  // Marks an int field as a pid; emits companion upid + process_name args.
-  optional bool pid = 73922;
+  // Marks an int field as a process id; trace_processor emits a companion
+  // "<...>upid" arg (a trailing "pid" replaced with "upid").
+  optional bool is_pid = 73922;
 
-  // Marks an int field as a tid; emits companion utid + thread_name args.
-  optional bool tid = 73923;
+  // Marks an int field as a thread id; trace_processor emits a companion
+  // "<...>utid" arg (a trailing "tid" replaced with "utid").
+  optional bool is_tid = 73923;
 }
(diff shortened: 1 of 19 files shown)


protos/perfetto/trace/field_options.proto
```diff
@@ -26,4 +26,10 @@ extend google.protobuf.FieldOptions {
   // Fully-qualified flags enum for a bitmask int field; trace_processor expands
   // the mask into one string arg per set flag.
   optional string flags_enum = 73921;
+
+  // Marks an int field as a pid; emits companion upid + process_name args.
+  optional bool pid = 73922;
```
LalitMaganti: We need something a bit more specific than "pid" - if I saw this annotation name in the proto I would be *extremely* confused.

At the least, this should be is_pid/is_tid.

src/trace_processor/importers/proto/args_parser.cc
```diff
@@ -40,12 +41,14 @@ ArgsParser::ArgsParser(int64_t packet_timestamp,
                        BoundInserter& inserter,
                        TraceStorage& storage,
                        PacketSequenceStateGeneration* sequence_state,
-                       bool support_json)
+                       bool support_json,
+                       ProcessTracker* process_tracker)
```
LalitMaganti: Two things:
1) Why is this a pointer when the other fields are references.
2) context should always come before configuration. This should live just beside TraceStorage.

src/trace_processor/util/proto_to_args_parser.cc
```diff
@@ -791,6 +803,41 @@ base::Status ProtoToArgsParser::AddFlags(uint32_t enum_descriptor_idx,
   return base::OkStatus();
 }
 
+namespace {
+// "caller_pid" + ("pid","upid") -> "caller_upid"; else appends "_<to>".
+std::string ReplaceKeySuffix(const std::string& key,
```
LalitMaganti: Can you avoid creating a string here? This is going to allocate all the time. Would be nicer at least to have a string buffer and reuse that instead.

src/trace_processor/importers/proto/args_parser.cc
```diff
@@ -86,6 +89,44 @@ void ArgsParser::AddBoolean(Id flat_key, Id key, bool value) {
...
+  }
+}
+
+void ArgsParser::AddProcessName(Id flat_key, Id key, int64_t pid) {
+  if (!process_tracker_)
+    return;
+  auto upid = process_tracker_->GetProcessOrNull(pid);
+  if (!upid)
+    return;
+  if (auto name = storage_.process_table()[*upid].name()) {
```
LalitMaganti: Sorry but this is *very* much not safe. Process names can appear *after* the event and/or they can change (and we keep the last one). We cannot do this except by buffering *all* args sets until the end of the trace and fixing them all up there (and extraordinarily expensive operation I don't see the justification for doing).

Why can this not be handled specially at the UI level if that's what you're worried about?

zezeozue: Removed process name. Will do it in ui as follow up

LalitMaganti:
Fwiw doing it in the UI will need discussion. It might need us to make some trace processor changes also to communicate metadata about the args (e.g. the fact that these fields reference upids etc).

LalitMaganti (approved):

PR #6758, opened 2026-07-16
ui: record page: extract a shared base for the adb recording targets

zezeozue:
The three adb-based RecordingTargets (WebUSB, WebSocket, Web Device Proxy) duplicate their connected / getServiceState / startTracing / disconnect logic and the AsyncLazy<AdbDevice> field, differing only in how they connect. Pull the shared behaviour into a generic AdbRecordingTarget<T> base so each transport only provides connectIfNeeded() plus its identity (id/name/preflight). AdbDevice gains an abstract `connected` getter so the base can read it generically.

Pure refactor, in preparation for sharing a runShellCommand() across the transports that an upcoming probe pre-flight hook needs. No behaviour change beyond consolidating the identical "<transport> transport disconnected" error into a single message.

### DIFF
--- a/ui/src/plugins/dev.perfetto.RecordTraceV2/adb/adb_device.ts
+++ b/ui/src/plugins/dev.perfetto.RecordTraceV2/adb/adb_device.ts
@@ -39,6 +39,9 @@ export abstract class AdbDevice {
 
   abstract close(): void;
 
+  // True while the underlying transport is connected.
+  abstract get connected(): boolean;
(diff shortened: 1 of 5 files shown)


stevegolton (commented):
I'm not entirely convinced TBH. The duplicated code is cheaper and clearer than an inheritance hierarchy IMO.

I'd just add the additional function in three places and factor out any common code into a helper.

PR #6767, opened 2026-07-17
tbV2: relocate fully-read scraped chunks instead of rewriting them

sashwinbalaji:
Fixes a data-loss race seen in 24h field traces with sporadic writers.
Fingerprint: orphan_continuation = overwrite = sequence_packet_loss = 1 on the
buffer, one packet with previous_packet_dropped = 73.

The race:
- A flush scrapes a writer's half-full chunk; the incomplete copy is pinned in
  the buffer and read over time.
- Hours later the writer commits the chunk. The recommit rewrote it in place,
  at the now-stale offset, with the write cursor almost a lap ahead: the
  never-read fragments could be overwritten before the next read pass.
- If the chunk ends with kFragBegin, that eviction destroys the begin fragment
  and its continuation later reads back as an orphan.

The fix:
- If the scraped copy is fully consumed and the final commit brings new
  fragments, erase the stale copy and relocate the chunk to the write cursor,
  reusing the re-admit path for evicted chunks (#5372).
- previously_consumed_payload skips what was already read, so nothing is
  duplicated; ChunkIDs stay consecutive, so no read gap.
- Rescrapes, partially-consumed and out-of-order recommits keep the in-place
  rewrite, as do kDiscard buffers (they never lap, and the write path could hit
  DiscardWrite() and drop the fragments being recovered).

New chunks_relocated stat (BufferStats field 20) counts these, mirroring
chunks_rewritten. Field 20 was never assigned in BufferStats: the CL below
added bytes_discarded_per_buffer = 20 to FilterStats but bumped the "Next id"
above BufferStats, leaving a phantom hole. https://r.android.com/2882818

Four new tests cover the race itself, the rescrape / unread-payload / kDiscard
exclusions, and the case where the continuation chunk is already buffered.

Bug: 518755701.

### DIFF
--- a/src/tracing/service/trace_buffer_v2.cc
+++ b/src/tracing/service/trace_buffer_v2.cc
@@ -881,9 +881,9 @@ void TraceBufferV2::CopyChunkUntrusted(
     const bool commit_adds_new_data =
         chunk_complete && all_frags_size > recommit_chunk->payload_size;
 
-    // EraseCurrentChunk() only supports the first chunk of a sequence. Chunks
-    // committed out of order keep the in-place rewrite: the race needs a stale
-    // offset, while out-of-order commits resolve within one scrape cycle.
+    // EraseCurrentChunk() only supports the first chunk of a sequence. Later
+    // chunks may stay physically ahead of the relocated one, which is fine:
+    // reads follow chunk_list, which is ordered by ChunkID and not by offset.
     const bool copy_is_first_chunk_of_seq =
         *chunk_list.begin() == OffsetOf(recommit_chunk);
 
@@ -924,7 +924,7 @@ void TraceBufferV2::CopyChunkUntrusted(
     stats_.set_chunks_relocated(stats_.chunks_relocated() + 1);
     previously_consumed_payload = recommit_chunk->payload_size;
     internal::ChunkSeqIterator(this, &seq).EraseCurrentChunk();
-  }
+  }  // if (recommit_chunk)
 
   // If there isn't enough room from the given write position: write a padding
   // record to clear the end of the buffer, wrap and start at offset 0.
--- a/src/tracing/service/trace_buffer_v2.cc
+++ b/src/tracing/service/trace_buffer_v2.cc
@@ -863,23 +863,67 @@ void TraceBufferV2::CopyChunkUntrusted(
       PERFETTO_DCHECK(suppress_client_dchecks_for_testing_);
       return;
     }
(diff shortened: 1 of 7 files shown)


src/tracing/service/trace_buffer_v2.cc
```diff
@@ -863,23 +863,67 @@ void TraceBufferV2::CopyChunkUntrusted(
...
+    const bool copy_fully_consumed = recommit_chunk->payload_avail == 0;
+
+    // The producer's final commit (not another scrape), with new fragments.
+    const bool commit_adds_new_data =
+        chunk_complete && all_frags_size > recommit_chunk->payload_size;
+
+    // EraseCurrentChunk() only supports the first chunk of a sequence. Chunks
+    // committed out of order keep the in-place rewrite: the race needs a stale
+    // offset, while out-of-order commits resolve within one scrape cycle.
+    const bool copy_is_first_chunk_of_seq =
```
primiano, line 887: Hmm can we discuss this offline?
I'm puzzled between:
A. We can safely do this only if this is the *ONLY* chunk of the sequence
B. We can do what you are doing here

Essentially i'm thinking of the case where we scrape both C1 and C2, and then recommit C1. Now in the buffer we'll rewrite them as [C2,... C1 (new)]

Hmm but then the out-of-order logic should still work because in the chunk_list they would still be stored as [C1, C2new] even though they are in different physical positions in the buffer.

EDIT: okay I think this can work actually. I just kept the comment to reflect my loud thinking. If there is something that doesn't sound right ping me over IM.

src/tracing/service/trace_buffer_v2.cc
```diff
@@ -863,23 +863,67 @@ void TraceBufferV2::CopyChunkUntrusted(
...
-    stats_.set_chunks_rewritten(stats_.chunks_rewritten() + 1);
-    return;
+
+    // Erase the stale copy and fall through to the write path below, which
+    // re-creates the chunk at the write cursor.
+    TRACE_BUFFER_V2_DLOG("  Relocating consumed scraped chunk %u", chunk_id);
+    stats_.set_chunks_relocated(stats_.chunks_relocated() + 1);
+    previously_consumed_payload = recommit_chunk->payload_size;
+    internal::ChunkSeqIterator(this, &seq).EraseCurrentChunk();
   }
```
primiano: can you add a `// if (recommit_chunk)` next to the close curly brace? this block became quite bigger now, it helps readability

primiano (approved):

PR #6770, opened 2026-07-17
ui: Upgrade Playwright and fix CodeMirror EditorView import

stevegolton:
Upgrade @playwright/test to 1.61.1. In addition, update EditorView import in ui/src/widgets/editor.ts to be imported directly from @codemirror/view instead of the codemirror umbrella package, avoiding type conflicts from version mismatches.


LalitMaganti (approved):

stevegolton:
The tests just flat out don't work with playwright 1.61.1 - pinned to the previous 1.58.2 for now

PR #6775, opened 2026-07-18
tp: add async attribution to stack sampling

LalitMaganti:
Add TaskContext.async_id for attributing a sample to a stackful async
context (goroutine, fiber, ...), described by an AsyncContextDescriptor
interned via the StackSampleInternedData extension.

The descriptor's name and kind fold onto __intrinsic_stack_sample_task_context
as async_name / async_kind. Generate the extension descriptor and register
it with the diff-test harness so interned extension fields can be used in
textproto traces.


sashwinbalaji (approved):
tests need update

PR #6777, opened 2026-07-18
tools: add trace_video_conv.py to extract display-video to mp4

zezeozue:
Reads the android.display.video frames from a trace via the trace_processor python API and muxes them into an .mp4 with ffmpeg: list the video streams, convert the whole video, clip to a time range or a SQL-selected region, change playback speed, and lay two traces' videos side by side (captioned) to compare them. Documents it in the Screen Recording data-source page.

### DIFF
--- a/python/tools/trace_video_conv.py
+++ b/python/tools/trace_video_conv.py
@@ -51,14 +51,13 @@ import subprocess
 import sys
 import tempfile
 
-# Bootstrap the in-repo perfetto python library (same as the other tools/).
-PYTHON_DIR = os.path.join(
-    os.path.dirname(
-        os.path.dirname(os.path.dirname(os.path.abspath(__file__)))), 'python')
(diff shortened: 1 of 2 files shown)


python/tools/trace_video_conv.py
```diff
@@ -0,0 +1,411 @@
...
+import shutil
+import statistics
+import subprocess
+import sys
+import tempfile
+
+# Bootstrap the in-repo perfetto python library (same as the other tools/).
+PYTHON_DIR = os.path.join(
+    os.path.dirname(
+        os.path.dirname(os.path.dirname(os.path.abspath(__file__)))), 'python')
```
LalitMaganti: This should not add the Perfetto library just the Python one and should import python.perfetto.trace_processor.

LalitMaganti (approved):

PR #6788, opened 2026-07-20
fix: missing Runnable states and wakeup info due to sched_waking timing

zhubin3:
## Problem

sched_waking fires while the thread is still in Running state, causing PushWakingEvent to incorrectly mark it as a spurious wakeup. This results in missing Runnable (R) states in the thread_state table.

## Fix

Use sched_wakeup instead of sched_waking for non-compact format traces. sched_wakeup fires after the thread enters S state, so PushWakingEvent correctly creates R states.

Interrupt context (irq_context) is preserved by storing common_flags from sched_waking in a pending map and retrieving it when sched_wakeup fires.

## Changes

- : Replace sched_waking case with sched_wakeup, add ParseSchedWakeup function
- : Add ParseSchedWakeup declaration
- : Add pending_waking_common_flags_ storage and lookup logic
- : Add pending_waking_common_flags_ member

## Testing

Verified with multiple traces:
- test.pftrace: missing R states reduced from 3390 to 72 (-97.9%)
- retest.pftrace: missing R states reduced from 2944 to 38 (-98.7%)
- 74-1: missing R states reduced from 117219 to 0 (-100%)

Remaining missing R states are all information loss (trace boundary), not code issues.

## Note

- Compact format traces are not affected (use PushSchedWakingCompact path)
- No unit tests added yet, can add if needed

### DIFF
--- a/src/trace_processor/importers/common/thread_state_tracker.cc
+++ b/src/trace_processor/importers/common/thread_state_tracker.cc
@@ -79,6 +79,12 @@ void ThreadStateTracker::PushWakingEvent(int64_t event_ts,
     // in the |thread_state| table but we track in the |sched_wakeup| table.
     // The |thread_state_id| in |sched_wakeup| is the current running/runnable
     // event.
+    //
+    // Store common_flags so that a later sched_wakeup (which fires when the
+    // thread is actually blocked) can pick up the interrupt context.
+    if (common_flags.has_value()) {
+      pending_waking_common_flags_[utid] = *common_flags;
+    }
     std::optional<uint32_t> irq_context =
         common_flags
             ? std::make_optional(CommonFlagsToIrqContext(*common_flags))
@@ -90,9 +96,20 @@ void ThreadStateTracker::PushWakingEvent(int64_t event_ts,
   }
 
   // Close the sleeping state and open runnable state.
+  // If common_flags was not provided (e.g. sched_wakeup from non-compact
+  // format), try to pick up the flags from a preceding sched_waking event
+  // that fired while the thread was still running.
+  std::optional<uint16_t> effective_common_flags = common_flags;
+  if (!common_flags.has_value()) {
+    auto it = pending_waking_common_flags_.find(utid);
+    if (it != pending_waking_common_flags_.end()) {
+      effective_common_flags = it->second;
+      pending_waking_common_flags_.erase(it);
+    }
+  }
   ClosePendingState(event_ts, utid, false);
   AddOpenState(event_ts, utid, runnable_string_id_, std::nullopt, waker_utid,
-               common_flags);
+               effective_common_flags);
 }
 
 void ThreadStateTracker::PushNewTaskEvent(int64_t event_ts,
--- a/src/trace_processor/importers/common/thread_state_tracker.h
+++ b/src/trace_processor/importers/common/thread_state_tracker.h
@@ -17,6 +17,8 @@
 #ifndef SRC_TRACE_PROCESSOR_IMPORTERS_COMMON_THREAD_STATE_TRACKER_H_
 #define SRC_TRACE_PROCESSOR_IMPORTERS_COMMON_THREAD_STATE_TRACKER_H_
 
+#include <unordered_map>
+
 #include "src/trace_processor/storage/trace_storage.h"
 #include "src/trace_processor/types/destructible.h"
 #include "src/trace_processor/types/trace_processor_context.h"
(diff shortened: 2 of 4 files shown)


LalitMaganti:
Sorry but we cannot accept this PR. If you go through the history of that us parsing ftrace you'll actually see that we used to parse sched_wakup but we very intentionally changed to use waking instead after lots of discussions with Linux kernel scheduler maintainers on this being the right event to use. We're not going to change that unless you have proof from them that wakeup has suddenly changed to be the correct event.

zhubin3:
Thank you for the response. I'd like to clarify the issue I'm trying to solve.

  My main goal is to fix the missing Runnable states problem.

  1. The vast majority of Runnable states already start from sched_waking, which is the same as the current behavior.

  2. Only for the cases where Runnable states are missing, the current Perfetto parsing behavior is: no Runnable state exists, the thread transitions directly from Sleep to Running, and the wakeup information is not visible. I believe this might be an issue, so I only modified this specific case to start the Runnable state from sched_wakeup instead, so that Perfetto can display the wakeup information. I think this would be more user-friendly for developers analyzing performance issues.

  In summary, I'm not changing how the majority of wakeups are handled - I'm only addressing the edge cases where Runnable states are currently missing.

  ## Root Cause Analysis

  The issue occurs in `PushWakingEvent` (thread_state_tracker.cc). When `sched_waking` fires while the thread is still in Running state, the code marks it as a spurious wakeup and skips creating the Runnable state:

  ```cpp
  if (!IsBlocked(last_row_ref.state())) {
      // Thread is still Running → marked as spurious wakeup
      // R state is NOT created
      storage_->mutable_spurious_sched_wakeup_table()->Insert(...);
      return;
  }

  The problem is: sched_waking fires during the wakeup path (try_to_wake_up entry), before the thread actually enters S state via sched_switch. So when PushWakingEvent checks IsBlocked(), the thread is still Running, and the wakeup is incorrectly marked as spurious.

  Analysis: ugc_coldstart.pftrace

  Missing Runnable States (2331 cases)

  ┌────────────┬────────────┬───────────┬─────────────┐
  │ Transition │ Before Fix │ After Fix │ Improvement │
  ├────────────┼────────────┼───────────┼─────────────┤
  │ S→Running  │ 1776       │ 0         │ -100%       │
  ├────────────┼────────────┼───────────┼─────────────┤
  │ D→Running  │ 459        │ 0         │ -100%       │
  ├────────────┼────────────┼───────────┼─────────────┤
  │ I→Running  │ 96         │ 0         │ -100%       │
  ├────────────┼────────────┼───────────┼─────────────┤
  │ Total      │ 2331       │ 0         │ -100%       │
  └────────────┴────────────┴───────────┴─────────────┘

  Example: Douyin main thread (tid=28639)

  Before fix (no Runnable state, no wakeup info):
<img width="2724" height="1212" alt="before" src="https://github.com/user-attachments/assets/29fd23e7-c474-4d8b-99b1-e6616e9e8ce7" />

  After fix (Runnable state with wakeup info):
<img width="2566" height="1304" alt="after" src="https://github.com/user-attachments/assets/4ccf12e1-378b-460f-a1e2-ef70f8c71f68" />

 Scenario where the issue occurs:
<img width="2750" height="1148" alt="Reason" src="https://github.com/user-attachments/assets/e91e7dfe-d9b5-44a0-b3f0-34579c2f0009" />

  This issue happens when sched_waking fires before sched_switch. For this scenario, we use sched_wakeup to start the Runnable state. There are 2,330 such cases in this trace, and the count is likely higher under high load.

  The fix makes it possible for developers to see the wakeup chain when analyzing performance issues, which was previously invisible due to missing Runnable states.

zhubin3:
[trace_and_png.zip](https://github.com/user-attachments/files/30186003/trace_and_png.zip)

LalitMaganti:
From speaking with the scheduler folks, this is infact not an issue and the behavior of Perfetto in this case I believe to be correct. There are indeed cases where wakeups fail to go on runqueues which can cause these events to be missed and in that case, it's correct that you should *not* mark the process as runnable at that point becuase they are kernel bugs that the transition is missing. I have discussed this a lot with people over the years, I don't want to change this without their approval

I would only change this behaviour if those folks told me actually wakeup was correct the correct thing to use. This also causes lots of problems for us though because e.g. compact scheduling uses waking, all our configs deployed everywhere use waking etc.

zhubin3:
Thank you for the explanation. I understand the design decision now

Thanks for the review!

zhubin3:
Closing this PR based on the feedback. Thanks again.

PR #6789, opened 2026-07-20
ui: DNS: Experimental sticky parent tracks

stevegolton:
This is an experimental attempt to make all parent tracks sticky while also displaying their track canvas content (e.g. slices, counters, etc).


stevegolton:
Synthetic trace to test with (featuring deeply nested tracks with flows):

[nested_tracks.perfetto-trace.zip](https://github.com/user-attachments/files/30182960/nested_tracks.perfetto-trace.zip)

Use the flag #!/flags/dev.perfetto.TimelineControlledVerticalScrolling to disable and enable controlled scrolling. With it off we use native browser scrolling in the timeline but the sticky track content lags the DOM content. When on (default) we control the scroll offset of the element manually which results in slightly more laggy scrolling but the track content and DOM elements are are coherent.

PR #6793, opened 2026-07-20
perfetto: make the StackSample inline callstack fully inline

LalitMaganti:
The inline callstack variant of StackSample reused the profiling
Callstack message, whose frames are interning references. That defeats
the point of an inline variant: a producer should be able to emit a
sample without setting up interned data, matching the intent behind
the original TrackEvent inline callstack discussion.

Hoist TrackEvent's nested Callstack message to a top-level
InlineCallstack in its own inline_callstack.proto and use it for both
TrackEvent and StackSample. Inline stack sample frames are interned
into a dummy mapping, the same way TrackEvent inline callstacks are.

This is a breaking change for code referencing the generated
TrackEvent.Callstack types; the wire format is unchanged. Add changelog
entries for this and for the transport-neutral stack sampling work.


sashwinbalaji (approved):
lgtm but yeah need to monitor the autorolls

PR #6794, opened 2026-07-20
tp: add support for multiple statements in query

LalitMaganti:
Separate the queries by a single blank line. This has become necessary
because AI these days really loves doing N queries simultaneously
when gaining context.


src/trace_processor/trace_processor_impl.cc
```diff
@@ -647,6 +647,58 @@ Iterator TraceProcessorImpl::ExecuteQuery(const std::string& sql) {
...
+    const std::string& sql,
+    uint32_t* offset) {
+  PERFETTO_CHECK(offset);
+  PERFETTO_TP_TRACE(metatrace::Category::API_TIMELINE, "EXECUTE_NEXT_STATEMENT",
+                    [&](metatrace::Record* r) { r->AddArg("query", sql); });
+
+  // Same normalization as ExecuteQuery. Offsets are relative to the
+  // normalized SQL: since the normalization is deterministic, an offset
+  // returned by one call stays valid when the same |sql| is passed back.
+  std::string non_breaking_sql = base::ReplaceAll(sql, "\u00A0", " ");
```
sashwinbalaji: hmm do we want to run replace all on entire sql for each statement and also we have a substr below for every call.

LalitMaganti: So yes you're right it would be better if we didn't do this here and were more optimized. On the other hand, O(sql) is *not* our problem in Perfetto so for now I think it's OK.

LalitMaganti: filed https://github.com/google/perfetto/issues/6799 as followup

sashwinbalaji (approved):

PR #6795, opened 2026-07-20
tp: add minimal flatbuffer reader and writer

LalitMaganti:
Trace processor needs to emit and parse small flatbuffer-encoded
metadata (Arrow IPC framing) without taking a dependency on the
flatbuffers library, which would have to be vendored for every
embedder including the Wasm and Android builds.

The writer builds buffers back-to-front so offsets are naturally
forward-pointing. It tracks the maximum requested alignment and pads
the final size to a multiple of it: element positions are aligned
relative to the buffer end, so an aligned total size is what makes
them aligned relative to the start, which is what strict flatbuffers
verifiers check. String length prefixes are padded the same way.

The reader is designed for untrusted input: every table, vtable,
string and vector access is bounds-checked in 64-bit offset arithmetic
and malformed or absent data uniformly reads back as defaults.

### DIFF
--- a/src/trace_processor/util/flatbuffer_reader.h
+++ b/src/trace_processor/util/flatbuffer_reader.h
@@ -1,0 +1,202 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_TRACE_PROCESSOR_UTIL_FLATBUFFER_READER_H_
+#define SRC_TRACE_PROCESSOR_UTIL_FLATBUFFER_READER_H_
+
+#include <cstdint>
+#include <cstring>
+#include <optional>
+#include <string_view>
+
+#include "perfetto/base/compiler.h"
+
+namespace perfetto::trace_processor::util {
+
+// Minimal read-only FlatBuffer API, sufficient for parsing Arrow IPC metadata
+// and similar small schemas without pulling in the flatbuffers library.
+//
+// Every access is bounds-checked against the buffer passed at construction:
+// this reader is designed to be safe on untrusted input. Absent or malformed
+// fields uniformly read back as defaults (empty views, invalid readers),
+// never out-of-bounds accesses.
+//
+// FlatBuffer wire format recap:
+//   - A Table starts with an int32 "soffset" pointing backwards to its vtable.
+//   - VTable layout: [u16 vtable_size, u16 table_size, u16 field0_off, ...]
+//   - Field offsets are relative to the table start; 0 means "not present".
+//   - Strings: u32 length followed by UTF-8 bytes + null terminator.
+//   - Vectors: u32 element count followed by the elements.
+//   - Offsets (to strings, tables, vectors) are u32 relative to the offset's
+//     own position: target = &offset + *offset.
+//   - Structs are stored inline (no vtable indirection).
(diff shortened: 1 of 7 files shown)


src/trace_processor/util/flatbuffer_writer.cc
```diff
@@ -0,0 +1,224 @@
...
+  }
+  uint32_t num_slots = fields_.empty() ? 0 : max_index + 1;
+  uint32_t vtable_size = 4 + num_slots * 2;
+  uint32_t table_size = table_off - table_body_start_;
+
+  // Each vtable entry is the byte offset from the table start (soffset
+  // position) to the field data. 0 means "not present".
+  std::vector<uint16_t> vt_entries(num_slots, 0);
+  for (const auto& f : fields_) {
+    vt_entries[f.index] = static_cast<uint16_t>(table_off - f.off_from_end);
```
safayat-google, line 189: nits, log some warning when table_off - f.off_from_end) > uint16_t max?

src/trace_processor/util/flatbuffer_reader.h
```diff
@@ -0,0 +1,202 @@
...
+
+  // Reads a sub-table field (offset -> child table). Invalid reader if
+  // absent or malformed.
+  FlatBufferReader Table(uint32_t field_index) const;
+
+  // Reads a vector of inline scalars or packed structs of type T. Empty view
+  // if absent, malformed or extending past the end of the buffer.
+  template <typename T>
+  FlatBufferScalarVec<T> VecScalar(uint32_t field_index) const {
+    uint32_t count = 0;
```
safayat-google, line 158: nits, should we also check for out of bound similar to `FlatBufferTableVec`?

src/trace_processor/util/flatbuffer_reader_unittest.cc
```diff
@@ -0,0 +1,395 @@
...
+  auto reader =
+      FlatBufferReader::GetRoot(buf.data(), static_cast<uint32_t>(buf.size()));
+  ASSERT_TRUE(reader.has_value());
+  EXPECT_EQ(reader->String(0), "");
+  EXPECT_FALSE(reader->Table(0));
+  EXPECT_EQ(reader->VecScalar<int32_t>(0).size(), 0u);
+  EXPECT_EQ(reader->VecTable(0).size(), 0u);
+  EXPECT_EQ(reader->VecString(0).size(), 0u);
+}
+
```
safayat-google, line 452: nits, should we add more tests to cover index access for out size or maximum boundary for major supported types?

safayat-google (approved):

src/trace_processor/util/flatbuffer_writer.h
```diff
@@ -0,0 +1,137 @@
...
+
+  // Prepends a single scalar value.
+  template <typename T>
+  void Prepend(T val) {
+    GrowIfNeeded(sizeof(T));
+    head_ -= sizeof(T);
+    memcpy(&buf_[head_], &val, sizeof(T));
+  }
+
+  uint32_t GetSize() const {
```
sashwinbalaji: driveby comment: this is exactly same as Size() func above so can be removed.

src/trace_processor/util/flatbuffer_writer.cc
LalitMaganti: Done - EndTable now logs an error when the table size overflows the 16-bit vtable encoding. That single check also bounds every vtable entry, since each entry is smaller than table_size.

src/trace_processor/util/flatbuffer_reader.h
LalitMaganti: Done - FlatBufferScalarVec::operator[] now returns a default-constructed value for out-of-range indices, matching FlatBufferTableVec/FlatBufferStringVec.

src/trace_processor/util/flatbuffer_reader_unittest.cc
LalitMaganti: Done - added VecIndexOutOfBoundsReturnsDefault covering the last valid index, one-past-the-end and extreme indices for scalar (int64/double/uint8), string and table vectors.

src/trace_processor/util/flatbuffer_writer.h
LalitMaganti: Done - removed GetSize(); the writer uses size() internally now.

PR #6806, opened 2026-07-21
tp: add support for multiple statements in query

LalitMaganti:
Separate the queries by a single blank line. This has become necessary
because AI these days really loves doing N queries simultaneously
when gaining context.


LalitMaganti:
Created by accident, this change is tracked in #6794.

PR #6808, opened 2026-07-21
tp: allow TarWriter to write to an abstract sink

LalitMaganti:
TarWriter could only write to a file descriptor, which makes it
unusable for streaming an archive over an RPC connection. Introduce a
TarWriterSink interface with buffer- and callback-backed
implementations; the existing path/fd constructors keep working via an
internal fd sink.

Also make the error handling suitable for streaming consumers: a
failing callback aborts the write and propagates its status,
finalization is an explicit Status-returning operation (the destructor
only finalizes best-effort and never crashes on a broken pipe), and
the new StreamFile API streams a member's content into the archive
without buffering it in memory first.

### DIFF
--- a/src/trace_processor/util/tar_writer.cc
+++ b/src/trace_processor/util/tar_writer.cc
@@ -17,12 +17,14 @@
 #include "src/trace_processor/util/tar_writer.h"
 
 #include <fcntl.h>
-#include <algorithm>
+#include <cstdint>
 #include <cstdio>
 #include <cstring>
 #include <ctime>
+#include <memory>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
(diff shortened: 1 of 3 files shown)


src/trace_processor/util/tar_writer.cc
```diff
@@ -152,43 +130,202 @@ base::Status TarWriter::CreateAndWriteHeader(const std::string& filename,
...
+  if (!writer_) {
+    return;
+  }
+  base::Status status = Finalize();
+  PERFETTO_CHECK(status.ok());
+}
+
+base::Status TarWriter::ScopedFileWriter::Write(const void* data, size_t len) {
+  PERFETTO_CHECK(writer_);
+  return writer_->WriteToSink(data, len);
```
safayat-google: nits, should we also check how many bytes already written and ensure it's not overflow the `size_`?
Finally we can also validate it in the Finalize if the total byte written matches with the size_.
Also we can add a test coverage to cover case when data < len or total written byte is less than size_.

safayat-google (approved):

PR #6809, opened 2026-07-21
tp: introduce the generic profiler_sample table

LalitMaganti:
Callstack samples were spread across four parallel tables (perf_sample,
cpu_profile_stack_sample, instruments_sample and the plugin-owned
stack_sample family), each with its own flavour of the same schema and,
in the StackSample case, a parallel reinvention of counters. This is the
first step of collapsing them onto one model.

Introduce __intrinsic_profiler_sample: one row per sample from any
profiler source, with flat context columns (utid, upid, async context,
ucpu, cpu_mode), an optional callstack, a session and an optional
counter set. The perf mechanisms become the generic ones:

- perf_session generalizes to profiler_session, one row per sampling
  data source instance of any source. Each session records its source
  and the unit of the timebase counter it sampled on, so the per-source
  views stay well formed (perf_session only exposes linux perf
  sessions) and consumers can tell time/cycle/instruction sampling
  apart from other sampling types. StackSample streams allocate one
  session per packet sequence.
- perf_counter_set generalizes to profiler_counter_set, linking each
  sample to the counter rows (timebase and followers) recorded at its
  sample point.

The linux perf importers (traced_perf proto and perf.data) write
profiler_sample directly; the perf_sample and perf_session views keep
their exact schemas, and the perf counter tracks, counter sets, stdlib
functions and the perf_counter plugin keep working unchanged.

The StackSample packet's counter descriptors now materialize as counter
tracks instead of a bespoke descriptor table. A counter track models a
counter instance, so the proto gains an explicit scope on
CounterDescriptor: SCOPE_GLOBAL (a single instance, one track per
sampling stream) and SCOPE_CPU (one instance per cpu, one track per
stream and cpu). Other scopes are added on demand via feature requests.
Per-sample values are pushed as counter rows and linked via the counter
set, exactly like perf.

A new stack_sample view exposes every sample which captured a
callstack, across all sources; per-source views (perf_sample today, the
cpu_profile and instruments tables in follow-ups) select their slice of
the same table. Samples which did not capture a callstack (e.g.
counter-only perf samples) stay visible through the per-source views.


src/trace_processor/plugins/perf_counter/perf_counter.cc
```diff
@@ -41,9 +41,9 @@ namespace perfetto::trace_processor {
 class PerfCounterExtractor {
  public:
   explicit PerfCounterExtractor(
-      const tables::PerfCounterSetTable& perf_counter_set_table)
+      const tables::ProfilerCounterSetTable& perf_counter_set_table)
```
sashwinbalaji: nit: rename variable to be profile_counter_set_table

src/trace_processor/plugins/storage_tables/storage_tables.cc
```diff
@@ -126,14 +126,16 @@ class StorageTablesPlugin : public Plugin<StorageTablesPlugin> {
...
-    AddDataframe(out, s->mutable_perf_session_table());
+    AddDataframe(out, s->mutable_profiler_session_table());
     AddDataframe(out, s->mutable_process_memory_snapshot_table());
     AddDataframe(out, s->mutable_profiler_smaps_table());
     AddDataframe(out, s->mutable_protolog_table());
     AddDataframe(out, s->mutable_winscope_trace_rect_table());
     AddDataframe(out, s->mutable_winscope_rect_table());
     AddDataframe(out, s->mutable_winscope_fill_region_table());
     AddDataframe(out, s->mutable_winscope_transform_table());
+    AddDataframe(out, s->mutable_profiler_sample_table());
```
sashwinbalaji: nit: move it up along wioth mutable_profiler_session_table

src/trace_processor/importers/proto/deobfuscation_tracker.cc
```diff
@@ -479,11 +479,15 @@ void DeobfuscationTracker::GuessPackages(
...
+      context_->storage->profiler_sample_table();
+  for (auto sample = profiler_sample_table.IterateRows(); sample; ++sample) {
+    if (!sample.utid().has_value() || !sample.callsite_id().has_value()) {
+      continue;
+    }
     auto thread = context_->storage
-                      ->thread_table()[tables::ThreadTable::Id(sample.utid())];
-    if (!thread.upid().has_value() || !sample.callsite_id().has_value()) {
+                      ->thread_table()[tables::ThreadTable::Id(*sample.utid())];
+    if (!thread.upid().has_value()) {
```
sashwinbalaji: upid in sample itself can be used and this can be removed

src/trace_processor/importers/perf/record_parser.cc
```diff
@@ -188,24 +191,25 @@ base::Status RecordParser::InternSample(Sample sample) {
...
-      {sample.trace_ts, utid, sample.cpu,
-       context_->storage->InternString(
-           ProfilePacketUtils::StringifyCpuMode(sample.cpu_mode)),
-       callsite_id, std::nullopt, session_id, counter_set_id});
+  row.cpu_mode =
+      sample.cpu_mode ==
+              protos::pbzero::perfetto_pbzero_enum_Profiling::MODE_UNKNOWN
+          ? context_->storage->InternString("")
+          : context_->storage->InternString(
+                ProfilePacketUtils::StringifyCpuMode(sample.cpu_mode));
```
sashwinbalaji: create StringId once in constructor ?

src/trace_processor/importers/proto/profile_module.cc
```diff
@@ -365,10 +352,19 @@ void ProfileModule::ParsePerfSample(
...
-      ts, utid, sample.cpu(), cpu_mode_id, cs_id, unwind_error_id,
-      sampling_stream.perf_session_id, counter_set_id);
-  context_->storage->mutable_perf_sample_table()->Insert(sample_row);
+
+  tables::ProfilerSampleTable::Row row;
+  row.ts = ts;
+  row.source = storage->InternString("linux.perf");
+  row.utid = utid;
+  row.upid = upid;
+  row.ucpu = context_->cpu_tracker->GetOrCreateCpu(sample.cpu()).value;
```
sashwinbalaji: add a check sample cpu is present ?

src/trace_processor/perfetto_sql/stdlib/linux/perf/counters.sql
```diff
@@ -69,8 +69,10 @@ SELECT
   pcs.counter_id,
   c.track_id,
   c.value AS counter_value
-FROM __intrinsic_perf_sample AS ps
-JOIN __intrinsic_perf_counter_set AS pcs
-  ON ps.counter_set_id = pcs.perf_counter_set_id
+FROM perf_sample AS ps
+JOIN __intrinsic_profiler_sample AS psi
```
sashwinbalaji: since perf_sample is created from __intrinsic_profiler_sample (with filter) can we just use that

sashwinbalaji (approved):

PR #6811, opened 2026-07-21
ui: download display-video frames and regions

zezeozue:
Clicking a video frame's download button (right of the playback controls in the details panel) saves that frame as a PNG. Selecting a time range over a video track adds a 'Video' tab with a 'Download region' button that saves that part of the video as an .mp4. The region is trimmed to exactly the selection: it is muxed from the enclosing key frame (so it decodes) but plays from where the selection starts. Muxing copies the coded frames with no re-encode (both H.264 and HEVC); the one exception is a selection starting partway through a group of pictures, where dropping the lead-in frames requires re-encoding from the start point via mediabunny's converter. A 'Download all display videos' command saves every stream's full video. Documented in the Screen Recording page.


LalitMaganti:
My 2c: I'm really not a fan of how much code this is adding nore the extra dependency for waht I think is a very niche feature (the area selection + download of that region). The png frame download I can get behind but the other seems a bit overkill to me.

zezeozue:
> My 2c: I'm really not a fan of how much code this is adding nore the extra dependency for waht I think is a very niche feature (the area selection + download of that region). The png frame download I can get behind but the other seems a bit overkill to me.

Yea, I'm on the fence too. The motivation is for quickly showing side by side comparisons (without having to use the trace_video_conv) script. If we think this is not worth the cost. Happy to close

stevegolton:
I'm with @LalitMaganti on this one, mediabunny expands the bundle by about a meg before compression, and it's anther dependency to manage.

If you're on the fence about the feature anyway, maybe let's just drop it and keep the screenshot download?

zezeozue:
Closing this off. Will repost a png only pr

PR #6815, opened 2026-07-21
docs: fix perfetto.dev asset compression

### LINKED ISSUE #6739: perfetto.dev serves all assets uncompressed
The App Engine proxy copies the stored object's `content_encoding` onto every response. Since `download_as_bytes()` always returns decompressed bytes, this stamps a bogus `Content-Encoding` (either `gzip` or the literal `None`), which makes the App Engine frontend skip its own gzip. 

Result: HTML, JS, CSS, and the search index all ship uncompressed (~2.5-3x larger than they should be).

For the docs search we worked around this in https://github.com/google/perfetto/pull/6702 by shipping the index pre-gzipped and decompressing it in the browser. That workaround should be removed once the proxy is fixed.

Better fix: align the docs-site serving with how the UI server  already handles compression.

Yong-yuan-X:
Fixes #6739.

## Problem

The perfetto.dev App Engine proxy downloads GCS objects in decompressed
form, but copies the stored `Content-Encoding` onto every response. This
makes the response body disagree with its headers and prevents assets from
being served with the expected compression.

The docs search worked around this in #6702 by generating a pre-gzipped
`search_index.json.gz` and decompressing it explicitly in the browser.

## Changes

- Negotiate the stored content encoding against the request's
  `Accept-Encoding` header.
- Use `raw_download=True` when the client accepts the stored encoding, so
  the response body remains compressed.
- Only set `Content-Encoding` when returning the corresponding compressed
  representation.
- Add `Vary: Accept-Encoding` to prevent caches from mixing compressed and
  identity responses.
- Restore the search index as `search_index.json`.
- Remove build-time gzip and browser-side `DecompressionStream` handling.
- Compress JSON assets through the existing gsutil deployment pipeline and
  serve them as `application/json`.
- Keep the precomputed search index and loading improvements introduced by
  #6702.

## Testing

- Exercised the App Engine handler with Flask's test client and mocked GCS
  blobs for HTML, JavaScript, CSS, and the search index.
- Verified gzip responses contain the raw gzip body with matching
  `Content-Encoding` and `Content-Length`.
- Verified identity responses contain the decompressed body without a
  `Content-Encoding` header.
- Verified objects without stored compression never produce
  `Content-Encoding: None`.
- Ran Python and JavaScript syntax checks.
- Verified JSON MIME detection and `git diff --check`.

The full docs build is covered by `repo-checks`.

### DIFF
--- a/infra/perfetto.dev/BUILD.gn
+++ b/infra/perfetto.dev/BUILD.gn
@@ -399,11 +399,11 @@ group("all_mdfiles") {
   deps = mdtargets
 }
 
-# Builds assets/search_index.json.gz, the client-side full-text search index
+# Builds assets/search_index.json, the client-side full-text search index
 # consumed by setupSearch() in src/assets/script.js. Hand-written docs are
 # indexed in full; the huge auto-generated reference pages (SQL tables, stdlib,
 # protos) are indexed by title + headings only to keep the shipped index small.
-gen_search_index_out = "${perfetto_website_out_dir}/assets/search_index.json.gz"
+gen_search_index_out = "${perfetto_website_out_dir}/assets/search_index.json"
 
 nodejs_script("gen_search_index") {
   script = "src/gen_search_index.js"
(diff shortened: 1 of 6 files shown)


LalitMaganti:
I didn't want anyone external to work on this issue because it's quite subtle to get AppEngine working correctly and we need to test this ourselves on our real instance :)

Will close this out, please ask on issues in the future before sending PRs.

PR #6818, opened 2026-07-21
tp: parse start_seq_id for Android process lifecycle events

parthsane:
Add optional int64 start_seq_id to AndroidProcessStartEvent, AndroidProcessDiedEvent, and AndroidBinderDiedEvent messages in frameworks_base_track_event.proto to match upstream Android frameworks/base.

Expose start_seq_id in __intrinsic_android_track_event_process table and populate it when handling process start track events in the android_framework_track_event plugin. Also add missing <cstdint> include in src/traceconv/utils.h for GCC builds.


src/traceconv/utils.h
```diff
@@ -19,6 +19,7 @@
 
 #include <stdio.h>
 
+#include <cstdint>
```
sashwinbalaji: nit: change seems unrelated

sashwinbalaji (approved):

PR #6819, opened 2026-07-21
tp: Import V8 CPU profiles as StackSample streams

### LINKED ISSUE #5673: Migrating V8 cpu profiler from legacy phase 'P' to modern TrackEvent
## Context

V8 is migrating to modern events in https://issues.chromium.org/issues/498378089. The CPU profiler emits `Profile` and `ProfileChunk` events under the `disabled-by-default-v8.cpu_profiler` category, which currently use the legacy `TRACE_EVENT_SAMPLE_WITH_ID1` macro (phase `'P'`).

The modern [`TrackEvent.Type`](https://github.com/google/perfetto/blob/95454ef4fe980334e44eb723090c54c6d17c0cb2/protos/perfetto/trace/track_event/track_event.proto#L130-L162) has no `TYPE_SAMPLE`.

Current plan is to use `TRACE_EVENT_INSTANT` https://chromium-review.googlesource.com/c/v8/v8/+/7787230/5/src/profiler/profile-generator.cc since a sample event is structurally similar to instant event with a differentiating phase byte.

```
// Legacy (phase 'P'):
TrackEvent {
  legacy_event {
    phase: 'P'
    unscoped_id: 42
  }
  debug_annotation { ... } ← JSON payload
}

// Modern (TYPE_INSTANT):
TrackEvent {
  type: TYPE_INSTANT
  name: "Profile"
  debug_annotation {       ← JSON payload (same format)
    name: "data"
    string_value: "{...\"id\":42...}"  ← id is HERE (or in correlation_id)
  }
}
```

### Tokenizer
The tokenizer special-cases V8 CPU samples at https://github.com/google/perfetto/blob/95454ef4fe980334e44eb723090c54c6d17c0cb2/src/trace_processor/importers/proto/track_event_tokenizer.cc#L465-L476 based on the phase type. This extracts the embedded V8 CPU profile data and feeds it to `V8Tracker`, which populates the `cpu_profile_stack_sample` table.

### Event importer
Phase `'P'` has no dedicated case in the phase switch https://github.com/google/perfetto/blob/95454ef4fe980334e44eb723090c54c6d17c0cb2/src/trace_processor/importers/proto/track_event_event_importer.h#L209-L244 — it defaults to storing the event in `chrome_raw_table` as an opaque blob.

## The problem

Using `TYPE_INSTANT` there's no way to distinguish a "regular" instant event from a CPU profile sample event at the proto level. The tokenizer must now match by **event name** (`"Profile"`, `"ProfileChunk"`) rather than by type. Similarly, DevTools must explicitly exclude these events from generic instant event handling, https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/7800921.

1. Should `TYPE_INSTANT` be the recommended modern replacement for phase `'P'`?

The current approach works but requires name-based detection in the tokenizer and explicit exclusions in consumers. As part of this migration, I would need to special case tokenizer like
```cpp
if (event.type() == TYPE_INSTANT) {
    if (name == "Profile" || name == "ProfileChunk") {
        TokenizeModernSampleEvent(...);
    }
}
```

Note: `TokenizeModernSampleEvent` will be using `TrackEvent::Decoder` instead of `LegacyEvent::Decoder` rest of the mechanics from the current tokenizer will remain the same.

2. Would a dedicated `TYPE_SAMPLE` be considered?

    Adding `TYPE_SAMPLE` to the `TrackEvent.Type` enum would:
- Allow the tokenizer to continue to detect CPU profile events by type alone
- Eliminate the need for downstream consumers (DevTools) to special-case instant events by name

cc @camillobruni

deepak1556:
Doc https://docs.google.com/document/d/1d6qo9DM_09MpgG4BzSNpZ6PcYUJredSannjUwItbKnc/edit?usp=sharing

V8 CPU profile samples flow through StreamingProfilePacket, Frame, and Callstack instead of legacy phase-'P' trace events.

V8-specific attributes use proto2 extensions:
- Frame: tier, column number, deopt-reason IID, script ID, and is-inlined.
- StreamingProfilePacket: per-sample kind, leaf line/column, and owning CPU-profile session ID.

V8CpuProfileSession carries start/end metadata: session ID, source, and wall/thread-time bounds.

Trace processor adds a V8 importer, sidecar tables for V8 frame, sample, session metadata, and internal v8.cpu_profile SQL views. The JSON exporter reconstructs DevTools compatible Profile and ProfileChunk events from those views.

Fixes https://github.com/google/perfetto/issues/5673
Bug: chromium:498378089

cc @camillobruni


deepak1556:
V8 side changes that will consume this https://chromium-review.googlesource.com/c/v8/v8/+/7805673

LalitMaganti:
Things have moved forward since we last discussed this issue. We now have https://github.com/google/perfetto/discussions/6027 which are vendor-neutral APIs designed for *exactly* the problem of "I have arbitrary CPU profiling data I'm looking into Perfetto".

We should rediscuss if StreamingProfilePacket is still the right thing to use. I know Chromium is using this for its callstack sampling but now with StackSample, I'd live if we could --(number of callstack like data we need to parse in Perfetto).

At the very least, we should really move StreamingProfilePacket to be entirely living in the Chromium side and simply roll into Perfetto to make it clearer this is a Chromium-only API.

Separately, https://github.com/google/perfetto/pull/6809 is about land which will signifcantly conflict with this CL so you might want to wait on this.

deepak1556:
Thanks for the context. After reading the RFC, I have a few questions about how `StackSample` could support V8's current CPU profiler as an alternative to `StreamingProfilePacket`.

**1. Logical V8 profile identity and lifecycle**

`TaskContext` and `ExecutionContext` identify what was sampled: the observed V8 thread/task and its execution state. They do not identify which logical profile collected the sample.

V8 permits multiple concurrent CPU profiles on the same isolate thread. Those profiles can overlap, observe identical task/execution contexts, and must still export as independent DevTools profiles. The current implementation carries a logical V8 profile ID per sample to avoid inferring membership from `(tid, timestamp range)`.

`profiler_session` appears to represent the sampling stream/data-source instance. V8 additionally has user-visible logical profile requests layered on top of that stream.

Each logical V8 profile needs:

* a stable profile ID
* source (`Inspector`, `SelfProfiling`, `Internal`)
* start/end wall-clock time
* optional start/end thread CPU time
* an exact association from every sample to its owning logical profile

I am considering the following Trace Processor representation:

```
profiler_session
  # Generic sampling stream / data-source instance.

profiler_sample
  profiler_session_id
  callsite_id
  ...

v8_cpu_profile_session
  id
  v8_profile_id
  source
  start/end metadata

v8_profiler_sample
  profiler_sample_id
  v8_cpu_profile_session_id
  sample_kind
  leaf_line
  leaf_column
```

In this model, the V8 tables are sidecars: generic samples remain in the common profiling model, while the V8 rows express the logical profile grouping and V8 only metadata needed for the existing DevTools CPU-profile JSON export.

Is this the expected approach for a self-emitting profiler with several overlapping logical profiles on one sampling stream ? Or is there an intended generic way to represent this additional level of profile identity and lifecycle ?

**2. Per-sample metadata, source position, and timebase**

V8 needs several attributes whose values vary per sample:

* sample kind: normal, `(program)`, GC, or idle
* leaf source line and column, used for DevTools line-level profiling
* wall-clock delta from the preceding sample, used to reconstruct legacy `timeDeltas`

The leaf source position cannot live on `Frame`: one interned frame can be sampled at multiple positions. Likewise, sample kind cannot be inferred reliably from frame names or task context.

For the imported representation, the `v8_profiler_sample` sidecar above could be keyed by the generic `profiler_sample` row and hold the V8 only fields.

For the timebase, should V8 declare a wall-time / nanoseconds `CounterDescriptor` and use `StackSample.primary_weight` for the inter-sample delta ?

What is the intended wire mechanism for the V8 only attributes? Would `StackSample` gain an extension range with V8 specific fields, or is there another annotation mechanism ?

**3. AoS versus batching cost**

The RFC prefers AoS: one self-contained `StackSample` per packet. `StreamingProfilePacket` is SoA: one packet batches parallel arrays of
callstacks, deltas, and metadata.

V8 currently samples with a 100 us configured interval and batches pending samples. A direct `StackSample` migration may therefore create many more trace packets and writer commits.

Is one `StackSample` per sample the intended initial implementation, with overhead measured afterward ? Or is there a planned canonical batching representation for high-rate `StackSample` producers?

**4. Importer and SQL layering after #6809**

Assuming `StackSample` is the right transport, would this be the intended Trace Processor layering ?

```
StackSample
  -> generic profiler_sample / profiler_session
  -> V8 frame, sample, and logical-profile sidecars
  -> internal V8 SQL views
  -> existing DevTools CPU-profile JSON exporter
```

LalitMaganti:
> profiler_session appears to represent the sampling stream/data-source instance. V8 additionally has user-visible logical profile requests layered on top of that stream.

But why make your life complicated like this? Why not just make each sampling stream at the V8 level == profiler_session at the trace processor level?

It's much simpler to split them into separate sampling streams and then have some "grouping id" if you want to preserve the fact that they were all started by the same logical "source".

> Is this the expected approach for a self-emitting profiler with several overlapping logical profiles on one sampling stream ? Or is there an intended generic way to represent this additional level of profile identity and lifecycle ?

Sessions *are* this mechanism and I think your life will be easier if you use it as such.

> For the timebase, should V8 declare a wall-time / nanoseconds CounterDescriptor and use StackSample.primary_weight for the inter-sample delta ?

Probably yeah. I think that would make sense. Though we should probably prototype and see if this does what you want.

> What is the intended wire mechanism for the V8 only attributes? Would StackSample gain an extension range with V8 specific fields, or is there another annotation mechanism ?

Yes we should probably allow for extensions. It would actaully be nice if we wrote something in TP where extensions were generically parsed as arguments in the args table (similar to what happens in track event) so anyone in the future doesn't need to add parsing code at all.

> Is one StackSample per sample the intended initial implementation, with overhead measured afterward ? Or is there a planned canonical batching representation for high-rate StackSample producers?

Unless there's evidence that this actually mattters, I doubt that you actually *need* SOAs. If you have proof you do, I'm happy to create a "bundled" version of StackSample which does what you need and bless that for people to use generically.

> Assuming StackSample is the right transport, would this be the intended Trace Processor layering ?

Well the devtools exporter cannot use SQL so it cannot use the views you write for SQL querying. Instead it will look at the generic tables + V8 Tables. It's not really much different from what you've implemented in this CL really.

deepak1556:
@LalitMaganti PTAL, thanks!

include/perfetto/tracing/data_source.h
```diff
@@ -336,6 +336,19 @@ class DataSource : public DataSourceBase {
       return tls_inst_->trace_writer->NewTracePacket();
     }
 
+    // Creates a new writer for this data source instance. The returned writer
```
LalitMaganti, line 339: This is not OK to add AFAIK. @sashwinbalaji is there a good way for a datasource to create a tracewriter? Not sure it was ever exposed and if not, why not. You might need to ask Primiano about this.

src/trace_processor/importers/common/v8_cpu_profile_tracker.cc
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
```
LalitMaganti, line 2: THis header doesn't seem to be correct. Please follow elsewhere in the project.

python/generators/diff_tests/models.py
```diff
@@ -54,6 +54,7 @@ class Config:
   winscope_extensions: str
   gpu_extensions: str
   gpu_interned_data_extensions: str
+  v8_profile_extensions: str
```
LalitMaganti, line 57: I don't know why this is called "v8 profile extensions"...

src/trace_processor/importers/proto/stack_profile_sequence_state.cc
```diff
@@ -373,5 +378,18 @@ StackProfileSequenceState::LookupInternedSourcePath(
   return *str;
 }
 
+void StackProfileSequenceState::AttachRuntimeFrameExtensions(
```
LalitMaganti, line 381: Sorry but this is a huge violation of layering. Why is core proto importing suddently having dependenices on v8? This seems not designed properly.

src/trace_processor/importers/proto/v8_cpu_profile_module.cc
```diff
@@ -0,0 +1,200 @@
...
+    return;
+
+  bool any = false;
+  std::optional<int32_t> tier;
+  std::optional<bool> is_inlined;
+  std::optional<uint32_t> column;
+  std::optional<uint64_t> deopt_reason_iid;
+  std::optional<int32_t> script_id;
+
+  protozero::ProtoDecoder dec(frame_bytes, frame_size);
```
LalitMaganti, line 149: Why exactly are we doing this rather than using the deocder for V8?

src/trace_processor/perfetto_sql/stdlib/sched/thread_executing_span_with_slice.sql
```diff
@@ -334,9 +334,7 @@ WITH
       spans.root_utid
     FROM relevant_spans AS spans
     LEFT JOIN slice AS sl
-      ON sl.id = spans.self_slice_id
```
LalitMaganti, line 337: Not sure why this code is changing?

src/trace_processor/perfetto_sql/stdlib/viz/summary/processes.sql
```diff
@@ -41,7 +41,8 @@ SELECT
   sum(running_count) AS running_count,
   sum(slice_count) AS slice_count,
   sum(perf_sample_count) AS perf_sample_count,
-  sum(instruments_sample_count) AS instruments_sample_count
+  sum(instruments_sample_count) AS instruments_sample_count,
```
LalitMaganti, line 44: Can you please hold off on changes here, I'm working through this in a separate stack. Please focus on v8 changes only.

src/trace_processor/plugins/stack_sample_importer/module.cc
```diff
@@ -118,6 +119,27 @@ const char* StringifyProfileUnit(protos::pbzero::StackSample::Unit unit) {
   return "";
 }
 
+std::optional<int32_t> GetV8SampleKind(protozero::ConstBytes blob) {
+  protozero::ProtoDecoder decoder(blob.data, blob.size);
```
LalitMaganti, line 123: As before this is a huge layering violation.

LalitMaganti (commented):
Sorry but this PR is quite far being landable. There several layering violations which are  present where core code is depending on V8. This needs to be properly thought through and factored out.

I don't mind if you are totally doing your own thing decoupled from us what you do. But if you're touching core code, you need to do proper layering not just add improper dependencies because it's convinient.

src/trace_processor/importers/proto/stack_profile_sequence_state.cc
deepak1556: Would the proper way to achieve this be something similar to `TrackEventExtensionParser` have a generic frame extension parser that V8 layer can then specialize upon ?

```
void StackProfileSequenceState::DispatchFrameExtensions(
    PacketSequenceStateGeneration* state,
    FrameId frame_id,
    protozero::ConstBytes frame_bytes) {
  protozero::ProtoDecoder decoder(frame_bytes.data, frame_bytes.size);

  for (auto field = decoder.ReadField(); field.valid(); field = decoder.ReadField()) {
    auto* parser = extension_context_->parsers_by_field.Find(field.id());
    if (!parser) {
      continue;
    }

    parser->OnFrameExtension(
        FrameExtensionField(field), frame_id, state);
  }
}
```

python/generators/diff_tests/models.py
deepak1556: is `v8_cpu_profile_extensions` better ?

src/trace_processor/importers/proto/v8_cpu_profile_module.cc
deepak1556: The current `V8FrameExtensions` doesn't generate a specialized decoder, is the ask to make the frame extension something like so a V8 decoder can be used ?

```
message V8Frame {
  optional V8Tier tier = 1;
  optional bool is_inlined = 2;
  optional uint32 column_number = 3;
  optional uint64 deopt_reason_iid = 4;
  optional int32 script_id = 5;
}

message V8FrameExtensions {
  extend Frame {
    optional V8Frame v8 = 1000;
  }
}
```

src/trace_processor/perfetto_sql/stdlib/sched/thread_executing_span_with_slice.sql
deepak1556: Unsure, I just ran `tools/format-sources`. Will undo it.

src/trace_processor/perfetto_sql/stdlib/viz/summary/processes.sql
deepak1556: Understood will revert them, I was testing this change E2E and saw the samples not showing up in the UI, even if it did was grouped to default workspace without this change.

deepak1556:
Apologies for the layering violations introduced, it was not the intention. I just went onto quickly prototype the new transport design E2E. Will address all the review feedback.

LalitMaganti:
> I just went onto quickly prototype the new transport design E2E. 

I wouldn't have a problem with those if this was marked as a prototype, I didn't see any indication it was. It seemed to me this is code you wanted to land :) 

Obviously I would treat prototype code in a very different way!

src/trace_processor/importers/proto/v8_cpu_profile_module.cc
LalitMaganti: It should generate if you add the GN dependency on the V8 protos.

src/trace_processor/importers/proto/stack_profile_sequence_state.cc
LalitMaganti: Yes something along those lines. Specifically all of this code should really be going in a plugin as well so ideally we plumb it through the plguin API just kike track event.

Though I don't think it should go in StackProfileSequenceState - this code should be inlined directly into the parser. This has njothing to do with the sequence state as such.

python/generators/diff_tests/models.py
LalitMaganti: It's unclear to me a) why this has anythign to do with v8 b) why you need this at all...

include/perfetto/tracing/data_source.h
deepak1556: V8 needs one independent packet sequence per logical CPU profile, so trace processor maps each stream directly to one generic profiler_session. The active v8.cpu_profiler data source instance currently exposes only its shared TLS writer via `TraceContext::NewTracePacket`. Is there a supported SDK API for a data source to create an additional trace writer bound to its active instance ?

LalitMaganti: If we cannot create a trace writer, we could always add a field in the proto for adding the logical sequence id, separate from the packet sequene. We assume both today but there's no reason we could not add something separate for you.

python/generators/diff_tests/models.py
deepak1556: sorry this is a stale artifact from previous iteration of sql diff tests, will remove it.

include/perfetto/tracing/data_source.h
deepak1556: That would actually make lifecycle management of the writer simpler on the V8 side. I will try this approach.

PR #6822, opened 2026-07-21
tp: migrate the cpu_profile sample sources onto profiler_sample

LalitMaganti:
Chrome streaming profiles, legacy V8 cpu profiles, gecko profiles,
simpleperf proto traces and perf script text all wrote their samples
into the cpu_profile_stack_sample table. Write them into the generic
profiler_sample table instead, tagged with their source and with process
attribution where the importer knows the pid.

cpu_profile_stack_sample keeps its exact schema as a view over
profiler_sample; the chrome-only process_priority column moves to a slim
extras table keyed by sample, populated only for samples with a
non-default priority.

### DIFF
--- a/src/trace_processor/importers/common/legacy_v8_cpu_profile_tracker.cc
+++ b/src/trace_processor/importers/common/legacy_v8_cpu_profile_tracker.cc
@@ -30,6 +30,7 @@
 #include "src/trace_processor/importers/common/mapping_tracker.h"
 #include "src/trace_processor/importers/common/parser_types.h"
 #include "src/trace_processor/importers/common/process_tracker.h"
+#include "src/trace_processor/importers/common/profiler_sample_tracker.h"
 #include "src/trace_processor/importers/common/stack_profile_tracker.h"
 #include "src/trace_processor/importers/common/stats_tracker.h"
 #include "src/trace_processor/storage/stats.h"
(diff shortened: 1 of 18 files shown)


src/trace_processor/importers/common/legacy_v8_cpu_profile_tracker.cc
```diff
@@ -184,8 +185,14 @@ base::Status LegacyV8CpuProfileTracker::AddSample(int64_t ts,
...
   }
   UniqueTid utid = context_->process_tracker->UpdateThread(tid, pid);
-  auto* samples = context_->storage->mutable_cpu_profile_stack_sample_table();
-  samples->Insert({ts, *id, utid, 0});
+  tables::ProfilerSampleTable::Row row;
+  row.ts = ts;
+  row.source = context_->storage->InternString("legacy_v8");
+  row.utid = utid;
+  row.upid = context_->process_tracker->GetOrCreateProcess(pid);
+  row.cpu_mode = context_->storage->InternString("");
```
sashwinbalaji: curious why empty string instead of letting it be null (was plannign to ask in previous pr but forgot..)

src/trace_processor/importers/proto/profile_module.cc
```diff
@@ -224,10 +224,18 @@ void ProfileModule::ParseStreamingProfileSample(
     return;
   }
 
-  tables::CpuProfileStackSampleTable::Row sample_row{ts, *opt_cs_id, utid,
-                                                     event.process_priority};
-  context_->storage->mutable_cpu_profile_stack_sample_table()->Insert(
-      sample_row);
+  tables::ProfilerSampleTable::Row row;
+  row.ts = ts;
+  row.source = context_->storage->InternString("chrome");
```
sashwinbalaji: intern once (other places also) ?

sashwinbalaji (approved):

PR #6827, opened 2026-07-22
tp: add Arrow support for dataframes

LalitMaganti:
Provide an Arrow serializer and deserializer for Dataframes using a
standard Arrow file containing one record batch. Primitive columns map
to Arrow numeric arrays, strings map to Utf8, nullable columns carry
validity bitmaps, and sparse storage uses Arrow's logical row layout.

The serializer computes metadata and variable-width sizes before
streaming column contents with bounded scratch memory. The deserializer
validates file framing, record batch metadata, and buffer extents before
rebuilding column storage, null state, implicit Id columns, and strings
in the target StringPool.


src/trace_processor/core/dataframe/arrow_serializer.h
```diff
@@ -0,0 +1,86 @@
...
+  // Computes Arrow buffer metadata and the exact output size for a finalized
+  // dataframe. Arrow places buffer lengths before buffer contents, so
+  // variable-width strings must be measured here before Write() can stream the
+  // file. This retains only metadata; it does not materialize column contents.
+  base::StatusOr<size_t> Prepare(const Dataframe&, const StringPool&);
+
+  // Streams the dataframe prepared by the preceding Prepare() call using
+  // bounded scratch memory. Column contents are not materialized as a complete
+  // Arrow body. The dataframe and string pool must not change between calls.
+  base::Status Write(const Dataframe&, const StringPool&, const WriteFn&);
```
safayat-google, line 57: I think it's great desing to provide the WriteFn to avoid copying a big memory.

But now we always need to call two functions, and provide the writeFn, assuming that in most cases the user just wants to write it to a string or file, Can we still provide another api similar to the `deserializer` to do all of this in function call? It can be convenient to the user of the api.

i.e.
`base::StatusOr<std::vector<uint8_t>> SerializeToArrow(const Dataframe& df, 
                                                       const StringPool& pool) {`

safayat-google (approved):

src/trace_processor/core/dataframe/arrow_serializer.h
LalitMaganti: Unfortunately Prepare/Write are needed separately because we need to compute the size of the file *before* we generate it so we can write it into the TAR header.

You're right in theory something like what you describe would be nice but no one would actually use that function in practice :) I don't want to land code folks won't use!

safayat-google: Acknowledged.

PR #6829, opened 2026-07-22
proto_merger: add support for proto extensions

BogdanLionte:


dragosp95 (approved):

src/tools/proto_merger/proto_file.cc
```diff
@@ -89,32 +83,75 @@ std::string SimpleFieldTypeFromDescriptor(
...
   field_type += ",";
-  field_type += FieldTypeFromDescriptor(parent, *desc.message_type()->field(1),
-                                        packageless_type);
+  field_type += FieldTypeInScope(
+      scope_full_name, *desc.message_type()->field(1), packageless_type);
   field_type += ">";
   return field_type;
 }
 
+std::string MinimizeExtendee(const std::string& target_full_name,
```
sashwinbalaji: don't think this wrapper one line function (used in one place) is needed and can be inlined

src/tools/proto_merger/proto_file_serializer.cc
```diff
@@ -16,7 +16,7 @@
 
 #include "src/tools/proto_merger/proto_file_serializer.h"
 
-#include "perfetto/ext/base/string_utils.h"
+#include <google/protobuf/descriptor.h>
```
sashwinbalaji: why this change ??

src/tools/proto_merger/proto_merger.cc
```diff
@@ -379,14 +379,18 @@ base::Status MergeFields(const std::vector<ProtoFile::Field>& input,
 
 // We call both of these just "Merge" so that |MergeRecursive| below can
 // reference them with the same name.
-base::Status Merge(const ProtoFile::Oneof& input,
+base::Status Merge(const Allowlist& root_allowlist,
```
sashwinbalaji: why are we plumbing root_allowlist and package ? I am not seeing them being used anywhere.

src/tools/proto_merger/proto_merger.cc
```diff
@@ -483,21 +493,27 @@ base::Status Merge(const ProtoFile::Message& input,
...
-  status = MergeRecursive(input.oneofs, upstream.oneofs, allowlist.oneofs,
-                          known_enums, allowlisted_options, out.oneofs);
+  status = MergeRecursive(root_allowlist, package, input.oneofs,
+                          upstream.oneofs, allowlist.oneofs, known_enums,
+                          allowlisted_options, out.oneofs);
   if (!status.ok())
     return status;
 
   // Finish by merging the list of fields.
-  return MergeFields(input.fields, upstream.fields, allowlist.fields,
```
sashwinbalaji, line 526: can we keep this why get status and return the same again ?

src/tools/proto_merger/proto_file_serializer_unittest.cc
```diff
@@ -640,6 +640,153 @@ TEST(ProtoFileSerializerTest, PassthroughDeepRecursion) {
...
+  };
+
+  base::TempDir temp_dir = base::TempDir::Create();
+  std::string input_path = temp_dir.path() + "/input.proto";
+  std::string upstream_base_path = temp_dir.path() + "/upstream_base.proto";
+  std::string upstream_ext_path = temp_dir.path() + "/upstream_ext.proto";
+
+  ScopedUnlink unlink_input{input_path};
+  ScopedUnlink unlink_upstream_base{upstream_base_path};
+  ScopedUnlink unlink_upstream_ext{upstream_ext_path};
```
sashwinbalaji: nit: this part is repeated >3 times in this file can we move it into a shared func pls

src/tools/proto_merger/proto_merger.cc
```diff
@@ -536,8 +553,136 @@ base::Status MergeProtoFiles(const ProtoFile& input,
...
+      }
+    }
+
+    for (auto& msg : ext_file.messages) {
+      ProcessMessageExtensions(msg, base_file);
+    }
+
+    for (auto& msg : ext_file.messages) {
+      if (!IsMessageEmpty(msg)) {
+        base_file.messages.push_back(std::move(msg));
```
sashwinbalaji: we should check for conflct before adding. also wont this msg's type be in reference to extension file not the base one ?

src/tools/proto_merger/proto_merger.h
```diff
@@ -44,6 +44,9 @@ base::Status MergeProtoFiles(
     ProtoFile& out,
     const std::set<std::string>& allowlisted_options = {});
 
+void InlineExtensions(ProtoFile& base_file,
```
sashwinbalaji: pls add a comment explainign the func similar to above MergeProtoFiles one

 also this destroys extension_files so can we take it by value and std::move at the callsite to be clearer

src/tools/proto_merger/proto_merger.cc
```diff
@@ -536,8 +553,136 @@ base::Status MergeProtoFiles(const ProtoFile& input,
...
+      return fq_type;
+    std::string key = fq_type.substr(4, comma - 4);
+    std::string val = fq_type.substr(comma + 1, fq_type.size() - comma - 2);
+    return "map<" + MinimizeFqType(key, scope) + "," +
+           MinimizeFqType(val, scope) + ">";
+  }
+  if (base::StartsWith(fq_type, ".")) {
+    std::string type_without_dot = fq_type.substr(1);
+    auto opt = MinimizeType(type_without_dot, scope);
+    return opt.value_or(type_without_dot);
```
sashwinbalaji: this is different behavior then what we have existingly in [SimpleFieldTypeFromDescriptor](https://crsrc.org/c/third_party/perfetto/src/tools/proto_merger/proto_file.cc;l=70) which return the name only we should align it.

src/tools/proto_merger/proto_file.cc
```diff
@@ -89,32 +83,75 @@ std::string SimpleFieldTypeFromDescriptor(
...
+std::vector<ProtoFile::Extension> ParseExtensions(
+    const DescriptorType& desc,
+    const std::string& scope_name) {
+  std::map<std::string, ProtoFile::Extension> extensions_map;
+  for (int i = 0; i < desc.extension_count(); ++i) {
+    const auto* ext_desc = desc.extension(i);
+    std::string extendee = MinimizeExtendee(
+        std::string(ext_desc->containing_type()->full_name()), scope_name,
+        std::string(ext_desc->containing_type()->name()));
+    auto& ext = extensions_map[extendee];
```
sashwinbalaji: so the key is MinimizeExtendeed name which can just return the target_name which is possible to collide. should we use fill_name instead ?

src/tools/proto_merger/proto_merger.cc
```diff
@@ -536,8 +553,136 @@ base::Status MergeProtoFiles(const ProtoFile& input,
...
+  }
+  return fq_type;
+}
+
+ProtoFile::Message* FindMessageByFullName(ProtoFile& file,
```
sashwinbalaji: I think existing FindMessageTypeByName already does this. so can the inling happen at descriptor time before protofilefromdescriptor ?

src/tools/proto_merger/proto_merger.cc
```diff
@@ -536,8 +553,136 @@ base::Status MergeProtoFiles(const ProtoFile& input,
...
+  for (auto& ext_file : extension_files) {
+    auto it = ext_file.extensions.begin();
+    while (it != ext_file.extensions.end()) {
+      ProtoFile::Message* target =
+          FindMessageByFullName(base_file, it->full_extendee);
+      if (target) {
+        InlineExtensionIntoMessage(*target, *it);
+        it = ext_file.extensions.erase(it);
+      } else {
+        ++it;
```
sashwinbalaji: right now we are skipping the unreslved ones should we pass it back to report it ?

src/tools/proto_merger/allowlist.cc
```diff
@@ -141,6 +145,15 @@ base::Status AllowlistFromFieldList(
     const auto* current = &desc;
     for (size_t i = 0; i < pieces.size(); ++i) {
       const auto* field = current->FindFieldByName(pieces[i]);
+      if (!field) {
```
sashwinbalaji, line 202: since FindFieldByName won't return extensions this also needs the fallback flow like above in passthrough , maybe need to use FindAllExtensions ?

src/tools/proto_merger/proto_file.h
```diff
@@ -52,42 +52,61 @@ struct ProtoFile {
...
     std::vector<Oneof> oneofs;
     std::vector<Field> fields;
+    std::vector<ExtensionRange> extension_ranges;
+    std::vector<Extension> extensions;
 
     std::vector<Enum> deleted_enums;
     std::vector<Message> deleted_nested_messages;
     std::vector<Oneof> deleted_oneofs;
     std::vector<Field> deleted_fields;
+    std::vector<Extension> deleted_extensions;
```
sashwinbalaji: where are we using these ?

also I think extension_ranges can be dropped ?

src/tools/proto_merger/proto_file.cc
```diff
@@ -275,14 +348,30 @@ ProtoFile ProtoFileFromDescriptor(
...
     file.enums.push_back(EnumFromDescriptor(*desc.enum_type(i)));
   }
   for (int i = 0; i < desc.message_type_count(); ++i) {
     file.messages.push_back(MessageFromDescriptor(*desc.message_type(i)));
   }
+  file.extensions = ParseExtensions(desc, std::string(desc.package()));
   return file;
 }
 
+std::optional<std::string> MinimizeType(const std::string& a,
```
sashwinbalaji: minor nit: place it where it was before. so it's before the callers.

src/tools/proto_merger/proto_file.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file_serializer.cc
BogdanLionte: string_utils.h was unused. I also removed <google/protobuf/descriptor.h> added by mistake.

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Removed

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file_serializer_unittest.cc

src/tools/proto_merger/allowlist.cc
BogdanLionte: It uses FindExtensionByName

src/tools/proto_merger/proto_file.h
BogdanLionte: Done

src/tools/proto_merger/proto_file.cc
BogdanLionte: Done

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Not relevant anymore after moving the inlining.

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file.cc
BogdanLionte: Done

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_merger.h
BogdanLionte: Done

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done. The msg's package will be changed.

src/tools/proto_merger/proto_file.cc
```diff
@@ -29,6 +29,24 @@ namespace perfetto {
...
+}
+
+bool IsMessageEmpty(const ProtoFile::Message& msg) {
+  return msg.fields.empty() && msg.nested_messages.empty() &&
+         msg.enums.empty() && msg.oneofs.empty();
+}
+
+ProtoFile::Field FieldFromDescriptor(
+    const google::protobuf::Descriptor& parent,
+    const google::protobuf::FieldDescriptor& desc);
```
sashwinbalaji: don't think this decl is needed anymore

src/tools/proto_merger/main.cc
```diff
@@ -42,19 +42,24 @@ namespace proto_merger {
...
+  dst->MapPath("", proto_dir_path);
 
-  google::protobuf::compiler::DiskSourceTree dst;
-  dst.MapPath("", proto_dir_path);
+  auto mfe = std::make_unique<protozero::MultiFileErrorCollectorImpl>();
 
   ImportResult result;
-  result.importer.reset(new google::protobuf::compiler::Importer(&dst, &mfe));
+  result.importer.reset(
+      new google::protobuf::compiler::Importer(dst.get(), mfe.get()));
```
sashwinbalaji: nit: can we use either make_unique or reset not both in same func?

src/tools/proto_merger/proto_file.cc
```diff
@@ -190,13 +216,31 @@ Output InitFromDescriptor(const Descriptor& desc) {
...
+    case google::protobuf::FieldDescriptor::TYPE_MESSAGE:
+      return "." + std::string(desc.message_type()->full_name());
+    case google::protobuf::FieldDescriptor::TYPE_ENUM:
+      return "." + std::string(desc.enum_type()->full_name());
+    default:
+      return kTypeToName[desc.type()];
+  }
+}
+
+ProtoFile::Field FieldFromDescriptorHelper(
```
sashwinbalaji: there is just one caller of this can we inline it pls

src/tools/proto_merger/proto_file.cc
```diff
@@ -190,13 +216,31 @@ Output InitFromDescriptor(const Descriptor& desc) {
...
+    const google::protobuf::FieldDescriptor& desc,
+    std::string type,
+    std::string packageless_type) {
   auto field = InitFromDescriptor<ProtoFile::Field>(desc);
   field.is_repeated = desc.is_repeated();
-  field.packageless_type = FieldTypeFromDescriptor(parent, desc, true);
-  field.type = FieldTypeFromDescriptor(parent, desc, false);
+  field.packageless_type = std::move(packageless_type);
+  field.type = std::move(type);
+  field.fq_type = FullyQualifiedFieldType(desc);
```
sashwinbalaji: fq_type and FullyQualifiedFieldType can be removed now ?

src/tools/proto_merger/proto_file.cc
```diff
@@ -67,50 +85,58 @@ std::optional<std::string> MinimizeType(const std::string& a,
...
+template <typename DescriptorType>
+std::string TypeNameInScope(const std::string& base_package,
+                            const std::string& scope_full_name,
+                            const google::protobuf::FieldDescriptor& desc,
+                            const DescriptorType* type,
+                            bool packageless_type) {
+  std::string full_name = std::string(type->full_name());
+  if (type->file() != desc.containing_type()->file()) {
+    full_name = base_package + "." + std::string(type->name());
+  }
```
sashwinbalaji: ohh this will break all cross file refs to nested types within the repo also and dropping nesting for them would be a problem. Pls verify this once

src/tools/proto_merger/proto_merger.cc
```diff
@@ -510,6 +510,7 @@ base::Status MergeProtoFiles(const ProtoFile& input,
   // The preamble is taken directly from upstream. This allows private stuff
   // to be in the preamble without being present in upstream.
   out.preamble = input.preamble;
+  out.package = input.package;
```
sashwinbalaji: what's the purpose of this field it's set once in src/tools/proto_merger/proto_file.cc but not used ? once you remove it also pls remove the newline deletion to redice the number of files changed

src/tools/proto_merger/proto_file.cc
```diff
@@ -265,22 +323,64 @@ ProtoFile::Message MessageFromDescriptor(
...
+      auto en = EnumFromDescriptor(*ext_file_desc->enum_type(i));
+      if (!FindByName(file.enums, en.name)) {
+        file.enums.push_back(std::move(en));
+      }
+    }
+    for (int i = 0; i < ext_file_desc->message_type_count(); ++i) {
+      auto msg = MessageFromDescriptor(*ext_file_desc->message_type(i),
+                                       extension_files);
+      if (!IsMessageEmpty(msg) && !FindByName(file.messages, msg.name)) {
+        file.messages.push_back(std::move(msg));
```
sashwinbalaji, line 382: should we have some error / exit if this condition is not met overall for the !FindByName case ?

sashwinbalaji:
Also just to be safe once the change stabilizes pls verify it by cherry-picking it to g3 code(http://google3/third_party/perfetto/src/tools/proto_merger/) and running the update_perfetto_log_proto.sh script

BogdanLionte:
> Also just to be safe once the change stabilizes pls verify it by cherry-picking it to g3 code(http://google3/third_party/perfetto/src/tools/proto_merger/) and running the update_perfetto_log_proto.sh script

I ran it and it fails because the proto imported sees ChromeCompositorSchedulerState defined in perfetto_trace.proto (the fused file). Then, while parsing the extension, it is forced to parse track_event.proto, which uses that message (as a field type).

The importer then complains because track_event.proto uses a message that the importer now believes belongs to perfetto_trace.proto, but track_event.proto doesn't have an import "perfetto_trace.proto"; statement (it imports  chrome_compositor_scheduler_state.proto for ChromeCompositorSchedulerState instead).

Error third_party/perfetto/protos/perfetto/trace/track_event/track_event.proto 363:11: "perfetto.protos.ChromeCompositorSchedulerState" seems to be defined in "third_party/perfetto/protos/perfetto/trace/perfetto_trace.proto", which is not imported by "third_party/perfetto/protos/perfetto/trace/track_event/track_event.proto".  To use it here, please add the necessary import.

src/tools/proto_merger/proto_file.cc
BogdanLionte: I think maybe we should only if they the messages have different fields, and just ignore them if they are the same. But that would require implementing message equality comparison. Should we do that?

src/tools/proto_merger/proto_merger.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file.cc
BogdanLionte: Done

src/tools/proto_merger/main.cc
BogdanLionte: Done

src/tools/proto_merger/proto_file.cc
BogdanLionte: Removed

PR #6838, opened 2026-07-22
perfetto: keep TrackEvent's inline callstack out of the profiling package

sashwinbalaji:
top-level InlineCallstack in a new profiling/inline_callstack.proto, so both TrackEvent and StackSample could use it. That made track_event.proto import a file outside its own directory for the first time, which broke out-of-tree Android.bp genrules that only depended on the narrow, self-contained track_event proto filegroup (e.g. vendor consumers and frameworks/native/tracing), since Soong's sbox sandboxes only see explicitly listed srcs.

Revert TrackEvent's side: restore its own nested TrackEvent.Callstack message (same field numbers, wire-compatible) instead of importing InlineCallstack. StackSample keeps using the shared InlineCallstack from profiling/inline_callstack.proto, since that's a same-package dependency and was never the problem.

### DIFF
--- a/protos/perfetto/trace/track_event/track_event.proto
+++ b/protos/perfetto/trace/track_event/track_event.proto
@@ -294,7 +294,7 @@ message TrackEvent {
   oneof callstack_field {
     // Inline callstack data. Use this for simplicity when interning is not
     // needed (e.g., for unique callstacks or when trace size is not critical).
-    Callstack callstack = 55;
+    InlineCallstack callstack = 55;
 
     // Reference to interned Callstack (see InternedData.callstacks).
     // This is the efficient option when callstacks are repeated.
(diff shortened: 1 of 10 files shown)


protos/perfetto/trace/track_event/track_event.proto
```diff
@@ -305,6 +303,33 @@ message TrackEvent {
...
 
+  // Inline callstack for TrackEvents when interning is not needed.
+  // This is a simplified version of the profiling Callstack/Frame messages,
+  // designed for cases where trace size is not critical or callstacks are
+  // unique.
+  //
+  // Use this for simple callstacks with function names and source locations.
+  // For binary/library information (mappings, build IDs, relative PCs), use
+  // interned callstacks via callstack_iid instead.
+  message Callstack {
```
LalitMaganti: Please keep calling it Inline callstack

LalitMaganti (approved):

PR #6839, opened 2026-07-22
tp: add streaming table export

LalitMaganti:
Add a bounded-memory Export API for statically registered Trace Processor
tables. ExportOutput is a virtual streaming interface with an optional file
path alternative for formats which need direct random-access output later.
Runtime SQL tables and views are deliberately outside this API.

Support two explicit tar formats. kArrowTar is a stable, forwards-compatible,
export-only archive containing standard Arrow files, including implicit ID
columns and empty static tables for external consumers. kPerfetto is a
version-coupled archive with an internal manifest which can restore tables
into a compatible fresh Trace Processor instance; its representation carries
no cross-version compatibility guarantee.

Expose export through RPC, HTTP, the shell export subcommand, and Python. Keep
the existing SQLite export path unchanged; moving it behind this API waits for
the separate file-I/O abstraction work.


src/trace_processor/rpc/httpd.cc
```diff
@@ -267,6 +267,39 @@ void Httpd::OnHttpRequest(const base::HttpRequest& req) {
...
+        conn.SendResponseBody(buf, len);
+        conn.SendResponseBody("\r\n", 2);
+      }
+      if (!has_more) {
+        conn.SendResponseBody("0\r\n\r\n", 5);
+      }
+    };
+    base::Status status = global_trace_processor_rpc_.Export(*format, on_chunk);
+    if (!status.ok()) {
+      return conn.SendResponseAndClose("500 Internal Server Error",
```
safayat-google: since line 281 already sent the response header `200 ok`, will it produce a malformed HTTP response?

src/trace_processor/plugins/trace_export/perfetto_export_reader.cc
```diff
@@ -0,0 +1,174 @@
...
+  size_t table_idx = static_cast<size_t>(table - state_->tables.begin());
+  if (state_->seen[table_idx]) {
+    return base::ErrStatus("perfetto_export: duplicate archive member '%s'",
+                           path.c_str());
+  }
+  state_->seen[table_idx] = true;
+
+  ASSIGN_OR_RETURN(auto dataframe,
+                   core::dataframe::DeserializeFromArrow(
+                       buffer_, context_->storage->mutable_string_pool(),
```
safayat-google, line 108: nits, if line 114 fails, `context_->storage->mutable_string_pool()` will be polluted. can we fail early for such case or use a temporary string pool?

src/trace_processor/plugins/trace_export/trace_export.cc
```diff
@@ -0,0 +1,158 @@
...
+      std::string manifest =
+          SerializePerfettoExportManifest(tables, base::GetVersionString());
+      RETURN_IF_ERROR(tar.AddFile(
+          kPerfettoManifestFileName,
+          reinterpret_cast<const uint8_t*>(manifest.data()), manifest.size()));
+    }
+    core::dataframe::ArrowSerializer serializer(
+        perfetto ? core::dataframe::ArrowSerializer::IdColumnMode::kOmit
+                 : core::dataframe::ArrowSerializer::IdColumnMode::kInclude);
+    for (const PluginDataframe* table : tables) {
```
safayat-google, line 151: nits, should we stop early if the client have closed the connection midway?

safayat-google (approved):

src/trace_processor/rpc/httpd.cc
LalitMaganti: Nice catch! Done.

src/trace_processor/plugins/trace_export/perfetto_export_reader.cc
LalitMaganti: I tried to fix this but honestly I felt like the fix was more ugly than any benefit it gave. So kept as is.

src/trace_processor/plugins/trace_export/trace_export.cc
LalitMaganti: Yes nice catch! Done!

PR #6843, opened 2026-07-22
ui: fix Heap Dump Explorer deeplink class filter being dropped

zezeozue:
Deeplinking to the Objects tab filtered by class (a Memscope class link) dropped the filter: the class was a query param, but the router parses everything after '?' as global route args and discards unknown keys, so it never reached the explorer.

Encode the class in the path (objects_<class>) like the object and bitmaps views, so the raw link survives the router, and update Memscope's link to match. The class is still consumed as a one-shot grid filter via the existing clearNavParam, which now also drops the param from the URL — it only edited the store before, relying on the router to strip the query, but path-encoded params aren't stripped.


stevegolton (approved):
Nice, thanks!

PR #6846, opened 2026-07-23
ui: Toggle task colors based on niceness

wusamuel6:
Add a hotkey (shift + C) to toggle color of task slices based on their niceness. This distinguishes tasks based on realtime, high-priority CFS tasks (usually foreground tasks), and low priority tasks.

The task slice visualization Perfetto UI offers is invaluable, as the color representation helps to easily spot patterns or trends with tasks being run or scheduled. By color coding different priority tasks, it's significantly easier to visualize how different priority tasks are being scheduled, which is especially helpful when developing new scheduling features, or debugging scheduling issues.

Bug: 537887463
Test: ui/run-unittests --no-build

### DIFF
--- a/ui/src/components/tracks/slice_track.ts
+++ b/ui/src/components/tracks/slice_track.ts
@@ -244,6 +244,11 @@ export interface SliceTrackAttrs<T extends DatasetSchema> {
   colorizer?(row: T): ColorScheme;
 
   /**
+   * Optional function returning a key for invalidating cached slice data frames when track attributes/modes change.
+   */
+  readonly getKey?: () => string;
+
+  /**
    * Override the text displayed on each event (title).
    */
   sliceName?(row: T): string;
@@ -825,6 +830,7 @@ export class SliceTrack<T extends RowSchema> implements TrackRenderer {
         start: bounds.start,
         end: bounds.end,
         resolution: bounds.resolution,
+        key: this.attrs.getKey?.(),
       },
       queryFn: async (signal) => {
         const promise = (async () => {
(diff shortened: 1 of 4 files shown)


wusamuel6:
Please let me know if this is too intrusive of a change into the core code or if it needs more discussion. I thought it would be useful to be able to visualize threads grouped by their niceness/priority, and this was small code change to make.

ui/src/components/tracks/slice_track.ts
```diff
@@ -499,7 +497,7 @@ export class SliceTrack<T extends RowSchema> implements TrackRenderer {
     for (let j = 0; j < count; j++) {
       const slice = slices[j];
       const colorVariant = colorVariants[j];
-      const cs = slice.colorScheme;
+      const cs = this.getColor(slice.row, slice.title);
```
stevegolton: What effect does calling getColor() every frame have on performance? Some if not most tracks do some hashing of the slice name to obtain the colors.

stevegolton (changes requested):
The concept looks good but I worry about the perf - see comment.

ui/src/components/tracks/slice_track.ts
wusamuel6: ACK, agree with you. Updated commit so that it should only recalc color only if there is the hotkey press

ui/src/components/tracks/slice_track.ts
```diff
@@ -243,6 +243,11 @@ export interface SliceTrackAttrs<T extends DatasetSchema> {
    */
   colorizer?(row: T): ColorScheme;
 
+  /**
```
stevegolton, line 246: Yeah this works, though I think we could just rename this to `getKey()`. It doesn't have to have anything to do with colour. It'll work much like the keys in Mithril components.

stevegolton (changes requested):
I like the approach. Looks good % one nit about the name.

stevegolton (approved):

PR #6852, opened 2026-07-23
ui: Fix fuzzy search matching by replacing MiniSearch with fuzzysort

stevegolton:
Fix odd fuzzy search behavior where search terms like 'gpucompute' did not match 'GpuCompute'. Replace MiniSearch with fuzzysort, which is built for omniboxes, and adds some better support for other things such as multi key seraching.

Summary of changes:
- Switch MiniSearch for fuzzysort (0 deps, bundle size reduced by 35K).
- Modify the interface to allow multiple key lookup functions to be passed which is now handled natively by fuzzysort.
- Add fuzzy demo page in the widgets page (#!/widgets/fuzzy-search)
- Sprinkly readonly on arrays liberally where appropriate.

Update FuzzyFinder to support single and multi-key lookups, allowing Omnibox command search to filter across both command names and sources with per-field match highlights. Add a Fuzzy Search demo to WidgetsPage.


ui/src/base/fuzzy.ts
```diff
@@ -12,129 +12,175 @@
...
 
-  // Return a list of items that match any of the search terms.
-  // Search terms separated by spaces are matched independently (any order).
-  find(searchTerm: string): FuzzyResult<T>[] {
+  // Return a list of items that match the search term.
+  find(searchTerm: string): R[] {
     if (searchTerm === '') {
-      return this.items.map((item) => {
-        const normalisedTerm = this.keyLookup(item);
+      if (this.isMulti) {
```
LalitMaganti: nit: the is multi branch. here and below feels a little ugly. Any easy way of refactoring this to make this a bit nicer? I was thinking maybe splitting into two functions but that that might be too much. WIll leave it to you to decide.

LalitMaganti (approved):

ui/src/base/fuzzy.ts
stevegolton: Split the multi case out into a function, and use the main function as a router. When multi key - just forward to the helper, else in the single key case also forward to the helper but adapt the input and output. This is DRY and reduces the mess.

PR #6857, opened 2026-07-23
tp: stdlib: Ensure SF-less frames are associated with input

ivan-chong:
This PR changes our frame association logic to use android_frames_choreographer_do_frame. This fixes an issue where input events, which are delivered to a frame which did not invoke SF, would be speculatively correlated as android_frames filters them out.

Bug: 537643056
Bug: 454761692


ivan-chong:
Friendly ping, thanks!

LalitMaganti (approved):

PR #6859, opened 2026-07-23
ui: fix integration tests with rootless Docker

LalitMaganti:
Run Playwright containers as container root when Docker
uses rootless user namespacing, keeping owner-only build
dependencies accessible through bind mounts. Also pass an
explicit snapshot update mode for current Playwright versions.

### DIFF
--- a/ui/run-integrationtests
+++ b/ui/run-integrationtests
@@ -144,7 +144,7 @@ def run_native(args):
       return 1
     cmd += ['--ui']
   elif args.rebaseline:
-    cmd += ['--update-snapshots=all']
+    cmd += ['--update-snapshots']
 
   if args.workers:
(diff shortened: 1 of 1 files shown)


ui/run-integrationtests
```diff
@@ -127,7 +144,7 @@ def run_native(args):
       return 1
     cmd += ['--ui']
   elif args.rebaseline:
-    cmd += ['--update-snapshots']
+    cmd += ['--update-snapshots=all']
```
stevegolton: What effect does this have? Seems unrelated.

stevegolton (approved):

ui/run-integrationtests
LalitMaganti: Removed.

PR #6865, opened 2026-07-23
ui: show OOME heap-dump metadata as a key-value grid

zezeozue:
Render the OutOfMemoryError summary shown above the heap-dump callstack flamegraph as a Property/Value grid (allocation size, free until OOM, and the raw error message) instead of just the raw string.

The same grid is reused in the HeapProfile details panel, the Heap Dump Explorer callstack page, and a new
"Out of Memory Error" section in the overview


zezeozue:
New
image
image
Old
image

stevegolton (approved):

PR #6866, opened 2026-07-23
ui: Open aggregation drilldowns in new tab

### LINKED ISSUE #6817: Area selection datagrid -> id selection UX is suboptimal
Here's a screenrecord showing my problem:  https://github.com/user-attachments/assets/2c259590-09dc-4a40-bfc7-732d0b9c40d5

Basically if I do an area select then use the datagrid to get to a single selection, I immediately loose the area select which I might have spent a long time working on perfecting.

It would be nice if there was some way to fix this UX so we didn't lose the area select: persisting this in another tab for example would be very nice.

stevegolton:
Clicking the drill-down button in an aggregator now copies the entire pivot model + selection to a new tab with the drill-down applied. The original aggregation tab's model stays as-is.

Also add 'Add debug track' menu button to the drill-down view, and add 'ts' columns to relevant aggregators to make debug tracks work.

This means that when clicking when subiquently clicking on a slice link to reveal the slice on the timeline from one of these drilldown tabs, the drill-down list is not lost the selection changes.

Fixes: #6817


ui/src/plugins/dev.perfetto.Sched/thread_state_selection_aggregator.ts
```diff
@@ -117,27 +119,29 @@ export class ThreadStateSelectionAggregator implements Aggregator {
           area.end,
         );
 
-        await engine.query(`
-          include perfetto module android.cpu.cluster_type;
+        const table = await createPerfettoTable({
+          engine,
+          as: `
+            include perfetto module android.cpu.cluster_type;
```
LalitMaganti: I don't remember: is includes perfetto module allowed in `as`?

LalitMaganti (approved):

ui/src/plugins/dev.perfetto.Sched/thread_state_selection_aggregator.ts
stevegolton: Good catch, no! Fixed.

PR #6868, opened 2026-07-23
ui: don't crash the flamegraph measure picker when no metric is selected

zezeozue:
During the loading state the metric list can be empty, so ensureExists on the selected metric threw. Render nothing instead. Hit by opening the Heap Dump Explorer callstack tab for an OutOfMemoryError with no recorded callstack (data loss).

### DIFF
--- a/ui/src/widgets/flamegraph.ts
+++ b/ui/src/widgets/flamegraph.ts
@@ -895,11 +895,12 @@ export class Flamegraph implements m.ClassComponent<FlamegraphAttrs> {
   }
 
   private renderMeasurePicker(attrs: FlamegraphAttrs) {
-    const selected = ensureExists(
-      attrs.metrics.find(
-        (metric) => metricId(metric) === attrs.state.selectedMetricId,
-      ),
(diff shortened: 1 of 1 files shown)


ui/src/widgets/flamegraph.ts
```diff
@@ -895,11 +895,12 @@ export class Flamegraph implements m.ClassComponent<FlamegraphAttrs> {
   }
 
   private renderMeasurePicker(attrs: FlamegraphAttrs) {
-    const selected = ensureExists(
-      attrs.metrics.find(
-        (metric) => metricId(metric) === attrs.state.selectedMetricId,
-      ),
+    const selected = attrs.metrics.find(
+      (metric) => metricId(metric) === attrs.state.selectedMetricId,
```
LalitMaganti, line 899: Won't this just cause infinite spin loading?

zezeozue: Tested on data loss trace and seems fine:

image

LalitMaganti (approved):

PR #6871, opened 2026-07-23
ui: add a table view to flamegraphs

zezeozue:
Add a Flame/Table toggle to the flamegraph filter bar. Table mode shows the same tree as the canvas — one expandable row per node with self and cumulative values and the share of the root total — so metric selection, filters and the top-down/bottom-up/pivot views apply to both modes.

The table is a DataGrid in id-based tree mode. QueryFlamegraph backs it with a source over the nodes the canvas already holds, so nothing extra is loaded: expanding walks only the visible part of the tree and the grid virtualises the DOM. Callers without a resident node array can pass a SQL data source instead, which fetches one level per expansion. The Flamegraph widget delegates the rendering to a renderTableView hook, keeping the widgets layer free of DataGrid dependencies.

### DIFF
--- a/ui/src/components/flamegraph_table.ts
+++ b/ui/src/components/flamegraph_table.ts
@@ -1,0 +1,177 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


LalitMaganti:
I have very strong thoughts here I was actually going to send this myself. The right thing is not to couple with flamegraph but instead:
1) pull out the filter bar and share it between flamegraph and this tree view
2) we should really also then have a flattened representation similar to pprof which should be normal datagrid
3) we should present this all as a unified super widget for the details panel sharable between all the tree like data.

PR #6873, opened 2026-07-24
tp: add __intrinsic_zstd(X) to compress a blob/string in SQL

zezeozue:
Compresses a string or blob into a zstd frame and returns it as a BLOB, the compression counterpart of the existing zstd decompression. Intended for compressing a payload generated in SQL (e.g. Chrome JSON assembled from the tables) before storing it; the frame is understood by the trace importer, so the blob re-opens as a trace. Uses zstd level 9, a good ratio/speed balance for the compressible text this targets.

Test: diff test PerfettoFunction:zstd asserts the zstd frame magic, non-empty output and NULL passthrough; verified a Chrome JSON compresses to a frame that zstd -d restores byte-for-byte and trace_processor re-opens as a trace.

### DIFF
--- a/src/trace_processor/plugins/zstd_functions/zstd.cc
+++ b/src/trace_processor/plugins/zstd_functions/zstd.cc
@@ -1,0 +1,109 @@
+// Copyright (C) 2026 The Android Open Source Project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/trace_processor/plugins/zstd_functions/zstd_functions.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
+#include <memory>
+#include <vector>
+
+#include "perfetto/base/logging.h"
+#include "perfetto/ext/base/utils.h"
+#include "src/trace_processor/core/plugin/plugin.h"
+#include "src/trace_processor/perfetto_sql/engine/perfetto_sql_connection.h"
+#include "src/trace_processor/sqlite/bindings/sqlite_function.h"
+#include "src/trace_processor/sqlite/bindings/sqlite_result.h"
+#include "src/trace_processor/sqlite/bindings/sqlite_type.h"
+#include "src/trace_processor/sqlite/bindings/sqlite_value.h"
+#include "src/trace_processor/sqlite/sqlite_utils.h"
+#include "src/trace_processor/util/zstd_compressor.h"
+
+namespace perfetto::trace_processor {
+
+namespace {
+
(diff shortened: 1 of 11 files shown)


LalitMaganti:
Overall idea is ok but Ashwin is the expert on this. In particular not clear if there's not already a compressor you should be using.

primiano:
Out of curiosity how/where is this going to be used? What is it for?

zezeozue:
> Out of curiosity how/where is this going to be used? What is it for?

This is for APC metrics where we want to replace the SVGs [here](https://brush-staging.corp.google.com/?metric_id=android_startup&charts=gallery&gallerySvgColumn=svg&galleryMetricColumn=dur_ms&galleryMetricNameColumn=process_name) 
image
 that show scrollable pictures of startups across multiple trace  with a chrome json trace containing only the relevant snippets of app startups across multiple traces. See cl/952735635.

The json outputs are larger than needed but compress well so this will save on storage on the APC side.

src/trace_processor/plugins/zstd_functions/zstd.cc
```diff
@@ -0,0 +1,109 @@
...
+namespace {
+
+// A good ratio/speed balance for the compressible text (Chrome JSON etc.) this
+// is meant to store.
+constexpr int kLevel = 9;
+
+// __intrinsic_zstd(X) -> zstd(X) as a BLOB. X is a string or blob; NULL yields
+// NULL.
+struct Zstd : public sqlite::Function<Zstd> {
+  static constexpr char kName[] = "__intrinsic_zstd";
```
sashwinbalaji: nit: name it as __intrinsic_zstd_compress

src/trace_processor/plugins/zstd_functions/zstd.cc
```diff
@@ -0,0 +1,109 @@
...
+        src = reinterpret_cast<const uint8_t*>(sqlite::value::Blob(argv[0]));
+        src_size = static_cast<size_t>(sqlite::value::Bytes(argv[0]));
+        break;
+    }
+
+    size_t out_size = 0;
+    auto out =
+        util::ZstdCompressor::CompressFully(src, src_size, &out_size, kLevel);
+    if (!out) {
+      return sqlite::utils::SetError(
```
sashwinbalaji: nit: you can use IsZstdSupported to see if compile and provide the error

src/trace_processor/util/zstd_compressor.cc
```diff
@@ -0,0 +1,62 @@
...
+// static
+std::unique_ptr<uint8_t, base::FreeDeleter> ZstdCompressor::CompressFully(
+    const uint8_t* data,
+    size_t len,
+    size_t* out_size,
+    int level) {
+  size_t bound = ZSTD_compressBound(len);
+  std::unique_ptr<uint8_t, base::FreeDeleter> out(
+      static_cast<uint8_t*>(malloc(bound)));
+  size_t n = ZSTD_compress(out.get(), bound, data, len, level);
```
sashwinbalaji, line 48: nit/depends on usecase: if we are going to have just a few rows this is fine, but you can also use ZSTD_compressCCtx to avoid ZSTD_CCtx alloc for each call of ZSTD_compress. If not for now, maybe just leave a comment for future.

sashwinbalaji (commented):
so sorry for the delay, was thinking of ways to unify the existing ones with this. But I think yeah a separate contained one makes sense, we can make it cpode agnostic in future if needed.

sashwinbalaji (approved):

PR #6890, opened 2026-07-27
tp: base: use saturating multiple to handle timestamp overflows

LalitMaganti:
Negative timestamps were overflying and wrapping back round to being
positive numbers causing crashes in the UI.

Use saturating multiple instead which preserve the sign meaning that
we accurately drop these events as negative timestamps .


src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -456,7 +457,7 @@ ModuleResult TrackEventTokenizer::TokenizeTrackEventPacket(
       return ModuleResult::Handled();
     }
     timestamp = track_event->IncrementAndGetTrackEventTimeNs(
-        event.timestamp_delta_us() * 1000);
+        base::SaturatingMultiply(event.timestamp_delta_us(), 1000));
```
sashwinbalaji, line 461: should addition within IncrementAndGetTrackEventTimeNs be also made saturating as otherwise the same overflow can happen

```
  int64_t IncrementAndGetTrackEventTimeNs(int64_t delta_ns) {
    PERFETTO_DCHECK(timestamps_valid());
    timestamp_ns_ += delta_ns;
    return timestamp_ns_;
  }
```

sashwinbalaji (approved):

PR #6891, opened 2026-07-27
tp: extract reusable core row operations

LalitMaganti:
Move distinct estimation, deduplication, and row-layout sorting into
shared span-based kernels. Reuse them from Dataframe and the bytecode
interpreter to remove duplicate implementations.


src/trace_processor/core/util/span.h
```diff
@@ -36,10 +38,38 @@ struct Span {
...
   T* begin() const { return b; }
   T* end() const { return e; }
+  T* data() const { return b; }
   size_t size() const { return static_cast<size_t>(e - b); }
   bool empty() const { return b == e; }
+  T& operator[](size_t index) const { return b[index]; }
 };
 
+template <typename T, size_t N>
+Span<T> MakeMutableSpan(T (&values)[N]) {
```
safayat-google, line 48: nits, declare these functions in this file as inline?

src/trace_processor/core/util/ops.cc
```diff
@@ -0,0 +1,177 @@
...
+  if (result > total) {
+    result = total;
+  }
+  return static_cast<uint32_t>(result);
+}
+
+void DistinctRows(Span<const uint8_t> row_layout,
+                  uint32_t row_stride,
+                  Span<uint32_t>* indices) {
+  if (indices->empty()) {
```
safayat-google, line 119: nits, null pointer check for indices? also for SortRowLayout.

safayat-google (approved):

src/trace_processor/core/util/span.h
LalitMaganti: You don't need to declare template functions inline, they automatically always are :)

PR #6892, opened 2026-07-27
proto_merger: allow message-to-message type transitions

BogdanLionte:


dragosp95 (approved):

src/tools/proto_merger/proto_file_serializer_unittest.cc
```diff
@@ -640,6 +640,91 @@ TEST(ProtoFileSerializerTest, PassthroughDeepRecursion) {
...
+  ProtoFile upstream_file = ProtoFileFromDescriptor("", *upstream_desc);
+
+  ProtoFile merged;
+  // This should succeed because transitioning from one message type to another
+  // is allowed.
+  ASSERT_TRUE(MergeProtoFiles(input_file, upstream_file, allowed, merged).ok());
+
+  std::string out = ProtoFileToDotProto(merged);
+  // Expect output to use the new upstream type name.
+  EXPECT_THAT(out, HasSubstr("Callstack callstack = 55;"));
```
sashwinbalaji: nit: this would still pass for `InlineCallstack callstack`

src/tools/proto_merger/proto_file_serializer_unittest.cc
```diff
@@ -640,6 +640,91 @@ TEST(ProtoFileSerializerTest, PassthroughDeepRecursion) {
   EXPECT_THAT(out, HasSubstr("proto_filter_merge_passthrough) = true"));
 }
 
+TEST(ProtoFileSerializerTest, MessageToMessageTypeTransitionAllowed) {
```
sashwinbalaji, line 643: can we follow the existing style of TypeTransitionAllowedSucceeds which would also make the test shorter.
http://ac/external/perfetto/src/tools/proto_merger/proto_file_serializer_unittest.cc?l=250-274

src/tools/proto_merger/proto_file_serializer_unittest.cc

src/tools/proto_merger/proto_file_serializer_unittest.cc
BogdanLionte: Done

sashwinbalaji:
I don't think this will fix the issue.

I tried running it on g3 head and it's updating the field but not the msg the field points at.  

You will find `InlineCallstack callstack = 55;` in the file but the defintion would still point to old Callstack only which will cause build failure

> "InlineCallstack" is not defined.

I think it's because MergeRecursive finds nested msgs by name, so they don't match and are also not in allowlist so it's dropped.

BogdanLionte:
> I don't think this will fix the issue.
> 
> I tried running it on g3 head and it's updating the field but not the msg the field points at.
> 
> You will find `InlineCallstack callstack = 55;` in the file but the defintion would still point to old Callstack only which will cause build failure
> 
> > "InlineCallstack" is not defined.
> 
> I think it's because MergeRecursive finds nested msgs by name, so they don't match and are also not in allowlist so it's dropped.

Actually, this might not be needed at all since the G3 copybara proto with InlineCallstack won't be used anymore.

PR #6894, opened 2026-07-27
tp: add opt-in abstract filesystem access

LalitMaganti:
Add minimal synchronous File and FileSystem interfaces for Trace Processor
features which write named files. A serializable config knob gates SQL access,
while a borrowed PlatformInterface supplies the filesystem capability.

Let shell embedders provide the implementation and gate SQL file access behind
--allow-sql-file-access, with an equivalent Python option. Route EXPORT_JSON
and __intrinsic_file_write through the interface and remove the legacy
file-descriptor form of EXPORT_JSON.


src/traceconv/trace_to_json.cc
```diff
@@ -31,48 +28,53 @@ namespace trace_to_text {
...
-  auto file = base::TempFile::Create();
-  base::StackString<100> query("select export_json(\"%s\")",
-                               file.path().c_str());
-  auto it = tp->ExecuteQuery(query.ToStdString());
-
-  if (!it.Next()) {
-    auto status = it.Status();
-    PERFETTO_CHECK(!status.ok());
+  TraceWriterOutputWriter output(writer);
+  base::Status status = trace_processor::json::ExportJson(tp, &output);
```
sashwinbalaji, line 69: ExportJson takes storage we are passing tp ?

>  json::ExportJson(trace_context->storage.get(), &output);

src/trace_processor/local_file_system.cc
```diff
@@ -0,0 +1,73 @@
...
+namespace {
+
+class LocalFile final : public File {
+ public:
+  LocalFile(std::string path, base::ScopedFile file)
+      : path_(std::move(path)), file_(std::move(file)) {}
+
+  base::Status Write(const void* data, size_t size) override {
+    ssize_t written = base::WriteAll(file_.get(), data, size);
+    if (written < 0 || static_cast<size_t>(written) != size) {
```
sashwinbalaji: set errno to 0 before call

src/trace_processor/local_file_system_unittest.cc
sashwinbalaji, line 1: rename file to be local_file_system_unittest

src/trace_processor/shell/common_flags.cc
```diff
@@ -56,8 +56,14 @@
 #include "src/trace_processor/rpc/remote_trace_processor.h"
 #endif
 
-namespace perfetto::trace_processor::shell {
+namespace perfetto::trace_processor {
+
+TraceProcessorShell_PlatformInterface::
```
sashwinbalaji, line 63: hmm what's the reason for this ? any specific ordering , if so pls add a comment.

src/traceconv/trace_to_json.cc
```diff
@@ -100,11 +102,10 @@ int TraceToJson(std::istream* input,
   }
 
   // TODO(eseckler): Support truncation of userspace event data.
-  if (ExportUserspaceEvents(tp.get(), trace_writer.get())) {
-    trace_writer->Write(",\n");
-  } else {
-    trace_writer->Write(kTraceHeader);
+  if (!ExportUserspaceEvents(tp.get(), trace_writer.get())) {
```
sashwinbalaji, line 105: is this behavior change safe of not writning kTraceHeader in case of failure ?

src/trace_processor/trace_processor_storage_impl.h
```diff
@@ -30,10 +30,12 @@
 namespace perfetto::trace_processor {
 
 class ForwardingTraceParser;
+class TraceProcessor_PlatformInterface;
 
 class TraceProcessorStorageImpl : public TraceProcessorStorage {
  public:
-  explicit TraceProcessorStorageImpl(const Config&);
+  TraceProcessorStorageImpl(const Config&,
```
sashwinbalaji: nit: can we keep this explicit

include/perfetto/trace_processor/trace_processor.h
```diff
@@ -35,16 +35,35 @@
...
+}  // namespace io
+
+class PERFETTO_EXPORT_COMPONENT TraceProcessor_PlatformInterface {
+ public:
+  virtual ~TraceProcessor_PlatformInterface();
+
+  // Returns the filesystem exposed to Trace Processor. The returned object
+  // must outlive the TraceProcessor instance. Supplying a filesystem does not
+  // enable SQL file access unless Config::enable_sql_file_access is also set.
+  virtual io::FileSystem* GetFileSystem() const { return nullptr; }
```
sashwinbalaji: do we need this to be const as then we are force to make file_system_ mutable ?

src/trace_processor/trace_processor_context.cc
```diff
@@ -301,6 +307,8 @@ TraceId TraceProcessorContext::trace_id() const {
 }
 
 void TraceProcessorContext::DestroyParsingState() {
+  Config _config = std::move(config);
```
sashwinbalaji, line 312: nit: comment to explain why keeping this ?

src/trace_processor/plugins/utils_functions/utils_functions.cc
```diff
@@ -56,57 +53,80 @@
...
+  }
+  return context->platform->GetFileSystem();
+}
+
+class FileOutputWriter final : public json::OutputWriter {
+ public:
+  explicit FileOutputWriter(std::unique_ptr<io::File> file)
+      : file_(std::move(file)) {}
+
+  base::Status AppendString(const std::string& value) override {
```
sashwinbalaji, line 70: Maybe need some buffering as AppendString is called once per event and each will invoke write() syscall.

src/traceconv/trace_to_json.cc
LalitMaganti: No it takes a TraceProcessorStorage of which traceProcessor is a subclass.

src/trace_processor/local_file_system.cc
LalitMaganti: THat's a very weird pattern we don't do elsehwere. Instead I've separated ot short writes.

sashwinbalaji (approved):

PR #6898, opened 2026-07-27
tp: Handle empty CpuInfo packets without crashing

safayat-google:
When a trace packet contains an empty CpuInfo message (0 CPUs), SystemProbesParser previously logged an error but did not return early. Because std::all_of on an empty container returns true, it attempted out-of-bounds indexing on the empty cpu_infos vector, leading to a SIGILL crash.
This change adds an early return and increments a new cpu_info_empty analysis stat when an empty CpuInfo packet is encountered.

Bug: 539414702


src/trace_processor/importers/proto/system_probes_parser.cc
```diff
@@ -1158,6 +1158,11 @@ void SystemProbesParser::ParseCpuInfo(ConstBytes blob) {
     cpu_infos.push_back(current_cpu_info);
   }
 
+  if (cpu_infos.empty()) {
+    context_->stats_tracker->IncrementStats(stats::cpu_info_empty);
```
LalitMaganti: Please use import logging. And generally we're following the process of, when you touch a file adding annew import log, you should also migrate all usages in that file to import logs as well.

src/trace_processor/storage/stats.h
```diff
@@ -193,6 +193,8 @@ namespace perfetto::trace_processor::stats {
   F(missing_disk_io_event_name,           kSingle,  kError,    kAnalysis, Scope::kMachineAndTrace,      \
        "ETW Disk IO tracker encountered an event with an opcode for which it " \
         "didn't have a name."),                                                \
+  F(cpu_info_empty,                       kSingle,  kError,    kAnalysis, Scope::kMachineAndTrace,      \
```
LalitMaganti, line 196: Please follow the style of modern stats

LalitMaganti, line 196: Please follow the style of modern stats. That means adding what happened, why it might happen and what used can do.

src/trace_processor/importers/proto/system_probes_parser.cc
safayat-google: Updated with import logging, also replaced other places in this file.

LalitMaganti (approved):

PR #6900, opened 2026-07-27
tp: introduce a typed core Tree representation

LalitMaganti:
Replace TreeColumns with a move-only Tree whose columns expose typed accessors, and migrate tree SQL conversion to the new representation. Keep row ordering and SQL behavior unchanged.


src/trace_processor/core/tree/tree_algorithms_unittest.cc
```diff
@@ -0,0 +1,69 @@
...
+  EXPECT_EQ(tree.parent[2], 1u);
+
+  const Tree::Column& values = tree.columns[2];
+  EXPECT_EQ(values.unchecked_span<int64_t>()[1], 30);
+  EXPECT_EQ(values.unchecked_span<int64_t>()[2], 20);
+  EXPECT_FALSE(values.null_bv.is_set(0));
+  EXPECT_TRUE(values.null_bv.is_set(1));
+  EXPECT_TRUE(values.null_bv.is_set(2));
+}
+
```
safayat-google, line 67: nits, add tests for self loop or cyclic reference?

safayat-google (approved):

PR #6907, opened 2026-07-28
tp: add android process state and freezer dump parsing

parthsane:
Update frameworks_base_trace_packet.proto from upstream Android tree and add Trace Processor parsing for trace-stop dumps.

Specifically, this adds:
  - AndroidProcessStateSnapshot message and nested Record message:
    - pid (1), uid (2), proc_state (3), oom_score (4), capability_flags (5)
  - AndroidFreezerStateSnapshot message and nested Record message:
    - pid (1), freeze_unfreeze_time (2), process_name (3), isFreezeSticky (4)
  - FrameworksBaseTracePacket extensions:
    - android_process_state (1003)
    - android_freezer_state (1004)
  - AndroidProcessDumpModule to parse fields 1003 and 1004, handling ProtoOutputStream zero-value omissions for proc_state and capability_flags by defaulting to 0 when omitted.

### DIFF
--- a/protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
+++ b/protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
@@ -20,6 +20,7 @@
 syntax = "proto2";
 
 import public "protos/perfetto/trace/trace_packet.proto";
+import "protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_common.proto";
 import "protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_interned_data.proto";
 
 package com.android.internal;
@@ -106,10 +107,43 @@ message VideoFrameError {
   optional Reason reason = 2;
 }
 
+message AndroidProcessStateSnapshot {
+  message Record {
+    optional int32 pid = 1;
+    optional int32 uid = 2;
+    optional int32 proc_state = 3;
+    optional int32 oom_score = 4;
+    optional int32 capability_flags = 5;
+  }
+  repeated Record record = 1;
+}
+
+message AndroidFreezerStateSnapshot {
+  message Record {
+    // The uid of the process being frozen or unfrozen.
+    optional int32 uid = 1;
+
+    // The pid of the process being frozen or unfrozen.
+    optional int32 pid = 2;
+
+    // Time since last unfrozen.
+    optional int64 unfrozen_dur_ms = 3;
+
+    // Time since last frozen.
+    optional int64 frozen_dur_ms = 4;
+
+    // Unfreeze reason.
+    optional UnfreezeReason unfreeze_reason = 5;
+  }
+  repeated Record record = 1;
+}
+
 message FrameworksBaseTracePacket {
   extend perfetto.protos.TracePacket {
     optional AppWakelockBundle app_wakelock_bundle = 116;
     optional VideoFrame video_frame = 1000;
     optional VideoFrameError video_frame_error = 1001;
+    optional AndroidProcessStateSnapshot android_process_state = 1002;
+    optional AndroidFreezerStateSnapshot android_freezer_state = 1003;
   }
 }
--- a/src/trace_processor/plugins/android_process_state/android_process_state.cc
+++ b/src/trace_processor/plugins/android_process_state/android_process_state.cc
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/trace_processor/plugins/android_process_state/android_process_state.h"
+
+#include <memory>
+#include <vector>
+
+#include "src/trace_processor/core/plugin/plugin.h"
+#include "src/trace_processor/core/plugin/registration.h"
+#include "src/trace_processor/importers/proto/proto_importer_module.h"
+#include "src/trace_processor/importers/proto/track_event_extension_parser.h"
+#include "src/trace_processor/plugins/android_process_state/android_process_state_module.h"
+#include "src/trace_processor/plugins/android_process_state/android_process_state_tracker.h"
+#include "src/trace_processor/plugins/android_process_state/tables_py.h"
+#include "src/trace_processor/storage/trace_storage.h"
+#include "src/trace_processor/types/trace_processor_context.h"
+
+namespace perfetto::trace_processor::android_process_state {
+namespace {
+
+using tables::AndroidFreezerStateTable;
+using tables::AndroidProcessStateTable;
+
+class AndroidProcessState : public Plugin<AndroidProcessState> {
+ public:
+  ~AndroidProcessState() override;
+
+  void RegisterDataframes(std::vector<PluginDataframe>& out) override {
+    EnsureTables();
+    out.push_back({&process_state_table_->dataframe(),
+                   AndroidProcessStateTable::Name(),
+                   {}});
+    out.push_back({&freezer_state_table_->dataframe(),
+                   AndroidFreezerStateTable::Name(),
+                   {}});
+  }
+
+  void RegisterProtoImporterModules(
+      ProtoImporterModuleContext* module_context,
+      TraceProcessorContext* trace_context) override {
+    EnsureTables();
+    module_context->modules.emplace_back(
+        std::make_unique<AndroidProcessStateModule>(
+            module_context, EnsureTracker(trace_context)));
+  }
+
+  void RegisterTrackEventExtensions(
+      TrackEventExtensionParserContext* context,
+      TraceProcessorContext* trace_context) override {
+    EnsureTables();
+    context->parsers.emplace_back(
+        std::make_unique<AndroidProcessStateExtensionParser>(
+            context, trace_context, EnsureTracker(trace_context)));
+  }
+
+ private:
+  void EnsureTables() {
+    if (!process_state_table_) {
+      process_state_table_ = std::make_unique<AndroidProcessStateTable>(
+          trace_context_->storage->mutable_string_pool());
+    }
+    if (!freezer_state_table_) {
+      freezer_state_table_ = std::make_unique<AndroidFreezerStateTable>(
+          trace_context_->storage->mutable_string_pool());
+    }
+  }
+
+  AndroidProcessStateTracker* EnsureTracker(TraceProcessorContext* ctx) {
+    EnsureTables();
+    if (!tracker_) {
+      tracker_ = std::make_unique<AndroidProcessStateTracker>(
+          ctx, process_state_table_.get(), freezer_state_table_.get());
+    }
+    return tracker_.get();
+  }
+
+  std::unique_ptr<AndroidProcessStateTable> process_state_table_;
+  std::unique_ptr<AndroidFreezerStateTable> freezer_state_table_;
+  std::unique_ptr<AndroidProcessStateTracker> tracker_;
+};
+
+AndroidProcessState::~AndroidProcessState() = default;
+
+}  // namespace
+
+void RegisterPlugin() {
+  PERFETTO_TP_REGISTER_PLUGIN(AndroidProcessState);
+}
+
+}  // namespace perfetto::trace_processor::android_process_state
--- a/src/trace_processor/plugins/android_process_state/android_process_state.cc
+++ b/src/trace_processor/plugins/android_process_state/android_process_state.cc
@@ -1,0 +1,91 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/trace_processor/plugins/android_process_state/android_process_state.h"
+
+#include <memory>
+#include <vector>
+
+#include "src/trace_processor/core/plugin/plugin.h"
+#include "src/trace_processor/core/plugin/registration.h"
+#include "src/trace_processor/importers/proto/proto_importer_module.h"
+#include "src/trace_processor/importers/proto/track_event_extension_parser.h"
+#include "src/trace_processor/plugins/android_process_state/android_process_state_module.h"
+#include "src/trace_processor/plugins/android_process_state/tables_py.h"
+#include "src/trace_processor/storage/trace_storage.h"
+#include "src/trace_processor/types/trace_processor_context.h"
+
+namespace perfetto::trace_processor::android_process_state {
+
+using tables::AndroidFreezerStateTable;
+using tables::AndroidProcessStateTable;
+
+AndroidProcessState::~AndroidProcessState() = default;
+
+void AndroidProcessState::RegisterDataframes(
+    std::vector<PluginDataframe>& out) {
+  EnsureTables();
+  out.push_back({&process_state_table_->dataframe(),
+                 AndroidProcessStateTable::Name(),
+                 {}});
+  out.push_back({&freezer_state_table_->dataframe(),
+                 AndroidFreezerStateTable::Name(),
+                 {}});
+}
+
+void AndroidProcessState::RegisterProtoImporterModules(
+    ProtoImporterModuleContext* module_context,
+    TraceProcessorContext* trace_context) {
+  EnsureTables();
+  module_context->modules.emplace_back(new AndroidProcessStateModule(
+      module_context, trace_context, EnsureTracker(trace_context),
+      freezer_state_table_.get()));
+}
+
+void AndroidProcessState::RegisterTrackEventExtensions(
+    TrackEventExtensionParserContext* context,
+    TraceProcessorContext* trace_context) {
+  EnsureTables();
+  context->parsers.emplace_back(new AndroidProcessStateExtensionParser(
+      context, trace_context, EnsureTracker(trace_context)));
+}
+
+void AndroidProcessState::EnsureTables() {
+  if (!process_state_table_) {
+    process_state_table_ = std::make_unique<AndroidProcessStateTable>(
+        trace_context_->storage->mutable_string_pool());
+  }
+  if (!freezer_state_table_) {
+    freezer_state_table_ = std::make_unique<AndroidFreezerStateTable>(
+        trace_context_->storage->mutable_string_pool());
+  }
+}
+
+AndroidProcessStateTracker* AndroidProcessState::EnsureTracker(
+    TraceProcessorContext* ctx) {
+  EnsureTables();
+  if (!tracker_) {
+    tracker_ = std::make_unique<AndroidProcessStateTracker>(
+        ctx, process_state_table_.get());
+  }
+  return tracker_.get();
+}
+
+void RegisterPlugin() {
+  PERFETTO_TP_REGISTER_PLUGIN(AndroidProcessState);
+}
+
+}  // namespace perfetto::trace_processor::android_process_state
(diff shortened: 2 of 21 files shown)


protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
```diff
@@ -106,10 +106,34 @@ message VideoFrameError {
...
+    optional int32 oom_score = 4;
+    optional int32 capability_flags = 5;
+  }
+  repeated Record record = 1;
+}
+
+message AndroidFreezerStateSnapshot {
+  message Record {
+    optional int32 pid = 1;
+    // Last time app was (un)frozen, 0 for never.
```
zezeozue: Maybe follow the fields here?

https://source.corp.google.com/piper///depot/google3/third_party/perfetto/protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_track_event.proto;l=662?q=frameworks_base_track_event.proto

separate field for time since last freeze/unfreeze and also freeze reason if frozen?

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
```diff
@@ -106,10 +106,34 @@ message VideoFrameError {
...
+    optional int64 freeze_unfreeze_time = 2;
+    optional string process_name = 3;
+    optional bool isFreezeSticky = 4;
```
zezeozue: What does this tell us?

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
```diff
@@ -106,10 +106,34 @@ message VideoFrameError {
...
+  }
+  repeated Record record = 1;
+}
+
+message AndroidFreezerStateSnapshot {
+  message Record {
+    optional int32 pid = 1;
+    // Last time app was (un)frozen, 0 for never.
+    optional int64 freeze_unfreeze_time = 2;
+    optional string process_name = 3;
```
zezeozue: We have pid so should be able to reconstruct process_name, lets use that

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
```
zezeozue, line 3: Please add these as plugins. See the pattern I used here: https://github.com/google/perfetto/commit/9220f063f05948f5dd03aa1438e2edec0da7be50

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
parthsane: This was an old field that I think is no longer being used, removed it.

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
```diff
@@ -106,10 +107,43 @@ message VideoFrameError {
   optional Reason reason = 2;
 }
 
+message AndroidProcessStateSnapshot {
```
zezeozue, line 112: Maybe add a comment to see the event in frameworks_base_track_event.proto. Same there.

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
```diff
@@ -21,6 +21,7 @@ syntax = "proto2";
 
 import public "protos/perfetto/trace/trace_packet.proto";
 import "protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_interned_data.proto";
+import "protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_track_event.proto";
```
zezeozue: Why do you import this? (you dont seem to use any message from there?

src/trace_processor/importers/proto/BUILD.gn
```diff
@@ -232,6 +232,7 @@ source_set("full") {
     "../../../../protos/perfetto/trace/system_info:zero",
     "../../../../protos/perfetto/trace/translation:zero",
     "../../../../protos/third_party/android/art:art_heap_graph_zero",
+    "../../../../protos/third_party/android/frameworks/base/proto/tracing:frameworks_base_trace_packet_zero",
```
zezeozue: do you (still) need this change? ditto below?

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
zezeozue, line 112: Also, can you land these protos first in the platform (since thats the source of truth)

src/trace_processor/plugins/android_process_state/tables.py
```diff
@@ -0,0 +1,114 @@
...
+from python.generators.trace_processor_table.public import CppUint32
+from python.generators.trace_processor_table.public import Table
+from python.generators.trace_processor_table.public import TableDoc
+from python.generators.trace_processor_table.public import WrappingSqlView
+from src.trace_processor.tables.metadata_tables import PROCESS_TABLE
+
+ANDROID_PROCESS_STATE_DUMP_TABLE = Table(
+    python_module=__file__,
+    class_name='AndroidProcessStateDumpTable',
+    sql_name='__intrinsic_android_process_state_dump',
```
zezeozue: Can you avoid creating 2 different tables, for 'snapshot' and for incremental changes? it becomes a pain to query. Just create one synthesized table based on both data sources. Can you see how it was done here: https://github.com/google/perfetto/compare/main...dev/zezeozue/psc_snap

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
parthsane: Ack. Turning this into a draft cl.

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
parthsane: I'm importing the UnfreezeReason enum.

Should I copy it here instead?

zezeozue: I'm not too sure if we should have this proto dep. If importing is fine we should also maybe just import for the process state changed and the FreezerChanged message itself. maybe check with Ashwin

protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_trace_packet.proto
parthsane: Edit: got it. will update

src/trace_processor/plugins/android_process_state/android_process_state.cc
```diff
@@ -0,0 +1,143 @@
...
+    }
+    if (freezer_state_table_) {
+      for (auto it = freezer_state_table_->IterateRows(); it; ++it) {
+        if (it.ts()) {
+          start_ns = std::min(*it.ts(), start_ns);
+          end_ns = std::max(*it.ts(), end_ns);
+        }
+      }
+    }
+    if (start_ns > end_ns) {
```
zezeozue: Is this necessary?

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
```diff
@@ -0,0 +1,475 @@
...
+void AndroidProcessStateTracker::ParseChange(int64_t ts,
+                                             protozero::ConstBytes bytes) {
+  fb::AndroidProcessStateChangedEvent::Decoder p(bytes);
+  if (!p.has_pid()) {
+    return;
+  }
+  int32_t pid = p.pid();
+  UniquePid upid =
+      context_->process_tracker->GetOrCreateProcess(static_cast<uint32_t>(pid));
+  if (p.has_uid()) {
```
zezeozue: Why are you setting process uid here?

src/trace_processor/plugins/android_process_state/tables.py
```diff
@@ -0,0 +1,125 @@
...
+    python_module=__file__,
+    class_name='AndroidProcessStateTable',
+    sql_name='__intrinsic_android_process_state',
+    columns=[
+        C('ts',
+          CppOptional(CppInt64()),
+          cpp_access=CppAccess.READ,
+          cpp_access_duration=CppAccessDuration.POST_FINALIZATION),
+        C('upid', CppTableId(PROCESS_TABLE), cpp_access=CppAccess.READ),
+        C('pid', CppInt32()),
```
zezeozue: We can drop this. Pid, uid and process_name can be recovered from the process table, lets just have upid as join key

src/trace_processor/plugins/android_process_state/tables.py
```diff
@@ -0,0 +1,125 @@
...
+    python_module=__file__,
+    class_name='AndroidFreezerStateTable',
+    sql_name='__intrinsic_android_freezer_state',
+    columns=[
+        C('ts',
+          CppOptional(CppInt64()),
+          cpp_access=CppAccess.READ,
+          cpp_access_duration=CppAccessDuration.POST_FINALIZATION),
+        C('upid', CppTableId(PROCESS_TABLE), cpp_access=CppAccess.READ),
+        C('pid', CppInt32()),
```
zezeozue: ditto, drop these

src/trace_processor/plugins/android_process_state/android_process_state_module.h
```diff
@@ -0,0 +1,152 @@
...
+}  // namespace perfetto::trace_processor
+
+namespace perfetto::trace_processor::android_process_state {
+
+// Fills __intrinsic_android_process_state from the
+// AndroidProcessStateChangedEvent deltas (one row each) and, at end of trace,
+// synthesizes one initial-state row per process (is_initial = 1, ts NULL):
+// processes that emitted a delta use the prev_* state of their earliest delta,
+// the rest use their state from the trace-stop AndroidProcessState dump.
+class AndroidProcessStateTracker {
```
zezeozue: Can you move the tracker stuff to its own header and cc

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
```diff
@@ -0,0 +1,475 @@
...
+    if (rec.has_uid()) {
+      v.uid = rec.uid();
+      context_->process_tracker->SetProcessUid(v.upid,
+                                               static_cast<uint32_t>(*v.uid));
+    }
+
+    auto& process_table = *context_->storage->mutable_process_table();
+    auto prr = process_table[v.upid];
+
+    if (rec.has_process_name()) {
```
zezeozue: Can you leave the process name, time and start_seq id handling out of this pr. We'll deal with that in a different pr. No point having that in a half state

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
```diff
@@ -0,0 +1,475 @@
...
+      continue;
+    }
+    UniquePid upid = context_->process_tracker->GetOrCreateProcess(
+        static_cast<uint32_t>(rec.pid()));
+
+    tables::AndroidFreezerStateTable::Row row;
+    row.ts = std::nullopt;
+    row.upid = upid;
+    row.pid = rec.pid();
+    if (rec.has_uid()) {
```
zezeozue: ditto, why are you setting uid on process table from here?

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
```diff
@@ -0,0 +1,475 @@
...
+    // during serialization on Android, so unset fields in the dump snapshot
+    // represent 0.
+    row.unfrozen_dur_ms = rec.has_unfrozen_dur_ms() ? rec.unfrozen_dur_ms() : 0;
+    row.frozen_dur_ms = rec.has_frozen_dur_ms() ? rec.frozen_dur_ms() : 0;
+    row.unfreeze_reason =
+        InternEnum(context_, unfreeze_reason_cache_,
+                   ".com.android.internal.UnfreezeReason",
+                   static_cast<int32_t>(
+                       rec.has_unfreeze_reason() ? rec.unfreeze_reason() : 0));
+    row.is_initial = 1;
```
zezeozue: For freezer, are you doing the same initial state computation that we do for process state, it seems the values marked as initial here are from the dump at the end?

zezeozue: Note that unfreeze/freeze dur will be incorrect as initial state values. So for now, maybe we can leave those out of the table at least for the initial state

src/trace_processor/plugins/android_process_state/android_process_state.cc
parthsane: Removed it.

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
parthsane: Removed this.

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
```diff
@@ -0,0 +1,162 @@
...
+    default:
+      break;
+  }
+}
+
+void AndroidProcessStateModule::OnEventsFullyExtracted() {
+  tracker_->Finalize();
+}
+
+void AndroidProcessStateModule::ParseAndroidFreezerState(
```
zezeozue: Maybe move the freezer dump + delta handling into the tracker, mirroring process
state, with an EmitInitialRow equivalent

src/trace_processor/plugins/android_process_state/android_process_state_module.cc
```diff
@@ -0,0 +1,162 @@
...
+    }
+    UniquePid upid = context_->process_tracker->GetOrCreateProcess(
+        static_cast<uint32_t>(rec.pid()));
+
+    tables::AndroidFreezerStateTable::Row row;
+    row.ts = std::nullopt;
+    row.upid = upid;
+    row.unfrozen_dur_ms = std::nullopt;
+    row.frozen_dur_ms = std::nullopt;
+    row.unfreeze_reason = std::nullopt;
```
zezeozue: Can you please add the unfreeze_reason (if there is). This way we can differentiate whether the inital state is in frozen or unfrozen state

src/trace_processor/plugins/android_process_state/android_process_state_tracker.cc
```diff
@@ -0,0 +1,191 @@
...
+  fb::AndroidProcessStateSnapshot::Decoder dump(blob);
+  for (auto it = dump.record(); it; ++it) {
+    fb::AndroidProcessStateSnapshot::Record::Decoder rec(*it);
+    if (!rec.has_pid()) {
+      continue;
+    }
+    ProcessStateValues v;
+    v.upid = context_->process_tracker->GetOrCreateProcess(
+        static_cast<uint32_t>(rec.pid()));
+    if (rec.has_uid()) {
```
zezeozue: You are still setting uid here. Please don't change anything on the process table in this pr

src/trace_processor/plugins/android_process_state/tables.py
```diff
@@ -0,0 +1,110 @@
...
+        C('ts',
+          CppOptional(CppInt64()),
+          cpp_access=CppAccess.READ,
+          cpp_access_duration=CppAccessDuration.POST_FINALIZATION),
+        C('upid', CppTableId(PROCESS_TABLE), cpp_access=CppAccess.READ),
+        C('proc_state', CppOptional(CppString())),
+        C('oom_score', CppOptional(CppInt32())),
+        C('capability_flags', CppOptional(CppInt32())),
+        C('reason', CppOptional(CppString())),
+        C('seq_id', CppOptional(CppInt64())),
```
zezeozue: We shouldn't have seq_id here. This is for the android_track_event_process table. The only join key to that and the process table should be upid

src/trace_processor/plugins/android_process_state/tables.py
```diff
@@ -0,0 +1,110 @@
...
+          CppOptional(CppString()),
+          cpp_access=CppAccess.READ),
+        C('is_initial', CppUint32(), cpp_access=CppAccess.READ),
+    ],
+    tabledoc=TableDoc(
+        doc='Process freezer events from snapshots and track events.',
+        group='Android',
+        columns={
+            'ts':
+                'Timestamp of freezer event.',
```
zezeozue: nit: NULL for initial backfill row.', as per process state docs

src/trace_processor/plugins/android_process_state/android_process_state_tracker.cc
```diff
@@ -0,0 +1,258 @@
...
+        static_cast<uint32_t>(rec.pid()));
+
+    // Note: android.util.proto.ProtoOutputStream ignores/omits 0 data points
+    // during serialization on Android, so unset fields in the dump snapshot
+    // represent 0.
+    //
+    // TODO: Consider setting process_name and uid on the core `process` table
+    // from dump records in a future update.
+    v.proc_state =
+        rec.has_proc_state() ? static_cast<int32_t>(rec.proc_state()) : 0;
```
zezeozue: Can these use the default enum constants?

src/trace_processor/plugins/android_process_state/android_process_state_tracker.cc
```diff
@@ -0,0 +1,258 @@
...
+    fb::AndroidFreezerStateSnapshot::Record::Decoder rec(*it);
+    if (!rec.has_pid()) {
+      continue;
+    }
+    FreezerStateValues v;
+    v.upid = context_->process_tracker->GetOrCreateProcess(
+        static_cast<uint32_t>(rec.pid()));
+    // Note: android.util.proto.ProtoOutputStream ignores/omits 0 data points
+    // during serialization on Android, so unset fields represent 0 (UFR_NONE).
+    v.unfreeze_reason = rec.has_unfreeze_reason()
```
zezeozue: what if this is a freeze event so unfreeze reason is really not set?

src/trace_processor/plugins/android_process_state/tables.py
```diff
@@ -0,0 +1,107 @@
...
+    sql_name='__intrinsic_android_process_state',
+    columns=[
+        C('ts',
+          CppOptional(CppInt64()),
+          cpp_access=CppAccess.READ,
+          cpp_access_duration=CppAccessDuration.POST_FINALIZATION),
+        C('upid', CppTableId(PROCESS_TABLE), cpp_access=CppAccess.READ),
+        C('proc_state', CppOptional(CppString())),
+        C('oom_score', CppOptional(CppInt32())),
+        C('capability_flags', CppOptional(CppInt32())),
```
zezeozue, line 40: We should eventually represent this as flag_set with args on the process state table using the: http://google3/third_party/perfetto/src/trace_processor/util/descriptors.cc;l=700;rcl=949143191 which already parses from the proto flag_enum annotation. But can do this as a follow up

zezeozue (approved):

src/trace_processor/plugins/android_process_state/android_process_state_tracker.cc
parthsane: this is only for the dump and those can only be frozen apps.

For unfreeze events, this isn't the function that will be called. For those, the ParseFreezerEvent is called which doesn't force set 0 if null.

src/trace_processor/plugins/android_process_state/android_process_state_tracker.cc
parthsane: Done

src/trace_processor/plugins/android_process_state/tables.py
parthsane: Ack. will do a followup pr

src/trace_processor/plugins/android_process_state/android_process_state_tracker.cc
```diff
@@ -0,0 +1,262 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/trace_processor/plugins/android_process_state/android_process_state_tracker.h"
+
+#include <string>
+
+#include "perfetto/ext/base/string_view.h"
+#include "protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_common.pbzero.h"
```
LalitMaganti: nit: proto includes should be separated and placed below.

src/trace_processor/plugins/android_process_state/android_process_state_module_unittest.cc
```diff
@@ -0,0 +1,257 @@
...
+  EXPECT_FALSE(rr0.unfrozen_dur_ms().has_value());
+  EXPECT_FALSE(rr0.frozen_dur_ms().has_value());
+  ASSERT_TRUE(rr0.unfreeze_reason().has_value());
+  EXPECT_STREQ(context_.storage->GetString(*rr0.unfreeze_reason()).c_str(),
+               "1");
+
+  auto rr1 = (*freezer_state_table_)[1];
+  EXPECT_EQ(rr1.id().value, 1u);
+  EXPECT_FALSE(rr1.ts().has_value());
+  EXPECT_EQ(rr1.is_initial(), 1u);
```
LalitMaganti: I wouldn't bother adding unittests here unless ther'es actual business logic. Just add diff tests.

parthsane: removed the tests

LalitMaganti (approved):

PR #6908, opened 2026-07-28
tp: parse exit reason and connect process death events via start_seq_id

parthsane:
Parse AndroidProcessDiedEvent in the android_framework_track_event
plugin.
Add exit_reason and exit_sub_reason columns to __intrinsic_android_track_event_process table. 

Use start_seq_id to reliably connect AndroidProcessStartEvent, AndroidBinderDiedEvent,
and AndroidProcessDiedEvent events to the correct process record.


src/trace_processor/plugins/android_framework_track_event/android_framework_track_event.cc
```diff
@@ -149,26 +180,80 @@ class Parser : public TrackEventExtensionParser {
...
+      row = FindRow(evt.start_seq_id());
+    }
+
+    if (!row) {
+      auto upid = GetProcess(static_cast<uint32_t>(evt.pid()));
+      std::optional<int64_t> start_seq_id;
+      if (evt.has_start_seq_id()) {
+        start_seq_id = evt.start_seq_id();
+      }
+      row = GetOrInsertRow(upid, start_seq_id);
```
zezeozue, line 243: This will create new upids same pids that were already running when the trace was created. This is why we need to solve the pid attribution first

PR #6911, opened 2026-07-28
Support parsing mergedClasses metadata to better de-obfuscate heap dumps.

MichaelFQuigley:
This supports parsing of JSON comments in the Proguard mapping (.map) files, such as:
```
# {"id":"com.android.tools.r8.mergedClasses", "classIdField": "$cid", "mergedClasses": [{ "name": "ClassA", "classId": 0 }, { "name": "ClassB", "classId": 1 }]}
```
This will allow for disambiguation of de-obfuscated classes in heap dumps.


src/trace_processor/util/deobfuscation/deobfuscator.cc
```diff
@@ -332,9 +447,25 @@ std::map<std::string, std::string> ObfuscatedClass::deobfuscated_methods()
...
   auto first_ch_pos = line.find_first_not_of(" \t");
-  if (first_ch_pos == std::string::npos || line[first_ch_pos] == '#')
+  if (first_ch_pos == std::string::npos)
     return base::Status();
 
+  if (line[first_ch_pos] == '#') {
+    if (current_class_ == nullptr) {
+      return base::Status();
+    }
+    size_t json_start = line.find('{');
```
safayat-google, line 461: There can be other metadata with this prefix `# {` .
So, trying to parse this json can be costly. What about a early substring check for the `kMergedClassesId`?

src/trace_processor/util/deobfuscation/deobfuscator.cc
```diff
@@ -332,9 +447,25 @@ std::map<std::string, std::string> ObfuscatedClass::deobfuscated_methods()
...
+    if (json_start != std::string::npos) {
+      std::string_view json_sv = std::string_view(line).substr(json_start);
+      base::Status s = ParseMergedClassesComment(json_sv, *current_class_);
+      if (!s.ok()) {
+        PERFETTO_DLOG("Failed to parse merged classes comment: %s\non line %s",
```
safayat-google: nit, push a `import_logs` like RecordAnalysisLog instead.

src/trace_processor/util/deobfuscation/deobfuscator.cc
MichaelFQuigley: Thanks. Added a .contains check for the kMergedClassesId key in ParseMergedClassesComment. I did this instead of a starts with check since it looks like the key can be used in different ways based on some examples in R8's specification. eg. single quotes vs double quotes vs. no quotes for the "id".

https://r8.googlesource.com/r8/+/refs/tags/4.0.9-dev/doc/retrace.md

src/trace_processor/util/deobfuscation/deobfuscator.cc
MichaelFQuigley: I raised this to an ELOG to be more consistent with the parse failures in other parts of this file.
Note that I left the other DLOG a DLOG since it's just an FYI and not a failure.

There is currently not an import_logs instance in this file, and injecting it seemed like it could add a lot of extra code to handle this log. Is my alternative acceptable?

LalitMaganti (approved):

safayat-google (approved):

PR #6912, opened 2026-07-29
Trace size optimization: Remove unreferenced intern table entries in redacted traces

edgararriagag:
This optimization reduces trace size by eliminating every unreferenced thread name from the compact sched intern table which is results from redaction, as redaction currently leaves dangling references that consume significant trace size as they are lists of strings.

The average gains in trace size with this optimization are ~9.6% of trace size.

Note: Internal tracking bug b/539563625


src/trace_redaction/redact_sched_events.cc
```diff
@@ -552,44 +560,42 @@ base::Status RedactSchedEvents::OnCompactSchedWaking(
...
     if (waking_filter_->Includes(context, ts_absolute, *it_pid)) {
       // Now that the waking event will be kept, it can be modified using the
       // same rules as switch events.
       auto pid = *it_pid;
-      comm.assign(intern_table->Find(*it_comm_index));
+      comm.assign(source_intern_table->Find(*it_comm_index));
       modifier_->Modify(context, ts_absolute, *it_target_cpu, &pid, &comm);
 
-      auto comm_it = intern_table->Push(comm.data(), comm.size());
+      auto found = std::find(target_intern_table->begin(),
```
LalitMaganti: Isn't this going to be o n^2 since you are already in a loop?

edgararriagag: You are correct, let me see if I can find a way to make this an O(1) search operation with maps.... will circle back once figure it out if it can be done

edgararriagag: My PR got messed up, i found it easier to just open a new one here https://github.com/google/perfetto/pull/7129, and I handled the issue you pointed out now it uses a map

PR #6919, opened 2026-07-29
gen_merged_protos : add new project root parameter to allow merging protos from custom paths

dragosp95:
Add new project root parameter in tools/gen_merged_protos to allow merging protos from custom paths.

This will be used for dynamically merging copies of perfetto files.

This change will not change the default behaviour.

### DIFF
--- a/tools/gen_merged_protos
+++ b/tools/gen_merged_protos
@@ -16,13 +16,29 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+import argparse
 import os
 import re
 import sys
 from codecs import open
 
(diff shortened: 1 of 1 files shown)


tools/gen_merged_protos
```diff
@@ -16,13 +16,29 @@
...
+    description='Merge all Perfetto protos into a single file.')
+parser.add_argument(
+    '--project-root',
+    default=DEFAULT_PROJECT_ROOT,
+    help='Path to the protos project root directory.')
+parser.add_argument(
+    '--check-only',
+    action='store_true',
+    help='Check if merged protos are up to date without modifying them.')
+args, unknown_args = parser.parse_known_args()
```
sashwinbalaji, line 38: should we just do args = parser.parse_args() safer than discarding unknwon ones ?

PR #6923, opened 2026-07-30
Adds generic GPU TrackEvent support across Perfetto

dreveman:
## Summary

Adds generic GPU timelines built from standard `TrackDescriptor` and
`TrackEvent` packets. Producers can control GPU hierarchy, naming, ordering,
annotations, counters, and state tracks without adding API-specific
hierarchy fields.

GPU hierarchy uses ordinary `parent_uuid` relationships. Process-scoped
concepts such as CUDA streams are associated with a process by parenting the
GPU root to a process `TrackDescriptor`.

## Render-stage integration

`GpuTrackDescriptor` can bind an exact authored track to either:

```proto
oneof render_stage_queue {
  uint64 hw_queue_iid = 2;
  uint64 logical_queue_id = 3;
}
```

Matching `GpuRenderStageEvent`s are projected onto that authored track while
remaining available as canonical `gpu_slice` rows. Authored annotations and
projected render-stage events use normal TrackEvent sibling merging:

- tracks merge by name by default;
- `sibling_merge_key` provides an explicit merge identity;
- `SIBLING_MERGE_BEHAVIOR_NONE` keeps the sources separate.

This lets producers add annotations directly to hardware or logical queue
tracks. For example, a producer can emit iteration ranges, scheduling
phases, queue-saturation intervals, or instant markers on the bound
descriptor. Annotations are displayed on the same visual track as the
matching GPU work, without duplicating render-stage events in the producer.

Flow relationships are mirrored onto projected representations without
creating cross-hierarchy or Cartesian duplicate flows.

Hardware queue IDs are scoped to the trusted packet sequence. Logical queue
IDs are scoped to the process resolved from the render event's graphics
context.

## UI behavior

A global GPU descriptor introducing a concrete `gpu_id` acts as the
placement anchor for that GPU. Existing frequency, memory, counters, and
other GPU-specific tracks become direct children of the anchor. This avoids
synthetic structures such as `Counters -> GPU N`.

For example, a producer can create:

```text
NVIDIA A100
  Frequency
  Memory
  Utilization
  Channels
    Channel #1
    Channel #2
```

Process-owned hierarchies can similarly represent logical concepts:

```text
CUDA
  Device #0
    Stream #2
```

Only successfully bound hardware or logical queues move into authored
locations. Unbound sources remain in the existing fallback hierarchy, so
authored and inferred tracks can coexist.

Traces without generic GPU descriptors retain the current UI hierarchy,
naming, and ordering.

## Other changes

- Exposes generic GPU scope and process ownership in Trace Processor.
- Keeps GPU placement policy in the GPU plugins while `TrackEventPlugin`
  remains generic.
- Resolves projected GPU Compute selections back to canonical slices.
- Updates visualization SQL, generated Python/Rust bindings, tests, and GPU
  documentation.
- Bumps `perfetto-protos-gpu` to `1.0.3`.


dreveman:
@LalitMaganti You might want to look at the individual commits as the full set of commits does a lot. This PR provides a complete picture but we can land each commit as a separate PR if you prefer.

LalitMaganti:
Still out on vacation for another week and a bit so will take a look when I'm back.

dreveman:
> Still out on vacation for another week and a bit so will take a look when I'm back.

No worries. I iterated on this approach some and I think the current version is now something I'm pretty happy with and provide all the features I was missing from perfetto when it comes to GPU workloads.

protos/perfetto/trace/gpu/gpu_render_stage_event.proto
```diff
@@ -139,6 +139,10 @@ message GpuRenderStageEvent {
   // optional. Compute kernel launch parameters for this event.
   optional ComputeKernelLaunch launch = 21;
 
+  // Optional producer-defined logical queue identifier. This identifier is
```
LalitMaganti, line 142: This seems a bit orthogonola to the rest of this PR?

src/trace_processor/importers/common/gpu_tracker.cc
```diff
@@ -126,25 +132,86 @@ void GpuTracker::AddRenderStageWait(uint64_t event_id, SliceId slice_id) {
...
     }
   }
 }
 
-void GpuTracker::AddEventWait(uint64_t waited_event_id, SliceId slice_id) {
+void GpuTracker::AddPendingRenderStageQueue(std::function<void()> materialize) {
+  pending_render_stage_queues_.push_back(std::move(materialize));
+}
+
+void GpuTracker::MaterializePendingRenderStageQueues() {
```
LalitMaganti, line 145: Uhhh this seems *very* suspicious. This at least needs some comment on why we need this or please remove it unless it's really necessary.

src/trace_processor/importers/proto/gpu_event_parser.cc
```diff
@@ -934,10 +937,48 @@ void GpuEventParser::ParseGpuRenderStageEvent(
         context_->gpu_tracker->AddGpuRenderStageSlice(event.event_id(),
                                                       slice_id);
       }
-
       for (auto it = event.event_wait_ids(); it; ++it) {
         context_->gpu_tracker->AddEventWait(*it, slice_id);
       }
+
+      auto add_projection =
```
LalitMaganti, line 944: What is "projection" - first time I'm hearing this word. Can we avoid concept like this unless it's a common thing?

src/trace_processor/importers/proto/track_event_parser.cc
```diff
@@ -327,6 +327,8 @@ void TrackEventParser::ParseTrackDescriptor(
     context_->stats_tracker->IncrementStats(stats::track_event_parser_errors);
     return;
   }
+  track_event_tracker_->InternGpuRenderStageQueueDescriptor(decoder.uuid(),
```
LalitMaganti, line 330: This feels like a hack. Why do we need to intercept like this here? Why is it not following the has_thread/has_process flow below?

src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -128,11 +130,36 @@ ModuleResult TrackEventTokenizer::TokenizeTrackDescriptorPacket(
...
   using Reservation = TrackEventTracker::DescriptorTrackReservation;
-  TrackDescriptorProto::Decoder track(
-      args.field.Cast<protos::pbzero::TracePacket::kTrackDescriptor>());
+  protozero::ConstBytes track_bytes =
+      args.field.Cast<protos::pbzero::TracePacket::kTrackDescriptor>();
+  TrackDescriptorProto::Decoder track(track_bytes);
 
   Reservation reservation;
 
+  protozero::Field gpu_track =
```
LalitMaganti, line 139: Any find filed like this is *definitely* a hack. Please fix this.

src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -378,6 +405,15 @@ ModuleResult TrackEventTokenizer::TokenizeTrackDescriptorPacket(
   }
 
   track_event_tracker_->ReserveDescriptorTrack(track.uuid(), reservation);
+  if (reservation.gpu_hw_queue_iid || reservation.gpu_logical_queue_id) {
```
LalitMaganti, line 408: THis should live beside the has_process etc above? This feels like hack once again.

src/trace_processor/importers/proto/track_event_tracker.cc
```diff
@@ -69,6 +70,20 @@ constexpr auto kGlobalCounterTrackBlueprint = tracks::CounterBlueprint(
     tracks::DimensionBlueprints(tracks::LongDimensionBlueprint("track_uuid")),
     tracks::DynamicNameBlueprint());
 
+constexpr auto kGpuCounterTrackBlueprint = tracks::CounterBlueprint(
```
LalitMaganti, line 73: it's while I'm reading this file I'm feeling like maybe we should figure out a more scalable way to do this :)

Especially with the rise of custom dimensions, this problem is only going to get worse. Maybe we should figure out a programatic way to define these blueprints etc to avoid this from just blowing up interms of number of permutations.

LalitMaganti (commented):
First round of comments. I have concerns on how much code this seems to be adding honestly. I was hoping it would be a lot lower.

I'm also concerned on the interaction with custom dimensions and how much duplication there is going to be between this and that.

dreveman:
@LalitMaganti This has grown into a relatively large set of commits, which I've used to prove out the idea and verify that they give me the flexibility to support all pytorch use-cases I have and also cleanup the the gpu-by-process code and make it perform better. 

I think it would be a good to start over after finalizing the generic dimension rfc (maybe even land that first) so we have all that context and then review/merge this in smaller pieces. e.g. proto+tp changes first, then UI changes, then UI cleanup and performance improvements. I think if we do that it will actually turn out much simpler than it looks now. Does that sound good to you?

LalitMaganti:
>  think if we do that it will actually turn out much simpler than it looks now. Does that sound good to you?

Yes sounds good! Thanks a lot!

PR #6928, opened 2026-07-31
Support Perfetto trace recovery across device reboot

safayat-google:
Adds support for persistent traces to survive device reboots on Android. 
When `persist_trace_across_reboots` is enabled in `TraceConfig`, Perfetto saves persistent trace files to disk and recovers/uploads them on the next boot via `--upload-after-reboot`.

Summary of changes:
- Added `persist_trace_across_reboots` to TraceConfig and `AfterRebootTraceEvent` proto header to annotate recovered trace metadata.
- Unlinks trace files upfront before processing in `UploadPersistentTracesAfterReboot` to prevent crash loops if parsing or upload fails.
- Synchronizes status via `traced.reboot_trace_status` sysprop so new tracing sessions wait for active uploads to finish before starting.
- Emits StatSD atoms for trace recovery (atom 66), timeouts (atom 67), and framework handoff (atom 51).
Tested on a device and verified new `perfetto_unittests`

Bug: 382209797
Test: atest perfetto_unittests;


protos/perfetto/config/perfetto_config.proto
```diff
@@ -6026,6 +6030,22 @@ message TraceConfig {
...
+  // Android-only. Not for general use. If set to true, writes trace chunks to a
+  // persistent file
+  // (/data/misc/perfetto-traces/persistent/$unique_session_name.tmp) that
+  // survives device reboots (e.g. during OTA updates or battery depletion).
+  //
+  // Introduced in: perfetto v58.
+  // Supported on: Android 26Q4+.
+  optional bool persist_trace_across_reboots = 49;
+
+  // If set to true, forces explicit buffer flushes prior to periodic write
```
LalitMaganti: Isn't this the default now? Why do we need this?

protos/perfetto/trace/android/after_reboot_trace_event.proto
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// TODO: or should we just reuse an existing an existing proto file?
```
LalitMaganti: It would be nice to use an existing file if possible.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1838,6 +1864,36 @@ void PerfettoCmd::CloneAllBugreportTraces(
...
+          base::OpenFile(persistent_path.c_str(), O_RDONLY | O_CLOEXEC);
+      if (fd_in) {
+        base::ScopedFile fd_out = base::OpenFile(
+            out_path.c_str(), O_CREAT | O_WRONLY | O_TRUNC | O_CLOEXEC, 0600);
+        if (fd_out) {
+          PERFETTO_LOG(
+              "Copying persistent trace %s for write_into_file session %" PRIu64
+              " into bugreport path %s",
+              persistent_path.c_str(), it->tsid, out_path.c_str());
+          base::CopyFileContents(*fd_in, *fd_out);
```
LalitMaganti: Should this not be on a thread? This seems inconsistent with how clone works otherwise.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -46,6 +54,33 @@ namespace {
...
 static constexpr int32_t kTrustedUid = 9999;
 
+// State: <status>:<boot_time_ns>
+// Empty : Not started yet
+// 1:ts  : Started but not finished yet
+// 2:ts  : Finished
+// 3:ts  : Timed out
+const char* kRebootTraceStatusProp = "traced.reboot_trace_status";
+constexpr uint64_t kUploadWaitTimeoutNs =
+    5ULL * 60 * 1000 * 1000 * 1000;  // 5 minutes
```
LalitMaganti: This seems like a very large timeout on the surface. Is there some motivation? Maybe add a comment?

LalitMaganti: Oh is this the time to wait for boot? Feels this variable is misnamed...

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +278,254 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    PERFETTO_FATAL(
+        "Timed out waiting for uploader to set property status for session "
+        "'%s'!",
+        session_name.c_str());
+  }
+
+  // If property is set, but the persistent trace file STILL exists on disk,
+  // log error, unlink file, and crash.
+  if (base::FileExists(target_file_path)) {
+    PERFETTO_ELOG(
```
LalitMaganti: We need statsd attoms for any case like this...

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +278,254 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  std::string clean_name = session_name.empty() ? "default" : session_name;
+  base::StackString<256> file_path("%s/%s.tmp", dir_path.c_str(),
+                                   clean_name.c_str());
+
+  // Wait for any previous pending reboot trace upload for this session name to
+  // complete if the persistent trace file exists on disk.
+  WaitForPreviousRebootTraceUpload(clean_name, file_path.c_str());
+
+  // Unlink any pre-existing instance of this persistent trace file.
+  unlink(file_path.c_str());
```
LalitMaganti: If we did O_TRUNC in the flags below, wouldn't that be stricly better than unlink + open?

LalitMaganti (commented):
First pass through these changes. Overall looking in the right direction but some questions.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1838,6 +1864,36 @@ void PerfettoCmd::CloneAllBugreportTraces(
...
+          PERFETTO_PLOG("Failed to open bugreport output file %s",
+                        out_path.c_str());
+        }
+      } else {
+        PERFETTO_LOG("Persistent trace %s for write_into_file session %" PRIu64
+                     " not found",
+                     persistent_path.c_str(), it->tsid);
+      }
+      sync_fn();
+      continue;
```
sashwinbalaji: this continue seems a bit off to me, if we hit then we'll never invoke CloneSessionOnThread below? I think we should not guard on basis of is_write_into_file.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Renamed and added comment.\
We are keeping it long intentionally since after the timeout we are going to crash anyway.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  }
+
+  // Inject AfterRebootTraceEvent packet into trace stream with recovery info.
+  protozero::HeapBuffered<protos::pbzero::TracePacket> packet;
+  auto* after_reboot_evt = packet->set_after_reboot_trace_event();
+  after_reboot_evt->set_original_file_size(
+      static_cast<uint64_t>(mmap.length()));
+  after_reboot_evt->set_bytes_truncated(bytes_truncated);
+
+  std::vector<uint8_t> packet_bytes = packet.SerializeAsArray();
```
sashwinbalaji: I think packet needs to be a protos::pbzero::Trace

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1838,6 +1857,7 @@ void PerfettoCmd::CloneAllBugreportTraces(
     // Clone the tracing session into the bugreport file.
     std::string out_path = GetBugreportTraceDir() + "/" + actual_fname;
     remove(out_path.c_str());
+
```
sashwinbalaji: nit: not needed change

src/perfetto_cmd/BUILD.gn
```diff
@@ -148,6 +150,9 @@ perfetto_unittest_source_set("unittests") {
     "../../protos/perfetto/config:cpp",
     "../../protos/perfetto/config/ftrace:cpp",
     "../../protos/perfetto/trace:cpp",
+    "../../protos/perfetto/trace/android:cpp",
+    "../../protos/perfetto/trace/android:zero",
```
sashwinbalaji: why the android ones are needed (line 58) also ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -46,6 +54,36 @@ namespace {
...
+// and unlink pre-existing trace files from disk. Sized to 5 minutes to
+// accommodate worst-case full device boot and service scheduling.
+constexpr uint64_t kBootTraceCleanupTimeoutNs =
+    5ULL * 60 * 1000 * 1000 * 1000;  // 5 minutes
+
+enum class RebootTraceUploadState {
+  kTraceUploadUninitialized = 0,
+  kTraceUploadStarted = 1,
+  kTraceUploadFinished = 2,
+  kTraceUploadTimedout = 3,
```
sashwinbalaji: don't see any usage of this state, so lets remove it ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -17,24 +17,32 @@
...
+#include <thread>
 
 #include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
+#include "perfetto/ext/base/android_utils.h"
 #include "perfetto/ext/base/file_utils.h"
 #include "perfetto/ext/base/scoped_mmap.h"
 #include "perfetto/ext/base/string_utils.h"
 #include "perfetto/ext/base/uuid.h"
+#include "perfetto/ext/base/watchdog.h"
```
sashwinbalaji: why this include ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+}
+
+int PerfettoCmd::UploadPersistentTracesAfterReboot() {
+  base::StackString<256> persistent_dir("%s/persistent", kStateDir);
+
+  // Initialize property to empty string if unset to ensure Bionic property node
+  // exists.
+  if (base::GetAndroidProp(kRebootTraceStatusProp).empty()) {
+    SetRebootTraceStatusProp(RebootTraceUploadState::kTraceUploadUninitialized);
+  }
```
sashwinbalaji: don't think kTraceUploadUninitialized and its flow is needed as GetAndroidProp handles a non-existing property same as a property with an empty value. This will also clean up the custom branch in SetRebootTraceStatusProp

https://crsrc.org/c/third_party/perfetto/include/perfetto/ext/base/android_utils.h;l=31-35

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  if (!base::FileExists(persistent_dir.c_str())) {
+    return 0;
+  }
+
+  std::vector<std::string> files;
+  base::Status status = base::ListFilesRecursive(persistent_dir.c_str(), files);
+  if (!status.ok()) {
+    PERFETTO_ELOG("reboot-trace: Failed to list files in %s: %s",
+                  persistent_dir.c_str(), status.c_message());
+    return 0;
```
sashwinbalaji, line 420: for these early returns should we update the state ? we can use OnScopeExit

protos/perfetto/config/trace_config.proto
```diff
@@ -987,4 +987,13 @@ message TraceConfig {
...
   optional TraceAttributes trace_attributes = 48;
+
+  // Android-only. Not for general use. If set to true, writes trace chunks to a
+  // persistent file
+  // (/data/misc/perfetto-traces/persistent/$unique_session_name.tmp) that
+  // survives device reboots (e.g. during OTA updates or battery depletion).
+  //
+  // Introduced in: perfetto v59.
+  // Supported on: Android 26Q4+.
+  optional bool persist_trace_across_reboots = 49;
```
sashwinbalaji, line 998: we should add a check that persist_trace_across_reboots is set only if write_into_file is also set during the parsing phase

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+      PERFETTO_PLOG("reboot-trace: Failed to mmap persistent trace %s",
+                    pending.file_name.c_str());
+      continue;
+    }
+
+    std::optional<TraceConfig> config = ParseTraceConfigFromMmapedTrace(mmap);
+    if (!config.has_value()) {
+      android_stats::MaybeLogUploadEvent(
+          PerfettoStatsdAtom::kRebootTraceParseFailed, /*uuid_lsb=*/0,
+          /*uuid_msb=*/0, pending.file_name);
```
sashwinbalaji, line 483: not a big fan of this as, we are (ab)using trigger_name to pass file_name and session_name here and above

```
void MaybeLogUploadEvent(PerfettoStatsdAtom atom,
                         int64_t uuid_lsb,
                         int64_t uuid_msb,
                         const std::string& trigger_name = "");
```

protos/perfetto/trace/perfetto/tracing_service_event.proto
```diff
@@ -92,3 +92,13 @@ message TracingServiceEvent {
...
+
+// Injected into TracePacket by perfetto --upload-after-reboot to record
+// post-reboot recovery metadata.
+message AfterRebootTraceEvent {
+  // Raw trace file size (in bytes) on disk prior to post-reboot recovery.
+  optional uint64 original_file_size = 1;
+
+  // Amount of trailing corrupted/incomplete bytes truncated during recovery.
+  optional uint64 bytes_truncated = 2;
+}
```
sashwinbalaji: hmm this file is for "// Events emitted by the tracing service." but this emitted by perfett_cmd so maybe a separate file ?

src/perfetto_cmd/perfetto_cmd.h
```diff
@@ -211,6 +222,14 @@ class PerfettoCmd : public Consumer {
   std::function<void()> on_session_cloned_;
   bool cloned_session_was_write_into_file_ = false;
 
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+  static base::ScopedFile CreatePersistentTmpFile(
+      const std::string& session_name,
+      std::string* out_file_path);
+  int UploadPersistentTracesAfterReboot();
+  std::string persistent_file_path_;
+#endif
```
sashwinbalaji, line 227: nit: place this above in the existing android block along wiht  CreateUnlinkedTmpFile?

src/perfetto_cmd/perfetto_cmd.h
```diff
@@ -75,6 +75,17 @@ class PerfettoCmd : public Consumer {
   void OnObservableEvents(const ObservableEvents&) override;
   void OnSessionCloned(const OnSessionClonedArgs&) override;
 
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+  static size_t SanitizeAndAnnotatePersistentTrace(
```
sashwinbalaji: do these need to be public ?

src/perfetto_cmd/perfetto_cmd.h
```diff
@@ -211,6 +222,14 @@ class PerfettoCmd : public Consumer {
   std::function<void()> on_session_cloned_;
   bool cloned_session_was_write_into_file_ = false;
 
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
+  static base::ScopedFile CreatePersistentTmpFile(
+      const std::string& session_name,
+      std::string* out_file_path);
+  int UploadPersistentTracesAfterReboot();
```
sashwinbalaji: can this also be static?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+      pending_traces.push_back({std::move(fd), file_name});
+    }
+  }
+
+  // Set property status to 1:TS after all persistent trace files are unlinked.
+  // At this point any new reboot-aware trace session can proceed immediately,
+  // since the target persistent file on disk has been unlinked.
+  SetRebootTraceStatusProp(RebootTraceUploadState::kTraceUploadStarted);
+
+  size_t traces_found = pending_traces.size();
```
sashwinbalaji: nit: there is just one caller of this and way below so we can just inline it ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    base::Uuid uuid = extract_trace_uuid(target_file_path);
+    android_stats::MaybeLogUploadEvent(
+        PerfettoStatsdAtom::kRebootTraceUploadTimeout, uuid.lsb(), uuid.msb(),
+        session_name);
+    PERFETTO_ELOG(
+        "Timed out waiting for uploader to set property status for session "
+        "'%s'! Unlinking '%s'.",
+        session_name.c_str(), target_file_path.c_str());
+    remove(target_file_path.c_str());
+    PERFETTO_FATAL(
```
sashwinbalaji: do we need both elog and fatal ? also if we already did cleanup by removing why fatal ?

src/android_stats/perfetto_atoms.h
```diff
@@ -84,6 +84,10 @@ enum class PerfettoStatsdAtom {
   kOnTracingDisabled = 4,
   kFinalizeTraceAndExit = 11,
   kCmdFwReportBegin = 49,
+  kRebootTraceRecovered = 66,
```
sashwinbalaji, line 87: just fyi once this change lands you also need to lang enable these protos westworld side like below:

- http://ag/35014059
- http://cl/801156060

protos/perfetto/trace/perfetto/tracing_service_event.proto
```diff
@@ -92,3 +92,13 @@ message TracingServiceEvent {
     uint32 buffer_cloned = 11;
   }
 }
+
+// Injected into TracePacket by perfetto --upload-after-reboot to record
+// post-reboot recovery metadata.
+message AfterRebootTraceEvent {
+  // Raw trace file size (in bytes) on disk prior to post-reboot recovery.
+  optional uint64 original_file_size = 1;
```
sashwinbalaji: nit: call it original_file_size_bytes ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -46,6 +54,36 @@ namespace {
 // We only trust packages written by traced.
 static constexpr int32_t kTrustedUid = 9999;
 
+// State: <status>:<boot_time_ns>
+// Empty : Not started yet
+// 1:ts  : Started but not finished yet
+// 2:ts  : Finished
+// 3:ts  : Timed out
+const char* kRebootTraceStatusProp = "traced.reboot_trace_status";
```
sashwinbalaji: we would need sepolicy changes for this right ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  // Only block if a persistent trace file with the SAME session name exists on
+  // disk.
+  if (!base::FileExists(target_file_path)) {
+    return;
+  }
+
+  std::string cur_prop = base::GetAndroidProp(kRebootTraceStatusProp);
+
+  // If the file exists on disk and property is empty, wait up to 5 minutes
+  // for property to be set.
```
sashwinbalaji: can we use __system_property_wait + move it after daemonize  ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +281,269 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+
+  std::string clean_name = session_name.empty() ? "default" : session_name;
+  base::StackString<256> file_path("%s/%s.tmp", dir_path.c_str(),
+                                   clean_name.c_str());
+
+  // Wait for any previous pending reboot trace upload for this session name to
+  // complete if the persistent trace file exists on disk.
+  WaitForPreviousRebootTraceUpload(clean_name, file_path.c_str());
+
+  auto fd = base::OpenFile(file_path.c_str(),
```
LalitMaganti, line 350: a comment here about O_TRUNC would be a good idea.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1331,7 +1345,12 @@ bool PerfettoCmd::OpenOutputFile(bool no_clobber) {
   base::ScopedFile fd;
   if (trace_out_path_.empty()) {
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
-    fd = CreateUnlinkedTmpFile();
+    if (trace_config_ && trace_config_->persist_trace_across_reboots()) {
+      fd = CreatePersistentTmpFile(trace_config_->unique_session_name(),
```
sashwinbalaji: would this also work for clone snapshot cases ? Because it will also have same unique session name and so we might remove the parent's ongoing file right because they match and we will fatal also

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Removed using the trigger name here

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Added comment

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Yes right. ag/41007699 to update the sepolicy.

src/perfetto_cmd/perfetto_cmd.cc
safayat-google: Nice catch! Updated the logic to create tmp file for clones.

src/perfetto_cmd/BUILD.gn
safayat-google: Tried adding `if android`, but the including the header with # if guard still fails at the setup_all_config validation. So, still had to included it for the test binary.

src/android_stats/perfetto_atoms.h
safayat-google: Thanks for reminding. Prepared the following changes to merge after merging this one.
* http://ag/41586892
* http://cl/966091835

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Done with OnScopeExit

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: `__system_property_wait`. is used

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Done. Initially splited into multiple func to avoid very large function.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: > also if we already did cleanup by removing why fatal ?
Removed the fatal.

This was from the design feedback. But yeah it makes sense to not crash, since we are cleaning up + pushing atoms. cc: @primiano in case you have thoughts on this.

LalitMaganti (approved):

protos/perfetto/trace/android/after_reboot_trace_event.proto
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Injected into TracePacket by perfetto --upload-after-reboot to record
+// post-reboot recovery metadata.
+message AfterRebootTraceEvent {
```
primiano: naming-wise, I'd call this RecoveredTraceInfo, to make it clear this is about traced recovering a trace file.

I'd also add an enum with reason (today let's have only this one): UNEXPECTED_REBOOT, in case we have more in future, to disambiguate.
Note you'll also need to add an UNSPECIFIED = 0 as usual to be compliant with proto rules.

"TraceEvent" is very misplaced here naming-wise, as it means something very different in perfetto.

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -866,8 +877,17 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
              (trace_config_->write_into_file() &&
               !trace_config_->output_path().empty())) {
     open_out_file = false;
-  } else if (trace_out_path_.empty() && !upload_flag_) {
-    PERFETTO_ELOG("Either --out or --upload is required");
+  } else if (trace_out_path_.empty() && !upload_flag_ &&
+             !trace_config_->persist_trace_across_reboots()) {
```
primiano: I'm not sure about this extra condition. this part of the logic shoudln't be changed.
Either one passed --out or --upload. the "persist trace across reboot"  feature is an add-on, shouln't affect this.
Also I'm inclined to say that for now the persist_trace_across_reboots shoudl be supported only when we do --upload. I wouldn't add the complexity of trying to support the case of that + just writing into a file, as we'll never need it (if i'm wrong we can add that part later, but that's extr aomclexity)

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -866,8 +877,17 @@ std::optional<int> PerfettoCmd::ParseCmdlineAndMaybeDaemonize(int argc,
...
+    PERFETTO_ELOG(
+        "Either --out, --upload, or persist_trace_across_reboots in trace "
+        "config is required");
+    return 1;
+  } else if (trace_config_->persist_trace_across_reboots() &&
+             !trace_config_->write_into_file()) {
+    PERFETTO_ELOG(
```
primiano, line 889: here houls should also check that --upload is passed.
so persist_trace_across_reboots requires both write_into_file and --upload

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+        target_file_path.c_str(), cur_prop.c_str());
+  }
+}
+
+// static
+base::ScopedFile PerfettoCmd::CreatePersistentTmpFile(
+    const std::string& session_name,
+    std::string* out_file_path) {
+  base::StackString<256> dir_path("%s/persistent", kStateDir);
+  base::Mkdir(dir_path.c_str());
```
primiano: can we please make the mkdir in perfetto.rc?
so it' socnsistent with the other cases, and we don't have a magic dir that is created in a different way here. remove the mkdir

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+
+  std::string clean_name = session_name.empty() ? "default" : session_name;
```
primiano: we should do some basic sanitization and at very least StripChars('/') to avoid this to become some weird chained attack (e.g. via session_name=../../etc/passwd).
I think we should have some basic sanitizaion that strips all chars that are not just a-zA-Z0-9-_ (without using regex)

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  base::StackString<256> file_path("%s/%s.tmp", dir_path.c_str(),
+                                   clean_name.c_str());
+
+  // Wait for any previous pending reboot trace upload for this session name to
+  // complete if the persistent trace file exists on disk.
+  WaitForPreviousRebootTraceUpload(clean_name, file_path.c_str());
+
+  // Use O_TRUNC to atomically truncate and overwrite any pre-existing file on
```
primiano: why? it's more reliable to unlink + create a new file. 
If you race and see the file before the other one unliks it and pass this file to gmscore, you are going to mangle a file that the other process is uploading.
just unlink unconidtionally (if it doesn't exist... fine) and then O_CREAT + O_EXCL.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    remove(target_file_path.c_str());
+    PERFETTO_ELOG(
+        "Persistent trace file '%s' still exists on disk even though property "
+        "is set to '%s'! Unlinked file.",
+        target_file_path.c_str(), cur_prop.c_str());
+  }
+}
+
+// static
+base::ScopedFile PerfettoCmd::CreatePersistentTmpFile(
```
primiano: This function must have "wait" in the name or somethign similar. 
The fact that this blocks until the other upload is done is not obvious at all when looking at the caller in bool PerfettoCmd::OpenOutputFile(bool no_clobber).
I think this shoudl be called

WaitForUploadCompleteAndCreatePersistentTmpFile

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  if (!base::FileExists(target_file_path)) {
+    return;
+  }
+
+  std::string cur_prop = base::GetAndroidProp(kRebootTraceStatusProp);
+
+  // If the file exists on disk and property is empty, wait up to 5 minutes
+  // for property to be set.
+  auto start_time = base::GetBootTimeNs();
+#if __ANDROID_API__ >= 26
```
primiano: I think the attempt of trying to use sys_prop_wait creates just more chaos and maintinenace cost than what it gains.
Frankly I don't see a problem with just polling (just every 500ms is fine).
Keep the code more compact and readable

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    } else {
+      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    }
+#else
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+#endif
+    cur_prop = base::GetAndroidProp(kRebootTraceStatusProp);
+  }
+
+  auto extract_trace_uuid = [](const std::string& path) -> base::Uuid {
```
primiano: this is not obvious to me. why in a funciton where we are suppsoed to wait for another trace, we are bothering to extrace the uuid? this is adding a lot of complexity for a reason i don't fully understand.
Is it only for atom logging? what do we buy by logging the UUID of the other trace that didnt' complete?
We sorta know already, as half of the UUID really depends on the subscription name.
I'm unconvinced we'll ever do anything with this uuid (happy to be objected, but i need some rationale) and hence i'd remove this complexithy

src/perfetto_cmd/perfetto_cmd.cc
```diff
@@ -1331,7 +1351,13 @@ bool PerfettoCmd::OpenOutputFile(bool no_clobber) {
   base::ScopedFile fd;
   if (trace_out_path_.empty()) {
 #if PERFETTO_BUILDFLAG(PERFETTO_OS_ANDROID)
-    fd = CreateUnlinkedTmpFile();
+    if (trace_config_ && trace_config_->persist_trace_across_reboots() &&
+        !is_clone()) {
+      fd = CreatePersistentTmpFile(trace_config_->unique_session_name(),
+                                   &persistent_file_path_);
```
primiano: See my other comment, this funciton shoudl have  a name that makes it clearer that you are going to block and wait here. Otherwise one would never tell by just looking at this code.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -159,6 +190,12 @@ void PerfettoCmd::ReportTraceToAndroidFrameworkOrCrash() {
   if (!status.ok()) {
     PERFETTO_FATAL("ReportTraceToAndroidFramework: %s", status.c_message());
   }
+
+  if (!persistent_file_path_.empty()) {
+    PERFETTO_LOG("Unlinking persistent trace file %s after framework handoff",
+                 persistent_file_path_.c_str());
+    unlink(persistent_file_path_.c_str());
```
primiano: please PERFETTO_CHECK (With a !, as success == 0) here.
if this fails, I really want to know via a crash report

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+int PerfettoCmd::UploadPersistentTracesAfterReboot() {
+  base::StackString<256> persistent_dir("%s/persistent", kStateDir);
+
+  // Ensure property status is updated to Finished (2:TS) on any exit path,
+  // preventing newly started tracing sessions from timing out.
+  auto on_exit = base::OnScopeExit([] {
+    SetRebootTraceStatusProp(RebootTraceUploadState::kTraceUploadFinished);
+  });
+
+  if (!base::FileExists(persistent_dir.c_str())) {
```
primiano: this check seems not very useful. the mkdir shoudl be done by init (see my other comemnt) so the dir will always exist.
and if it doesn't exist, ListFilesRecursive should just fail, so you regurn 0 anyways below

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    base::ScopedFile fd = base::OpenFile(full_path.c_str(), O_RDWR | O_CLOEXEC);
+    if (!fd) {
+      PERFETTO_PLOG("reboot-trace: Failed to open persistent trace %s",
+                    full_path.c_str());
+    }
+    // Unlink immediately regardless of open status to guarantee disk cleanup.
+    unlink(full_path.c_str());
+
+    if (fd) {
+      pending_traces.push_back({std::move(fd), file_name});
```
primiano: emplace_back

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +274,283 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    std::optional<uint64_t> file_size = base::GetFileSize(fd);
+    if (!file_size.has_value() || *file_size == 0) {
+      PERFETTO_LOG("reboot-trace: Skipping empty persistent trace %s",
+                   pending.file_name.c_str());
+      continue;
+    }
+
+    base::ScopedFile dup_fd(dup(fd));
+    base::ScopedMmap mmap = base::ScopedMmap::FromHandle(
+        std::move(dup_fd), static_cast<size_t>(*file_size));
```
primiano: imho more readable if you do FromHandle(based::ScopedFile(dup(fd))) inline in the call

primiano (commented):
I have a few comments.
My general advice is: please avoid dealing with niche use cases that require extra code complexity unless they are stricly needed.
There are a few cases in the code where you make the code more complex while trying to deal with things that imho don't justify that complexty.

protos/perfetto/trace/android/after_reboot_trace_event.proto
safayat-google: Renamed as `RecoveredTraceInfo` and added the reason. Yes `AfterRebootTraceEvent` was a confusing naming choice.

src/perfetto_cmd/perfetto_cmd.cc
safayat-google: Removed the extra condition from here.
`persist_trace_across_reboots` is supported only when `--upload` is provided.

src/perfetto_cmd/perfetto_cmd.cc
safayat-google: Added explicit check that `--upload` is provided.

src/perfetto_cmd/perfetto_cmd.cc
safayat-google: function is renamed to `WaitForUploadCompleteAndCreatePersistentTmpFile`

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Removed from here.
`perfetto.rc` already have the mkdir.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Added simple sanitization to be safe.\
Ideally selinux should block perfetto from any read/write outside of the designated folders.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Unlinked unconditionally. Thanks for catching this race scenario!

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: The check is removed

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Renamed to `WaitForUploadCompleteAndCreatePersistentTmpFile`

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Added `PERFETTO_CHECK`

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Yes agree, this is also aligned with my initial thoughts. Additionally these part has very low chance of being hit.

since the unlinking alone should not take even 50ms ideally, what about choosing 200ms as a middle ground?

ccing @sashwinbalaji here since this part is diverging again from the last review.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Removed the uuid calculation.

Yes, it was added only for the purpose of atom logging for the failure paths.  (`kRebootTraceUploadTimeout` / `kRebootTraceUploadLeftover`). Since the session name is pushed already, we can prevent this complexity.

protos/perfetto/trace/android/recovered_trace_info.proto
```diff
@@ -0,0 +1,36 @@
...
+
+syntax = "proto2";
+
+package perfetto.protos;
+
+// Injected into TracePacket by perfetto post-reboot recovery to record
+// recovered trace metadata.
+message RecoveredTraceInfo {
+  enum Reason {
+    UNSPECIFIED = 0;
```
primiano: nit: use REASON_ prefix: REASON_UNSPECIFIED, REASON_UNEXPECTED_REBOOT
because these enums gets added to the scoped of the upper container (in this case message RecoverTraceInfo) they mix up.
It' sa general practice in proto to use some common prefix. it doesn't need to match the enum class name, just something that makes codesearch and autocompletion work nicely.
In this case "REASON_" is short enough . In other cases we have done things liek "LID_RADIO" and the like

src/perfetto_cmd/perfetto_cmd.h
```diff
@@ -153,8 +153,19 @@ class PerfettoCmd : public Consumer {
...
+      const std::string& session_name,
+      std::string* out_file_path);
+  static int UploadPersistentTracesAfterReboot();
+  static size_t SanitizeAndAnnotatePersistentTrace(
+      int fd,
+      const base::ScopedMmap& mmap,
+      const std::string& file_name);
+  static void WaitForPreviousRebootTraceUpload(
+      const std::string& session_name,
+      const std::string& target_file_path);
```
primiano: I think the only functions that need to be in this header are

UploadPersistentTracesAfterReboot and 
WaitForUploadCompleteAndCreatePersistentTmpFile

which need to be seen by perfetto_cmd.cc.
the other two are internal implementation details and should go into the anonymous namespace of perfetto_cmd_android.cc

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -46,8 +54,43 @@ namespace {
 // We only trust packages written by traced.
 static constexpr int32_t kTrustedUid = 9999;
 
+// State: <status>:<boot_time_ns>
+// Empty : Not started yet
+// 1:ts  : Started but not finished yet
+// 2:ts  : Finished
+const char* kRebootTraceStatusProp = "traced.reboot_trace_status";
+// Maximum duration to wait for the boot recovery service to start up
```
primiano, line 63: nit: \n to separate comment from line above

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -159,6 +202,12 @@ void PerfettoCmd::ReportTraceToAndroidFrameworkOrCrash() {
   if (!status.ok()) {
     PERFETTO_FATAL("ReportTraceToAndroidFramework: %s", status.c_message());
   }
+
+  if (!persistent_file_path_.empty()) {
+    PERFETTO_LOG("Unlinking persistent trace file %s after framework handoff",
```
primiano: i'd remove this LOG, imho adds very little value and just adds too much noise to logs.

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  // If the file exists on disk and property is empty, wait up to 5 minutes
+  // for property to be set.
+  auto start_time = base::GetBootTimeNs();
+  while (cur_prop.empty()) {
+    auto elapsed_ns =
+        static_cast<uint64_t>((base::GetBootTimeNs() - start_time).count());
+    if (elapsed_ns >= kBootTraceCleanupTimeoutNs) {
+      break;
+    }
+    base::SleepMicroseconds(200 * 1000);
```
primiano: 500ms is enough

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  // disk.
+  if (!base::FileExists(target_file_path)) {
+    return;
+  }
+
+  std::string cur_prop = base::GetAndroidProp(kRebootTraceStatusProp);
+
+  // If the file exists on disk and property is empty, wait up to 5 minutes
+  // for property to be set.
+  auto start_time = base::GetBootTimeNs();
```
primiano: I think this whole code can become more compact and readable if you:

1. Declare kBootTraceCleanupTimeout `constexpr auto kBootTraceCleanupTimeoutNs = std::chrono::minutes(5);`
2. Here you do
```
 const auto deadline = GetBootTimeNs() + kBootTraceCleanupTimeout;
 std::string cur_prop;

 do {
    base::SleepMicroseconds(200 * 1000);
    cur_prop = base::GetAndroidProp(kRebootTraceStatusProp);
 } while(cur_prop.empty() && base::GetBootTimeNs() < deadline);
```

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  if (lseek(fd, static_cast<off_t>(valid_offset), SEEK_SET) == -1) {
+    PERFETTO_PLOG("Failed to lseek trace file %s", file_name.c_str());
+  } else {
+    base::WriteAll(fd, packet_bytes.data(), packet_bytes.size());
+    valid_offset += packet_bytes.size();
+  }
+  return valid_offset;
+}
+
+int PerfettoCmd::UploadPersistentTracesAfterReboot() {
```
primiano, line 406: Add a comment saying this is the function called when --upload-after-reboot is passed to perfetto_cmd

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    base::ScopedMmap mmap = base::ScopedMmap::FromHandle(
+        base::ScopedFile(dup(fd)), static_cast<size_t>(*file_size));
+    if (!mmap.IsValid()) {
+      PERFETTO_PLOG("reboot-trace: Failed to mmap persistent trace %s",
+                    pending.file_name.c_str());
+      continue;
+    }
+
+    std::optional<TraceConfig> config = ParseTraceConfigFromMmapedTrace(mmap);
+    if (!config.has_value()) {
```
primiano: Here you should add also ` || !config->has_android_report_config()`

as that would be a weird case we should know about

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+          pending.file_name.c_str(), report_status.c_message());
+    } else {
+      traces_uploaded++;
+      PERFETTO_LOG(
+          "reboot-trace: Successfully uploaded recovered reboot trace %s",
+          pending.file_name.c_str());
+    }
+  }
+
+  if (traces_found > 0) {
```
primiano: do we need this log? you log each trace individually, so this block is redundant

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+    int fd,
+    const base::ScopedMmap& mmap,
+    const std::string& file_name) {
+  // Compute the valid boundary of complete trace packets in the file.
+  size_t valid_offset = 0;
+  protozero::ProtoDecoder trace_decoder(mmap.data(), mmap.length());
+  for (auto packet = trace_decoder.ReadField(); packet;
+       packet = trace_decoder.ReadField()) {
+    valid_offset = trace_decoder.read_offset();
+  }
```
primiano, line 371: CAn you plz doublecheck that this doens't drop the last packet?
I don't remember how ProtoDecoder works. does it increase the read_offset to the end of the current ReadField?
I am 50:50 between "This is correct" or "this will point to the beginning of the last packet, so you will miss the last packet from each trace"

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+        "reboot-trace: Truncating incomplete trailing trace packet from %zu to "
+        "%zu bytes (%" PRIu64 " bytes truncated) in %s",
+        mmap.length(), valid_offset, bytes_truncated, file_name.c_str());
+    if (ftruncate(fd, static_cast<off_t>(valid_offset)) != 0) {
+      PERFETTO_PLOG("Failed to ftruncate trace file %s", file_name.c_str());
+    }
+  }
+
+  // Inject RecoveredTraceInfo packet into trace stream with recovery info.
+  protozero::HeapBuffered<protos::pbzero::Trace> trace;
```
primiano: `trace` is a bit of an odd name as a reader could assume this is the whole mmaped trace. call id `extra` ?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+  auto* packet = trace->add_packet();
+  auto* recovered_info = packet->set_recovered_trace_info();
```
primiano: just extra->add_packet()->set_recovered_trace_info()?
I don't htink you need packet anywhere else right?

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -237,9 +286,253 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
...
+          pending.file_name.c_str());
+      continue;
+    }
+
+    base::Uuid uuid(config->trace_uuid_lsb(), config->trace_uuid_msb());
+    android_stats::MaybeLogUploadEvent(
+        PerfettoStatsdAtom::kRebootTraceRecovered, uuid.lsb(), uuid.msb());
+
+    size_t valid_offset =
+        SanitizeAndAnnotatePersistentTrace(fd, mmap, pending.file_name);
```
primiano: I think a better name would be TruncateAndAnnotatePersistentTrace

otherwise Sanitize feels like you might do some sort of filtering.

primiano (approved):
LGTM with few final comments

protos/perfetto/trace/android/recovered_trace_info.proto
safayat-google: Reason_ prefix is added.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Removed the log.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Updated to `500ms`

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Updated with the suggested code.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: comment added

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Removed the extra log

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Confirming that this is correct.\
Also updated the unit tests to verify that all the completed packets are recovered by this function.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: renamed as `extra`

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Renamed to `TruncateAndAnnotatePersistentTrace`

PR #6935, opened 2026-07-31
perfetto_cmd: fix typo and usage formatting in trigger_perfetto

safayat-google:


src/perfetto_cmd/trigger_perfetto.cc
```diff
@@ -82,7 +82,7 @@ int PERFETTO_EXPORT_ENTRYPOINT TriggerPerfettoMain(int argc, char** argv) {
     triggers_to_activate.push_back(std::string(argv[i]));
 
   if (triggers_to_activate.empty()) {
-    PERFETTO_ELOG("At least one trigger must the specified.");
+    PERFETTO_ELOG("At least one trigger must be specified.");
     return PrintUsage(argv[0]);
```
sashwinbalaji, line 86: lgtm on the typo

but changing elog to printf though makes sense but it would split the error logs now like it will be now:

```
PERFETTO_ELOG("At least one trigger must be specified.");
  fprintf(stderr, R"(
Usage: %s TRIGGER...
  -h|--help  Show this message
)",
          argv0);
```

I think would be better to keep both of them consistent so let the printusage be elog only ?

sashwinbalaji (approved):

PR #6938, opened 2026-08-01
[API] Fix build for windows ubsan client api

etiennep-chromium:
I'm trying to build angle with perfetto client API: https://crrev.com/c/7295789
Ubsan bot complains about
1- `undefined symbol: public: static struct perfetto::Track const perfetto::internal::TrackEventInternal::kDefaultTrack`
and many more
Fix : move default_deps from public_deps to deps

2- `undefined symbol: typeinfo for perfetto::internal::TracingTLS`
Fix: Add PERFETTO_EXPORT_COMPONENT on TracingTLS.


gn/proto_library.gni
```diff
@@ -134,7 +134,7 @@ template("protozero_cpp_library") {
     generator_plugin_label =
         perfetto_root_path + "src/protozero/protoc_plugin:cppgen_plugin"
     generator_plugin_suffix = ".gen"
-    if (build_with_chromium) {
+    if (build_with_chromium || perfetto_build_with_embedder) {
```
sashwinbalaji: nit: perfetto_build_with_embedder is a superset of build_with_chromium so this can be just perfetto_build_with_embedder

https://crsrc.org/c/third_party/perfetto/gn/perfetto.gni;l=34-39

sashwinbalaji (approved):

gn/proto_library.gni
etiennep-chromium: Done.
I also had to ignore `component_build_force_source_set` in gn/standalone/proto_library.gni for standalone build where `perfetto_build_with_embedder = true`.
This works in chromium's library (v8, webrtc, angle) because they all use chromium's proto_library (and already support `component_build_force_source_set`).
This will work with embedder using perfetto's proto_library too.
But this would possibly break an embedder that defines (and overrides) a custom proto_library that doesn't support component_build_force_source_set.

etiennep-chromium: Actually I changed to `build_with_chromium || (perfetto_build_with_embedder && !is_perfetto_build_generator)`

This seems what we already do elsewhere for the same reason:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/perfetto/gn/protozero_descriptor_diff.gni;l=18?q=build_with_chromium%20file:perfetto%20-file:angle%20-file:dawn%20-file:crossbench%20-file:openscreen&ss=chromium

sashwinbalaji (approved):

PR #6939, opened 2026-08-02
ui: add "Copy main thread tracks to workspace" command

zezeozue:
Copy every process's main-thread (thread.is_main_thread) thread_state and slice tracks into a workspace, grouped under their process.

### DIFF
--- a/ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
+++ b/ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
@@ -381,8 +381,13 @@ export default class TrackUtilsPlugin implements PerfettoPlugin {
     ctx.commands.registerCommand({
       id: 'dev.perfetto.CopyMainThreadTracksToWorkspace',
       name: 'Copy main thread tracks to workspace',
-      callback: (workspaceNameArg: unknown) =>
-        copyMainThreadTracksToWorkspace(ctx, workspaceNameArg),
+      callback: (workspaceNameArg: unknown) => {
+        const workspaceName =
+          typeof workspaceNameArg === 'string'
+            ? workspaceNameArg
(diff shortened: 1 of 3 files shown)


stevegolton (changes requested):
We already pull out `is_main_thread` when adding these tracks in the first place, we use them to create the chips in the timeline (see screenshot).

Image

However we should also add isMainThread tags to the track so that we can use these tags to filter on them later, then this command can just search for tracks with these tags synchronously rather than running another query.

zezeozue:
> We already pull out `is_main_thread` when adding these tracks in the first place, we use them to create the chips in the timeline (see screenshot).
> 
> Image
> However we should also add isMainThread tags to the track so that we can use these tags to filter on them later, then this command can just search for tracks with these tags synchronously rather than running another query.

Thanks for the tip

ui/src/core_plugins/dev.perfetto.TracklUtils/index.ts
```diff
@@ -673,3 +680,24 @@ async function resolveTracksFromSliceQuery(
     .map((event) => ctx.currentWorkspace.getTrackByUri(event.trackUri))
     .filter((track) => track !== undefined);
 }
+
+async function copyMainThreadTracksToWorkspace(
+  ctx: Trace,
+  workspaceNameArg: unknown,
+): Promise<void> {
+  const workspaceName =
+    typeof workspaceNameArg === 'string' ? workspaceNameArg : 'Main threads';
```
stevegolton: small nit - maybe pull the argument parsing from unknown -> string into the command callaback - it's a bit odd to have a naked function with command specific arg parsing.

Also I don't think this needs to be aync any more right?

stevegolton (approved):
LGTM!

PR #6940, opened 2026-08-02
ui: add "Open in default workspace" button to slice details

zezeozue:
Show a button in the slice details panel that switches to the default workspace and scrolls to the selected slice there. It appears only when the selection is in a non-default workspace and the track also exists in the default workspace.

### DIFF
--- a/ui/src/components/details/thread_slice_details_tab.ts
+++ b/ui/src/components/details/thread_slice_details_tab.ts
@@ -285,10 +285,7 @@ export class ThreadSliceDetailsPanel implements TrackEventDetailsPanel {
       {
         title: 'Slice',
         description: slice.name,
-        buttons: [
-          this.renderOpenInDefaultWorkspaceButton(),
-          this.renderContextButton(slice),
-        ],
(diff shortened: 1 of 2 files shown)


stevegolton (changes requested):
Can you find a way to do this more generically? E.g. if you're not on the main workspace, and your selected track exists on a the main workspace, add a button to all details panels.

I think it might be awkward to inject a button into all details panels, but if you can think of a nice way to do it that could be a much neater solution.

The UI really suffers from inconsistent features like this that are implemented for some tracks/plugins but not others.

zezeozue:
> Can you find a way to do this more generically? E.g. if you're not on the main workspace, and your selected track exists on a the main workspace, add a button to all details panels.
> 
> I think it might be awkward to inject a button into all details panels, but if you can think of a nice way to do it that could be a much neater solution.
> 
> The UI really suffers from inconsistent features like this that are implemented for some tracks/plugins but not others.

Lmk what you think of the current approach. Seems to work fine for slices, thread_state, counters

PR #6942, opened 2026-08-02
tp: rebuild flamegraphs from typed trees

LalitMaganti:
Consume the core Tree representation directly and preserve typed values while
merging top-down, bottom-up, pivot, and filtered flamegraph views.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/src/trace_processor/core/tree/tree_column_ops.h
+++ b/src/trace_processor/core/tree/tree_column_ops.h
@@ -33,13 +33,6 @@ void UpdateRowHashes(const Tree::Column&,
 // its type and nullability.
 Tree::Column Gather(const Tree::Column&, Span<const uint32_t> rows);
 
-// Concatenates selected rows from two same-typed columns into a new dense
-// column while preserving nullability.
-Tree::Column GatherConcat(const Tree::Column& first,
-                          Span<const uint32_t> first_rows,
-                          const Tree::Column& second,
-                          Span<const uint32_t> second_rows);
-
 }  // namespace perfetto::trace_processor::core::tree_ops
 
 #endif  // SRC_TRACE_PROCESSOR_CORE_TREE_TREE_COLUMN_OPS_H_
--- a/src/trace_processor/plugins/flamegraph/flamegraph.cc
+++ b/src/trace_processor/plugins/flamegraph/flamegraph.cc
@@ -17,670 +17,1618 @@
 #include "src/trace_processor/plugins/flamegraph/flamegraph.h"
 
 #include <algorithm>
-#include <cmath>
+#include <cinttypes>
+#include <cstddef>
 #include <cstdint>
 #include <cstring>
 #include <string>
-#include <string_view>
+#include <tuple>
+#include <type_traits>
 #include <utility>
 #include <vector>
 
 #include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
+#include "perfetto/ext/base/endian.h"
 #include "perfetto/ext/base/flat_hash_map.h"
-#include "perfetto/ext/base/regex.h"
+#include "perfetto/ext/base/hash.h"
+#include "perfetto/ext/base/murmur_hash.h"
 #include "perfetto/ext/base/status_macros.h"
 #include "perfetto/ext/base/status_or.h"
-#include "src/trace_processor/core/dataframe/adhoc_dataframe_builder.h"
-#include "src/trace_processor/core/dataframe/dataframe.h"
+#include "perfetto/ext/base/string_utils.h"
+#include "perfetto/ext/base/utils.h"
+#include "src/trace_processor/core/tree/tree_column_ops.h"
+#include "src/trace_processor/core/tree/tree_path_interner.h"
+#include "src/trace_processor/core/util/bit_vector.h"
 #include "src/trace_processor/core/util/flex_vector.h"
+#include "src/trace_processor/core/util/slab.h"
+#include "src/trace_processor/core/util/sort.h"
 
 namespace perfetto::trace_processor::flamegraph {
 namespace {
 
-using core::FlexVector;
+// A bit outside the SHOW_STACK mask. Once set, equality with the required
+// SHOW_STACK mask is impossible, rejecting this path and all its descendants.
+constexpr uint64_t kHideStackBit = uint64_t{1} << 63;
 
-// Show-stack and show-from filters are tracked as per-frame bitmasks, one
-// bit per filter, ORed along each path: a stack satisfies them when every
-// bit is present.
-constexpr size_t kMaxBitFilters = 63;
-
-// Row helpers for the flattened metric columns; explicit loops so they
-// inline instead of dispatching into libc for a handful of bytes.
-void AddRow(double* dst, const double* src, uint32_t k) {
-  for (uint32_t m = 0; m < k; ++m) {
-    dst[m] += src[m];
-  }
-}
-
-void CopyRow(double* dst, const double* src, uint32_t k) {
-  for (uint32_t m = 0; m < k; ++m) {
-    dst[m] = src[m];
-  }
-}
-
-bool AnyNonZero(const double* row, uint32_t k) {
-  for (uint32_t m = 0; m < k; ++m) {
-    if (row[m] < 0 || row[m] > 0) {
-      return true;
-    }
-  }
-  return false;
-}
-
-// The compiled filter regexes. The pivot pattern is an extra show-stack
-// filter (only pivot-containing stacks are shown), remembered by index so
-// pivot frames can be identified from the match bitmask.
-struct Filters {
-  std::vector<base::Regex> show_stack;
-  std::vector<base::Regex> hide_stack;
-  std::vector<base::Regex> show_from;
-  std::vector<base::Regex> hide_frame;
-  uint32_t pivot_idx = 0;
-  uint64_t show_stack_mask = 0;
-  uint64_t show_from_mask = 0;
-  // A bit outside the mask: assigning it to a frame's bits makes its
-  // stacks unable to ever satisfy the mask, implementing hide-stack.
-  uint64_t impossible_bits = 0;
-
-  bool any() const {
-    return !show_stack.empty() || !hide_stack.empty() || !show_from.empty() ||
-           !hide_frame.empty();
-  }
-};
-
-base::StatusOr<Filters> CompileFilters(const Config& config) {
-  Filters f;
-  for (const auto& spec : config.filters) {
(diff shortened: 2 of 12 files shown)


src/trace_processor/core/tree/tree_column_ops.h
```diff
@@ -25,6 +25,9 @@
 
 namespace perfetto::trace_processor::core::tree_ops {
 
+// Creates a null-typed column of |rows| rows: no payload, every row null.
+Tree::Column NullColumn(uint32_t rows);
```
safayat-google: nit: move this inside `Tree::Column` as a static method?

src/trace_processor/plugins/flamegraph/flamegraph_benchmark.cc
```diff
@@ -39,254 +38,125 @@ struct Rng {
...
-}
-BENCHMARK(BM_FlamegraphTreeTopDownRows)->Arg(100000)->Arg(1000000);
-
-void BM_FlamegraphTreeTopDownPropertiesRows(benchmark::State& state) {
-  StringPool pool;
-  BenchInput in =
-      GenForest(&pool, static_cast<uint32_t>(state.range(0)), 1, true);
-  RunWithRows(state, in, Config{}, &pool);
+void BM_FlamegraphBottomUpAggregate(benchmark::State& state) {
+  Run(state, Config::View(Config::BottomUp{}), false, false, true);
```
safayat-google: nit: `false, false, true` are difficult to read. consider adding a `struct RunOptions {}` where those values are default false.

```
Run(state, Config::View(Config::BottomUp{}), { .aggregate = true });
```

src/trace_processor/plugins/flamegraph/flamegraph_unittest.cc
```diff
@@ -16,511 +16,257 @@
...
+  core::Tree input =
+      MakeTree(&pool, {{"main", std::nullopt, 1}, {"hide", 0, 2}, {"b", 1, 4}});
+  Config config = MakeConfig(input, pool);
+  config.hide_frame_filters.push_back(base::Regex::CreateOrCheck("hide"));
+  auto result = Build(std::move(input), config);
+  ASSERT_TRUE(result.ok());
+  ASSERT_EQ(result->row_count, 2u);
+  EXPECT_EQ(Value<int64_t>(*result, "self_value", 0), 3);
+  EXPECT_EQ(Value<int64_t>(*result, "cumulative_value", 0), 7);
 }
```
safayat-google, line 370: nit: add tests to cover `Pivot` views, `FromFrame`/`show_stack_filters`/`hide_stack_filters`?

src/trace_processor/plugins/flamegraph/flamegraph.cc
```diff
@@ -17,670 +17,1359 @@
...
+  const core::Span<const uint32_t> packed_sources = packed.source_node_span();
+  const bool keep_all =
+      packed.size() == downward_.tree.size() + upward_.tree.size();
+  const auto take_or_gather = [&](core::Tree::Column& column) {
+    return keep_all ? std::move(column)
+                    : core::tree_ops::Gather(column, packed_sources);
+  };
+  for (uint32_t i = 0; i < config_.value_columns.size(); ++i) {
+    const std::string name(input_.ColumnName(config_.value_columns[i]));
+    output.names.push_back("self_" + name);
```
safayat-google, line 1473: nit: should we use the same column naming. not sure if this will break somewhere if we have hard coded column name,
```
- columns.push_back("selfValue" + suffix);
- columns.push_back("cumulativeValue" + suffix);
```

src/trace_processor/plugins/flamegraph/flamegraph.cc
```diff
@@ -17,670 +17,1359 @@
...
+base::Status FlamegraphBuilder::MarkActiveBottomUpAnchors(
+    const core::Tree::Column& column,
+    core::Span<uint8_t> active) {
+  core::Span<const T> values = column.unchecked_span<T>();
+  for (uint32_t row = 0; row < input_.row_count; ++row) {
+    if (column.null_bv.size() > 0 && !column.null_bv.is_set(row)) {
+      continue;
+    }
+    const T value = values[row];
+    if (value < 0) {
```
safayat-google, line 923: nit: check against epsilon for `double`. also on Line 855

src/trace_processor/plugins/flamegraph/flamegraph.cc
```diff
@@ -17,670 +17,1359 @@
...
+// Lays out siblings widest first: a stable radix sort orders all nodes by
+// descending width, and distributing that order into per-parent buckets
+// keeps it within every sibling list. A breadth-first walk then packs each
+// node's children left-to-right from the node's own position; the walk
+// order is the render order.
+Layout ComputeLayout(const core::Tree& tree,
+                     const core::Tree::Column& cumulative,
+                     const core::Tree::Column& depth) {
+  const uint32_t row_count = tree.row_count;
+  const core::Span<const int64_t> depths = depth.unchecked_span<int64_t>();
```
safayat-google, line 1508: nit: early return for 0 rows?

safayat-google (commented):
Overall LGTM but I will have another look before approving since it's a large change.
left some nits.

src/trace_processor/plugins/flamegraph/flamegraph.cc
LalitMaganti: We don't need to do that for zero, we don't care about closeness, non-negative is purely for making sure it's properly additive.

src/trace_processor/plugins/flamegraph/flamegraph.cc
LalitMaganti: No this is acutally totally fine. Note that this is all *only* going to be used by the UI behind the scenes, it's not (and is not going to be) a public API.

src/trace_processor/plugins/flamegraph/flamegraph_unittest.cc
```diff
@@ -16,511 +16,321 @@
...
-  TreeSummary s = Summarize(fg.down, f);
-  ASSERT_EQ(s.size(), 2u);
-  EXPECT_EQ(std::get<1>(s["main"]), std::vector<double>{5});
-  EXPECT_EQ(std::get<1>(s["main/ok"]), std::vector<double>{4});
+TEST(FlamegraphTest, StringAggregates) {
+  StringPool pool;
+  core::Tree input =
+      MakeTree(&pool, {{"main", std::nullopt, 1}, {"x", 0, 2}, {"x", 0, 3}});
+  AddStringColumn(&input, &pool, "source", {"root", "a", "b"});
+  input.names.emplace_back("num");
```
safayat-google: nit: add a function similar to `AddStringColumn` to do this two line at once `AddValueColum(name, vector<value_type>)`. or/also a generic one `AddColumn(name, Tree::Column)`

src/trace_processor/plugins/flamegraph/flamegraph.cc
```diff
@@ -17,670 +17,1363 @@
...
+  UpwardHalf upward_;
+  std::vector<MetricColumns> metrics_;
+  std::vector<core::Tree::Column> aggregate_columns_;
+};
+
+bool IsValidConfig(const Config& config) {
+  if (!config.name || !config.name->type.Is<core::String>()) {
+    return false;
+  }
+  if (config.value_columns.empty()) {
```
safayat-google: nit: also verify that individual value columns are not empty / non-numeric?

src/trace_processor/plugins/flamegraph/flamegraph.cc
```diff
@@ -17,670 +17,1363 @@
...
+  const uint32_t downward_size = downward_.tree.size();
+  std::vector<State> merged(downward_size + upward_.tree.size());
+  if (config_.view.IsAnyOf<Config::DownwardViews>()) {
+    for (uint32_t frame = 0; frame < input_.row_count; ++frame) {
+      const PathState& state = analysis_.states[frame];
+      if (!state.IsRetained(frame) || !retained_states[frame].has_value() ||
+          !ContributesToDownwardCumulative(frame)) {
+        continue;
       }
+      merged[downward_.node[frame]].Merge(retained_states[frame], config_.pool);
```
safayat-google: nits, for `10,000` unique string rows, temporary allocation+deallocation all these strings looks a little waste to finally get a string like 'x an 9999 others'. If there's no such existing needs maybe we can optimise memory usage here.

src/trace_processor/plugins/flamegraph/flamegraph.cc
```diff
@@ -17,670 +17,1363 @@
...
+                    : core::tree_ops::Gather(column, packed_sources);
+  };
+  for (uint32_t i = 0; i < config_.value_columns.size(); ++i) {
+    const std::string name(input_.ColumnName(config_.value_columns[i]));
+    output.names.push_back("self_" + name);
+    output.columns.push_back(take_or_gather(metrics_[i].self));
+    output.names.push_back("cumulative_" + name);
+    output.columns.push_back(take_or_gather(metrics_[i].cumulative));
+  }
+  for (uint32_t i = 0; i < config_.aggregate_columns.size(); ++i) {
```
safayat-google, line 1478: nit: check for duplicate column names?

src/trace_processor/plugins/flamegraph/flamegraph.cc
```diff
@@ -17,670 +17,1363 @@
...
+  core::Slab<uint64_t> keys = core::Slab<uint64_t>::Alloc(row_count);
+  core::Slab<uint32_t> by_width = core::Slab<uint32_t>::Alloc(row_count);
+  for (uint32_t node = 0; node < row_count; ++node) {
+    uint64_t bits;
+    memcpy(&bits, &width[node], sizeof(bits));
+    keys[node] = base::HostToBE64(~bits);
+    by_width[node] = node;
+  }
+  core::Slab<uint32_t> scratch = core::Slab<uint32_t>::Alloc(row_count);
+  core::Slab<uint32_t> radix_counts = core::Slab<uint32_t>::Alloc(1u << 16);
```
safayat-google, line 1535: nit: for a few number of rows can we use a different sort to avoid allocating this 256kb histogram buffer?

safayat-google (approved):

src/trace_processor/plugins/flamegraph/flamegraph.cc
LalitMaganti: No radix sort is too efficeint (and good for binary size) to give this up.

PR #6944, opened 2026-08-02
tp: add reusable tree transformation helpers

LalitMaganti:


src/trace_processor/core/tree/tree.h
```diff
@@ -58,6 +63,23 @@ struct Tree {
...
+    auto it = std::find(names.begin(), names.end(), find_name);
+    if (it == names.end()) {
+      return std::nullopt;
+    }
+    return columns.data() + std::distance(names.begin(), it);
+  }
+
+  // Returns the name associated with a column in this tree.
+  std::string_view ColumnName(const Column* column) const {
+    const auto index = static_cast<size_t>(column - columns.data());
```
safayat-google: > column - columns.data()

nits, for consistency use `std::distance` similar to the previous function?

src/trace_processor/core/util/span.h
```diff
@@ -42,6 +42,9 @@ struct Span {
   size_t size() const { return static_cast<size_t>(e - b); }
   bool empty() const { return b == e; }
   T& operator[](size_t index) const { return b[index]; }
+  Span<T> subspan(size_t offset, size_t count) const {
+    return Span<T>(b + offset, b + offset + count);
```
safayat-google, line 50: nits, add bound check?

src/trace_processor/core/tree/tree.h
```diff
@@ -89,6 +94,23 @@ struct Tree {
     BitVector null_bv;  // non-empty for nullable columns
```
safayat-google, line 94: sorry, I missed this before.

The name looks a little confusing `null_bv[x] = true` sounds like value at index x is null while it's the opposite. Should we rename with something like `non_null_bv` / `valid_bv` or `validity_bv`?

src/trace_processor/core/tree/tree_column_ops.cc
```diff
@@ -0,0 +1,149 @@
...
+}
+
+Tree::Column Gather(const Tree::Column& input, Span<const uint32_t> rows) {
+  if (input.type.Is<Int64>()) {
+    return GatherTyped<int64_t>(input, rows);
+  }
+  if (input.type.Is<Double>()) {
+    return GatherTyped<double>(input, rows);
+  }
+  return GatherTyped<StringPool::Id>(input, rows);
```
safayat-google: nits, explicitly validate the type before falling back to `StringPool::Id`? also on the next function.

safayat-google (approved):

src/trace_processor/core/tree/tree.h
LalitMaganti: Keeping the name as-is: it matches core::dataframe's null_bv, for consistency across the two modules.

PR #6947, opened 2026-08-02
tp: enforce parent-before-child Tree ordering

LalitMaganti:
Reorder arbitrary DataFrame input while constructing Trees so every parent
precedes its children. Preserve the fast path for already ordered input and
detect cycles while computing the reordered traversal.


LalitMaganti:

PR #6949, opened 2026-08-02
tp: rebuild flamegraphs from typed trees

LalitMaganti:
Consume the core Tree representation directly and preserve typed values while
merging top-down, bottom-up, pivot, and filtered flamegraph views.


LalitMaganti:

PR #6950, opened 2026-08-02
tp: expose flamegraphs through PerfettoSQL

LalitMaganti:
Expose the typed flamegraph builder as a create-only virtual table. The
operator ingests its immutable source once and accepts view, filter, metric,
grouping, and aggregation configuration at query time.

Add the stdlib layout macro and end-to-end coverage for all views, regex
flags, metrics, and folded aggregate modes.


LalitMaganti:

PR #6951, opened 2026-08-02
ui: use persistent flamegraph operators

LalitMaganti:
Replace the seven-macro viz.flamegraph pipeline with a persistent virtual
table per metric. Each operator materializes its source once and receives
views, filters, grouping columns, and aggregate modes through its call-time
configuration.

The old viz.flamegraph module is no longer used and is deleted.


LalitMaganti:

PR #6956, opened 2026-08-03
[DNS] ai: add workflow for debugging slow PerfettoSQL queries

### LINKED ISSUE #6591: Add workflow for debugging slow queries to AI skill
We already have a (out of date) doc internally for this. We need to take this approach and essentially modernize it for e.g. create perfetto table and then make it a full e2e workflow.

LalitMaganti:
Agents that hit a slow query today either give up, silently simplify
the analytical intent, or re-run one-shot trace_processor invocations
that re-pay trace parsing on every attempt.

Add a workflow with a measure -> diagnose -> fix loop: time queries
against a long-running server so parse time is excluded, read EXPLAIN
QUERY PLAN and relation sizes to find exploding joins, then apply fixes
in order of leverage (filter early, materialize with CREATE PERFETTO
TABLE, index with CREATE PERFETTO INDEX, purpose-built interval
operators instead of inequality joins, GLOB over LIKE), re-measuring
after each change.

Fixes #6591.

### DIFF
--- a/ai/skills/perfetto/SKILL-template.md
+++ b/ai/skills/perfetto/SKILL-template.md
@@ -75,6 +75,10 @@ If you want to load a trace and write custom PerfettoSQL queries:
 *   Read [querying.md]($SKILL_ROOT/infra-references/querying.md) to learn about running
     one-shot queries, using the long-running RPC mode, discovering schemas, and
     writing efficient PerfettoSQL.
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Do not submit: this is pretty poor quality advice, needs to be signifcantly reworked.

Issue #6927, opened 2026-07-31
Improving error handling on traceconv bundle


LalitMaganti:
See https://github.com/google/perfetto/issues/6383#issuecomment-5141756157

We should really drastically improve our error messages to prevent mistakess like this and guide the user better.

LalitMaganti:
Fixed by https://github.com/google/perfetto/pull/6969

PR #6969, opened 2026-08-03
tp: overhaul error handling and diagnostics for the shell and bundle

LalitMaganti:
Issue: https://github.com/google/perfetto/issues/6927

`trace_processor bundle` (formerly `traceconv bundle`) gave cryptic or
misleading feedback when things went wrong:

- Traces with nothing to enrich (e.g. a pure atrace trace, or a
  function_graph trace recorded **with** `symbolize_ksyms`) failed with a
  bare `bundle: failed to create bundle.` and no explanation.
- A function_graph trace recorded **without** `symbolize_ksyms` got no
  feedback at all, even though its kernel addresses can never be
  symbolized offline.
- Unwritable output paths (missing parent dir, read-only, path-is-a-dir)
  crashed the whole process with a `PERFETTO_CHECK`.
- Extra positional arguments were silently ignored/misparsed (e.g. passing
  each `--symbol-paths` directory as a separate argument).
- Conversion failures produced double error messages (an internal
  `PERFETTO_ELOG` + a generic "conversion failed" wrapper), and several
  paths crashed with `PERFETTO_FATAL` on ordinary user error.

## Changes

- **bundle** now always produces the bundle and explains what could not be
  enriched, with actionable hints. It detects function_graph traces
  recorded without `symbolize_ksyms`, fails gracefully (instead of
  crashing) on unwritable output paths, and reports reading/enrichment
  progress.
- **CLI**: every fixed-arity subcommand (`bundle`, `convert`, `util`,
  `export`, `query`, `server`, `metrics`) now rejects extra positional
  arguments with a hint, instead of silently misinterpreting them.
- **traceconv**: the conversion entry points (`TraceToJson`,
  `TraceToSystrace`, `TraceToText`, `TraceToProfile`,
  `TraceToFirefoxProfile`, `UnpackCompressedPackets`, `SymbolizeProfile`,
  `DeobfuscateProfile`) now return descriptive `base::Status` errors
  instead of int/bool returns paired with internal `PERFETTO_ELOG`/`FATAL`,
  so failures are reported once with the actual reason.
- **TarWriter**: fails gracefully when the output path cannot be opened
  (also fixes the same crash in `util merge`).
- **Parse errors**: no longer append a redundant "The trace file is
  corrupt." sentence; the UI error-dialog extraction is updated for the
  new message format.

### DIFF
--- a/src/trace_processor/read_trace_internal.cc
+++ b/src/trace_processor/read_trace_internal.cc
@@ -591,8 +591,10 @@ base::Status ReadTraceUnfinalized(
 #endif  // PERFETTO_HAS_MMAP()
   if (bytes_read == 0) {
     base::ScopedFile fd(base::OpenFile(filename, O_RDONLY));
-    if (!fd)
-      return base::ErrStatus("Could not open trace file (path: %s)", filename);
+    if (!fd) {
+      return base::ErrStatus("could not open trace file (errno: %d, %s)", errno,
+                             strerror(errno));
+    }
     RETURN_IF_ERROR(
         ReadTraceUsingRead(tp, *fd, &bytes_read, progress_callback));
   }
--- a/src/trace_processor/read_trace_internal.cc
+++ b/src/trace_processor/read_trace_internal.cc
@@ -592,8 +592,8 @@ base::Status ReadTraceUnfinalized(
   if (bytes_read == 0) {
     base::ScopedFile fd(base::OpenFile(filename, O_RDONLY));
     if (!fd) {
-      return base::ErrStatus("could not open trace file (errno: %d, %s)", errno,
-                             strerror(errno));
+      return base::ErrStatus("could not open trace file %s (errno: %d, %s)",
+                             filename, errno, strerror(errno));
     }
     RETURN_IF_ERROR(
         ReadTraceUsingRead(tp, *fd, &bytes_read, progress_callback));
--- a/src/trace_processor/shell/bundle_integrationtest.cc
+++ b/src/trace_processor/shell/bundle_integrationtest.cc
@@ -100,7 +100,7 @@ std::map<std::string, std::string> ReadTarMembers(const std::string& path) {
   return out;
 }
 
-class TraceconvBundleTest : public ::testing::Test {
+class TraceconvShellBundleTest : public ::testing::Test {
  protected:
   void SetUp() override {
     input_trace_ = base::GetTestDataPath(
@@ -137,7 +137,7 @@ class TraceconvBundleTest : public ::testing::Test {
 // The bundle should contain the unmodified input trace plus a parseable
 // deobfuscation.pb whose `deobfuscation_mapping` reflects the supplied
 // mapping.txt (package, class, method names).
-TEST_F(TraceconvBundleTest, BundleWithProguardMap) {
+TEST_F(TraceconvShellBundleTest, BundleWithProguardMap) {
   base::TempFile mapping = WriteTempFile(
       "com.example.Foo -> a.a:\n"
       "    void bar() -> b\n");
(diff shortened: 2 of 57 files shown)


src/trace_processor/shell/bundle_integrationtest.cc
```diff
@@ -271,5 +282,302 @@ TEST_F(TraceconvBundleTest, BundleNoAutoProguardMapsWithExplicit) {
...
+  ArgvInvoker invoker;
+  invoker.Add("trace_processor_shell");
+  invoker.Add("bundle");
+  // Note: no --no-auto-symbol-paths here, so the symbolizer searches the
+  // mappings' absolute paths (which don't exist in this environment) and
+  // reports them as "no matching symbols".
+  invoker.Add(input_trace_);
+  invoker.Add(output_path_);
+
+  ScopedStderrCapture capture;
```
sashwinbalaji, line 458: this will fail on windows right as GTEST_SKIP is runtime ?

src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -394,17 +392,26 @@ std::map<std::string, FoundBinary> BuildIdIndex(
...
+    WalkDirectories(std::move(dirs), [&result, &corrupt_file_count](
+                                         const char* fname, size_t size) {
+      ProcessBinaryFile(fname, size, result, &corrupt_file_count);
     });
   }
 
   // Process individual files
   for (const std::string& file_path : files) {
-    ProcessBinaryFile(file_path.c_str(), 0, result);
+    ProcessBinaryFile(file_path.c_str(), 0, result, &corrupt_file_count);
```
sashwinbalaji: for my understanding, won't this call always  log corrupt ?

```
-> ProcessBinaryFile(file_path.c_str(), 0, result, &corrupt_file_count);
      ->   std::optional<BinaryInfo> binary_info = GetBinaryInfo(fname, size);
   if (!binary_info) {
    // Passed the magic check but failed to parse: corrupt or truncated.
    ++(*corrupt_file_count);
    return;
  }
       -> std::optional<BinaryInfo> GetBinaryInfo(const char* fname, size_t size) {
  static_assert(EI_CLASS > EI_MAG3, "mem[EI_MAG?] accesses are in range.");
  if (size <= EI_CLASS) {
    return std::nullopt;
  }

```

src/traceconv/trace_to_profile.cc
```diff
@@ -210,30 +220,35 @@ int TraceToProfile(std::istream* input,
...
+    return base::OkStatus();
   }
 
   // Write profiles to files.
-  std::string dst_dir = GetDestinationDirectory(output_dir, dir_prefix);
+  ASSIGN_OR_RETURN(std::string dst_dir,
+                   GetDestinationDirectory(output_dir, dir_prefix));
   for (const auto& profile : profiles) {
     std::string filename = dst_dir + "/" + filename_fn(profile);
     base::ScopedFile fd(base::OpenFile(filename, O_CREAT | O_WRONLY, 0700));
```
sashwinbalaji: should it be O_TRUNC also considering similar issue we hit recently b/537440297

src/traceconv/trace_to_systrace.cc
```diff
@@ -131,6 +131,7 @@ class QueryWriter {
       }
       global_writer_.AppendStringView(line_writer.GetStringView());
     }
+    EndProgressLine();
```
sashwinbalaji, line 134: nit: this might print too many empty lines ?

src/trace_processor/read_trace_internal.cc
```diff
@@ -591,8 +591,10 @@ base::Status ReadTraceUnfinalized(
 #endif  // PERFETTO_HAS_MMAP()
   if (bytes_read == 0) {
     base::ScopedFile fd(base::OpenFile(filename, O_RDONLY));
-    if (!fd)
-      return base::ErrStatus("Could not open trace file (path: %s)", filename);
+    if (!fd) {
+      return base::ErrStatus("could not open trace file (errno: %d, %s)", errno,
```
sashwinbalaji: nit: keep printing the filename

sashwinbalaji (approved):

PR #6973, opened 2026-08-03
tp: partially revert "tp: remove the standalone traceconv binary"

sashwinbalaji:
Partial revert of https://github.com/google/perfetto/commit/2d5a0c77f71d053501ea8d6b655f7546e4ed7a82 (https://github.com/google/perfetto/pull/6410).

Downstream build targets still depend on the standalone traceconv
binary (e.g. //third_party/perfetto:traceconv in google3, which is
breaking the autoroll), so restore it while those callers migrate to
trace_processor_shell.

Reverted:
 - src/traceconv/{main,traceconv}.cc, include/perfetto/ext/traceconv
   and the executable/main/traceconv_lib GN targets. src/traceconv:lib
   is now linked by both traceconv and trace_processor_shell.
 - src/traceconv/wasm_main.cc is dropped; the UI's traceconv WASM
   module goes back to being built from traceconv.cc.
 - The bundle CLI test moves back to
   src/traceconv/traceconv_bundle_integrationtest.cc.
 - Release plumbing: the LUCI recipe, package-github-release-artifacts,
   gen_android_bp and gen_bazel, plus the regenerated Android.bp and
   bazel BUILD.

Not reverted:
 - python/tools/{cpu_profile,heap_profile}.py and the amalgamated
   tools/{cpu_profile,heap_profile}. https://github.com/google/perfetto/pull/6410 pointed these back at the
   traceconv prebuilt and its flat MODE arguments, which fixes a real
   issue independent of this revert: both manifests are pinned at the
   v57.2 release, and the v57.2 trace_processor_shell prebuilt predates
   the convert/util subcommands, so `trace_processor_shell util
   symbolize` fails against the binary the scripts download.

The traceconv-name compat shim in src/trace_processor/shell is left in
place: it came from https://github.com/google/perfetto/pull/6400, not https://github.com/google/perfetto/pull/6410, so trace_processor_shell keeps
accepting traceconv-style command lines when invoked under that name.


python/tools/cpu_profile.py
```diff
@@ -31,7 +31,7 @@
 import time
```
LalitMaganti, line 31: Please remove this + heap_profile.py +heap_profile + cpu_profile from this patch. They are fixing real issue.

LalitMaganti (approved):

PR #6975, opened 2026-08-03
Fix dangling reference in UnwindingMetadata::GetBuildId

Jake-The-Human:
Modify UnwindingMetadata::GetBuildId to return std::string by value instead of const std::string&.

The underlying libunwindstack function frame.map_info->GetBuildID() returns SharedString by value. Returning a const reference to the string object resulted in a dangling reference to a destroyed temporary, causing undefined behavior.

This change also removes the unused `empty_string_` member variable from UnwindingMetadata.


sashwinbalaji (approved):

safayat-google (approved):
LGTM.

Seems like initially we had return by value.
Then libunwinding updated their `GetBuildId` to return by reference. So we updated it. But again in [this aosp change](https://android.googlesource.com/platform/system/unwinding/+/b16492b1200f6dfcf7ca7928c0fe8b7d5240244d%5E%21/libunwindstack/MapInfo.cpp) they updated it to return `SharedString`. So, returning by value makes sense now.

PR #6977, opened 2026-08-03
ui: add contained-traces merge panel

zezeozue:
Add an area-selection tab to the ContainedTraces plugin that lists the contained ("other") traces under the selection in a grid (self flag, uuid link, name, track, duration) with per-row checkboxes, prepends the current trace as a "This trace" row (uuid from the trace_uuid metadata), and a "Merge selected traces" button that opens the checked uuids as go/trace-uuids/<uuids>.

### DIFF
--- a/ui/src/plugins/com.android.ContainedTraces/index.ts
+++ b/ui/src/plugins/com.android.ContainedTraces/index.ts
@@ -24,7 +24,7 @@ import {
 } from '../../trace_processor/query_result';
 import {SourceDataset, type Dataset} from '../../trace_processor/dataset';
 import SupportPlugin from '../com.android.AndroidLongBatterySupport';
-import {AsyncLimiter} from '../../base/async_limiter';
+import {AsyncMemo} from '../../base/async_memo';
 import {Duration, Time, type duration} from '../../base/time';
 import {DurationWidget} from '../../components/widgets/duration';
 import {selectTracksAndGetDataset} from '../../components/aggregation_adapter';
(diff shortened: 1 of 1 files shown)


zezeozue:
image

ui/src/plugins/com.android.ContainedTraces/index.ts
```diff
@@ -28,6 +50,171 @@ interface ContainedTrace {
...
+      this.previousSelection === undefined ||
+      !areaSelectionsEqual(this.previousSelection, selection)
+    ) {
+      this.previousSelection = selection;
+      this.loading = true;
+      this.limiter.schedule(async () => {
+        this.rows = await this.queryRows(dataset, selection);
+        this.selected = new Set(this.rows.map((r) => r.uuid));
+        this.loading = false;
+        m.redraw();
```
stevegolton: Shouldn't need this m.redraw()

ui/src/plugins/com.android.ContainedTraces/index.ts
```diff
@@ -28,6 +50,171 @@ interface ContainedTrace {
...
+      CONTAINED_TRACE_SPEC,
+    );
+    if (dataset === undefined) {
+      this.previousSelection = undefined;
+      return undefined;
+    }
+
+    if (
+      this.previousSelection === undefined ||
+      !areaSelectionsEqual(this.previousSelection, selection)
```
stevegolton: Could you use an AsyncMemo for this? It would kill the long lived member variables loading, data, selected, and previousSelection.

I think you'll need to pull out the important fields into the key to make it JSON serialisable. Don't spread as you'll pull in tracks as well which are not serialisable.

Something like this:

```ts
memo.use({
  key: {start: selection.start, end: selection.end, trackUris: selection.trackUris},
  compute: async () => {
    const rows = await this.queryRows(dataset, selection);
    const selected = new Set(this.rows.map((r) => r.uuid));
    return {rows, selected};
  },
});
```

stevegolton (changes requested):

stevegolton (approved):

PR #6979, opened 2026-08-03
Add multi-pass redactor support and implement process tree merging with it

edgararriagag:
The process tree merging achieves trace savings of 12.2% of redacted trace size on average.

Internal tracking bug b/539590970


src/trace_redaction/trace_redactor.cc
```diff
@@ -173,6 +185,81 @@ base::Status TraceRedactor::Transform(
   return base::OkStatus();
 }
 
+base::Status TraceRedactor::AugmentReduce(const Context* context,
```
LalitMaganti: Sorry but I don't understand this last but at all? Why are we loading the file and writing it out? The whole design of the trace refactor is to do all the processing in memory - you seem to be breaking that completely here?

edgararriagag: The trace redactor right now has several issue, here are some of the current key issues in the architecture which this change overcomes (I also tagged you in the design doc which goes into more detail)
* Current redactor architecture is a single pass architecture which modifies packets as it goes, this creates redundancies due to the redaction Transform phases. Those redundancies cannot be cleaned up because transform only looks at the current packet, it cannot look into the future (so it cannot have the context of the future)
* Context is global, so any context for a custom operation is shared with everything else, so that harms escalability because we pollute the context with things that are specific to an operation, so this better helps encapsulate by keeping local contexts in this type of primitives
* You cannot generate new packets as the goal of transforms is only to modify existing packets
* Having a second pass makes it easier to reason about as to which changes depend on operating on the redacted trace (post-processing it), and which changes are part of redaction.

How this change overcomes this:
* It adds local context to the AugmentReduce phases, so they won't pollute the global Context when operating
* Adds posibility to insert new trace packets (this is especially helpful when doing merging operations efficiently where we just collect, drop, and issue a new packet which is cleaner and more performant) 
* Allows to have a global vision of the redacted trace, it still operates sequentially like the first pass, but it has the possibility to Collect() based on the entire trace contents. 
* The added cost of this specific primitive along with the overhead of re-reading the trace and creating intermediate file, etc is small, it sits at about 1.27% so its not a significant increase in redaction time.

Now with all this context I will reply to the question:
>Why are we loading the file and writing it out?
This is required because the redactor streams the packets as it goes into a file, redactor currently does not keep the output trace in memory while redacting, it streams it out packet by packet

edgararriagag: I reworked the architecture, it should now be ready for review. It is now greatly simplified and now it should be more generic and it reuses existing arch for multi pass flows.

src/trace_redaction/trace_redactor.h
```diff
@@ -72,51 +71,118 @@ class TraceRedactor {
...
   base::Status Collect(Context* context,
                        const trace_processor::TraceBlobView& view) const;
 
+  // Runs all validators once on the context.
+  base::Status Validate(const Context& context) const;
+
   // Runs builders once.
-  //
-  // ```
-  //  with context:
```
LalitMaganti, line 99: These style of comments was very usefl. Why are we removing them?

src/trace_redaction/trace_redaction_framework.h
```diff
@@ -407,6 +434,29 @@ class TransformPrimitive {
...
+// Augment primitives are executed after the Transform phase, and all generated
+// packets are appended to the end of the trace.
+class AugmentPrimitive {
+ public:
+  virtual ~AugmentPrimitive();
+
+  // Generates a trace packet using data from the context. Called repeatedly
+  // once per trace until an empty string is returned. Each non-empty string
+  // returned should be a serialized TracePacket to be appended to the trace.
+  virtual std::string Augment(const Context& context) = 0;
```
LalitMaganti: You don't want to return a std::string on every time, this will churn allocations like crazy. There was a reason why we took packets as pointers in the past, it's exactly to avoid this sort of memory churn.

src/trace_redaction/trace_redactor.h
edgararriagag: They went away during a refactoring, no particular reason to get rid of them, though re-reading them they could be improved, so I brought them back and I also added extra information to explain the phase being executed.

src/trace_redaction/trace_redaction_framework.h
edgararriagag: Good catch, fixed.

src/trace_redaction/trace_redactor.cc
```diff
@@ -153,74 +168,134 @@ base::Status TraceRedactor::Transform(
...
+  for (size_t i = 0; i < passes_.size(); ++i) {
+    std::string next_buffer;
+    RETURN_IF_ERROR(passes_[i]->Redact(current_view, context, &next_buffer));
+    current_buffer = std::move(next_buffer);
+
+    // Only wrap the output in a TraceBlobView if there is a subsequent pass
+    // that needs to read it. The final pass writes directly from
+    // current_buffer.
+    if (i + 1 < passes_.size()) {
+      auto pass_blob = trace_processor::TraceBlob::CopyFrom(
```
LalitMaganti: This copy seems very suspicious. Why are we doing this? Please try to avoid copies and allocaitons as much as possible.

src/trace_redaction/trace_redactor.cc
```diff
@@ -153,74 +168,134 @@ base::Status TraceRedactor::Transform(
...
+  std::string packet;
+  for (const auto& augmenter : augmenters_) {
+    // Keep augmenting until the augmenter returns an empty packet.
+    while (true) {
+      packet.clear();
+      RETURN_IF_ERROR(augmenter->Augment(context, &packet));
+      if (packet.empty()) {
+        break;
+      }
+      protozero::HeapBuffered<protos::pbzero::Trace> serializer;
```
LalitMaganti: This is extremely wasteful. You are allocating, a buffer, copying a bunch of data, then serializing to a string only to append it to a bigger buffer. Please be more mindful of allocations and copies, as we've discussed, this code is quite perf sensitive.

src/trace_redaction/merge_process_tree.cc
```diff
@@ -0,0 +1,142 @@
...
+  for (const auto& it : merged_tree.threads_by_tid) {
+    const Context::ProcessTreeThread& thread = it.second;
+    auto* thread_field = tree->add_threads();
+    thread_field->set_tgid(thread.tgid);
+    thread_field->set_tid(thread.tid);
+    if (!thread.name.empty()) {
+      thread_field->set_name(thread.name);
+    }
+  }
+  packet->assign(message.SerializeAsString());
```
LalitMaganti: Instead of serializing here, you can use GetSlices and write this in a way which reduces one allocation and copy.

src/trace_redaction/trace_redactor.cc
edgararriagag: I had done it this way to keep using the TraceBlobView currently used in other APIs, but now that we are making this multi pass arch, it makes sense to actually be more efficient here. I went ahead and avoided this copy by making a double string buffer ping-pong approach when doing redaction passes to only have 2 buffers allocated at most (one for reading and one for writting) without copying when flipping between them.

src/trace_redaction/trace_redactor.cc
edgararriagag: I was originally going off the way that Transform worked, it was doing this same approach on every packet, which I agree is wasteful. I went ahead with your suggestion and avoided this by appending directly to the output buffer both the field premable and payload and avovid creating this HeapBuffered wrapper, which avoids that intermediate string creation.

src/trace_redaction/merge_process_tree.cc
edgararriagag: Done

src/trace_redaction/merge_process_tree.cc
```diff
@@ -0,0 +1,143 @@
...
+  const auto& merged_tree = context.merged_process_tree.value();
+  if (merged_tree.processes_by_pid.empty() &&
+      merged_tree.threads_by_tid.empty()) {
+    return base::OkStatus();
+  }
+
+  emitted_ = true;
+
+  protozero::HeapBuffered<protos::pbzero::TracePacket> message;
+  message->set_timestamp(merged_tree.timestamp);
```
LalitMaganti, line 114: Just as an FYI, this *will* mess with the timestamps in trace processor. Maybe it doesn't matter for profilingmanager purposes but you are giving up the quality of the trace somewhat by doing this.

LalitMaganti (approved):

src/trace_redaction/merge_process_tree.cc
edgararriagag: Ack, I can't yet devise a practical case were app devs would care about them, but will keep it in mind as a caveat.

etiennep-chromium:
These tests are failing on Windows blocking perfetto -> chromium roll:
TraceRedactorTest.MultiPassPipelineExecution
TraceRedactorTest.SinglePassAppendsAugmentAtEnd
TraceRedactorTest.ThreePassPipelineExecution

https://ci.chromium.org/ui/p/chromium/builders/try/win-rel/1663709/overview

PR #6988, opened 2026-08-04
Ensure `object-src: 'none'` CSP directive is applied as intended

Svetlitski:
[According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/object-src#none) the single-quotes surrounding `'none'` are mandatory in the `object-src` CSP directive. Unlike the other CSP directives that take care to preserve these single-quotes when required (e.g. `'self'` in `connect-src`), the `object-src` was passing a plain `none` without quotes, which is treated as the bare hostname `none`.

### DIFF
--- a/ui/src/bigtrace/index.ts
+++ b/ui/src/bigtrace/index.ts
@@ -60,7 +60,7 @@ function setupContentSecurityPolicy() {
     // wasm-unsafe-eval: the SQL formatter runs as WebAssembly; this allows
     // Wasm compilation without enabling JS eval().
     'script-src': [`'self'`, `'wasm-unsafe-eval'`],
-    'object-src': ['none'],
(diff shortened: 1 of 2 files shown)


stevegolton (approved):
Good catch, thanks!

LalitMaganti (approved):

sashwinbalaji (approved):

rsavitski (approved):

stevegolton:
@Svetlitski I think we just need you to sign the CLA.

Svetlitski:
@stevegolton Sorry for the delay, CLA is now signed.

LalitMaganti:
Thanks! All merged :)

PR #6991, opened 2026-08-04
ui: merge duplicate per-cpu/gpu tracks on multi-machine open

zezeozue:
A merged trace can contain several track rows for one logical entity: two traces on the same machine each intern their own cpu_frequency, gpufreq, cpu-per-uid totals and gpu work-period tracks. Keying a track URI off a single row's dimension made registerTrack() throw RegistryError on the duplicate; keying it off the row id avoided the crash but split one CPU/GPU into several tracks.

Follow the Sched plugin: enumerate the logical entity (grouping by machine plus its machine-unique key) and have the single resulting track query across all of that entity's track ids. This removes the collision and unions the rows into one contiguous track, keeping distinct machines separate.

  CpuFreq       group by ucpu; union freq/idle track ids
  Gpu frequency group by (machine, ugpu|gpu_id); union track ids
  GpuWorkPeriod group by (machine, ugpu|gpu_id, uid); union track ids
  CpuPerUid     group by (machine, type, cluster); union track ids

Sched, CpuidleTimeInState, CoarseCpu, LinuxPerf and the Gpu counter/slice tracks already enumerate distinct keys or suffix the track id, so they are unaffected.

### DIFF
--- a/ui/src/plugins/com.android.CpuPerUid/index.ts
+++ b/ui/src/plugins/com.android.CpuPerUid/index.ts
@@ -17,7 +17,7 @@ import type {PerfettoPlugin} from '../../public/plugin';
 import {TrackNode} from '../../public/workspace';
 import {CounterTrack} from '../../components/tracks/counter_track';
 import StandardGroupsPlugin from '../dev.perfetto.StandardGroups';
-import {NUM, STR} from '../../trace_processor/query_result';
+import {NUM, NUM_NULL, STR} from '../../trace_processor/query_result';
 
 export default class implements PerfettoPlugin {
   static readonly id = 'com.android.CpuPerUid';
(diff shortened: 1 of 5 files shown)


LalitMaganti:
Sorry but this is just *not* the correct way to do this. I don't want to support mutlple tracks for the same machine like this at all, it's intentional this is crashing. Realistically, we should actually, in trace processor, drop the duplicate tracks entirely.

Having duplicate frequency tracks for the same machine just does *not* make sense and we should fudnamentally disallow it.

LalitMaganti:
The semantics of multiple traces contributing to the same "shared" CPU frequency/schedling breaks my brain and is just not something I want to support fuill stop. Not unless someone goes threough all the edge cases and convinces me of them.

zezeozue:
Well, this breaks any usecases that involve merging field traces (from the same device) with sched? It's a low priority thing for me anyways, so I'll close this. But I thought this is a use case you'd like to support?

LalitMaganti:
The usecase I want to support is *different* datasources being merged together collected a the same time. It doesn't work for *same* dataousrces being merged, especially at the same time but even at different times. There's too many edge cases and gotchas.

zezeozue:
How hard to drop the data when there's a conflict like this in the TP side? It would be nice even if there's a same data source the ui doesn't crash and there's a clear warning about why data is dropped

LalitMaganti:
> How hard to drop the data when there's a conflict like this in the TP side? It would be nice even if there's a same data source the ui doesn't crash and there's a clear warning about why data is dropped

Should be straightforward. Ultimately what you need is a tracker to keep track of the "first trace to claim scheduling/CPU frequency for machine X" and then prevent any other trace from using that, dropping events as necessary (and incrementing a stat of course).

PR #6995, opened 2026-08-04
tp: read only the ELF headers needed to extract the build-id

LalitMaganti:
The local symbolizer mmap'd every candidate file in full and walked its
section header table to find the GNU build-id note. For large binaries
the section headers sit at the end of the file, so this faults in pages
far from the ELF header.

Parse the ELF from a handful of small preads instead: the ELF header,
the program header table and the build-id note found via PT_NOTE
segments, with a section-table fallback for files whose note is
section-only. The parse also reuses the fd that verified the magic,
dropping a redundant stat and file open.

Measured on bundle's real symbol paths (~5300 files): index build
300 ms -> 157 ms (~1.9x); a 60 MB bundle run 1.42 s -> 1.28 s.

### DIFF
--- a/src/trace_processor/util/symbolizer/local_symbolizer.cc
+++ b/src/trace_processor/util/symbolizer/local_symbolizer.cc
@@ -31,7 +31,6 @@
 #include <optional>
 #include <string>
 #include <string_view>
-#include <type_traits>
 #include <utility>
 #include <vector>
 
@@ -39,7 +38,6 @@
 #include "perfetto/base/logging.h"
 #include "perfetto/ext/base/file_utils.h"
 #include "perfetto/ext/base/scoped_file.h"
-#include "perfetto/ext/base/scoped_mmap.h"
 #include "perfetto/ext/base/string_splitter.h"
 #include "perfetto/ext/base/string_utils.h"
 #include "perfetto/ext/base/utils.h"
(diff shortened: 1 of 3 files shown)


src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -77,77 +77,77 @@ std::string GetLine(const std::function<int64_t(char*, size_t)>& fn_read) {
...
     return std::nullopt;
-  }
-  for (size_t i = 0; i < ehdr->e_shnum; ++i) {
-    typename E::Shdr* shdr = GetShdr<E>(mem, ehdr, i);
-    if (!InRange(mem, size, shdr, sizeof(typename E::Shdr))) {
-      return std::nullopt;
-    }
-
-    if (shdr->sh_type != SHT_NOTE)
+  const size_t shdr_table =
```
sashwinbalaji: nit: call it shdr_table_size (same for phdr_table)

src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -77,77 +77,77 @@ std::string GetLine(const std::function<int64_t(char*, size_t)>& fn_read) {
...
-          }
-          return std::string(value, nhdr->n_descsz);
-        }
-      }
-      offset += sizeof(*nhdr) + base::AlignUp<4>(nhdr->n_namesz) +
-                base::AlignUp<4>(nhdr->n_descsz);
+    // Build-id notes are tiny; cap the read to avoid huge allocations from
+    // corrupt headers.
+    const size_t note_size = static_cast<size_t>(shdr.sh_size);
+    if (note_size == 0 || note_size > (1u << 20))
```
sashwinbalaji: nit: can this be some helper function as 1u << 20 is used in couple of places

src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -77,77 +77,77 @@ std::string GetLine(const std::function<int64_t(char*, size_t)>& fn_read) {
...
-
-    if (shdr->sh_type != SHT_NOTE)
+  const size_t shdr_table =
+      static_cast<size_t>(ehdr.e_shnum) * sizeof(typename E::Shdr);
+  std::string shdrs;
+  shdrs.resize(shdr_table);
+  // Tolerate a truncated section header table (declared e_shnum larger than
+  // the actual table): process only the complete entries.
+  const size_t num_shdrs =
+      ReadFileAt(fd, static_cast<size_t>(ehdr.e_shoff), &shdrs[0], shdr_table) /
```
sashwinbalaji: can we use shdrs.data() insteadof &shdrs[0]

src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -328,26 +382,20 @@ void ProcessBinaryFile(const char* fname,
...
+  char magic[EI_MAG3 + 1];
+  if (base::Read(*fd, &magic, sizeof(magic)) != sizeof(magic) ||
+      (!IsElf(magic, sizeof(magic)) && !IsMachO64(magic, sizeof(magic)))) {
+    // Too small to be a binary, or not an ELF/Mach-O file: skip silently.
+    PERFETTO_DLOG("%s not an ELF or Mach-O 64.", fname);
+    return;
   }
-  std::optional<BinaryInfo> binary_info = GetBinaryInfo(fname, size);
+  std::optional<BinaryInfo> binary_info =
+      GetBinaryInfoFromFd(fd.get(), fname, size);
```
sashwinbalaji: nit: can we optimise the number of read calls. we read 

(base::Read(*fd, &magic, sizeof(magic)) != sizeof(magic) 

then again we read in GetBinaryInfoFromFd the size is different but maybe get once and reuse

src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -328,26 +382,20 @@ void ProcessBinaryFile(const char* fname,
                        std::map<std::string, FoundBinary>& result,
                        uint32_t* corrupt_file_count) {
   static_assert(EI_MAG3 + 1 == sizeof(kMachO64Magic));
-  char magic[EI_MAG3 + 1];
-  // Scope file access. On windows OpenFile opens an exclusive lock.
-  // This lock needs to be released before mapping the file.
```
sashwinbalaji, line 333: won't this be a problem now, it's a corner case but still as we now hold the fd across read

src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -77,77 +77,77 @@ std::string GetLine(const std::function<int64_t(char*, size_t)>& fn_read) {
...
+// file is smaller than |len| (e.g. a truncated header).
+size_t ReadFileAt(base::PlatformHandle fd,
+                  size_t offset,
+                  void* buf,
+                  size_t len) {
+  if (len == 0)
+    return 0;
+  if (!base::SeekFile(fd, offset))
+    return 0;
+  ssize_t rd = base::Read(fd, buf, len);
```
sashwinbalaji, line 88: this iwll not work on windows PlatformHandle is void* there

[PlatformHandle](https://crsrc.org/c/third_party/perfetto/include/perfetto/base/platform_handle.h;l=44)

sashwinbalaji (approved):

PR #6997, opened 2026-08-04
tp: add SQLite VFS backed by abstract I/O

LalitMaganti:
Add a random-access FileSystem interface and a SQLite VFS backed by it.

Use process-wide local and no-op filesystem implementations, with reusable seek and truncate utilities.

### DIFF
--- a/src/base/file_utils.cc
+++ b/src/base/file_utils.cc
@@ -267,24 +267,28 @@ bool FlushFile(int fd) {
 }
 
 bool SeekFile(int fd, uint64_t offset) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
   if (offset > static_cast<uint64_t>(std::numeric_limits<int64_t>::max())) {
     errno = EOVERFLOW;
     return false;
   }
-#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
   return _lseeki64(fd, static_cast<int64_t>(offset), SEEK_SET) != -1;
 #else
+  if (offset > static_cast<uint64_t>(std::numeric_limits<off_t>::max())) {
+    errno = EOVERFLOW;
+    return false;
+  }
   return lseek(fd, static_cast<off_t>(offset), SEEK_SET) !=
          static_cast<off_t>(-1);
 #endif
 }
 
 bool TruncateFile(int fd, uint64_t size) {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
   if (size > static_cast<uint64_t>(std::numeric_limits<int64_t>::max())) {
     errno = EOVERFLOW;
     return false;
   }
-#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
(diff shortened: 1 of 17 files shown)


src/base/file_utils.cc
```diff
@@ -265,6 +266,36 @@ bool FlushFile(int fd) {
 #endif
 }
 
+bool SeekFile(int fd, uint64_t offset) {
+  if (offset > static_cast<uint64_t>(std::numeric_limits<int64_t>::max())) {
```
safayat-google, line 271: nit: check against `std::numeric_limits<off_t>::max()` for non-windows. also in `TruncateFile()`

src/base/file_utils.cc
```diff
@@ -265,6 +266,36 @@ bool FlushFile(int fd) {
 #endif
 }
 
+bool SeekFile(int fd, uint64_t offset) {
```
safayat-google, line 269: nit: check for invalid fd?

src/base/file_utils.cc
```diff
@@ -265,6 +266,36 @@ bool FlushFile(int fd) {
...
+  }
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  int result = _chsize_s(fd, size);
+  if (result != 0) {
+    errno = result;
+    return false;
+  }
+  return true;
+#else
+  return ftruncate(fd, static_cast<off_t>(size)) == 0;
```
safayat-google: nit: wrap with `PERFETTO_EINTR()`?

src/base/utils_unittest.cc
```diff
@@ -364,6 +364,28 @@ TEST(UtilsTest, CopyFileContents) {
...
+  ASSERT_EQ(WriteAll(*file, "abcdef", 6), 6);
+
+  ASSERT_TRUE(TruncateFile(*file, 3));
+  EXPECT_EQ(GetFileSize(*file), 3u);
+
+  std::string contents;
+  ASSERT_TRUE(ReadFile(file.path(), &contents));
+  EXPECT_EQ(contents, "abc");
+}
+
```
safayat-google, line 404: nit and optional: add tests to cover with growing size(TruncateFile), invalid fd and overflow case?

safayat-google (approved):

src/base/file_utils.cc
LalitMaganti: Leaving this to the syscall, which rejects invalid descriptors with EBADF.

src/base/file_utils.cc
LalitMaganti: Wrapped ftruncate. lseek is not interruptible and does not report EINTR.

src/base/utils_unittest.cc
LalitMaganti: Done; added growth, invalid-fd, and overflow coverage.

src/base/file_utils.cc
LalitMaganti: Done. This matters on platforms or build configurations where off_t is narrower than 64 bits.

PR #7003, opened 2026-08-05
ui: Extend plugin selection detail tabs to track events

ivan-chong:
This PR extends the subtab concept from area selection tabs to also track event selections. This allows plugins to register their custom subtabs which can be displayed in the selection details bottom drawer even on regular track slices.

Bug: 289343169

### DIFF
--- a/ui/src/public/selection.ts
+++ b/ui/src/public/selection.ts
@@ -35,7 +35,7 @@ export interface ContentWithLoadingFlag {
   readonly buttons?: m.Children;
 }
 
-export interface AreaSelectionTab {
+export interface SelectionTab<T = Selection> {
   // Unique id for this tab.
   readonly id: string;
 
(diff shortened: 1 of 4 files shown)


ui/src/public/selection.ts
```diff
@@ -154,10 +157,20 @@ export interface SelectionManager {
...
+
   /**
    * Register a new tab under the area selection details panel.
    */
   registerAreaSelectionTab(tab: AreaSelectionTab): void;
+
+  /**
+   * Register a new tab under the track event selection details panel.
+   */
+  registerTrackEventSelectionTab(tab: TrackEventSelectionTab): void;
```
stevegolton: Could you remove this function and just keep the registerSelectionTab and registerAreaSelectionTab for compatibility? Then when registering the track event specific tab in your next PR just do the check manually and return undefined if it's not a track event selection type.

I think ultimately everything is simpler with the single generic entry point, and we should port all the area selection tabs over to it to at some point. But that's out of scope for this PR.

Maybe add a jsdoc deprecation notice around registerAreaSelectionTab() as well to avoid new usages of it?

stevegolton (approved):
Looks awesome, thanks!

PR #7004, opened 2026-08-05
ui: Show lock contention breakdown in selection details

ivan-chong:
This PR displays the lock contention breakdown even on the regular lock contention slices, which makes the plugin more discoverable and doesn't require the user to stay on the debug lock owner slice to view these details.

Bug: 289343169


ui/src/plugins/com.android.AndroidLockContention/lock_owner_details_panel.ts
```diff
@@ -38,82 +43,155 @@ import {Callout} from '../../widgets/callout';
...
 
   constructor(
     private readonly trace: Trace,
-    private readonly eventId: number,
     private readonly plugin: AndroidLockContentionPlugin,
   ) {}
 
-  async load() {
+  render(selection: TrackEventSelection): ContentWithLoadingFlag | undefined {
+    if (this.currentSelection?.eventId !== selection.eventId) {
```
stevegolton, line 74: Could you use an AsyncMemo instead of the limiter and the manual caching? It should wrap all that up for you. Use the eventid as the key.

ui/src/plugins/com.android.AndroidLockContention/lock_owner_details_panel.ts
```diff
@@ -38,82 +43,155 @@ import {Callout} from '../../widgets/callout';
...
-    const uniqueBlockedThreads = new Set(
-      rows.map((r) => r.blockedThreadTid ?? r.blockedThreadName),
-    );
     return m(
-      DetailsShell,
-      {
-        title: 'Lock Owner Contention Breakdown',
-        description: `Owner is blocking ${uniqueBlockedThreads.size} threads`,
-      },
+      'div',
```
stevegolton, line 193: Nit: '.pf-lock-owner-panel' is a bit tidier.

Also below.

stevegolton (approved):
Looks great, thanks!

ui/src/plugins/com.android.AndroidLockContention/lock_owner_details_panel.ts
```diff
@@ -38,119 +43,201 @@ import {Callout} from '../../widgets/callout';
...
-  >();
+  >;
+}
+
+export class LockContentionDetailsTab implements TrackEventSelectionTab {
+  readonly id = 'android_lock_contention';
+  readonly name = 'Lock Contention';
+  readonly priority = 10;
+
+  private readonly memo = new AsyncMemo<ContentionBreakdownData | null>();
```
stevegolton: Nit: Prefer undefined over null purely for consistency with the rest of the project.

ui/src/plugins/com.android.AndroidLockContention/lock_owner_details_panel.ts
```diff
@@ -38,119 +43,201 @@ import {Callout} from '../../widgets/callout';
...
   ) {}
 
-  async load() {
+  render(selection: TrackEventSelection): ContentWithLoadingFlag | undefined {
+    const {data} = this.memo.use({
+      key: {eventId: selection.eventId},
+      compute: () => this.loadData(selection),
+    });
+
+    if (!data || data.rows.length === 0) {
```
stevegolton, line 96: UX nit: I'd return a loading spinner (not undefined) here while loading. This makes the tab appear immediately after making the selection. Returning undefined while loading hides the tab entirely until it's loaded which produces unpleasant pop-in.

stevegolton (approved):

ui/src/plugins/com.android.AndroidLockContention/lock_owner_details_panel.ts
ivan-chong: Thanks for spotting, done!

ui/src/plugins/com.android.AndroidLockContention/lock_owner_details_panel.ts
ivan-chong: I see thanks, makes sense - done!

PR #7006, opened 2026-08-05
ui: Add thread/process fields to slice aggregations

stevegolton:
This patch adds additional optional aggregations to the generic slice aggregation panel. Users can now choose to group on process and thread properties.

Explanation:

Slice tracks are generic tracks. Anything that can be represented by slices on a timeline use slice tracks to render their data into the UI. If were to be dogmatic about it, then a truly generic slice aggregation panel shouldn't include any opinionated joins on process and thread.

However, there are several reasons why this makes sense pragmatically:
- In 99% of traces, slice tracks are thread execution tracks, where process and thread are useful fields to have available.
- The vast majority of slice tracks use the 'ThreadSliceDetailsPanel' which contains (among other opinionated details) thread and process information.
- If a given track has no upid or utid dimension, then the table simply shows NULL anyway.

In the future, it would make sense to pull out the track kinds into separate categories, allowing pure slice tracks to exist without having an opinionated aggregation panel bolted on, but this is out of scope. This is the pragmatic change.

Summary of changes:
- Add the process and thread field as optional joinable columns.
- Rename SliceSelectionAggregator to ThreadSliceAggregator (note added the prefix 'Thread' and drop 'Selection' for brevity).
- While in the area, also add optional joinable process fields to the frames aggregator.


ui/src/plugins/dev.perfetto.TraceProcessorTrack/thread_slice_aggregator.ts
```diff
@@ -0,0 +1,383 @@
...
+          thread_track.utid AS utid,
+          COALESCE(thread.upid, process_track.upid) AS upid,
+          thread.name AS thread_name,
+          process.name AS process_name
+        FROM ${iiTable.name} AS slice
+        LEFT JOIN ${childDurTable.name} USING (id)
+        LEFT JOIN thread_track ON (slice.track_id = thread_track.id)
+        LEFT JOIN thread USING (utid)
+        LEFT JOIN process_track ON (slice.track_id = process_track.id)
+        LEFT JOIN process ON (
```
LalitMaganti: This is a very non-ideal join. Please do a double join and coalesce on the output side. Never do a non-eqiality join like this.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/thread_slice_aggregator.ts
```diff
@@ -0,0 +1,383 @@
...
+        WHERE parent_id IS NOT NULL
+        GROUP BY parent_id
+      `,
+    });
+
+    // Create a coalesced table containing
+    return await createPerfettoTable({
+      engine,
+      as: `
+        SELECT
```
LalitMaganti, line 148: I feel like there's a viz table which largely does exactly this same query...

stevegolton: If you're referring to `slices.with_context.thread_slice`, it does an inner join which drops any non thread slice tracks. I started off with this, but switched to the manual left join.

If we select over non thread/process tracks, we don't want to drop the slices, we just want to show NULL instead.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/thread_slice_aggregator.ts
stevegolton: Good catch, this was left over from a previous world. The join is now formed dynamically on demand when the user adds one of the process details columns, so we no longer need to join in the original table.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/thread_slice_aggregator.ts
LalitMaganti: Slices.slices (_slice_with_thread_and_process_info).

stevegolton: Apparently this was removed over a year ago in favour of `thread_or_process_slice`, which itself takes the thread_track table and the process_track table and unions them together + joining on slice, so it has the same problem.

LalitMaganti: Might have been just looking at some outdated codesearch :( LGTM then

LalitMaganti (approved):

PR #7015, opened 2026-08-06
profiling:x86_64: use stack pointer instead of frame address for heapprofd

### LINKED ISSUE #4801: heap profiling (53.0) not working with absl time lib (Linux)

safayat-google:
On x86_64, the hardware call instruction pushes the return address directly to the stack at %rsp. When heapprofd intercepts allocations in Client::RecordMalloc, sampling the stack starting from __builtin_frame_address(0) can miss the bottom of the active frame [RSP, __builtin_frame_address(0)) (e.g. when the interceptor library is built with release optimizations where __builtin_frame_address(0) returns the caller's CFA).

During asynchronous unwinding, if the target thread or process exits shortly after an allocation (or under PERFETTO_HEAPPROFD_BLOCKING_INIT=1), heapprofd cannot fall back to reading /proc/<pid>/mem, leading to ERROR MEMORY_INVALID.

This change extends the stack pointer capture mechanism already used for RISC-V to x86_64 as well. It samples stack memory starting from the hardware stack pointer (RSP), ensuring 100% of the active frame is present in the shared memory snapshot buffer regardless of optimization or frame pointer flags.

Fixes: #4801

### DIFF
--- a/src/profiling/memory/client.cc
+++ b/src/profiling/memory/client.cc
@@ -345,9 +345,6 @@ bool Client::IsPostFork() {
   return false;
 }
 
-#if (PERFETTO_BUILDFLAG(PERFETTO_ARCH_CPU_RISCV) ||   \
-     PERFETTO_BUILDFLAG(PERFETTO_ARCH_CPU_X86_64)) && \
-    !PERFETTO_HAS_BUILTIN_STACK_ADDRESS()
 ssize_t Client::GetStackRegister(unwindstack::ArchEnum arch) {
   ssize_t reg_sp, reg_size;
   switch (arch) {
@@ -383,8 +380,6 @@ uintptr_t Client::GetStackAddress(char* reg_data, unwindstack::ArchEnum arch) {
     return reinterpret_cast<uintptr_t>(nullptr);
   return *reinterpret_cast<uintptr_t*>(&reg_data[reg]);
 }
-#endif /* (PERFETTO_ARCH_CPU_RISCV || PERFETTO_ARCH_CPU_X86_64) && \
-          !PERFETTO_HAS_BUILTIN_STACK_ADDRESS() */
 
 // The stack grows towards numerically smaller addresses, so the stack layout
 // of main calling malloc is as follows.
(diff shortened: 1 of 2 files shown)


safayat-google:
here's an internal [diff-text](https://diff.googleplex.com/#key=ajhaOSQ1E0Qw) to show the difference of stack sample + read through from the libunwinding before and after the change.

[Also issue reproduce stats in different platform + scenarios](https://paste.googleplex.com/5226144910671872#)

src/profiling/memory/client.cc
```diff
@@ -426,9 +430,15 @@ bool Client::RecordMalloc(uint32_t heap_id,
   }
 #endif /* PERFETTO_HAS_BUILTIN_STACK_ADDRESS() */
 #else
+  // On ARM/ARM64, active registers (such as the return address in LR) are
```
rsavitski: nit: I don't get what this is trying to highlight. AsmGetRegs() grabs active registers on all platforms, so why call this out for arm64 specifically?

src/profiling/memory/client.cc
```diff
@@ -426,9 +430,15 @@ bool Client::RecordMalloc(uint32_t heap_id,
...
+  // captured directly in the register snapshot (metadata.register_data).
+  // Any stack-saved registers and caller frames reside at or above the Frame
+  // Pointer (FP), so capturing stack memory from __builtin_frame_address(0)
+  // upwards is sufficient.
   const char* stackptr = reinterpret_cast<char*>(__builtin_frame_address(0));
   unwindstack::AsmGetRegs(metadata.register_data);
```
rsavitski: Can we please take this as an opportunity to clean up and move the RISCV clutter out of the main function.

I believe the uregs+ustack code can be something like:
```
      AllocMetadata metadata;
      unwindstack::AsmGetRegs(metadata.register_data);

    #if PERFETTO_BUILDFLAG(PERFETTO_ARCH_CPU_ARM64)
      // On ARM64, the prologue allocates the callee-saved register block first and
      // sets x29 (FP) to the base of that block. Starting the stack copy at FP
      // captures all saved registers while omitting ~450 bytes of RecordMalloc's
      // own local stack frame to save ring buffer space.
      const char* stackptr = reinterpret_cast<char*>(__builtin_frame_address(0));
    #else
      // On other architectures (such as x86_64 and RISC-V), callee-saved registers
      // are pushed below the frame pointer (or calling conventions differ). We must
      // copy from the sampled stack pointer (SP) to include all saved registers.
      const char* stackptr = reinterpret_cast<char*>(
          GetStackAddress(metadata.register_data, unwindstack::Regs::CurrentArch()));
      if (!stackptr) {
        PERFETTO_ELOG("Failed to get stack address.");
        shmem_.SetErrorState(SharedRingBuffer::kInvalidStackBounds);
        return false;
      }
    #endif

      const char* stackend = GetStackEnd(stackptr);
      if (!stackend) {
        PERFETTO_ELOG("Failed to find stackend.");
        shmem_.SetErrorState(SharedRingBuffer::kInvalidStackBounds);
        return false;
      }
```

rsavitski (changes requested):

LalitMaganti (approved):

rsavitski (approved):
LGTM but needs a fixup for ASAN. I believe the code is correct, and the simplest change to make the sanitiser happy is to initialise `AllocMetadata metadata = {};`.

PR #7029, opened 2026-08-07
ui: focus slices on double click

### LINKED ISSUE #7017: Should double clicking on slice zoom into that (same as f)?
In other profilers, I've seen that they use double click as a way to "snap zoom" into that region of the slice and select it. I felt the behaviour was nice and I cannot think of any other thing we might want to double click.

Should we switch to the same?

LalitMaganti:
Select and focus timeline slices when they are double-clicked, matching
the existing FocusSelection command behavior. Add a dedicated integration
test and synthetic trace covering the interaction.

Fixes #7017


ui/src/components/tracks/slice_track.ts
```diff
@@ -300,9 +300,15 @@ export interface SliceTrackAttrs<T extends DatasetSchema> {
...
-   * Called when a slice is clicked. Return false to prevent default selection.
+   * Called when a slice is clicked. Overrides the default selection behavior.
    */
   onSliceClick?(args: OnSliceClickArgs<T>): void;
+
+  /**
+   * Called when a slice is double-clicked. Overrides the default select and
+   * focus behavior.
+   */
+  onSliceDoubleClick?(args: OnSliceClickArgs<T>): void;
```
stevegolton: I'd remove this override unless someone actually needs it - keeps things simple.

ui/src/test/slice_double_click_zoom.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  pth = new PerfettoTestHelper(page);
+  await pth.openTraceFile(traceName);
+});
+
+test('double-clicking a slice zooms into it', async () => {
+  const track = pth.locateTrack('Global Track Events/Slices');
+  const canvas = track.locator('.pf-track__canvas');
+  const box = await canvas.boundingBox();
+  if (box === null) throw new Error('Track canvas bounding box is null');
+  await canvas.dblclick({position: {x: box.width / 2, y: box.height / 2}});
```
stevegolton, line 34: Do you think it makes sense to also test the timeline bounds here to see whether we've actually zoomed in or not? The golden screenshot example could fail quite transparently.

stevegolton (approved):

ui/src/test/slice_double_click_zoom.test.ts
LalitMaganti: Done!

PR #7031, opened 2026-08-07
ui: Support filtering Day Explorer flamegraph by selected tracks

sanath099:
Pass track_id and root status to Day Explorer tracks, and use them in computeDayExplorerFlameGraph to construct a recursive query filtering by selected tracks and their descendants.

Bug: 482033855

Welcome to Perfetto!

For more details please see

### DIFF
--- a/ui/src/plugins/com.android.DayExplorer/index.ts
+++ b/ui/src/plugins/com.android.DayExplorer/index.ts
@@ -100,11 +100,13 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
       `;
       const groupKey = `_day_explorer_ui_hierarchy_under_${parentId}`;
       const trackName = `${childIter.display_name} - ${childIter.energy_mwh}mWh`;
+
       const node = await this.createDayExplorerTrack(
         ctx,
         trackName,
         groupKey,
         query,
+        childIter.track_id,
       );
       parent.addChildInOrder(node);
       await this.addDayExplorerRecursive(ctx, node, limit, childIter.track_id);
@@ -116,6 +118,7 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
     name: string,
     groupKey: string,
     query: string,
+    trackId: bigint,
   ): Promise<TrackNode> {
     const uri = `/day_explorer_${uuidv4()}`;
     const renderer = await CounterTrack.createMaterialized({
@@ -130,6 +133,7 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
       renderer,
       tags: {
         kinds: [DAY_EXPLORER_TRACK_KIND],
+        trackId: Number(trackId),
       },
     });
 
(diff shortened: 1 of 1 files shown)


ui/src/plugins/com.android.DayExplorer/index.ts
```diff
@@ -130,6 +134,9 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
       renderer,
       tags: {
         kinds: [DAY_EXPLORER_TRACK_KIND],
+        trackId: trackId.toString(),
```
rzuklie: Does this need to be a string? Does Perfetto not accept this being a number?

ui/src/plugins/com.android.DayExplorer/index.ts
```diff
@@ -130,6 +134,9 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
...
+        isLabelRoot: isRoot && name.startsWith('[Label]'),
```
rzuklie: This is pretty tightly coupled with the google3 side. I think it would make more sense to encode this into the hierarchy. For example, it could be that every hierarchy root is a single root that avoids overlapping count. E.g.

```
- Regular (root 1)
  - Screen on
  - Screen off
- [Label] Media playback (root 2)
  - etc
```

ui/src/plugins/com.android.DayExplorer/index.ts
```diff
@@ -178,27 +185,66 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
...
+    const selectedTrackIds: bigint[] = [];
+    let hasPhysicalSelected = false;
+
+    // First pass: identify if any physical roots are selected
+    for (const trackInfo of currentSelection.tracks) {
+      if (
+        trackInfo?.tags?.kinds?.includes(DAY_EXPLORER_TRACK_KIND) &&
+        trackInfo.tags.trackId !== undefined
+      ) {
+        if (trackInfo.tags.isPhysicalRoot === true) {
```
rzuklie: nit: why the additional nested if?

```
if (
  trackInfo?.tags?.kinds?.includes(DAY_EXPLORER_TRACK_KIND) &&
  trackInfo.tags.trackId !== undefined &&
  trackInfo.tags.isPhysicalRoot === true) {
```

ui/src/plugins/com.android.DayExplorer/index.ts
```diff
@@ -178,27 +185,66 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
...
   ): ReadonlyArray<QueryFlamegraphMetric> | undefined {
     // The flame graph will be shown when any day explorer track is in the area
-    // selection. The selection is used to filter by time, but not by track. All
-    // day explorer tracks are considered for the graph.
-    let hasDayExplorer = false;
+    // selection. The selection is used to filter by time, and we filter the graph
+    // to only include energy from the selected tracks and their recursive descendants.
+    // If a physical track is selected, we exclude label roots to avoid double-counting.
+    const selectedTrackIds: bigint[] = [];
+    let hasPhysicalSelected = false;
```
rzuklie: This is only looking for roots, so this can still overcount. For example:

```
[ ] - Screen on
[x]     - Compute
[x]     - Networking
[ ] - Screen off
[ ]     - Compute
[ ]     - Networking
[ ] - [Label] Media
[x]     - Compute
[ ]     - Networking
```

And this is probably worse even, since you lose the context of being under the label. Though, it's arguably a bad idea for someone to select like this :)

The original motivation for not filtering by track was to ensure all area selections had context. Additional filtering could be done on the flame chart.

I wonder if a better solution is to try to find the root for every track in the area selection. It would still allow overcounting, but only if you explicitly select tracks from regular and label tracks. And, if you do, you would at least have the appropriate context. But it would be better than today where it _always_ shows the label tracks.

ui/src/plugins/com.android.DayExplorer/index.ts
sanath099: Done. I've updated trackId in the tags to be passed as a number (Number(trackId)), and updated the parser in computeDayExplorerFlameGraph to support both number and string types when constructing the query.

ui/src/plugins/com.android.DayExplorer/index.ts
sanath099: Removed

ui/src/plugins/com.android.DayExplorer/index.ts
sanath099: I've updated the logic to address this.  Removed the TS-level filter that discarded label tracks when physical tracks were selected. Now, the SQL query recursively resolves ancestors for all selected tracks, allowing both physical and label hierarchies to render side-by-side with their correct root context. If a user explicitly selects both, they will see both trees (with potential double-counting at the root level, but preserving visual context as discussed).

ui/src/plugins/com.android.DayExplorer/index.ts
sanath099: Done The UI now treats all selected tracks uniformly, and they will render side-by-side with their correct root context. If a user mixes selections, they will see both trees (with potential double-counting, which we agreed is the preferred tradeoff to preserve context).

ui/src/plugins/com.android.DayExplorer/index.ts
```diff
@@ -178,27 +182,61 @@ export default class DayExplorerPlugin implements PerfettoPlugin {
...
+              SELECT track_id FROM descendants
+              UNION
+              SELECT track_id FROM ancestors
+            ),
             total_energy AS (
               SELECT track_id, parent_id, display_name, SUM(energy_uws) AS energy_uws
               FROM day_explorer_ui_hierarchy_per_ts
               WHERE ts >= ${currentSelection.start}
                 AND ts <= ${currentSelection.end}
+                AND track_id IN (SELECT track_id FROM all_nodes)
```
rzuklie, line 239: Overall looks good! I recall you mentioning that this got slower. Did you end up measuring that at all? Did you try using a materialized CTE for `all_nodes`?

rzuklie (approved):

ui/src/plugins/com.android.DayExplorer/index.ts
sanath099: I will do the optimization part in subsequent set of patches

LalitMaganti (approved):

PR #7036, opened 2026-08-10
tools/proto_merger: handle reserved fields by marking them deprecated

andrabivolaru:
When proto fields are removed upstream and their tag numbers are declared as `reserved` (e.g. `reserved 5;`), `proto_merger` previously treated them as deleted and commented them out in the merged output.

However, downstream consumers (such as Google3's `perfetto_log.proto`) must retain these fields to maintain wire compatibility for decoding historical traces and supporting legacy configurations.

### Changes:
- **`proto_file.h` / `proto_file.cc`**: Extract `reserved_range()` from message descriptors into `ProtoFile::Message::reserved_numbers`.
- **`proto_merger.cc`**: When a downstream field is missing from upstream's active field list:
  - If the field number is present in `upstream.reserved_numbers`, preserve it in the active fields list and annotate it with `[deprecated = true]`.
  - Otherwise, preserve existing behavior by moving it to `deleted_fields` (commented out).
  - Sort fields by tag number in `MergeFields` to ensure canonical tag ordering in the serialized proto.
- **`proto_file_serializer_unittest.cc`**: Added unit and end-to-end regression tests verifying that reserved upstream fields are serialized with `[deprecated = true]`.

Test: `out/linux_debug/perfetto_unittests --gtest_filter="ProtoFileSerializerTest.*"`


src/tools/proto_merger/proto_file.h
```diff
@@ -69,6 +69,8 @@ struct ProtoFile {
     std::vector<Oneof> oneofs;
     std::vector<Field> fields;
 
+    std::set<int> reserved_numbers;
```
LalitMaganti: nit: unordered_set

LalitMaganti (approved):

PR #7049, opened 2026-08-11
tp: add flamechart operator for sampled stack rendering

LalitMaganti:
Adds a flamechart operator computing, for a stream of (ts, leaf callsite) samples, the time ranges over which each frame was continuously present at each stack depth. Exposed to SQL as `_stack_sample_flamechart_runs!` in the `std.stack_sample.flamechart` module.

Bug: https://github.com/google/perfetto/issues/7401


src/trace_processor/plugins/flamechart/flamechart.cc
```diff
@@ -0,0 +1,140 @@
...
+#include "src/trace_processor/core/tree/tree.h"
+#include "src/trace_processor/core/util/span.h"
+
+namespace perfetto::trace_processor::flamechart {
+namespace {
+
+// One open (unclosed) segment per stack depth during the sweep.
+struct OpenSegment {
+  // Tree row index of the frame at this depth.
+  uint32_t row;
```
sashwinbalaji: nit: default initialize them

src/trace_processor/plugins/flamechart/flamechart_function.cc
```diff
@@ -0,0 +1,161 @@
...
+    bool failed = false;
+    std::vector<int64_t> ts;
+    std::vector<int64_t> leaf_id;
+  };
+
+  static void Step(sqlite3_context* ctx, int argc, sqlite3_value** argv) {
+    PERFETTO_DCHECK(argc == kArgCount);
+    auto& agg = AggCtx::GetOrCreateContextForStep(ctx);
+    const auto* tree =
+        sqlite::value::Pointer<core::Tree>(argv[0], kTreePointerType);
```
sashwinbalaji: should this be MovePointer as in http://ac/external/perfetto/src/trace_processor/plugins/tree_functions/tree_conversion.cc?l=177-180?

sashwinbalaji (approved):

src/trace_processor/plugins/flamechart/flamechart.cc
LalitMaganti: Done.

src/trace_processor/plugins/flamechart/flamechart_function.cc
LalitMaganti: Yes. It only worked because the tree is the first member of MovePointer. Fixed: the tree is taken on the first row and owned by the aggregate.

PR #7052, opened 2026-08-11
tp: persist an id -> row index on Tree

LalitMaganti:
Builds an id -> row index once during tree construction (identity / dense vector / hash map depending on the id shape) so consumers can resolve node ids without rebuilding the map. Identity is the default, so trees without an id column resolve row indices directly.

### DIFF
--- a/src/trace_processor/core/tree/tree.h
+++ b/src/trace_processor/core/tree/tree.h
@@ -28,6 +28,7 @@
 #include <vector>
 
 #include "perfetto/base/logging.h"
+#include "perfetto/ext/base/flat_hash_map.h"
 #include "src/trace_processor/core/common/storage_types.h"
 #include "src/trace_processor/core/util/bit_vector.h"
 #include "src/trace_processor/core/util/slab.h"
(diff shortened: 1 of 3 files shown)


src/trace_processor/core/tree/tree.h
```diff
@@ -48,6 +49,52 @@ namespace perfetto::trace_processor::core {
...
+    std::vector<uint32_t> dense;  // dense uint32 ids -> row, kNullParent absent
+    // Arbitrary ids -> row. Optional because FlatHashMap's constructor is
+    // explicit, which would break aggregate initialization of Tree.
+    std::optional<base::FlatHashMap<int64_t, uint32_t>> hash;
+
+    // Resolves an id to a row index within a table of |num_rows| rows, or
+    // kNullParent if absent (returned as a sentinel rather than wrapped in
+    // std::optional as this is on the hot path of consumers). |num_rows|
+    // bounds the identity case, where ids are the row indices themselves.
+    uint32_t Find(int64_t id, uint32_t num_rows) const {
```
safayat-google, line 72: nit: add test coverage?

safayat-google (approved):

src/trace_processor/core/tree/tree.h
LalitMaganti: Done. Added TreeFromDataframeTest.FindRowIdentityIds / FindRowDenseIds / FindRowHashIds covering all three index storages, ReorderRebuildsIdIndex for the rebuild after a topological reorder, plus the identity default and duplicate-id rejection in 66e08b99c6.

PR #7063, opened 2026-08-11
Exclude waitForBufferRelease from DrawFrames blocking calls

XSJoJo:
This recalculates the duration and end time for DrawFrames slices by explicitly subtracting the time spent in any child waitForBufferRelease slices. This helps stabilize metrics that are sensitive to EGL swap buffer overhead differences.

Bug: 540803616

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
@@ -87,19 +87,59 @@ RETURNS BOOL
 AS
 SELECT _is_relevant_blocking_call_expr!($name);
 
+CREATE PERFETTO TABLE _android_critical_blocking_calls_draw_frames AS
+SELECT utid, id, ts, dur
+FROM thread_slice
+WHERE
+  name GLOB 'DrawFrames*'
+ORDER BY
+  ts;
+
+CREATE PERFETTO TABLE _android_critical_blocking_calls_wait_for_buffer_release
+AS
+SELECT utid, ts, dur
+FROM thread_slice
+WHERE
+  name = 'waitForBufferRelease'
+ORDER BY
+  ts;
+
 --Extract critical blocking calls from all processes.
 CREATE PERFETTO TABLE _android_critical_blocking_calls AS
+WITH
+  _wait_for_buffer_release_dur AS (
+    SELECT s.id, SUM(child.dur) AS dur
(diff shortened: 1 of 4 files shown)


src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
```diff
@@ -19,89 +19,91 @@ INCLUDE PERFETTO MODULE android.binder;
...
+  -- Handler pattern heuristics
+  $name GLOB '*Handler: *$*'
+  OR $name GLOB '*.*.*: *$*'
+  OR $name GLOB '*.*$*: #*'));
 
 --Extract critical blocking calls from all processes.
 CREATE PERFETTO TABLE _android_critical_blocking_calls AS
+WITH
+  _wait_for_buffer_release_dur AS (
+    SELECT s.id, SUM(child.dur) AS dur
```
LalitMaganti, line 111: What you're doing here is an expensive operation. You should pull out all the waitForBufferRelease in a separate table before you do the below (making sure to order by ts).

src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
```diff
@@ -19,89 +19,91 @@ INCLUDE PERFETTO MODULE android.binder;
...
+    FROM thread_slice AS child
+    JOIN thread_slice AS s
+      ON child.utid = s.utid
+      AND child.ts >= s.ts
+      AND child.ts < s.ts + s.dur
+      AND s.name GLOB 'DrawFrames*'
```
LalitMaganti: This should be in the where clause.

src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
```diff
@@ -19,89 +19,91 @@ INCLUDE PERFETTO MODULE android.binder;
...
+    WHERE
+      child.name = 'waitForBufferRelease'
```
LalitMaganti: It would be more efficient to materialze these two drawFramea and wait for buffer release tables separately (or even better, if you can reuse something which surely already exists :))

LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/critical_blocking_calls.sql
XSJoJo: done

PR #7064, opened 2026-08-11
trace_processor: Add support for tracking string system properties

sadrulhc:
Currently Android system properties that evaluate to strings (rather than numbers) are silently dropped during trace_processor ingestion.

This modifies the android_probes_parser to catch values that fail StringToInt32 conversion and emit them instead as an instantaneous (0-duration) slice dynamically utilizing a sysprop_string dimension track named after the property name (e.g., debug.wallpaper.package_name).

This natively extends Perfetto's android_system_property capabilities without requiring upstream proto changes and exposes string states seamlessly to SQL.

Bug: 415757470
Test: built trace_processor

### DIFF
--- a/src/trace_processor/importers/proto/android_probes_parser.cc
+++ b/src/trace_processor/importers/proto/android_probes_parser.cc
@@ -604,6 +604,15 @@ void AndroidProbesParser::ParseAndroidSystemProperty(int64_t ts,
       continue;
     }
 
+    if (name == "debug.tracing.wallpaper.package_name" ||
+        name == "debug.tracing.wallpaper.class_name") {
+      StringId name_id = context_->storage->InternString(name);
+      StringId value_id = context_->storage->InternString(kv.value());
+      context_->metadata_tracker->SetDynamicMetadata(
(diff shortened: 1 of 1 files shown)


src/trace_processor/importers/proto/android_probes_parser.cc
```diff
@@ -607,6 +607,17 @@ void AndroidProbesParser::ParseAndroidSystemProperty(int64_t ts,
     std::optional<int32_t> state =
         base::StringToInt32(kv.value().ToStdString());
     if (!state) {
+      static constexpr auto kStringBlueprint = tracks::SliceBlueprint(
```
LalitMaganti: Please follow the dvice stat example above instead of doing this.

Also please don't make a generic "string" fallback - add something specific for the strings you're adding.

Likely for your string, you want an entry in the metadata table anyway rather than a slice (unless you really want to track wallpaper changes *during* the trace itself).

sadrulhc: Done.

LalitMaganti (approved):

PR #7072, opened 2026-08-11
ui: Revamp AndroidLockContention track layout

ivan-chong:
To reduce the visual clutter on traces with high number of lock
contentions, this PR moves the blocking contention slices to a nested
track under a counter slice which represents the number of contentions
at a given time period.

We also separate the details track by thread to make it easier to see
the name of the particular thread being blocked and to reduce nesting of
tracks.

Additional related bug fixes:
- Gracefully handle when lock_name is null
- Fix lock name regex when lock name doesn't end with lock (e.g. Lock
  contention on linear alloc)
- Only create owner slice for monitor contention slice if there is a
  monitor contention slice nested above a lock contention on monitor
slice to prevent duplicate owner slices
- Remove colour override on owner slices which helps visualise the
  distribution of lock types

Bug: 289343169


stevegolton (commented):
Nice, I like this approach!

zezeozue:
> * https://storage.googleapis.com/perfetto-ci-artifacts/gh-33867695665-1-ui/ui/index.html

+1. It would also be really nice to have the blocked threads with their slices on different tracks. To be easier to see the thread names and not have overlapping slices in a track

ivan-chong:
> > * https://storage.googleapis.com/perfetto-ci-artifacts/gh-33867695665-1-ui/ui/index.html
> 
> +1. It would also be really nice to have the blocked threads with their slices on different tracks. To be easier to see the thread names and not have overlapping slices in a track

@zezeozue Great idea, thanks! I've now added separate tracks per thread under the counter track.

I've also removed the duplicate owner slice when there is a nested monitor contention and lock contention slice (to prevent overlapping slices) and added back the default colouring of slices to visualise the distribution of the lock type better - WDYT?

stevegolton (approved):
I'm relatively un-opinionated about the UX change within the scope of the plugin, but code LGTM.

Only question is why did the screenshot change?

ivan-chong:
> I'm relatively un-opinionated about the UX change within the scope of the plugin, but code LGTM.
> 
> Only question is why did the screenshot change?

Thanks! Yep the screenshot initially had contention tracks and with this change the tracks' heights increased, due to now being counter tracks, and so shifted the timeline up slightly:

https://storage.googleapis.com/perfetto-ci-artifacts/gh-34487538767-1-ui/ui-test-artifacts/index.html#?testId=eb597fbff9090bb5667a-2d2809ddfa960bd50ed2

zezeozue (approved):

PR #7079, opened 2026-08-12
Fix trace redactor bug that causes data from target process to be dropped

edgararriagag:
This fixes a timeline bug where the main thread can cause the entire process or many slices within the process to be dropped due to a bug in the timeline where the main thread loops forever in itself which also causes performance degradation, this fix significantly improves trace quality by keeping that slices and processes that would otherwise be wrongly dropped and significantly improves redaction performance by reducing redaction cpu cost by around 34% due to the avoidance of looping in the main thread during UID belonging tests.

Internal bug: b/543435350 (Contains in-detail investigation)


src/trace_redaction/collect_timeline_events.cc
```diff
@@ -50,6 +50,18 @@ void MarkOpen(uint64_t ts,
 void MarkOpen(uint64_t ts,
               const ProcessTree::Thread::Decoder& thread,
               ProcessThreadTimeline* timeline) {
+  if (thread.tid() == thread.tgid()) {
```
LalitMaganti, line 53: I'm not sure I understand the motivation for this fix nor the comment explaining it here. Why are main threads in any way different? Surely the fix should be on the timeline lookup side not on the timeline building side

LalitMaganti (approved):

src/trace_redaction/collect_timeline_events.cc
LalitMaganti: Discussed offline.

edgararriagag: Synced offline, summarizing here. The test could be performed on the lookup side, but the practical effect would be the same, threads do not serve other purpose than navigating to the closest PID, which can still be achieved to reach the PID without the presence of the main thread regardless of whether this is done at lookup or build time. It is however better at build time because we avoid that extra condition during lookups, which happens so much to actually affect performance of redactor

PR #7080, opened 2026-08-13
CI: add a buildtools cache probe step

D1rk9ghT:
Small change to the install-build-deps composite action to log cache-key computation timing. No functional impact.


LalitMaganti:
We're not going to allow trying to exploit our infra :)

PR #7085, opened 2026-08-13
tp: update live_tid cache when reaffirming thread parent process

gignat-dev:
When an existing thread is reaffirmed with its parent process in GetOrCreateThreadWithParentInternal(), update live_tid_[tid] and move the utid to the back of tids_[tid].

Background:
ProcessTracker tracks threads across time using two data structures:
- tids_[tid]: An ordered vector of all UniqueTids allocated for a tid, scanned backwards (from rbegin()) to prioritize newer incarnations.
- live_tid_[tid]: A fast-path cache mapping a tid to its active UniqueTid, queried by bare ftrace events (like sched_switch) that do not carry a parent PID.

Problem:
If a TID is recycled across processes—for instance, when a thread is created by Process A, then an out-of-order /proc dump associates that TID with Process B, and later a ProcessTree dump confirms the thread belongs to Process A:
1. Process A's reaffirmation matches the existing utid via GetThreadOrNull(tid, pid), so StartNewThread() is not called.
2. Because StartNewThread() was skipped, live_tid_[tid] was not updated and remained pointing to Process B's utid.
3. In addition, Process B's utid stayed at the back of tids_[tid], so any subsequent RefreshLiveTid() calls would also pick Process B.

Impact:
Subsequent bare lookups from sched_switch and sched_waking routed all CPU scheduling states to Process B. Furthermore, when the thread spawned worker threads via task_newtask, the child threads inherited Process B's upid, mistakenly parenting them to the wrong process.

Fix:
In GetOrCreateThreadWithParentInternal(), when an existing utid is matched for (tid, pid), explicitly update live_tid_[tid] to this utid and move it to the back of tids_[tid]. This ensures both direct lookups and RefreshLiveTid() correctly resolve to the active thread.

BUG: b/546056884

### DIFF
--- a/src/trace_processor/importers/common/process_tracker.cc
+++ b/src/trace_processor/importers/common/process_tracker.cc
@@ -166,6 +166,17 @@ UniqueTid ProcessTracker::GetOrCreateThreadWithParentInternal(
   auto opt_utid = GetThreadOrNull(tid, ps.pid());
   UniqueTid utid = opt_utid ? *opt_utid : StartNewThread(std::nullopt, tid);
 
+  // When an existing thread is reaffirmed with its parent process, update
+  // the live_tid_ cache and move it to the back of tids_ so subsequent
+  // bare lookups and RefreshLiveTid() calls resolve to this active thread.
+  if (opt_utid.has_value()) {
+    live_tid_[tid] = utid;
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Nice find! LGTM!

LalitMaganti (approved):

PR #7089, opened 2026-08-14
perfetto: Fix TracingMuxerImpl::Shutdown() crash

mmeenaks-gh:
At benchmark exit calls "Tracing::Shutdown()"
(triggered via _util_perfetto_fini registered at exit process).

then DestroyStoppedTraceWritersForCurrentThread() function destroys TraceWriterImpl for data sources that have already been explicitly stopped. If any data source not stopped still active at shutdown time retains alive. Then muxer is deleted, which frees all shared memory.

Later PlatformPosix->Shutdown() -> ~TraceWriterImpl() -> Flush() -> ReturnCompletedChunk() which try to access the already freed memory and causes SIGSEGV crash.

Fix: Added DestroyAllTraceWritersForCurrentThread() new helper function to reset all trace writer on the calling thread before shared memory destoryed.

Bug: 534222391


src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -2793,7 +2808,13 @@ void TracingMuxerImpl::Shutdown() {
 
   // Shutting down on the muxer thread would lead to a deadlock.
   PERFETTO_CHECK(!muxer->task_runner_->RunsTasksOnCurrentThread());
-  muxer->DestroyStoppedTraceWritersForCurrentThread();
+  // Destroy ALL trace writers for the calling thread while
+  // the shared memory are still alive.
+  // DestroyStoppedTraceWritersForCurrentThread()
+  // would only clean up writers for already-stopped data sources, leaving
+  // active writers to be destroyed later in ~PlatformPosix() after the
+  // memory have been freed, causing a SIGSEGV crash.
```
sashwinbalaji: nit: how about phrasing as below so it's more generic and also refers to the internal bug so in future know we know why this change:

```
  // Destroy this thread's trace writers, including those of still-running data
  // sources: the SMB dies with `delete muxer` below, but this thread's TLS only
  // at platform->Shutdown() further down. Flushing then is a UAF (b/534222391).
```

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -1842,6 +1842,21 @@ void TracingMuxerImpl::SyncProducersForTesting() {
   PERFETTO_DCHECK(all_producers_connected);
 }
 
+void TracingMuxerImpl::DestroyAllTraceWritersForCurrentThread() {
+  // Forcefully reset all trace writers on the current thread, regardless of
+  // whether their data source instances are stopped.
+  auto* root_tls = GetOrCreateTracingTLS();
+  auto reset_all = [](DataSourceThreadLocalState& tls) {
+    for (auto& inst_tls : tls.per_instance) {
```
sashwinbalaji: nit: pls rename ds_tls to be tls, inst_tls to be ds_tls and we have lambda below called destroy_stopped_instances so name this destroy_all_instances. Aligning the naming across nearby functions makes it a bit easier to read.

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -1842,6 +1842,21 @@ void TracingMuxerImpl::SyncProducersForTesting() {
...
+      if (inst_tls.trace_writer)
```
sashwinbalaji: I think it's safer to not have this if guard and just run reset always

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -136,6 +136,7 @@ class TracingMuxerImpl : public TracingMuxer {
       DataSourceState*,
       BufferExhaustedPolicy buffer_exhausted_policy) override;
   void DestroyStoppedTraceWritersForCurrentThread() override;
+  void DestroyAllTraceWritersForCurrentThread();
```
sashwinbalaji: I understand you wanted to club both the destory functions together but these are all overrides + this doesn't have to be a public function . 

So pls move it below and also add a comment explaining the function pls.

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -1842,6 +1842,21 @@ void TracingMuxerImpl::SyncProducersForTesting() {
...
+        inst_tls.Reset();
+    }
+  };
+  for (auto& ds_tls : root_tls->data_sources_tls)
+    reset_all(ds_tls);
+  reset_all(root_tls->track_event_tls);
+}
```
sashwinbalaji, line 1860: DestroyStoppedTraceWritersForCurrentThread updates the generation but we aren't doing so. I think that is a right choice as we have this in [TracePrologue](https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:external/perfetto/include/perfetto/tracing/internal/data_source_type.h;bpv=1;bpt=1;l=103?q=include%2Fperfetto%2Ftracing%2Finternal%2Fdata_source_type.h&ss=h%2Fgoogleplex-android%2Fplatform%2Fsuperproject%2Fmain%2F%2B%2Frefs%2Fheads%2Fmain:external%2Fperfetto%2F&gsn=TracePrologue&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fgoogleplex-android.googlesource.com%2Fplatform%2Fsuperproject%2Fmain%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dexternal%2Fperfetto%2Finclude%2Fperfetto%2Ftracing%2Finternal%2Fdata_source_type.h%2370rqkyDtAqyhn0uL1uAkd3nztnv5NF3pPF8ec72Pe1w)
```
    if (PERFETTO_UNLIKELY(
            (*tls_state)->root_tls->generation !=
            tracing_impl->generation(std::memory_order_relaxed))) {
      // Will update root_tls->generation.
      tracing_impl->DestroyStoppedTraceWritersForCurrentThread();
    }
```

so now for a post shutdown trace point (if any) we will invoke DestroyStoppedTraceWritersForCurrentThread but since by then instance will be TracingMuxerFake we will crash.  Tracing after shutdown resulting in abort makes sense but for a reader the difference between 2 functions would be not that clear. So pls add a comment something like below:

```
   // Unlike DestroyStoppedTraceWritersForCurrentThread(), we deliberately don't
   // sync |root_tls->generation|: this only runs at shutdown, after which any
   // trace point on this thread aborts in TracingMuxerFake anyway.
```

src/tracing/internal/tracing_muxer_impl.cc
sashwinbalaji: should this also be updated to be DestroyAllTraceWritersForCurrentThread because we invoke 
owned_task_runner.reset(); after deleting muxer.

So I think the same UAF can happen here also if we have some trace events on the muxer thread.

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, Added these comments

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, Modified the names

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, removed the if condition

src/tracing/internal/tracing_muxer_impl.h
mmeenaks-gh: Done, Moved inside the Private

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, Added comments

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, Replaced to DestroyAllTraceWritersForCurrentThread

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -1842,6 +1842,23 @@ void TracingMuxerImpl::SyncProducersForTesting() {
   PERFETTO_DCHECK(all_producers_connected);
 }
 
+void TracingMuxerImpl::DestroyAllTraceWritersForCurrentThread() {
+  // Forcefully reset all trace writers on the current thread, regardless of
+  // whether their data source instances are stopped.
+  auto* root_tls = GetOrCreateTracingTLS();
+  auto reset_all = [](DataSourceThreadLocalState& tls) {
```
sashwinbalaji: nit: rename of this to destroy_all_instances missed from previous review comment.

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -1842,6 +1842,23 @@ void TracingMuxerImpl::SyncProducersForTesting() {
...
+    for (auto& ds_tls : tls.per_instance) {
+        ds_tls.Reset();
+    }
+  };
+  // Unlike DestroyStoppedTraceWritersForCurrentThread(), we deliberately don't
+  // sync |root_tls->generation|: this only runs at shutdown, after which any
+  // trace point on this thread aborts in TracingMuxerFake anyway.
```
sashwinbalaji, line 1859: nit: pls place this comment after the reset_all call basically just before the closing braces.

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -2808,10 +2828,10 @@ void TracingMuxerImpl::Shutdown() {
         PERFETTO_CHECK(!consumer->service_);
       }
     }
-    // Make sure no trace writers are lingering around on the muxer thread. Note
-    // that we can't do this for any arbitrary thread in the process; it is the
-    // caller's responsibility to clean them up before shutting down Perfetto.
-    muxer->DestroyStoppedTraceWritersForCurrentThread();
+    // Make sure no trace writers are lingering around on the muxer thread.
+    // DestroyStoppedTraceWriters would miss active writers; those would UAF
```
sashwinbalaji: nit: DestroyAllTraceWritersForCurrentThread

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -1842,6 +1842,23 @@ void TracingMuxerImpl::SyncProducersForTesting() {
   PERFETTO_DCHECK(all_producers_connected);
 }
 
+void TracingMuxerImpl::DestroyAllTraceWritersForCurrentThread() {
+  // Forcefully reset all trace writers on the current thread, regardless of
+  // whether their data source instances are stopped.
+  auto* root_tls = GetOrCreateTracingTLS();
+  auto reset_all = [](DataSourceThreadLocalState& tls) {
+    for (auto& ds_tls : tls.per_instance) {
+        ds_tls.Reset();
```
sashwinbalaji: yeah pls run tools/format-sources to format the change.

CI is fialing https://github.com/google/perfetto/actions/runs/32017253380/job/95352183277#step:5:12

src/tracing/internal/tracing_muxer_impl.h
sashwinbalaji, line 494: pls add `  void DestroyAllTraceWritersForCurrentThread();` here along with existing functions.

sashwinbalaji (commented):
since this file is critical, sorry for being a bit pedantic.

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, Renamed.

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, Moved the comments before exit function

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, updated comments to DestroyAllTraceWritersForCurrentThread

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Thanks for sharing this tool, Ran this format-source tool and updated the changes

src/tracing/internal/tracing_muxer_impl.h
mmeenaks-gh: Done, Moved this function

mmeenaks-gh:
Hi @sashwinbalaji, your comments are addressed, can you please review latest changes?

LalitMaganti:
@sashwinbalaji is out on vacation until later this week.

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -2808,10 +2828,10 @@ void TracingMuxerImpl::Shutdown() {
         PERFETTO_CHECK(!consumer->service_);
       }
     }
-    // Make sure no trace writers are lingering around on the muxer thread. Note
-    // that we can't do this for any arbitrary thread in the process; it is the
-    // caller's responsibility to clean them up before shutting down Perfetto.
-    muxer->DestroyStoppedTraceWritersForCurrentThread();
+    // Make sure no trace writers are lingering around on the muxer thread.
+    // DestroyAllTraceWritersForCurrentThread will destory the active writers
+    // Before when the thread exits, later delete muxer frees the SMB.
```
sashwinbalaji: nit: pls rephrase this sentence, it's not coherent.

sashwinbalaji (commented):
sorry for the delay, @primiano wanted to take a look into this so was waiting for his review.

src/tracing/internal/tracing_muxer_impl.cc
mmeenaks-gh: Done, Updated comments.

primiano (approved):
I am fine with this change as it is (Assuming doesn't bounce back because breaks some test).

But I just want to point out that this moves the problem only a bit more further down the line. This happens to work only if you have a single thread tracing (Which happens to be the one that calls shutdown), which is probably your case.

In the more general case of a multithreaded app, this issue will keep happening.
This is the reason why I was opposed to the notion of a Shutdown() in the first place.

Substantially it's impossible (*) to have a properly working Shutdown() unless we accept to take a mutex on the tracing path of every TraceWriter. BUt that would mean hurting (a lot) the hot production path, to deal with a case that is used only in tests.

Also Shutdown is this weird thing that never gets any love as it's used only rarely in tests, which means bugs will always creep in there.
 

(*) technically this could be solved with some userspace-RCU, but that has non-trivial complexity and still adds cost to the tracing path

Anyways, we are here now, and you are making Shutdown a little bit better than what it is, so I guess it's strictly better than what we have (hoping it doesn't uncover some other bug that somebody else has elsewhere)


At last, a small aside rant (unrelated with this CL), as this topic is one of my pet peeves:
clean exit/shutdown is one of the canonically complicated problems in tech, which is made complicated by the fact that once somebody in the stack decides "i'm going to handle shutdown cleanly" then every other library / piece of code in the stack must follow suit.
The reality is that exit() (or any notion of clean shutdown) was a big mistake in the history of operating systems.
The only API that shoudl ever existed is _exit() (The one that kills the process immediately without running any destructors). In the real world, processes crash all the time, so code must be robust to tolerate unclean shutdown anyways.

sashwinbalaji (approved):

PR #7093, opened 2026-08-14
tp: speed up SQLite joins on ID columns

LalitMaganti:
Speed up common SQLite joins against trace tables when an ID equality constraint means the joined side can return at most one row.

The query optimizer now emits a compact zero-or-one-row program for ID equality lookups, optionally followed by additional equality predicates. This avoids materializing intermediate ranges and allocating index storage, and lets failed predicates short-circuit immediately. The SQLite virtual-table path also avoids repeated argument copying and moves one-time cursor preparation off the hot path.

## Performance

Representative SQLite join benchmark with 1,024 lookups:

| Path | Before | After |
|---|---:|---:|
| Hit | ~67 us | ~58 us |
| Secondary predicate miss | ~49 us | ~39 us |
| ID miss | ~48 us | ~36 us |

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -17604,6 +17604,7 @@ filegroup {
     srcs: [
         "src/trace_processor/core/interpreter/bytecode_builder.cc",
         "src/trace_processor/core/interpreter/bytecode_interpreter_impl.cc",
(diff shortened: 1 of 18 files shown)


LalitMaganti:
Superseded by the clean-room stack starting at #7097.

PR #7094, opened 2026-08-14
tp: reduce Trace Processor binary size

LalitMaganti:
## Summary

Reduce native and WebAssembly Trace Processor binary size without giving up the point-query performance added by the preceding PR.

The general query interpreter was previously instantiated separately for each filter-value provider. This change type-erases that boundary so the large general execution loop is emitted once, while retaining concrete specialization for the small latency-sensitive zero-or-one-row path.

## Size and performance

- Stripped native binary: 15,284,000 to 15,216,864 bytes for this PR (-67,136 bytes)
- Full stack: approximately 89 KB smaller raw release Wasm than the rebuilt baseline
- No measurable change to the optimized SQLite join timings from the preceding PR

## Testing

- 183 focused interpreter, query, and SQLite connection tests
- Native unit-test, benchmark, and Trace Processor shell builds
- GN dependency check
- Perfetto presubmit

This is the second PR in the stack and depends on #7093.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@ Unreleased:
    *
   Trace Processor:
     * Improved performance of SQLite joins using ID columns.
+    * Reduced native and WebAssembly Trace Processor binary size.
   UI:
     * Double-clicking a timeline slice now selects and zooms into it.
   SDK:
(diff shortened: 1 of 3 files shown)


LalitMaganti:
Superseded by the clean-room stack ending at #7098.

PR #7096, opened 2026-08-14
tp: unify stdlib object discovery

LalitMaganti:
Perfetto's stdlib discovery metadata is currently split across separate module,
table, function, and macro intrinsics. Global discovery therefore requires
agents to enumerate modules and issue repeated metadata queries before they can
find the relevant API.

Replace those intrinsics with a single searchable `__intrinsic_stdlib_objects`
table covering modules, tables, views, functions, table functions, and macros.
Each row includes structured metadata and a precomputed human-readable
`summary`, allowing agents to search all documentation with one
case-insensitive regexp and inspect a candidate without further joins or string
construction.

Also update the Perfetto AI skill to use the unified catalog and add coverage
for all object kinds, summaries, regexp discovery, internal objects, and dotted
package names.

### DIFF
--- a/src/trace_processor/plugins/stdlib_docs/stdlib_docs.cc
+++ b/src/trace_processor/plugins/stdlib_docs/stdlib_docs.cc
@@ -16,10 +16,12 @@
 
 #include "src/trace_processor/plugins/stdlib_docs/stdlib_docs.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <limits>
 #include <memory>
 #include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


src/trace_processor/plugins/stdlib_docs/stdlib_docs.cc
```diff
@@ -48,17 +50,117 @@ template <typename Entry>
...
+        .append(")");
+    if (!entry.description.empty()) {
+      summary->append(": ").append(entry.description);
+    }
+  }
+}
+
+void Append(std::string_view value, std::string* output) {
+  output->append(value.data(), value.size());
+}
```
sashwinbalaji: dont think this is needed, output->append(value) should work

src/trace_processor/plugins/stdlib_docs/stdlib_docs.cc
```diff
@@ -82,388 +184,124 @@ base::StatusOr<stdlib_doc::ParsedModule> ParseModule(
...
  public:
   class Cursor : public StaticTableFunction::Cursor {
    public:
-    explicit Cursor(StringPool*, const PerfettoSqlConnection*);
-    bool Run(const std::vector<SqlValue>& arguments) override;
-
-   private:
-    StringPool* string_pool_ = nullptr;
-    const PerfettoSqlConnection* engine_ = nullptr;
-    tables::StdlibDocsModulesTable table_;
```
sashwinbalaji, line 96: should table_ be part of cursor and not StdlibDocsObjects? Otherwise single table would be used multiple cursor which can be problem in case of self join.

sashwinbalaji (approved):

src/trace_processor/plugins/stdlib_docs/stdlib_docs.cc
LalitMaganti: Done — the generated table now belongs to each cursor, and I added self-join coverage.

src/trace_processor/plugins/stdlib_docs/stdlib_docs.cc
LalitMaganti: Done — removed the helper and append the string views directly.

PR #7100, opened 2026-08-14
ui: Handle macro command errors gracefully and add validation in settings

safayat-google:
Catch exceptions when executing user-defined macros and show a dedicated
macro error dialog pointing to documentation and settings instead of
triggering unhandled rejection errors.

Add an asynchronous "Validate" button and validation result banner to
JsonSettingsEditor alongside "Save", enabling on-demand dry-run checks
for macro configurations in CoreCommands to catch syntax or command
execution failures before saving.

Bug: 540964526

Screenshot 2026-08-14 at 14 20 35
Screenshot 2026-08-14 at 14 21 47
Screenshot 2026-08-14 at 14 25 52


ui/src/core_plugins/dev.perfetto.CoreCommands/index.ts
```diff
@@ -181,9 +182,79 @@ export default class CoreCommands implements PerfettoPlugin {
...
+            message: 'No macros configured.',
+            intent: Intent.Success,
+            icon: 'check_circle',
+          };
+        }
+
+        // Disable prompts during macro dry-run execution
+        using _ = ctx.omnibox.disablePrompts();
+
+        for (const macro of macros) {
```
LalitMaganti, line 209: Validating should not cause the macros to run. This will do very weird things and cause unexpected results.

safayat-google: Makes sense. Also the same macro may pass with one trace and fail for another.\
Updated the logic to do light validation like id not found / duplicate / etc.

LalitMaganti (approved):

PR #7132, opened 2026-08-19
tp: split access-path selection out of dataframe bytecode emission

LalitMaganti:
Pure refactor, splitting query planning into two stages to allow for alternative lowering from logical plan other than
to bytecode.

### DIFF
--- a/src/trace_processor/core/common/op_types.h
+++ b/src/trace_processor/core/common/op_types.h
@@ -58,6 +58,30 @@ struct In {};
 using Op =
     core::TypeSet<Eq, Ne, Lt, Le, Gt, Ge, Glob, Regex, IsNotNull, IsNull, In>;
 
+// Subsets of Op describing which operations a given kind of value or access
+// path supports. Used to pick how a filter is evaluated.
+
+// Set of operations applicable to non-null values.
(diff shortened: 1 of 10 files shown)


src/trace_processor/core/common/op_types.h
```diff
@@ -58,6 +58,30 @@ struct In {};
...
+// path supports. Used to pick how a filter is evaluated.
+
+// Set of operations applicable to non-null values.
+using NonNullOp = core::TypeSet<Eq, Ne, Lt, Le, Gt, Ge, Glob, Regex>;
+
+// Set of operations applicable to non-string values.
+using NonStringOp = core::TypeSet<Eq, Ne, Lt, Le, Gt, Ge>;
+
+// Set of operations applicable to string values.
+using StringOp = core::TypeSet<Eq, Ne, Lt, Le, Gt, Ge, Glob, Regex>;
```
safayat-google, line 71: nit: this is identical to `NonNullOp`. Use one name for both or just copy one with another?

safayat-google (approved):

src/trace_processor/core/common/op_types.h
LalitMaganti: No it's intentionally separate for readability in the actual usage of these structs.

PR #7135, opened 2026-08-19
tp: add a batch execution pipeline

LalitMaganti:
Trace processor moves rows one at a time through a cursor. That interface
works well for callers, but upstream it costs a virtual call and bounds check
per row and prevents operators from working over a run of values.

Add a batch pipeline alongside the cursor. RowBatch holds column views over
producer-owned storage, so filtering changes a selection instead of moving
values. Operators hold only plan data; mutable execution state is owned
separately, which makes plans safe to reuse.

Values normally live until the producer's next pull. RowStore copies them into
fixed-size chunks when an operator needs to keep them longer. Later changes add
sources and operators on top of these primitives.

### DIFF
--- a/src/trace_processor/core/exec/column_view.h
+++ b/src/trace_processor/core/exec/column_view.h
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_TRACE_PROCESSOR_CORE_EXEC_COLUMN_VIEW_H_
+#define SRC_TRACE_PROCESSOR_CORE_EXEC_COLUMN_VIEW_H_
+
+#include <cstdint>
+#include <type_traits>
+
+#include "perfetto/base/compiler.h"
+#include "perfetto/base/logging.h"
+#include "src/trace_processor/core/common/storage_types.h"
+#include "src/trace_processor/core/exec/row_selection.h"
+#include "src/trace_processor/core/exec/variant.h"
+#include "src/trace_processor/core/util/bit_vector.h"
+
+namespace perfetto::trace_processor::core::exec {
+
+// One column of a RowBatch: a reference to storage the batch does not own,
+// plus the row selection currently applied to it.
+class ColumnView {
+ public:
+  enum class Kind : uint8_t {
+    kFlat,
+    kSequence,
+    // `data()` is a Variant array and `type()` is meaningless.
+    kVariant,
+  };
+
+  ColumnView() = default;
+
+  static ColumnView Reference(StorageType type,
+                              const void* data,
+                              const BitVector* validity = nullptr) {
+    ColumnView view;
+    view.type_ = type;
+    if (type.Is<Id>()) {
+      PERFETTO_DCHECK(data == nullptr && validity == nullptr);
+      view.kind_ = Kind::kSequence;
+      return view;
+    }
+    view.kind_ = Kind::kFlat;
+    view.data_ = data;
+    view.validity_ = validity;
+    return view;
+  }
+
+  // A column carrying a type per row rather than one for the whole column.
+  static ColumnView Variants(const Variant* data) {
+    ColumnView view;
+    view.kind_ = Kind::kVariant;
+    view.data_ = data;
+    return view;
+  }
+
+  Kind kind() const { return kind_; }
+  StorageType type() const { return type_; }
+  RowSelection selection() const { return selection_; }
+
+  // Composes `selection` with the selection already applied, materializing the
+  // result into a block of `pool` if one is needed. The ordinals in
+  // `selection` must be strictly increasing.
+  void Slice(RowSelection selection, uint32_t count, SelectionPool& pool);
+
+  // Points this column at physical rows the caller owns. Nothing is copied, so
+  // `rows` has to outlive the batch's current contents.
+  void SetBorrowedRows(Span<const uint32_t> rows) {
+    selection_ = RowSelection::Indices(rows);
+    block_ = nullptr;
+  }
+
+  // Forgets which batch's block this selection was composed into. A batch
+  // adopting a column from another batch must call this, or narrowing the
+  // column would write into the other batch's storage.
+  void DisownBlock() { block_ = nullptr; }
+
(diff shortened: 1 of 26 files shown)


src/trace_processor/core/util/bit_vector.h
```diff
@@ -268,6 +269,31 @@ struct BitVector {
   // The && qualifier allows reusing the existing words_ buffer in-place.
   BitVector Compact(const BitVector& keep) && { return CompactInPlace(keep); }
 
+  // Sets bits [at, at + count) from `src` bits [from, from + count). Bits in
+  // the destination range which are unset in the source are left alone, so the
+  // range has to start clear to be a copy rather than a merge.
+  //
+  // Word at a time: a bit at a time serialises on the read-modify-write of the
+  // destination word, which costs more than the whole rest of an append.
+  void SetBitsFrom(uint64_t at,
```
LalitMaganti, line 279: This should be improved because we can break it up nito a "before", "during" and "after" and avoid all the min/max div operations etc.

src/trace_processor/core/exec/row_store.h
```diff
@@ -0,0 +1,109 @@
...
+  // Points `batch` at the rows `rows` picks out, in that order. They can come
+  // from any chunk, which no single view can span, so the values are gathered
+  // into reused storage. The result stays valid until the next indexed View()
+  // on this store.
+  void View(RowBatch* batch, Span<const uint32_t> rows);
+
+  // Drops all the rows but keeps the allocated chunks.
+  void Clear();
+
+  // kChunkRows rows of one column. Public only so the .cc can name it.
```
LalitMaganti: This seems like a hack and should be fixed.

src/trace_processor/core/exec/row_store.h
```diff
@@ -0,0 +1,109 @@
...
+        values{FlexVector<uint32_t>()};
+    BitVector validity;
+  };
+
+ private:
+  struct Column {
+    StorageType type{Uint32{}};
+    bool variant = false;
+    bool nullable = false;
+    std::vector<std::shared_ptr<Chunk>> chunks;
```
LalitMaganti: Why is this a shared_ptr?

src/trace_processor/core/exec/variant.h
```diff
@@ -0,0 +1,81 @@
...
+  double AsDouble() const {
+    PERFETTO_DCHECK(type == Type::kDouble);
+    return double_;
+  }
+  StringPool::Id AsString() const {
+    PERFETTO_DCHECK(type == Type::kString);
+    return string_;
+  }
+
+  Type type;
```
sashwinbalaji, line 73: nit: type = kNull

src/trace_processor/core/exec/row_batch.cc
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+bool RowBatch::Slice(RowSelection selection, uint32_t count) {
+  PERFETTO_DCHECK(count <= cardinality_);
+  for (uint32_t row = 0; row < count; ++row) {
+    PERFETTO_DCHECK(selection.GetIndex(row) < cardinality_);
+    PERFETTO_DCHECK(row == 0 ||
+                    selection.GetIndex(row - 1) < selection.GetIndex(row));
+  }
```
sashwinbalaji, line 69: nit: wrap in PERFETTO_DCHECK_IS_ON to be sure

src/trace_processor/core/util/bit_vector.h
```diff
@@ -279,8 +305,14 @@ struct BitVector {
...
   }
 
+  // Makes room for `new_size` bits without changing the size. Growth here is
+  // geometric where resize allocates exactly what was asked for, so anything
+  // filling a bit vector a chunk at a time has to come through here first.
+  void reserve(uint64_t new_size) { words_.reserve((new_size + 63) / 64); }
+
   // Resizes the vector to the specified size. If shrinking, bits past the new
   // size are cleared. If growing, new bits are set to the given value.
+
```
sashwinbalaji: nit:remove newline

src/trace_processor/core/exec/row_cursor.h
```diff
@@ -0,0 +1,78 @@
...
+  // Rewinds and moves to the first row.
+  bool Open() {
+    source_.Rewind(*state_);
+    index_ = 0;
+    size_ = 0;
+    return Pull();
+  }
+
+  // Moves to the next row, pulling a new batch if this one is exhausted.
+  bool Next() { return ++index_ != size_ || Pull(); }
```
sashwinbalaji: nit: what if we are already at eof ?

src/trace_processor/core/exec/column_view.h
```diff
@@ -0,0 +1,139 @@
...
+#include "src/trace_processor/core/exec/variant.h"
+#include "src/trace_processor/core/util/bit_vector.h"
+
+namespace perfetto::trace_processor::core::exec {
+
+// One column of a RowBatch: a reference to storage the batch does not own,
+// plus the row selection currently applied to it.
+class ColumnView {
+ public:
+  enum class Kind : uint8_t {
```
sashwinbalaji, line 36: maybe in future how about constant_vector (simiale to https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/common/enums/vector_type.hpp#L18)

src/trace_processor/core/exec/row_batch.h
```diff
@@ -0,0 +1,110 @@
...
+    owners_ = other.owners_;
+    cardinality_ = other.cardinality_;
+    selections_.Reset();
+    for (ColumnView& column : columns_) {
+      column.DisownBlock();
+    }
+  }
+
+  // Replaces `column` and the owner keeping its values alive.
+  void SetColumn(uint32_t column,
```
sashwinbalaji, line 65: don't we need to call DisownBlock here also ?

src/trace_processor/core/exec/row_store.cc
```diff
@@ -0,0 +1,340 @@
...
+  bool variant = from.kind() == ColumnView::Kind::kVariant;
+  StorageType type = from.type();
+  if (!variant && type.Is<Id>()) {
+    type = StorageType{Uint32{}};
+  }
+  bool nullable = into.nullable || from.validity() != nullptr;
+  if (nullable && !into.nullable) {
+    for (const std::shared_ptr<Chunk>& chunk : into.chunks) {
+      if (chunk) {
+        chunk->validity = BitVector::CreateWithSize(kChunkRows, true);
```
sashwinbalaji: we are setting validity true for all slots including the future unwritten ones. so CopyValidity would persist it and if any null comes next we would treat it as non-null?

src/trace_processor/core/exec/pipeline.cc
```diff
@@ -0,0 +1,106 @@
...
+  }
+  return source_.status(*s.source);
+}
+
+bool Pipeline::GetData(RowBatch& out, OperatorState& state) const {
+  State& s = state.Cast<State>();
+  if (operators_.empty()) {
+    return source_.GetData(out, *s.source);
+  }
+  for (;;) {
```
sashwinbalaji, line 76: should we add (or in future) some interrupt check to support cancel ?

sashwinbalaji (approved):

PR #7136, opened 2026-08-19
 base: use intptr_t for ssize_t on Windows 

Archkon:
Use intptr_t for Perfetto's Windows ssize_t definition. This preserves the width and value range on all Windows architectures and makes the definition compatible with common embedders.

Electron has carried this downstream fix for Win32 builds since 2022.

### DIFF
--- a/include/perfetto/ext/base/sys_types.h
+++ b/include/perfetto/ext/base/sys_types.h
@@ -41,11 +41,7 @@ using gid_t = int;
 using pid_t = int;
 #endif  // !GCC
 
-#if defined(_WIN64)
-using ssize_t = int64_t;
-#else
-using ssize_t = long;
-#endif  // _WIN64
+using ssize_t = intptr_t;
 
 #endif  // __MINGW32__
 


LalitMaganti (approved):

primiano:
Do you have any link to the Electron's code you are mentioning?

From https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types I see

```
SSIZE_T | A signed version of SIZE_T
This type is declared in BaseTsd.h as follows:
typedef LONG_PTR SSIZE_T;
``` 

followed by

```
#if defined(_WIN64)
 typedef __int64 LONG_PTR; 
#else
 typedef long LONG_PTR;
#endif
```

So MSDN headers suggests it should be a long

Archkon:
https://github.com/electron/electron/blob/main/patches/perfetto/define_ssize_t_to_be_intptr_t_to_match_libuv.patch
The crux is the define would conflict with libuv and maybe not only limited to electron as nodejs is working on integrating with perfetto

primiano:
Hmm I see.
This is going to be tricky. I did some research.

  - MinGW-w64 basetsd.h:148: typedef LONG_PTR SSIZE_T → long on Win32 ✅ (matches MSDN)
  - MinGW-w64 corecrt.h:43-51: #ifndef _SSIZE_T_DEFINED ... typedef int ssize_t; on Win32, __int64 on Win64

  So the same headers deliberately use long for the Win32 API type and int for the POSIX-ish lowercase one. libuv's intptr_t matches the CRT exactly on both
  bitnesses, and it's guarded with _SSIZE_T_/_SSIZE_T_DEFINED and also supplies SSIZE_MAX. Perfetto's is long, unguarded, and doesn't set _SSIZE_T_DEFINED, so it
  can never lose or defer a race.

so mingw seems to be on your side

-  CPython uses _W64 int on Win32;
- libssh2 does typedef SSIZE_T ssize_t (the long camp) guarded by HAVE_SSIZE_T/defined(ssize_t) rather than
  _SSIZE_T_DEFINED, so libssh2 collides with libuv too.

Also worth knowing:

This already landed once: #4800 (Feb 2026) changed it to int, then #4830 reverted the whole PR as it broke  google3 internal build.

I would probably do this instead

```
  // intptr_t is `int` on 32-bit and `long long` on 64-bit Windows. This matches
  // the MinGW CRT (corecrt.h) and libuv, which both define ssize_t this way.
  // Using `long` on 32-bit clashes with those (C2371) as int != long in C++.
  #if !defined(_SSIZE_T_) && !defined(_SSIZE_T_DEFINED)
  using ssize_t = intptr_t;
  #define _SSIZE_T_
  #define _SSIZE_T_DEFINED
  #endif
```

Archkon:
libuv also define SSIZE_MAX  in guard block so your example code would cause some problem 
how about just  using ssize_t = intptr_t; as what original electron patch do to be easier to maintain?

Archkon:
@primiano What is the correct way to push this PR forward ?

LalitMaganti:
Personally I think the smaller patch is the way to go (I.e. ssize_t = intptr_t).

primiano:
Why still int and not intprt_t as I suggested?
As I wrote, you are re-proposing #4800 which got reverted.
This will likely get reverted again

Plz look ay my comment above https://github.com/google/perfetto/pull/7136#issuecomment-5344828014

Archkon:
I believe the intptr_t is the type alias for int on win32.The #4800 also changes many other codes and I can't access the google3 link what #4830 metioned. So would it really work if change the int to intptr_t to avoid trigger the broken in internal build  ?

LalitMaganti:
I believe that yes it should work. Internally I don't think anyone compiles for win32

LalitMaganti (approved):

sashwinbalaji (approved):

PR #7138, opened 2026-08-19
perfetto: update frameworks/base tracing protos

parthsane:
Add AndroidProcessStateSnapshot and AndroidFreezerStateSnapshot AndroidSelfBroadcastEvent trace packet extension protos. Extract UnfreezeReason enum into
frameworks_base_common.proto and register track event extensions.

Mirrors frameworks/base/proto/tracing/frameworks_base_track_event.proto in the upstream Android tree.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -2982,6 +2982,7 @@ cc_test {
         ":perfetto_protos_perfetto_trace_translation_zero_gen",
         ":perfetto_protos_third_party_android_art_art_heap_graph_zero_gen",
         ":perfetto_protos_third_party_android_connectivity_connectivity_network_trace_zero_gen",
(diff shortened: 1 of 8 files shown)


primiano:
Out of curiosity why do you need this CL?
these days you should touch directly the protos in android internal repo.
Everything should work automatically from there

parthsane:
I was under the impression that this would be required if we wanted to add special parsing for this in trace processor, is that not the case?

primiano:
> I was under the impression that this would be required if we wanted to add special parsing for this in trace processor, is that not the case?


yes if that's what you want to do you have to sync them. Lalit explained me you were after this after I commented here.

primiano (approved):

PR #7143, opened 2026-08-19
tp: run dataframe scans with a filter on the operator executor

LalitMaganti:
Routes real queries through the interfaces added previously: a scan with
numeric filters now runs as a tree of streaming operators rather than a
bytecode program.

The predicate loop is not rewritten. Two implementations of a predicate is one
more than a query engine should have, since any difference between them is a
difference in what a query returns, so the interpreter's loop moves to
core/util and both backends call it.

A comparison narrows the caller's value to the column's type itself, since it
is the only thing that knows what that type is. The rules for doing so move
out of the interpreter and next to the fetcher, so neither backend restates
them. Narrowing is also where a comparison can stop being about a value at all
-- 1e300 as an int32 matches nothing -- so an operator can decline when armed,
and a query that cannot match ends before a chunk is read.

The lowering reads the logical plan and ignores the planner's access-path
strategy, except where that strategy is sub-linear: turning a binary search
into a full scan is ruinous for a cursor a join runs once per outer row.
Anything it does not accept stays on the interpreter, and that set shrinking is
how the executor takes over.


LalitMaganti:
Dropping this approach; the work is not being pursued.

PR #7145, opened 2026-08-19
tp: make the dataframe value fetcher a virtual interface

LalitMaganti:
ValueFetcher was a duck-typed base with declared-but-undefined methods, so
every consumer had to be a template just to name it. Cursor and Interpreter
were templated for no other reason, which meant their implementations lived in
headers and the type leaked into every caller.

It is now an ordinary interface. Filter values are fetched a handful of times
per execution, so the dispatch costs nothing next to running the query, and
Cursor and Interpreter become plain classes with their bodies in a .cc.

Emitting one interpreter instead of one per fetcher takes 70KB off the
stripped shell.

The dataframe builders keep the duck-typed form. There is a single
instantiation of the row builder, so an interface would remove no code -- it
measured 144 bytes larger and no faster -- while putting a virtual call on
every cell of every row built.

The type tags are SQLite's own, so the fetcher SQLite backs returns its source
type rather than translating it.


LalitMaganti:
Dropping this approach; the work is not being pursued.

PR #7147, opened 2026-08-20
tp: recycle row batches through a pool

LalitMaganti:
An operator which holds several chunks at once -- anything that has to
see every row before it can emit one -- cannot borrow the source's,
because a source refills its chunk on the next pull. It needs chunks of
its own, and allocating one per chunk of input makes the amount of
allocation a property of the input size rather than of the query.

A pool makes it a property of the query: an operator holding four
chunks at a time allocates four, whatever it is run over. Batches come
back reset, so nothing carries over from one query into the next.


LalitMaganti:

PR #7148, opened 2026-08-20
tp: read a row cursor's rows per column

LalitMaganti:
A cursor took the row view from column zero and applied it to every
other column, and RowBatch said in as many words that all of them shared
one. Nothing enforced it, and the first operator to want otherwise
breaks quietly: an operator which adds a computed column has nowhere to
put it in the index space its input arrived in, because that space
belongs to the storage the input was read from. It adds the column in
its own space, the two disagree, and the cursor reads the new column at
the old column's rows -- off the end of it, in release, with no
complaint.

Columns of one batch now agree on how many rows there are and on nothing
else. The cost is resolving a row per column instead of once per batch,
which a row-at-a-time reader can afford; what it buys is that adding a
column stops being a thing which happens to work when the offset is
zero.

### DIFF
--- a/src/trace_processor/core/exec/column_view.h
+++ b/src/trace_processor/core/exec/column_view.h
@@ -77,6 +77,9 @@ class ColumnView {
     block_ = other.block_;
   }
 
+  // The values this column reads from, before its row view is applied.
+  const void* data() const { return data_; }
+
  private:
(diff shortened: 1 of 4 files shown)


LalitMaganti:

PR #7149, opened 2026-08-20
tp: let a batch own the values it carries

LalitMaganti:
A column view is a view, and a batch built from views is only good for
as long as the storage behind them stands still. That is true of a
source reading a dataframe and false of one that materialises rows into
buffers it refills, which is what reading from SQL will do. Anything
holding a batch past the pull that produced it -- every breaker, by
definition -- would be reading whatever the source wrote next.

Owning the values is optional and decided per column by the operator,
because the operator is the only thing that knows how long it needs them
for. A streaming step still adds views and pays nothing.

The copy lands dense from row zero, which settles a second question: an
operator adding a computed column has nowhere to put it while its
input's index space belongs to the storage that input was read from.
Once the payload is the operator's own, the computed column sits beside
it in the same space.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -17601,6 +17601,7 @@ filegroup {
     srcs: [
         "src/trace_processor/core/exec/column_view.cc",
         "src/trace_processor/core/exec/from.cc",
+        "src/trace_processor/core/exec/owned_column.cc",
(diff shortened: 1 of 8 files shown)


LalitMaganti:

PR #7150, opened 2026-08-20
tp: add a sink and a breaker, and fold a tree over chunks

LalitMaganti:
An operator is handed a chunk and answers for that chunk, which is most
of what a pipeline does and not all of it. A sort, a subtree fold, or
anything else whose first row depends on the last one cannot be one, and
until now there was no name for what it is instead.

A breaker is a sink while it fills and a source once it is full. Where
the breakers are is where a pipeline stops being a pipeline: the memory
a query needs at once and the latency before its first row are decided
by them and by nothing else.

TREE ACCUMULATE UP is the first of them. Its state keeps the rows in a
store and one total per row; the plan is two column indices. The columns
it folds over are read back out of the store rather than copied aside,
so the input is held once however many columns the fold looks at.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -17599,6 +17599,7 @@ filegroup {
 filegroup {
     name: "perfetto_src_trace_processor_core_exec_exec",
     srcs: [
+        "src/trace_processor/core/exec/breaker.cc",
         "src/trace_processor/core/exec/column_view.cc",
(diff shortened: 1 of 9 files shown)


LalitMaganti:
Superseded: both accumulate directions now stream, so the sink and breaker are gone. The fold lives in https://github.com/google/perfetto/pull/7152.

PR #7151, opened 2026-08-20
tp: scan SQL queries into batches

LalitMaganti:
Queries which filter, join, group, or compute still need SQLite. SqlScan steps
a prepared statement and fills RowBatch instances instead of exposing the
row-at-a-time cursor to the rest of the pipeline.

Lineage decides how each result column is stored. Columns traced to a
dataframe come out flat at the proven type. Other columns use Variant because
SQLite may return a different type on each row.

Preparing, stepping, and rewinding the statement all follow the Source status
contract. Type validation is left to the operator which knows how a column
will be used.

### DIFF
--- a/src/trace_processor/perfetto_sql/exec/sql_scan.cc
+++ b/src/trace_processor/perfetto_sql/exec/sql_scan.cc
@@ -1,0 +1,350 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/trace_processor/perfetto_sql/exec/sql_scan.h"
+
+#include <sqlite3.h>
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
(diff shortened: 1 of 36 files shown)


src/trace_processor/perfetto_sql/exec/sql_scan.cc
```diff
@@ -0,0 +1,334 @@
...
+
+void SqlScan::Rewind(core::exec::OperatorState& state) const {
+  Prepare(state.Cast<State>());
+}
+
+bool SqlScan::ReadValue(State& s,
+                        sqlite3_stmt* stmt,
+                        uint32_t index,
+                        uint32_t row) const {
+  auto col = static_cast<int>(index);
```
LalitMaganti, line 224: This function is hot but inefficiently written. It should be improved.

src/trace_processor/perfetto_sql/exec/sql_scan.h
```diff
@@ -0,0 +1,117 @@
...
+ private:
+  // One column's values. Only the buffer matching the column's type is ever
+  // filled.
+  struct Column {
+    std::vector<core::exec::Variant> variants;
+    std::vector<uint32_t> uint32s;
+    std::vector<int32_t> int32s;
+    std::vector<int64_t> ints;
+    std::vector<double> doubles;
+    std::vector<StringPool::Id> strings;
```
sashwinbalaji: can these be a variant ?

src/trace_processor/perfetto_sql/exec/sql_scan.cc
```diff
@@ -0,0 +1,334 @@
...
+  void operator()(SyntaqliteParser* parser) const {
+    syntaqlite_parser_destroy(parser);
+  }
+};
+using ScopedParser = std::unique_ptr<SyntaqliteParser, ParserDeleter>;
+
+// The types lineage established, lined up with the query's columns. If the two
+// disagree on the number of columns they are not describing the same query, so
+// no type is claimed for any of them.
+std::vector<std::optional<StorageType>> Prove(const SqlSource& sql,
```
sashwinbalaji: can we name it like ResolveTypes

sashwinbalaji (approved):

PR #7153, opened 2026-08-20
tp: sort a stream of tree rows which arrives in no order at all

LalitMaganti:
Streaming and turning round between them cover a stream that is already
in one of the two tree orders. A stream in neither is not a harder case
of the same thing: no reordering of chunks and no reversal of rows
produces an order that was not in the input, so the rows have to be put
in one.

What makes that affordable is that the rows are held anyway, and held as
one column each. Sorting them is another array of row numbers, exactly
like turning them round is, and handing a batch over stays a matter of
pointing every column at a slice of it. Nothing is moved for any of the
three.

A tree also has to be a tree, and a stream in no order is the first one
where that is worth checking properly: a walk from the roots reaches
every node exactly once, and any it does not reach are in a cycle.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -18771,6 +18771,22 @@ filegroup {
     ],
 }
 
+// GN: //src/trace_processor/perfetto_sql/exec:exec
+filegroup {
+    name: "perfetto_src_trace_processor_perfetto_sql_exec_exec",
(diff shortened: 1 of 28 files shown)


LalitMaganti:

PR #7155, opened 2026-08-20
tp: say what a row batch is and where rows are kept

LalitMaganti:
Who owns the values a pipeline is carrying was never written down, and
the answer was drifting: some columns pointed into a source's storage,
some into a batch's, and an operator that wanted rows for longer than a
batch lasts declared whatever vectors it felt like. That is the sort of
thing which is fine until two operators disagree about it.

There are two rules and this is both of them.

A batch owns nothing. Its columns are views over storage belonging to
whatever produced them, so a batch is good until the next pull from that
source and not one instruction longer.

An operator that needs the rows after that copies them into a RowStore
it names as a member. There is no other way to keep rows, so the memory
a query holds is the sum of its stores, and finding it is reading the
operators' member lists.

Two things follow. Columns of one batch stop having to agree about which
rows they mean: a cursor took the row view from column zero and applied
it to the rest, which quietly reads a computed column off the end of
itself, because the space its input arrived in belongs to the storage
the input was read from. And a store reads back dense from row zero, so
a fold over a column is a walk over an array and a computed column can
sit beside it in the same space.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -17604,6 +17604,7 @@ filegroup {
         "src/trace_processor/core/exec/pipeline.cc",
         "src/trace_processor/core/exec/row_batch.cc",
         "src/trace_processor/core/exec/row_cursor.cc",
+        "src/trace_processor/core/exec/row_store.cc",
     ],
 }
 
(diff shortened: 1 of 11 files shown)


LalitMaganti:
Superseded: the plan/state split replaces this.

PR #7157, opened 2026-08-20
tp: make an operator a plan and move what it is doing out of it

LalitMaganti:
An operator held what it was in the middle of doing: how far through its
input it had got, the buffers it was filling, the batch it was handing
out. That makes an operator a thing you can only use once, ties the
lifetime of a query's values to the lifetime of the objects describing
the query, and leaves no answer to the question of who owns what beyond
"whoever happens to have declared it".

So an operator is now a plan and nothing else. It is const while it
runs, holds no values, and answers one question about itself: what state
does an execution of you need. The state is made by the operator, owned
by whatever is running it, and holds everything that changes -- which
for a source that materialises rows is also the values, and for a
breaker will be everything it has buffered.

The consequences are the point. A plan can be run twice, or twice at
once, and the two runs cannot see each other. Dropping a plan cannot
invalidate a running query's data, because the plan never had any.
Where the memory went is answerable by looking at the states rather than
at the objects. And a batch stops being something a source hands out and
becomes something the executor owns and asks to have filled, so the
answer to how long a batch is good for stops depending on which operator
produced it.

The cursor is the executor: it makes the state, owns the batch, and
holds the plan by const reference. Rewriting it also fixes the row view
it was handing out, which came from column zero and was applied to every
other column -- fine until an operator adds a computed column, which has
nowhere to go in the index space its input arrived in. Columns of one
batch now agree on how many rows there are and on nothing else.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -17601,6 +17601,7 @@ filegroup {
     srcs: [
         "src/trace_processor/core/exec/column_view.cc",
         "src/trace_processor/core/exec/from.cc",
+        "src/trace_processor/core/exec/operator.cc",
         "src/trace_processor/core/exec/pipeline.cc",
         "src/trace_processor/core/exec/row_batch.cc",
         "src/trace_processor/core/exec/row_cursor.cc",
--- a/Android.bp
+++ b/Android.bp
@@ -17595,6 +17595,26 @@ filegroup {
     ],
 }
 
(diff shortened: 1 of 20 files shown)


LalitMaganti:
Folded into #7135, where the model it changes is introduced.

PR #7158, opened 2026-08-20
tp: keep an operator's rows in a store rather than in vectors

LalitMaganti:
A plan holds no values and a batch is a borrow, which leaves the
question of where the values are. They are in a state, and this is the
one thing a state keeps rows in.

The alternative is what was happening before: whichever operator needed
rows for longer than a batch lasts declared whatever vectors it felt
like, so the same column ended up stored twice under different names and
the memory a query needed at once was not findable without reading every
operator. A store makes it one thing, once, in one place.

Growing it is the whole of why it is a class rather than a vector.
Resize allocates exactly what it is asked for and only reserve grows
geometrically, so filling a column a chunk at a time reallocates and
recopies the column on every chunk -- quadratic, and invisible until the
input is big. Two million rows of one column took 2.7 seconds and now
take 23 milliseconds. The test for it counts how often the buffer moves
rather than how long it takes, so it does not depend on a clock.

The rows read back dense from row zero, so a fold over a column is a
walk over an array, and a column computed from them can sit beside them
in the same index space.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -17595,6 +17595,27 @@ filegroup {
     ],
 }
 
+// GN: //src/trace_processor/core/exec:exec
+filegroup {
+    name: "perfetto_src_trace_processor_core_exec_exec",
+    srcs: [
+        "src/trace_processor/core/exec/column_view.cc",
+        "src/trace_processor/core/exec/from.cc",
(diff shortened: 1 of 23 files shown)


LalitMaganti:
Folded into #7135: it is the same model.

PR #7165, opened 2026-08-21
tp: scan dataframes without SQLite

LalitMaganti:
A query which only reads a dataframe gives SQLite nothing to do. The values
already have storage types and are laid out the way a batch needs them.

DataframeScan returns views over the dataframe's storage without copying.
Columns without one slot per row are expanded one batch at a time into a
reused buffer, so they still look dense to downstream operators.

The planner only takes this path for passthrough dataframes. Filtering,
joining, grouping, and expressions still go through SQLite.

### DIFF
--- a/src/trace_processor/core/exec/BUILD.gn
+++ b/src/trace_processor/core/exec/BUILD.gn
@@ -18,6 +18,8 @@ source_set("exec") {
   sources = [
     "column_view.cc",
     "column_view.h",
+    "dataframe_scan.cc",
+    "dataframe_scan.h",
     "operator.cc",
     "operator.h",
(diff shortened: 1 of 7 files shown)


src/trace_processor/core/exec/BUILD.gn
```diff
@@ -36,6 +38,7 @@ source_set("exec") {
     "../../../base",
     "../../containers",
     "../common",
+    "../dataframe",
```
sashwinbalaji, line 41: hmm anyway to avoid exec(generic) depending on dataframe ?

sashwinbalaji (approved):

PR #7166, opened 2026-08-21
tp: fold tree values in a streaming pipeline

LalitMaganti:
Once rows are in tree order, a fold only needs one running value per node.
TreeAccumulateUp consumes child-first rows and adds each completed child into
its parent. TreeAccumulateDown consumes parent-first rows and carries each
parent total into its children.

Both operators append the accumulated value and stream batches without keeping
another copy of the input.

The SqlScan test covers the complete path from SQLite variants through type
validation, node numbering, ordering, and accumulation.


src/trace_processor/core/exec/tree_accumulate.h
```diff
@@ -0,0 +1,95 @@
...
+#include "src/trace_processor/core/exec/operator.h"
+#include "src/trace_processor/core/exec/row_batch.h"
+#include "src/trace_processor/core/util/flex_vector.h"
+
+namespace perfetto::trace_processor::core::exec {
+
+// The columns holding the tree structure and the values being summed. Node and
+// parent columns must be flat, non-null Uint32 columns; values must be flat
+// Int64. A null value contributes zero.
+struct AccumulateSpec {
```
sashwinbalaji, line 34: how about TreeAccumulateSpec as it's in exec namespace

src/trace_processor/core/exec/tree_accumulate.h
```diff
@@ -0,0 +1,95 @@
...
+  uint32_t node_column = 0;
+  uint32_t parent_column = 1;
+  uint32_t value_column = 2;
+};
+
+// What one execution carries between batches: a running total per node, and
+// the totals computed for the current batch.
+class AccumulateState : public OperatorState {
```
sashwinbalaji, line 42: Do we need this to be in .h as we will anywya talk in terms of OperatorState right?

src/trace_processor/core/exec/tree_accumulate.cc
```diff
@@ -0,0 +1,228 @@
...
+  const uint32_t* nodes =
+      Flatten<uint32_t>(in.column(spec_.node_column), count, &s.node_scratch);
+  const uint32_t* parents = Flatten<uint32_t>(in.column(spec_.parent_column),
+                                              count, &s.parent_scratch);
+  const int64_t* values =
+      FlattenValues(in.column(spec_.value_column), count, &s.value_scratch);
+
+  s.totals->resize(count);
+  int64_t* totals = s.totals->data();
+  for (uint32_t row = 0; row < count; ++row) {
```
sashwinbalaji: hmm, as per https://github.com/google/perfetto/pull/7214:

```
Folding up a tree needs every child before its parent, and needs the
order to be a depth first post-order so the fold can carry a stack of
the current path instead of an array indexed by node. 
```

so can we use a current path stack here instead of by_node?

src/trace_processor/core/exec/tree_accumulate.cc
```diff
@@ -0,0 +1,228 @@
...
+    uint32_t node = nodes[row];
+    Grow(&s.by_node, node);
```
sashwinbalaji, line 170: ok so I was thinking why we need all the nodes (by_node) as I thought (maybe incorrectly) we wanted DFS post-order and also #7214 said,

```
Folding up a tree needs every child before its parent, and needs the order
to be a depth first post-order so the fold can carry a stack of the current
path instead of an array indexed by node.
```

But then I checked the pr again and see in tree_order some early exits which won't necessarily guarantee DFS post order right ? 

```
if (s.child_first) {
  // Already in the requested order, so no reordering is needed.
  s.order.clear();
  return true;
}
if (s.parent_first) {
  for (uint32_t row = 0; row < rows; ++row) {
    s.order[row] = rows - 1 - row;
  }
  return true;
}
```

Just to make sure, I asked AI to find a possible scenario and it said:

for example, we can have this tree:

```
      0
     / \
    1   2
    |   |
    3   4
```

`3, 4, 1, 2, 0` has every child before its parent, so we would pass it through, but it is not DFS post-order. A DFS post-order would be `3, 1, 4, 2, 0`.

So I guess this is why we need by_node here. Wanted to just make sure this is intentional. And, yes I see the comment on TreeParentFirst saying below so I think we just need to add similar comment for child first also.

```
// The order is parent first and nothing more: not a pre-order, so a fold
// down keeps a value per node rather than a path. The input columns are node
// numbers, which TreeNumberNodes produces. No column is added.
```

sashwinbalaji (approved):
There are CI failures as test written accoding to the old API

PR #7167, opened 2026-08-21
tp: validate variant column types

LalitMaganti:
A SqlScan column uses Variant when lineage cannot prove one storage type.
Typed operators need to check those values before reading them as a flat
column.

AssertType checks every selected row against the requested type and produces a
flat column. Nulls remain null. Narrow integers can widen when the conversion
is exact; other mismatches stop the pipeline and name the offending column.

This is an operator rather than part of SqlScan because the consumer, not the
source, decides what type a column must have.

### DIFF
--- a/src/trace_processor/core/exec/assert_type.cc
+++ b/src/trace_processor/core/exec/assert_type.cc
@@ -1,0 +1,318 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/trace_processor/core/exec/assert_type.h"
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+
+#include "perfetto/base/logging.h"
+#include "perfetto/base/status.h"
+#include "src/trace_processor/containers/string_pool.h"
+#include "src/trace_processor/core/common/storage_types.h"
+#include "src/trace_processor/core/exec/column_chunk.h"
+#include "src/trace_processor/core/exec/column_view.h"
+#include "src/trace_processor/core/exec/operator.h"
+#include "src/trace_processor/core/exec/row_batch.h"
+#include "src/trace_processor/core/exec/row_selection.h"
+#include "src/trace_processor/core/exec/variant.h"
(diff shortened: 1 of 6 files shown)


src/trace_processor/core/exec/assert_type.h
```diff
@@ -0,0 +1,86 @@
...
+
+  std::unique_ptr<OperatorState> MakeState() const override;
+  OpResult Execute(const RowBatch& in,
+                   RowBatch& out,
+                   OperatorState&) const override;
+  void Rewind(OperatorState&) const override;
+  base::Status status(const OperatorState&) const override;
+
+ private:
+  struct Values {
```
sashwinbalaji: simialr thing before ?

src/trace_processor/core/exec/assert_type.h
```diff
@@ -0,0 +1,86 @@
...
+// single type, failing on any row which disagrees.
+//
+// A source reading SQLite cannot promise a column's type, because a declared
+// type in SQLite is not binding. AssertType is needed for variant columns. A
+// flat column already of the requested type passes through unchanged; a flat
+// integer column of any other type is copied into a widened one. Integers
+// widen to a float only where the conversion is exact; nothing else converts.
+class AssertType : public Operator {
+ public:
+  AssertType(uint32_t column, AssertTypeTarget type, std::string name);
```
sashwinbalaji: converttype?

src/trace_processor/core/exec/assert_type.cc
```diff
@@ -0,0 +1,311 @@
...
+    case Variant::Type::kDouble:
+      return "a float";
+    case Variant::Type::kString:
+      return "a string";
+  }
+  return "something";
+}
+
+const char* Name(StorageType type) {
+  if (type.IsAnyOf<IntegerType>()) {
```
sashwinbalaji: nit:this also switch like above one

src/trace_processor/core/exec/assert_type.cc
```diff
@@ -0,0 +1,311 @@
...
+    };
+    return WidenAs<int64_t>(column, count, exact, &values.doubles,
+                            &values.validity);
+  }
+  PERFETTO_DCHECK(from.Is<Id>() || from.Is<Uint32>());
+  return WidenAs<uint32_t>(column, count, Cast<uint32_t, double>,
+                           &values.doubles, &values.validity);
+}
+
+OpResult AssertType::Execute(const RowBatch& in,
```
sashwinbalaji, line 229: so each column will invoke asserttpye separately ? just thinking anyway to batch that to avoid the row copying for each col separately

sashwinbalaji (approved):

src/trace_processor/core/exec/assert_type.cc
LalitMaganti: Each AssertType only touches its own column. CopyFrom on the RowBatch is a shallow copy of the column views and owners, not of row data. A flat column of the right type passes through with no per-row work; only a variant or narrow integer column has its values materialised, and that is a single pass over that one column. So chaining one per column costs N shallow descriptor copies rather than N passes over the rows. Happy to add a multi-column variant later if a consumer needs it.

PR #7182, opened 2026-08-21
tp: record responsible thread on process-scoped counters

zezeozue:
Counters delivered in thread context but attributed to a process (rss_stat, oom_score_adj, mm_event, ...) go through PushProcessCounterForThread, which resolves the emitting utid to a upid and drops the thread. The thread is often the signal you want ("which thread grew the process's anon RSS"), so record it as a "utid" arg on each counter sample.

The counter stays process-scoped; the responsible thread now survives in counter.arg_set_id, is queryable via EXTRACT_ARG, and shows in the UI details panel with no UI change -- mirroring the utid/end_utid args already attached to async slices in systrace_parser.

### DIFF
--- a/src/trace_processor/importers/common/event_tracker.cc
+++ b/src/trace_processor/importers/common/event_tracker.cc
@@ -30,7 +30,8 @@
 namespace perfetto::trace_processor {
 
 EventTracker::EventTracker(TraceProcessorContext* context)
-    : context_(context) {}
+    : context_(context),
+      thread_arg_key_(context->storage->InternString("utid")) {}
 
 EventTracker::~EventTracker() = default;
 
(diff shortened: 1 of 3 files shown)


LalitMaganti:
There a serious problem with this: it's going to add a *lot* of memory for storing this data, especially because you're putting it in the args table. 

Do we really need this?

zezeozue:
> There a serious problem with this: it's going to add a _lot_ of memory for storing this data, especially because you're putting it in the args table.
> 
> Do we really need this?

It's pretty handy to see what threads are allocating anon or file pages. You can often guess from looking at CPU but it's hard to write queries against. I don't know if that's worth the cost

LalitMaganti:
Could we quantify the memory bloat a bit?

zezeozue:
> Could we quantify the memory bloat a bit?

From my synthetic trace measurements, the size seems to scale by about 5bytes per process counter sample. 

     samples     delta        
     600,000     3.11 MB      
     2,400,000  10.49 MB    
     9,600,000  51.20 MB    

From AOT traces, the size seems to be about ~100k samples for about 15s of sched. Found one with 200k samples.

```
select count() from counter
join process_counter_track p on p.id = counter.track_id 
```

This comes out to under 1MB increase in memory for AOT traces.

LalitMaganti (approved):

PR #7184, opened 2026-08-22
proto: change WindowManager child_tokens format

rukkal:
This commit changes the WindowManager's child_tokens field from a
repeated int32 to a repeated IdentifierProto message.
Thus the WindowManager incremental tracing can signal to the ProtoVM
when the list becomes empty, emitting an empty IndentifierProto message.

This is a breaking change, but acceptable as the field is not being used
yet.

Bug: 512718675


LalitMaganti (approved):

protos/third_party/android/frameworks/base/proto/tracing/winscope/server/windowmanagerservice.proto
```diff
@@ -622,7 +627,7 @@ message WindowContainerProto {
   optional IdentifierProto identifier = 6;
   optional SurfaceControlProto surface_control = 7;
   optional bool visible_requested = 8;
-  repeated int32 child_tokens = 9;
+  repeated IdentifierProto child_tokens = 9;
```
sashwinbalaji, line 630: this will cause issues I think.

- stable UI still thinks child_tokens is int32, so any new traces with IdentifierProto type will crash on it
- similarly, in autopush if it's IdentifierProto, the traces collected till now will have the old int type and will crash

rukkal: As discussed offline, the issue is that even if there exist no traces containing the field (it is not used yet), we do have traces containing the field descriptor

PR #7186, opened 2026-08-23
ui: Handle null navigator.keyboard

iw4p:
## Problem
Screenshot 2026-08-23 at 23 05 46

In some browsers, the `keyboard` property exists on `navigator` but its value is `null`.

The existing property-presence check succeeds and then calls `getLayoutMap()` on `null`, crashing the Perfetto home page:

> TypeError: Cannot read properties of null (reading 'getLayoutMap')

Observed with Perfetto UI v58.2 on Chrome 150 on macOS.

## Fix

Declare `navigator.keyboard` as nullable and check its value before calling `getLayoutMap()`.

Unsupported environments now throw the existing `NotSupportedError`, allowing the UI to use its fallback keyboard layout.

I solved it with declare global because I saw they've been used in other parts, could be fixed with cast but ugly, or runtime guard which's verbose..

Additionally getLayoutMap() genuinely returns a Promise<KeyboardLayoutMap> according to the [Keyboard Map specification](https://wicg.github.io/keyboard-map/#keyboard-interface), this is the reason I changed it.

### DIFF
--- a/ui/src/base/keyboard_layout_map.ts
+++ b/ui/src/base/keyboard_layout_map.ts
@@ -18,7 +18,13 @@ export interface KeyboardLayoutMap {
 }
 
 interface Keyboard {
-  getLayoutMap(): KeyboardLayoutMap;
+  getLayoutMap(): Promise<KeyboardLayoutMap>;
+}
+
+declare global {
+  interface Navigator {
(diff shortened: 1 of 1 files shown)


stevegolton (approved):
Thanks for the patch!

safayat-google (approved):

PR #7188, opened 2026-08-24
ui: Add Google Drive integration and 'Open from GDrive' sidebar button

stevegolton:
This patch adds a new non-default enabled Google Drive (core) plugin, which adds the following features:

- Open a trace directly from Google drive.
- Allow opening pftrace files in Perfetto directly from drive (after initial setup).

> Note: Saving and sharing will be added in a followup PR.

For now these features are only available for Googlers - and app secrets are delivered by a GCS hosted .js file with internal ACLs applied.

# Details

## Open a trace from google drive

This feature adds a new 'Open from Gdrive' button in the sidebar, next to the 'Open from file' button.

image

This launches the Perfetto drive authentication workflow, which installs Perfetto. The permissions asked for are low risk:
- Installs itself as the default opener for various different file extensions - (.trace, .pftrace, .perfetto-trace, + .gz variants).
- Allows Perfetto to read specific files that the user has granted access to. *This bit is crucial, perfetto cannot help itself to any file, or even see the directory structure - it's only allowed to access files that the user specifically decides to share with it*.

gdrive-auth

Once set up, the user can pick a file using the native drive file picker:
gdrive-picker

## Allow opening pftrace files in Perfetto directly from drive

After Perfetto is registered with drive, users can now open files directly from the drive UI. Any file with an extension like .trace, .pftrace, .perfetto-trace, + .gz variants will show up with an 'Open in Perfetto' button when inspecting that specific file in drive.

image

Note: Since the landing url is ui.perfetto.dev - this feature will not work until this patch is merged.


ui/src/core_plugins/dev.perfetto.GoogleDrive/index.ts
```diff
@@ -0,0 +1,279 @@
...
+import {GoogleDriveClient, type GoogleDriveFile} from './gdrive_client';
+import type {Result} from '../../base/result';
+import {defer} from '../../base/deferred';
+import {Intent} from '../../widgets/common';
+
+// The Google Drive credentials are not checked into the repo. They are loaded
+// at runtime from an internal script (the same way the internal userscript is
+// loaded) via a <script> tag, which sidesteps the CORS restrictions a
+// cross-origin fetch() would hit. The script populates window.gDriveConfig.
+const CONFIG_URL =
```
LalitMaganti, line 30: Could we please plumb this from the embedder server instead of doing this?

stevegolton: Discussed offline. We can't put this in the embedders server as we need it to be available for everyone (eventually), and there is no global embedders server yet.

Since this is a one-off for the drive integration only, we'll leave it being fetched from the plugin specifically. Embedders not hosting on ui.perfetto.dev can just disable the plugin - as it won't work anyway. If we ever move drive integration into the core we'll make it part of the embedders API so that embedders can customise the project used.

LalitMaganti: +1 but as agreed we should move this to the perfetto-ui bucket *and* ideally make this a JSON object which can be fetched instead of JS object which needs to be run.

PR #7207, opened 2026-08-25
perfetto: renumber recovered_trace_info field ID to 138

BogdanLionte:
Update RecoveredTraceInfo field ID in TracePacket from 136 to 138 and regenerate merged protos, C-SDK headers, and Python protobuf bindings.

### DIFF
--- a/include/perfetto/public/protos/trace/trace_packet.pzc.h
+++ b/include/perfetto/public/protos/trace/trace_packet.pzc.h
@@ -576,7 +576,7 @@ PERFETTO_PB_FIELD(perfetto_protos_TracePacket,
                   MSG,
                   perfetto_protos_RecoveredTraceInfo,
                   recovered_trace_info,
-                  136);
+                  138);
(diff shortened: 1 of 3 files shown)


dragosp95 (approved):

safayat-google (approved):
LGTM.

Assuming that this is due to proto merger getting conflicts on G3, We should think about a solution to avoid this types race conditions in the future.

primiano:
Can I ask what was the context/rationale of this change?
I don't have too many objections, but for changes like this it would be nice to say in the commit message why we are changing it,

primiano:
Oh is it for the presence of some old

```
   AndroidProcessStateSnapshot android_process_state = 136;
    AndroidFreezerStateSnapshot android_freezer_state = 137;
```

in the internal logs proto?
In that case, this PR is fine, but please please:
- mark those fields as `[deprecated=true]` in google3
- mark those fields ids as reserved in the upstream proto adding a comment explaning that they have been used by accident in the internal logs proto.

Mistakes were made, and it's okay, but we should prevent that other people stumble on it.

PR #7208, opened 2026-08-25
proto_merger: Keep reserved fields of deleted types in deleted_fields

andrabivolaru:
When a field is reserved upstream and its type is a deleted nested message/enum (not present upstream), `proto_merger` previously moved it to `fields`, causing it to be serialized *before* its type definition.

This change keeps reserved fields of deleted types in `deleted_fields` with `[deprecated = true]` so they are serialized after their nested type definitions.

Added unit test `ReservedUpstreamFieldWithDeletedTypeIsPlacedInDeletedFields` in `proto_file_serializer_unittest.cc`.

### DIFF
--- a/src/tools/proto_merger/proto_file_serializer_unittest.cc
+++ b/src/tools/proto_merger/proto_file_serializer_unittest.cc
@@ -900,6 +900,90 @@ TEST(ProtoFileSerializerTest, MapFieldsDoNotGetRepeatedOrEntryMessages) {
   EXPECT_NE(merged_desc, nullptr);
 }
 
+TEST(ProtoFileSerializerTest,
+     ReservedUpstreamFieldWithDeletedTypeIsPlacedInDeletedFields) {
+  ProtoFile input;
+  {
+    ProtoFile::Message message{};
+    message.name = "Container";
+
+    ProtoFile::Message note_msg{};
+    note_msg.name = "Note";
+    note_msg.fields.push_back(MakeField("string", "key", 1));
+    note_msg.fields.push_back(MakeField("string", "value", 2));
+    message.nested_messages.push_back(note_msg);
+
+    ProtoFile::Field notes_field = MakeField("Note", "notes", 46);
+    notes_field.is_repeated = true;
+    message.fields.push_back(notes_field);
+
+    input.messages.push_back(message);
+  }
+
+  ProtoFile upstream;
+  {
+    ProtoFile::Message message{};
+    message.name = "Container";
+    message.reserved_numbers.insert(46);
+    upstream.messages.push_back(message);
+  }
+
+  ProtoFile merged;
+  ASSERT_TRUE(MergeProtoFiles(input, upstream, Allowlist{}, merged).ok());
+
+  std::string out = ProtoFileToDotProto(merged);
+
+  size_t pos_note_msg = out.find("message Note {");
+  size_t pos_notes_field = out.find("repeated Note notes = 46");
+
+  EXPECT_NE(pos_note_msg, std::string::npos);
+  EXPECT_NE(pos_notes_field, std::string::npos);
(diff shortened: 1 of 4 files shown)


dragosp95 (approved):

BogdanLionte (approved):

src/tools/proto_merger/proto_merger.cc
```diff
@@ -366,6 +366,7 @@ base::Status MergeFields(const std::vector<ProtoFile::Field>& input,
                          const std::unordered_set<int>& reserved_numbers,
                          const std::set<std::string>& known_enums,
                          const std::set<std::string>& allowlisted_options,
+                         const std::set<std::string>& deleted_type_names,
```
sashwinbalaji: nit: this and above 2 (known_enums and allowlisted_options) can be unordered_set +
can we create a struct for these params, it's upto 8 now, feels like time.

src/tools/proto_merger/proto_merger.cc
```diff
@@ -476,7 +479,7 @@ base::Status Merge(const ProtoFile::Oneof& input,
 
   // Finish by merging the list of fields.
   return MergeFields(input.fields, upstream.fields, allowlist, {}, known_enums,
-                     allowlisted_options, out.fields);
+                     allowlisted_options, {}, out.fields);
```
sashwinbalaji: hmm why not handle deleted_type_names here also ?

src/tools/proto_merger/proto_file_serializer_unittest.cc
```diff
@@ -882,6 +882,92 @@ TEST(ProtoFileSerializerTest, MapFieldsDoNotGetRepeatedOrEntryMessages) {
...
+  };
+
+  base::TempDir temp_dir = base::TempDir::Create();
+  std::string input_path = temp_dir.path() + "/input.proto";
+  std::string upstream_path = temp_dir.path() + "/upstream.proto";
+
+  ScopedUnlink unlink_input{input_path};
+  ScopedUnlink unlink_upstream{upstream_path};
+
+  std::string input_content = R"(
```
sashwinbalaji: nit: follow the way of MergeKeepsFieldsDeletedUpstream to avoid writing to actual files and to shorten the code a bit

src/tools/proto_merger/proto_merger.cc
andrabivolaru: Done.

src/tools/proto_merger/proto_merger.cc
```diff
@@ -399,29 +415,75 @@ base::Status MergeFields(const std::vector<ProtoFile::Field>& input,
...
+                   ProtoFile::Oneof& out) {
+  PERFETTO_CHECK(input.name == upstream.name);
+  out.name = input.name;
 
-base::Status Merge(const ProtoFile::Message& input,
-                   const ProtoFile::Message& upstream,
-                   const Allowlist::Message& allowlist,
-                   const std::set<std::string>& known_enums,
-                   const std::set<std::string>& allowlisted_options,
-                   ProtoFile::Message& out);
```
sashwinbalaji, line 416: hmm if we remove this forward-decl how will MergeRecursive work for message case ? Maybe I am missing somethign but is this building ?

andrabivolaru: Yes because MergeRecursive is a template: Merge is resolved at instantiation time, at which point Merge(Message) is already defined.

Happy to add the forward-declaration back if you prefer it for readability

sashwinbalaji: interesting, learned somethign new https://eel.is/c++draft/basic.lookup.argdep#4. Since it builds (and is in toos/) yeah I think fine to keep it as it's.

sashwinbalaji (approved):

PR #7212, opened 2026-08-25
tp: trace SQL columns to dataframes

LalitMaganti:
SQLite declarations do not bind result types, so the executor needs semantic
column origins before it can select flat storage.

Reuse the shared PerfettoSQL relation analyzer for that work. The analyzer owns
parsing view definitions and reports canonical leaf relation names as origins.
Views are parsed on demand for now; a TODO records that caching those parsed
definitions would avoid repeated parser allocation if this becomes measurable.

The trace processor adapter is limited to three boundary operations:
- expose a live dataframe as a leaf relation;
- return SQLite's stored CREATE VIEW SQL;
- map resolved leaf columns to dataframe storage types.

Aliases, subqueries, joins, compound queries, and row-origin semantics remain
owned by `src/perfetto_sql/analysis`.

### DIFF
--- a/src/perfetto_sql/analysis/relation.cc
+++ b/src/perfetto_sql/analysis/relation.cc
@@ -17,14 +17,16 @@
 #include "src/perfetto_sql/analysis/relation.h"
 
 #include <algorithm>
-#include <cctype>
 #include <cstddef>
+#include <cstdint>
 #include <memory>
 #include <optional>
+#include <string>
 #include <string_view>
 #include <utility>
 #include <vector>
 
+#include "perfetto/base/status.h"
 #include "perfetto/ext/base/status_macros.h"
 #include "perfetto/ext/base/status_or.h"
 #include "perfetto/ext/base/string_utils.h"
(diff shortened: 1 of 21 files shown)


src/trace_processor/perfetto_sql/lineage/connection_catalog.cc
```diff
@@ -0,0 +1,119 @@
...
+std::optional<core::StorageType> ConnectionCatalog::ColumnType(
+    const analysis::ColumnLineage& column) const {
+  std::optional<core::StorageType> result;
+  for (const analysis::ColumnOrigin& origin : column.origins) {
+    const dataframe::Dataframe* dataframe =
+        connection_->GetDataframeOrNull(std::string(origin.relation_name));
+    if (!dataframe) {
+      return std::nullopt;
+    }
+    std::string column_name(origin.column_name);
```
sashwinbalaji: hmm we can avoid this by providing the string_view overload of CaseInsensitiveEqual

src/perfetto_sql/analysis/relation.cc
```diff
@@ -171,8 +188,8 @@ class RelationAnalyzer::Impl {
                        std::string_view table,
                        std::string_view column) const;
 
-  const Program& program_;
   const Catalog& catalog_;
+  std::vector<std::unique_ptr<OwnedView>> views_;
```
sashwinbalaji, line 187: nit: pls add a comment explainign why it's unique_ptr

src/trace_processor/perfetto_sql/lineage/connection_catalog.cc
```diff
@@ -0,0 +1,119 @@
...
+
+}  // namespace
+
+ConnectionCatalog::ConnectionCatalog(PerfettoSqlConnection* connection)
+    : connection_(connection) {}
+
+std::optional<analysis::LeafRelation> ConnectionCatalog::FindLeafRelation(
+    std::string_view name) const {
+  const dataframe::Dataframe* dataframe =
+      connection_->GetDataframeOrNull(std::string(name));
```
sashwinbalaji: how will this work if there is a difference in case

sashwinbalaji (approved):

src/perfetto_sql/analysis/relation.cc
sashwinbalaji: how do we handle one side being a literal? 

AppendUniqueOrigins keeps only the other side's origins -> SqlScan::Prove() (#7151) passes tham as ColumnType. but should it be unknown ?

PR #7213, opened 2026-08-25
tp: assign dense tree node numbers

LalitMaganti:
Tree IDs can be sparse, strings, or row IDs from a table much larger than the
query result. Arrays indexed by those IDs would scale with the ID range rather
than the tree being processed.

TreeNumberNodes assigns stable dense numbers as rows arrive and appends the
numbered node and parent columns to each batch. A parent is numbered when first
referenced, so child-first input works even when the parent row appears later.

The original columns stay in the batch. Later tree operators get compact
indices without losing the values needed in the result.

### DIFF
--- a/src/trace_processor/core/exec/tree_number_nodes.h
+++ b/src/trace_processor/core/exec/tree_number_nodes.h
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


src/trace_processor/core/exec/tree_number_nodes.h
```diff
@@ -0,0 +1,105 @@
...
+  void Rewind(OperatorState&) const override;
+  base::Status status(const OperatorState&) const override;
+
+  // Whether every id seen so far was already its own node number.
+  bool IsDenseForTesting(const OperatorState&) const;
+
+ private:
+  struct Key {
+    int64_t value;
+    bool string;
```
sashwinbalaji: is_string

sashwinbalaji (approved):

PR #7216, opened 2026-08-25
Eliminate cuj_id remapping in android_jank_latency_cujs

nicomazz:
In android_jank_latency_cujs, remove the ROW_NUMBER() OVER (...) remapping and original_cuj_id column. Instead, make the table a direct UNION ALL of android_sysui_jank_cujs and android_sysui_latency_cujs, preserving the original cuj_id from the underlying jank and latency sources.

This ensures that only one place generates row_number() for Jank CUJs (_jank_cujs_slices) and only one place generates row_number() for Latency CUJs (android_sysui_latency_cujs). Duplicated cuj_id values across jank and latency rows in android_jank_latency_cujs are permitted.

If anyone does join on the table containing both latency and jank cujs, they are expected to also join on the cuj_type.

---- 

fully human part: In practise, this was not really an issue for jank cujs, even if they were accessed from `android_jank_latency_cujs`.. What was happening in `android_jank_latency_cujs` was that jank cuj were keeping the existing id, while latency cujs were remapped after them. We didn't really notice the problem as most of the infra we have is for jank cujs, and not for latency ones. In the blocking calls, the cuj id was not really used anywhere.

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
@@ -276,8 +276,6 @@ SELECT
   end_vsync,
   'jank' AS cuj_type
 FROM android_sysui_jank_cujs
-ORDER BY
-  id
 UNION ALL
 SELECT
   cuj_id,
@@ -297,7 +295,7 @@ SELECT
   NULL AS begin_vsync,
   NULL AS end_vsync,
   'latency' AS cuj_type
-FROM android_sysui_latency_cujs ORDER BY id;
+FROM android_sysui_latency_cujs ORDER BY cuj_id;
 
 -- Slices corresponding to critical blocking calls that occurred during a CUJ,
 -- clipped to the CUJ time boundaries.
--- a/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
@@ -276,6 +276,8 @@ SELECT
   end_vsync,
   'jank' AS cuj_type
 FROM android_sysui_jank_cujs
+ORDER BY
+  id
 UNION ALL
 SELECT
   cuj_id,
--- a/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
@@ -215,7 +215,9 @@ WHERE
 
 -- Table tracking all jank/latency CUJs information.
 CREATE PERFETTO TABLE android_jank_latency_cujs(
-  -- Unique incremental ID for each CUJ.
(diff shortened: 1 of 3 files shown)


src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -288,8 +274,28 @@ SELECT
...
+  ts_end,
+  dur,
+  state,
+  -- upid is used as the ui_thread as it's the tid of the main thread.
+  upid AS ui_thread,
+  NULL AS layer_id,
+  NULL AS begin_vsync,
+  NULL AS end_vsync,
+  'latency' AS cuj_type
+FROM android_sysui_latency_cujs;
```
LalitMaganti: nit: you should probably order by id here to make this fast to query.

test/trace_processor/diff_tests/stdlib/android/sysui_cujs_test.py
```diff
@@ -69,3 +88,37 @@ def test_android_sysui_latency_cujs_state(self):
...
+
+  def test_android_cuj_blocking_calls(self):
+    return DiffTestBlueprint(
+        trace=Path(
+            '../../metrics/android/android_blocking_calls_cuj_per_frame_metric.py'
+        ),
+        query="""
+        INCLUDE PERFETTO MODULE android.cujs.sysui_cujs;
+        SELECT *
+        FROM android_cuj_blocking_calls
```
LalitMaganti, line 110: Please don't check in ids into tests, it makes them brittle. Only expose non-id columns.

test/trace_processor/diff_tests/stdlib/android/sysui_cujs_test.py
```diff
@@ -54,6 +54,25 @@ def test_android_sysui_latency_cujs(self):
...
+        trace=Path(
+            '../../metrics/android/android_blocking_calls_cuj_per_frame_metric.py'
+        ),
+        query="""
+        INCLUDE PERFETTO MODULE android.cujs.sysui_cujs;
+        SELECT *
+        FROM android_jank_latency_cujs;
+        """,
+        out=Csv("""
+        "cuj_id","id","upid","process_name","cuj_slice_name","cuj_name","slice_id","ts","ts_end","dur","state","ui_thread","layer_id","begin_vsync","end_vsync","cuj_type"
```
LalitMaganti: Please don't check in ids into tests, it makes them brittle. Only expose non-id columns.

src/trace_processor/perfetto_sql/stdlib/android/cujs/sysui_cujs.sql
```diff
@@ -276,6 +276,8 @@ SELECT
   end_vsync,
   'jank' AS cuj_type
 FROM android_sysui_jank_cujs
+ORDER BY
```
LalitMaganti: This is incorrect, this should happen *after* the union all.

nicomazz (commented):
Comments addressed

LalitMaganti (approved):

PR #7217, opened 2026-08-25
build: add `perfetto_use_system_re2` flag

aduh95:
I'm trying to package Perfetto into NixOS / nixpkgs (https://github.com/NixOS/nixpkgs/pull/553888), and one of the requirement on that system is that downloading things at build time is not allowed, so I'd need a flag to use the system re2.

Given the patch is simple enough, I figured I should send it upstream rather than floating it on Nix end.

### DIFF
--- a/gn/BUILD.gn
+++ b/gn/BUILD.gn
@@ -532,8 +532,27 @@ if (enable_perfetto_pcre2) {
 }
 
 if (enable_perfetto_re2) {
+  # pkg_deps selects the appropriate pkg-config based on current_toolchain and
+  # this is a no-op if |perfetto_use_pkgconfig| is false.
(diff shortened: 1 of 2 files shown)


LalitMaganti:
Looks great thanks! I'd also like @sashwinbalaji to take a look, he's out until later this week so this might take a couple of days before we get back to you!

LalitMaganti (approved):

primiano (approved):
Very nice contrib. Looks okay to me and inline with the protobuf precedent we have!

sashwinbalaji (commented):
Thanks for the change!

sashwinbalaji (approved):

PR #7220, opened 2026-08-25
build: add `perfetto_use_system_zstd` flag

aduh95:
Similar to https://github.com/google/perfetto/pull/7217, I was working on the 57.2 branch, and realized when trying to update to 58.2 that zstd needs the same treatment.

### DIFF
--- a/gn/BUILD.gn
+++ b/gn/BUILD.gn
@@ -461,15 +461,26 @@ if (enable_perfetto_zlib) {
   }
 }
 
-config("system_zstd_config") {
-  libs = [ "zstd" ]
-}
-
 # Zstd is used both by trace_processor and by the tracing service.
 if (enable_perfetto_zstd) {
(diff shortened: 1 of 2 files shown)


LalitMaganti (approved):

gn/BUILD.gn
```diff
@@ -461,10 +461,16 @@ if (enable_perfetto_zlib) {
   }
 }
 
+config("system_zstd_config") {
+  libs = [ "zstd" ]
```
sashwinbalaji: hmm why not use pkg_config for this also like for re2?

sashwinbalaji (approved):

gn/BUILD.gn
aduh95: I failed to consider that pkg-config of zstd is actually called `libzstd` x) pushed a fixup commit

sashwinbalaji (approved):

primiano (approved):

PR #7222, opened 2026-08-25
wattson: Add support to extract SoC model from devicetree on Linux

willmcvicker:
On Android, the `ro.soc.model` system property is used to identify the target's SoC which the wattson stdlib uses to determine if a given target supports Wattson. Since Linux distributions don't use system properties, this patch introduces a new SystemInfo attribute -- linux_device. For arm64 Linux devices, we can populate the linux_device attribute with the Device Tree base compatible. This attribute is then passed through via the metadata to allow the Wattson stdlib to create a mapping between the linux_device and wattson_device properies.

With this change, you can successfully collect Wattson metrics on supported devices running Linux. As an example, use tracebox on your Linux target like this:

  $ tracebox --txt -c wattson.cfg --background-wait -o trace.pb

Bug: 446195032
Test: collect trace on Pixel 6 with Linux and verify Wattson track
Test: verify power metrics wattson_markers_rails,wattson_markers_threads

### DIFF
--- a/include/perfetto/ext/base/android_utils.h
+++ b/include/perfetto/ext/base/android_utils.h
@@ -20,6 +20,7 @@
 #include <cstdint>
 #include <optional>
 #include <string>
+#include <vector>
 
 #include "perfetto/base/build_config.h"
 
@@ -57,6 +58,7 @@ struct SystemInfo {
   std::string android_storage_model;
   std::string android_ram_model;
   std::string android_serial_console;
+  std::vector<std::string> device_tree_compatibles;
 };
 
 // Returns the device's utsname information.
--- a/protos/perfetto/common/system_info.proto
+++ b/protos/perfetto/common/system_info.proto
@@ -25,7 +25,7 @@ message Utsname {
   optional string machine = 4;
 }
 
-// Next id: 18
+// Next id: 19
 message SystemInfo {
   optional Utsname utsname = 1;
   optional string android_build_fingerprint = 2;
@@ -38,6 +38,12 @@ message SystemInfo {
 
   // The SoC model from which trace is collected
   optional string android_soc_model = 9;
+
+  // Device tree compatible strings for Linux devices. This field is populated
+  // from the firmware file /sys/firmware/devicetree/base/compatible on the
+  // target device. The ordering matches the ordering in the device tree.
+  // Introduced after 26Q3 in Sept 2026 and is not supported on older versions.
+  repeated string device_tree_compatibles = 18;
 
   // The guest SoC model from which trace is collected in case of VMs
   optional string android_guest_soc_model = 13;
--- a/protos/perfetto/trace/perfetto_trace.proto
+++ b/protos/perfetto/trace/perfetto_trace.proto
@@ -6103,7 +6103,7 @@ message Utsname {
   optional string machine = 4;
 }
 
-// Next id: 18
+// Next id: 19
 message SystemInfo {
   optional Utsname utsname = 1;
   optional string android_build_fingerprint = 2;
@@ -6116,6 +6116,12 @@ message SystemInfo {
 
   // The SoC model from which trace is collected
   optional string android_soc_model = 9;
+
+  // Device tree compatible strings for Linux devices. This field is populated
+  // from the firmware file /sys/firmware/devicetree/base/compatible on the
+  // target device. The ordering matches the ordering in the device tree.
+  // Introduced after 26Q3 in Sept 2026 and is not supported on older versions.
+  repeated string device_tree_compatibles = 18;
 
   // The guest SoC model from which trace is collected in case of VMs
   optional string android_guest_soc_model = 13;
--- a/src/base/android_utils.cc
+++ b/src/base/android_utils.cc
@@ -15,6 +15,8 @@
  */
 
 #include "perfetto/ext/base/android_utils.h"
+#include "perfetto/ext/base/file_utils.h"
+#include "perfetto/ext/base/string_splitter.h"
 
 #include "perfetto/base/build_config.h"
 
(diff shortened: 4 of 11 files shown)


src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
```diff
@@ -104,6 +104,57 @@ WITH
...
+          "google,ZUMA PRO KOMODO", -- Downstream compatible
+          "Tensor G4"
+        ),
+        (
+          "google,ZUMA PRO TOKAY", -- Downstream compatible
+          "Tensor G4"
+        ),
+        (
+          "google,ZUMA PRO TEGU", -- Downstream compatible
+          "Tensor G4"
```
wusamuel6, line 140: Do we need `comet` and `stallion` for Zuma Pro?

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
```diff
@@ -104,6 +104,57 @@ WITH
...
+          "google,lga-frankel", -- Downstream compatible
+          "Tensor G5"
+        ),
+        (
+          "google,lga-blazer", -- Downstream compatible
+          "Tensor G5"
+        ),
+        (
+          "google,lga-mustang", -- Downstream compatible
+          "Tensor G5"
```
wusamuel6, line 160: Do we need `rango` for lga?

wusamuel6 (approved):

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
willmcvicker: Yeah I can add those.

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
willmcvicker: Yeah I can add rango as well

protos/perfetto/trace/perfetto_trace.proto
```diff
@@ -6102,6 +6102,9 @@ message SystemInfo {
   // The SoC model from which trace is collected
   optional string android_soc_model = 9;
 
+  // The Linux device base compatible string
+  optional string linux_device = 18;
```
sashwinbalaji, line 6106: When I looked at the proto I though this is just one single device which is not the case, it's a null-separated list of strings that identify the exact hardware platform or board mode.

I feel like it's better to keep all (But since you have more info about it, pls correct me if this is not needed) insted of just silently keeping the first entry by doing.

```
info.linux_device = device_compatible.c_str();
```

you can use StringSplitter if needed as it handles null

so should this be repeated string device_tree_compatibles (I think linux_device is just too broad and doesn't depict this field clearly) and change the name and processing everywhere accordinghly. Including:
- updating the proto repeated string device_tree_compatibles 
- in tp making it a kMulti

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
```diff
@@ -121,6 +184,7 @@ SELECT * FROM data;
 
 CREATE PERFETTO TABLE _wattson_device AS
 WITH
+  linux_device AS (SELECT str_value FROM metadata WHERE name = 'linux_device'),
```
sashwinbalaji, line 187: since this is machine scoped, I think you need to select machine id and use it in join also

sashwinbalaji (commented):
Really sorry for the delay, this just got missed.

protos/perfetto/trace/perfetto_trace.proto
willmcvicker: After talking with the team, we agree it'll be better to take all the compatible strings. I'll update this accordingly.

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
willmcvicker: CMIIW, but I don't think we need to use machine_id when grabbing the compatible from the metadata. We are just passing in data from the target.

include/perfetto/ext/base/android_utils.h
```diff
@@ -57,6 +58,7 @@ struct SystemInfo {
   std::string android_storage_model;
   std::string android_ram_model;
   std::string android_serial_console;
+  std::optional<std::vector<std::string>> device_tree_compatibles;
```
sashwinbalaji: don't think this needs to be an optional, empty vectory by itself should cover it and also removes some checks in code below?

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
```diff
@@ -131,6 +176,14 @@ SELECT * FROM data;
 
 CREATE PERFETTO TABLE _wattson_device AS
 WITH
+  dt_compatibles AS (
+    SELECT id, str_value AS compatible
+    FROM metadata
+    WHERE
+      name = 'device_tree_compatible'
+    ORDER BY
+      id DESC
```
sashwinbalaji, line 194: curious why in reverse order

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
```diff
@@ -131,6 +176,14 @@ SELECT * FROM data;
 
 CREATE PERFETTO TABLE _wattson_device AS
 WITH
+  dt_compatibles AS (
```
sashwinbalaji, line 187: I think machine_id is needed due to android having vm enabled + also for merging traces and stuff

maybe I don't understand the usecase completely, @LalitMaganti pls correct me if it ain't required.

protos/perfetto/common/system_info.proto
```diff
@@ -39,6 +39,9 @@ message SystemInfo {
   // The SoC model from which trace is collected
   optional string android_soc_model = 9;
 
+  // Device tree compatible strings for Linux devices
```
sashwinbalaji: pls also mention the sysfs this depends on, the ordering matching the output and also add introduced in perfetto ... part

src/base/android_utils.cc
```diff
@@ -119,7 +122,26 @@ SystemInfo GetSystemInfo() {
     info.system_ram_bytes = static_cast<uint64_t>(pages) * (*info.page_size);
   }
 #endif
+
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_LINUX_BUT_NOT_QNX)
+  // Extract the device compatible strings from the device tree (if available)
+  const std::string path = "/sys/firmware/devicetree/base/compatible";
+  std::string device_compatible;
+  bool read_device_compatible = ReadFile(path, &device_compatible);
```
sashwinbalaji: nits/optional(feel free to ignore): 
- call the string device_tree_compatible_data (to indicate it's multiple)
- directly do if(ReadFile..) removing the bool read_device_compatible (as not adding much)

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
willmcvicker: I dug into this a bit and am fine adding a check for the machine_id. Wattson currently is only targeting the host machine at the moment. So we can filter by (machine_id == 0 OR machine_id IS NULL)

protos/perfetto/common/system_info.proto
willmcvicker: sure, I can add that.

src/trace_processor/perfetto_sql/stdlib/wattson/device_infos.sql
willmcvicker: The upstream device tree compatible strings are ordered from most-specific to least specific. For Wattson, our models are generic across SoCs which would be the least generic compatible. So using the reverse order gets us a match quicker as we will ignore the more specific compatible strings.

include/perfetto/ext/base/android_utils.h
willmcvicker: thanks

src/base/android_utils.cc
willmcvicker: thanks

sashwinbalaji (approved):

PR #7223, opened 2026-08-26
perfetto: reserve TracePacket field IDs 136 and 137

BogdanLionte:
Reserve IDs 136 and 137 in TracePacket as they were accidentally used in downstream Google internal protos.


LalitMaganti (approved):

primiano:
I think at this point we can abandon this. I suggested doing this as I thought they were genuinely used at some point in android with the 136-7 number. but turns out that never happened. So these extension numbers are free to use as usual.

Please shout if I'm missing something. otherwise i'm going to just close and abandon this cl

PR #7224, opened 2026-08-26
tp: fix the strace importer on real -f and -T output

Socialpranker:
I ran a plain `strace -ttt -T -f sh -c "ls /usr/bin > /dev/null; sleep 0.05; echo done"` (strace 6.13) and fed the 228-line result to trace_processor. It produced zero slices and 152 parse failures. Four separate things are wrong, all in the tokenizer I contributed in #6588.

**`[pid N]` prefixes.** Writing to stderr, which is what a piped or redirected capture records, strace prints the pid as `[pid    75] ` rather than the bare `75 ` form it uses with `-o`. The tokenizer only knew the bare form, so every line belonging to a followed process failed to parse. Both forms are accepted now.

**strace's own diagnostics.** strace interleaves messages like `strace: Process 75 attached` with the trace. The `:` in `strace:` made the timestamp parser classify them as `-t`/`-tt` output, which raises `strace_unsupported_timestamp_format` telling the user to re-run with `-ttt` on a trace that already used `-ttt`. Those lines are recognised and skipped now, and format sniffing looks past them, so a log that opens with one (as `strace -f -p <pid>` does) is still detected as strace instead of failing type detection outright.

**Return value split.** The split ran forwards to the first `)` followed by `=`, which lands in the middle of the arguments whenever an argument contains one. wait4 prints its status through macros, so `<... wait4 resumed>[{WIFEXITED(s) && WEXITSTATUS(s) == 0}], 0, NULL) = 75` gave a return value of `= 0}], 0, NULL) = 75`. The scan runs backwards from the end of the line now, which also covers a string argument containing `) =` verbatim.

**`-T` durations.** `-T` prints how long each call took (`= 0 <1.004879>`) and that was ignored, so every completed call became a zero-duration slice with the timing glued onto the return value. The suffix becomes the slice duration now. Without `-T` slices stay zero-duration, exactly as before.

The same log imports as 139 slices across 3 threads, 135 of them with durations, after the change.

The new diff-test fixture is lines from that real strace 6.13 output rather than something hand-written, with the root process's lines given the same `[pid]` prefix strace uses when it is attached to several processes.

Ran `perfetto_trace_processor_unittests --gtest_filter="*Strace*"` (37 pass, 14 of them new), the full unit suite (2067 pass), `diff_test_trace_processor.py --name-filter ".*Strace.*"` (11 pass, 3 new) and the same for `(Systrace|Atrace|PerfText|Ftrace)` to confirm the sniffing change doesn't disturb the neighbouring text formats (28 pass). The repo's pre-push presubmit passes.

Written with Claude Code (Claude Opus 5); I reviewed and tested everything before submitting.

### DIFF
--- a/src/trace_processor/plugins/strace/strace_trace_tokenizer.cc
+++ b/src/trace_processor/plugins/strace/strace_trace_tokenizer.cc
@@ -230,8 +230,7 @@ ParseStraceLineResult ParseStraceLine(std::string_view line) {
     size_t sp = rest.find(' ');
     if (sp != std::string_view::npos) {
       std::string_view head = rest.substr(0, sp);
-      bool all_digits =
-          !head.empty() && head.size() < 10 && IsAllDigits(head);
+      bool all_digits = !head.empty() && head.size() < 10 && IsAllDigits(head);
       if (all_digits) {
         auto pid = base::StringViewToNumber<uint32_t>(base::StringView(head));
         if (!pid)
--- a/src/trace_processor/plugins/strace/strace_trace_tokenizer.cc
+++ b/src/trace_processor/plugins/strace/strace_trace_tokenizer.cc
@@ -55,6 +55,17 @@ std::string_view ToStringView(const TraceBlobView& tbv) {
   return {reinterpret_cast<const char*>(tbv.data()), tbv.size()};
 }
 
+// Whether `s` consists only of ASCII digits. Vacuously true for an empty
+// `s`, matching a `for` loop over it that never finds a non-digit; callers
+// that require at least one digit check emptiness separately.
+bool IsAllDigits(std::string_view s) {
+  for (char c : s) {
+    if (!isdigit(static_cast<unsigned char>(c)))
+      return false;
+  }
+  return true;
+}
+
 // The largest number of whole seconds since the epoch for which
 // `seconds * kNsPerSec` cannot overflow int64_t. Anything larger is
 // rejected outright rather than silently saturating/overflowing; genuine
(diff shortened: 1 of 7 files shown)


LalitMaganti (approved):

src/trace_processor/plugins/strace/strace_trace_tokenizer.cc
```diff
@@ -102,6 +106,69 @@ std::optional<int64_t> ParseEpochTimestamp(std::string_view s) {
...
+  if (whole.empty() || whole.size() > 10 || frac.size() > 9)
+    return std::nullopt;
+  for (char c : whole) {
+    if (!isdigit(static_cast<unsigned char>(c)))
+      return std::nullopt;
+  }
+  for (char c : frac) {
+    if (!isdigit(static_cast<unsigned char>(c)))
+      return std::nullopt;
+  }
```
sashwinbalaji: optional: you can pull this for loop check into a anon namespace helpere function as there are now 3 such copies

sashwinbalaji (approved):

Socialpranker:
Sorry for the churn — I force-pushed a **comment-only** fix on top of the
approved commit, which reset the two approvals. Nothing executable changed.

The parser comment named a flag that does not exist:

```diff
-// fractional digits by default and up to nine with
-// `--syscall-times=precision:ns`, so the fraction is scaled by its digit
-// count rather than assumed to be microseconds.
+// fractional digits by default and up to nine with `--syscall-times=ns`, so
+// the fraction is scaled by its digit count rather than assumed to be
+// microseconds.
```

`strace --syscall-times=precision:ns` errors out with
`invalid --syscall-times argument`; the option takes the precision directly
(`s`/`ms`/`us`/`ns`, added in strace 5.6). The same wrong spelling was
repeated in the unit test comment. Verified against strace 6.13, and the
nine-digit path itself now checked end to end rather than only in the unit
test: `--syscall-times=ns` produces `<0.000159167>`, which imports as
`dur = 159167`.

`git diff 06d3332f..4073f29e` is those two comment blocks and nothing else.
Re-ran locally on the new SHA: 37 `*Strace*` unit tests, 2067 in the full
trace_processor unit suite, 11 `.*Strace.*` diff tests, and the pre-push
presubmit — all pass.

src/trace_processor/plugins/strace/strace_trace_tokenizer.cc
LalitMaganti: +1 I would ask for this too.

Socialpranker: Done in 2a06d8713b — pulled it into an `IsAllDigits()` helper in the anon namespace, three call sites now use it.

LalitMaganti (approved):

sashwinbalaji (approved):

LalitMaganti:
The CI appears to still be failing.

Socialpranker:
That was clang-format on the new helper call site — fixed in fe2f5187f1, CI is green now (only enforce-reviews left, which needs a re-approve after the push).

PR #7225, opened 2026-08-26
tp: avoid using heap_graph_object when heap_graph would do the job

LalitMaganti:
Since we introduced the heap_graph table as the denormalized table to
keep track of information about heap graphs, we should just use it when
looking for information about the number/ts/upid etc of the heap dump.


zezeozue (approved):

stevegolton (approved):
Nice

PR #7226, opened 2026-08-26
docs: document the strace textual format

Socialpranker:
**Depends on #7224 — please land that first.** This describes the strace
importer's behaviour *after* that fix; on current `main` a `-T` capture still
imports with zero-duration slices and the measured numbers below differ. Happy
to hold this until #7224 is in.

---

Perfetto has had an strace importer since the plugin landed, but the format is
not mentioned anywhere under `docs/`. A reader has no way to find out that an
strace log can be opened at all, and — more painfully — no way to find out
which strace flags produce a log that imports cleanly. Get either of `-ttt` or
`-f` wrong and the trace loads to an empty timeline, with the explanation
buried in the `stats` table.

This adds a section to the external-formats guide covering:

- what the importer maps into — thread-track slices, the `strace` category,
  `args`/`ret`, `-T` durations, and blocked calls spanning their
  `<unfinished ...>`/`<... resumed>` pair;
- the two flags that are not optional (`-ttt`, `-f`) and why each is required,
  each pointing at the stat that fires when it is missing;
- the limitations worth stating up front — opaque argument strings, and the
  fact that a `ptrace`-based tool perturbs the timings it reports;
- how to capture a log in the first place.

Two things in it are measured rather than asserted:

- The example block is verbatim lines from an strace 6.13 capture, not
  hand-written — same method as the diff-test fixture in #7224.
- The "use `-o FILE`, don't redirect stderr" advice comes from tracing
  `sh -c 'ls /usr/bin >/dev/null; sleep 0.1'` both ways: with `-o`, all 214
  syscalls import and none are dropped; over stderr, 140 import and 76 are
  dropped as `strace_missing_pid`, because strace leaves the process it
  started unprefixed until a second one is attached. For a program that never
  forks, the stderr capture imports nothing at all.

`--syscall-times` is documented with the version it appeared in (strace 5.6,
per the project's `NEWS`) rather than left for the reader to discover by
trying it.

Verified: `tools/run_presubmit` passes on this branch.

### DIFF
--- a/docs/getting-started/other-formats.md
+++ b/docs/getting-started/other-formats.md
@@ -696,6 +696,145 @@ source directly.**
 - **Tracefs Documentation:**
   [The Tracefs Pseudo Filesystem (kernel.org)](https://www.kernel.org/doc/html/latest/trace/tracefs.html)
 
+## {#strace-format} Linux `strace` textual format
+
+**Description:** [`strace`](https://strace.io/) is the standard Linux tool for
+observing the system calls a process makes, which it does by attaching to the
+process with `ptrace`. Perfetto ingests the textual log strace writes, in the
+form produced by `strace -ttt -f`: one line per system call, carrying the pid
+of the calling thread, a Unix epoch timestamp, the system call name, its
+arguments, the return value and — when `-T` is used — the time spent inside
+the call.
+
+```
+66    1787745825.395990 execve("/usr/bin/sh", ["sh", "-c", "ls /usr/bin > /dev/null; sleep 0"...], 0xffffe3fb3ce0 /* 4 vars */) = 0 <0.003849>
+66    1787745825.527876 clone(child_stack=0xffffede25a00, flags=CLONE_VM|CLONE_VFORK|SIGCHLD <unfinished ...>
+67    1787745825.534500 execve("/usr/bin/ls", ["ls", "/usr/bin"], 0xaaaaf807e498 /* 4 vars */ <unfinished ...>
+66    1787745825.543303 <... clone resumed>) = 67 <0.014914>
+67    1787745825.550192 <... execve resumed>) = 0 <0.014576>
+66    1787745825.554319 wait4(-1,  <unfinished ...>
+67    1787745826.332621 exit_group(0)   = ?
+66    1787745826.348956 <... wait4 resumed>[{WIFEXITED(s) && WEXITSTATUS(s) == 0}], 0, NULL) = 67 <0.794037>
+```
+
+When a call blocks, strace splits it across two lines — the `<unfinished ...>`
+above and the matching `<... wait4 resumed>` — so that the calls made by other
+threads in between stay in chronological order.
+
+**Common Scenarios:** This format is useful when:
+
+- You want to see *where* a process's wall-clock time goes inside the kernel:
+  which `read` blocked and for how long, how long a `futex` wait lasted, how
+  many `openat` calls a startup path makes — as a timeline rather than as
+  thousands of lines of text. The same question can be asked in SQL, for
+  example the calls a run spent the longest inside:
+
+  ```sql
+  select name, count(*) as calls, sum(dur) as total_dur
+  from slice
+  where category = 'strace'
+  group by name
+  order by total_dur desc
+  limit 10;
+  ```
+
+- Perfetto's own tracing is not available on the machine you are debugging — a
+  locked-down container, a customer's box, a distro without `tracefs` — but
+  `strace` is.
+- Someone has attached an strace log to a bug report and you would like to
+  analyse it rather than read it.
+- You want to follow system call behaviour across a whole process tree
+  captured with `strace -f`.
+
+**Perfetto Support:** strace logs are recognised by their content, so no
+particular file extension is required.
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/other-formats.md
```diff
@@ -696,6 +696,136 @@ source directly.**
...
+    than a low-overhead measurement.
+  - Only what strace prints is available: no scheduling, CPU frequency or
+    userspace instrumentation data. Where Perfetto's own tracing is an option,
+    prefer it.
+
+Lines that could not be imported are counted in the `stats` table, each with
+an explanation of what to change:
+
+```sql
+select name, value from stats where name like 'strace%' and value > 0
```
sashwinbalaji: use glob and nit add a semicolon

docs/getting-started/other-formats.md
```diff
@@ -696,6 +696,136 @@ source directly.**
...
+When a call blocks, strace splits it across two lines — the `<unfinished ...>`
+above and the matching `<... wait4 resumed>` — so that the calls made by other
+threads in between stay in chronological order.
+
+**Common Scenarios:** This format is useful when:
+
+- You want to see *where* a process's wall-clock time goes inside the kernel:
+  which `read` blocked and for how long, how long a `futex` wait lasted, how
+  many `openat` calls a startup path makes — as a timeline rather than as
+  thousands of lines of text.
```
sashwinbalaji: would be nice to have a e.g. query for this.

docs/getting-started/other-formats.md
```diff
@@ -696,6 +696,136 @@ source directly.**
...
+  - **`-ttt`.** Timestamps must be Unix epoch seconds. `-t` and `-tt` print a
+    wall-clock time of day with no date, which cannot be turned back into an
+    absolute point in time, so those lines are rejected rather than silently
+    placed decades away from every other trace they are merged with. See the
+    `strace_unsupported_timestamp_format` stat.
+  - **`-f`.** strace only prints pids while it is following processes. Without
+    a pid there is no thread to attribute a system call to, so such lines are
+    dropped; see the `strace_missing_pid` stat.
+- **Limitations:**
+  - Syscall arguments are kept as the single opaque string strace printed.
```
sashwinbalaji: nit: pls use either system call or syscall

docs/getting-started/other-formats.md
```diff
@@ -696,6 +696,136 @@ source directly.**
...
+  sudo strace -ttt -f -T -o my_trace.strace -p 1234
+  ```
+
+- **Use `-o FILE`; do not redirect stderr.** Given `-o`, strace prefixes every
+  line with the pid it belongs to. Writing to stderr instead, it leaves the
+  process it started unprefixed until a second process is attached, and those
+  unprefixed lines have to be dropped on import. Tracing
+  `sh -c 'ls /usr/bin >/dev/null; sleep 0.1'` both ways, `-o` imports all 214
+  system calls and drops none, while the stderr capture imports 140 and drops
+  76. Tracing a program that never forks at all, the stderr capture imports
```
sashwinbalaji: pls remove these numbers 214 and 76 as those vary system to system.

docs/getting-started/other-formats.md
Socialpranker: Done: `where name glob 'strace*' and value > 0;`

docs/getting-started/other-formats.md
Socialpranker: Added one right under that bullet — top system calls by time spent inside them:

```sql
select name, count(*) as calls, sum(dur) as total_dur
from slice
where category = 'strace'
group by name
order by total_dur desc
limit 10;
```

Checked it runs against a real strace capture.

docs/getting-started/other-formats.md
Socialpranker: Done — "system call" everywhere in prose now; `--syscall-times` stays as-is since that's the flag's actual name.

docs/getting-started/other-formats.md
Socialpranker: Removed. Replaced with the part that doesn't depend on the machine: on a program that never forks, no second process is ever attached, so a stderr capture has no pid prefix on any line and imports nothing.

docs/getting-started/other-formats.md
```diff
@@ -696,6 +696,145 @@ source directly.**
...
+  ```
+
+- **Use `-o FILE`; do not redirect stderr.** Given `-o`, strace prefixes every
+  line with the pid it belongs to. Writing to stderr instead, it leaves the
+  process it started unprefixed until a second process is attached, and those
+  unprefixed lines have to be dropped on import. On a program that never forks
+  at all no second process is ever attached, so a stderr capture is dropped in
+  its entirety and imports nothing.
+
+- **For nanosecond-resolution durations,** ask `-T` for more precision:
```
LalitMaganti: This line is confusing and does not match below.

Socialpranker: good catch, that bullet says -T but the command under it doesn't use -T at all. reworded so it says --syscall-times=ns replaces it

LalitMaganti (approved):

sashwinbalaji (approved):

PR #7240, opened 2026-08-26
ui: offer ART heap dumps as pprof from the flamegraph export menu

LalitMaganti:
traceconv can convert a Java heap graph to pprof, but the UI only ever
asked it for allocator profiles, so ART heap dumps had no pprof export
at all. Map each profile type to the flag traceconv expects rather than
hardcoding --alloc.

The download also sat in a button of its own beside the snapshot time,
leaving two unrelated download affordances in one panel. Fold it into
the flamegraph's Export menu, which grows a way for a host to contribute
a download the widget cannot build from the data it is showing.

That export comes from traceconv reading the whole trace, so nothing the
tree explorer does to the displayed tree reaches it. Say so under the
menu entry. Drop the entry for area selections, whose time range does
not name a snapshot traceconv can convert.


stevegolton (approved):
Should we also add the download button to the flamegraph in HDE?

PR #7242, opened 2026-08-26
ui: GpuCompute: correct the Summary tab and rename Workload Analysis

zytyz:
Small fixes so a vendor plugin can register its own metrics into the
GpuCompute tab.

- Fix the Summary tab so a dependent plugin's metric ids reach it. The
  prefetch ran before those plugins had registered, so the columns came out
  empty and stayed empty.

- Scale the compute and memory throughput bars against the hardware peak
  rather than the largest value in the table, and label them with a percent
  sign. On a capture spanning 0.03% to 0.93% of peak, a kernel at the mean
  0.47% filled 51% of its bar, and the busiest always pinned at 100%.

- Rename "Workload Analysis" to "Compute Workload Analysis". It covers
  per-pipeline compute utilization only, and the plain name collides with the
  memory counterpart a vendor plugin registers alongside it.

- Rename the matching workflow in the Perfetto agent skill, so an agent and a
  user read one name for one section.

### DIFF
--- a/ai/skills/perfetto/workflows/gpu/compute/compute_workload_analysis.md
+++ b/ai/skills/perfetto/workflows/gpu/compute/compute_workload_analysis.md
@@ -1,4 +1,4 @@
-# Compute kernel — workload analysis (which pipeline is the limit?)
+# Compute kernel — Compute Workload Analysis (which pipeline is the limit?)
 
 For a compute-bound kernel, "compute-bound" isn't the end of the story — *which*
 execution pipeline is saturated determines the lever. This workflow looks at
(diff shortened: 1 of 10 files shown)


zytyz:
@dreveman ptal. Thanks!

dreveman (approved):

zytyz:
Hi @stevegolton, Just checking in on this. Let me know if you have any feedback when you get a chance to review. Thanks!

stevegolton (approved):
Apologies for the delay - LGTM.

LalitMaganti (approved):

PR #7244, opened 2026-08-26
tp: remove recursive Perfetto SQL functions

LalitMaganti:
They were never used in practice and/or achieved popularity and slow
down non-recursive functions. There's also a subtle UAF (see b/552636369).

Fix it by just removing support for them altogether.

### DIFF
--- a/src/trace_processor/perfetto_sql/engine/created_function.cc
+++ b/src/trace_processor/perfetto_sql/engine/created_function.cc
@@ -20,15 +20,11 @@
 #include <cstdint>
 #include <memory>
 #include <optional>
-#include <queue>
-#include <stack>
 #include <string>
 #include <utility>
-#include <vector>
 
 #include "perfetto/base/logging.h"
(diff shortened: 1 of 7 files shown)


src/trace_processor/perfetto_sql/engine/created_function.cc
```diff
@@ -674,29 +289,21 @@ void CreatedFunction::Step(sqlite3_context* ctx,
...
+          state->connection()->sqlite_connection()->db(), state->prototype());
+      !status.ok()) {
+    sqlite::utils::SetError(ctx, status.c_message());
   }
 }
 
 base::Status CreatedFunction::Prepare(CreatedFunction::UserData* ctx,
                                       FunctionPrototype prototype,
-                                      sql_argument::Type return_type,
+                                      sql_argument::Type,
```
safayat-google: since it's not used, remove completely from .h as well?

safayat-google (approved):

PR #7257, opened 2026-08-27
tools: remove all obsolete stacked PR tools

LalitMaganti:
Now GitHub tools exist, no reason for these to exist anymore.


docs/AGENTS.md
```diff
@@ -192,15 +192,6 @@ For example, to run `test_my_cool_test`, use the filter `MyTestSuite.my_cool_tes
   
 - **Remove `test_` prefix for diff tests.** When using the `--name-filter` flag for diff tests, do not include `test_` in the filter. The test runner automatically drops this prefix. For example, to run `test_my_cool_test`, use the filter `MyTestSuite.my_cool_test`.
 
-## Getting Diffs
```
stevegolton, line 195: This change seems unrelated?

stevegolton (approved):

docs/AGENTS.md
LalitMaganti: It's related because it only exists because of those tools. Without it, we would need a different approach *but* the models are actually smart enough to understand the correct thing to do so we just don't need anything at all.

PR #7264, opened 2026-08-28
protozero: let transports read directly into ProtoRingBuffer

LalitMaganti:
Every transport kept a staging buffer, read into it, then handed those
bytes to Append(), which copied them again. BeginWrite()/EndWrite() lets
read(2) deposit them in the ring buffer directly, so they are copied
once.

The reservation is handed out as a move-only WriteHandle that must be
consumed exactly once, by EndWrite() to keep what was written or
AbortWrite() to discard it; dropping one fails a CHECK. Holding it in a
value the compiler tracks is what makes the pattern safe to use: the
buffer refuses to recompact, grow or hand out a second reservation while
one is outstanding, so bytes cannot move under a pointer someone still
holds. Rpc::RequestHandle carries the same contract out to the
transports.

```
  out/linux_clang_release/perfetto_benchmarks \
      --benchmark_filter='ProtoRingBufferIngest|ProtoRingBufferDispatch'
  ingest, 4KB reads    142 ns -> 98.3 ns
  ingest, 1MB reads   38.4 us -> 19.2 us
  dispatch             171 ns -> 99.1 ns
```

traceconv also drops a 64MB read buffer: peak RSS 62.7MB -> 13.6MB.

### DIFF
--- a/include/perfetto/ext/protozero/proto_ring_buffer.h
+++ b/include/perfetto/ext/protozero/proto_ring_buffer.h
@@ -17,8 +17,10 @@
 #ifndef INCLUDE_PERFETTO_EXT_PROTOZERO_PROTO_RING_BUFFER_H_
 #define INCLUDE_PERFETTO_EXT_PROTOZERO_PROTO_RING_BUFFER_H_
 
+#include <stddef.h>
 #include <stdint.h>
 
+#include "perfetto/base/compiler.h"
 #include "perfetto/ext/base/paged_memory.h"
 
 namespace protozero {
@@ -115,6 +117,44 @@ class RingBufferMessageReader {
   // Will invaildate the pointers previously handed out.
   void Append(const void* data, size_t len);
 
+  // Returned by BeginWrite() and must be given back to EndWrite() (to keep
+  // what was written) or AbortWrite() (to discard it). It enforces that
+  // writes are neither left pending nor interleaved: only one handle can
+  // exist at a time, and letting it go out of scope without passing it to
+  // EndWrite()/AbortWrite() causes a CHECK. The buffer will not recompact or
+  // grow while a handle is outstanding.
+  class WriteHandle {
+   public:
+    WriteHandle() = default;
+    ~WriteHandle();
+
+    WriteHandle(WriteHandle&&) noexcept;
+    WriteHandle& operator=(WriteHandle&&) noexcept;
+    WriteHandle(const WriteHandle&) = delete;
+    WriteHandle& operator=(const WriteHandle&) = delete;
+
+    uint8_t* data() const { return data_; }
+    size_t size() const { return size_; }
+
+    void EndWrite(size_t size_written);
+    void AbortWrite();
+
+    explicit operator bool() const { return reader_ != nullptr; }
+
+   private:
+    friend class RingBufferMessageReader;
+    WriteHandle(RingBufferMessageReader* reader, uint8_t* data, size_t size)
+        : reader_(reader), data_(data), size_(size) {}
+
+    RingBufferMessageReader* reader_ = nullptr;
+    uint8_t* data_ = nullptr;
+    size_t size_ = 0;
+  };
+
+  // Zero-copy counterpart of Append(): the caller reads straight into the
+  // returned reservation (e.g. by passing data() to read(2)).
+  PERFETTO_WARN_UNUSED_RESULT WriteHandle BeginWrite(size_t size);
+
   // If a message can be read, it returns the boundaries of the message
   // (without including the preamble) and advances the read cursor.
   // If no message is available, returns a null range.
@@ -131,11 +171,16 @@ class RingBufferMessageReader {
   virtual Message TryReadMessage(const uint8_t* start, const uint8_t* end) = 0;
 
  private:
+  friend class WriteHandle;
+
+  void FinishWrite(size_t size_written);
+
   perfetto::base::PagedMemory buf_;
   Message fastpath_{};
   bool failed_ = false;  // Set in case of an unrecoverable framing faiulre.
   size_t rd_ = 0;        // Offset of the read cursor in |buf_|.
   size_t wr_ = 0;        // Offset of the write cursor in |buf_|.
+  bool write_in_flight_ = false;
 };
 
 class ProtoRingBuffer final : public RingBufferMessageReader {
--- a/src/protozero/proto_ring_buffer.cc
+++ b/src/protozero/proto_ring_buffer.cc
@@ -16,6 +16,10 @@
 
 #include "perfetto/ext/protozero/proto_ring_buffer.h"
 
+#include <new>
+#include <utility>
+
+#include "perfetto/base/compiler.h"
 #include "perfetto/base/logging.h"
 #include "perfetto/ext/base/paged_memory.h"
 #include "perfetto/protozero/proto_utils.h"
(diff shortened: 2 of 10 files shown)


primiano:
I'm going to take these reviews if you don't mind, as I wrote these classes

src/trace_processor/rpc/unixd.cc
```diff
@@ -322,12 +320,6 @@ void UnixRpcServer::OnDataAvailable(base::UnixSocket* self) {
 void UnixRpcServer::DispatchMessage(base::UnixSocket* sock,
                                     const uint8_t* data,
                                     size_t len) {
-  namespace pu = protozero::proto_utils;
-  uint8_t preamble[16];
-  uint8_t* preamble_end = preamble;
-  preamble_end = pu::WriteVarInt(pu::MakeTagLengthDelimited(1), preamble_end);
```
primiano, line 328: i'm not sure what happened here? How can teh code works if you stop emitting the proto preamble?
I doi't see it readded anywhere else

primiano (commented):
Very nice. LGTM % one question on what hapepend to unixd.cc, i'm confused on how that can still work.

src/trace_processor/rpc/unixd.cc
LalitMaganti: Your comment actually made me realize that there's a bunch of complexity here where we were framing once per connection and then once again per RPC. I made a half change where I short-circuited the framing in the unixd making the preamble write unncesssary

But there's an even better thign we can do which is kill the framing buffers entirely. So I've done that now.

LalitMaganti:
> Very nice. LGTM % one question on what hapepend to unixd.cc, i'm confused on how that can still work.

Replied, PTAL

src/protozero/proto_ring_buffer.cc
```diff
@@ -133,21 +122,53 @@ void RingBufferMessageReader::Append(const void* data_void, size_t data_len) {
       while (data_len > new_size - wr_)
         new_size += kGrowBytes;
       if (new_size > kMaxMsgSize * 2) {
+        // These bytes can never amount to a message (e.g. a never-ending
+        // varint). |buf_| exceeds kMaxMsgSize by now, so dropping them leaves
+        // room for the write.
         failed_ = true;
-        return;
+        rd_ = wr_ = 0;
+        return static_cast<uint8_t*>(buf_.Get());
```
primiano: you need to update writable_ here as well no? otherwise you create another cases where a wrongly framed message will crash the whole thing

include/perfetto/ext/protozero/proto_ring_buffer.h
```diff
@@ -115,6 +116,14 @@ class RingBufferMessageReader {
   // Will invaildate the pointers previously handed out.
   void Append(const void* data, size_t len);
 
+  // Zero-copy counterpart of Append(), for transports that can deposit their
+  // bytes straight into the ring buffer rather than into a staging buffer of
+  // their own. BeginWrite() reserves `size` bytes at the write cursor; the
+  // caller writes at most that many there (e.g. by passing the pointer to
+  // read(2)) and calls EndWrite() with the number actually written.
+  uint8_t* BeginWrite(size_t size);
```
primiano: btw one thing I was noticing is that while i like this patter of "please give me a buffer to write" rather than buffering at the destination, it makes code way more brittle and harder to think about lifetime of those allocs.
I think (I haven't tried) it could be made a bit more robust if instead of returning a uint8_t* from Begin Write, you have to return a wrapper object that has move-only semantics (and which has detection if you let it go out of scope)

so the whole game is that BeginWrite returns this move-only handkle, and you have to std::move it back and pass to EndWrite. so you guarantee that both memory isn't not freed under your hoods, and we don't have accidental interleaving. The move only object could also hold the "hoe many bytes have been actually written" here.

src/trace_processor/rpc/unixd.cc
```diff
@@ -291,54 +287,32 @@ UnixRpcServer::ClientConn* UnixRpcServer::FindConn(base::UnixSocket* self) {
...
+    self->Send(resp, resp_len);
   });
-  rpc_.OnRpcRequest(preamble, static_cast<size_t>(preamble_end - preamble));
-  rpc_.OnRpcRequest(data, len);
+
+  // Read straight into Rpc's tokenizer. A session serves one client at a time,
+  // so there is no second byte stream that could interleave with this one.
+  for (;;) {
+    constexpr size_t kReadSize = 4096;
+    size_t n = self->Receive(rpc_.BeginRpcRequest(kReadSize), kReadSize);
```
primiano: as much as I like compact code, imho this becomes more readable if you break into:

uint8_t* dst_buf = rpc_.BeginRpcRequest(kReadSize);
size_t rx_bytes = self->Receive(dst_buf, kReadSize);
rpc_.EndRpcRequest(rx_bytes);

primiano (commented):
much better, LG with some minor questions

LalitMaganti:
Reworked it as you suggested. PTAL.

include/perfetto/ext/protozero/proto_ring_buffer.h
```diff
@@ -115,6 +117,41 @@ class RingBufferMessageReader {
   // Will invaildate the pointers previously handed out.
   void Append(const void* data, size_t len);
 
+  // Must be consumed exactly once, by EndWrite() to keep what was written or
```
primiano: I'd say here:

This handle is returned by BeginWrite and must be returned to EndWrite. It enforces that we don't leave writes pending, or try to interleave writes. Only one handle can exist, and letting it go out of scope causes a CHECK (unless passed to EndWrite).

src/protozero/proto_ring_buffer.cc
```diff
@@ -133,21 +144,69 @@ void RingBufferMessageReader::Append(const void* data_void, size_t data_len) {
...
-  memcpy(&buf[wr_], data, data_len);
-  wr_ += data_len;
+  write_in_flight_ = true;
+  return WriteHandle(this, static_cast<uint8_t*>(buf_.Get()) + wr_, data_len);
+}
+
+void RingBufferMessageReader::WriteHandle::EndWrite(size_t size_written) {
+  PERFETTO_CHECK(reader_ != nullptr);
+  PERFETTO_CHECK(size_written <= size_);
+  RingBufferMessageReader* reader = reader_;
```
primiano: more of a stylistic nit, here and elsewhere, but these lines can be easily replaced, I believe, with

std::exchange(reader_, nullptr)->FinishWrite(size_written)

same below and in the other files

src/trace_processor/rpc/rpc.cc
```diff
@@ -253,6 +253,44 @@ void Rpc::ResetTraceProcessorInternal(const Config& config) {
...
+}
+
+void Rpc::RequestHandle::AbortRequest() {
+  PERFETTO_CHECK(rpc_ != nullptr);
+  rpc_ = nullptr;
+  write_.AbortWrite();
+}
+
+Rpc::RequestHandle Rpc::BeginRpcRequest(size_t size) {
+  return RequestHandle(this, rxbuf_.BeginWrite(size));
```
primiano, line 295: here I would also add a counter to catch the case of somebody interleaving two requests and creating twice BeginRcpRequests.

The handle protects you about losing track of the pointer, but doesn't protect about somebody creating (and holding onto) two distinct handles.

I think you need a boolean that you set here on BeginRpc, and clear in End/AbortWrite, and in both places you CHECK that is false (in Begin) and TRUE (in end)

In the other file you have a bool write_in_flight. I think you need the same here to be more robust

primiano (commented):
LGTM with few minor comments (just land after addressing no need to re-wait)

primiano (approved):
I forgot to cllick Approve on the previous one

LGTM with few minor comments (just land after addressing no need to re-wait)

PR #7265, opened 2026-08-28
ui: write wasm RPC requests directly into the tokenizer

LalitMaganti:
trace_processor_rpc_init() allocated a 32MB request buffer that JS wrote into
and C++ copied out of on every chunk. The tokenizer holds a standing
reservation instead and hands back the address for the next write, so nothing
is copied and it stays at one JS->Wasm call per request.

Entry points are resolved once with cwrap() instead of ccall(), which redoes
the symbol lookup and argument marshalling on every call. For a numeric
signature cwrap() returns the raw wasm export, so this is as direct as it gets:
29.6 ns -> 4.5 ns per call.

  memory64, vs the staging buffer: trace load -61.5%, queries -23.7%

addFunction() was declared as returning void; it returns a function pointer.
ccall()'s any[] hid that, cwrap()'s number[] does not.

### DIFF
--- a/gn/standalone/wasm.gni
+++ b/gn/standalone/wasm.gni
@@ -47,11 +47,12 @@ template("wasm_lib") {
       _environment = "worker"
     }
     if (_enable_fs) {
-      _exports = "['ccall', 'callMain', 'addFunction', 'FS', 'FS_mkdir', " +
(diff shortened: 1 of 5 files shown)


stevegolton (approved):
Woah, nice!

PR #7266, opened 2026-08-28
base: stream HTTP and websocket payloads into the handler's buffer

LalitMaganti:
This makes it such that instead of the HttpServer providing the buffer
for HTTP/Websocket payloads, it instead asks the handler for where they
would like the payload to be stored.

The flow looks like this: socket read -> HTTPServer reads the
headers/framing -> once it knows the size it calls OnHttpRequestBody or
OnWebsocketPayload depending on the message type -> the handler provides
the storage bytes -> httpserver reads into those provided buffers.

The motivation for this is that it removes one copy allowing for
"zero-copy" parsing and tokenization etc.

The motivation for this is multi-threaded trace processor which requires
fast-handoffs of buffers between threads and reducing the latency of the
parse path (because we're going to be adding latency by a cross thread
hop). This change is one step on that journey.

Because a payload is now read across several socket reads, more than one
can be in flight at a time, so every buffer the handler hands out has to
belong to the connection asking for it. RPC payloads go into that
connection's Rpc::Stream, added in the previous change; the bodies of
the non-RPC endpoints go into a per-connection buffer alongside it.
Sharing either across connections lets one connection's payload be
dispatched into another's, or freed under it when the buffer grows.

The origin check moves above the point where the body is requested, so a
request that is about to be refused with a 403 is never handed a payload
sink to write into.

The reservation is held in the connection's state between the two calls,
so a peer that disappears mid-payload gives it back explicitly rather
than leaving the tokenizer believing a write is still in flight.

### DIFF
--- a/src/base/http/http_server.cc
+++ b/src/base/http/http_server.cc
@@ -15,6 +15,8 @@
  */
 #include "perfetto/ext/base/http/http_server.h"
 
+#include <algorithm>
+
 #include <cinttypes>
 
 #include <cstddef>
@@ -41,7 +43,9 @@ namespace perfetto::base {
 
 namespace {
 constexpr size_t kMaxPayloadSize = 64 * 1024 * 1024;
-constexpr size_t kMaxRequestSize = kMaxPayloadSize + 4096;
+// |rxbuf| holds only headers, so this also bounds how much of a payload can
+// arrive in the same read() as them and has to be moved to the handler.
+constexpr size_t kMaxHeadersSize = 64 * 1024;
 
 enum WebsocketOpcode : uint8_t {
   kOpcodeContinuation = 0x0,
@@ -142,22 +146,20 @@ void HttpServer::OnDataAvailable(UnixSocket* sock) {
   PERFETTO_CHECK(conn);
 
   char* rxbuf = reinterpret_cast<char*>(conn->rxbuf.Get());
+  // Reading and parsing interleave: it's the parser that discovers that the
+  // bytes still on the socket are a payload, not headers.
   for (;;) {
-    size_t avail = conn->rxbuf_avail();
-    PERFETTO_CHECK(avail <= kMaxRequestSize);
-    if (avail == 0) {
-      conn->SendResponseAndClose("413 Payload Too Large");
-      return;
+    if (conn->payload_used_ < conn->payload_size_) {
+      size_t rsize = sock->Receive(conn->payload_ + conn->payload_used_,
+                                   conn->payload_size_ - conn->payload_used_);
+      if (rsize == 0)
+        return;
+      conn->payload_used_ += rsize;
+      continue;
     }
-    size_t rsize = sock->Receive(&rxbuf[conn->rxbuf_used], avail);
-    conn->rxbuf_used += rsize;
-    if (rsize == 0 || conn->rxbuf_avail() == 0)
-      break;
-  }
 
-  // At this point |rxbuf| can contain a partial HTTP request, a full one or
-  // more (in case of HTTP Keepalive pipelining).
-  for (;;) {
+    // At this point |rxbuf| can contain a partial HTTP request, a full one or
+    // more (in case of HTTP Keepalive pipelining).
     size_t bytes_consumed;
 
     if (conn->is_websocket()) {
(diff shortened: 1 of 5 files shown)


primiano:
Sry this needs some more commentary i'm a bit lost in trying to follow what the intended new usage of onWebsocket** is, before I even look into the code itself.

At very least the commit message (or the comments in the code) need to explain what you have in mind here.

LalitMaganti:
> Sry this needs some more commentary i'm a bit lost in trying to follow what the intended new usage of onWebsocket** is, before I even look into the code itself.

The idea is that the caller provides the buffer for the httpserver to read the websocket message/http body payload into. This removes the need to first copy into the httpserver staging and *then* copy into another buffer.

Motivation is trace processor multi-threading which needs buffers which can move between threads.

I've changed the commit message, PTAL.

src/trace_processor/rpc/httpd.cc
```diff
@@ -63,12 +63,19 @@ class Httpd : public base::HttpRequestHandler {
...
+                              size_t size) override;
   void OnHttpRequest(const base::HttpRequest&) override;
   void OnWebsocketMessage(const base::WebsocketMessage&) override;
 
   static void ServeHelpPage(const base::HttpRequest&);
 
   Rpc& global_trace_processor_rpc_;
+  // Bodies for the non-RPC endpoints, which just want to read them. Only one
+  // is ever in flight: httpd serves a single TraceProcessor instance.
+  std::unique_ptr<uint8_t[]> payload_buf_;
```
primiano: I don't understand how you can have only one of thi for the whole HTTP server.
You can receive N connections and each of them will overlap on the same payload buf

src/trace_processor/rpc/httpd.cc
```diff
@@ -127,6 +134,24 @@ void Httpd::Run(const std::string& listen_ip,
...
 
+// RPC payloads go straight into the tokenizer, so a trace chunk is never
+// copied.
+uint8_t* Httpd::OnHttpRequestBody(const base::HttpRequest& req, size_t size) {
+  if (req.uri == "/rpc")
+    return global_trace_processor_rpc_.BeginRpcRequest(size);
+  if (size > payload_buf_size_) {
+    // Deliberately not value-initialized: |size| comes from Content-Length, so
+    // touching it here would make an unsent body resident.
+    payload_buf_.reset(new uint8_t[size]);
```
primiano: I don't think this is right. You are sharing the same payload_buf_ for the whole HTTPd, which is one per TP. But you can definitely interleave different OnHttpRequestBody while the connections is still being used no? it's not that every http request resolves within one packet buffer.
So here you will end up resetting an older payload_buf_ while another interleaved requests is still using it.
I'm NOT talking about concurrency (I know it's single threaded).
I'm talkinga bout this scenario:

- Request 1 starts,  OnHttpRequestBody, you allocate payload_buf_
- REquest 2 starts, OnHttpRequestBody, you reallocate a new payload_buf_ and free the older one
- Some data arrives from req 1: now the machinery internally is holding a pointer on memory that you freed in the previous step.


I asked AI to check that I'm not allucinating and seems to agree with me. THis is what it says:

```
I verified two connections really can be mid-payload at once (a test handler counting outstanding buffers reports 2), and then reproduced it end-to-end: two concurrent POST /rpc with 1 MB bodies, interleaved in 64 KB slices.

[PR]   PERFETTO_CHECK(size_written <= writable_)  proto_ring_buffer.cc:144
       #01 RingBufferMessageReader::EndWrite
       #02 Rpc::EndRpcRequest                      rpc.cc:264
       #03 Httpd::OnHttpRequest                    httpd.cc:214
       → process aborts; 14463 log lines of "RPC framing error" /
         "Unknown request type" first, i.e. the two streams corrupted each other
[main] server survives, each connection gets its own body, one clean
       "RPC framing error" per connection as expected

There are three distinct failure modes from the same root cause:

Aliasing (equal sizes, no realloc): both connections get the same pointer and shred each other's bytes — that's the corruption above.
Use-after-free (growing size): BeginWrite reallocates and PagedMemory munmaps the old block, so the first connection's payload_ dangles. I saw this as recv() returning EFAULT and the connection being silently dropped with no HTTP response; if the region gets re-mapped instead, it's a silent OOB write.
Same for payload_buf_: payload_buf_.reset(new uint8_t[size]) frees the buffer another connection is streaming into.
```

src/base/http/http_server.cc
```diff
@@ -251,18 +273,41 @@ size_t HttpServer::ParseOneHttpRequest(HttpServerConnection* conn) {
...
-  if (!all_headers_received || buf_view.size() < body_size)
+  if (!all_headers_received)
     return 0;
 
-  http_req.body = buf_view.substr(0, body_size);
+  // Only what arrived in the same read() as the headers has to be moved; the
+  // rest is read from the socket into |dst|. The headers stay in |rxbuf| and
+  // are re-parsed on completion, rather than kept alive across the read.
+  if (body_size > 0 && conn->payload_ == nullptr) {
+    uint8_t* dst = req_handler_->OnHttpRequestBody(http_req, body_size);
```
primiano, line 299: shouldn't we do the origin check here before passing it down?
it's a bit weird that now we can end up passing down teh request to the handler even if few lines below we might end up with a 403.
(Also not sure if that will end up with embalancing Begin/End writes, didn't look deep enough, but feels like the origin check below should happen before this point)

LalitMaganti:
https://github.com/google/perfetto/pull/7274 should have addressed the stream tearing problem completel leaving this CL focusing on just migrating the HTTP server to be zero copy. PTAL.

primiano (approved):

PR #7267, opened 2026-08-28
protozero: remove Append() and its fastpath

LalitMaganti:
Append()'s fastpath returned messages pointing into the caller's buffer,
valid only until the next call. That is why RemoteTraceProcessor had to
hoist its read buffer out of the loop, and it is what stops a Message
from owning its storage. Every caller now writes through BeginWrite(),
so both can go.

Also collapses RingBufferMessageReader into ProtoRingBuffer: the virtual
had a single production implementation, and a test-only subclass whose
coverage the ProtoRingBuffer tests subsume.

The benchmarks added to motivate the switch go too: their "before" arm
is the staging buffer this removes.


primiano (approved):

primiano (approved):
still LG

PR #7269, opened 2026-08-28
tp: let TraceBlob wrap memory it does not own

LalitMaganti:
TraceBlob can hold heap memory it allocated or an mmap region, so a
caller that already has the bytes somewhere else has no way to hand
them over: it has to copy them into a blob first.

Add an ownership mode that wraps a pointer and holds a shared_ptr to
whoever owns it, for as long as the blob and every TraceBlobView onto
it. Give Rpc::Parse() an overload taking a TraceBlobView so callers can
use it; the pointer-and-length overload stays and now copies through
TraceBlob::CopyFrom() rather than open-coding new[] and memcpy.

No behaviour change: nothing adopts anything yet.

### DIFF
--- a/src/trace_processor/rpc/rpc.cc
+++ b/src/trace_processor/rpc/rpc.cc
@@ -43,6 +43,8 @@
 #include "perfetto/protozero/scattered_heap_buffer.h"
 #include "perfetto/public/compiler.h"
 #include "perfetto/trace_processor/basic_types.h"
+#include "perfetto/trace_processor/trace_blob.h"
+#include "perfetto/trace_processor/trace_blob_view.h"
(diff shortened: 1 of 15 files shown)


src/trace_processor/rpc/rpc.cc
```diff
@@ -683,13 +684,16 @@ base::Status Rpc::Parse(const uint8_t* data, size_t len) {
   bytes_parsed_ += len;
   MaybePrintProgress();
 
-  if (len == 0)
+  if (blob.size() == 0)
```
sashwinbalaji: nit: this can be len(as before)

sashwinbalaji (approved):

PR #7274, opened 2026-08-28
tp: give the RPC layer one framing stream per connection

LalitMaganti:
Rpc holds a single tokenizer for the whole process, even though several
transports serve more than one peer at a time. A tokenizer holds the
message that arrived in pieces, so interleaving the bytes of two peers
into one corrupts both.

unixd.cc had already worked around this by keeping a ProtoRingBuffer per
connection and forwarding only whole messages, and httpd.cc is about to
need the same thing: once it reads payloads straight into the tokenizer,
two connections mid-payload at once alias each other's memory.

Introduce Rpc::Stream, owning the framing state of one byte-pipe: the
tokenizer and the response function. Transports that serve one peer
(Wasm, stdiod) hold one stream; those that serve many (httpd, unixd)
hold one per connection. unixd loses its hand-rolled tokenizer and
message-forwarding loop, as the stream does that now.

The message sequence ids deliberately stay on Rpc, shared by every
stream. They are not framing state: a second peer numbering its requests
from its own 0 breaks the sequence, and that is how two clients on one
TraceProcessor - two UI tabs on the same --httpd instance, say - are
detected and told so. Per-connection tokenizers make that check more
useful rather than replacing it, because the peers now reach it with
their framing intact instead of first shredding each other's messages.

This also retires the SetRpcResponseFunction(fn)/SetRpcResponseFunction
(nullptr) dance every transport had to perform around each dispatch: the
response function is handed to the stream once, at construction.

### DIFF
--- a/src/trace_processor/rpc/rpc.h
+++ b/src/trace_processor/rpc/rpc.h
@@ -81,9 +81,18 @@ class Rpc {
   // exchanged on these pipes are TraceProcessorRpc protos (defined in
   // trace_processor.proto). This has been introduced in Perfetto v15.
 
-  // A reservation inside the tokenizer for a transport to read into. Must be
-  // consumed exactly once: EndRequest() reports how many bytes were written
-  // and dispatches whatever messages completed, AbortRequest() discards it;
+  // The size argument is a uint32_t and not size_t to avoid ABI mismatches
+  // with Wasm, where size_t = uint32_t.
+  // (nullptr, 0) has the semantic of "close the channel" and is issued when an
+  // unrecoverable wire-protocol framing error is detected.
+  using RpcResponseFunction =
+      std::function<void(const void* /*data*/, uint32_t /*len*/)>;
+
+  class Stream;
+
+  // A reservation inside a Stream's tokenizer to read into. Must be consumed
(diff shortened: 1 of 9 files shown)


src/trace_processor/rpc/rpc.h
```diff
@@ -81,56 +81,85 @@ class Rpc {
...
+  //
+  // A Stream owns only the framing state: the tokenizer holding a message that
+  // arrived in pieces. Interleaving the bytes of two peers into one tokenizer
+  // corrupts both, hence one per connection.
+  //
+  // The message sequence ids stay on Rpc and shared: a second peer numbering
+  // from its own 0 breaks the sequence, which is how two clients on one
+  // TraceProcessor (e.g. two UI tabs on the same --httpd instance) are
+  // detected and told so.
+  class Stream {
```
primiano, line 137: not very srrong on this, but over time I regretted any time I created a nested class with more than one level, as it makes eyeballing the code very hard.

I'd probably consider making Stream a sibling of ReqHandle, and using friends, purely to be able to figure out which method belongs to what when reading code

primiano (approved):

PR #7275, opened 2026-08-28
protozero: let tokenized messages outlive the read loop

LalitMaganti:
ProtoRingBuffer hands out messages pointing into a buffer that the next
BeginWrite() is free to recompact or grow, so a message is only valid
until the following read. Passing one to another thread - what running
trace processor's RPC off the frontend thread needs - is therefore not
expressible, and the fallback is copying every request out, including
32 MB APPEND_TRACE_DATA chunks.

Refcount the buffer and hand out messages as slices holding a reference.
The write path is unchanged (a single non-wrapping buffer, rewound
whenever the read cursor catches up) except that recompacting, growing
and rewinding are now only allowed while nothing points into the buffer.
When something does, the write moves onto another buffer and the last
message releases the old one.

Measured over 2000 messages, buffers allocated:

```
  msg size  retained    before   after
      512B         0         1       1
      512B         1         8       2
      512B        64         8       2
       64K         1      2000       2
       64K        64      2000    2000
```

out/linux_clang_release/perfetto_benchmarks \
    --benchmark_filter='BM_ProtoRingBuffer.*'

### DIFF
--- a/include/perfetto/ext/protozero/proto_ring_buffer.h
+++ b/include/perfetto/ext/protozero/proto_ring_buffer.h
@@ -20,8 +20,9 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include <memory>
+
 #include "perfetto/base/compiler.h"
-#include "perfetto/ext/base/paged_memory.h"
 
 namespace protozero {
 
@@ -91,22 +92,56 @@ namespace protozero {
 //
 // In the next ReadMessage() the R cursor will chase the W cursor. When this
 // happens (very frequent) we can just reset both cursors to 0 and restart.
-// If we are unlucky and get to the end of the buffer, two things happen:
-// 1. We try first to recompact the buffer, moving everything left by R.
-// 2. If still there isn't enough space, we expand the buffer.
+// If we reach the end of the buffer, we calculate the space required for the
+// unread bytes and the next write. If compaction provides enough space, we move
+// unread bytes left by R; otherwise we copy them directly into a larger buffer.
 // Given that each message is expected to be at most kMaxMsgSize (64 MB), the
 // expansion is bound at 2 * kMaxMsgSize.
+//
+// Rewinding and recompacting are only possible while no Message points into
+// the buffer. Messages own references to the underlying allocation. If it is
+// still shared when a write needs more room, unread bytes move to another
+// buffer. One previous buffer is kept for reuse, provided it is large enough.
+// This avoids repeated allocations when alternating between two buffers;
+// deeper queues of retained messages can still cause allocation churn.
 
 class ProtoRingBuffer {
+ private:
+  class Buffer;
+  struct BufferDeleter {
+    void operator()(Buffer*) const;
+  };
+  // Each handle owns one reference to the allocation, released by its deleter.
+  using BufferPtr = std::unique_ptr<Buffer, BufferDeleter>;
+
  public:
   static constexpr size_t kMaxMsgSize = 64 * 1024 * 1024;
-  struct Message {
-    const uint8_t* start = nullptr;
-    uint32_t len = 0;
-    uint32_t field_id = 0;
-    bool fatal_framing_error = false;
-    const uint8_t* end() const { return start + len; }
-    inline bool valid() const { return !!start; }
+  // The payload of one field, without its preamble. Keeps the buffer it points
+  // into alive, so it stays valid across further writes and can be moved to
+  // another thread.
+  class Message {
+   public:
+    Message() = default;
+    ~Message();
+    Message(Message&&) noexcept;
+    Message& operator=(Message&&) noexcept;
+    Message(const Message&) = delete;
+    Message& operator=(const Message&) = delete;
+
+    const uint8_t* data() const { return start_; }
+    const uint8_t* end() const { return start_ + len_; }
+    uint32_t size() const { return len_; }
+    uint32_t field_id() const { return field_id_; }
+    bool valid() const { return !!start_; }
+    bool fatal_framing_error() const { return fatal_framing_error_; }
+
+   private:
+    friend class ProtoRingBuffer;
+    BufferPtr buf_;
+    const uint8_t* start_ = nullptr;
+    uint32_t len_ = 0;
+    uint32_t field_id_ = 0;
+    bool fatal_framing_error_ = false;
   };
 
   ProtoRingBuffer();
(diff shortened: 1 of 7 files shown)


include/perfetto/ext/protozero/proto_ring_buffer.h
```diff
@@ -155,24 +183,27 @@ class ProtoRingBuffer {
...
+  void SwitchToFreshBuffer(size_t capacity);
   void FinishWrite(size_t size_written);
 
-  perfetto::base::PagedMemory buf_;
+  // Refcounted; the ring buffer holds one reference, each Message another.
+  Buffer* buf_ = nullptr;
+  // The buffer left behind by the last SwitchToFreshBuffer(), reused once the
+  // Messages pinning it are released.
+  Buffer* spare_ = nullptr;
   bool failed_ = false;  // Set in case of an unrecoverable framing faiulre.
```
safayat-google: nit: failure

src/protozero/proto_ring_buffer.cc
safayat-google: nit: `ended up`

src/protozero/proto_ring_buffer.cc
```diff
@@ -100,21 +189,28 @@ ProtoRingBuffer::WriteHandle ProtoRingBuffer::BeginWrite(size_t data_len) {
...
 
   // Nothing can be tokenized any more, so recycle the whole buffer for the
   // bytes EndWrite() is about to drop.
-  if (PERFETTO_UNLIKELY(failed_))
+  if (PERFETTO_UNLIKELY(failed_) && buf_->IsUniquelyOwned())
     rd_ = wr_ = 0;
 
+  // Messages handed out still point into |buf_|, so neither the recompaction
+  // nor the rewind below may touch its bytes while any of them is alive.
+  if (PERFETTO_UNLIKELY(!buf_->IsUniquelyOwned() &&
```
safayat-google: it is completely exclusive from the `if (data_len > avail)` condition on line 214?
If not, we may allocate twice (another in `SwitchToFreshBuffer` on line 250.)

src/protozero/proto_ring_buffer.cc
```diff
@@ -76,9 +86,88 @@ ProtoRingBuffer::Message TryReadProtoMessage(const uint8_t* start,
...
+  if (this == &other)
+    return *this;
+  Buffer::DecRef(buf_);
+  buf_ = other.buf_;
+  start_ = other.start_;
+  len_ = other.len_;
+  field_id_ = other.field_id_;
+  fatal_framing_error_ = other.fatal_framing_error_;
+  other.buf_ = nullptr;
+  other.start_ = nullptr;
```
safayat-google: Reset other's `len_` `field_id_`, and `fatal_framing_error_` as well?

Also can we use std::exchange to do copy+reset in one line?

LalitMaganti:
Friendly ping!

src/protozero/proto_ring_buffer_unittest.cc
```diff
@@ -272,5 +281,161 @@ TEST_F(ProtoRingBufferTest, GivingUpDoesNotAbort) {
...
+  for (int i = 0; i < 1000; i++) {
+    Write(&buf, last_msg_.data(), last_msg_.size());
+    inflight = buf.ReadMessage();  // Releases the previous one.
+    ASSERT_EQ(inflight, expected);
+  }
+  // Each message takes up nearly a whole buffer, so every write has to move
+  // onto another one. Handing them off must still not allocate per message.
+  EXPECT_EQ(buf.num_buffers_for_testing(), 2u);
+}
+
```
safayat-google, line 443: nit: should we also add some multithreaded test case?

include/perfetto/ext/protozero/proto_ring_buffer.h
```diff
@@ -155,24 +190,29 @@ class ProtoRingBuffer {
...
+  void RecycleBuffer(BufferPtr buffer);
   void FinishWrite(size_t size_written);
 
-  perfetto::base::PagedMemory buf_;
-  bool failed_ = false;  // Set in case of an unrecoverable framing faiulre.
+  // Refcounted; the ring buffer holds one reference, each Message another.
+  BufferPtr buf_;
+  // At most one previous allocation, at least as large as the active buffer.
+  // It can be reused once all Messages referring to it have been released.
+  BufferPtr spare_;
```
safayat-google, line 210: nit: should we explore double/triple buffering? If the const memory addition is not high, this should reduce overall allocation for longer queue of threads.

src/protozero/proto_ring_buffer.cc
```diff
@@ -76,10 +87,86 @@ ProtoRingBuffer::Message TryReadProtoMessage(const uint8_t* start,
...
+    return std::move(spare_);
+  ++num_buffers_;
+  return Buffer::Create(capacity);
+}
+
+void ProtoRingBuffer::RecycleBuffer(BufferPtr buffer) {
+  // Keep only the most recent buffer, and only if it can satisfy future
+  // replacements. In particular, growth must not cache an undersized buffer.
+  spare_.reset();
+  if (buffer->size() >= buf_->size())
```
safayat-google, line 166: So, the current buffer and the spare buffer size will never decrease?
If that is the case, and suddenly we receive a large packets (say 32mb), and later all the packets are very small, shall we end up using extra memory here?

safayat-google (approved):
LGTM.

There are some scope of improvement of some edge case.
For example, say a thread has only one remaining reference for a small message in a huge 64Mb ring buffer. Should we shrink the ring buffer size in that case?
Not sure how frequently we will hit those and definitely it will make things complex if we want to handle it. So, I don't have a strong opinion on it.

src/protozero/proto_ring_buffer_unittest.cc
LalitMaganti: Done, see MessagesCanBeConsumedOnAnotherThread.

include/perfetto/ext/protozero/proto_ring_buffer.h
LalitMaganti: Deferring. Allocation only happens when a write needs room while both the active and spare buffers are still shared, and the header notes this. If profiling the zero-copy stack shows churn, turning the spare into a small pool is a straightforward follow-up.

src/protozero/proto_ring_buffer.cc
LalitMaganti: Yes, they never shrink, same as before this PR. Shrinking would just trade this for hysteresis, and releasing pages needs bookkeeping around the refcounts. Neither feels worth it, so I've left it.

PR #7284, opened 2026-08-28
tp: dispatch all registered extension fields in ParseTracePacket

parthsane:
When a TracePacket contains multiple registered extension fields (e.g., Android's ProcessStateDataSource emitting both android_process_state and android_freezer_state in a single snapshot packet), ParseTracePacket previously returned on the first registered field, dropping any subsequent extension fields in the same packet.

Although canonical Perfetto packets usually contain a single payload in the oneof data field, out-of-tree extensions are independent optional fields and can be co-located in a single packet.

This change iterates through all registered fields present in the packet before returning, ensuring all extension payloads are dispatched.

Welcome to Perfetto!

For more details please see


LalitMaganti:
> Although canonical Perfetto packets usually contain a single payload in the oneof data field, out-of-tree extensions are independent optional fields and can be co-located in a single packet.

No sorry this is not what we want to enable. We should class this as a producer bug and keep maintaining the principle that only a single extension to TracePacket can be set. 

If you want to send multiple fields, break them into multiple packets.

parthsane:
Ack. will fix

PR #7291, opened 2026-09-01
perfetto_cmd: Support parsing TraceConfig from compressed persistent traces

safayat-google:
When compression is enabled (e.g. COMPRESSION_TYPE_DEFLATE), the initial
TraceConfig packet is stored inside `TracePacket.compressed_packets`.
Update `PerfettoCmd::ParseTraceConfigFromMmapedTrace` to decompress the
first chunk using `ZlibDecompress()` and extract the TraceConfig so
compressed persistent traces can be uploaded after a reboot.
    
Test: perfetto_unittests --gtest_filter="PerfettoCmdlineUnitTest.*"
Bug: 382209797


src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -530,26 +535,56 @@ std::optional<TraceConfig> PerfettoCmd::ParseTraceConfigFromMmapedTrace(
...
+        trusted_uid_field.as_int32() == kTrustedUid) {
+      TraceConfig trace_config;
+      if (trace_config.ParseFromArray(trace_config_field.data(),
+                                      trace_config_field.size())) {
+        return trace_config;
+      }
+    }
 
-    int32_t uid_value = trusted_uid_field.as_int32();
+#if PERFETTO_BUILDFLAG(PERFETTO_ZLIB)
```
sashwinbalaji, line 552: you would need to handle zstd also  kZstdCompressedPacketsFieldNumber

in that case maybe worth reusing the existing tp decompressor (would have to move to a common place)? Or should we have a separate metadata fiel to avoid decompressing altogether especially during boot time

cc: @primiano

primiano: TL;DR we need to pick up aosp/2815935. I intnded to do that a while ago, then as usually happen never had the time to do it fully.

This is the Nth time we hit this. 
I think the right solution is to not compress this stuff so we can inspect it in various places

sashwinbalaji: trace config packet is  32KB. compressed is around 10K so we will adding an extra 22KB

safayat-google: #7300 is uploaded from the cherrypick of aosp/2815935 and I am closing this one.

PR #7294, opened 2026-09-01
tp+ui: scope pprof archive members and intern fallback mappings

gignat-dev:
### Summary

  Improves multi-pprof archive support in Trace Processor by correctly scoping members and interning fallback mappings, and fixes a UI profile selection bug on the Aggregate Profiles page:

  - Archive member scoping (TP): Scopes imported pprofs by their member filename within archives (walking past unnamed gzip layers) instead of defaulting all to pprof_file. The file ID is passed directly to the PprofTraceReader constructor from the parser, with ancestor naming encapsulated in TraceFileTracker.
  - Fallback mapping & frame interning (TP): Names the interned fallback mapping "[pprof unknown]" and uses VirtualMemoryMapping::InternFrameAndCheckIfNew() to skip re-symbolizing frames already interned by earlier profiles in an archive, preventing frames/mappings/symbols from scaling linearly with profiles × locations.
  - Profile selector fix (UI): Fixes a state clobbering bug in AggregateProfilesPage where switching profiles in the dropdown caused the selector to snap back to the previous profile.
  - Controls layout & styles (UI):
    - Consolidates the profile selector onto the left of the controls row alongside help buttons, eliminating a redundant row of chrome above the flamegraph.
    - Corrects the stylesheet root class (.pf-pprof-page → .pf-aggregate-profiles-page) so page padding, selector sizing, and control styles take effect.

  ### Visual comparison on a ZIP with 100 pprofs
  
  ### Before: can't select the pprof
image

### After: can select the pprof
image

  ### Testing

  • TP diff tests: tools/diff_test_trace_processor.py <out>/trace_processor_shell --name-filter=".\*pprof.\*" (13/13 passed)
  • UI unit tests: ui/run-unittests -n (all passed)


zezeozue (approved):

src/trace_processor/importers/common/trace_file_tracker.h
```diff
@@ -44,6 +45,15 @@ class TraceFileTracker {
   void StartParsing(tables::TraceFileTable::Id id, TraceImporterId trace_type);
   void DoneParsing(tables::TraceFileTable::Id id, size_t size);
```
LalitMaganti, line 46: This is unnecessary. The correct fix is to pass this into the constructor from forwarding trace parser.

src/trace_processor/importers/pprof/pprof_trace_reader.cc
```diff
@@ -182,7 +188,11 @@ base::Status PprofTraceReader::ParseProfile() {
       }
     }
     if (!mapping) {
-      mapping = &context_->mapping_tracker->CreateDummyMapping("[unknown]");
+      // Interned so profiles in an archive share one mapping and their
+      // frames dedupe.
+      CreateMappingParams params;
```
LalitMaganti, line 191: We ideally should make this pprof unknown.

src/trace_processor/importers/pprof/pprof_trace_reader.cc
```diff
@@ -214,6 +224,15 @@ base::Status PprofTraceReader::ParseProfile() {
         mapping->InternFrame(rel_pc, storage->GetString(frame_name_id));
     location_to_frame[loc_decoder.id()] = frame_id;
 
+    // Frames interned by an earlier profile already carry their symbols.
```
LalitMaganti, line 225: This is a hack. I need to understand what this is patching to suggest how to fix this.

src/trace_processor/importers/pprof/pprof_trace_reader.cc
```diff
@@ -283,6 +302,23 @@ base::Status PprofTraceReader::ParseProfile() {
...
   }
 
+  // Scope each profile by its source file, so archive members stay
+  // distinguishable. Gzipped members parse under an unnamed decompression
+  // file, so walk up to the nearest named ancestor.
+  StringId scope_id = pprof_file_string_id_;
+  const auto& trace_files = context_->storage->trace_file_table();
+  auto cur = file_id_;
+  while (cur.has_value()) {
+    auto row = trace_files[*cur];
```
LalitMaganti: Sorry I don't like this at all, this is also a hack. We shouldn't be reading tables, if we need this, it should be passed into the constructor.

src/trace_processor/importers/pprof/pprof_trace_reader.cc
gignat-dev: Since frames get interned across profiles in an archive via `mapping->InternFrame()`, later profiles seeing the `same frame` would otherwise keep creating `duplicate rows in symbol_table` and overwriting `frame.symbol_set_id` (leaving the older symbol rows orphaned).

For archives with lots of profiles (e.g. 100 heapdumps), that blows up symbol_table with orphaned duplicates and wastes memory. 

This check just skips re-creating the symbol entries if the interned frame already has them populated.

gignat-dev: E.g. 

Executing this sql on the same `100_heapdumps` archive
```
SELECT
      -- Total distinct frames:
      (SELECT COUNT(*) FROM stack_profile_frame) AS total_frames,
      (SELECT COUNT(symbol_set_id) FROM stack_profile_frame) AS frames_with_symbols,

      -- Symbol table entries:
      (SELECT COUNT(*) FROM stack_profile_symbol) AS total_symbol_rows,
      (SELECT COUNT(DISTINCT symbol_set_id) FROM stack_profile_symbol) AS total_symbol_sets,

      -- Orphaned symbol sets (symbol sets created that no frame references):
      (SELECT COUNT(DISTINCT symbol_set_id) FROM stack_profile_symbol) -
      (SELECT COUNT(DISTINCT symbol_set_id) FROM stack_profile_frame) AS orphaned_symbol_sets;
```

Without the skip block:

| total_frames | frames_with_symbols | total_symbol_rows | total_symbol_sets | orphaned_symbol_sets |
| --- | --- | --- | --- | --- |
| 5225 | 5225 | 17611 | 17611 | 12386 |

so a lot of symbol_table entries are just recreated unnecessarily.


With the skip block:

| total_frames | frames_with_symbols | total_symbol_rows | total_symbol_sets | orphaned_symbol_sets |
| --- | --- | --- | --- | --- |
| 5225 | 5225 | 5225 | 5225 | 0 |

gignat-dev: This is still a hack (i.e. reading from raw storage tables), but I hope the issue we try to solve is clearer.

LalitMaganti: We need to add a proper deduplicating tracker not just read the table. 

In general reading the tables directly like this is heavily discouraged.

gignat-dev: I'm thinking of making 
```
FrameId frame_id =
        mapping->InternFrame(rel_pc, storage->GetString(frame_name_id))
```

also return a boolean `is_new` that says if the frame is newly created or reused. If new, then we create the symbol_table entries.

Wdyt?

LalitMaganti: We cannot change the InternFrame method itslef but I would be in favor of adding another method which did this for example.

gignat-dev: Yeah, we can add a `InternFrameAndCheckIfNew`, which I guess will use InternFrame in a way or another.

Thanks for the directions, I'll come back with a potential solution.

gignat-dev: Done.

Updated `VirtualMemoryMapping` to expose `InternFrameAndCheckIfNew()`, returning `{frame_id, is_new_frame}` (using the internal `was_inserted` flag it already tracks). 

The reader now just checks `if (!is_new_frame) continue;` instead of querying the frame table row.

src/trace_processor/importers/pprof/pprof_trace_reader.cc
gignat-dev: Done

src/trace_processor/importers/common/trace_file_tracker.h
gignat-dev: Done. 
Removed `CurrentFile()` and passed `file_id` directly into `PprofTraceReader`'s constructor via `PprofImporter::CreateReader()`.

src/trace_processor/importers/pprof/pprof_trace_reader.cc
gignat-dev: Done. 

Removed the table walk from `PprofTraceReader`. 
The reader now uses the `file_id` passed to the constructor and queries `TraceFileTracker::GetName(file_id)`, which handles ancestor name resolution internally across decompression layers like this:

```cc
  StringId inherited_name = parent ? names_[parent->value] : kNullStringId;
  bool named = !name.is_null() && !context_->storage->GetString(name).empty();
  names_.push_back(named ? name : inherited_name);
```

Please let me know if this solution is better.

LalitMaganti (approved):

PR #7296, opened 2026-09-01
ui: Add 'Show only direct flows' setting to FlowEvents plugin

### LINKED ISSUE #1004: Add UI control over only rendering flow one step forward
Hello, I have noticed an issue.
There are three slices. Slice 1 has `track_event.flow_ids` as `[1]`, slice 3 has `track_event.flow_ids` as `[2]`, and slice 2 has `track_event.flow_ids` as `[1, 2]`.
When I click on slice 1 with my mouse, based on my understanding, the arrow between slice 2 and slice 3 should not appear, but it does.
Additionally, the arrow on the left is thicker, while the arrow on the right is thinner. Is this intended by design? 
If I want to hide the arrow on the right, is it possible to achieve that?

example UI:
Image

stevegolton:
Add a new setting `dev.perfetto.DirectFlowsOnly` which, when enabled, only renders the immediate flows for a given slice when selected, rather than following the entire flow chain via BFS. This is useful for traces with dense flow graphs (e.g. build dependencies) where the full-chain traversal causes the UI to grind to a halt.

The setting is registered by the FlowEvents plugin and defaults to off (preserving current behaviour). When enabled, the sliceSelected query uses a simple `WHERE` clause on the flow table instead of the `directly_connected_flow()` table function.

Fixes https://github.com/google/perfetto/issues/1004

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -19,6 +19,10 @@ Unreleased:
     * Memscope: native callstack snippet table no longer truncates frames.
     * Intelletto: registered tools with Chrome's WebMCP API.
     * Serialize workspaces to permalinks.
+    * Added a "Show only direct flows" setting to the FlowEvents plugin: when
+      enabled, selecting a slice shows only the flows directly connected to it
+      (one step) instead of following the entire flow chain. Useful for traces
+      with dense flow graphs (e.g. build dependencies).
   SDK:
     *
 
--- a/ui/src/core_plugins/dev.perfetto.FlowEvents/flow_manager.ts
+++ b/ui/src/core_plugins/dev.perfetto.FlowEvents/flow_manager.ts
@@ -25,6 +25,7 @@ import type {
 } from '../../public/selection';
 import type {Engine} from '../../trace_processor/engine';
 import type {Raf} from '../../public/raf';
+import type {Setting} from '../../public/settings';
 
 const SHOW_INDIRECT_PRECEDING_FLOWS_FLAG = featureFlags.register({
   id: 'showIndirectPrecedingFlows',
@@ -49,6 +50,7 @@ export class FlowManager {
     private trackMgr: TrackManager,
     private selectionMgr: SelectionManager,
     private raf?: Raf,
+    private directFlowsOnlySetting?: Setting<boolean>,
   ) {}
 
   // TODO(primiano): the only reason why this is not done in the constructor is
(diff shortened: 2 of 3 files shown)


BGR360:
Hey @stevegolton, thank you for giving this a shot!

I tried it out on [this trace (77 MB uncompressed)](https://github.com/user-attachments/files/31750521/2026-06-17T18_30_21-07_00.json.zip), but the webpage became unresponsive after clicking one of the slices. This happened both when the new setting was disabled and when it was enabled.

When I last tried this trace in June, the UI did slow down a bit when I clicked a slice with a huge flow closure, but it never locked up like this.

Is there some extra CPU or memory cost being paid with this change? It doesn't seem like there should be based on looking at the diff...

Let me know if there's any logs or traces I could collect to help you diagnose the issue, if you're unable to reproduce.

BGR360:
Just checked, and the latest canary on ui.perfetto.dev does not lock up with that same trace. For reference, I clicked the big `qonstruct/rustc.py` slice in the middle:

image

stevegolton:
Ah, that doesn't bode well!

Would you mind testing on the tip of autopush? I'm curious as to whether https://github.com/google/perfetto/pull/7030 might have introduced a bug.

You can go [here](https://ui.perfetto.dev/#!/flags/releaseChannel) to switch to autopush.

stevegolton:
Ah, I think it might have been this bug: https://github.com/google/perfetto/pull/7334

I also somehow missed the fact that you uploaded a trace - thank you very much! After #7334 the performance seems a lot better, so I'm pretty sure it's the cause. I'll rebase this PR on top of that patch after it lands.

stevegolton:
Just rebased - the CI build should appear any minute.

image

LalitMaganti (approved):

BGR360:
Nice!! Looks like the gh-actions bot didn't send the new CI build, but I found it in the actions output: https://storage.googleapis.com/perfetto-ci-artifacts/gh-33785029854-1-ui/ui/index.html

Playing around with it, it works great!

stevegolton:
> Nice!! Looks like the gh-actions bot didn't send the new CI build, but I found it in the actions output: https://storage.googleapis.com/perfetto-ci-artifacts/gh-33785029854-1-ui/ui/index.html
> 
> Playing around with it, it works great!

Thanks for testing!

Going forward it would be good to improve the UX. I think the first step is to co-locate all flow related panels into the same plugin, which would involve changing some of the plugin API. That would allow us to put the setting toggle on the single slice selection details panel next to the flow events.

These API changes could be happening anyway at some point soon for other reasons, but I can't really guarantee a timeline.

Anyway, glad the setting does the job for now.

PR #7300, opened 2026-09-01
traced: Don't compress clock snapshot, trace config and uuid

safayat-google:
These packets can be useful for services that consume the trace on-device. Keep them uncompressed even in presence of a compression_type setting, so that on-device services can inspect trace metadata without having to uncompress the trace.

- cherrypick of aosp/2815935

Test: perfetto_unittests --gtest_filter=TracingServiceImplTest.*Compr*
Bug: 302357655


src/tracing/service/tracing_service_impl.cc
```diff
@@ -2939,7 +2946,8 @@ void TracingServiceImpl::MaybeFilterPackets(TracingSession* tracing_session,
 
 void TracingServiceImpl::MaybeCompressPackets(
     TracingSession* tracing_session,
-    [[maybe_unused]] std::vector<TracePacket>* packets) {
+    [[maybe_unused]] std::vector<TracePacket>* packets,
```
sashwinbalaji, line 2954: maybe_unused can eb removed

src/tracing/service/tracing_service_impl.cc
sashwinbalaji, line 2734: we will be uncompressing way more than just trace config

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2948,21 +2956,44 @@ void TracingServiceImpl::MaybeCompressPackets(
...
+
+  const auto nskip = static_cast<ssize_t>(skip_compression_of_first_n_packets);
+  std::vector<TracePacket> packets_to_compress(
+      std::make_move_iterator(packets->begin() + nskip),
+      std::make_move_iterator(packets->end()));
+  packets->erase(packets->begin() + nskip, packets->end());
+  compress_fn(&packets_to_compress);
+  packets->insert(packets->end(),
+                  std::make_move_iterator(packets_to_compress.begin()),
+                  std::make_move_iterator(packets_to_compress.end()));
```
sashwinbalaji: nit: if no compressions configured we will still be doing all the steps of move, erase and insert which can be avoided.

src/tracing/service/tracing_service_impl.cc
safayat-google: Thanks, while resolving the conflict added it to wrong place, updated

src/tracing/service/tracing_service_impl.cc
safayat-google: Updated to check the compression mode and skipped all when no compression is enabled.

src/tracing/service/tracing_service_impl.cc
sashwinbalaji, line 2722: we should skip these also ?

sashwinbalaji: just trace config and uuid should be enough right ?

safayat-google: I think the intention was to skip compressing the `triggers` as well from the original commit message.

@primiano is it ok to keep the triggers as compressed?

safayat-google: Only the trace config and UUID are uncompressed and the PR title is updated.

src/tracing/service/tracing_service_impl_unittest.cc
```diff
@@ -2802,36 +2804,8 @@ TEST_F(TracingServiceImplTest, CompressionZstdReadIpc) {
...
   EXPECT_THAT(compressed_packets, Each(PacketIsCompressed()));
-
-  // Test that the trace config and trace uuid are NOT compressed and can be
-  // read in clear.
-  EXPECT_THAT(
-      compressed_packets,
-      Contains(Property(&protos::gen::TracePacket::has_trace_config, true)));
-  EXPECT_THAT(
-      compressed_packets,
-      Contains(Property(&protos::gen::TracePacket::has_trace_uuid, true)));
```
sashwinbalaji: nit: can keep this.

src/tracing/service/tracing_service_impl.h
```diff
@@ -344,7 +344,8 @@ class TracingServiceImpl : public TracingService {
 
   // If `*tracing_session` has compression enabled, compress `*packets`.
```
sashwinbalaji, line 345: nit: this can be updated to cover the skipping

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2685,8 +2685,22 @@ std::vector<TracePacket> TracingServiceImpl::ReadBuffers(
   *has_more = false;
 
   std::vector<TracePacket> packets;
+  size_t skip_compression_of_first_n_packets = 0;
```
sashwinbalaji: nit: you can just do const size_t skip_compression_of_first_n_packets = packets.size(); in line 2702 directly.

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2948,21 +2963,59 @@ void TracingServiceImpl::MaybeCompressPackets(
...
-    ZlibCompressFn(packets);
-    return;
+    is_compression_enabled = true;
   }
 #endif
+  if (!is_compression_enabled)
+    return;
+
+  auto compress_fn = [&]([[maybe_unused]] std::vector<TracePacket>* target) {
+#if PERFETTO_BUILDFLAG(PERFETTO_ZSTD)
```
sashwinbalaji: don't have a direct answer but anyway to write this such that the ifdefs for compression is in one place only instead of 2

src/tracing/service/tracing_service_impl.h
safayat-google: Comment is updated

src/tracing/service/tracing_service_impl.cc
safayat-google: Added the compression + skip logic in a separate func to decouple the logics

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2937,9 +2949,34 @@ void TracingServiceImpl::MaybeFilterPackets(TracingSession* tracing_session,
       static_cast<uint64_t>((end - start).count());
 }
 
+// Helper to re-use logics for different compression (zlib, zstd, etc.).
+template <typename CompressFn>
+void MaybeCompressPacketsWithCompressionFn(
```
sashwinbalaji: instead of a free function in this file, it can be in anon namespace or a lambda within MaybeCompressPackets

sashwinbalaji (approved):

src/tracing/service/tracing_service_impl.cc
safayat-google: moved inside the same func as a lambda

src/tracing/service/tracing_service_impl.cc
```diff
@@ -2687,6 +2687,19 @@ std::vector<TracePacket> TracingServiceImpl::ReadBuffers(
   std::vector<TracePacket> packets;
   packets.reserve(1024);  // Just an educated guess to avoid trivial expansions.
 
+  if (!tracing_session->config.builtin_data_sources().disable_trace_config()) {
```
primiano, line 2696: can we also keep the clock snapshot first?
so you do:
- snapshot
- config
- uuid

and the rest compressed?

TraceProcessor makes assu,ptioms left and right about ClockSnap and at some poimt maybe somebody adds a dep from there to the TraceConfig. I would not change the relative order of those

primiano (approved):
LG with one comment

src/tracing/service/tracing_service_impl.cc
safayat-google: moved clock snapshot into the uncompressed block before config and uuid

PR #7303, opened 2026-09-01
Promote BatteryStatus and PlugType to top level tracks.

### LINKED ISSUE #7302: Multi-machine migration broke battery counters
The migration in #7243 is causing battery counters such as the charging state to break:

Image

I'm working on a quick fix, but this whole area is quite fragile so want to use this to discuss a better long term plan.

The change in #7243 sensibly tries to use `counter_leading_intervals` partitioned by `track_id`. The issue is that many of the battery related counters are populated by two data sources and and up in two tracks:
1. We get them from regular atrace.
2. We poll them using the `android.system_property` datasource. This is primarily to populate the start state, but is often configured to still poll multiple times per trace.

Because they're in two tracks, we end up with two concurrent spans like the picture above, even on a single-machine trace. The quick fix is to change `counter.track_id` to `counter.machine_id AS track_id` (and downstream) so that counter leading intervals partitions only by machine.

@LalitMaganti: do you have any ideas on how to make this less fragile? The [importer here](https://github.com/google/perfetto/blob/56b3c449c968296bda3831d7cd2e0d2140acc9e9/src/trace_processor/importers/proto/android_probes_parser.cc#L670) can't know which process is going to emit the atrace. We _could_ have the atrace importer find the same events and emit it using the same track blueprint, but that doesn't seem great.

rzuklie:
These previously were top level when emit by the periodic poller, but process level when emit by atrace. This leads to two separate track_ids that can lead to incorrect results when trying to use them normally. This fixes #7302

### DIFF
--- a/src/trace_processor/importers/common/tracks_common.h
+++ b/src/trace_processor/importers/common/tracks_common.h
@@ -210,6 +210,18 @@ inline constexpr auto kAndroidScreenStateBlueprint =
                              tracks::DimensionBlueprints(),
                              tracks::StaticNameBlueprint("ScreenState"));
 
+inline constexpr auto kAndroidBatteryStatusBlueprint =
+    tracks::CounterBlueprint("battery_status",
+                             tracks::UnknownUnitBlueprint(),
+                             tracks::DimensionBlueprints(),
+                             tracks::StaticNameBlueprint("BatteryStatus"));
+
+inline constexpr auto kAndroidPlugTypeBlueprint =
+    tracks::CounterBlueprint("plug_type",
+                             tracks::UnknownUnitBlueprint(),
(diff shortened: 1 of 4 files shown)


LalitMaganti (approved):

LalitMaganti (changes requested):
https://github.com/google/perfetto/pull/7308 seems to conflict with the purpose of the change.

stevegolton (approved):

stevegolton:
> #7308 seems to conflict with the purpose of the change.

I think they complement each other. We can land both.

LalitMaganti (approved):
I personally think this sort of merging should be happening at the SQL level only but to unblock things as they are we can land this now. 

We will need to revisit this once PerfettoSQL next is live.

PR #7308, opened 2026-09-02
tp: Merge android battery/screen state tracks from the same machine

stevegolton:
PR #7243 made battery charging, doze, and screen state tables multi-machine aware by passing counter.track_id to counter_leading_intervals, presumably under the impression that each machine would only have single track. This caused intervals to be partitioned by individual track rather than by machine, leading to overlapping intervals on devices with multiple tracks per machine and breaking downstream SPAN_JOIN metrics.

Fix this by passing COALESCE(counter_track.machine_id, 0) as the track_id to counter_leading_intervals so tracks on the same machine are merged into a single non-overlapping timeline.

Rolled from google3 cl/975030561.

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/battery/charging_states.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/battery/charging_states.sql
@@ -39,7 +39,16 @@ AS
 -- union is populated but not both.
 WITH
   _counter AS (
-    SELECT counter.id, ts, counter.track_id, value
+    SELECT
+      counter.id,
+      ts,
+      -- Note: counter_leading_intervals! partitions on track_id. We alias
+      -- machine_id to track_id to partition intervals by machine rather
+      -- than by individual track, merging all tracks on the same machine.
+      -- TODO(stevegolton): Fix this when PerfettoSQL supports customizable
+      -- partition columns in counter_leading_intervals!.
+      COALESCE(counter_track.machine_id, 0) AS track_id,
+      value
     FROM counter
     JOIN counter_track
       ON counter_track.id = counter.track_id
@@ -51,7 +60,7 @@ WITH
       intervals.id,
       intervals.ts,
       intervals.dur,
-      counter_track.machine_id,
+      intervals.track_id AS machine_id,
       CASE intervals.value
         WHEN 2 THEN 'charging'
         WHEN 3 THEN 'discharging'
@@ -67,8 +76,6 @@ WITH
         WHEN 5 THEN 'Full'
       END AS charging_state
     FROM counter_leading_intervals!(_counter) AS intervals
-    JOIN counter_track
-      ON counter_track.id = intervals.track_id
     WHERE
       dur > 0
   )
@@ -80,7 +87,7 @@ SELECT
   ROW_NUMBER() OVER () AS id,
   ts,
   dur,
-  machine_id,
+  COALESCE(machine_id, 0) AS machine_id,
   COALESCE(short_charging_state, 'unknown') AS short_charging_state,
   COALESCE(charging_state, 'Unknown') AS charging_state
 FROM _intervals_fill_gaps!((machine_id), (short_charging_state, charging_state), _intervals);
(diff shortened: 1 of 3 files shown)


src/trace_processor/perfetto_sql/stdlib/android/battery/charging_states.sql
```diff
@@ -39,7 +39,11 @@ AS
 -- union is populated but not both.
 WITH
   _counter AS (
-    SELECT counter.id, ts, counter.track_id, value
+    SELECT
+      counter.id,
+      ts,
+      COALESCE(counter_track.machine_id, 0) AS track_id,
```
LalitMaganti, line 50: Why are we coalescing? We should leave it machine_id nul. Also why are we renaming machine_id as track_id? I'm quite confused by this.

src/trace_processor/perfetto_sql/stdlib/android/battery/charging_states.sql
```diff
@@ -51,7 +55,7 @@ WITH
       intervals.id,
       intervals.ts,
       intervals.dur,
-      counter_track.machine_id,
+      intervals.track_id AS machine_id,
```
LalitMaganti, line 63: Uhhhh we seem to now be renaming back? This is super confusing.

src/trace_processor/perfetto_sql/stdlib/android/battery/doze.sql
```diff
@@ -32,7 +32,11 @@ CREATE PERFETTO TABLE android_light_idle_state(
 AS
 WITH
   _counter AS (
-    SELECT counter.id, ts, counter.track_id, value
+    SELECT
+      counter.id,
+      ts,
```
LalitMaganti, line 37: As above. I don't understand this b

LalitMaganti (changes requested):

src/trace_processor/perfetto_sql/stdlib/android/battery/charging_states.sql
stevegolton: We're (ab)using the fact that counter_leading_intervals!() partitions by track_id to partition by machine_id

Here's the process:
1. Rename `machine_id` to `track_id`.
2. Run it through `counter_leading_intervals!()` - tracks will be partitioned by machine_id disguised as track_id.
3. Rename `track_id` back to `machine_id` for the resulting table.

Yes it's a hack.

Before https://github.com/google/perfetto/pull/7243, these tables used `0 as track_id` to merge all counter tracks together, but that assumes only one machine. Now with multi machine, we want to merge track with the same machine id together.

Edit: Please do let me know if there's a better way to do this that doesn't involve the hack!

LalitMaganti: Ryans patch in https://github.com/google/perfetto/pull/7303 seems much more proper fix for this issue.

stevegolton: Let's go with that then.

Ryan's patch didn't exist when I was fixing this in google3. This is simply a roll of what I fixed internally.

stevegolton: Actually, I think this patch is still necessary as Ryan's patch only handles the BatteryStatus track - we still need to fix the test failures caused by the other tracks.

LalitMaganti: I think we need to document these hacks properly and file a bug to fix this. We should stop hacking this when PerfettoSQL Next is around.

LalitMaganti: (document as in both a bug and comments).

stevegolton: Ack.

stevegolton: Added a comment inline in the three files.

LalitMaganti (approved):

src/trace_processor/perfetto_sql/stdlib/android/battery/charging_states.sql
rzuklie: > Actually, I think this patch is still necessary as Ryan's patch only handles the BatteryStatus track - we still need to fix the test failures caused by the other tracks.

What other tracks are failing?

AFAICT, the ScreenState track shouldn't have this issue because it also hoisted. The Doze tracks shouldn't have this issue either because they aren't pulled by the initial state data source, so never end up in a separate track.

stevegolton:
Ryan's PR https://github.com/google/perfetto/pull/7303 fixed the issue. The other two sql files touched in this PR were not necessary to fix the tests.

PR #7309, opened 2026-09-02
ui: AndroidLockContention: link held locks to their slice

gignat-dev:
Emit slice_id and table_name in the held locks debug track query so DebugSliceTrackDetailsPanel can link each held lock back to its underlying slice. Also explicitly include intervals.overlap module.

### DIFF
--- a/ui/src/plugins/com.android.AndroidLockContention/index.ts
+++ b/ui/src/plugins/com.android.AndroidLockContention/index.ts
@@ -158,9 +158,16 @@ export default class AndroidLockContentionPlugin implements PerfettoPlugin {
       id: 'com.android.visualiseHeldLocks',
       name: 'Lock Contention: Visualise held locks',
       callback: async () => {
+        await trace.engine.query('INCLUDE PERFETTO MODULE intervals.overlap;');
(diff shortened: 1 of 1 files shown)


zezeozue (approved):

ui/src/plugins/com.android.AndroidLockContention/index.ts
```diff
@@ -171,13 +178,19 @@ export default class AndroidLockContentionPlugin implements PerfettoPlugin {
...
                     )
                     SELECT
-                    row_number() OVER () AS id,
-                    name AS thread_name,
-                    lock_name,
-                    utid,
-                    ts,
-                    MIN(LEAD(ts) OVER(PARTITION BY lock_name ORDER BY ts), ts + dur) - ts AS dur
-                    FROM lock_held_slices
+                    (
```
LalitMaganti, line 181: Drive by: this is a *very* inefficient queyr you're doing here...

gignat-dev: Will be fixed in #7340

PR #7311, opened 2026-09-02
perfetto: add sdk variant of libperfetto_client_experimental

sashwinbalaji:
Modules built with sdk_version can only link deps that have an sdk
variant, and no perfetto library provided one.

- Add sdk_version: "current" to libperfetto_client_experimental.
  Existing platform/vendor/host consumers are unaffected.
- Make perfetto_flags_c_lib force-read-only and give it sdk_version too.
  Its production-mode deps (libbase, aconfig storage) have no sdk
  variant. All flags read from C++ are fixed read-only, so no behaviour
  change.
- Set stl: "libc++_static" on both. sdk variants otherwise default to
  the NDK "system" STL, which has no libc++.
- Delete the obsolete, non read-only save_all_traces_in_bugreport flag.

Build and Tests pass:
  - http://ag/q/topic:%22libperfetto_sdk_variant%22
  -  http://android-build/abtd/run/L39000030178677719/
  - http://android-build/abtd/run/L87100030178538702/

### DIFF
--- a/perfetto_flags.aconfig
+++ b/perfetto_flags.aconfig
@@ -2,12 +2,6 @@ package: "perfetto.flags"
 container: "system"
 
 flag {
-  name: "save_all_traces_in_bugreport"
-  namespace: "perfetto"
-  description: "This flag controls whether dumpstate invokes perfetto --save-for-bugreport (when disabled, old behavrior) or --save-all-for-bugreport (when enabled, new behavior). save-all serializes all the eligible active traces, rather than only the one with the highest bugreport score"
-  bug: "321196572"
-}
-flag {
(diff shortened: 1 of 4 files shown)


perfetto_flags.aconfig
```diff
@@ -1,12 +1,6 @@
 package: "perfetto.flags"
 container: "system"
 
-flag {
-  name: "save_all_traces_in_bugreport"
```
sashwinbalaji, line 5: need to wait for android cl removing dependency on this flag to be approved before submitting.

safayat-google (approved):

PR #7313, opened 2026-09-02
watchdog_posix: monitor anon RSS from proc/self/statm

### LINKED ISSUE #7338: memory watchdog easily hit in traced_probes when systemd_journald source is enabled
tring to use linux.systemd_journald source, and after a few minutes traced_probes service crashed reporting memory watchdog hit.

traced_probes[33514]: -----BEGIN PERFETTO PRE-CRASH LOG-----
traced_probes[33514]: wdog_reason: 2
traced_probes[33514]: wdog_timeout: 0
traced_probes[33514]: wdog_actual_boot: 0
traced_probes[33514]: wdog_actual_mono: 0
traced_probes[33514]: wdog_actual_cpu: 0
traced_probes[33514]: ds_instance_count: 4
traced_probes[33514]: tracing_session_count: 1
traced_probes[33514]: [608.093]   watchdog_posix.cc:411 Memory watchdog trigger. Memory window of 103649280.000000 bytes is above the 69206016 bytes limit.
Sep 04 13:40:42 ALANLAPTOP traced_probes[33514]: -----END PERFETTO PRE-CRASH LOG-----

safayat-google:
Previously Watchdog measured the memory by reading RSS from proc/self/stat. This may lead to false guardrail violations in case of total RSS is increased from shared library

This change reads anon RSS (rss - shared) from proc/self/statm Also watchdog memory slack is safely returned to 32 Mib as before.

FIXED_READ_ONLY_FLAG_REASON=NATIVE_FLAG_PRE_METADATA_MOUNT
Bug: 540573436
Flag: use_anon_rss_in_watchdog
Fixes: #7338


src/base/watchdog_posix.cc
```diff
@@ -101,6 +101,30 @@ bool ReadProcStat(int fd, ProcStat* out) {
...
+      return false;
+    }
+    if (rd == 0)
+      break;
+    pos += static_cast<size_t>(rd);
+  }
+  PERFETTO_CHECK(pos < sizeof(str));
+  str[pos] = '\0';
+
+  if (sscanf(str, "%lu %lu %lu %lu %lu %lu %lu", &out->size, &out->resident,
```
sashwinbalaji: driveby: you should use PRId64

sashwinbalaji: driveby: I think this won't work on android, you would need PRId64

safayat-google: Done. Updated with `SCNu64` to be precise for sscanf

src/base/watchdog_posix.cc
```diff
@@ -240,6 +267,11 @@ void Watchdog::ThreadMain() {
     PERFETTO_ELOG("Failed to open stat file to enforce resource limits.");
     return;
   }
+  base::ScopedFile statm_fd(base::OpenFile("/proc/self/statm", O_RDONLY));
```
sashwinbalaji: As per https://man7.org/linux/man-pages/man5/proc_pid_statm.5.html,  both resident and shared values will be inaccruate (not sure by how much, need to check) so is using them as the stable one safe ?

Image

src/base/watchdog_posix.cc
```diff
@@ -101,6 +101,33 @@ bool ReadProcStat(int fd, ProcStat* out) {
...
+    if (rd < 0) {
+      PERFETTO_ELOG("Failed to read statm file to enforce resource limits.");
+      return false;
+    }
+    if (rd == 0)
+      break;
+    pos += static_cast<size_t>(rd);
+  }
+  PERFETTO_CHECK(pos < sizeof(str));
+  str[pos] = '\0';
```
sashwinbalaji: we could've a small helper for as it's exactly the same as above one for ReadProcStat

safayat-google: moved into a func

src/base/watchdog_posix.cc
safayat-google: Yes, right. It's a cached inaccurate value.\
in the theoriticaly worst case scenario it can be +/-256kb per thread. Since traced runs a few threads, the number is ~1MB for the whole process. And practically, most of the case the number is just the same as the smaps.

we could use other options for more accurate number, but as the doc mentioned, they are slower ( O(n) compared to O(1) for statm).\
For the guardrail value this should be fine, we are already not including the swap memory into account.

sashwinbalaji (approved):

include/perfetto/ext/base/watchdog_posix.h
```diff
@@ -43,6 +43,23 @@ struct ProcStat {
 
 bool ReadProcStat(int fd, ProcStat* out);
 
+// all are measured in pages
```
primiano, line 46: why this lives here and not in the anonymous namespace in the cc file?

src/base/watchdog_posix.cc
```diff
@@ -240,6 +263,11 @@ void Watchdog::ThreadMain() {
     PERFETTO_ELOG("Failed to open stat file to enforce resource limits.");
     return;
   }
+  base::ScopedFile statm_fd(base::OpenFile("/proc/self/statm", O_RDONLY));
+  if (!statm_fd) {
+    PERFETTO_ELOG("Failed to open statm file to enforce resource limits.");
+    return;
```
primiano: maybe here don't return? i think the rest of the code can handle properly the absence of the FD (check).
beteter to keep going with reduced functionality than bailing out compeltely

primiano (changes requested):
The code itself LGTM but i feel this is a subtle enough change that shoudl go behind a flag plz.
If we get sometjing wrong (selinux, weird kernels, etc) the blast radious is too wide.
The risk is low, but the impact if something happens is to high sorry.

LalitMaganti:
FWIW this also fixes https://github.com/google/perfetto/issues/7338. You might want to add that to the PR description @safayat-google

include/perfetto/ext/base/watchdog_posix.h
safayat-google: We are using these definitions in `watchdog_posix_unittest.cc`

src/base/watchdog_posix.cc
safayat-google: Updated.
it will not return and fallback to rss as before when statm read fails

safayat-google (commented):
> The code itself LGTM but i feel this is a subtle enough change that shoudl go behind a flag plz.

Android flag `use_anon_rss_in_watchdog` is added and statm reading is completely controlled by the flag. Also the guardrail value is set to 32/64 mb base on it.

> FWIW this also fixes https://github.com/google/perfetto/issues/7338. You might want to add that to the PR description @safayat-google

Thanks, The description is updated to include the bug id

src/base/watchdog_posix.cc
```diff
@@ -316,15 +351,29 @@ void Watchdog::ThreadMain() {
     ProcStat stat;
     if (!ReadProcStat(stat_fd.get(), &stat))
       continue;
-    uint64_t cpu_time = stat.utime + stat.stime;
-    uint64_t rss_bytes =
-        static_cast<uint64_t>(stat.rss_pages) * base::GetSysPageSize();
+
+    const uint64_t cpu_time = stat.utime + stat.stime;
+    const uint64_t rss_bytes =
```
primiano, line 357: Why this change? If rss_pages is 0 the multiplication still returns 0 no?
If I see code like this I wonder "hmm there must be a reason why this has been done this way". So either add a comment if there is a mysterious reason or otherwise keep the code predictable (As it was)

src/base/watchdog_posix.cc
```diff
@@ -316,15 +351,29 @@ void Watchdog::ThreadMain() {
...
+        stat.rss_pages > 0
+            ? static_cast<uint64_t>(stat.rss_pages) * base::GetSysPageSize()
+            : 0;
+
+    std::optional<uint64_t> rss_anon_bytes = std::nullopt;
+    if constexpr (PERFETTO_FLAGS(USE_ANON_RSS_IN_WATCHDOG)) {
```
primiano: you don't need this gate here again. If the flag is off, then the statm_fd is already going to be invalid. So the internal if will protect you already.

What is the rationale of the outer flag?

primiano (approved):
LGTM with few minor changes left

src/base/watchdog_posix.cc
safayat-google: Comment added.\
`stat.rss_pages` is signed. So `max(signed_value, 0)` is taken

src/base/watchdog_posix.cc
safayat-google: Removed the outer guard, Thanks

PR #7325, opened 2026-09-03
tracing: add append-only proto encoding for v2 writers

sashwinbalaji:
Protozero normally back-patches nested-message lengths. The v2 ring can expose committed prefixes before their enclosing messages close, so packets written there need append-only framing.

- Add opt-in ProtoGroup framing to the C and C++ Protozero APIs while keeping length-delimited protobuf as the default.

- Use one private C++ framing discriminator for length-delimited messages, ProtoGroup roots and nested ProtoGroup messages without growing Message.

- Preserve the public C message layout and reject incremental STRING and PACKED fields in ProtoGroup mode until buffered construction is needed.

- Carry the selected encoding through C data-source packet creation while keeping the legacy entry point restricted to length-delimited writers.

- Pin the C++ Message size to its previous 32-bit and 64-bit debug and release budgets.

- Rewrite copied ProtoGroup packets into ordinary length-delimited protobuf, validating the wire format and bounding output and nesting links.

- Implement TraceWriterV2 fragmentation, continuation, loss reporting and append-only finalization on SharedRingBufferWriter.

- Make the retained TraceWriterV2 delegate provide the ring so its storage and coordination share one lifetime contract.

- Test exact C and C++ encodings, malformed input, output limits, chunk-boundary finalization, scraping, drops, direct stream reservations, delegate retention and unchanged defaults.


include/perfetto/protozero/message.h
```diff
@@ -59,8 +76,13 @@ class PERFETTO_EXPORT_COMPONENT Message {
   Message() = default;
 
   // Clears up the state, allowing the message to be reused as a fresh one.
+  // This overload predates NestedMessageEncoding and continues to write normal
+  // length-delimited protobuf.
```
sashwinbalaji: not needed comment

include/perfetto/protozero/message.h
```diff
@@ -246,6 +277,11 @@ class PERFETTO_EXPORT_COMPONENT Message {
 
   MessageState message_state_;
 
+  NestedMessageEncoding nested_message_encoding_;
+
+  // The root has no closing byte. Every nested proto-group message has one.
```
sashwinbalaji: why root has no closing byte ?

include/perfetto/protozero/proto_utils.h
```diff
@@ -48,6 +48,23 @@ enum class ProtoWireType : uint32_t {
...
+// support groups, so keep them out of ProtoWireType. Tracing v2 only borrows
+// the start-group wire type for the encoding below.
+constexpr uint32_t kWireTypeStartGroup = 3;
+constexpr uint32_t kWireTypeEndGroup = 4;
+
+// Append-only nested-message encoding used by tracing v2:
+//
+//   open nested field f:  varint((f << 3) | 3)
+//   close current nested: 0x04
+//   root:                 no wrapper and no close byte
```
sashwinbalaji: why so ?

include/perfetto/protozero/proto_utils.h
```diff
@@ -48,6 +48,23 @@ enum class ProtoWireType : uint32_t {
...
+//
+// This is similar to a protobuf group, but the closing byte has field id zero
+// instead of repeating |f|. The reader closes the most recently opened message
```
sashwinbalaji: not needed here: The reader closes the most recently opened message
// and rewrites it as length-delimited protobuf.

include/perfetto/protozero/root_message.h
```diff
@@ -39,9 +39,15 @@ class RootMessage : public T {
   RootMessage(RootMessage&&) = delete;
   RootMessage& operator=(RootMessage&&) = delete;
 
+  // Existing callers continue to write normal length-delimited protobuf.
   void Reset(ScatteredStreamWriter* writer) {
+    Reset(writer, NestedMessageEncoding::kLengthDelimited);
+  }
+
+  // Selects the encoding for this message and all its children.
```
sashwinbalaji: useless comment

include/perfetto/public/abi/data_source_abi.h
```diff
@@ -359,9 +359,27 @@ PERFETTO_SDK_EXPORT void PerfettoDsImplTraceIterateBreak(
...
+enum PerfettoDsPacketEncoding {
+  PERFETTO_DS_PACKET_ENCODING_LENGTH_DELIMITED = 0,
+  PERFETTO_DS_PACKET_ENCODING_PROTO_GROUP = 1,
+};
+
+// Starts a packet and returns its stream writer through `writer`. The return
+// value must be passed to PerfettoPbMsgInitWithEncoding(). Call
+// PerfettoDsTracerImplPacketEnd() when done.
+PERFETTO_SDK_EXPORT enum PerfettoDsPacketEncoding
+PerfettoDsTracerImplPacketBeginWithEncoding(
```
sashwinbalaji: why this needs writer ?

include/perfetto/public/abi/heap_buffer.h
```diff
@@ -42,6 +42,12 @@ PERFETTO_SDK_EXPORT void PerfettoHeapBufferCopyInto(
     void* dst,
     size_t size);
 
+// Appends the contents of the heap buffer to another stream writer.
```
sashwinbalaji: why two stream (ig input and out that's not clear here)

include/perfetto/public/pb_msg.h
```diff
@@ -52,13 +85,56 @@ struct PerfettoPbMsg {
   struct PerfettoPbMsg* parent;
 };
 
-static inline void PerfettoPbMsgInit(struct PerfettoPbMsg* msg,
-                                     struct PerfettoPbMsgWriter* writer) {
+// STRING and PACKED fields can be written piece by piece. Their length is not
+// known until the field closes, so proto-group messages temporarily write them
+// to a heap buffer. Keep |msg_writer| first: the field is recovered from the
```
sashwinbalaji: I dont understand:

 Keep |msg_writer| first: the field is recovered from the
// writer pointer stored in PerfettoPbMsg.

include/perfetto/public/pb_msg.h
```diff
@@ -52,13 +85,56 @@ struct PerfettoPbMsg {
...
+};
+
+// Returns the number of bytes written into |msg|.
+static inline uint32_t PerfettoPbMsgSize(const struct PerfettoPbMsg* msg) {
+  return msg->size & PERFETTO_PB_MSG_SIZE_MASK;
+}
+
+static inline bool PerfettoPbMsgUsesProtoGroup(
+    const struct PerfettoPbMsg* msg) {
+  return (msg->size & PERFETTO_PB_MSG_PROTO_GROUP_BIT) != 0;
```
sashwinbalaji: != 0 seems a bit off to me is it the established pattern ?

include/perfetto/public/pb_msg.h
```diff
@@ -52,13 +85,56 @@ struct PerfettoPbMsg {
...
-  msg->size = 0;
   msg->writer = writer;
   msg->nested = PERFETTO_NULL;
   msg->parent = PERFETTO_NULL;
+  switch (encoding) {
+    case PERFETTO_PB_MSG_ENCODING_LENGTH_DELIMITED:
+      msg->size = 0;
+      return;
+    case PERFETTO_PB_MSG_ENCODING_PROTO_GROUP:
+      msg->size = PERFETTO_PB_MSG_PROTO_GROUP_BIT;
```
sashwinbalaji: why is size storing protogroup bit ?

include/perfetto/public/pb_msg.h
```diff
@@ -72,7 +148,9 @@ static inline void PerfettoPbMsgPatch(struct PerfettoPbMsg* msg) {
 static inline void PerfettoPbMsgPatchStack(struct PerfettoPbMsg* msg) {
   uint8_t* const cur_range_end = msg->writer->writer.end;
   uint8_t* const cur_range_begin = msg->writer->writer.begin;
-  while (msg && cur_range_begin <= msg->size_field &&
+  // Proto-group messages have no size field. Check for null before the pointer
```
sashwinbalaji: phrase better:

Check for null before the pointer
  // comparisons below.

src/shared_lib/heap_buffer.cc
```diff
@@ -34,12 +48,12 @@ void PerfettoHeapBufferCopyInto(struct PerfettoHeapBuffer* buf,
                                 struct PerfettoStreamWriter* w,
                                 void* dst,
                                 size_t size) {
-  auto* shb = reinterpret_cast<protozero::ScatteredHeapBuffer*>(buf);
-  auto* sw = reinterpret_cast<protozero::ScatteredStreamWriter*>(w->impl);
-  sw->set_write_ptr(w->write_ptr);
+  auto* heap_buffer = reinterpret_cast<protozero::ScatteredHeapBuffer*>(buf);
```
sashwinbalaji: call it shb only ?

src/shared_lib/heap_buffer.cc
```diff
@@ -51,6 +65,18 @@ void PerfettoHeapBufferCopyInto(struct PerfettoHeapBuffer* buf,
   }
 }
 
+void PerfettoHeapBufferAppendToStream(struct PerfettoHeapBuffer* buf,
+                                      struct PerfettoStreamWriter* heap_writer,
+                                      struct PerfettoStreamWriter* dst) {
+  auto* heap_buffer = reinterpret_cast<protozero::ScatteredHeapBuffer*>(buf);
```
sashwinbalaji: shb

src/shared_lib/data_source.cc
```diff
@@ -575,16 +575,39 @@ void PerfettoDsImplTraceIterateBreak(
   ds_impl->cpp_type.TraceEpilogue(tls);
 }
 
-struct PerfettoStreamWriter PerfettoDsTracerImplPacketBegin(
-    struct PerfettoDsTracerImpl* tracer) {
+enum PerfettoDsPacketEncoding PerfettoDsTracerImplPacketBeginWithEncoding(
+    struct PerfettoDsTracerImpl* tracer,
+    struct PerfettoStreamWriter* writer) {
```
sashwinbalaji: why this func need writer now ?

src/shared_lib/data_source.cc
```diff
@@ -575,16 +575,39 @@ void PerfettoDsImplTraceIterateBreak(
...
+struct PerfettoStreamWriter PerfettoDsTracerImplPacketBegin(
+    struct PerfettoDsTracerImpl* tracer) {
+  struct PerfettoStreamWriter ret;
+  const enum PerfettoDsPacketEncoding encoding =
+      PerfettoDsTracerImplPacketBeginWithEncoding(tracer, &ret);
+  // Old C SDK callers always write length-delimited messages. Writer setup must
+  // keep them on tracing v1; this CHECK prevents mixed framing if that contract
+  // is ever broken.
+  // TODO(sashwinbalaji): advertise proto-group support before enabling tracing
+  // v2 for independently built C SDK clients.
```
sashwinbalaji: remove this todo

include/perfetto/public/track_event.h
```diff
@@ -380,9 +380,11 @@ static inline bool PerfettoTeLlDynCatEnabled(
 static inline void PerfettoTeLlPacketBegin(
     struct PerfettoTeLlIterator* iterator,
     struct PerfettoDsRootTracePacket* root) {
-  root->writer.writer =
-      PerfettoDsTracerImplPacketBegin(iterator->impl.ds.tracer);
-  PerfettoPbMsgInit(&root->msg.msg, &root->writer);
+  const enum PerfettoDsPacketEncoding encoding =
+      PerfettoDsTracerImplPacketBeginWithEncoding(iterator->impl.ds.tracer,
+                                                  &root->writer.writer);
+  PerfettoPbMsgInitWithEncoding(&root->msg.msg, &root->writer,
```
sashwinbalaji: I am not understand why we are passing the writer. is it because now we have two var encoding and wrtier. If so wrap it in struct or something instead of this kind of hack ?

include/perfetto/public/pb_msg.h
```diff
@@ -31,6 +34,34 @@
...
+  PERFETTO_PB_MSG_ENCODING_PROTO_GROUP = 1,
+};
+
+// PerfettoPbMsg is public ABI, so its layout cannot grow. In proto-group mode
+// the top three bits of `size` store state:
+//
+//   bit 31      proto-group encoding
+//   bit 30      message finalized
+//   bit 29      buffered length-delimited field
+//   bits 28..0  number of bytes written
```
sashwinbalaji: a block diagram would make it clearer than 4 lines I think

include/perfetto/public/pb_msg.h
```diff
@@ -228,7 +352,55 @@ static inline void PerfettoPbMsgEndNested(struct PerfettoPbMsg* parent) {
   parent->nested = PERFETTO_NULL;
 }
 
+static inline size_t PerfettoPbMsgFinalizeBufferedField(
```
sashwinbalaji: make sure this func alings with existing funcs and yeah add a func doc or somethign explaining why we are doiung this

include/perfetto/public/pb_msg.h
```diff
@@ -203,21 +288,60 @@ static inline void PerfettoPbMsgAppendCStrField(struct PerfettoPbMsg* msg,
...
-  PerfettoPbMsgInit(nested, parent->writer);
-  if (PERFETTO_UNLIKELY(
-          PROTOZERO_MESSAGE_LENGTH_FIELD_SIZE >
-          PerfettoStreamWriterAvailableBytes(&parent->writer->writer))) {
-    PerfettoPbMsgPatchStack(parent);
+// Starts a STRING or PACKED field whose contents will be appended
+// incrementally. Length-delimited messages keep their existing
+// reserve-and-patch path. A proto-group message buffers the field and emits one
+// ordinary length-delimited value when the field closes. Raw STRING and PACKED
+// bytes must not be mistaken for proto-group tags by the relay.
```
sashwinbalaji: phrase it better all this becaosue "Raw STRING and PACKED
// bytes must not be mistaken for proto-group tags by the relay." so phrase it cleartly to distinguish .,.. we do this also maybe add a small e.g. ?

include/perfetto/public/pb_utils.h
```diff
@@ -38,6 +38,25 @@ static inline uint32_t PerfettoPbMakeTag(int32_t field_id,
          PERFETTO_STATIC_CAST(uint32_t, wire_type);
 }
 
+// Append-only nested-message encoding used by tracing v2:
+//
+//   open nested field f:  varint((f << 3) | 3)
+//   close current nested: 0x04
+//   root:                 no wrapper and no close byte
```
sashwinbalaji: why so how we identify them then ?

include/perfetto/public/pb_packed.h
```diff
@@ -26,11 +26,11 @@
...
-// nested messages and allow zero-copy serialization. A protobuf message that
-// has a packed repeated field provides begin and end operations that accept a
-// PerfettoPbPackedMsg. The downside of this approach is that (like all
-// protozero nested messages), it reserves 4 bytes to encode the length, so it
-// might add overhead for lots of small messages.
+// nested messages. A protobuf message that has a packed repeated field provides
+// begin and end operations that accept a PerfettoPbPackedMsg. Normal protobuf
+// encoding writes entries directly to the stream and reserves 4 bytes for the
+// length. Append-only proto-group encoding buffers the entries until the field
+// closes, when its final length is known.
```
sashwinbalaji: is this change really needed, avoiding this would reduce number of files touched....

src/tracing/v2/proto_rewriter.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace perfetto::tracing_v2 {
+
+enum class RewriteResult {
+  kSuccess,
+  // Not a well-formed proto-group packet: a bad tag, value or length, a stray
+  // close marker, or a message left open.
+  kMalformedInput,
+  // The rewritten packet would exceed |max_output_size|, or a nested message
+  // outgrew its four-byte length field. The input itself may be valid.
```
sashwinbalaji: how is this possible "nested message
  // outgrew its four-byte length field."

src/tracing/v2/proto_rewriter.h
```diff
@@ -0,0 +1,69 @@
...
+//
+//   proto-group (hex):  0b 10 07 04
+//                       \_/       \/
+//                    start field 1  end current message
+//
+//   protobuf (hex):     0a 82 80 80 00 10 07
+//                       \_/ \_________/ \___/
+//                    field 1  length 2   contents
+//
+// The root message has no opening or closing marker. Ordinary protobuf fields
```
sashwinbalaji: just stating is not usefule, reason also important "The root message has no opening or closing marker."

same applies to: // The input must be the reader's private copy, not producer-owned shared
// memory. 

same for "which gain four per
// nested message, "

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,200 @@
...
+// then; the rest overflows the shift and a malformed varint reads as a small
+// legal value. Redundant encodings such as 80 00 for zero stay valid.
+const uint8_t* ParseVarIntWithinUint64(const uint8_t* start,
+                                       const uint8_t* end,
+                                       uint64_t* value) {
+  const uint8_t* pos = protozero::proto_utils::ParseVarInt(start, end, value);
+  if (pos == start)
+    return start;
+  if (pos - start == 10 && (start[9] & 0x7f) > 1)
+    return start;
```
sashwinbalaji: this if block no human can read and understand on what's going on

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,200 @@
...
+
+namespace perfetto::tracing_v2 {
+namespace {
+
+// Protobuf field ids are 29 bits.
+constexpr uint64_t kMaxFieldId = (1u << 29) - 1;
+
+// ParseVarInt() accepts any tenth byte, but a 64-bit value has one bit left by
+// then; the rest overflows the shift and a malformed varint reads as a small
+// legal value. Redundant encodings such as 80 00 for zero stay valid.
```
sashwinbalaji: 3 lines is a big func doc but at the same itme I am not able to even parse and understand this 3 lines so need to rewrite from scratch

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,200 @@
...
+}
+
+bool AppendToOutput(const uint8_t* source_begin,
+                    const uint8_t* source_end,
+                    std::vector<uint8_t>* output,
+                    size_t max_output_size) {
+  const size_t size = static_cast<size_t>(source_end - source_begin);
+
+  // Written as two comparisons rather than output->size() + size so the check
+  // cannot itself overflow.
```
sashwinbalaji: not useful comment!

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,200 @@
...
+      size > max_output_size - output->size()) {
+    return false;
+  }
+  output->insert(output->end(), source_begin, source_end);
+  return true;
+}
+
+// Never leave the caller with the prefix written before the error was found.
+RewriteResult Reject(std::vector<uint8_t>* output, RewriteResult result) {
+  PERFETTO_DCHECK(result != RewriteResult::kSuccess);
```
sashwinbalaji: unnecessary + adds unnecessary extra arg of result . it's in anon namespace and called from this file only what benefit it fives???

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,200 @@
...
+  output->clear();
+  return result;
+}
+
+}  // namespace
+
+RewriteResult RewriteProtoGroupToLengthDelimited(const uint8_t* input_begin,
```
sashwinbalaji: this function needs comments over different code blocks. I am going to trust you that this works fine but adding commetns explaining where what's happenign will help a lot!

src/tracing/v2/trace_writer_v2.h
```diff
@@ -0,0 +1,152 @@
...
+}  // namespace protos
+
+namespace tracing_v2 {
+
+class SharedRingBuffer;
+
+// TraceWriter implementation backed by a tracing v2 shared ring buffer.
+// Protozero writes packets in the append-only proto-group encoding; the reader
+// converts them to length-delimited protobuf after copying them out of shared
+// memory.
```
sashwinbalaji: these 3 lines what value they add? does writer need to tell about all this or even know about this ?

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,206 @@
...
+#include "protos/perfetto/trace/trace_packet.pbzero.h"
+
+namespace perfetto::tracing_v2 {
+namespace {
+
+// Start a new chunk if the current one cannot hold protozero's largest
+// scalar-field encoding. This avoids creating a fragment that immediately
+// needs another buffer.
+constexpr uint32_t kMinFragmentPayloadSize =
+    protozero::proto_utils::kMaxSimpleFieldEncodedSize;
```
sashwinbalaji: what does this mean ? why is this needed? why wasting this much size ?

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,206 @@
...
+  // Protozero never receives more than one chunk at a time. Reserving the same
+  // amount here keeps the drop path allocation-free.
+  drop_buffer_.resize(args.ring_buffer->chunk_size());
+  stream_writer_.Reset(
+      {drop_buffer_.data(), drop_buffer_.data() + drop_buffer_.size()});
+}
+
+TraceWriterV2::~TraceWriterV2() {
+  FinishTracePacket();
+  ring_buffer_writer_.FinishCurrentChunk();
```
sashwinbalaji: why finish current one ? is it the existing practice ?

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,206 @@
...
+  stream_writer_.Reset({nullptr, nullptr});
+  delegate_->OnWriterDestroyed(writer_id());
+}
+
+TraceWriter::TracePacketHandle TraceWriterV2::NewTracePacket() {
+  // Starting another packet while its handle is alive would let the old
+  // handle finalize the new packet when it eventually goes out of scope.
+  PERFETTO_CHECK(!packet_open_);
+
+  const SharedRingBufferWriter::FragmentSpan span =
```
sashwinbalaji: why use word "span" fragmentspan?

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,206 @@
...
+  // A completed fragment leaves its chunk cached so that the next packet can
+  // append to it. Give up that cache before the delegate samples write_pos.
+  ring_buffer_writer_.FinishCurrentChunk();
+  stream_writer_.Reset({nullptr, nullptr});
+  delegate_->Flush(writer_id(), std::move(callback));
+}
+
+void TraceWriterV2::OnMessageFinalized(protozero::Message*) {
+  FinishTracePacket();
+}
```
sashwinbalaji: this can be in .h ?

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,206 @@
...
+  ClosePacketFragment(/*continues_on_next=*/true);
+
+  // Notify before reserving the continuation. A packet can fill the entire
+  // ring without finishing; a stalling writer would otherwise wait for space
+  // before the reader had been asked to release any.
+  delegate_->NotifyReader();
+
+  // The reader may have scraped the fragment while it was being closed. If
+  // its unpublished suffix could not be relocated, the packet now has a gap.
+  if (in_drop_mode_)
```
sashwinbalaji: explain clearly this canhappein in ClosePacketFragment..

include/perfetto/public/pb_utils.h
```diff
@@ -38,6 +38,25 @@ static inline uint32_t PerfettoPbMakeTag(int32_t field_id,
...
+//
+// Opens carry the field number. 0x04 is field-zero/end-group, which cannot be
+// an ordinary field, and closes the innermost message. Packet framing bounds
+// the root, so it needs neither marker.
+enum {
+  PERFETTO_PB_WIRE_TYPE_START_GROUP = 3,
+  PERFETTO_PB_PROTO_GROUP_END_BYTE = 0x04,
+};
```
sashwinbalaji: why is this a enum ??? it ain't making sense to me

include/perfetto/public/pb_utils.h
```diff
@@ -38,6 +38,25 @@ static inline uint32_t PerfettoPbMakeTag(int32_t field_id,
...
+
+static inline uint32_t PerfettoPbMakeStartGroupTag(int32_t field_id) {
+  return ((PERFETTO_STATIC_CAST(uint32_t, field_id)) << 3) |
+         PERFETTO_STATIC_CAST(uint32_t, PERFETTO_PB_WIRE_TYPE_START_GROUP);
+}
```
sashwinbalaji: why use PERFETTO_PB_WIRE_TYPE_START_GROUP in one place and 3 in other ???

include/perfetto/protozero/message.h
```diff
@@ -58,16 +75,17 @@ class PERFETTO_EXPORT_COMPONENT Message {
   // static_assert()s in the Reset() method.
   Message() = default;
 
-  // Clears up the state, allowing the message to be reused as a fresh one.
+  // Clears the state using length-delimited encoding.
```
sashwinbalaji: can we try to minimise the changes and avoid updating the existing comment unless absolutely necessary pls!

include/perfetto/protozero/message.h
```diff
@@ -58,16 +75,17 @@ class PERFETTO_EXPORT_COMPONENT Message {
...
-  // size can be represented with fewer than
-  // proto_utils::kMessageLengthFieldSize bytes.
+  // As above, but selects the encoding used for nested messages.
+  void Reset(ScatteredStreamWriter*, MessageArena*, NestedMessageEncoding);
+
+  // Finalizes this message and its children:
+  // - Length-delimited children backfill their length and may compact it.
+  // - Proto-group children append one close byte; the root appends none.
+  // - Repeated calls emit nothing further.
+  // Returns the size including children and close bytes, excluding chunking.
```
sashwinbalaji: can we try to minimise the changes and avoid updating the existing comment unless absolutely necessary pls!

include/perfetto/protozero/message.h
```diff
@@ -79,6 +97,13 @@ class PERFETTO_EXPORT_COMPONENT Message {
   uint8_t* size_field() const { return size_field_; }
   void set_size_field(uint8_t* size_field) { size_field_ = size_field; }
 
+  // Encoding selected by the root and used by all its children.
+  NestedMessageEncoding nested_message_encoding() const {
```
sashwinbalaji: why can't just MessageFraming be used why we need NestedMessageEncoding ????

include/perfetto/protozero/message.h
```diff
@@ -226,7 +254,8 @@ class PERFETTO_EXPORT_COMPONENT Message {
 
   // [optional] Pointer to a non-aligned pre-reserved var-int slot of
   // kMessageLengthFieldSize bytes. When set, the Finalize() method will write
-  // the size of proto-encoded message in the pointed memory region.
+  // the size of proto-encoded message in the pointed memory region. Always null
+  // in kProtoGroup mode, which never reserves a length.
```
sashwinbalaji: yup follow this still of appedning/addition don't rewrite existing vcomment unnecessarly that's just churn!

include/perfetto/protozero/message.h
```diff
@@ -246,6 +275,16 @@ class PERFETTO_EXPORT_COMPONENT Message {
 
   MessageState message_state_;
 
+  // Determines child encoding and how this message ends. A proto-group root
+  // has no opening tag, so only nested proto-group messages append an end byte.
+  enum class MessageFraming : uint8_t {
```
sashwinbalaji: not clear at all why both MessageFraming and NestedMessageEncoding needed why not one is engouh ????

include/perfetto/protozero/proto_utils.h
```diff
@@ -200,6 +217,11 @@ constexpr uint32_t MakeTagLengthDelimited(uint32_t field_id) {
          static_cast<uint32_t>(ProtoWireType::kLengthDelimited);
 }
 
+// Opens a nested message in the proto-group encoding described above.
+constexpr uint32_t MakeTagStartGroup(uint32_t field_id) {
+  return (field_id << 3) | kWireTypeStartGroup;
```
sashwinbalaji: why nbot use kWireTypeStartGroup consistentl ????

include/perfetto/protozero/root_message.h
```diff
@@ -40,8 +40,12 @@ class RootMessage : public T {
   RootMessage& operator=(RootMessage&&) = delete;
 
   void Reset(ScatteredStreamWriter* writer) {
+    Reset(writer, NestedMessageEncoding::kLengthDelimited);
```
sashwinbalaji: why messagegramig can't be used here ???

include/perfetto/public/abi/data_source_abi.h
```diff
@@ -359,9 +359,30 @@ PERFETTO_SDK_EXPORT void PerfettoDsImplTraceIterateBreak(
 // Creates a new trace packet on `tracer`. Returns a stream writer that can be
 // used to write data to the packet. The caller must use
 // PerfettoDsTracerImplPacketEnd() when done.
+//
+// This entry point is kept for ABI compatibility and only supports tracing v1
+// writers. New callers should use
+// PerfettoDsTracerImplPacketBeginWithEncoding().
```
sashwinbalaji: this comment can be just "New callers should use
// PerfettoDsTracerImplPacketBeginWithEncoding()." I dont think any need fior "// This entry point is kept for ABI compatibility and only supports tracing v1
// writers"

include/perfetto/public/data_source.h
```diff
@@ -277,13 +277,38 @@ struct PerfettoDsRootTracePacket {
...
 };
 
+static inline enum PerfettoPbMsgEncoding PerfettoDsPacketEncodingToPbMsg(
+    uint32_t encoding) {
+  switch (encoding) {
+    case PERFETTO_DS_PACKET_ENCODING_LENGTH_DELIMITED:
+      return PERFETTO_PB_MSG_ENCODING_LENGTH_DELIMITED;
+    case PERFETTO_DS_PACKET_ENCODING_PROTO_GROUP:
+      return PERFETTO_PB_MSG_ENCODING_PROTO_GROUP;
+  }
```
sashwinbalaji: what the hell is this switch ?

include/perfetto/public/pb_macros.h
```diff
@@ -184,7 +184,7 @@
       struct PROTO * msg, struct PerfettoPbPackedMsg##C_TYPE * nested) {  \
     struct PerfettoPbMsg* nested_msg =                                    \
         PERFETTO_REINTERPRET_CAST(struct PerfettoPbMsg*, nested);         \
-    PerfettoPbMsgBeginNested(&msg->msg, nested_msg, NUM);                 \
+    PerfettoPbMsgBeginLengthDelimitedField(&msg->msg, nested_msg, NUM);   \
```
sashwinbalaji: why these changes ???

include/perfetto/public/pb_msg.h
```diff
@@ -31,6 +33,36 @@
 // proto_utils.h.
 #define PROTOZERO_MESSAGE_LENGTH_FIELD_SIZE 4
 
+// Encoding used for nested messages. The root chooses it when it is initialized
+// and all its children use the same encoding.
+enum PerfettoPbMsgEncoding {
+  // Normal protobuf. The nested-message length is filled in on finalization.
+  PERFETTO_PB_MSG_ENCODING_LENGTH_DELIMITED = 0,
```
sashwinbalaji: why so many duplciate enum defintions I am seeing ???

src/tracing/v2/trace_writer_v2.h
```diff
@@ -0,0 +1,152 @@
...
+    virtual void Flush(WriterID, std::function<void()> callback) = 0;
+
+    // Called after the writer publishes its last chunk. The WriterID must not
+    // be reused while unresolved ring positions can still name it.
+    virtual void OnWriterDestroyed(WriterID) = 0;
+  };
+
+  struct InitArgs {
+    // Must be non-null. Retained until the writer is destroyed.
+    std::shared_ptr<Delegate> delegate;
```
sashwinbalaji: should these be name writer_delegate or not needed

src/shared_lib/data_source.cc
```diff
@@ -575,17 +575,38 @@ void PerfettoDsImplTraceIterateBreak(
...
+
+struct PerfettoStreamWriter PerfettoDsTracerImplPacketBegin(
+    struct PerfettoDsTracerImpl* tracer) {
+  const struct PerfettoDsPacketBeginResult result =
+      PerfettoDsTracerImplPacketBeginWithEncoding(tracer);
+  // The legacy entry point cannot tell its caller to use proto-group encoding.
+  PERFETTO_CHECK(result.encoding ==
+                 PERFETTO_DS_PACKET_ENCODING_LENGTH_DELIMITED);
+  return result.writer;
 }
```
sashwinbalaji: shouldn't this be placed before PerfettoDsTracerImplPacketBegin PerfettoDsTracerImplPacketBeginWithEncoding ?

src/protozero/message.cc
```diff
@@ -45,8 +45,17 @@ std::atomic<uint32_t> g_generation;
 // Do NOT put any code in the constructor or use default initialization.
 // Use the Reset() method below instead.
 
-// This method is called to initialize both root and nested messages.
+// Keep the original exported overload for callers built before
+// NestedMessageEncoding was added.
```
sashwinbalaji: not needed

src/protozero/message.cc
```diff
@@ -55,12 +64,21 @@ void Message::Reset(ScatteredStreamWriter* stream_writer, MessageArena* arena) {
...
   stream_writer_ = stream_writer;
   arena_ = arena;
   size_ = 0;
   size_field_ = nullptr;
   nested_message_ = nullptr;
   message_state_ = MessageState::kNotFinalized;
+  message_framing_ =
+      nested_message_encoding == NestedMessageEncoding::kProtoGroup
+          ? MessageFraming::kProtoGroupRoot
+          : MessageFraming::kLengthDelimited;
```
sashwinbalaji: again why the dance of nesting and message ?

src/protozero/message.cc
```diff
@@ -125,11 +143,18 @@ uint32_t Message::Finalize() {
...
     EndNestedMessage();
 
-  // Write the length of the nested message a posteriori, using a leading-zero
-  // redundant varint encoding. This can be nullptr for the root message, among
-  // many reasons, because the TraceWriterImpl delegate is keeping track of the
-  // root fragment size independently.
-  if (size_field_) {
+  if (PERFETTO_UNLIKELY(message_framing_ != MessageFraming::kLengthDelimited)) {
+    PERFETTO_DCHECK(!size_field_);
+    if (message_framing_ == MessageFraming::kProtoGroupNested) {
```
sashwinbalaji: add a comment over this if explaing what we doing

src/protozero/message.cc
```diff
@@ -188,20 +213,28 @@ Message* Message::BeginNestedMessageInternal(uint32_t field_id) {
   if (nested_message_)
     EndNestedMessage();
 
-  // Write the proto preamble for the nested message.
```
sashwinbalaji: keep this comment ?

include/perfetto/public/pb_msg.h
```diff
@@ -72,7 +140,9 @@ static inline void PerfettoPbMsgPatch(struct PerfettoPbMsg* msg) {
 static inline void PerfettoPbMsgPatchStack(struct PerfettoPbMsg* msg) {
   uint8_t* const cur_range_end = msg->writer->writer.end;
   uint8_t* const cur_range_begin = msg->writer->writer.begin;
-  while (msg && cur_range_begin <= msg->size_field &&
+  // A proto-group message has no length field to patch, so the walk stops at
+  // the first message without one.
```
sashwinbalaji: so won't this be cleaner to guard on basis of enum ??

include/perfetto/public/pb_msg.h
```diff
@@ -203,24 +273,44 @@ static inline void PerfettoPbMsgAppendCStrField(struct PerfettoPbMsg* msg,
...
-      PerfettoStreamWriterReserveBytes(&nested->writer->writer,
-                                       PROTOZERO_MESSAGE_LENGTH_FIELD_SIZE));
   nested->parent = parent;
-  parent->size += PROTOZERO_MESSAGE_LENGTH_FIELD_SIZE;
   parent->nested = nested;
 }
 
+// Starts a STRING or PACKED field whose value is written in several calls.
+// - Length-delimited: reserve the length and backfill it when the field closes.
+// - Proto-group: abort before the tag. Published fragments cannot be patched.
```
sashwinbalaji: you need to add ref to PerfettoPbMsgAppendBytes or somethign to make this clearer

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,199 @@
...
+                          args.target_buffer,
+                          args.buffer_exhausted_policy,
+                          delegate_.get()),
+      stream_writer_(this),
+      cur_packet_(new protozero::RootMessage<protos::pbzero::TracePacket>()) {
+  // Protozero never receives more than one chunk at a time. Reserving the same
+  // amount here keeps the drop path allocation-free.
+  drop_buffer_.resize(delegate_->ring_buffer().chunk_size());
+  stream_writer_.Reset(
+      {drop_buffer_.data(), drop_buffer_.data() + drop_buffer_.size()});
```
sashwinbalaji: we can do the smae inline commetn thingh like {/*.. */ .. , .....}

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,193 @@
...
+#include <vector>
+
+#include "perfetto/base/logging.h"
+#include "perfetto/protozero/proto_utils.h"
+
+namespace perfetto::tracing_v2 {
+namespace {
+
+// Protobuf field ids are 29 bits.
+constexpr uint64_t kMaxFieldId = (1u << 29) - 1;
```
sashwinbalaji: add reason for this pls...

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,193 @@
...
+
+constexpr size_t kMaxVarIntBytes = 10;
```
sashwinbalaji: add reason coment for this also pls

src/tracing/v2/proto_rewriter.cc
```diff
@@ -0,0 +1,193 @@
...
+// continuing to accept redundant encodings, as protobuf does.
+const uint8_t* ParseVarIntWithinUint64(const uint8_t* start,
+                                       const uint8_t* end,
+                                       uint64_t* value) {
+  const uint8_t* pos = protozero::proto_utils::ParseVarInt(start, end, value);
+  if (pos == start)
+    return start;
+  const size_t num_bytes = static_cast<size_t>(pos - start);
+  // The first nine bytes hold 63 bits, so byte ten may only contain 0 or 1.
+  if (num_bytes == kMaxVarIntBytes && start[kMaxVarIntBytes - 1] > 1)
```
sashwinbalaji: sjhould thois be num_bytes >= ... if not comment should ssay why nto >

include/perfetto/protozero/message.h
```diff
@@ -227,6 +259,10 @@ class PERFETTO_EXPORT_COMPONENT Message {
   // [optional] Pointer to a non-aligned pre-reserved var-int slot of
   // kMessageLengthFieldSize bytes. When set, the Finalize() method will write
   // the size of proto-encoded message in the pointed memory region.
+  //
+  // In proto-group mode, a nested message ends with a closing byte and the
+  // root ends at the packet boundary. Neither writes a length before its
+  // contents, so there is no length field to fill in later.
```
primiano, line 265: I'm assuming that in group mode the size_field is always null?
In this case make the comemnt stronger. Say "this is unused and always null in group mode".
Or maybe it's set but ignored (which is also fine). but then say which one is which.

EDIT after reading the message.cc looks like it's always null.

include/perfetto/protozero/message.h
```diff
@@ -246,6 +282,33 @@ class PERFETTO_EXPORT_COMPONENT Message {
...
+  //        |                                        |
+  //        v                                        v
+  //        +-------------+----------------+---------+
+  //        | START child | child's fields | END     |
+  //        +-------------+----------------+---------+
+  //
+  // END closes the child. No additional byte closes the root.
+  // Both messages use the same encoding, so we need is_root_ to tell them
+  // apart.
+  bool is_root_;
```
primiano, line 310: I'm going to be a bit handwavy here, but I wonder if we can leverage the fact that we have already the notion of a RootMessage (for arena purposes).
TraceWritrer (and also your V2) has a reference to a root message.
So I am thinking something on the lines of:
- We make Message::Finalize take an optional argument, so Finalize(bool is_root=false)
- We override (without making it virtual) the Finalize method in RootMessage (keeping it a void)
so something like:

```
class RootMessage : public T {
  ...
  Finalize() {
    Finalize(/*is_root=*/true);
  }
}
```

I think this shoudl have the same effect without adding further state tracking.
But I have not tought deeply to this, so don't take it as granted.
I wouldn't be surprised if you found some good reason why we have to do your way.

src/protozero/message.cc
```diff
@@ -188,20 +223,30 @@ Message* Message::BeginNestedMessageInternal(uint32_t field_id) {
...
+  uint32_t tag;
+  switch (encoding_) {
```
primiano, line 227: Imho the switch(encoding_) is 1. overkill, 2. inconsistent with what you did above (where you had an if(UNLIKELY)). Pick one and be consistent.
I'd honestly just go with a plain if.

siwtch adds more nesting, and sometimes causes problems with older compilers about "what if it's option 3".

include/perfetto/protozero/proto_utils.h
```diff
@@ -48,6 +48,23 @@ enum class ProtoWireType : uint32_t {
   kFixed32 = 5,
 };
 
+// Wire types 3 and 4 delimit a protobuf group. Protozero decoders do not
+// support groups, so keep them out of ProtoWireType. Tracing v2 only borrows
+// the start-group wire type for the encoding below.
+constexpr uint32_t kWireTypeStartGroup = 3;
```
primiano, line 54: When commenting on kWireTypeStartGroup explain that we bastardized group (okay not in those terms) and don't follow the official encoding for it (you can link at the RFC)

src/tracing/v2/trace_writer_v2.h
```diff
@@ -0,0 +1,156 @@
...
+// before nested-message lengths are known.
+//
+// A packet normally occupies one fragment. If it crosses a chunk boundary, the
+// writer sets the continuation flags and carries on in the next chunk. The
+// reader joins those fragments before rewriting the packet.
+//
+// As with TraceWriterImpl, an instance may be used by only one thread at a
+// time. Packet writes go straight to the ring; the delegate handles reader
+// notifications, flush completion and WriterID retirement.
+class TraceWriterV2 : public TraceWriter,
```
primiano, line 58: Maybe purely for naming consistency (when comparing and doing codesearch) this should be called TraceWriterV2Impl, as thi feeels the moral equivalent of TraceWriterImpl, right?

src/tracing/v2/trace_writer_v2.h
```diff
@@ -0,0 +1,156 @@
...
+  // Provides storage and coordinates reader progress for the writer's lifetime.
+  // Calls run on SDK writer threads and may overlap across writers.
+  // Post reader work to a shared sequence if it needs to run serially.
+  class Delegate : public SharedRingBufferWriter::Delegate {
+   public:
+    ~Delegate() override;
+
+    // Returns the same ring on every call. The ring and its backing memory must
+    // remain valid for the delegate's lifetime.
+    virtual SharedRingBuffer& ring_buffer() = 0;
```
primiano, line 71: Why the ring_buffer is not just a pointer passed in InitArgs? feels a bit weird to have a delefate method just to return the same thing over and over.
Or maybe there is an architectural reason why doing that has less plumbing?
I'm not too obsessed by this as you ude it only in the ctor anyways, so follow my suggestion only if it doesn't cause more plumbing. if not stick with yours

src/tracing/v2/trace_writer_v2.h
```diff
@@ -0,0 +1,156 @@
...
+    virtual void Flush(WriterID, std::function<void()> callback) = 0;
+
+    // Called after the writer publishes its last chunk. The WriterID must not
+    // be reused while unconsumed ring positions can still name it.
+    virtual void OnWriterDestroyed(WriterID) = 0;
+  };
+
+  struct InitArgs {
+    // Must be non-null. Retained until the writer is destroyed.
+    std::shared_ptr<Delegate> delegate;
```
primiano, line 86: out of curiosity why the Deletgate is a shared_ptr and not a unique_ptr<Delegate>?
A TraceWriter is not thread-safe (it's supposed to be used on the same thread, or caller must use a lock), so I'm not sure on why you need the refcounting here?

src/tracing/v2/trace_writer_v2.h
```diff
@@ -0,0 +1,156 @@
...
+
+    // Called after this writer publishes its data. For a non-empty callback:
+    // - Wait for service acknowledgement before invoking it.
+    // - The delegate chooses the callback sequence.
+    // - Disconnect may discard it, as allowed by TraceWriter::Flush().
+    virtual void Flush(WriterID, std::function<void()> callback) = 0;
+
+    // Called after the writer publishes its last chunk. The WriterID must not
+    // be reused while unconsumed ring positions can still name it.
+    virtual void OnWriterDestroyed(WriterID) = 0;
```
primiano, line 81: uh why the delegate has this notirication? isn't the owner of the delegate supposed to know when the writer is destoryed?

I think I don't fully understand how ownership/lifecycle works for TWv2

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,202 @@
...
+      ring_buffer_writer_(&delegate_->ring_buffer(),
+                          args.writer_id,
+                          args.target_buffer,
+                          args.buffer_exhausted_policy,
+                          delegate_.get()),
+      stream_writer_(this),
+      cur_packet_(new protozero::RootMessage<protos::pbzero::TracePacket>()) {
+  // Protozero never receives more than one chunk at a time. Reserving the same
+  // amount here keeps the drop path allocation-free.
+  drop_buffer_.resize(delegate_->ring_buffer().chunk_size());
```
primiano, line 55: can we lazily allocate this if we ever need it, rather than pre-serving one chunk per thread? feels too aggressive memory-wise.
Or actually can the drop buffer be one for all (i.e. a global in the anonymous name space here) rather than one per writer?
is it like the garbage chunk? or do we care about ownerhsip of the writes ? is this going to do CAS & co on this drop_buffer?

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,202 @@
...
+  stream_writer_.Reset(
+      {drop_buffer_.data(), drop_buffer_.data() + drop_buffer_.size()});
+}
+
+TraceWriterV2::~TraceWriterV2() {
+  FinishTracePacket();
+  // Publish the last chunk before notifying the delegate of destruction.
+  ring_buffer_writer_.FinishCurrentChunk();
+  stream_writer_.Reset({nullptr, nullptr});
+  delegate_->OnWriterDestroyed(writer_id());
```
primiano, line 65: isn't the delegate the same obhect that owns TraceWriterV2 (which means it could tell precisely when it's going to be destroyed?)

src/tracing/v2/trace_writer_v2.cc
```diff
@@ -0,0 +1,202 @@
...
+      kMinFragmentPayloadSize, /*continues_from_prev=*/false);
+  if (range.result == SharedRingBufferWriter::BeginFragmentResult::kSuccess) {
+    fragment_begin_ = range.begin;
+    stream_writer_.Reset({range.begin, range.end});
+    in_drop_mode_ = false;
+  } else {
+    stream_writer_.Reset(EnterDropMode());
+  }
+
+  // Every nested message inherits the root's encoding.
```
primiano, line 85: i'm not sure I understood this comment

primiano (commented):
I have some limited time today but i'm sending some intiial comments.
I only looked at message.h/cc, trace_writer_v2.h/cc, which I believe are the more intetesting ones and worth getting comments sooner.

I didn't look at the rewriter, but givenm teh signature I suppose it's non controversial, to the point where I'd blindly trust a decent AI Agent to deal with that.

I dind't look at the shared_lib stuff because i'm not fmailiar with it and woudl take me too time today to look. but again I don't think there is a lot of juicy stuff there

sashwinbalaji:
>I dind't look at the shared_lib stuff because i'm not fmailiar with it and woudl take me too time today to look. but again I don't think there is a lot of juicy stuff there

There is a slilghtly controversial thing in include/perfetto/public/pb_msg.h , I would like your opinion on. About how I am (ab)using PerfettoPbMsg::size field

PR #7326, opened 2026-09-03
tracing: route in-process SDK writers through the v2 ring

sashwinbalaji:
Add a temporary producer-side adapter so the v2 writer can run in complete
tracing sessions while traced still reads the v1 shared-memory format:

    SDK -> TraceWriterV2 -> v2 ring -> reader/rewriter
                                      -> v1 writer -> v1 SMB -> traced

- Select v2 per data-source instance. Keep v1 as the default and fail loudly
  for unsupported startup tracing, interceptors and platforms.

- Size the ring from the producer's shared-memory configuration and make the
  chunk size configurable through TraceConfig.

- Drain, reassemble and rewrite v2 packets on a relay sequence. Forward each
  WriterID through its own v1 writer.

- Keep the bridge and endpoint from the same connection together so a
  reconnect cannot mix their writers or WriterIDs.

- Drain earlier ring data before completing flush, stop and writer retirement.
  Process large drains in bounded batches so other relay work can run.

- Test v1/v2 selection, packet forwarding, data loss, no_flush, exhaustion,
  reconnect, shutdown and control ordering.


src/shared_lib/data_source.cc
```diff
@@ -602,11 +602,9 @@ struct PerfettoStreamWriter PerfettoDsTracerImplPacketBegin(
   struct PerfettoStreamWriter ret;
   const enum PerfettoDsPacketEncoding encoding =
       PerfettoDsTracerImplPacketBeginWithEncoding(tracer, &ret);
-  // Old C SDK callers always write length-delimited messages. Writer setup must
-  // keep them on tracing v1; this CHECK prevents mixed framing if that contract
-  // is ever broken.
-  // TODO(sashwinbalaji): advertise proto-group support before enabling tracing
-  // v2 for independently built C SDK clients.
+  // This entry point cannot return the writer's encoding, so it only works
```
sashwinbalaji: should be in scaffolding cl

include/perfetto/public/abi/producer_abi.h
sashwinbalaji: in commit it should v2 smb not v2 ring to make it easier to read:

SDK -> TraceWriterV2 -> v2 ring -> relay reader/rewriter
                                      -> v1 writer -> v1 SMB -> traced

include/perfetto/tracing/tracing.h
```diff
@@ -100,6 +101,13 @@ struct TracingInitArgs {
   // larger than kMaxShmSize (32MB) or not a multiple of 4KB.
   uint32_t shmem_size_hint_kb = 0;
 
+  // [Optional] Experimental, default off. Routes every TraceWriter of the
+  // producer backends initialized here through the tracing v2 ring; startup
+  // tracing writers stay on v1. Every data source must then use the current
+  // C++ or C SDK; older C headers and the Rust SDK are not supported. Wider
+  // rollout needs per-data-source capability negotiation.
+  bool enable_tracing_v2 = false;
```
sashwinbalaji: maybe call this experimental_enable_tracing_v2 or someting to be clearer (need to update everywhere)

include/perfetto/tracing/tracing.h
```diff
@@ -200,6 +208,8 @@ struct TracingInitArgs {
   BackendFactoryFunction in_process_backend_factory_ = nullptr;
   ProducerBackendFactoryFunction system_producer_backend_factory_ = nullptr;
   ConsumerBackendFactoryFunction system_consumer_backend_factory_ = nullptr;
+  const internal::TracingV2EndpointFunctions* tracing_v2_endpoint_functions_ =
```
sashwinbalaji: why only this from internal:: ? and also why only this const ?

include/perfetto/tracing/internal/tracing_v2_endpoint_functions.h
```diff
@@ -0,0 +1,58 @@
...
+// Keep this table limited to operations that cannot go through
+// ProducerEndpoint itself.
+struct TracingV2EndpointFunctions {
+  bool (*is_supported)() = nullptr;
+  std::unique_ptr<ProducerEndpoint> (*create_endpoint)(
+      std::unique_ptr<ProducerEndpoint>,
+      base::TaskRunner* muxer_task_runner,
+      base::TaskRunner* relay_task_runner) = nullptr;
+  void (*start_releasing_writers)(ProducerEndpoint*) = nullptr;
+  bool (*retains_writers)(const ProducerEndpoint*) = nullptr;
```
sashwinbalaji: this func name(haven't read the .c file yet) is not conveying to me what it does and why it exisits

src/shared_lib/producer.cc
```diff
@@ -59,6 +60,12 @@ void PerfettoProducerBackendInitArgsSetMachineId(
   backend_args->machine_id = machine_id;
 }
 
+void PerfettoProducerBackendInitArgsSetTracingV2Enabled(
+    struct PerfettoProducerBackendInitArgs* backend_args,
+    bool enabled) {
```
sashwinbalaji: do we need this bool if the func is called ...enabled( why not just set it to true directlyt ?

src/tracing/internal/tracing_muxer_impl.cc
sashwinbalaji: CI failures:

- https://github.com/google/perfetto/actions/runs/33741404721/job/100604013235
- https://github.com/google/perfetto/actions/runs/33741404721/job/100604013134

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -275,6 +284,10 @@ class TracingMuxerImpl : public TracingMuxer {
 
     std::map<FlushRequestID, std::set<DataSourceInstanceID>> pending_flushes_;
 
+    // Non-owning pointer to the v2 decorator held by |service_|. Null when the
```
sashwinbalaji: why use term like "decorator"

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -501,17 +514,21 @@ class TracingMuxerImpl : public TracingMuxer {
     FindDataSourceRes(DataSourceStaticState* a,
                       DataSourceState* b,
                       uint32_t c,
-                      bool d)
+                      bool d,
+                      bool e)
```
sashwinbalaji: curious why variables are name like this here ?

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -501,17 +514,21 @@ class TracingMuxerImpl : public TracingMuxer {
...
+          requires_callbacks_under_lock(d),
+          no_flush(e) {}
     explicit operator bool() const { return !!internal_state; }
 
     DataSourceStaticState* static_state = nullptr;
     DataSourceState* internal_state = nullptr;
     uint32_t instance_idx = 0;
     bool requires_callbacks_under_lock = false;
+    // The registered descriptor's no_flush, whatever the service was told.
+    bool no_flush = false;
```
sashwinbalaji: seems like new feature support to me. somethign not even v1 supported ?

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -544,6 +561,10 @@ class TracingMuxerImpl : public TracingMuxer {
 
   // WARNING: If you add new state here, be sure to update ResetForTesting.
   std::unique_ptr<base::TaskRunner> task_runner_;
+  // Shared by all temporary v2 relays and created on first use. Declared before
```
sashwinbalaji: "v2 relays" is difficult to understand, make it clearer.

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -275,6 +284,10 @@ class TracingMuxerImpl : public TracingMuxer {
 
     std::map<FlushRequestID, std::set<DataSourceInstanceID>> pending_flushes_;
 
+    // Non-owning pointer to the v2 decorator held by |service_|. Null when the
+    // current endpoint uses v1. Accessed only on the muxer sequence.
+    ProducerEndpoint* tracing_v2_endpoint_ = nullptr;
```
sashwinbalaji: i see no v1_endpoint_ and no v1_endpoint_functions_ so why only v2 needs them ? what's going on ?

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -204,6 +207,29 @@ void TracingMuxerImpl::ProducerImpl::Initialize(
     std::unique_ptr<ProducerEndpoint> endpoint) {
   PERFETTO_DCHECK_THREAD(thread_checker_);
   PERFETTO_DCHECK(!connected_);
+  tracing_v2_endpoint_ = nullptr;
+  // The temporary consumer runs on a thread in this process, regardless of the
+  // selected backend. Startup writers bypass
+  // ProducerEndpoint::CreateTraceWriter() and remain on v1.
```
sashwinbalaji: I dont understand this comment. it's combining two things with no break in between and is vague and not clear.

src/tracing/internal/tracing_muxer_impl.cc
sashwinbalaji: just curious who invokes this ? and who crearted producerimpl passing the needed functions ?

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -204,6 +207,29 @@ void TracingMuxerImpl::ProducerImpl::Initialize(
...
+          /*name_for_debugging=*/"TracingV2Relay"};
+      muxer_->tracing_v2_relay_task_runner_ =
+          muxer_->platform_->CreateTaskRunner(std::move(args));
+    }
+    std::unique_ptr<ProducerEndpoint> v2_endpoint =
+        tracing_v2_endpoint_functions_->create_endpoint(
+            std::move(endpoint), muxer_->task_runner_.get(),
+            muxer_->tracing_v2_relay_task_runner_.get());
+    tracing_v2_endpoint_ = v2_endpoint.get();
+    endpoint = std::move(v2_endpoint);
```
sashwinbalaji: why not write this as below why bring in v2_endpoint ? 

endpoint =
        tracing_v2_endpoint_functions_->create_endpoint(
            std::move(endpoint), muxer_->task_runner_.get(),
            muxer_->tracing_v2_relay_task_runner_.get());
    tracing_v2_endpoint_ = endpoint.get();

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+//
+// SDK threads write into a producer-local ring. A separate relay sequence reads
+// the ring, joins packet fragments, converts proto-group fields to ordinary
+// protobuf and writes the result through v1. One v1 writer per v2 WriterID
+// preserves packet-sequence identity while the endpoint is connected.
+//
+// Step 1 uses this extra hop to exercise the v2 writer without changing the
+// service:
+//
+//   SDK -> TraceWriterV2 -> v2 ring
```
sashwinbalaji: use v2 smb?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+//                                |
+//                +---------------v------------------+
+//                | temporary bridge on the relay    |
+//                | reader -> rewriter -> v1 writer  |
+//                +---------------+------------------+
+//                                |
+//                                v
+//                             v1 SMB -> traced
+//
+// Once traced reads the v2 ring, the bridge, relay and v1 writer disappear.
```
sashwinbalaji: maybe explicitly say this file and whatver other file is going to disappear.

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+// The reader and rewriter move to the service:
+//
+//   SDK -> TraceWriterV2 -> v2 ring -> traced
+//                                      reader -> rewriter -> trace buffer
+//
+// The bridge also owns the mapping behind the ring for now. SharedRingBuffer is
+// only a view over it, as SharedMemoryABI is in v1. In the cross-process design
+// the producer-side shared-memory setup owns the mapping, and the bridge goes
+// away with the rest of this hop.
```
sashwinbalaji: of this hop ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+//
+// One relay sequence serves all bridges owned by the muxer. A v1 writer using
```
sashwinbalaji: make it clear what this means from end view like each v1 writer shared one or somethign (similar to how reader) and draw some fan in , fan out diagram

and they ksall can delay other writer ring drainss would be good.

Also split into different sentence why this coupling is fine for prototype.

And another new sentence for not waiting with reasn pls

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+// kStall can therefore delay other ring drains. Using kDrop for that temporary
+// hop would break the writer's requested lossless policy, so the initial
+// prototype accepts the coupling. The muxer must never wait for the relay from
+// its own sequence. |relay_task_runner| is owned by the muxer and outlives all
+// bridges that use it.
+class InProcessTracingV2Bridge
+    : public TraceWriterV2::Delegate,
+      public SharedRingBufferReader::Delegate,
```
sashwinbalaji: also mention in future who will own these delegates

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+      public std::enable_shared_from_this<InProcessTracingV2Bridge> {
+ public:
+  // 1024 x 256-byte chunks provide 256 KiB of payload capacity. The mapping is
+  // slightly larger because it also contains the ring header and page padding.
+  static constexpr uint32_t kDefaultNumChunks = 1024;
+  static constexpr uint32_t kDefaultChunkSize = 256;
```
sashwinbalaji: why is this hardcoded why not take from trace config ? how is it done in v1 ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  ~InProcessTracingV2Bridge() override;
+
+  InProcessTracingV2Bridge(const InProcessTracingV2Bridge&) = delete;
+  InProcessTracingV2Bridge& operator=(const InProcessTracingV2Bridge&) = delete;
+  InProcessTracingV2Bridge(InProcessTracingV2Bridge&&) = delete;
+  InProcessTracingV2Bridge& operator=(InProcessTracingV2Bridge&&) = delete;
+
+  // Thread-safe. Reuses |v1_writer|'s WriterID for the v2 writer and retains
+  // |v1_writer| as the relay destination. Returns it unchanged if it
+  // cannot be represented by the v2 ABI.
```
sashwinbalaji: I cannot understand what unchaged means for a func returing uique_ptr: "Returns it unchanged if it
  // cannot be represented by the v2 ABI."

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  std::unique_ptr<TraceWriter> CreateTraceWriter(
+      std::unique_ptr<TraceWriter> v1_writer,
+      BufferID target_buffer,
+      BufferExhaustedPolicy);
+
+  // Stops accepting new writers and queues the release of all retained v1
+  // writers on the relay sequence. Returns without waiting. The caller must
+  // keep their SharedMemoryArbiter alive until holds_v1_writers() is
+  // false.
+  void StartReleasingV1Writers();
```
sashwinbalaji: why this func needed is not clear

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+
+  // Thread-safe. Once false after release has started, remains false.
+  bool holds_v1_writers() const;
```
sashwinbalaji: why this in different case than other funcoins and why placed here ? and what is it doing

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+
+  // Thread-safe. |completion| runs on the relay sequence once every position
```
sashwinbalaji: should bridge know about relay muxer and all ? and can we use some common terminlogy (need to think of it) to make it clear what "relay seqece" means and same for muxer

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  InProcessTracingV2Bridge(base::PagedMemory ring_memory,
+                           uint32_t chunk_size,
+                           base::TaskRunner* relay_task_runner);
+
+  struct WriterState {
+    std::unique_ptr<TraceWriter> v1_writer;
+    BufferID target_buffer = 0;
+
+    // Reassembly state. Relay sequence only.
+    std::vector<uint8_t> pending_packet;
```
sashwinbalaji: why this needed? how pending possible ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+    // the flag has to ride the next packet. If there is no next packet, account
+    // for the loss separately in the service stats.
+    // TODO(sashwinbalaji): add that service-side accounting.
+    uint32_t pending_data_loss = 0;
+  };
+
+  // One control-plane barrier: drain the ring to |target_read_pos|, flush the
+  // v1 writers it names, wait for their acknowledgements, run |completion|.
+  // Barriers run one at a time in enqueue order. Relay sequence only.
+  struct Barrier {
```
sashwinbalaji: Barrier (I know) but in comments not clear exactly why this is needed. what purpose it serves

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+    uint64_t id = 0;
```
sashwinbalaji: explain why giving this a id

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+    // write_pos when the barrier was requested.
+    uint32_t target_read_pos = 0;
```
sashwinbalaji: target_read_pos but comment says write_pos ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  // TraceWriterV2::Delegate:
+  void NotifyReader() override;
+  void Flush(WriterID, std::function<void()> callback) override;
+  void OnWriterDestroyed(WriterID) override;
+
+  // SharedRingBufferReader::Delegate:
+  void OnChunkRead(const SharedRingBufferReader::ChunkContents&) override;
+  void OnDataLoss(WriterID) override;
+
+  // Ring draining. Relay sequence unless noted otherwise.
```
sashwinbalaji: where is this the case "Relay sequence unless noted otherwise." and why so ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  void ScheduleDrain();
+  void DrainOnRelaySequence();
+  // Drains up to |target_read_pos|. False means the reader asked to retry
+  // later.
```
sashwinbalaji: say returns true if drains succedd and false if drain failed and need to retr

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  // Thread-safe. Samples write_pos, then queues the barrier on the relay.
+  void EnqueueBarrier(WriterID, std::function<void()> completion);
+  // Runs barriers from the front until one has to wait.
+  void RunFrontBarrier();
+  void OnFrontBarrierFlushAcknowledged(uint64_t barrier_id);
+  void CompleteFrontBarrier();
+
+  // Writer release. Relay sequence only.
+  void ReleaseV1Writer(WriterID);
+  void ReleaseV1WritersOnRelaySequence();
```
sashwinbalaji: why 2 ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  // Downstream v1 flushing. Relay sequence only.
+  void FlushV1Writer(WriterID, std::function<void()> callback);
+
+  // Packet forwarding. Relay sequence only.
+  void ForwardPacket(WriterID, WriterState*);
+
+  bool HasPendingRingData() const;
+  // Keeps the last reference in the queued task, so discarding the task during
+  // shutdown destroys the bridge instead of leaking it.
+  static void DeleteWhenQuiescent(
```
sashwinbalaji: why this static ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+      std::shared_ptr<InProcessTracingV2Bridge> owner);
+
+  // The mapping, the view over it, then the reader on the view. Destruction
+  // runs in reverse, so the mapping goes last. A TraceWriterV2 keeps this
+  // bridge alive through its delegate, so its pointer to |ring_buffer_| cannot
+  // outlive the mapping either.
+  base::PagedMemory ring_memory_;
```
sashwinbalaji: would be great to say somehre here (maybe above only with other ocmment) where all this will move in future

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  SharedRingBuffer ring_buffer_;
+  SharedRingBufferReader ring_buffer_reader_;
+
+  // Reused across packets.
+  std::vector<uint8_t> rewritten_packet_;
+
+  // Writers with data in an open v1 chunk. Relay sequence only.
+  std::vector<WriterID> dirty_writers_;
```
sashwinbalaji: ongoing_writers_ ?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+
+  // Only the front barrier is ever in progress. Relay sequence only.
+  std::deque<Barrier> barriers_;
```
sashwinbalaji: barriers are not at all explained and need to named bertter like flush_/ack_ barrier?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,264 @@
...
+  uint64_t next_barrier_id_ = 1;
+
+  // CreateTraceWriter() inserts from SDK threads. WriterState contents are
+  // otherwise confined to the relay sequence.
```
sashwinbalaji: this comment directly doesntc convey why the mutex and what it protects

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+// sequence. A pass that stops with work left reschedules itself, so this only
+// bounds how long a single task runs.
+constexpr uint32_t kMaxPositionsPerPass = 256;
+
+// Limit how much memory the relay uses for one packet. Rewriting can add four
+// bytes per nested message, so it can still reject a reassembled packet near
+// this limit.
+// TODO(sashwinbalaji): replace with the service's aggregate per-producer budget
+// when the consumer moves into traced; a per-packet cap alone does not bound
+// what a producer can make the reader hold across many writers.
```
sashwinbalaji: this commet todo is difficult for me to understand, need ot be humanized + made clearer

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+constexpr size_t kMaxPacketSize = protozero::proto_utils::kMaxMessageLength;
+
+// Bytes to map for a ring of |num_chunks| chunks of |chunk_size| bytes, or
+// nullopt if SharedRingBuffer would reject that layout or the mapping cannot
+// be sized. PagedMemory rounds the request up to a page and adds two guard
+// pages with unchecked size_t arithmetic, so leave room for both.
```
sashwinbalaji: this two lines tooo AI'e and can be made clearer.

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+std::optional<size_t> RingMappingSize(uint32_t num_chunks,
```
sashwinbalaji: func can be named better RingPageMemorySize or better ?

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+                                      uint32_t chunk_size) {
+  if (!base::IsPowerOfTwo(num_chunks) || num_chunks > kMaxChunksPerRing ||
+      chunk_size < kMinChunkSize || chunk_size % kChunkAlignmentBytes != 0) {
+    return std::nullopt;
+  }
+  // Two 32-bit factors and the header cannot overflow 64 bits. size_t can be
+  // narrower than that.
```
sashwinbalaji: comment no valu...

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+// static
+std::shared_ptr<InProcessTracingV2Bridge> InProcessTracingV2Bridge::Create(
+    base::TaskRunner* relay_task_runner,
+    uint32_t num_chunks,
+    uint32_t chunk_size) {
+  PERFETTO_CHECK(relay_task_runner);
+  const std::optional<size_t> mapping_size =
+      RingMappingSize(num_chunks, chunk_size);
+  if (!mapping_size) {
+    PERFETTO_ELOG("Tracing v2: %u x %u is not a valid ring; staying on v1",
```
sashwinbalaji: fatal pls

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+                  num_chunks, chunk_size);
+    return nullptr;
+  }
+  // PagedMemory is zero-filled. Zero is both (write_pos=0, read_pos=0) and
+  // Free(0), so the ring needs no initialization pass.
+  base::PagedMemory ring_memory =
+      base::PagedMemory::Allocate(*mapping_size, base::PagedMemory::kMayFail);
+  if (!ring_memory.IsValid()) {
+    PERFETTO_ELOG(
```
sashwinbalaji: fatal pls if any such failure stop it, dont switch to v1.

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+    // their sequence. The task owns the bridge: discarding the task during
+    // shutdown still destroys it.
+    std::shared_ptr<InProcessTracingV2Bridge> owner(bridge);
+    // Move the only reference into the task. The relay can run it before
+    // PostTask() returns; keeping another copy here could then make the bridge
+    // destructor run on this thread instead of the relay.
+    relay_task_runner->PostTask([owner = std::move(owner)]() mutable {
+      DeleteWhenQuiescent(std::move(owner));
+    });
+  };
```
sashwinbalaji: i dont undertstand the deleter. is it actualk possiblity or hypothesis. can the comments be made clearer ?

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+  return std::shared_ptr<InProcessTracingV2Bridge>(
+      new InProcessTracingV2Bridge(std::move(ring_memory), chunk_size,
+                                   relay_task_runner),
+      std::move(deleter));
+}
+
+InProcessTracingV2Bridge::InProcessTracingV2Bridge(
```
sashwinbalaji: why have both static create and  a constructor ? is consturcotr private if so what benefit it gives ?

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+      // Only stores the pointer; nothing is called back before this
+      // constructor returns.
+      ring_buffer_reader_(&ring_buffer_, this),
+      weak_runner_(relay_task_runner) {
+  // Construction happens on the muxer sequence. All subsequent bridge work is
+  // confined to the relay.
+  PERFETTO_DETACH_FROM_THREAD(thread_checker_);
+}
+
+InProcessTracingV2Bridge::~InProcessTracingV2Bridge() {
```
sashwinbalaji: i dont understand this destructior. is it a established pattern in the codebase ?

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+  const WriterID writer_id = v1_writer->writer_id();
+  // Zero means the v1 arbiter could not allocate an id.
+  if (writer_id == 0)
+    return v1_writer;
+
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    // Writer creation races endpoint teardown. Checking and inserting under
+    // the same lock gives the two operations one order.
+    if (!accepting_writers_)
```
sashwinbalaji: seems like using accepting_writers_ and related complexity we are trying to make shutdown gracefule which goes agaiuont https://github.com/google/perfetto/pull/7089#pullrequestreview-5051655372

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (!accepting_writers_)
+      return;
+    accepting_writers_ = false;
+  }
+  if (weak_runner_.task_runner()->RunsTasksOnCurrentThread()) {
+    ReleaseV1WritersOnRelaySequence();
+    return;
+  }
+  weak_runner_.PostTask([this] { ReleaseV1WritersOnRelaySequence(); });
```
sashwinbalaji: why this dance ?

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+    ScheduleDrain();
+}
+
+bool InProcessTracingV2Bridge::DrainTowardsWatermark(uint32_t target_read_pos) {
+  PERFETTO_DCHECK_THREAD(thread_checker_);
+  for (;;) {
+    const uint32_t positions_to_watermark = NumOutstandingPositions(
+        target_read_pos, ring_buffer_reader_.read_pos());
+    // An ordinary drain queued ahead of this task can pass the sampled
+    // watermark. In that case the unsigned distance is close to UINT32_MAX,
```
sashwinbalaji: what valie this sentence adds and is so confusing:
In that case the unsigned distance is close to UINT32_MAX,
    // whereas a watermark still ahead is at most one ring long.

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+    // whereas a watermark still ahead is at most one ring long.
+    if (positions_to_watermark == 0 ||
+        positions_to_watermark > ring_buffer_.num_chunks()) {
+      return true;
+    }
+
+    // Positions at or beyond |target_read_pos| belong to a later drain.
+    const SharedRingBufferReader::DrainResult result =
+        ring_buffer_reader_.Drain(
+            std::min(positions_to_watermark, kMaxPositionsPerPass));
```
sashwinbalaji: watermark as term is difficult to understna. and doesnt give much info / context

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+    PERFETTO_DCHECK_THREAD(thread_checker_);
+    barriers_.emplace_back();
+    Barrier& barrier = barriers_.back();
+    barrier.id = next_barrier_id_++;
+    barrier.target_read_pos = target_read_pos;
+    barrier.writer_id = writer_id;
+    barrier.completion = std::move(completion);
+    // Only the front barrier runs; a queue of one has nothing running yet.
+    if (barriers_.size() == 1)
+      RunFrontBarrier();
```
sashwinbalaji: this optimisaton is it really needed ?

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+// pass. After teardown every queued barrier completes, in order.
+void InProcessTracingV2Bridge::RunFrontBarrier() {
+  PERFETTO_DCHECK_THREAD(thread_checker_);
+  while (!barriers_.empty()) {
+    Barrier& barrier = barriers_.front();
+    // Waiting for v1 acknowledgements; the last one resumes this.
+    if (barrier.pending_flush_acks != 0)
+      return;
+
+    if (!stopped_forwarding_) {
```
sashwinbalaji: again seems like dance / complexity for gracefult shutdown which is just impossible....

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+        });
+        return;
+      }
+
+      std::vector<WriterID> writers;
+      if (barrier.writer_id != 0) {
+        writers.push_back(barrier.writer_id);
+      } else {
+        writers.swap(dirty_writers_);
+      }
```
sashwinbalaji: what is going on here....

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+  }
+}
+
+void InProcessTracingV2Bridge::OnFrontBarrierFlushAcknowledged(
+    uint64_t barrier_id) {
+  PERFETTO_DCHECK_THREAD(thread_checker_);
+  // Teardown may have completed this barrier already.
+  if (barriers_.empty() || barriers_.front().id != barrier_id)
+    return;
+  PERFETTO_DCHECK(barriers_.front().pending_flush_acks > 0);
```
sashwinbalaji: more teardown shenanigans

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+  // Release can remove the writer while its old SDK writer is still alive.
+  // There is no v1 connection left to acknowledge that flush.
+  if (v1_writer) {
+    v1_writer->Flush(std::move(callback));
+  } else if (callback) {
+    callback();
+  }
+}
+
+void InProcessTracingV2Bridge::OnChunkRead(
```
sashwinbalaji: this is the most critical func and comes at line 434 anyone reading wuld be tired by now. can we order better. also some commetns or diagrams woul help a lot

src/tracing/v2/in_process_tracing_v2_bridge.cc
```diff
@@ -0,0 +1,633 @@
...
+      packet->set_previous_packet_dropped(state->pending_data_loss);
+      state->pending_data_loss = 0;
+    }
+  }  // The handle's destructor finalizes the packet.
+  ++num_packets_forwarded_;
+
+  if (std::find(dirty_writers_.begin(), dirty_writers_.end(), writer_id) ==
+      dirty_writers_.end()) {
+    dirty_writers_.push_back(writer_id);
+  }
```
sashwinbalaji: yeah I am struggling with dirty dance

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+namespace perfetto {
+
+namespace tracing_v2 {
+class InProcessTracingV2Bridge;
+}
+
+namespace internal {
+
+// Number of TraceWriters routed through v2 since process start. Test-only.
+uint64_t GetTracingV2WritersCreatedForTesting();
```
sashwinbalaji: why not make it static ?

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+//
+// Method contracts:
+//   - CreateTraceWriter() is thread-safe and runs on arbitrary SDK threads.
+//   - The other ProducerEndpoint methods and the teardown methods run on the
+//     muxer sequence.
+//   - WaitForRelayQuiescenceForTesting() runs outside both internal sequences.
+//
+// NotifyFlushComplete(), NotifyDataSourceStopped() and Sync() first drain
+// through the ring position observed by the call. The control request is sent
+// only after the downstream writers have acknowledged their flushes. An
```
sashwinbalaji: dont say downstream it will confuse. use v1/v2 where suitable pls

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+class InProcessTracingV2Bridge;
+}
+
+namespace internal {
+
+// Number of TraceWriters routed through v2 since process start. Test-only.
+uint64_t GetTracingV2WritersCreatedForTesting();
+
+// ProducerEndpoint decorator for the temporary in-process v2 path. Ordinary
+// TraceWriters use the v2 ring and are relayed through a v1 writer from
```
sashwinbalaji: a quick block diagram would be great

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+// Method contracts:
+//   - CreateTraceWriter() is thread-safe and runs on arbitrary SDK threads.
+//   - The other ProducerEndpoint methods and the teardown methods run on the
+//     muxer sequence.
+//   - WaitForRelayQuiescenceForTesting() runs outside both internal sequences.
+//
+// NotifyFlushComplete(), NotifyDataSourceStopped() and Sync() first drain
+// through the ring position observed by the call. The control request is sent
+// only after the downstream writers have acknowledged their flushes. An
+// unbound v1 SharedMemoryArbiter (startup tracing) parks that acknowledgement
```
sashwinbalaji: why care about starup tracig when we are not supporting it ? also why we not supoprting ?

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+// in pending_flush_callbacks_ and runs it after the deferred CommitData, so
+// the order holds there too.
+//
+// |relay_task_runner| is separate from the muxer sequence. TracingMuxerImpl
+// owns both task runners.
```
sashwinbalaji: jsut saying sepataye is not enough explain them also... and explain why separate also and what purpose

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+  void Sync(std::function<void()> callback) override;
+
+  // Test-only. Waits until existing ring data has passed through v1 and the
+  // service has acknowledged it. Must not run on the muxer or relay sequence.
+  void WaitForRelayQuiescenceForTesting();
+
+  // Starts releasing the v1 writers retained by the relay. Returns immediately
+  // and is idempotent. The endpoint's SharedMemoryArbiter must stay alive until
+  // relay_retains_writers() is false.
+  void StartReleasingRelayWriters();
```
sashwinbalaji: again saying v1 would be great instead of relay. also explain why this needed and that too as public

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+
+ private:
+  // The descriptor |v1_endpoint_| registers, with no_flush cleared while the
+  // ring is in the way.
+  DataSourceDescriptor DescriptorForService(const DataSourceDescriptor&) const;
+
+  // Forwards a control request after the relay has resolved every ring
+  // position reserved before this call. The request runs on the muxer
+  // sequence, with a null endpoint if this decorator is gone by then.
+  void ForwardAfterPendingRingData(
```
sashwinbalaji: explain why and what purpose

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+      std::function<void(ProducerEndpoint*)> request);
+
+  // Pushes v1 commits issued by the relay. Used by the blocking test helper;
+  // muxer sequence only.
+  void FlushPendingCommitDataRequests();
```
sashwinbalaji: if for testing only, name it so

src/tracing/internal/tracing_v2_producer_endpoint.h
```diff
@@ -0,0 +1,135 @@
...
+
+  const std::unique_ptr<ProducerEndpoint> v1_endpoint_;
+  base::TaskRunner* const muxer_task_runner_;
+  // Null if ring allocation failed; all calls then use v1.
```
sashwinbalaji: we shouldn't this if alloc failed just fatal

src/tracing/internal/tracing_v2_producer_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+
+void WaitForQuiescenceForTesting(ProducerEndpoint* endpoint) {
+  static_cast<TracingV2ProducerEndpoint*>(endpoint)
+      ->WaitForRelayQuiescenceForTesting();
+}
+
+const TracingV2EndpointFunctions kTracingV2EndpointFunctions{
+    /*is_supported=*/&IsSupported,
+    /*create_endpoint=*/&CreateProducerEndpoint,
+    /*start_releasing_writers=*/&StartReleasingWriters,
```
sashwinbalaji: need to name this functions better if these are api especially start releasing and retain and also explain more clearly why they needed

src/tracing/internal/tracing_v2_producer_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+      weak_ptr_factory_(this) {
+  PERFETTO_CHECK(v1_endpoint_);
+  PERFETTO_CHECK(muxer_task_runner_);
+}
+
+TracingV2ProducerEndpoint::~TracingV2ProducerEndpoint() {
+  // Retained writers return their WriterIDs to the v1 arbiter during
+  // destruction, so they must be gone before |v1_endpoint_|.
+  PERFETTO_DCHECK(!relay_retains_writers());
+  StartReleasingRelayWriters();
```
sashwinbalaji: yeah confusin becaudse of using v1 somewhere and relay somehwere I see we have releaserelay and releasev1 now even more confused

src/tracing/internal/tracing_v2_producer_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+    const DataSourceDescriptor& descriptor) const {
+  if (!in_process_bridge_ || !descriptor.no_flush())
+    return descriptor;
+  // no_flush lets the service skip the producer flush and scrape the shared
+  // memory itself. The scrape cannot see the ring, so a consumer flush or
+  // FlushAndCloneSession could miss the ring tail. Clear the bit on the copy
+  // the service sees; the muxer still never calls OnFlush() for a no_flush
+  // data source (TracingMuxerImpl::FlushDataSource_AsyncBegin).
+  DataSourceDescriptor for_service = descriptor;
+  for_service.set_no_flush(false);
```
sashwinbalaji: explain this to me, i dont undertsand this at all!

src/tracing/internal/tracing_v2_producer_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+
+std::unique_ptr<TraceWriter> TracingV2ProducerEndpoint::CreateTraceWriter(
+    BufferID target_buffer,
+    BufferExhaustedPolicy buffer_exhausted_policy) {
+  std::unique_ptr<TraceWriter> v1_writer =
+      v1_endpoint_->CreateTraceWriter(target_buffer, buffer_exhausted_policy);
+  if (!in_process_bridge_)
+    return v1_writer;
+
+  // Use the same policy on both hops. Otherwise the relay could drop a packet
```
sashwinbalaji: both hops wdym be direct and clear

src/tracing/internal/tracing_v2_producer_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+  // already accepted by a kStall v2 writer. This means a stalling v1 writer
+  // can delay other rings on the shared relay sequence. The coupling belongs
+  // to this temporary v1 forwarding hop and disappears with the adapter.
+  TraceWriter* const v1_writer_raw = v1_writer.get();
+  std::unique_ptr<TraceWriter> writer = in_process_bridge_->CreateTraceWriter(
+      std::move(v1_writer), target_buffer, buffer_exhausted_policy);
+  // The bridge hands the plain v1 writer back when it cannot use the WriterID.
+  if (writer.get() != v1_writer_raw)
+    g_tracing_v2_writers_created.fetch_add(1, std::memory_order_relaxed);
```
sashwinbalaji: ohh this seems hacky to me. is it good design ? if so why?

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -190,11 +191,13 @@ TracingMuxerImpl::ProducerImpl::ProducerImpl(
     TracingMuxerImpl* muxer,
     TracingBackendId backend_id,
     uint32_t shmem_batch_commits_duration_ms,
-    bool shmem_direct_patching_enabled)
+    bool shmem_direct_patching_enabled,
+    const TracingV2EndpointFunctions* tracing_v2_endpoint_functions)
```
sashwinbalaji: to undewrstand the flow who calls impls supplying this

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -267,12 +293,29 @@ void TracingMuxerImpl::ProducerImpl::OnDisconnect() {
...
+  // TryShutdown() keeps the endpoint alive until every WriterID is returned.
+  bool v2_release_pending = false;
+  if (tracing_v2_endpoint_) {
+    tracing_v2_endpoint_functions_->start_releasing_writers(
+        tracing_v2_endpoint_);
+    // That call closed the accept set, so this can only go from true to false
+    // (when the relay runs the release). False means there is nothing for
+    // SweepDeadServices() to wait for - and the sweep must not run on an
+    // endpoint that never set up tracing: an in-process endpoint without an
+    // arbiter aborts in MaybeSharedMemoryArbiter().
```
sashwinbalaji: i dont undertsand any of this comment make this clearer and explain to me first!

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -1716,6 +1766,12 @@ bool TracingMuxerImpl::FlushDataSource_AsyncBegin(
     PERFETTO_ELOG("Could not find data source to flush");
     return true;
   }
+  // A no_flush data source never gets OnFlush(). The service still asks for
+  // it when the tracing v2 endpoint cleared the bit (see
+  // TracingV2ProducerEndpoint::DescriptorForService) so that the ring is
+  // drained before the acknowledgement. Report it as flushed.
+  if (ds.no_flush)
+    return true;
```
sashwinbalaji: this seriously looks like a side feature not related to tracing v2 to me. I am not saying we should not do it but we should clutter already big cl

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -2211,6 +2267,38 @@ bool TracingMuxerImpl::EnableDirectSMBPatchingForTesting(
   return true;
 }
 
+void TracingMuxerImpl::SetMaxProducerReconnectionsForTesting(uint32_t count) {
```
sashwinbalaji: why the function moved here

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -2211,6 +2267,38 @@ bool TracingMuxerImpl::EnableDirectSMBPatchingForTesting(
...
+  max_producer_reconnections_.store(count);
+}
+
+void TracingMuxerImpl::WaitForTracingV2QuiescenceForTesting() {
```
sashwinbalaji: why this above nontesting func findconsumer ?

src/tracing/internal/tracing_muxer_impl.cc
```diff
@@ -2831,15 +2947,26 @@ void TracingMuxerImpl::Shutdown() {
...
+  // tasks before joining it; destroying a ThreadTaskRunner does not drain its
+  // immediate-task queue. The muxer thread stays alive while they run.
+  if (owned_relay_task_runner) {
+    base::WaitableEvent relay_drained;
+    owned_relay_task_runner->PostTask(
+        [&relay_drained] { relay_drained.Notify(); });
+    relay_drained.Wait();
+  }
+
+  owned_relay_task_runner.reset();
```
sashwinbalaji: do this in if block why not in if ?

src/shared_lib/test/utils.cc
```diff
@@ -73,6 +73,9 @@ std::vector<uint8_t> TracingSession::Builder::BuildProtoConfig() {
 
       perfetto_protos_DataSourceConfig_set_cstr_name(&ds_cfg,
                                                      data_source_name_.c_str());
+      if (use_tracing_v2_) {
+        perfetto_protos_DataSourceConfig_set_use_tracing_v2(&ds_cfg, true);
+      }
```
sashwinbalaji: why not just do below, avoids need for if block ?
```
 perfetto_protos_DataSourceConfig_set_use_tracing_v2(&ds_cfg, use_tracing_v2_)
```

protos/perfetto/config/trace_config.proto
```diff
@@ -229,6 +229,20 @@ message TraceConfig {
...
+    // from 256 through 32768 bytes (the current v1 SMB page-size limit).
+    // At least two chunks must fit within the producer's SMB size. The adapter
+    // uses this size as the ring's chunk-storage budget and adds a ring header.
+    //
+    // The value is fixed when shared memory is set up. Changing it requires the
+    // producer to reconnect. A producer-provided SMB is set up at connection
+    // time, before any TraceConfig, and always uses the default chunk size.
+    // V1 writers do not use this setting. The service rejects invalid values
+    // even if no data source selects v2.
+    optional uint32 tracing_v2_chunk_size_bytes = 4;
```
primiano, line 245: add experimental_ prefix and clarify this will go away eventually and is only for testing. You need the prefix only here, you can leave thigns as they arae in the .h and .cc code elsewhere

src/tracing/service/tracing_service_impl.cc
```diff
@@ -458,8 +474,12 @@ TracingServiceImpl::ConnectProducer(Producer* producer,
           "Adopting producer-provided SMB of %zu kB for producer \"%s\"",
           shm_size / 1024, endpoint->name_.c_str());
       auto shmem_mode = GetShmemMode(client_identity, in_process);
-      endpoint->SetupSharedMemory(std::move(shm), page_size,
-                                  /*provided_by_producer=*/true, shmem_mode);
+      endpoint->SetupSharedMemory(
+          std::move(shm), page_size,
+          /*provided_by_producer=*/true, shmem_mode,
+          // This connection has no TraceConfig at SMB setup, so a
```
primiano, line 480: imho the comment here is distracting. I feel it's obvious and also breaks too much the reading of the function args. jus keep the /*...sizE_bytes=/=0 part

src/tracing/service/tracing_service_impl.cc
```diff
@@ -656,6 +676,21 @@ base::Status TracingServiceImpl::EnableTracing(ConsumerEndpointImpl* consumer,
     }
   }
 
+  // Validate the chunk size before sending it to a selected producer, whose
+  // SDK aborts setup if the value is invalid.
+  for (const auto& producer : cfg.producers()) {
+    const uint32_t chunk_size = producer.tracing_v2_chunk_size_bytes();
+    if (chunk_size != 0 && (chunk_size < kMinTracingV2ChunkSize ||
+                            chunk_size > kMaxTracingV2ChunkSize ||
+                            chunk_size % kTracingV2ChunkAlignment != 0)) {
```
primiano, line 685: doesn't it have to be a pow2 btw? if the full buffer must be a pow2, and you pick a chunk size of 384 bytes, isn't it going to cause problems because of the leftover?

src/tracing/service/tracing_service_structs.h
```diff
@@ -106,6 +106,18 @@ struct DataSourceInstance {
   DataSourceInstance(const DataSourceInstance&) = delete;
   DataSourceInstance& operator=(const DataSourceInstance&) = delete;
 
+  // The v2 ring buffer lives in the producer process and is not shared with
+  // the service. Ask the producer to drain it even for no_flush instances.
+  bool RequiresProducerFlush() const {
+    return !no_flush || config.use_tracing_v2();
```
primiano, line 112: hmm this could be problematic.
The fact is that there is a difference betwene
1) We ask the data source to use V2 (whichj is what you are inspecting here when looking at the config)

2) The data source actually supports V2 (which depends on the SDK version)

I feel this should be left as-is in traced, and instead we should change the behaviour at the SDK level, so if a dat source uses V2, we should set no_flush and will_notify, not imply from the traced side, as we don't have enough info  in traced to tell if will be the case

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -244,6 +252,47 @@ class TracingMuxerImpl : public TracingMuxer {
     void SendOnConnectTriggers();
     void NotifyFlushForDataSourceDone(DataSourceInstanceID, FlushRequestID);
 
+    // --- Tracing v2. ---
+    // A v1-only connection never creates a TracingV2Connection or calls these
```
primiano, line 256: i mean, this is kinda expected/obvious.
Also the "tracing v2 -- marker". the name of the struct makes it quite obvious

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -244,6 +252,47 @@ class TracingMuxerImpl : public TracingMuxer {
...
+
+    // Publishes the endpoint and bridge together so writer creation can check
+    // that both belong to the same connection. Declare |endpoint| first so this
+    // object's bridge reference is released before its endpoint reference.
+    // Other bridge owners can survive this object. ProducerImpl::dead_services_
+    // then retains the endpoint while the bridge's v1 writers still need it.
+    struct TracingV2Connection {
+      // Creates the downstream v1 writer and wraps it in a v2 writer. The
+      // caller's exhaustion policy applies only to the v2 ring buffer.
+      // Thread-safe.
```
primiano, line 267: "Can be called on any thread"

When I see "thread safe" i never know if I have to intrepret as "the caller must be thread-safe / use a lock" or "can be called on any thread"

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -244,6 +252,47 @@ class TracingMuxerImpl : public TracingMuxer {
...
+      std::unique_ptr<TraceWriter> CreateTraceWriter(BufferID,
+                                                     BufferExhaustedPolicy);
+
+      std::shared_ptr<ProducerEndpoint> endpoint;
```
primiano, line 271: why the shared_ptr?
the problem of shared_ptr is that they create this situation where an object can be destroyed on thread A or B depending on whoever is last. but then can ProducerEndpoint be destroyed on any thread?
If you really need a shared_ptr here then you need to make sure to PostTask the destruction of the last sharedptr on the right thread

src/tracing/internal/tracing_muxer_impl.h
```diff
@@ -273,14 +322,53 @@ class TracingMuxerImpl : public TracingMuxer {
     // expiration).
     std::list<std::pair<std::string, base::TimeMillis>> on_connect_triggers_;
 
-    std::map<FlushRequestID, std::set<DataSourceInstanceID>> pending_flushes_;
+    // Ring buffer work needed before acknowledging a flush request.
+    enum class RingBufferDrainState {
```
primiano, line 326: maybe call it RingBufferV2DrainState as this seems about V2?

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,281 @@
...
+namespace perfetto::test {
+class TracingMuxerImplV2Test;
+}
+
+namespace perfetto::tracing_v2 {
+
+class InProcessTracingV2BridgeTestPeer;
+
+// Temporary adapter that lets SDK data sources write through the tracing v2
+// ring buffer while traced reads the v1 shared memory buffer (SMB). The bridge
```
primiano, line 53: I'd say "and then rewrites them in the same process into a v1 SMB, which is what traced sees"

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,281 @@
...
+// reads the ring buffer in the producer process and forwards complete packets
+// through v1 TraceWriters. NotifyReader() schedules a drain on the relay
```
primiano, line 55: 1 V1 TraceWriter for each V2 tracewriter

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,281 @@
...
+//                                          v
+//                                    shared v1 SMB
+//                                          |
+//                                          v
+//                                       traced
+//                                          |
+//                                          v
+//                                 session TraceBuffer
+//
+// Once traced reads the ring buffer directly, the relay sequence, retained
```
primiano, line 80: make it clar this is a "FUTURE PLAN:"

src/tracing/v2/in_process_tracing_v2_bridge.h
```diff
@@ -0,0 +1,281 @@
...
+//
+// The bridge owns the ring buffer memory, its reader, and one WriterState per
+// v2 WriterID. Each WriterState holds packet reassembly state and a v1 writer.
+// SDK threads register writers under |mutex_|. The relay reads the ring buffer,
+// reassembles packets, forwards them to v1, and processes barriers.
+// A barrier delays a control operation until the reader reaches a sampled
+// ring buffer position. See Barrier below.
+//
+// Each TraceWriterV2Impl retains the bridge as its Delegate. Queued relay tasks
+// also hold shared references, so destruction can happen on any thread.
```
primiano, line 101: do we want this though (destruction on any thread?)

src/tracing/v2/relay_sequence.h
```diff
@@ -0,0 +1,86 @@
...
+    PERFETTO_CHECK(task_runner_);
+  }
+
+  RelaySequence(const RelaySequence&) = delete;
+  RelaySequence& operator=(const RelaySequence&) = delete;
+
+  // Queues |task| on the relay. Returns false and drops it after Close().
+  // The caller decides how to handle rejected work. Thread-safe.
+  bool PostTask(std::function<void()> task) {
+    std::lock_guard<std::mutex> lock(mutex_);
```
primiano, line 55: hmm let's discuss this. this seems to undo the lockfreetaskrunner. i'm not fully sure i understand the whole problem space this is solving though, so need some f2f discussion

primiano (commented):
sensding out partial comments

PR #7328, opened 2026-09-03
tp: add pipeline breakers

LalitMaganti:
A plan is push all the way up and only pulled from at the top. Some steps
cannot be pushed through because nothing can come out until everything
has gone in, sorting being the classic case. Until now each such step
had to be written as a Source wrapping a Source and drain its input by
hand, hiding that it is a sink for one pipeline and the source of the
next.

Breaker makes that shape explicit. The pipeline below pushes each batch
into Consume(), Finish() runs once after the last, and only then does the
pipeline above pull batches out. The base owns the input state, the drain
loop, error propagation from the input and rewinding, so a concrete
breaker only supplies the two halves.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -17766,6 +17766,7 @@ filegroup {
     name: "perfetto_src_trace_processor_core_exec_exec",
     srcs: [
         "src/trace_processor/core/exec/assert_type.cc",
+        "src/trace_processor/core/exec/breaker.cc",
         "src/trace_processor/core/exec/column_view.cc",
         "src/trace_processor/core/exec/dataframe_scan.cc",
         "src/trace_processor/core/exec/operator.cc",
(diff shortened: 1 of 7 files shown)


src/trace_processor/core/exec/operator.h
```diff
@@ -121,6 +121,57 @@ class Source {
...
+// A plan is push all the way up: a Pipeline pushes each batch of its source
+// through its operators. Only at the top is it pulled from. Some steps cannot
+// be pushed through, because nothing can come out until everything has gone
+// in; sorting is the classic case. A breaker is that step: the pipeline below
+// pushes every batch into Consume(), Finish() runs once after the last, and
+// only then does the pipeline above start pulling batches out.
+//
+// What drives the push is the first pull: GetData() drains the input before
+// serving anything, so a breaker sits in a plan wherever a Source does.
+class Breaker : public Source {
```
sashwinbalaji: should we move it to its own file ? or is it for later

src/trace_processor/core/exec/operator.cc
```diff
@@ -16,10 +16,61 @@
...
+  if (!s.status.ok()) {
+    return false;
+  }
+  if (!s.filled) {
+    while (input_.GetData(s.batch, *s.input)) {
+      if (!Consume(s.batch, s)) {
+        return false;
+      }
+    }
+    if (!input_.status(*s.input).ok()) {
```
sashwinbalaji: nit: creating input status twice

sashwinbalaji (approved):

PR #7329, opened 2026-09-03
tp: finish operators after the last input

LalitMaganti:
An operator which holds rows back has nowhere to let them go once the
source runs dry, and no way to report rows it is still holding. Finish()
is called once after the last batch and can return output, fan out, or
fail, just as Execute() can.

The pipeline finishes its operators from the bottom up. Whatever one lets
go is pushed through the operators above it as ordinary input before the
next one is finished, so an operator sees all of its input before its own
Finish() runs.

### DIFF
--- a/src/trace_processor/core/exec/operator.h
+++ b/src/trace_processor/core/exec/operator.h
@@ -66,7 +66,8 @@ enum class OpResult : uint8_t {
 // One input batch can produce more than one output batch: an operator which
 // fans out returns kHaveMoreOutput and is called again with the same input.
 // This is why input and output are separate batches: the input has to survive
-// being read more than once.
+// being read more than once. Output can also come after the last input, from
+// Finish().
 class Operator {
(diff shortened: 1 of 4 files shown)


src/trace_processor/core/exec/operator.h
```diff
@@ -81,6 +82,15 @@ class Operator {
                            RowBatch& out,
                            OperatorState& state) const = 0;
 
+  // Called once after the last input batch, for an operator which holds rows
+  // back to let them go. The results mean what they do for Execute(): keep
+  // returning kHaveMoreOutput to be called again. An operator which holds
+  // nothing back leaves `out` empty.
+  virtual OpResult Finish(RowBatch& out, OperatorState&) const {
```
sashwinbalaji: seems similar to breaker

sashwinbalaji (approved):

PR #7333, opened 2026-09-03
ui: add copy button for aggregate profile name and fix initial render

gignat-dev:
Add a copy button next to the aggregate profile selector so member filenames in multi-pprof archives can be copied from the native select. 

Also fix the initial page render (existing bug) by scheduling a redraw and propagating the selected profile state to createFlamegraph so the flamegraph mounts immediately on load.

### DIFF
--- a/ui/src/plugins/dev.perfetto.AggregateProfiles/aggregate_profiles_page.ts
+++ b/ui/src/plugins/dev.perfetto.AggregateProfiles/aggregate_profiles_page.ts
@@ -56,8 +56,19 @@ export class AggregateProfilesPage implements m.ClassComponent<AggregateProfiles
         selectedProfileId: selectedProfile?.id,
       });
       if (selectedProfile) {
-        this.createFlamegraph(attrs, selectedProfile);
+        // createFlamegraph spreads the state it is given, so hand it one
+        // already carrying the selection or it writes the old one back.
+        this.createFlamegraph(
+          {
+            ...attrs,
+            state: {...attrs.state, selectedProfileId: selectedProfile.id},
+          },
+          selectedProfile,
+        );
       }
+      // Editing the store notifies nobody and `attrs.state` is this render's
+      // snapshot, so the new state only shows on the next render.
+      attrs.trace.raf.scheduleFullRedraw();
     }
     return m(
       Stack,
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.AggregateProfiles/aggregate_profiles_page.ts
```diff
@@ -55,8 +56,19 @@ export class AggregateProfilesPage implements m.ClassComponent<AggregateProfiles
...
+          {
+            ...attrs,
+            state: {...attrs.state, selectedProfileId: selectedProfile.id},
+          },
+          selectedProfile,
+        );
       }
+      // Editing the store notifies nobody and `attrs.state` is this render's
+      // snapshot, so the new state only shows on the next render.
+      attrs.trace.raf.scheduleFullRedraw();
```
LalitMaganti: I don't think you need this.

gignat-dev: I tested without it and the flamegraph doesn't show up on initial load. 

`store.edit()` doesn't trigger a redraw, so without scheduling one explicitly, Mithril never re-renders with the initialized flamegraphState.

Did you mean this is not recommended?

LalitMaganti: Hmmmmm @stevegolton what would be the advice here?

stevegolton: I'm a bit confused if I'm honest. I'll need to look at the code a little deeper.

From a cursory glance I suspect that we're mutating some state that view depends on within the view function, so we need to schedule a redraw again to refresh the UI from that mutated state.

This is usually an anti-pattern; view functions should be side-effect free (aside from maybe some internal caching).

Taking a deeper look though... I could be entirely wrong.

stevegolton: I just uploaded a new commit on top. The whole page component was a little confused and my commit sorts the state management properly IMO.

Feel free to take it or roll it back or whatever, but I think the initial load is working now without any manual redraws.

stevegolton (approved):

ui/src/plugins/dev.perfetto.AggregateProfiles/aggregate_profiles_page.ts
gignat-dev: Thank you so much, Steve!

I can confirm the initial load is working properly.

PR #7336, opened 2026-09-03
ui: step through aggregate profiles with the arrow keys

gignat-dev:
Reading an archive one profile at a time meant reopening the dropdown for every step. Left and right now move to the neighbouring profile.

The dropdown steps through its own options when it has focus, so the handler stays out of the way while it, or any editable element, is focused. It also checks the page is on screen, since a page navigated away from stays mounted behind a closed gate.

### DIFF
--- a/ui/src/plugins/dev.perfetto.AggregateProfiles/aggregate_profiles_page.ts
+++ b/ui/src/plugins/dev.perfetto.AggregateProfiles/aggregate_profiles_page.ts
@@ -22,6 +22,7 @@ import {Button} from '../../widgets/button';
 import {Callout} from '../../widgets/callout';
 import {CopyToClipboardButton} from '../../widgets/copy_to_clipboard_button';
 import {EmptyState} from '../../widgets/empty_state';
+import {HotkeyContext} from '../../widgets/hotkey_context';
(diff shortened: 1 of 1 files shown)


ui/src/plugins/dev.perfetto.AggregateProfiles/aggregate_profiles_page.ts
```diff
@@ -73,6 +87,52 @@ export class AggregateProfilesPage implements m.ClassComponent<AggregateProfiles
...
+    attrs.onStateChange({
+      selectedProfileId: profile.id,
+      flamegraphState: updateTreeExplorerState(
+        attrs.state.flamegraphState,
+        profile.metrics,
+      ),
+    });
+  }
+
+  private readonly onKeyDown = (e: KeyboardEvent): void => {
```
stevegolton: The HotkeyContext component should be able to replace most if not all of this.

gignat-dev: Thank you! 
Done!

Indeed it's much cleaner

stevegolton (approved):

PR #7339, opened 2026-09-04
tools: add java_blob_emitter

rukkal:
This commit adds java_blob_emitter.py, a build-time codegen tool
analogous to the already-existing cpp_blob_emitter.py. The tool takes
binary data as input and emits a Java class with a
"public static byte[] getBytes()" method.

Because the JVM spec imposes a strict 64KB bytecode method size limit,
byte array literals containing roughly more than ~10,900 elements would
result in a too large static initializer and crash the compilation with
a "code too large" error. As a mitigation, the tool dynamically falls
back to storing the data into a base64 string constant when the binary
payload exceeds 8KB. If the threshold is not exceeded, the data is
stored as a byte array literal.

Bug: 512718675


python/tools/java_blob_emitter.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python3
```
sashwinbalaji, line 1: nit: copyright for this file also

sashwinbalaji (approved):

python/tools/java_blob_emitter.py
rukkal: Done

PR #7340, opened 2026-09-04
stdlib: add android.lock_held stdlib module

gignat-dev:
Add `android.lock_held` stdlib module to track intervals when named locks are held. 

Merge recursive holds, truncate exclusive holds across threads, and adopt in the `AndroidLockContention` UI plugin.


src/trace_processor/perfetto_sql/stdlib/android/lock_held.sql
```diff
@@ -0,0 +1,81 @@
...
+  replace(name, '_lock_held', '') AS lock_name,
+  utid
+FROM thread_slice
+WHERE
+  name GLOB '*_lock_held';
+
+-- Intervals during which a named lock was held by a thread.
+-- Overlapping holds on the same thread (recursive locks) are merged.
+-- Holds on exclusive locks are truncated when acquired by a successor thread.
+CREATE PERFETTO TABLE android_lock_held(
```
zezeozue: Can you please add the (short) *blocking method name* when it exists. Something like this:

  (SELECT mc.short_blocking_method
     FROM android_monitor_contention AS mc
     WHERE mc.blocking_utid = h.utid
       AND mc.ts < h.ts + h.dur
       AND mc.ts + mc.dur > h.ts
     LIMIT 1) AS blocking_method,

(or if theres a more efficient idea)

gignat-dev: You mean one of the blocking_method (they can be more then more if I understand correctly)?

zezeozue: The blocking method is the one holding the lock. Per *one* lock_held slice, there can only be one. Even though there can be multiple monitor contention slices, they should all have the same *blocking* method. But yes there can be multiple blocked methods. We care about the blocking

gignat-dev: Done. 

Did an interval intersect with monitor_contention table
```SQL
_interval_intersect!(
      (
        _android_lock_held_raw,
        (
          SELECT id, ts, dur, blocking_utid AS utid
          FROM android_monitor_contention
          WHERE blocking_utid IS NOT NULL
        )
      ),
      (utid)
    )
```

zezeozue (approved):

LalitMaganti (approved):

PR #7341, opened 2026-09-04
perfetto: filter app registration in Perfetto Java SDK

parthsane:
Filter which Android apps can connect to the system tracing backend (traced) when registering via PerfettoTrace.register(). Only allowlisted apps (com.google.android.youtube,
com.google.android.googlequicksearchbox, and system_server) or apps with persist.debug.perfetto.sdk_enable_tracing_all_apps set to "1" are allowed to register.

Registration using the in-process backend (isBackendInProcess=true) and registration on host environments remain unconditionally enabled.


src/android_sdk/jni/dev_perfetto_sdk_PerfettoTrace.cc
```diff
@@ -60,6 +61,20 @@ static void dev_perfetto_sdk_PerfettoTrace_activate_trigger(JNIEnv* env,
 void dev_perfetto_sdk_PerfettoTrace_register(JNIEnv*,
                                              jclass,
                                              bool is_backend_in_process) {
+  if (!IsAppRegistrationAllowed(is_backend_in_process)) {
```
zezeozue, line 64: can we just do this in java?

parthsane: using the java Sysprop functions in libcore is the constraint. if we do everything in java, then we would need either reflection or a separate jni call to get it via the bionic functions. Or we do what the current code does and does most of it in c++.

src/android_sdk/java/main/dev/perfetto/sdk/TracingPolicy.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  /**
+   * Returns argv[0] from {@code /proc/self/cmdline}, or "" if it can't be read. The framework caches
+   * the same value as {@code Process.myProcessName()} (both set by {@code Process.setArgV0}), but
+   * that isn't reachable here. {@code Files.readAllBytes} can't be used: /proc files report a size
+   * of 0, so it would return nothing.
+   */
+  private static String currentProcessName() {
+    try (FileInputStream in = new FileInputStream("/proc/self/cmdline")) {
```
sashwinbalaji: would this work fine in StrictMode ?

src/android_sdk/java/main/dev/perfetto/sdk/TracingPolicy.java
```diff
@@ -0,0 +1,122 @@
...
+      return !allowlist.isEmpty() && matches(processName, allowlist.split(","));
+    } catch (ReflectiveOperationException | ClassCastException e) {
+      return false;
+    }
+  }
+
+  /**
+   * Returns argv[0] from {@code /proc/self/cmdline}, or "" if it can't be read. The framework caches
+   * the same value as {@code Process.myProcessName()} (both set by {@code Process.setArgV0}), but
+   * that isn't reachable here. {@code Files.readAllBytes} can't be used: /proc files report a size
```
sashwinbalaji: > Files.readAllBytes can't be used: /proc files report a size

I think you can use readallbytes as per http://ac/libcore/ojluni/src/main/java/java/nio/file/Files.java?l=3159 it uses the size field given only as intialsize it would still read to the end

so something like  byte[] cmdline = Files.readAllBytes(Paths.get("/proc/self/cmdline")); should work and make the code a bit simpler ?

src/android_sdk/java/main/dev/perfetto/sdk/PerfettoTrace.java
```diff
@@ -334,18 +334,31 @@ public static void activateTrigger(String triggerName, int ttlMs) {
...
+   * Registers the process with Perfetto.
+   *
+   * <p>The in-process backend is always registered. The system backend is registered only if
+   * {@link TracingPolicy} allows this process; otherwise this is a no-op and Perfetto is left
+   * uninitialized. Callers must not register categories or emit events when this returns {@code
+   * false}.
+   *
+   * @return whether the process is registered with the requested backend
+   */
+  public static boolean register(boolean isBackendInProcess) {
```
sashwinbalaji, line 351: We need to make sure callers of this respect the return value. I worry about this case:

- If a caller ignores a denied register(false) and then registers a category like:

```
PerfettoTrace.register(false); 
category.register();
```

- the native code aborts because tracing was never initialized. 

Could we make category setup safe / update the affected callers before enabling this?

src/android_sdk/java/main/dev/perfetto/sdk/TracingPolicy.java
parthsane: I fixed it to work with strict mode

src/android_sdk/java/main/dev/perfetto/sdk/TracingPolicy.java
```diff
@@ -0,0 +1,134 @@
...
+   * Returns the current process name.
+   *
+   * <p>Checks in-memory framework caches first (e.g. {@code ActivityThread.currentProcessName()}) to
+   * avoid disk reads and StrictMode violations, falling back to reading {@code /proc/self/cmdline}
+   * on non-framework / host Linux environments.
+   */
+  private static String currentProcessName() {
+    // 1. Try in-memory framework process name (avoids disk read / StrictMode entirely).
+    try {
+      Class<?> activityThread = Class.forName("android.app.ActivityThread");
```
sashwinbalaji, line 125: this won't work for system_server (below analysis from internal AI):


For the `system_server` process in Android, this code snippet would **not return anything** (it would skip the `return` statement and fall through) because `ActivityThread.currentProcessName()` returns **`null`** in this context.

Here is a detailed breakdown of why this happens:

### 1. How `ActivityThread.currentProcessName()` works
Internally in AOSP, the implementation of `ActivityThread.currentProcessName()` looks like this:

```java
    public static String currentProcessName() {
        ActivityThread am = currentActivityThread();
        return (am != null && am.mBoundApplication != null)
            ? am.mBoundApplication.processName : null;
    }
```

It relies on the `mBoundApplication` field (which holds `AppBindData`) to get the process name.

### 2. Why it returns `null` for `system_server`
*   **Standard Apps:** When a standard application process starts, it goes through the `bindApplication` flow, which initializes the `mBoundApplication` field with the application's metadata (including its process name).
*   **System Server:** The `system_server` process is special. It initializes `ActivityThread` differently. Instead of the standard flow, it calls `ActivityThread.attach(true)` (where `true` indicates it is a system process). This special initialization sets up the system context but **does not** populate the `mBoundApplication` field. 

Since `mBoundApplication` is `null` in `system_server`, `ActivityThread.currentProcessName()` returns `null`.

### 3. Result of the snippet
1.  `currentProcessName.invoke(null)` yields `null`.
2.  The check `if (name != null && !name.isEmpty())` evaluates to `false`.
3.  The block skips the `return` statement.

### Alternative for `system_server`
If you are trying to get the process name of `system_server` reliably (which is typically `"system"` or `"system_server"` depending on how you query it), you should use standard Linux process utilities or `Application.getProcessName()` on API 28+, though file-based reads (`/proc/self/cmdline`) are often used as a fallback in multi-process utilities.

### References:
* [android.libraries.inputmethod.utils.ProcessUtil.getProcessName](https://source.corp.google.com/piper///depot/google3/java/com/google/android/libraries/inputmethod/utils/ProcessUtil.java)
* [ActivityThread.java](https://source.corp.google.com/h/googleplex-android/platform/superproject/main/+/main:frameworks/base/core/java/android/app/ActivityThread.java)

parthsane: Okay.  Fixed.

src/android_sdk/java/test/dev/perfetto/sdk/test/PerfettoTraceTest.java
```diff
@@ -1443,4 +1443,26 @@ private void collectTrackNames(TracePacket packet) {
...
+  @Test
+  public void testRegisterSystemBackendOnHost() {
+    // On host / default variant, system backend registration is unrestricted.
+    assertThat(PerfettoTrace.register(false)).isTrue();
+    assertThat(PerfettoTrace.getAttempedSystemRegistration()).isTrue();
+  }
+
+  @Test
+  public void testCategorySetupSafeWhenUnregistered() {
+    Category unregCategory = new Category("unregistered_test_category");
```
sashwinbalaji: before each test we do:

```
  @Before
  public void setUp() {
    PerfettoTrace.registerWithDebugChecks(true);
    // 'var unused' suppress error-prone warning
    var unused = FOO_CATEGORY.register();
```

so tracing is already intialized and we are not hitting the early return we're trying to test

```
    public synchronized Category register() {
      if (!sIsRegistered) {
        return this;
      }
```

sashwinbalaji (approved):

PR #7350, opened 2026-09-07
Support pinning metrics via parameter dictionaries in PinAndroidPerfMetrics

nicomazz:
Allow other tools to automatically pin perf metrics by passing an open parameter dictionary, without depending on the legacy structured metric key string format.

Introduce PinIntent and register a command to parse parameters and pin the corresponding CUJs, scoped jank, or blocking calls.


ui/src/plugins/com.android.PinAndroidPerfMetrics/index.ts
```diff
@@ -79,6 +85,25 @@ export default class implements PerfettoPlugin {
...
+
+    /**
+     * Generic entry point for other clients (e.g. Perfetto startup commands)
+     * that provide parameter dictionaries directly.
+     *
+     * Example URL usage:
+     * https://ui.perfetto.dev/#!/?startupCommands=[{"id":"com.android.PinAndroidPerfMetrics#pinRequests","args":["{\"cuj\":\"SHADE_EXPAND\",\"process\":\"systemui\"}"]}]
+     */
+    ctx.commands.registerCommand({
+      id: 'com.android.PinAndroidPerfMetrics#pinRequests',
```
stevegolton: Please remove the plugin name from this command - command names should be able to be moved between plugins or survive plugin renames. Think of a name that cleanly describes the intent of the command but keep the 'com.android' prefix.

ui/src/plugins/com.android.PinAndroidPerfMetrics/index.ts
```diff
@@ -79,6 +85,25 @@ export default class implements PerfettoPlugin {
...
+      name: 'Pin performance tracks from parameter dictionaries',
+      callback: async (...args: unknown[]): Promise<PinIntent[]> => {
+        const input = args.length === 1 ? args[0] : args;
+        const intents = parsePinIntents(input as PinRequestsInput);
```
stevegolton: What happens when this cast fails - probably better to parse it with a zod schema and fail fast.

stevegolton (changes requested):

stevegolton (approved):

PR #7354, opened 2026-09-07
tp: trim per-event work on the track event tokenize and import path

LalitMaganti:
Several costs were paid on every track event without doing anything
useful for it: two heap allocations to look at a single inline
category, two scans of the raw event bytes for out-of-tree fields
that are almost never there, a second typed decode of every packet in
the reader, a sequence-state insert per packet, a walk of the interned
submessage map to find the track event defaults on every call, six
repeated-field iterators for extra counter values, entering the
"debug" args dictionary for events without debug annotations, and
seven separate checks for legacy timestamp fields.

Each is skipped when it does not apply: the category is counted
through its iterator, the typed decoder records whether it skipped an
out-of-range field, the decoder is passed through, the defaults are
memoized per generation, and the legacy fields are tested with one
presence mask.

Instructions to load a 3.1 GB trace of 23.7M track events go from
540.05G to 473.96G (-12.2%), and for chrome_android_systrace from
11.87G to 10.62G (-10.5%). Output is identical on both.

### DIFF
--- a/include/perfetto/protozero/proto_decoder.h
+++ b/include/perfetto/protozero/proto_decoder.h
@@ -386,6 +386,24 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
     return kInvalidField;
   }
 
+  // True if the message contained at least one field with an id beyond the
+  // highest id known in-tree (typically an out-of-tree extension). Such
+  // fields are not stored here; this lets callers skip a selective re-decode
+  // via unknown_fields() when there is nothing to find.
+  bool has_out_of_range_fields() const { return has_out_of_range_fields_; }
+
+  // True if any field whose bit is set in |mask| is present. |mask| is laid
+  // out like the presence bitmap (bit i of word i / 64 for field id i) and
+  // holds |words| words; bits beyond the known field range are ignored.
+  bool HasAnyField(const uint64_t* mask, size_t words) const {
+    size_t presence_words = (num_fields_ + 63) / 64;
+    for (size_t i = 0; i < words && i < presence_words; ++i) {
+      if (presence_[i] & mask[i])
+        return true;
+    }
+    return false;
+  }
+
   // True if the field with the given id was seen while decoding: tests bit
   // |id| of the presence bitmap. The caller must check that |id| is within
   // the bitmap, i.e. <= the decoder's MAX_FIELD_ID. Note that |id| may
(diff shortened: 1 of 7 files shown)


include/perfetto/protozero/proto_decoder.h
```diff
@@ -392,6 +392,24 @@ class PERFETTO_EXPORT_COMPONENT TypedProtoDecoderBase : public ProtoDecoder {
   // exceed |num_fields_|: the selective decoding constructor shrinks
   // num_fields_ to the mask's extent but clears the full bitmap, so that
   // at<>() can keep testing the bit directly.
```
sashwinbalaji: this comment seems more sutiabled for HasAnyField

src/trace_processor/importers/proto/track_event_tokenizer.cc
```diff
@@ -589,23 +609,26 @@ ModuleResult TrackEventTokenizer::TokenizeTrackEventPacket(
...
-  }
-  if (!AddExtraCounterValues(
-          *args.state, data, index, event.extra_double_counter_values(),
-          event.extra_double_counter_track_uuids(),
-          defaults ? defaults->extra_double_counter_track_uuids()
-                   : kEmptyIterator,
-          seq_id, &data.trace_packet_data.packet)) {
-    return ModuleResult::Handled();
+  if (PERFETTO_UNLIKELY(event.has_extra_counter_values() ||
+                        event.has_extra_double_counter_values())) {
```
sashwinbalaji: should we split this guard to save one AddExtraCounterValues call when not needed

src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -259,47 +259,50 @@ class TrackEventEventImporter {
...
 
-    std::vector<uint64_t> category_iids;
-    for (auto it = event_.category_iids(); it; ++it) {
-      category_iids.push_back(*it);
+    // Count first: the common single-category cases below must not allocate.
+    auto category_iids = event_.category_iids();
+    auto category_strings = event_.categories();
+    size_t iid_count = 0;
+    for (auto it = category_iids; it; ++it) {
+      ++iid_count;
```
sashwinbalaji: very minor thing would it better if we stop the loop for both iid_count and string_count when > 1

sashwinbalaji (approved):

PR #7355, opened 2026-09-07
tp: move the tracing service's packet handlers into a module

LalitMaganti:
Service lifecycle events and extension descriptors were handled by
ProtoTraceReader itself although nothing about them needs the reader:
they consume their field and are done, which is what importer modules
are for. They move verbatim into TracingServiceModule. Clock snapshots
stay in the reader, since resolving any packet's timestamp, including
the snapshot's own, depends on them.

The two packet types now pass through the content analyzer like every
other module-handled packet, which moves the proto_content diff test
expectation by one packet. Instructions are unchanged.

### DIFF
--- a/src/trace_processor/importers/proto/default_modules.cc
+++ b/src/trace_processor/importers/proto/default_modules.cc
@@ -22,12 +22,16 @@
 #include "src/trace_processor/importers/proto/memory_tracker_snapshot_module.h"
 #include "src/trace_processor/importers/proto/metadata_minimal_module.h"
 #include "src/trace_processor/importers/proto/proto_importer_module.h"
+#include "src/trace_processor/importers/proto/tracing_service_module.h"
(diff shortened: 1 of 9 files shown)


src/trace_processor/importers/proto/default_modules.cc
```diff
@@ -31,6 +32,8 @@ void RegisterDefaultModules(ProtoImporterModuleContext* module_context,
   // Ftrace and Etw modules are special, because they have an extra method for
   // parsing the ftrace/etw packets. So we need to store a pointer to it
   // separately.
+  module_context->modules.emplace_back(
+      new TracingServiceModule(module_context, context));
```
sashwinbalaji: nit: move it above the comments for ftrace and etw

sashwinbalaji (approved):

PR #7356, opened 2026-09-07
tp: decode each packet once in the proto reader

LalitMaganti:
Every TracePacket was decoded twice on the tokenization path: with the
typed decoder in ParsePacket, which reads the reader's fields by name,
and again with SelectiveTracePacketDecoder for module dispatch.

The selective decoder's allowlist now covers every field the reader
reads by name, so ParsePacket decodes once with it and passes it down.
Its spill list remains exactly the payload fields modules are
dispatched on, in wire order; track events are read by name and handed
to their module directly.

Instructions to load a 3.1 GB trace of 23.7M track events go from
473.96G to 459.93G (-3.0%), and for chrome_android_systrace from
10.62G to 10.25G (-3.5%). Output is identical on both.

### DIFF
--- a/src/trace_processor/importers/proto/proto_trace_reader.cc
+++ b/src/trace_processor/importers/proto/proto_trace_reader.cc
@@ -51,9 +51,11 @@
 #include "src/trace_processor/importers/common/process_tracker.h"
 #include "src/trace_processor/importers/common/stats_tracker.h"
 #include "src/trace_processor/importers/proto/default_modules.h"
+#include "src/trace_processor/importers/proto/metadata_minimal_module.h"
 #include "src/trace_processor/importers/proto/packet_analyzer.h"
 #include "src/trace_processor/importers/proto/proto_importer_module.h"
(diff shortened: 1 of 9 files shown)


src/trace_processor/importers/proto/proto_trace_reader.cc
```diff
@@ -448,9 +448,9 @@ base::Status ProtoTraceReader::TimestampTokenizeAndPushToSorter(
       }
     }
 
-    if ((decoder.has_chrome_events() || decoder.has_chrome_metadata()) &&
-        (!timestamp_clock_id ||
-         timestamp_clock_id == protos::pbzero::BUILTIN_CLOCK_MONOTONIC)) {
+    if ((!timestamp_clock_id ||
```
sashwinbalaji, line 452: has_chrome_metadata check got removed ?

LalitMaganti:
PTAL again.

sashwinbalaji (approved):

PR #7364, opened 2026-09-08
tp: reduce token buffer allocation and extraction work

LalitMaganti:
Return each bump allocation together with its address and retain the
current chunk metadata to avoid repeated allocation-id lookups.

Resolve the data pointer and release bookkeeping together during
extraction, then free only after reading all fields. This avoids a
second chunk lookup without reading an allocation after freeing it.

Recorded buildprof results:
- Wall time: 30.52s -> 29.57s (-3.11%).
- User instructions: 455.313G -> 447.673G (-1.68%).


src/trace_processor/sorter/trace_token_buffer.cc
sashwinbalaji, line 357: should we have early return for erased = 0 ? Also since this func is the only one caller of InvalidateCurrentChunk maybe inline it and remove the func InvalidateCurrentChunk altogether ?

sashwinbalaji (approved):

PR #7365, opened 2026-09-08
tp: handle track event argument fields through extensions

LalitMaganti:
Give each argument field the same handler interface as extension fields.
Move built-in handlers out of the event importer and port the in-tree
plugins, making field ownership explicit before changing reflection.

Recorded buildprof results:
- Wall time: 29.57s -> 29.21s (-1.20%).
- User instructions: 447.673G -> 445.418G (-0.50%).

### DIFF
--- a/src/trace_processor/plugins/android_job_scheduler/android_job_scheduler_tracker.cc
+++ b/src/trace_processor/plugins/android_job_scheduler/android_job_scheduler_tracker.cc
@@ -46,14 +46,15 @@ AndroidJobSchedulerTracker::AndroidJobSchedulerTracker(
   RegisterTrackEventExtension(kJobSchedulerJobExtensionFieldId);
 }
 
-TrackEventExtensionParser::Result
-AndroidJobSchedulerTracker::OnTrackEventSliceExtension(
+TrackEventExtensionParser::Result AndroidJobSchedulerTracker::OnTrackEventField(
     const TrackEventExtensionField& field,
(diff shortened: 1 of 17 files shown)


src/trace_processor/plugins/android_job_scheduler/android_job_scheduler_tracker.cc
```diff
@@ -70,7 +71,7 @@ AndroidJobSchedulerTracker::OnTrackEventSliceExtension(
     }
   }
 
-  int64_t ts = trace_context_->storage->slice_table()[slice_id].ts();
+  int64_t ts = event.ts;
```
sashwinbalaji, line 74: this is a behavior change right, like earlier this used to be start timestamp but now we are going to use the end one ?

sashwinbalaji (approved):

src/trace_processor/plugins/android_job_scheduler/android_job_scheduler_tracker.cc
LalitMaganti: event.ts is still the begin timestamp? am I missing something?

sashwinbalaji: ahh ignore, I was thinking about the case of slice end event but that wont apply here.

sashwinbalaji (approved):

PR #7368, opened 2026-09-08
tp: stdlib: Add frame_event_time to android_input_events

ivan-chong:
The event time used in the frame can differ from the actual event time of the input event due to resampling. As a result, this PR explicitly partitions the frame event time to its own field. We then ensure that the exact "event_time" field is always populated with the real event time from InputReader which is important for event correlation.

Drive-by: Replace the nested str_split with regexp_extract in _input_read_time

Bug: 454761692

### DIFF
--- a/src/trace_processor/perfetto_sql/stdlib/android/input.sql
+++ b/src/trace_processor/perfetto_sql/stdlib/android/input.sql
@@ -70,13 +70,15 @@ WITH
   _extracted_input_read_args AS (
     SELECT
       name,
-      str_split(str_split(str_split(name, 'id=', 1), ',', 0), ')', 0) AS input_event_id,
-      ts AS read_time
+      regexp_extract(name, 'id=([^,\)]+)') AS input_event_id,
+      ts AS read_time,
+      cast_int!(regexp_extract(name, 'eventTime=(\d+)')) AS event_time
     FROM slice
     WHERE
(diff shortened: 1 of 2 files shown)


src/trace_processor/perfetto_sql/stdlib/android/input.sql
```diff
@@ -70,13 +70,15 @@ WITH
   _extracted_input_read_args AS (
     SELECT
       name,
-      str_split(str_split(str_split(name, 'id=', 1), ',', 0), ')', 0) AS input_event_id,
-      ts AS read_time
+      regexp_extract(name, 'id=([^,\)]+)') AS input_event_id,
+      ts AS read_time,
+      cast_int!(regexp_extract(name, 'eventTime=([^,\)]+)')) AS event_time
```
vishniakou: since the eventTime supposed to be just digits (i think?), should this be something like "'eventTime=(\\d+)'" instead?

vishniakou (approved):

src/trace_processor/perfetto_sql/stdlib/android/input.sql
ivan-chong: Good point, unless the eventTime is somehow injected or modified to be negative they should just be digits - I've updated this now, thanks!

ivan-chong:
Friendly ping, thanks!

LalitMaganti (approved):

PR #7370, opened 2026-09-08
perfetto: expose flush reason and flags in shared lib C ABI

parthsane:
Expose the flush reason (PerfettoDsFlushReason) and raw 64-bit flush flags via PerfettoDsOnFlushArgsGetReason() and
PerfettoDsOnFlushArgsGetFlags() in the public shared library C ABI.

This allows producers (such as the Android platform Perfetto data source bindings) to distinguish between periodic flushes, session stops, and clone snapshots (kTraceClone).

### DIFF
--- a/include/perfetto/public/abi/data_source_abi.h
+++ b/include/perfetto/public/abi/data_source_abi.h
@@ -127,6 +127,7 @@ typedef void (*PerfettoDsOnDestroyCb)(struct PerfettoDsImpl*,
                                       void* inst_ctx);
 
 // Opaque handle used to perform operations from the OnFlush callback.
+// Pointers to this struct are valid only during the OnFlush callback.
 struct PerfettoDsOnFlushArgs;
 
 // Opaque handle used to signal when the data source flush operation is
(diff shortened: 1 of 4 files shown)


include/perfetto/public/abi/data_source_abi.h
```diff
@@ -144,6 +144,22 @@ PerfettoDsOnFlushArgsPostpone(struct PerfettoDsOnFlushArgs*);
...
+  PERFETTO_DS_FLUSH_REASON_EXPLICIT = 4,
+};
+
+// Returns the reason for the flush operation (see enum PerfettoDsFlushReason).
+PERFETTO_SDK_EXPORT enum PerfettoDsFlushReason PerfettoDsOnFlushArgsGetReason(
+    struct PerfettoDsOnFlushArgs*);
+
+// Returns the raw 64-bit flush flags.
+PERFETTO_SDK_EXPORT uint64_t
+PerfettoDsOnFlushArgsGetFlags(struct PerfettoDsOnFlushArgs*);
```
zezeozue: Any need for this?

zezeozue (approved):

include/perfetto/public/abi/data_source_abi.h
```diff
@@ -144,6 +144,18 @@ PerfettoDsOnFlushArgsPostpone(struct PerfettoDsOnFlushArgs*);
...
+enum PerfettoDsFlushReason {
+  PERFETTO_DS_FLUSH_REASON_UNKNOWN = 0,
+  PERFETTO_DS_FLUSH_REASON_PERIODIC = 1,
+  PERFETTO_DS_FLUSH_REASON_TRACE_STOP = 2,
+  PERFETTO_DS_FLUSH_REASON_TRACE_CLONE = 3,
+  PERFETTO_DS_FLUSH_REASON_EXPLICIT = 4,
+};
+
+// Returns the reason for the flush operation (see enum PerfettoDsFlushReason).
+PERFETTO_SDK_EXPORT enum PerfettoDsFlushReason PerfettoDsOnFlushArgsGetReason(
```
sashwinbalaji: 1) would it be safer if the return type was uint64_t instead of enum similar to how clock id is already handled in this file to avoid depending on compiler for the actual type.
2) nit: also add a comment adding args is valid during onflush only

parthsane: Done.

include/perfetto/public/abi/data_source_abi.h
```diff
@@ -144,6 +145,19 @@ PerfettoDsOnFlushArgsPostpone(struct PerfettoDsOnFlushArgs*);
 // PerfettoDsOnFlushArgsPostpone).
 PERFETTO_SDK_EXPORT void PerfettoDsFlushDone(struct PerfettoDsAsyncFlusher*);
 
+enum PerfettoDsFlushReason {
```
sashwinbalaji, line 151: since this is the ABI file pls also add a comment (sorry for not adding this earlier) that this mirrors FlushFlags::Reason and should be kept aligned with http://ac/external/perfetto/include/perfetto/tracing/core/flush_flags.h?l=40

sashwinbalaji (approved):

PR #7375, opened 2026-09-08
tp: avoid redundant packet reader work

LalitMaganti:
Detect compressed bundles using the reader's existing decoder and read
contiguous packet headers directly. Retain validated sequence lookups
and inline the common single-chunk reader operations to reduce
per-packet work.

Move-slice blob views when consuming bytes so ownership transfers
without refcount updates or a separate consumed-byte offset. Keep
individual packet field checks instead of maintaining a second inventory
of rare fields.

On the interned buildprof trace, against its parent:
- Wall time: 21.55s -> 20.53s (-4.7%).
- User instructions: 323.492G -> 293.280G (-9.34%).
- Executable .text: 8,905,154 -> 8,906,946 bytes (+1,792).

### DIFF
--- a/src/trace_processor/importers/proto/selective_trace_packet_decoder.h
+++ b/src/trace_processor/importers/proto/selective_trace_packet_decoder.h
@@ -165,9 +165,16 @@ class SelectiveTracePacketDecoder {
   bool has_compressed_packets() const {
     return decoder_.at<TracePacket::kCompressedPacketsFieldNumber>().valid();
   }
+  protozero::ConstBytes compressed_packets() const {
+    return decoder_.at<TracePacket::kCompressedPacketsFieldNumber>().as_bytes();
+  }
   bool has_zstd_compressed_packets() const {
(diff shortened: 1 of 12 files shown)


src/trace_processor/importers/proto/selective_trace_packet_decoder.h
```diff
@@ -226,6 +233,12 @@ class SelectiveTracePacketDecoder {
   // Whether the packet was decoded to its end.
   size_t bytes_left() const { return decoder_.bytes_left(); }
 
+  // True if any field whose bit is set in |mask| is present. See
+  // TypedProtoDecoderBase::HasAnyField().
+  bool HasAnyField(const uint64_t* mask, size_t words) const {
```
sashwinbalaji: not used ?

sashwinbalaji (approved):
SerializingProtoTraceReader also needs update fialing CI tests)

PR #7377, opened 2026-09-08
tp: reduce setup on the ordinary track event import path

LalitMaganti:
Construct argument writers lazily and share the empty legacy decoder.
Handle single interned categories directly and keep uncommon legacy and
correlation handling out of line. Cache the last field handler,
invalidating the lookup when another handler is registered.

On the interned buildprof trace, against its parent:
- Wall time: 19.66s -> 18.53s (-5.7%).
- User instructions: 289.388G -> 278.381G (-3.80%).
- Executable .text: 8,907,474 -> 8,910,578 bytes (+3,104).

### DIFF
--- a/src/trace_processor/importers/proto/track_event_event_importer.h
+++ b/src/trace_processor/importers/proto/track_event_event_importer.h
@@ -25,9 +25,11 @@
 #include <utility>
 #include <vector>
 
+#include "perfetto/base/compiler.h"
 #include "perfetto/base/logging.h"
 #include "perfetto/base/status.h"
 #include "perfetto/ext/base/dynamic_string_writer.h"
+#include "perfetto/ext/base/no_destructor.h"
 #include "perfetto/ext/base/status_macros.h"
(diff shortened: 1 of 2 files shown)


src/trace_processor/importers/proto/track_event_event_importer.h
```diff
@@ -234,18 +254,8 @@ class TrackEventEventImporter {
...
-              protos::pbzero::InternedData::kEventCategoriesFieldNumber, iid)) {
-        category_id = *id;
-      } else {
-        base::DynamicStringWriter writer;
-        writer.AppendLiteral("unknown(");
-        writer.AppendUnsignedInt(iid);
-        writer.AppendChar(')');
-        category_id = storage_->InternString(writer.GetStringView());
-      }
+    if (!has_multiple_categories && category_iids) {
```
sashwinbalaji: won't this case be covered by ParseSingleInternedCategory above?

sashwinbalaji (approved):

PR #7384, opened 2026-09-09
ui: fix tree explorer queries after disposal

sashwinbalaji:
- Stop queued fetches when disposal starts.
- Recheck disposal after awaiting the metric table.
- Clean up tables created during disposal.
- Add regression tests for disposal races and cleanup.

Bug: 558973171

### DIFF
--- a/ui/src/components/tree_explorer_fetcher.ts
+++ b/ui/src/components/tree_explorer_fetcher.ts
@@ -186,6 +186,7 @@ export class TreeExplorerFetcher implements AsyncDisposable {
     SharedAsyncDisposable<AsyncDisposable>
   >;
   private readonly metricTables: MetricTable[] = [];
+  private isDisposed = false;
 
   constructor(
     private readonly trace: Trace,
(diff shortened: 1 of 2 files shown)


ui/src/components/tree_explorer_fetcher.ts
```diff
@@ -204,12 +207,15 @@ export class TreeExplorerFetcher implements AsyncDisposable {
...
   // Fetches the tree for the metric selected in `state`, with all of
-  // `state`'s filters and view applied. Returns undefined if the fetcher's
-  // dependencies were disposed while the fetch was in flight.
+  // `state`'s filters and view applied. Returns undefined if this fetcher was
+  // disposed before the tree query was issued.
   async fetch(
     metrics: ReadonlyArray<TreeExplorerQueryMetric>,
     state: TreeExplorerState,
   ): Promise<TreeExplorerData | undefined> {
+    if (this.isDisposed) {
```
stevegolton, line 216: I feel like a lot of this can be handled with an asyncmemo, as we've done elsewhere in the UI, but I don't really understand the code... having a look

sashwinbalaji: Just was trying out minimal possible change to fix the crash. Happy to drop/edit the change altogether.

PR #7385, opened 2026-09-09
tp: support thread-scoped PMU counters in perf importer

TudorCretu:
Previously, trace processor only created per-CPU (`perf_cpu_counter`) and global (`perf_global_counter`) perf counter tracks. In multi-threaded `perf` and `simpleperf` traces (e.g. `simpleperf record -p <pid> --add-counter cpu-cycles,instructions`), hardware PMU counters collected in read groups (`PERF_FORMAT_GROUP`) are opened per `(tid, cpu)` or per `tid`. Interleaving cumulative counter readouts from multiple threads onto a single shared per-CPU track corrupted counter values across context switches and tripped `PERFETTO_CHECK(count >= last_count_)` in `PerfCounter::AddCount`.

This PR models all four `perf_event_open(attr, pid, cpu, ...)` counter scoping modes in the perf importer (`CounterScope::{kGlobal, kCpu, kThread, kThreadAndCpu}`), extends `perf_counter_track` in the SQL stdlib with a `utid` column, and renders thread-scoped hardware counters beneath their respective thread rows in the UI.

## Commits

1. **tp: support thread-scoped PMU counters in perf importer**
   - Models the four kernel `perf_event_open` counter scopes via `CounterScope` in `PerfEventAttr` and `RecordParser::UpdateCountersInReadGroups`:
     - `kGlobal`: global counters (`perf_global_counter`).
     - `kCpu` (`pid == -1, cpu >= 0`): per-CPU counters (`perf_cpu_counter`).
     - `kThread` (`pid >= 0, cpu == -1`): per-thread CPU-unbound counters (`perf_thread_counter`).
     - `kThreadAndCpu` (`pid >= 0, cpu >= 0`): per-thread-and-CPU counters (`perf_thread_cpu_counter`), where each `(utid, cpu)` stream corresponds to a single kernel `event_id` (`PERFETTO_CHECK(count >= last_count_)` in `PerfCounter::AddCount` remains unchanged from `main`).
   - Parses `PERF_RECORD_ID_INDEX` (`69`) in `PerfDataTokenizer::ProcessIdIndexRecord` to determine `CounterScope` directly from `{id, idx, cpu, tid}` entries in Linux `perf.data` files.
   - Because `perf_event_attr` does not store the target `pid`/`cpu` passed to `perf_event_open` and Android `simpleperf` omits `PERF_RECORD_ID_INDEX`, `PerfInvocation::SetSimpleperfMetaInfo` uses `system_wide_collection` (`"true"` vs `"false"`) and `app_package_name` from `FEATURE_SIMPLEPERF_META_INFO` as the fallback when `PERF_RECORD_ID_INDEX` is absent.
   - Extends `perf_counter_track` in `counters.sql` with `utid JOINID(thread.id)` (`extract_arg(ct.dimension_arg_set_id, utid)`) and includes `'perf_thread_counter'` and `'perf_thread_cpu_counter'`.

2. **ui: show thread-scoped perf counters under their thread**
   - Registers `perf_thread_counter` and `perf_thread_cpu_counter` in `COUNTER_TRACK_SCHEMAS` under `topLevelGroup: 'THREAD'` with `mode: 'delta'` so thread-scoped hardware counters render beneath their thread row.
   - Filters `dev.perfetto.LinuxPerf` top-level `'Perf counters'` registration to `where utid is null` so CPU/global tracks remain under `'Perf counters'` while thread-scoped tracks are handled by `TraceProcessorTrack`.

## UI

**Thread-scoped PMU counters (`cpu-cycles`, `raw-inst-retired`) grouped beneath each thread's tracks (`mode: 'delta'`):**
Screenshot 2026-09-14 2 12 10 PM

**Backwards compatibility on CPU-scoped `perf_with_add_counter.data` (`dev.perfetto.LinuxPerf`):**
Screenshot 2026-09-14 2 14 19 PM


src/trace_processor/importers/perf/perf_counter.cc
```diff
@@ -16,20 +16,62 @@
...
   return counter_table_.Insert({ts, track_id_, last_count_}).id;
 }
 
-tables::CounterTable::Id PerfCounter::AddCount(int64_t ts, double count) {
-  PERFETTO_CHECK(count >= last_count_);
+tables::CounterTable::Id PerfCounter::AddCount(int64_t ts,
+                                               double count,
+                                               std::optional<uint32_t> cpu) {
+  // On heterogeneous multi-core architectures (e.g. big.LITTLE), cluster PMU
+  // readouts can experience baseline shifts across core migrations. Track
```
kraskevich: I'm not sure I understand what this means.

src/trace_processor/importers/perf/record_parser.cc
```diff
@@ -293,6 +301,55 @@ base::Status RecordParser::ParseComm(Record record) {
   return base::OkStatus();
 }
 
+base::Status RecordParser::ParseFork(int64_t, Record record) {
+  Reader reader(record.payload.copy());
+  uint32_t pid;
+  uint32_t ppid;
```
kraskevich: parent_pid?

src/trace_processor/importers/perf/record_parser.cc
```diff
@@ -293,6 +301,55 @@ base::Status RecordParser::ParseComm(Record record) {
...
+  uint32_t tid;
+  uint32_t ptid;
```
kraskevich: The same: parent_tid?

src/trace_processor/importers/perf/perf_event_attr.cc
```diff
@@ -152,18 +152,30 @@ PerfEventAttr::PerfEventAttr(TraceProcessorContext* context,
 
 PerfEventAttr::~PerfEventAttr() = default;
 
-PerfCounter& PerfEventAttr::GetOrCreateCounter(std::optional<uint32_t> cpu) {
-  if (!cpu) {
-    if (!global_counter_) {
-      global_counter_ = std::make_unique<PerfCounter>(CreateGlobalCounter());
+PerfCounter& PerfEventAttr::GetOrCreateCounter(std::optional<uint32_t> cpu,
+                                               std::optional<UniqueTid> utid) {
+  if (is_thread_scoped() && utid.has_value()) {
```
kraskevich, line 157: When we do system wide recording, don't we still have thread ids in samples? If so, why not use them? 

Real example for Chrome analysis: I record system wide PMUs for a loadline run. I want to know individual values for different Chrome threads, e.g. Browser vs GPU vs Renderer.

src/trace_processor/importers/perf/record_parser.cc
TudorCretu: Done. Renamed to parent_tid.

src/trace_processor/importers/perf/record_parser.cc
TudorCretu: Done. Renamed to parent_pid.

src/trace_processor/importers/perf/features.cc
```diff
@@ -114,7 +114,7 @@ bool ParseBuildId(const perf_event_header& header,
 base::Status ParseEventTypeInfo(std::string value, SimpleperfMetaInfo& out) {
   for (const auto& line : base::SplitString(value, "\n")) {
     auto tokens = base::SplitString(line, ",");
-    if (tokens.size() != 3) {
+    if (tokens.size() < 3) {
```
kraskevich: I believe I know what it is but I'd put it into a separate CL. This is orthogonal to multi thread PMUs.

kraskevich (approved):

src/trace_processor/importers/perf/features.cc
TudorCretu: Done!

src/trace_processor/importers/perf/perf_counter.cc
TudorCretu: Updated, thanks!

LalitMaganti:
This will need time to digest for me. Mainly I would need to research to understand how thread scoped counters work in perf to see if this is even the right approach at all. As a perf expert, I would also like @rsavitski to look at this to understand whether this is correct from a perf perspective

src/trace_processor/importers/common/tracks_common.h
```diff
@@ -296,6 +296,15 @@ inline constexpr auto kPerfCpuCounterBlueprint = tracks::CounterBlueprint(
         tracks::kNameFromTraceDimensionBlueprint),
     tracks::DynamicNameBlueprint());
 
+inline constexpr auto kPerfThreadCounterBlueprint = tracks::CounterBlueprint(
```
LalitMaganti, line 299: To my understanding, in general, perf has three modes of operation:
1) Per cpu (the mode we support today)
2) Per thread (what you are tyring to model here *but* inheritence does not work)
3) Per thread *and* cpu (what I think you are actually doing but not modelling correctly)

I would expect to see three track types here correspondingly but I only see 1 and 2. Where is 3 and why is it missing?

src/trace_processor/importers/perf/perf_counter.cc
```diff
@@ -16,20 +16,65 @@
...
   last_count_ += delta;
   return counter_table_.Insert({ts, track_id_, last_count_}).id;
 }
 
-tables::CounterTable::Id PerfCounter::AddCount(int64_t ts, double count) {
-  PERFETTO_CHECK(count >= last_count_);
+tables::CounterTable::Id PerfCounter::AddCount(int64_t ts,
+                                               double count,
+                                               std::optional<uint32_t> cpu) {
+  // In thread-targeted mode, perf_event_open creates a separate file
```
LalitMaganti: So there are two problems with this:
1) this is only true *if* you set inherit. There are ways to use perf/simpleperf which do *not* do this and instead truly migrate the counter between cpus and emit one per thread
2) In my opinion, in trace procesor our job is to accurately model what the perf.data file shows. If it is emitting events per-thread *and* per-cpu we should moel it as such not force it to fit in the "per-thread" model just becasue it is convinient.

src/trace_processor/importers/perf/perf_event_attr.cc
```diff
@@ -152,18 +152,38 @@ PerfEventAttr::PerfEventAttr(TraceProcessorContext* context,
...
+    global_counter_ = std::make_unique<PerfCounter>(CreateGlobalCounter());
   }
-  auto it = counters_.find(*cpu);
-  if (it == counters_.end()) {
-    it = counters_.emplace(*cpu, CreateCpuCounter(*cpu)).first;
+  return *global_counter_;
+}
+
+PerfCounter& PerfEventAttr::GetOrCreateThreadCounter(UniqueTid utid) {
+  if (auto* it = thread_counters_.Find(utid)) {
```
LalitMaganti: As above: where is per-cpu *and* per-thread?

src/trace_processor/importers/perf/perf_event_attr.h
```diff
@@ -98,9 +98,30 @@ class PerfEventAttr : public RefCounted {
...
   }
 
+  bool is_thread_scoped() const {
+    if (is_system_wide_) {
+      return false;
+    }
+    if (is_thread_scoped_override_.has_value()) {
+      return *is_thread_scoped_override_;
+    }
+    return attr_.inherit != 0;
```
LalitMaganti: I don't think this precludes it being thread scoped...

src/trace_processor/importers/perf/perf_invocation.cc
```diff
@@ -230,6 +230,32 @@ std::optional<BuildId> PerfInvocation::LookupBuildId(
...
+  bool is_thread_scoped = false;
+  for (const auto& arg : args) {
+    if (arg == "--") {
+      break;
+    }
+    if (arg == "-a" || arg == "--all-cpus" || arg == "--system-wide") {
+      is_system_wide = true;
+      break;
+    }
+    if (arg == "-p" ||
```
LalitMaganti: This seems like *very* hacky way to determine whether we're in thread scoped mode or not. Why are we not able to figure it out from the attrs themselves? Really not a fan of this.

src/trace_processor/importers/perf/perf_invocation_unittest.cc
```diff
@@ -364,5 +376,532 @@ TEST(PerfInvocationTest, SimpleperfMetaInfoEventTypeInfoWithExtraFields) {
   EXPECT_EQ(*name, "cpu-cycles");
 }
 
+class PerfThreadScopedCounterTest : public ::testing::Test {
```
LalitMaganti: In general in Perfetto we prefer writing diff tests to unittests like this. Please turn these into diff tests.

src/trace_processor/importers/perf/record_parser.cc
```diff
@@ -117,7 +117,14 @@ base::Status RecordParser::ParseRecord(int64_t ts, Record record) {
     case PERF_RECORD_COMM:
       return ParseComm(std::move(record));
 
+    case PERF_RECORD_FORK:
```
LalitMaganti, line 120: IMO the fork/exit handling is orthogonal to this PR and should be split out.

TudorCretu: Sure! Removed ParseFork and ParseExit from this PR and moved them to a separate branch for a standalone PR.

src/trace_processor/importers/perf/perf_invocation_unittest.cc
TudorCretu: Got it, I've turned these into diff tests

src/trace_processor/importers/common/tracks_common.h
TudorCretu: Good point, many thanks! I added kPerfThreadCpuCounterBlueprint alongside kPerfCpuCounterBlueprint and kPerfThreadCounterBlueprint

src/trace_processor/importers/perf/perf_event_attr.cc
TudorCretu: Thanks, added `GetOrCreateThreadCpuCounter(UniqueTid utid, uint32_t cpu)` and `CreateThreadCpuCounter` here.

src/trace_processor/importers/perf/perf_invocation.cc
TudorCretu: Thanks! I had an incorrect understanding and despite not liking it either, I didn't know how else to go about this. Many thanks for the comments and clarifications! I'm not changing SetCmdline anymore

src/trace_processor/importers/perf/perf_event_attr.h
TudorCretu: Got it, I separated `is_thread_scoped()` and `is_cpu_bound()` accordingly.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/counter_tracks.ts
```diff
@@ -298,6 +298,12 @@ export const COUNTER_TRACK_SCHEMAS: ReadonlyArray<CounterTrackTypeSchema> = [
     topLevelGroup: 'PROCESS',
     group: undefined,
   },
+  {
+    type: 'perf_thread_counter',
```
LalitMaganti, line 302: There's no process counter here?

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/counters.sql
```diff
@@ -252,11 +253,12 @@ FROM counter_track AS ct
 JOIN args
   ON ct.dimension_arg_set_id = args.arg_set_id
 WHERE
-  args.key = 'utid';
+  args.key = 'utid'
+  AND ct.type != 'perf_thread_cpu_counter';
```
LalitMaganti: Why are you removing this?

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/counters.sql
```diff
@@ -305,6 +307,51 @@ FROM counter_track AS ct
 WHERE
   ct.type IN ('perf_cpu_counter', 'perf_global_counter');
```
LalitMaganti: Shouldn't we modify this to include utid as well? And also the two new track types you're adding

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/counters.sql
```diff
@@ -305,6 +307,51 @@ FROM counter_track AS ct
...
 
+-- Tracks containing thread-scoped counter-like events collected from Linux perf.
+CREATE PERFETTO TABLE perf_thread_counter_track(
```
LalitMaganti: Is there a reason to have this table or is adding to perf_counter_track not enough?

src/trace_processor/importers/perf/perf_counter.cc
TudorCretu: Many thanks for the pointers and clarifications. I greatly appreciate it that you took the time to review and clarify in detail my wrong assumptions. I reverted back the check and I'm routing the `(utid, cpu)` streams to `perf_thread_cpu_counter` to keep each counter track monotonic without needing per-cpu tracking here.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/counters.sql
```diff
@@ -111,7 +111,8 @@ FROM counter_track AS ct
 JOIN args
   ON ct.dimension_arg_set_id = args.arg_set_id
 WHERE
-  args.key = 'cpu';
+  args.key = 'cpu'
+  AND ct.type != 'perf_thread_cpu_counter';
```
LalitMaganti: Why are you removing this?

src/trace_processor/importers/perf/perf_event_attr.cc
```diff
@@ -152,18 +152,42 @@ PerfEventAttr::PerfEventAttr(TraceProcessorContext* context,
 
 PerfEventAttr::~PerfEventAttr() = default;
 
-PerfCounter& PerfEventAttr::GetOrCreateCounter(std::optional<uint32_t> cpu) {
-  if (!cpu) {
-    if (!global_counter_) {
-      global_counter_ = std::make_unique<PerfCounter>(CreateGlobalCounter());
+PerfCounter& PerfEventAttr::GetOrCreateCpuCounter(std::optional<uint32_t> cpu) {
+  if (cpu.has_value()) {
+    if (auto* it = cpu_counters_.Find(*cpu)) {
```
LalitMaganti: You can avoid the find + instead for more performance. Do an insert with null and then use the boolean to decide whether to create the CPU counter.

src/trace_processor/importers/perf/perf_event_attr.cc
```diff
@@ -152,18 +152,42 @@ PerfEventAttr::PerfEventAttr(TraceProcessorContext* context,
...
   }
-  return it->second;
+  return *global_counter_;
+}
+
+PerfCounter& PerfEventAttr::GetOrCreateThreadCounter(UniqueTid utid) {
+  if (auto* it = thread_counters_.Find(utid)) {
+    return **it;
+  }
+  return **thread_counters_
```
LalitMaganti: As above

src/trace_processor/importers/perf/perf_invocation.cc
```diff
@@ -238,6 +257,58 @@ void PerfInvocation::SetCmdline(const std::vector<std::string>& args) {
   }
 }
 
+void PerfInvocation::SetSimpleperfMetaInfo(
+    const base::FlatHashMap<std::string, std::string>& entries) {
+  // If profiling a specific app (e.g. Chrome via --app), simpleperf records
```
LalitMaganti: Do we actually need this? Why is using the attrs not sufficient?

src/trace_processor/importers/perf/perf_event_attr.h
```diff
@@ -98,9 +98,42 @@ class PerfEventAttr : public RefCounted {
...
+  bool is_cpu_bound() const {
+    if (is_cpu_bound_override_.has_value()) {
+      return *is_cpu_bound_override_;
+    }
+    return num_ids_ > 1;
+  }
+
+  void set_num_ids(size_t num_ids) { num_ids_ = num_ids; }
+
+  void set_is_system_wide(bool is_system_wide) {
```
LalitMaganti: I'm not sure I understand the need for all these separate functions. I would expect to see an enum here modelling the four types (global, cpu, thread, CPU and thread) rather than all these booleans.

src/trace_processor/importers/perf/perf_event_attr.h
```diff
@@ -123,13 +158,26 @@ class PerfEventAttr : public RefCounted {
   std::optional<size_t> id_offset_from_start_;
   std::optional<size_t> id_offset_from_end_;
   size_t sample_id_size_;
-
-  // Counter not bound to a specific CPU. Might actually be an aggregate from
-  // multiple per-cpu counters in cases such as thread-scoped profiling.
+  size_t num_ids_ = 0;
+  bool is_system_wide_ = false;
```
LalitMaganti: Why do we need is_system_wide at all? Shouldn't the attr tell you everything you need to know? This seems a lot more complex that I would (naively) expect.

src/trace_processor/importers/perf/perf_event_attr.h
TudorCretu: Thanks! Replaced the booleans with enum class CounterScope { kGlobal, kCpu, kThread, kThreadAndCpu } and unified the lookup behind GetOrCreateCounter(cpu, utid).

src/trace_processor/importers/perf/perf_event_attr.cc
TudorCretu: Done, switched to Insert.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/counters.sql
TudorCretu: Removed my change here.

src/trace_processor/importers/perf/perf_event_attr.cc
TudorCretu: Done

src/trace_processor/importers/perf/perf_event_attr.h
TudorCretu: Removed is_system_wide_ and the other override flags in favor of CounterScope. Thanks!

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/counters.sql
TudorCretu: Done, added utid and the two new track types (perf_thread_counter and perf_thread_cpu_counter) to perf_counter_track.

src/trace_processor/perfetto_sql/stdlib/prelude/after_eof/counters.sql
TudorCretu: Removed my change here.

src/trace_processor/importers/perf/perf_invocation.cc
TudorCretu: If I understand correctly, the `perf_event_attr` struct in `perf.data` ends up identical between `simpleperf record -p <pid> --add-counter` and `simpleperf record -a --add-counter` (since the target pid/cpu is passed to `perf_event_open` rather than stored in attr, and inherit is 0 in both cases). Simpleperf only records whether it was system-wide or target-scoped in `SIMPLEPERF_META_INFO`.

ui/src/plugins/dev.perfetto.TraceProcessorTrack/counter_tracks.ts
TudorCretu: If I understand correctly, perf / simpleperf always attach counters to individual threads even when recording with -p <pid>, so we only get per-thread counter values in the samples rather than process-wide ones.

src/trace_processor/importers/perf/perf_invocation.cc
LalitMaganti: That is disappointing to hear. I really don['t like how simpleper has made these ad-hoc differences to perf in an inconsistent way.

LalitMaganti (approved):

src/trace_processor/importers/perf/perf_event_attr.h
```diff
@@ -123,13 +140,22 @@ class PerfEventAttr : public RefCounted {
...
 
-  // Counter not bound to a specific CPU. Might actually be an aggregate from
-  // multiple per-cpu counters in cases such as thread-scoped profiling.
-  std::unique_ptr<PerfCounter> global_counter_;
+  // Keyed by utid. Per-thread counters (pid >= 0, cpu == -1).
+  base::FlatHashMap<UniqueTid, std::optional<PerfCounter>> thread_counters_;
 
-  // Keyed by cpu index. Per-cpu counters.
-  std::unordered_map<uint32_t, PerfCounter> counters_;
+  // Keyed by (utid << 32) | cpu. Per-(thread, cpu) counters (pid >= 0, cpu >=
```
rsavitski: nit: is there a significant reason for this bitpacking? Afaiu `std::pair<UniqueTid, uint32_t>` key should work just as well, and is a bit more self-documenting (and won't require readers to double-check that UniqueTid is 32 bits).

src/trace_processor/importers/perf/perf_invocation.cc
```diff
@@ -238,6 +254,35 @@ void PerfInvocation::SetCmdline(const std::vector<std::string>& args) {
   }
 }
 
+void PerfInvocation::SetSimpleperfMetaInfo(
```
rsavitski: nit: this seems quite important for simpleperf, but the function name does not highlight that importance (it reads as if it just sets some metadata).

I'd rename this to `SetSimpleperfCounterScope()` or similar, so that its purpose is more obvious at the callsite.

src/trace_processor/importers/perf/record_parser.cc
rsavitski: does this need to pass `utid` as well?

src/trace_processor/importers/perf/perf_event_attr.cc
```diff
@@ -148,22 +148,62 @@ PerfEventAttr::PerfEventAttr(TraceProcessorContext* context,
       time_offset_from_end_(TimeOffsetFromEndOfNonSampleRecord(attr_)),
       id_offset_from_start_(IdOffsetFromStartOfSampleRecord(attr_)),
       id_offset_from_end_(IdOffsetFromEndOfNonSampleRecord(attr_)),
-      sample_id_size_(GetSampleIdSize(attr_)) {}
+      sample_id_size_(GetSampleIdSize(attr_)),
+      counter_scope_((attr_.sample_type & PERF_SAMPLE_CPU) != 0
```
rsavitski: Do we absolutely need to speculate on the initial scope, instead of leaving it indeterminate until concrete knowledge from PERF_RECORD_ID_INDEX? Or is this needed for backwards compatiblity?

It's a bit weird to rely on perf_event_attr contents since it's technically possible to do per-cpu sampling without sampling the cpu number itself.

src/trace_processor/importers/perf/perf_event_attr.cc
```diff
@@ -148,22 +148,62 @@ PerfEventAttr::PerfEventAttr(TraceProcessorContext* context,
...
 
 PerfEventAttr::~PerfEventAttr() = default;
 
-PerfCounter& PerfEventAttr::GetOrCreateCounter(std::optional<uint32_t> cpu) {
-  if (!cpu) {
-    if (!global_counter_) {
-      global_counter_ = std::make_unique<PerfCounter>(CreateGlobalCounter());
+PerfCounter& PerfEventAttr::GetOrCreateCounter(std::optional<uint32_t> cpu,
+                                               std::optional<UniqueTid> utid) {
+  if (!utid.has_value()) {
```
rsavitski: This part looks odd, why only utid optionality is special? If it's for backwards compat, then a short one-line comment would help future readers.

test/trace_processor/diff_tests/parser/simpleperf/perf_test.sql
```diff
@@ -32,7 +32,7 @@ FROM counter_delta AS c, perf_counter_track AS t
 ON c.track_id = t.id,
 perf_sample AS s
 ON c.ts = s.ts AND t.perf_session_id = s.perf_session_id AND t.cpu = s.cpu,
-thread USING (utid),
+thread ON s.utid = thread.utid,
```
rsavitski, line 35: Not blocking: hmm, a bit worried whether other existing queries with USING(utid) will now silently prefer perf_counter_track.utid. Probably fine as is unless Lalit objects.

src/trace_processor/importers/perf/perf_event_attr.h
TudorCretu: Done, switched to std::pair<UniqueTid, uint32_t>

src/trace_processor/importers/perf/perf_invocation.cc
TudorCretu: Done, renamed to SetSimpleperfCounterScope

src/trace_processor/importers/perf/record_parser.cc
TudorCretu: Done, passing utid here as well

src/trace_processor/importers/perf/perf_event_attr.cc
TudorCretu: Good point, changed counter_scope_ to std::optional<CounterScope> so it stays unset until PERF_RECORD_ID_INDEX or SetSimpleperfCounterScope sets it.

src/trace_processor/importers/perf/perf_event_attr.cc
TudorCretu: Done, made utid non-optional now that UpdateCounters passes it too.

test/trace_processor/diff_tests/parser/simpleperf/perf_test.sql
TudorCretu: Ack

rsavitski (approved):

PR #7389, opened 2026-09-09
tp: add terminal-aware progress and color controls

LalitMaganti:
Share throttled stderr progress reporting across trace loading,
bundling, and conversions. Pass --no-progress explicitly through
callers, preserve ordinary diagnostics, and support FORCE_COLOR and
NO_COLOR independently of progress.


docs/learning-more/symbolization.md
```diff
@@ -112,8 +112,17 @@ add `--no-auto-symbol-paths` and `--no-auto-proguard-maps`. Native paths from
 `PERFETTO_BINARY_PATH` still apply; unset that variable to restrict lookup to
 `--symbol-paths`.
 
+For scripts that capture diagnostics, redirect stderr to a log:
```
LalitMaganti, line 115: Is this something we really want to document?

include/perfetto/ext/base/progress_reporter.h
```diff
@@ -0,0 +1,51 @@
...
+class ProgressReporter {
+ public:
+  explicit ProgressReporter(bool enabled = true);
+  ~ProgressReporter();
+  ProgressReporter(const ProgressReporter&) = delete;
+  ProgressReporter& operator=(const ProgressReporter&) = delete;
+
+  void Update(const std::string& message);
+  void Clear();
+  static void ClearBeforeLog();
```
LalitMaganti, line 41: This seems like a bit of a hack.

include/perfetto/ext/base/progress_reporter.h
```diff
@@ -0,0 +1,51 @@
...
+bool StderrSupportsProgress();
+bool StderrSupportsColor();
+
+// A single-line, throttled stderr display. Updates must be plain ASCII without
+// terminal escapes. No output is produced for redirected stderr or TERM=dumb.
+// Use Clear() before ordinary stdio diagnostics; Perfetto logs clear it
+// automatically. This class does not print a final summary.
+class ProgressReporter {
+ public:
+  explicit ProgressReporter(bool enabled = true);
```
LalitMaganti, line 34: Surely we should have a GetInstance or similar instead of letting people create it?

src/base/progress_reporter.cc
```diff
@@ -0,0 +1,133 @@
...
+  if (ioctl(STDERR_FILENO, TIOCGWINSZ, &size) == 0 && size.ws_col)
+    return size.ws_col;
+#endif
+  return 80;
+}
+}  // namespace
+
+bool StderrSupportsProgress() {
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WASM)
+  return false;
```
LalitMaganti, line 61: Hmmmm actually from what I understand WASM *does* support progress and is relied uponby the UI.

src/base/progress_reporter.cc
```diff
@@ -0,0 +1,133 @@
...
+
+bool StderrSupportsColor() {
+  const char* force_color = getenv("FORCE_COLOR");
+  if (force_color && *force_color)
+    return true;
+  const char* no_color = getenv("NO_COLOR");
+  if (no_color && *no_color)
+    return false;
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WIN)
+  // Do not emit ANSI colors without first enabling virtual terminal output.
```
LalitMaganti, line 76: Are you sure windows shouldn't have colors? Why?

src/base/progress_reporter.cc
```diff
@@ -0,0 +1,133 @@
...
+  return false;
+#elif PERFETTO_BUILDFLAG(PERFETTO_CHROMIUM_BUILD)
+  return false;
```
LalitMaganti, line 79: Why?

src/trace_processor/shell/bundle_subcommand.cc
```diff
@@ -86,7 +87,11 @@ const char* BundleSubcommand::detailed_help() const {
 
 Outputs a TAR containing the trace plus the symbols and deobfuscation
 mappings needed to make it self-contained. Both <input> and <output> must be
-real file paths (stdin/stdout are not supported).)";
+real file paths (stdin/stdout are not supported).
+
+Live progress uses stderr only when it is a terminal (except TERM=dumb).
```
LalitMaganti, line 92: Do we really need to have this in the help text? Why? Could we trim it down?

src/trace_processor/shell/common_flags.cc
```diff
@@ -123,6 +123,10 @@ std::vector<FlagSpec> GetGlobalFlagSpecs(GlobalOptions* opts) {
   flags.push_back(BoolFlag("help", 'h', "Prints this guide.", &opts->help));
   flags.push_back(
       BoolFlag("version", 'v', "Prints the version.", &opts->version));
+  flags.push_back(BoolFlag(
+      "no-progress", '\0',
+      "Disable live progress (summaries and warnings are still printed).",
```
LalitMaganti, line 128: Does the parens comment add anything?

include/perfetto/ext/base/progress_reporter.h
```diff
@@ -0,0 +1,64 @@
...
+
+// Process-wide, throttled, single-line progress display on stderr. Updates are
+// only shown when StderrSupportsProgress(). Log messages clear the line before
+// printing, so diagnostics never get merged with progress text.
+class ProgressReporter {
+ public:
+  static ProgressReporter& GetInstance();
+
+  // Suppresses all progress output. Diagnostics are unaffected.
+  void set_enabled(bool enabled) { enabled_ = enabled; }
```
sashwinbalaji: should this all be guarded by mutex ?

edit: this is set in the beginning so I guess no need but just for completeness would be good

src/base/progress_reporter.cc
```diff
@@ -0,0 +1,152 @@
...
+  last_update_ms_ = now;
+#if PERFETTO_BUILDFLAG(PERFETTO_OS_WASM)
+  // stdio reaches the embedder only when a write ends with a newline.
+  fprintf(stderr, "%s\n", message.c_str());
+#else
+  // Leave one column unused to avoid wrapping, including on narrow terminals.
+  size_t width = TerminalWidth();
+  size_t new_width = std::min(message.size(), width > 1 ? width - 1 : 0);
+  if (new_width < visible_width_)
+    ClearLocked();
```
sashwinbalaji, line 135: there is a case where we don't respect last_update_ms_

- set last_update_ms_
- then if width is simaller call ClearLocked
- ClearLocked resets last_update_ms_
- so the next update won't have a last_update_ms_ and proceed

src/trace_processor/shell/common_flags.cc
```diff
@@ -435,15 +437,16 @@ base::StatusOr<base::TimeNanos> LoadTraceFile(
...
   double size_mb = 0;
+  auto& progress = base::ProgressReporter::GetInstance();
 
-  base::Status load_status =
-      platform->LoadTrace(tp, trace_file, [&size_mb](size_t parsed_size) {
+  base::Status load_status = platform->LoadTrace(
+      tp, trace_file, [&size_mb, &progress](size_t parsed_size) {
         size_mb = static_cast<double>(parsed_size) / 1E6;
-        fprintf(stderr, "\rLoading trace: %.2f MB\r", size_mb);
+        progress.Update(
```
sashwinbalaji: would it be better to add StringView version of update

src/base/progress_reporter.cc
```diff
@@ -0,0 +1,152 @@
...
+  return true;
+#else
+  return StderrIsInteractiveTerminal();
+#endif
+}
+
+bool StderrSupportsColor() {
+  const char* force_color = getenv("FORCE_COLOR");
+  if (force_color && *force_color)
+    return true;
```
sashwinbalaji, line 96: for windows we won't invoke EnableVirtualTerminalProcessing if force color true

sashwinbalaji (approved):

include/perfetto/ext/base/progress_reporter.h
LalitMaganti: Done, and I've added the locking annotations so the compiler checks it.

src/base/progress_reporter.cc
LalitMaganti: Good catch, fixed. The timestamp is now set after clearing, with a test.

src/trace_processor/shell/common_flags.cc
LalitMaganti: Done, it takes std::string_view now.

src/base/progress_reporter.cc
LalitMaganti: Fixed, VT processing is enabled before checking the env.

PR #7390, opened 2026-09-09
tp: publish completed bundles atomically

LalitMaganti:
Write bundles through a base AtomicFile helper and publish with Commit()
&& only after the archive is complete and flushed. Preserve existing
output on failure, reject identical input/output files, and clean up
temporary files on ordinary failures without changing signal or
subprocess handling.


src/trace_processor/shell/bundle_subcommand.cc
```diff
@@ -89,6 +89,10 @@ Outputs a TAR containing the trace plus the symbols and deobfuscation
 mappings needed to make it self-contained. Both <input> and <output> must be
 real file paths (stdin/stdout are not supported).
 
+The output is replaced only after the bundle is complete. Input and output
```
LalitMaganti, line 92: Does this comment add value? I don't think it does really.

src/base/atomic_file.cc
```diff
@@ -0,0 +1,115 @@
...
+#else
+  struct stat output{};
+  bool exists = lstat(destination_.c_str(), &output) == 0;
+  PERFETTO_MSAN_UNPOISON(&output, sizeof(output));
+  if (exists && !S_ISREG(output.st_mode))
+    return base::ErrStatus(
+        "output must be a regular file, not a directory or link: '%s'. Specify "
+        "the target path directly.",
+        destination_.c_str());
+  fd_ = base::OpenFile(temp_path_, O_CREAT | O_EXCL | O_RDWR, 0644);
```
sashwinbalaji: should the permission be 0600

sashwinbalaji (approved):

PR #7418, opened 2026-09-10
ui: Use zod for plugin permalink storage

stevegolton:
This PR migrates the old mountStore (permalink storage registration mechanism) extension point to look and feel more like the settings registration system. Both are essentially doing the same thing - registering a little persistent (i.e. survives the session) bucket of where plugins can put persistent things. Settings, however, does it a lot better the mountstore.

The main difference between them is the way they handle validation - which is important when handling version skew (data saved by an older version of the UI being opened in a newer version, and vice versa) and also generally untrusted JSON from the internet in the case of permalinks. Simply asserting it to be a specific shape is dangerous and can lead to mysterious crashes later down the line. Settings registration solves this problem by mandating that plugins provide a zod schema when at registration time.

Currently, `mountStore()` takes a migrate function which is expected to take an undefined object and pass back a new object in the correct shape to be used by the plugin. While zod could simply be called in this function, a lot of plugins just don't bother and simply assert the shape is correct. Mandating a zod schema helps a lot for backward compatibility here. What'smore, with zod's .transform() function, the old migrate() function can be emulated anyway which is a nice escape hatch.

Another big difference is how to mutate data from the plugin. MountStore relies heavily on the immutable store concept which is managed by immer, so modifying some mounted is done by passing a callback which modifies the state and immer applies the changes while ensuring the underlying state remains immutable. The settings interface doesn't use store or immer, and while the state is frozen for performance and caching reasons, the immutability is kicked down the road for the plugin to handle. This is actually the more intuitive path to be honest and settings are handled just fine.

This patch migrates the mountStore call and all plugins that use it.

## Before & after

Taking `com.example.State` as a representative example:

**Before** - `mountStore()` with a hand-rolled migrate function and immer-style edits:

```ts
interface State {
  counter: number;
}

  static readonly id = 'com.example.State';
  private store: Store<State> = createStore({counter: 0});

  private migrate(initialState: unknown): State {
    if (
      exists(initialState) &&
      typeof initialState === 'object' &&
      'counter' in initialState &&
      typeof initialState.counter === 'number'
    ) {
      return {counter: initialState.counter};
    } else {
      return {counter: 0};
    }
  }

    this.store = ctx.mountStore('com.example.SkeletonStore', (init: unknown) =>
      this.migrate(init),
    );

    const counter = this.store.state.counter;
    this.store.edit((draft) => {
      ++draft.counter;
    });
  }
}
```

**After** - `registerStorage()` with a zod schema and plain get/set:

```ts
const STATE_SCHEMA = z.object({
  counter: z.number().default(0),
});

type State = z.infer<typeof STATE_SCHEMA>;

  static readonly id = 'com.example.State';

    const storage = ctx.registerStorage({
      id: 'com.example.SkeletonStore',
      schema: STATE_SCHEMA,
      defaultValue: {counter: 0},
    });

    const counter = storage.get().counter;
    storage.set({counter: counter + 1});
  }
}
```

The interface itself changes as follows:

```diff
-  // Create a store mounted over the top of this plugin's persistent state.
-  mountStore<T>(id: string, migrate: Migrate<T>): Store<T>;
+  /**
+   * Register a persistent storage container backed by the trace's permalink
+   * state.
+   *
+   * When loading a permalink, the value will be restored from the permalink if
+   * valid according to the schema, or fall back to defaultValue.
+   */
+  registerStorage<T>(descriptor: StorageDescriptor<T>): Storage<T>;
```

```ts
export interface StorageDescriptor<T> {
  readonly id: string;
  readonly schema: z.ZodType<T>;
  readonly defaultValue?: T;
  readonly name?: string;
  readonly description?: string;
}

export interface Storage<T> {
  get(): T;
  set(value: T): void;
  reset(): void;
}
```


stevegolton:
It doesn't seem particularly compact if I'm honest - perhaps you could go into more detail about the intended design goals and how it operates. 

image

LalitMaganti (approved):

PR #7419, opened 2026-09-10
perfetto_cmd: protect persistent traces against duplicate session names

safayat-google:
Use advisory file locks (flock) on persistent .tmp trace files to prevent duplicate
sessions with the same name from deleting or corrupting traces of an active session.

- Hold LOCK_EX on the persistent trace file while session is active.
- Skip unlinking and fail startup if an active session holds the lock.
- Skip reboot upload if file is locked by an active session.
- Add unit tests for duplicate session rejection and stale cleanup.

Bug: 382209797
Test: perfetto_unittests --gtest_filter="PerfettoCmdlineUnitTest.*"

### DIFF
--- a/src/perfetto_cmd/perfetto_cmd_android.cc
+++ b/src/perfetto_cmd/perfetto_cmd_android.cc
@@ -16,6 +16,7 @@
 
 #include "src/perfetto_cmd/perfetto_cmd.h"
 
+#include <sys/file.h>
 #include <sys/sendfile.h>
 #include <sys/system_properties.h>
 
@@ -204,7 +205,10 @@ void PerfettoCmd::ReportTraceToAndroidFrameworkOrCrash() {
   }
 
   if (!persistent_file_path_.empty()) {
-    PERFETTO_CHECK(!unlink(persistent_file_path_.c_str()));
+    // Explicitly release the advisory lock before unlinking the file.
+    flock(trace_fd, LOCK_UN);
+    PERFETTO_CHECK(unlink(persistent_file_path_.c_str()) == 0 ||
+                   errno == ENOENT);
   }
 }
 
@@ -284,15 +288,16 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
   return fd;
 }
 
-void PerfettoCmd::WaitForPreviousRebootTraceUpload(
+base::Status PerfettoCmd::WaitForRebootTraceUploadOrCleanup(
     const std::string& session_name,
     const std::string& target_file_path) {
   // Only block if a persistent trace file with the SAME session name exists on
   // disk.
   if (!base::FileExists(target_file_path)) {
-    return;
+    return base::OkStatus();
   }
 
+  // Wait for after reboot task to unlink traces.
   const auto deadline = base::GetBootTimeNs() + kBootTraceCleanupTimeoutNs;
   std::string cur_prop;
   do {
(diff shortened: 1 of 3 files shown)


src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -302,6 +307,29 @@ void PerfettoCmd::WaitForPreviousRebootTraceUpload(
...
 
+  if (!base::FileExists(target_file_path)) {
+    return base::OkStatus();
+  }
+
+  // File exists after waiting for reboot task. Check if an active session is
+  // holding the lock before doing any cleanup.
+  {
+    base::ScopedFile existing_fd =
+        base::OpenFile(target_file_path, O_RDWR | O_CLOEXEC);
```
primiano: base::OpenFile already takes care of CLOEXEC

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -302,6 +307,29 @@ void PerfettoCmd::WaitForPreviousRebootTraceUpload(
     }
   } while (cur_prop.empty() && base::GetBootTimeNs() < deadline);
 
+  if (!base::FileExists(target_file_path)) {
+    return base::OkStatus();
+  }
+
+  // File exists after waiting for reboot task. Check if an active session is
```
primiano: I'd reword this comment as:

// This file might be lefover by a crashed instance of perfetto_cmd

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -302,6 +307,29 @@ void PerfettoCmd::WaitForPreviousRebootTraceUpload(
...
+      if (flock(existing_fd.get(), LOCK_EX | LOCK_NB) != 0) {
+        if (errno == EWOULDBLOCK || errno == EAGAIN) {
+          return base::ErrStatus(
+              "A persistent tracing session for '%s' is already running (%s is "
+              "locked).",
+              session_name.c_str(), target_file_path.c_str());
+        }
+        PERFETTO_PLOG("Failed to check flock on %s", target_file_path.c_str());
+      }
+      existing_fd.reset();
```
primiano: This is not neeed given you are using scopedfile

src/perfetto_cmd/perfetto_cmd_android.cc
```diff
@@ -284,15 +288,16 @@ base::ScopedFile PerfettoCmd::CreateUnlinkedTmpFile() {
   return fd;
 }
```
primiano, line 290: at this point woudl be nice to have a comment explaining all the possible scenarios and what they return, as this function now is a bit more complex

primiano (approved):
LGTM with a few comments

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Removed the redundant `CLOEXEC`.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Comments are updated.

The following block handles early return case when another instance of perfetto_cmd is still actively using it. Added your suggested comments where we push the `kRebootTraceUploadLeftover` atoms.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Removed the explicit `reset` call.

src/perfetto_cmd/perfetto_cmd_android.cc
safayat-google: Thanks. Added comments explaining all scenarios.

PR #7420, opened 2026-09-11
ui: add anew "compact" time format (9y8m7d6m5s4ms3us2ns)

LalitMaganti:
Playing aroudn and honestly feel this is the nicest format. I would like
this to become the default eventually but for now an option is fine.

### DIFF
--- a/ui/src/base/time.ts
+++ b/ui/src/base/time.ts
@@ -192,6 +192,12 @@ export class Time {
     return Time.toMicros(time).toString() + ' µs';
   }
 
+  // Format as a compact list of units relative to zero, e.g. "1h2m3s4ms".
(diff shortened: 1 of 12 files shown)


stevegolton:
It doesn't seem particularly compact to me: 
image

Perhaps you could describe some of the design decisions and how it operates?

LalitMaganti:
> Perhaps you could describe some of the design decisions and how it operates?

Was just an experiment inspired by other profilers.

I personally find the timecode format very difficult to work with and always get confused about what each position means. I was trying to get something which works similar but tells me the position in line. 

Compact was a poor name, the goal is not to be compact but to be readable. I would call this "human readable" (that was my initial name) but it clashed with human readable duration precision.

PR #7421, opened 2026-09-11
proto_merger: ignore LINT comments in proto_merger

dragosp95:
LINT comments shouldn't be forwarded in automatically generated files, as they depend on the original sources and so can't be updated at the same with them.

### DIFF
--- a/src/tools/proto_merger/proto_file.cc
+++ b/src/tools/proto_merger/proto_file.cc
@@ -217,9 +217,25 @@ Output InitFromDescriptor(const Descriptor& desc) {
   if (!desc.GetSourceLocation(&source_loc))
     return {};
 
+  auto filter_lint_comments =
+      [](const std::vector<std::string>& lines) -> std::vector<std::string> {
+    std::vector<std::string> parsed;
+    for (const auto& line : lines) {
+      std::string t = base::TrimWhitespace(line);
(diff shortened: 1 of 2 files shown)


src/tools/proto_merger/proto_file.cc
```diff
@@ -217,9 +217,25 @@ Output InitFromDescriptor(const Descriptor& desc) {
   if (!desc.GetSourceLocation(&source_loc))
     return {};
 
+  auto filter_lint_comments =
+      [](const std::vector<std::string>& lines) -> std::vector<std::string> {
+    std::vector<std::string> parsed;
+    for (const auto& line : lines) {
+      std::string t = base::TrimWhitespace(line);
+      if (base::StartsWith(t, "LINT.IfChange") ||
```
TParaschiv, line 225: nit: Why not just "LINT."?

TParaschiv (approved):

src/tools/proto_merger/proto_file.cc
dragosp95: I would like it to be more obvious what comments we are deleting.
If I were to see this code in 2 years there's a good chance i won't remember that `LINT.` is the beging of "LINT.IfChange" and "LINT.ThenChange"

PR #7424, opened 2026-09-11
ui: Group frame timelines by layer

lukaville:
## Summary
Groups Expected Timeline and Actual Timeline tracks by layer name as collapsible children under the process's 'Actual Timeline' track.

When expanding the 'Actual Timeline' track, each layer is displayed as a child node that can be further expanded to show its layer-scoped 'Expected Timeline' and 'Actual Timeline'.

Note on expected frames: some traces only tag expected frame slices with one of the layer names of a process. To attribute them to the right layer, expected frames are matched both by `layer_name` and via the `surface_frame_token` of the corresponding actual frame (deduped with a `UNION`).

## Testing
- Unit tests: `tools/pnpm --dir ui exec vitest run src/plugins/dev.perfetto.Frames/` (9/9 pass), covering the per-layer track hierarchy under 'Actual Timeline', layer-scoped SQL for both expected and actual frames, URI encoding, and the experimental jank track.
- Full suite: `vitest run` passes (2721 passed, 5 skipped).
- Code style: `tools/format-sources --check-only` passed with 0 issues.
- Manually verified on an Android trace with multiple layers per process (SystemUI status bar and notification shade).


LalitMaganti:
Drive-by: this seems overkill as a setting. Why not just automatically add the per layer breakdown as child tracks of the expected/actual timeline tracks?

lukaville:
I wasn't sure how useful this split could be for everyone, so wanted to not add much noise/overhead, but I guess it should be quite minimal to have it, so happy to remove the setting.

Re child tracks of expected/actual: I like the idea, I have tried it and it looks good (see screenshot)

One problem with it: I wouldn't group by layer separately in expected / actual tracks, since if we do that the matching layer tracks won't be shown together. So I added this grouping to "Actual timeline". But now it might be a bit odd that we also show 'expected' tracks under the root 'Actual timeline' track. Does anyone have any ideas how this could be improved? thanks!

image

nicomazz:
This is super useful, thanks for implementing it Nick! Having it as child tracks of the actual timeline looks good to me: I don't think we need any flag or setting for this. It's already hidden by default after all, and it doesn't seem to make the trace messy. If Lalit is fine with it, I would suggest having this enabled by default and iwhtout setting.

(btw, I can't find the nested tracks in the version generated by the bot)

nicomazz (approved):

LalitMaganti:
Steve for final approval.

lukaville:
In the latest patchset: removed the setting, it is now always enabled (collapsed by default under actual timeline)

PR #7439, opened 2026-09-13
tp: run PerfettoSQL pipelines end to end

LalitMaganti:
Execute standalone pipelines and pipeline bodies in `CREATE PERFETTO TABLE` through the batch executor. Move parser statements into execution without copying their plans, and preserve the original SQL for tracebacks.

Register one create-only SQLite module per connection. Each pipeline creates a TEMP virtual table with its own schema; unique positional columns are projected back to public names, including duplicate, empty, and quoted names. Bind the execution plan through a hidden pointer argument so SQLite owns its lifetime. Finalization releases the plan and retires its temporary table; a callback which `PreparedStatement` runs once the statement has been finalized then drops it. Cleanup waits while a transaction or active statement prevents a safe drop, and is retried when the next pipeline is finalized or prepared. No per-query module registration remains.

Cache typed column readers once per batch, specialize nullable reads, and pass known string lengths to SQLite. Output-rowid equality can be handled after tree accumulation; SQLite still evaluates general predicates so they do not change the fold's input. Use `Join` for column lists. Create `perfetto_tables` once per database, in `PerfettoSqlDatabase`, rather than once per connection, so forked connections can be constructed.

Validation: 235 selected parser, scanner, catalog, physical-plan, executor, and connection unit tests; all three pipeline diff tests. New coverage includes different schemas and overlapping statements, statement reset after schema changes, forked connections, deferred cleanup and rollback, failed table creation, output constraints, and mixed types across batches. GN header and formatting checks pass. Added and ran `BM_Connection_Pipeline_Scan` (32,768 rows × six columns; debug build, no release-performance claim).


src/trace_processor/perfetto_sql/engine/perfetto_sql_connection.cc
```diff
@@ -1017,6 +1058,42 @@ base::Status PerfettoSqlConnection::ExecuteCreateTable(
   return exec_res.status();
 }
 
+base::StatusOr<PerfettoSqlConnection::RegisteredPipeline>
+PerfettoSqlConnection::RegisterPipeline(pipeline::LogicalPlan logical) {
```
LalitMaganti: It would be better if instead of this we had a single eponymous only module and created instances of this. It would be *even* nicer if SQLite had an API to let us register a temporary virtual table for the lifetime of this statement or similar.

src/trace_processor/perfetto_sql/engine/pipeline_module.cc
```diff
@@ -0,0 +1,196 @@
...
+
+namespace perfetto::trace_processor {
+namespace {
+
+using core::exec::ColumnView;
+using core::exec::Variant;
+
+std::string CreateTableStmt(const pipeline::PhysicalPlan& plan) {
+  std::string stmt = "CREATE TABLE x(";
+  // SQLite requires unique virtual-table column names. Public names (which
```
LalitMaganti: Why not use JoinString or similar?

src/trace_processor/perfetto_sql/engine/pipeline_module.cc
```diff
@@ -0,0 +1,196 @@
...
+  return SQLITE_OK;
+}
+
+int PipelineModule::Disconnect(sqlite3_vtab* vtab) {
+  std::unique_ptr<Vtab> v(GetVtab(vtab));
+  return SQLITE_OK;
+}
+
+int PipelineModule::BestIndex(sqlite3_vtab*, sqlite3_index_info* info) {
+  // No constraints are pushed down; SQLite applies them all.
```
LalitMaganti: Why not? Why should we not support this?

src/trace_processor/perfetto_sql/engine/perfetto_sql_connection.cc
```diff
@@ -1017,6 +1058,42 @@ base::Status PerfettoSqlConnection::ExecuteCreateTable(
...
+    sql += '"';
+  }
+  sql += " FROM " + table;
+  auto ctx = std::make_unique<PipelineModule::Context>();
+  ctx->plan = std::move(plan);
+  ctx->pool = pool_;
+  connection_->RegisterVirtualTableModule(
+      table, &PipelineModule::kModule, ctx.release(),
+      [](void* p) { delete static_cast<PipelineModule::Context*>(p); });
+  return RegisteredPipeline{
```
LalitMaganti: Where is the deregistration for this also? How do solve multiple connections clashing on this.

src/trace_processor/perfetto_sql/engine/pipeline_module.cc
```diff
@@ -0,0 +1,196 @@
...
+  }
+  // Interned strings live as long as the pool.
+  sqlite::result::StaticString(ctx, pool->Get(id).c_str());
+}
+
+void ResultCell(sqlite3_context* ctx,
+                StringPool* pool,
+                const ColumnView& view,
+                uint32_t row) {
+  if (view.kind() == ColumnView::Kind::kVariant) {
```
LalitMaganti: This is going to be a very very very hot function. We should maybe spend some time optimizing this as such.

src/trace_processor/sqlite/sqlite_connection.h
```diff
@@ -71,11 +76,19 @@ class SqliteConnection {
     const base::Status& status() const { return status_; }
     sqlite3_stmt* sqlite_stmt() const { return stmt_.get(); }
 
+    // Keeps `dependency` alive until the statement is finalized, for things
```
LalitMaganti: Seems like a hack. Don't know why we need this.

src/trace_processor/perfetto_sql/engine/perfetto_sql_connection.cc
```diff
@@ -350,12 +353,20 @@ PerfettoSqlConnection::PerfettoSqlConnection(
...
+    RegisterVirtualTableModule<PipelineModule>(PipelineModule::kName,
+                                               std::move(ctx));
+  }
   // Initialize `perfetto_tables` table, which will contain the names of all of
   // the registered tables.
   char* errmsg_raw = nullptr;
-  int err = sqlite3_exec(connection_->db(),
-                         "CREATE TABLE perfetto_tables(name STRING);", nullptr,
-                         nullptr, &errmsg_raw);
+  int err =
```
LalitMaganti, line 366: Why if not exists? This should be removed.

src/trace_processor/perfetto_sql/engine/perfetto_sql_connection.cc
```diff
@@ -586,6 +597,7 @@ PerfettoSqlConnection::ExecuteNextStatement(SqlSource sql_source,
 base::StatusOr<std::optional<PerfettoSqlConnection::ExecutionResult>>
 PerfettoSqlConnection::ExecuteStatements(SqlSource sql_source,
                                          uint32_t* end_offset) {
+  RETURN_IF_ERROR(pipeline_context_->Cleanup(connection_->db()));
```
LalitMaganti, line 600: WHy is this here? this should be happening at the end of the preivous pipeline no?

src/trace_processor/perfetto_sql/engine/pipeline_module.h
```diff
@@ -0,0 +1,135 @@
...
+                    const char*,
+                    int,
+                    sqlite3_value**);
+  static int Next(sqlite3_vtab_cursor*);
+  static int Eof(sqlite3_vtab_cursor*);
+  static int Column(sqlite3_vtab_cursor*, sqlite3_context*, int);
+  static int Rowid(sqlite3_vtab_cursor*, sqlite_int64*);
+
+  // Refreshes per-column readers only when entering a new batch.
+  static void CacheColumnReaders(Cursor*);
```
sashwinbalaji: this can be in .cc anon namespace

sashwinbalaji (approved):

PR #7441, opened 2026-09-13
tp: fail gracefully when llvm-symbolizer cannot be run

LalitMaganti:
Probe llvm-symbolizer upfront and fail gracefully (instead of dying with SIGPIPE) when it can't be run.


src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -1128,6 +1141,16 @@ LocalSymbolizer::LocalSymbolizer(std::unique_ptr<BinaryFinder> finder)
 
 LocalSymbolizer::~LocalSymbolizer() = default;
 
+bool IsLlvmSymbolizerAvailable() {
+  return LLVMSymbolizerProcess(kDefaultSymbolizer).Probe();
```
sashwinbalaji, line 1145: can we cache the process LLVMSymbolizerProcess(kDefaultSymbolizer) ?

src/trace_processor/util/symbolizer/local_symbolizer.cc
```diff
@@ -995,6 +995,19 @@ std::vector<SymbolizedFrame> LLVMSymbolizerProcess::Symbolize(
...
+  // llvm-symbolizer replies with JSON even if the binary doesn't exist.
+  static constexpr char kProbeRequest[] = "\"/nonexistent\" 0x0\n";
+  if (subprocess_.Write(kProbeRequest, sizeof(kProbeRequest) - 1) < 0) {
+    return false;
+  }
+  auto line = GetLine([&](char* read_buffer, size_t buffer_size) {
+    return subprocess_.Read(read_buffer, buffer_size);
+  });
+  std::vector<SymbolizedFrame> frames;
+  return ParseLlvmSymbolizerJsonLine(line, &frames);
```
sashwinbalaji, line 1008: won't this cause a incorrect elog (Failed to symbolize) which might confuse users? http://ac/external/perfetto/src/trace_processor/util/symbolizer/local_symbolizer.cc?l=896

sashwinbalaji (approved):

src/trace_processor/util/symbolizer/local_symbolizer.cc
LalitMaganti: No. The probe runs once per symbolization run and costs ~6ms. Caching needs global state for no real gain.

src/trace_processor/util/symbolizer/local_symbolizer.cc
LalitMaganti: Good catch, fixed. The parser no longer logs; Symbolize() does.

PR #7442, opened 2026-09-13
tp: allow symbolizing unsymbolicated Firefox Profiler format profiles

LalitMaganti:
Import samply's unsymbolicated frames with their build ID and address so `bundle` can symbolize them; fix Mach-O address correction.


src/trace_processor/importers/gecko/gecko_trace_tokenizer.cc
```diff
@@ -658,20 +780,29 @@ void GeckoTraceTokenizer::ProcessLegacySamples(
...
     uint32_t name_str_idx = t.func_names[func_idx];
     if (name_str_idx >= strings.size()) {
       continue;
     }
-    frame_ids.push_back(InternFrame(base::StringView(strings[name_str_idx])));
+    base::StringView name(strings[name_str_idx]);
+    if (const GeckoLib* lib =
+            GetLibForUnsymbolizedFrame(t, i, func_idx, libs, name)) {
+      frame_ids.push_back(InternNativeFrame(
+          *lib, static_cast<uint64_t>(t.frame_addresses[i]), name));
```
sashwinbalaji: nit: can we rewrite this section such that we calculate mapping and build id once per lib and reuse it for all the frames of that lib.

src/trace_processor/importers/gecko/gecko_trace_tokenizer.cc
```diff
@@ -420,6 +466,65 @@ base::Status ParseMeta(json::SimpleJsonParser& reader, GeckoProfile& profile) {
...
+const GeckoLib* GetLibForUnsymbolizedFrame(const GeckoThread& t,
+                                           size_t frame_idx,
+                                           uint32_t func_idx,
+                                           const std::vector<GeckoLib>& libs,
+                                           base::StringView name) {
+  // Unsymbolicated frames are named by their address.
+  if (!name.StartsWith(base::StringView("0x"))) {
+    return nullptr;
+  }
+  if (frame_idx >= t.frame_addresses.size() ||
```
sashwinbalaji, line 514: thanks to AI: https://github.com/mstange/samply/pull/866/changes/84399b2903d69e9c582437bacfa0fe8798df0f2e

```
Version 70 moves the `lib` column off the resourceTable and onto the frameTable
```

so we should also read frameTable.lib

sashwinbalaji (approved):

src/trace_processor/importers/gecko/gecko_trace_tokenizer.cc
LalitMaganti: samply is chruning a bit too fast atm. I don't want to support the latest changes until it actually cuts a release.

src/trace_processor/importers/gecko/gecko_trace_tokenizer.cc
LalitMaganti: Done.

PR #7445, opened 2026-09-14
perfetto_cfg: copy the config instead of symlinking to it

alexander-scott:
`_perfetto_cfg_repository_impl` produced a repository whose only content was a symlink to `@perfetto//bazel/standalone:perfetto_cfg.bzl`, obtained via `repository_ctx.path(Label(...))`. Two problems, both of which surface under Bazel's remote repo contents cache (`--experimental_remote_repo_contents_cache`):

  * `repository_ctx.path(Label(...))` must produce a real filesystem path for the target, which forces the whole of `@perfetto` (~128 MB) to be materialized in order to read one ~7 KB file. When `@perfetto` is served lazily from the remote cache, materializing it costs a network round trip per file and takes a significant amount of time.

  * the resulting symlink pointed outside the repository's own root, at an absolute path into `@perfetto`'s directory in this output base. A repo containing such a symlink is never stored in the repo contents cache, so `@perfetto_cfg` was rebuilt from scratch in every output base, and the stored path would dangle if restored on another machine.

Read the file and write a real copy instead, which makes the repo self-contained.

The repo can then declare itself reproducible - its contents are a pure function of the `perfetto_cfg.bzl` file read above, which is a recorded input.

Note that `repository_ctx.read(Label(...))` is not itself cheaper than `path()` - both reach `ensureMaterialized` and pull the whole target repo. The saving is second-order: `@perfetto_cfg` becomes cacheable, and on a cache hit its implementation doesn't run at all, so nothing asks for `@perfetto`. A cold first fetch still materializes it.


sashwinbalaji (approved):

sashwinbalaji (approved):
sorry, I approved without adding the comments earlier.

Thanks for the PR! And yes switching from symliknk to read will help with caching

But, I don't think you can escape from materializing the entire perfetto repo when the rule runs. I assume you're using a newer version as you're using experimental_remote_repo_contents_cache 

So now tracing the code in bazel, readfile -> getPath -> ... -> ensureMaterialized [ctx](https://github.com/bazelbuild/bazel/blob/9.2.0/src/main/java/com/google/devtools/build/lib/bazel/repository/starlark/StarlarkBaseExternalContext.java). And this will lead to entire perfetto repo being materialzied.

So LGTM on the change but would ask you to confirm this once and update the pr description accordingly.

alexander-scott:
Hey @sashwinbalaji, thanks for the review. Yeah actually I think you're correct, the first part of the description is not true, both `path` & `read` will materialise the whole repo and one is not cheaper than the other. I think the claim came from how I did some tests locally, but re-running it now, I can confirm both `path` & `read` behave the same.

But the main fix that this PR does is still correct: avoiding creating a symlink that escapes the repos own root, allowing `@perfetto_cfg` to stay in the `remote_repo_contents_cache` and importantly avoiding the need to materialise `@perfetto` when there's a cache hit.

I'll update the PR descriprtion.

PR #7448, opened 2026-09-14
ai: add Android memory diff workflows for heap dumps

lukaville:
Introduce comparative diff skills for Android memory analysis in Perfetto:
- Add heap_dump_diff.md and diff_heap_graphs.py for comparing ART Java heap graphs across two traces, e.g. class-level and dominator-path deltas
- Register the comparative memory diff workflows in SKILL-template.md

Evaluation heap dumps are created using a fake sample app.

## Evaluation Results (`baseline-tp` vs `skill-local`)

| Eval Case | Metric | `baseline-tp` (No Skill) | `skill-local` (With Skill) | Delta / Improvement |
| :--- | :--- | :---: | :---: | :---: |
| **`heap-dump-diff`**<br>*(Perfetto `.perfetto-trace` Java heap diff)* | **Score / All-Pass Rate** | `0.75` / `0%` | **`1.00` / `100%`** | **+0.25 (+100% pass rate)** |
| | `leaked-class` | 100% | 100% | — |
| | `instance-count` (`20 -> 70`, `+50`) | 100% | 100% | — |
| | `dominated-size` (`+3,306,050 B`) | 0% | **100%** | **+100%** |
| | `retention-path` (`TileLifecycleController`) | 100% | 100% | — |
| | **Duration** | `306.4s` | **`78.7s`** | **3.9x faster (-74%)** |
| | **`trace_processor` invocations** | `24` (`16` SQL calls) | **`6` (`4` SQL calls)** | **4.0x fewer (-75%)** |
| | **SQL Errors** | `0` | **`0`** | — |
| **`heap-dump-hprof-diff`**<br>*(Android `.hprof` field-value attribution)* | **Score / All-Pass Rate** | `0.67` / `0%` | **`1.00` / `100%`** | **+0.33 (+100% pass rate)** |
| | `leaked-tile-spec` (`bluetooth_audio`) | 100% | 100% | — |
| | `primitive-fields` (`1500` / `autoRefreshEnabled`) | 100% | 100% | — |
| | `unchanged-tile-spec` (`wifi`) | 0% | **100%** | **+100%** |
| | **Duration** | `348.8s` | **`58.0s`** | **6.0x faster (-83%)** |
| | **`trace_processor` invocations** | `0` (parsed raw binary) | **`6` (`4` SQL calls)** | **Uses warm `tp` sessions** |
| | **SQL Errors** | `0` | **`0`** | — |


LalitMaganti:
I have some concerns taken over the approach in this change:
1) until now the Python scripts in the skills folder have been small and focused. We're now checking in skills which are huge, untested and only accessible to AI. This is troubling as imo anything accessible to AI should also be accessible to humans, from a capability perspective. 
2) you've shipped some tests but they don't run on CI nor are they integrated with our test suite. This seems unideal. 
3) I'm worried about the performance of these scripts. This is a lot of unbenchmarked code we're having to take responsibility for. I would feel better if this was smaller and more focused than dumping so much code here.

lukaville:
Thanks, Lalit! Yeah, these are valid points.

I see two alternative options:

1. 'Productionalize' the Python script to diff heap dumps/profiles and move it to `tools/`, so that everyone can use it. We can add proper functional / performance tests for it. Not sure though if user adoption for this script will be high, I think the most straightforward entry point for this should be Perfetto Web UI to diff traces, which as far as I understand planned but could take quite some time to have it implemented and ready?

2. I think we should be able to achieve similar results by just including some sample SQL queries to the skill itself instead of using the Python script. We won't need to support the code since everything will be in md files, but we won't have proper testing except AI evals

I'm leaning towards (1), what do you think about this? If you are okay with it, I'm happy to create a separate PR to introduce this tool, and then we can add AI skills for it

LalitMaganti:
In general, my preference is to *not* have big ad hoc scripts like this. These should simply be features of the trace processor cli. If you look at some of the RFCs and prototype PRs you will already see I've been thinking a lot about these things. 

If sysui have the bandwidth to commit time to working with me on this, I'd gladly take on that work. But otherwise if you just want to do the minimum thing then please have a small focused script/SQL which only does what you need for the skill.

LalitMaganti:
More generally: why should trace diffing be for heap dumps alone? Why not for any arbitrary aggregatable data? This is my problem with doing ad-hoc scripts. 

I think a good starting point if you did want to do the proper thing is to write up an RFC for diffing in general and use this as a motivation. 

Please see https://github.com/google/perfetto/tree/rfcs and https://github.com/google/perfetto/discussions/5300

lukaville:
Thanks Lalit for the context! I'm glad that this is something that is being worked on. I don't think we will have bandwidth to work on this actively, but I'm happy to review the proposals and provide any feedback from our side.

Re this change: I have removed the Python script and significantly slimed down the skill

ai/skills/perfetto/workflows/android_memory/scripts/diff_heap_csv.sql
```diff
@@ -0,0 +1,130 @@
+-- Diffs two heap CSVs exported by query_class_aggregation.sql.
+--
+-- Usage:
+--   sqlite3 :memory: -cmd ".mode csv" -cmd ".import base.csv base_raw" \
+--     -cmd ".import cand.csv cand_raw" -cmd ".mode markdown --wrap 0" < diff_heap_csv.sql
+--
```
nicomazz: aha so this script is not for the trace processor at all, but for sqlite3 binary itself?

is it worth having it in sql or would it look better as a python script?

ai/skills/perfetto/workflows/android_memory/heap_dump_diff.md
```diff
@@ -0,0 +1,136 @@
...
+  -cmd ".mode markdown --wrap 0" \
+  < $SKILL_ROOT/workflows/android_memory/scripts/diff_heap_csv.sql
+```
+
+Total heap size fluctuates on its own (caches, `SoftReference` eviction), so
+never dismiss a regression because the heap total looks flat.
+
+## Step 3: Inspect the leaked class (`<LEAKED_CLASS>`)
+
+Once Step 2 identifies `<LEAKED_CLASS>`, run targeted queries against `cand_tp`
```
nicomazz: are we ssuming it's a single one?

ai/skills/perfetto/workflows/android_memory/heap_dump_diff.md
```diff
@@ -0,0 +1,136 @@
...
+
+Groups reachable instances of `<LEAKED_CLASS>` by their `String` and primitive
+field values so the leaked configuration (e.g. `tag = 'camera_preview'`) can be
+distinguished from baseline instances (`tag = 'launcher_icon'`). Run on both
+`base_tp` and `cand_tp` to compare counts. Field values are only recorded in
+Android `.hprof` dumps, so this query is not applicable to Perfetto traces.
+
+```bash
+trace_processor query --remote cand_tp -f - << 'EOF'
+CREATE OR REPLACE PERFETTO TABLE target_objs AS
```
nicomazz: very very nit: should we add one indentation for this block so it's easier to parse for humans (and understand the previous line is a command)

ai/skills/perfetto/workflows/android_memory/scripts/diff_heap_csv.sql
LalitMaganti: I agree a focused python script is the way to go. My comments above was not to discourage Python scripts but instead of simply suggest we don't want big complex scripts.

ai/skills/perfetto/workflows/android_memory/heap_dump_diff.md
lukaville: Added explicit instructions that we might have more than one leaked class

ai/skills/perfetto/workflows/android_memory/scripts/diff_heap_csv.sql
lukaville: yeah, doing this in sql is a bit clunky, brought back a small python script for diffing

LalitMaganti (approved):

nicomazz (approved):

PR #7324, opened 2026-09-03
tracing: add the v2 shared-ring chunk protocol

sashwinbalaji:
Add the shared-memory ring that tracing v2 producers will write packet
fragments into, with a writer and a reader class on top of it:

- One 64-bit atomic holds both the write and the read position. Writers
  reserve positions with a lock-free CAS on it; the reader publishes
  read_pos once per drain pass.
- Every chunk starts with a 32-bit state word. The five states (Free,
  BeingWritten, Complete, RewriteRequested, RewriteAcknowledged) change
  hands only through exact-value CAS, so whoever loses a race sees the
  word that won.
- A Free word carries a 16-bit wrap count, so a delayed writer cannot
  claim a chunk for a position the reader already resolved. Only the
  reader ever writes Free.
- Writers append fragments to their chunk and publish by moving the
  state word. The reader may take the committed prefix of a chunk that
  is still being written; only the just-ended fragment can require
  relocation into a fresh chunk.
- Finishing a writer releases its cached chunk, so later fragments
  require a new reservation. This establishes the flush watermark
  boundary.
- Ring exhaustion follows BufferExhaustedPolicy: kDrop, kStall and
  kStallThenDrop. On Linux and Android a stalled writer waits on
  read_pos with a futex, under the same 30-second ceiling as v1.
- Tests cover the ABI encoding, reader/writer race outcomes, position
  and wrap-count rollover, and multi-writer stress against a concurrent
  reader.

PR #7449, opened 2026-09-14
perfetto: Add big-endian support

miladfarca:
The V8 Js Engine is working towards removing legacy tracing and using Perfetto instead. Node.js is also enabling it by default:
https://github.com/nodejs/diagnostics/issues/654
https://github.com/nodejs/node/pull/65794

Currently setting `v8_use_perfetto = true` in V8 fails a number of tests on big-endian platforms which this PR attempts to fix.

The protobuf wire format requires fixed-width fields (fixed32/fixed64, float/double) to be little-endian. protozero memcpy's them in host order, so on big-endian (s390x) traces are corrupt and unreadable by libprotobuf/trace_processor. BE was previously stubbed with `#error` guards.


sashwinbalaji:
Thanks for working on this! And excited to see Node.js using perfetto.

We have a lot of little-endian assumptions throughout the codebase and this doesn't cover all [cs](https://source.chromium.org/search?q=little.*endian&sq=&ss=chromium%2Fchromium%2Fsrc:third_party%2Fperfetto%2F). 

More so, adding support is one big thing but for us to maintain and consider for all future changes (for e.g this breaks one of the ongoing PR https://github.com/google/perfetto/pull/7324) is going to be a big commitment.

Please give us a little time to discuss internally, and we’ll get back to you about the possible path.

miladfarca:
Not a problem,
I just wanted to clarify that we are mainly concerned with making sure the emitter has big-endian support, not the trace_processor/analyzer at the moment, which only spans a small number of files.

We already maintain V8/Node.js and some of their dependencies (such as Abseil, Highway, etc.) on big-endian platforms and once Perfetto is enabled by default we will have to maintain this as well (only the emitter for now).

primiano:
Hey,
thanks for sending this out.
generally speaking I'm supportive adding support to big-endian if there is a clear case like Node that needs it (can you expand a little bit on which architecture btw?).

However:

1. This patch seems quite incomplete. if you want to go there you need to do better auditing. I remember trying my best to leave around notes via AssumeLittleEndian and PERFETTO_IS_LITTLE_ENDIAN. But for instance I'm seeing that in message.cc you are just removing the `#error` on PERFETTO_IS_LITTLE_ENDIAN but you are not touching any .cc code (why?). There are also various other place that have that.

2. My only request is that whoever does it has to maintain the BE port over time. We can't commit to maintain something that we can't even test (I am not even sure where I would find a BE VM these days)

So, overall supportive, but this PR seems incomplete

P.S. also I'm not sure I follow the rationale of the change in android.gni, how do you end up there with `target_os != "android"` ?

miladfarca:
> can you expand a little bit on which architecture btw?

This PR is only concerned with IBM platforms, mainly big-endian platforms, however perfetto usage will affect [all supported platforms](https://github.com/nodejs/node/blob/main/BUILDING.md#supported-platforms) by Node.

> This patch seems quite incomplete. 

To be clear on scope, this targets only the emitter, not a big-endian port of trace_processor/the analyzer. This PR is only created to make V8 compile with all its tests passing when [v8_use_perfetto](https://github.com/v8/v8/blob/3640a862c8e6638fe42235df945b29003d000bd2/BUILD.gn#L996) is set to true on s390x. 

message.cc has zero fixed width serialization code and the float/double memcpy the `#error` warned about now lives in message.h's AppendFixed template. Most/all other spots with `IS_LITTLE_ENDIAN` seem to be related to the trace_processor and do not affect the tracer (as tests seem to all pass on V8).

> whoever does it has to maintain the BE port over time.

As long as V8/Node.js use it as a dependency we will make sure tests are passing using perfetto on our platforms.

The `gn` change is added because I tried building perfetto by itself from source and got these errors:
```
ERROR at //gn/standalone/android.gni:47:5: Assertion failed.
    assert(false, "Need android libgcc support for this arch.")

ERROR at //gn/standalone/android.gni:63:5: Assertion failed.
    assert(false, "Unknown ABI: " + current_cpu)
    ^-----
Unknown ABI: s390x
```
However they can be removed from this PR if you prefer.

sashwinbalaji:
Thanks for the clarification and commitment. 

Now just to fully cover the emitter side you still need to make changes at below places which do a direct memcpy:

1) PackedFixedSizeInt::Append [include/perfetto/protozero/packed_repeated_fields.h:101–109](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/include/perfetto/protozero/packed_repeated_fields.h#L101), specifically:

```cpp
memcpy(reinterpret_cast<T*>(write_ptr_), &value, sizeof(T));
```

2) PackedRepeatedFieldIterator [include/perfetto/protozero/proto_decoder.h:240–246](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/include/perfetto/protozero/proto_decoder.h#L240):

```cpp
memcpy(&curr_value_, read_ptr_, sizeof(CppType));
```

3) update emitter side uuid assumption
 - [include/perfetto/ext/base/uuid.h:45–62](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/include/perfetto/ext/base/uuid.h#L45), [src/base/uuid.cc:89–103](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/src/base/uuid.cc#L89). The emitter constructs/logs this UUID at [tracing_service_impl.cc:631–637](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/src/tracing/service/tracing_service_impl.cc#L631) and emits its numeric halves at [lines 4150–4151](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/src/tracing/service/tracing_service_impl.cc#L4150).

Also you would've to

1) Fix the buggy is big endian endian check by adding parantheses (and that might show up more failures)

- [include/perfetto/base/compiler.h:115–119](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/include/perfetto/base/compiler.h#L115), [include/perfetto/ext/base/utils.h:160–165](https://github.com/google/perfetto/blob/f8d9726ce5127ecba790fbbfde47920da37f4054/include/perfetto/ext/base/utils.h#L160)

```
  // Current macro:
  #define PERFETTO_IS_LITTLE_ENDIAN() __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__

  // This check:
  #if !PERFETTO_IS_LITTLE_ENDIAN()

  // Expands to:
  #if !__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__

  // Since ! binds more tightly than ==, it means:
  #if (!__BYTE_ORDER__) == __ORDER_LITTLE_ENDIAN__

  // On big-endian:
  #if (!4321) == 1234  // -> 0 == 1234 -> false: the guard is skipped
```



If you run the perfetto unittests and integratiotests following https://perfetto.dev/docs/contributing/testing all these should show up.

sashwinbalaji:
~Also could we pls put this behind a default off GN flag. So if anybody wants big-endian support they need to explicitly op in? That would make it a bit easier to maintain and give us a simple way to turn it off if any issue comes up.~

See https://github.com/google/perfetto/pull/7449#issuecomment-5717779774

include/perfetto/protozero/message.h
```diff
@@ -135,7 +135,8 @@ class PERFETTO_EXPORT_COMPONENT Message {
     uint8_t* pos = buffer;
 
     pos = proto_utils::WriteVarInt(proto_utils::MakeTagFixed<T>(field_id), pos);
-    memcpy(pos, &value, sizeof(T));
+    T le_value = proto_utils::FixedToFromLittleEndian(value);
```
primiano: So rather than this I'd be more in favour of 

1. Make updates to base/endian.h
2. Implement those HostToBE16/32 and so on
3. Implement HostToLE16/32 and so on

The canonical pattern for that is:

1. Define macros for byteswap (they are there already in that file) to deal with msvc vs gcC/clang, and facotre them together
2. Implement those functions as "if little endian return value, else return swapped" and viceversa

Then use the right function here

gn/standalone/android.gni
```diff
@@ -56,6 +59,7 @@ declare_args() {
   } else if (current_cpu == "arm64") {
     android_app_abi = "arm64-v8a"
   } else {
-    assert(false, "Unknown ABI: " + current_cpu)
+    assert(target_os != "android", "Unknown ABI: " + current_cpu)
```
primiano, line 62: ack on this (and above).
seems fine after all

src/protozero/proto_decoder.cc
```diff
@@ -115,11 +111,14 @@ ParseOneField(const uint8_t* const buffer, const uint8_t* const end) {
     if (PERFETTO_UNLIKELY(new_pos > end))
       return res;
     memcpy(&int_value, pos, sizeof(uint64_t));
+    int_value = proto_utils::FixedToFromLittleEndian(int_value);
```
primiano: same suggestion here, this is going to be LE64ToHost

as the protobuf encoding is LE

primiano:
> Also could we pls put this behind a default off GN flag. So if anybody wants big-endian support they need to explicitly op in?
> 
> That would make it a bit easier to maintain and give us a simple way to turn it off if any issue comes up.

Hmm I think this might end up causing more churn than anything.
At the end of the day all the code is gated an `#if` which is only hit on BE, so we will never hit those code paths (and somebody else would have to maintain them).

I'm not sure a dedicated GN flag would buy us much.
I'd be with you if they were introducing extra .cc files. But I think all theyhave to do is to improve the endian.h file and invoke the right HostTo{BE,LE}{32,64} and  {LE,BE}ToHost{32,64} and that's it really

miladfarca:
Thank you for all the information. I have updated the PR accordingly. The way I'm building and running the tests is as follows:
I've added this to the end of `src/protozero/BUILD.gn` (not part of this PR):
```
test("protozero_emitter_unittests") {
  deps = [
    ":unittests",
    "../../gn:default_deps",
    "../../gn:gtest_main",
  ]
}
```
Then I build it with `ninja -C out/s390x protozero_emitter_unittests` and run tests with `out/s390x/protozero_emitter_unittests`. They all pass on my native machine.

I cannot build `perfetto_unittests` on its own because it pulls in code for the ipc/trace_processor (which we are not patching here) and causes build failures:
```
In file included from ../../src/ipc/buffered_frame_deserializer.h:26:
../../include/perfetto/ext/base/utils.h:163:17: error: static assertion failed: Unimplemented on big-endian archs
  163 |   static_assert(false, "Unimplemented on big-endian archs");
```

primiano:
> buffered_frame_deserializer.h:

Hmm but buffered_frame_deserializer.h is used in the standard IPC code required for tracing. How can you hit that path and still be able to build?
that is not some trace-processor only code, it's production tracing code.

P.S: I'll look anyways to the PR later today as soon as I get to the airport

miladfarca:
`src/ipc` isn't compiled in my setup. My local test target only depends on `src/protozero:unittests` ([gn block here](https://github.com/google/perfetto/pull/7449#issuecomment-5733893420)), none of those pull in `src/ipc`.

Maybe a better title for this PR is `Fix protozero on big-endian`.

include/perfetto/protozero/proto_utils.h
```diff
@@ -38,6 +41,48 @@
 namespace protozero {
 namespace proto_utils {
 
+// Converts a fixed32/fixed64/sfixed/float/double between host byte order and
+// the little-endian proto wire representation.
+template <typename T>
```
primiano: These two functions seems quite redundant in the way they are written, as they do too many memcpys. If you follow all the paths from message.h, there are now 3 memcpys, and I think you need only one total.

The general problem that requires memcpy in protozero is that in some places we do write into unaligned locations, so we need a memcpy regardless of endianness. but that's a reason for  one, not 3 

But in the way you are using these functions (which is correct at the call site) you are stashing the values on the stack, so you don't need these two extra memcpys, unless i'm missing something.

The only thing I would do here is to add a generic version of HostToLE / LeToHost in base/endian.h 

You don't even need templates, overloads should be enough, so something like

inline uint16_t HostToBE(uint16_t x) { return HostToBE16(x); }
inline uint32_t HostToBE(uint32_t x) { return HostToBE32(x); }
inline uint64_t HostToBE(uint64_t x) { return HostToBeE4(x); }


that should be enough, so you can avoid any change here

primiano (commented):
Overall looks in a good shape, I have only one final suggestion as the proto_utils helper look redundant

primiano:
> Maybe a better title for this PR is `Fix protozero on big-endian`.

More like "support big-endian in in-process tracing SDK code paths"

miladfarca:
Thanks again for reviewing. I've made the changes and have also changed the title of the PR.

primiano:
> Thanks again for reviewing. I've made the changes and have also changed the title of the PR.

See my previous comments on the triple memcpy. I don't think you need all those memcpys. Is there a reason why the generic version can't just be an overload with the same name that calls the right function?
(see the snippet i put in my previous comment)

miladfarca:
> I don't think you need all those memcpys.

Sorry which memcpy are you referring to? My last patch with templates is only using a memcpy for float types, int types are not using a memcpy. Are you asking I avoid using templates here?

primiano:
> > I don't think you need all those memcpys.
> 
> Sorry which memcpy are you referring to? My last patch with templates is only using a memcpy for float types, int types are not using a memcpy. Are you asking I avoid using templates here?

Aaah I didn't realize that branch was for floats and double sorry.
But why don't we keep things consistent? like:

1. Define HostTo{LE,BE}Float and HostTo{LE,BE}Double (and viceverse for ToHost) using byteswap
2. For the generic function, Rather than that template function, can we just use standard overloads as they are much much more readable?

Or is there a strong reason for using c++ templates, which overloads wouln't capture?

miladfarca:
Not a problem, I've modified it to use overloads instead.

include/perfetto/ext/base/endian.h
```diff
@@ -19,57 +19,207 @@
...
 }
 inline uint64_t BE64ToHost(uint64_t x) {
-  return __builtin_bswap64(x);
+  return x;
 }
 #endif
 
+inline float HostToLEFloat(float x) {
+  uint32_t u;
+  memcpy(&u, &x, sizeof(u));
```
primiano: these should be guarded like the other ones above.
Now on little endian arhitecture you are doing a double memcpy, while in reality you should just `return x`.

miladfarca: They should get optimized out by the compiler at -O1, but no problem I've guarded them for consistency.

primiano (approved):
Thanks LGTM

miladfarca:
Thanks for all your help.

PR #7460, opened 2026-09-16
protozero: let HeapBuffered release its slices

LalitMaganti:
A finished message can only leave its HeapBuffered by being stitched
into a fresh allocation, so anything that wants to keep the bytes past
the object copies them. The slices are already move-only heap buffers,
so add TakeSlices(), which hands them over as they are and leaves the
buffer empty and reusable, exactly like Reset() does.

This is what lets a trace processor RPC response be owned by the
transport rather than borrowed from the stack of whoever produced it.

out/linux_clang_release/perfetto_unittests \
    --gtest_filter=ScatteredHeapBufferTest.*


src/protozero/scattered_heap_buffer.cc
```diff
@@ -75,6 +75,13 @@ ScatteredHeapBuffer::GetSlices() {
   return slices_;
 }
 
+std::vector<ScatteredHeapBuffer::Slice> ScatteredHeapBuffer::TakeSlices() {
+  AdjustUsedSizeOfCurrentSlice();
+  std::vector<Slice> slices = std::move(slices_);
+  slices_.clear();
+  return slices;
```
safayat-google, line 84: nit: should we call `writer_ ->reset` here?
Even though the comments says- writer is not reset, other func does reset it.
>  // Like GetSlices() but hands them over, leaving this empty. The writer still
   // points into the last one: Reset() it before writing again.
   std::vector<Slice> TakeSlices();

safayat-google (approved):

PR #7465, opened 2026-09-16
tp: handle PERF_RECORD_FORK and PERF_RECORD_EXIT in perf record parser

TudorCretu:
Previously, `PERF_RECORD_FORK` (type 7) and `PERF_RECORD_EXIT` (type 4) records in `perf.data` traces were unhandled in `RecordParser::ParseRecord`, incrementing `perf_record_skipped[4]` and `perf_record_skipped[7]` and leaving forked child processes/threads unlinked or exited threads open across TID reuse.

- Handle `PERF_RECORD_FORK` in `RecordParser` by linking `child_upid` to `parent_upid` via `ProcessTracker::SetProcessParent` (when `pid != ppid`) and associating `(tid, pid)` via `ProcessTracker::UpdateThread`.
- Handle `PERF_RECORD_EXIT` in `RecordParser` by associating `(tid, pid)` and terminating the thread via `ProcessTracker::EndThread(ts, tid)` so subsequent TID reuse allocates a new `utid`.
- Fall back to the built-in `time` field in `PERF_RECORD_FORK` and `PERF_RECORD_EXIT` payloads in `PerfDataTokenizer::ReadTime` when `sample_id_all` is not enabled.
- Add diff test `Simpleperf.perf_fork_exit`.


src/trace_processor/importers/perf/record_parser.cc
```diff
@@ -293,6 +299,43 @@ base::Status RecordParser::ParseComm(Record record) {
...
+  uint32_t pid;
+  uint32_t ppid;
+  uint32_t tid;
+  uint32_t ptid;
+  if (!reader.Read(pid) || !reader.Read(ppid) || !reader.Read(tid) ||
+      !reader.Read(ptid)) {
+    return base::ErrStatus("Failed to parse PERF_RECORD_FORK");
+  }
+
+  if (pid != ppid) {
```
LalitMaganti, line 316: Why is this check there?

TudorCretu: If I understand correctly, PERF_RECORD_FORK is emitted for thread creation as well as process forks, and when spawning a thread inside the same process pid == ppid (e.g. (pid=26199, ppid=26199, tid=26200, ptid=26199) — 273 of the 306 PERF_RECORD_FORK records in my trace are like this). Without pid != ppid, SetProcessParent would either set the process as its own parent or log process_tracker_parent_pid_changed if its parent PID was already set. I can add a comment in the code explaining this.

LalitMaganti: Please do. This is highly non obvious

TudorCretu: Done.

LalitMaganti (approved):

PR #7475, opened 2026-09-17
stdlib + ui: add Android process state stdlib module and UI plugin

gignat-dev:
Adds support for analyzing and visualizing Android framework process states in long memory traces.

### Stdlib:
- Adds the `android.process_state` standard library module wrapping `__intrinsic_android_process_state`.
- Computes `_android_process_state_intervals` (per-process state durations and lifecycle boundaries) and `_android_process_state_concurrency` (state concurrency levels over time).
- Orders process states according to `ProcessStateEnum` protobuf declaration order.
- Includes diff tests covering interval reconstruction and concurrency aggregation.

### Perfetto UI:
- Adds an opt-in `com.android.ProcessState` plugin.
- Visualizes states under a dedicated top-level "Process states" track group containing:
  - **Concurrency by state:** Per-state counter tracks for all active states.
  - **By process:** Compact slice tracks ordered by UID and timestamp for respawning app continuity.
- Registers two area-selection aggregation tabs:
  - **Process State Residency:** Total duration and % occupancy per process/state.
  - **Process State Transitions:** Detailed state transitions table within the selection, including state change reasons, unclipped duration tooltips, and deep links to timeline slices.


src/trace_processor/perfetto_sql/stdlib/android/process_state.sql
```diff
@@ -0,0 +1,163 @@
...
+INCLUDE PERFETTO MODULE intervals.overlap;
+
+-- Ordering follows the declaration order of ProcessStateEnum in
+-- protos/third_party/android/frameworks/base/proto/tracing/frameworks_base_track_event.proto
+-- (lines 317-342).
+-- Numeric values do not match declaration order because
+-- PROCESS_STATE_BOUND_TOP was appended as 1020 while declared between
+-- TOP (1002) and FOREGROUND_SERVICE (1003).
+-- Explicit unknown states (UNSPECIFIED, UNKNOWN_TO_PROTO, UNKNOWN) sort last.
+CREATE PERFETTO TABLE _android_process_state_rank AS
```
zezeozue, line 28: We shouldn't do this here, maybe expose the raw proto value in addition to the process state string in the __intrinsic_android_process_state table

ui/src/plugins/com.android.ProcessState/index.ts
```diff
@@ -0,0 +1,201 @@
...
+  dur: LONG,
+  upid: NUM,
+  pid: NUM,
+  uid: NUM_NULL,
+  user_id: NUM_NULL,
+  process_name: STR_NULL,
+  package_name: STR_NULL,
+  version_code: NUM_NULL,
+  debuggable: NUM_NULL,
+  state: STR,
```
zezeozue, line 50: whats state vs state_name?

ui/src/plugins/com.android.ProcessState/index.ts
```diff
@@ -0,0 +1,201 @@
...
+  state_name: STR,
+  reason: STR_NULL,
+} as const;
+
+// Sort key for processes without a known UID so they sort after numbered UIDs.
+const NULL_UID_SORT_ORDER = 2147483647;
```
zezeozue: This seems weird? whats happening here?

src/trace_processor/perfetto_sql/stdlib/android/process_state.sql
gignat-dev: One small issue with using the raw proto int directly is that `PROCESS_STATE_BOUND_TOP` has wire value `1020` (it was added later after NONEXISTENT = 1019 to avoid renumbering 1003..1019, even though in ActivityManager it sits between `TOP` = 1002 and `FOREGROUND_SERVICE` = 1003), and the `UNKNOWN / UNSPECIFIED` values are < 1000.

That said, we only actually use state_rank in the UI plugin to order the concurrency tracks. 

We could either:
1. use the raw proto value and accept that the order will be a bit strange. (i.e. `UNKNOWN` will be the first track and `PROCESS_STATE_BOUND_TOP` will be the last track)
2. move the rank macro out of the stdlib module and into the UI plugin.

Wdyt?

ui/src/plugins/com.android.ProcessState/index.ts
gignat-dev: state was the raw string with the `PROCESS_STATE_` prefix and state_name had it stripped for display. 

Simplified to a single state column (without `PROCESS_STATE_`) in SQL.

ui/src/plugins/com.android.ProcessState/index.ts
gignat-dev: It was a workaround to sort processes with uid IS NULL to the bottom. 

Replaced it with `ORDER BY uid NULLS LAST, min(ts), pid`.

src/trace_processor/perfetto_sql/stdlib/android/process_state.sql
zezeozue: Fine, maybe leave as is then.

zezeozue (approved):

stevegolton (approved):
UI code LGTM

PR #7478, opened 2026-09-17
tools: skip gn formatting for chrome stdlib directory

rasikanavarange:
The Copybara sync from Chromium to Perfetto fails repo-checks (tools/run_presubmit) with a GN formatting error on src/trace_processor/perfetto_sql/stdlib/chrome/BUILD.gn.

Root Cause:
- In Chromium, we formatted base/tracing/stdlib/chrome/BUILD.gn using Chromium's latest gn binary. Newer gn versions insert a blank line after an import(...) statement when preceded by comments.
- Perfetto's tools/install-build-deps pins an older 2022 gn binary (gn-linux64-1968-0725d782), which formats BUILD.gn without that blank line.

Fix:
Skip src/trace_processor/perfetto_sql/stdlib/chrome/ in GnFormat (python/tools/code_format_gn.py), matching existing precedent in python/tools/code_format_sql.py and CheckAndroidBlueprint in tools/run_presubmit, since Chromium is the source of truth for this directory.


rasikanavarange:
An alternative would be to update Perfetto's gn binary, but Perfetto owners can comment on which approach is better :)

LalitMaganti:
I would prefer if we update GN instead.

rasikanavarange:
Thanks Lalit, I have raised https://github.com/google/perfetto/pull/7492 to update gn, PTAL!

PR #7480, opened 2026-09-17
Validate chunk index bounds in SharedMemoryArbiterImpl

etiennep-chromium:
Fix for b/513159433 where a malicious tracing service can trigger a TOCTOU.
This is a best effort fix; we can't flip perfetto trust model entirely (untrusted tracing service),
but this fix is relatively simple.


primiano (approved):
Thanks for fixing this. 
the PR itself is NOT controversial, so this LGTM (can I just ask you that you add a small comment around that check that points to the b/ bug?)

My only overall comment is that the SDK was never designed with the idea that a service is not trusted (i.e. a producer can be more privileged than the service itself) so I wouldn't rule out that there are similar issues lurking around.

BUt for now, hey. this LG

etiennep-chromium:
> (can I just ask you that you add a small comment around that check that points to the b/ bug?)

Done

> My only overall comment is that the SDK was never designed with the idea that a service is not trusted (i.e. a producer can be more privileged than the service itself) so I wouldn't rule out that there are similar issues lurking around.

I mostly agree; perfetto being in a different process in chrome is more about stability. I think we can merely fix this kind of issue as a best effort if trivial or ignore otherwise.

PR #7486, opened 2026-09-18
tp: preserve retained executor values and selections

LalitMaganti:
Tree ordering needs to retain batches across producer calls. The current borrowed-until-next-call contract requires defensive copies, even when the underlying dataframe already outlives the query.

Make owned values, validity and selections stable while retained, and recycle producer buffers only after consumers release them. This enables the copy removal in #7487; it is an ownership prerequisite, not a demonstrated query speedup. The cost is ownership bookkeeping and additional live buffers while outputs are retained.

PR #7487, opened 2026-09-18
tp: retain tree input views and gather reordered output

LalitMaganti:
Tree ordering copies retained rows even when their finalized dataframe storage outlives the query. Retain owned input views to avoid that copy; materialize borrowed input before retaining it.

Range output shares backing. Reordered output always gathers contiguous columns, avoiding scattered downstream reads without shared-backing or partial-run heuristics. A dense row-to-batch directory replaces binary searches at four bytes per retained row.

In the full-stack synthetic 100k-row comparison, wide shuffled DOWN improves 209→158 ms (original: 152 ms), while wide ordered UP→DOWN regresses 110→124 ms (original: 119 ms). These compare stack versions, not this PR in isolation.

Tests cover backing reuse, borrowed input, ordering, mixed validity and retained-output lifetime. All 171 affected tests pass across the stack.

Draft / DNM. Depends on #7486.


LalitMaganti:
@sashwinbalaji this and all dependent CLs are good to review also!

src/trace_processor/core/exec/column_chunk.cc
```diff
@@ -0,0 +1,78 @@
...
+  } else if (view.type().Is<Double>()) {
+    CopyValues<double>(view, count, offset, *this);
+  } else {
+    CopyValues<StringPool::Id>(view, count, offset, *this);
+  }
+  validity.resize(kMaxBatchRows);
+  for (uint32_t i = 0; i < count; ++i) {
+    if (!view.validity() ||
+        view.validity()->is_set(view.selection().GetIndex(i)))
+      validity.set(offset + i);
```
sashwinbalaji: nit: multi line if braces would be better

src/trace_processor/core/exec/row_store.cc
```diff
@@ -17,323 +17,148 @@
...
-void Copy(const ColumnView& column,
-          uint32_t from,
-          uint32_t at,
-          uint32_t count,
-          ColumnChunk& chunk) {
-  T* dest = chunk.Values<T>().data() + at;
-  const auto* data = static_cast<const T*>(column.data());
-  RowSelection selection = column.selection();
-  if (selection.is_range()) {
-    memcpy(dest, data + selection.offset() + from, count * sizeof(T));
```
sashwinbalaji: We had this fast path for range can we keep this in the new CopyFrom

PR #7403, opened 2026-09-09
tp: add explicit debuginfod lookup for native symbols

LalitMaganti:
Resolve missing native symbols by build ID without requiring local
binaries. Try local sources first, then validated cache entries and
ordered debuginfod servers. Use curl with configurable connection and
stall timeouts, and publish downloaded files atomically.

Require --debuginfod to enable lookups. Add URL and cache path overrides,
warn about configured URLs when disabled, and filter LLVM's environment
to prevent implicit downloads or option overrides. Report aggregates,
verbose lookup details, and optional terminal progress.

Document the workflow separately from the shared CLI reference.

PR #7494, opened 2026-09-18
tp: fix Chromium roll build errors in debuginfod and subprocess_windows

rasikanavarange:
The Perfetto → Chromium autoroller (crrev.com/c/8421777 https://chromium-review.googlesource.com/c/chromium/src/+/8421777) is currently failing due to two compile errors introduced by PR #7403 https://github.com/google/perfetto/pull/7403 "tp: add explicit debuginfod lookup for native symbols"

- In debuginfod.cc: wrap the post-return body of ResolveDebuginfodOptions() in an #else branch when !PERFETTO_BUILDFLAG(PERFETTO_LOCAL_SYMBOLIZER) to fix -Werror,-Wunreachable-code on Android/iOS/Fuchsia Chromium builds, ignore unused config parameter, and move #if PERFETTO_BUILDFLAG(PERFETTO_LOCAL_SYMBOLIZER) above ParseSeconds() to avoid -Wunused-function.
- In subprocess_windows.cc: #undef GetEnvironmentStrings and call GetEnvironmentStrings() instead of GetEnvironmentStringsA(), since Windows SDK <processenv.h> only defines the GetEnvironmentStringsA macro when UNICODE is not defined (whereas Chromium Windows builds define UNICODE).


src/trace_processor/util/symbolizer/subprocess_windows.cc
```diff
@@ -27,6 +27,11 @@
 
 #include "perfetto/base/logging.h"
 
+// <processenv.h> declares the ANSI function as GetEnvironmentStrings() and only
+// defines GetEnvironmentStringsA when UNICODE is not set. When UNICODE is set,
+// it defines GetEnvironmentStrings to GetEnvironmentStringsW.
+#undef GetEnvironmentStrings
```
LalitMaganti: Wouldn't it be better to use the non-macro function instead?

LalitMaganti (approved):

PR #7497, opened 2026-09-20
tp: bounds-check dual-clock ART method record before reading thread ts_delta

evilgensec:
## What
`ArtMethodTokenizer::ParseRecord` reads the dual-clock thread `ts_delta` at body offset `[8, 12)`:

```cpp
uint32_t ts_delta = clock_ == kDual ? ToInt(record.slice_off(8, 4))
                                    : ToInt(record.slice_off(4, 4));
```

For v1/v2 traces the record size is fixed by the version (`record_size_` = 9/10, so the body handed to `ParseRecord` is `slice_off(2, record_size_ - 2)` = **7/8 bytes**). Unlike v3 — which validates the record size against the dual-clock layout at header time via `kMinV3RecordSize` (2 tid + 4 method + 4 wall + 4 thread = 14), per the comment *"so later fixed-offset reads in ParseRecord cannot read past the record"* — v1/v2 are never checked against the clock mode.

The clock mode comes from a text option line (`clock=dual`) parsed independently. A **v1/v2 trace that declares `clock=dual`** therefore drives `slice_off(8, 4)` past the end of the 7/8-byte body. `slice_off` is only `PERFETTO_DCHECK`-guarded, so in release builds this returns an over-length view and `ToInt`'s `memcpy(&x, tbv.data(), tbv.size())` reads out of bounds (v2 dual: 4 bytes past; v1 dual: 5 bytes past). For the final record in the buffer this is a heap-buffer-overflow read.

## Fix
Reject the malformed dual-clock/small-record combination before the read. Real dual-clock traces are v3 (record size ≥ 14), so this rejects only malformed input; single-clock v1/v2 parsing is unaffected.

## Reachability
`ArtMethodImporter::Sniff` routes versions 1/2/3 to this tokenizer (only 0x0004/0x0005/0x00f4/0x00f5 go to the hardened v2 tokenizer), so a crafted ART method trace reaches this path. CWE-125.


src/trace_processor/importers/art_method/art_method_tokenizer.cc
```diff
@@ -213,6 +213,18 @@ base::Status ArtMethodTokenizer::ParseRecord(uint32_t tid,
   }
 
   uint32_t methodid_action = ToInt(record.slice_off(0, 4));
+  // A dual clock reads the thread ts_delta at body offset [8, 12). For v1/v2
```
LalitMaganti: Please remove this comment. It's very self explanatory, the comment is unnecessary

evilgensec: Done, removed.

LalitMaganti (approved):

sashwinbalaji (approved):

PR #7509, opened 2026-09-21
tp: compute stdlib tree totals with PerfettoSQL pipelines

LalitMaganti:
Replace the _graph_aggregating_scan! based subtree totals in the stdlib
with TREE ACCUMULATE UP pipelines:
 - android_heap_graph_class_summary_tree
 - _heap_graph_object_tree_aggregation
 - android_heap_profile_summary_tree
 - cpu_profiling_summary_tree
 - linux_perf_samples_summary_tree
 - appleos_instruments_samples_summary_tree

Public tables keep their columns and are ordered by id, as a projection
over the pipeline's table. The sample summary trees accumulate straight
over _callstacks_for_callsites!, which removes their raw callstack
tables and parent_id indexes. _callstacks_self_to_cumulative! has no
users left and is deleted; callstacks.stack_profile stops including
graphs.scan as a result, so the UI query which relied on that includes
it directly.

_heap_graph_object_tree_aggregation also becomes correct: the scan summed
a node's children but never added the node itself, so every non-leaf
object reported a total which left out its own size and count.

On a 60MB system_server heap dump, median of 7 interleaved runs:
 - class summary tree: 21011ms -> 147ms
 - object tree aggregation: 2988ms -> 750ms
On smaller traces: heap profile summary tree 19.6ms -> 2.5ms and
cpu_profiling_summary_tree 203ms -> 79ms.


src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/class_summary_tree.sql
```diff
@@ -15,45 +15,11 @@
 
 INCLUDE PERFETTO MODULE android.memory.heap_graph.class_tree;
 
-INCLUDE PERFETTO MODULE graphs.scan;
```
sashwinbalaji, line 18: can we also not cleanup the scan.sql file removing _graph_aggregating_scan

sashwinbalaji (approved):
nice!!!

src/trace_processor/perfetto_sql/stdlib/android/memory/heap_graph/class_summary_tree.sql
LalitMaganti: I'd like to do a global cleanup when we get close to the end.

PR #7519, opened 2026-09-22
ui: build the Wattson task summary once per selection

wusamuel6:
WattsonProcessSelectionAggregator and WattsonCpuPackageSelectionAggregator both read from wattson_plugin_thread_summary and previously relied on WattsonThreadSelectionAggregator's prepareData() running first to create it. On top of that implicit ordering dependency, switching between the thread, process, and package tabs within the same selection would re-run the full _wattson_threads_aggregation pipeline every time.

Extract buildTaskSummary() into task_summary.ts that keys its cached Promise on (engine, start, end, selected CPUs, selected utids) and have all three aggregators await it. Concurrent prepareData() calls for the same selection now share a single in-flight query, and subsequent tab switches reuse the materialized wattson_plugin_thread_summary table. Also partition _wattson_ui_selected_tasks by cpu in SPAN_JOIN so overlapping intervals across CPUs do not drop threads from the summary.

Bug: 454944449


ui/src/plugins/org.kernel.Wattson/task_summary.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+    if (trackInfo?.tags?.kinds?.includes(WATTSON_THREAD_TRACK_KIND)) {
+      exists(trackInfo.tags.utid) && utids.push(trackInfo.tags.utid);
+    }
+  }
+  return {cpus, utids};
+}
+
+// The most recent build, keyed by the engine and selection it was built for.
+let inFlight:
```
stevegolton: Could you avoid the global variable here by making the memo a class that you recreate in the onTraceLoad() function, and pass into the various area selection aggregators?

That way we'd avoid the global variable and avoid having to key on engine, as the cache will be recreated every trace.

stevegolton (changes requested):

ui/src/plugins/org.kernel.Wattson/task_summary.ts
```diff
@@ -0,0 +1,154 @@
...
+      [...utids].sort((a, b) => a - b),
+    ]);
+
+    if (this.inFlight?.key === key) {
+      return this.inFlight.done;
+    }
+
+    const prev = this.inFlight?.done ?? Promise.resolve();
+    const done = prev
+      .catch(() => {})
```
stevegolton, line 83: What is this catch for?

ui/src/plugins/org.kernel.Wattson/task_summary.ts
```diff
@@ -0,0 +1,154 @@
...
+      .then(() => this.createTaskSummary(area, cpus, utids))
+      .catch((e) => {
```
stevegolton, line 85: Under what circumstances might `createTaskSummary` fail? Would it not be better to crash loudly?

ui/src/plugins/org.kernel.Wattson/process_aggregator.ts
```diff
@@ -13,38 +13,36 @@
...
+
   probe(area: AreaSelection) {
-    const selectedCpus: number[] = [];
-    for (const trackInfo of area.tracks) {
-      trackInfo?.tags?.kinds?.includes(CPU_SLICE_TRACK_KIND) &&
-        exists(trackInfo.tags.cpu) &&
-        selectedCpus.push(trackInfo.tags.cpu);
-    }
-    if (selectedCpus.length === 0) return undefined;
+    const {cpus} = getWattsonTrackSelection(area);
```
stevegolton, line 39: Nit: It seems like we call this twice on every probe. Could we push the result of getWattsonTrackSelection() into taskSummary so it doesn't have to call it internally? This is kind of the point of having the probe step return the prepareData step so that we can capture any intermediate objects and avoid having to recalculate them.

stevegolton (changes requested):
Looks much better re the cache rework, but I just had some questions about the error handling.

ui/src/plugins/org.kernel.Wattson/task_summary.ts
wusamuel6: Not needed

ui/src/plugins/org.kernel.Wattson/task_summary.ts
wusamuel6: Agree, not needed

